PR #998, opened 2015-12-22
Removes the un-used settings view controller

orta:
It was Laura's first project, but we ended up having to work on something else and never got back to it.

PR #1000, opened 2015-12-22
@alloy => Convert App back to Frameworks

orta:
Initial work on converting us back to frameworks. Relies on #998 & #99.


orta:
I've been wondering about whether this will cause too much problems now that we're using testflight for betas and not enterprise hockey builds. Will ship a test flight beta once I've got a working build.

orta:
Currently blocked by https://github.com/CocoaPods/CocoaPods/issues/4723 - will see where this goes

orta:
OK, hrm, looks like some master merging needs doing.

orta:
This is all good.

alloy:
Have you by any chance tested whether removing the various provisioning profiles from your device makes it launch fast, as mentioned in the startup time ticket?

orta:
I'm not seeing slow-down, either in dev or on deploys (in testflight. ) The issue showed up on enterprise builds. Could be cause I have a relatively new phone, so no profiles around.

alloy:
Well that’s good then :+1:

PR #1003, opened 2015-12-27
Use new podfile target inheritance

segiddins:


Podfile
```diff
@@ -3,6 +3,8 @@ source 'https://github.com/CocoaPods/Specs.git'
 
 platform :ios, '8.0'
 
+install! 'cocoapods', :deterministic_uuids => false
```
orta: Should this include the `inhibit_all_warnings` below`?

segiddins: No, since inhibiting warnings is per target and the installation options are per-podfile.

orta:
Thanks @segiddins  - I've merged this in to #1000 ( this predates the `:head` remova, so I did that. ) Closing, but you get your commits in that PR.

PR #1019, opened 2016-01-08
[CI] Remove references to ENV vars in Podfile / circle.yml

orta:
Betas would get OSS keys, they would work... Betas get deployed to the app store. They don't work for a reviewer. Phew.

This also adds a check that we're not deploying those keys through fast lane.

Richer [context in slack](https://artsy.slack.com/archives/eigen/p1452257377000467)

PR #1021, opened 2016-01-11
Use the one true format for CHANGELOGs

orta:
Note! this PR goes to #1019 - this should be merged first.


Dangerfile
```diff
@@ -1,6 +1,6 @@
 warn("PR is classed as Work in Progress") if pr_title.include? "[WIP]"
 
-if lines_of_code > 50 && !files_modified.include?("docs/BETA_CHANGELOG.md")
+if lines_of_code > 50 && !files_modified.include?("CHANGELOG.yml")
```
ashfurrow: Such a shouty name! I'm pretty sure I'm losing the war on this front, though.

ashfurrow:
:+1:

Dangerfile
orta: DEF

PR #1024, opened 2016-01-13
Support latest facebook SDK

orta:
This will have to be merged into master too.
See: https://developers.facebook.com/docs/ios/ios9


orta:
Fail is an old build of danger, it was fixed in a later commit.

ashfurrow:
Looks :+1:

orta:
good point, this PR has shipped

PR #1039, opened 2016-01-18
Use match for all the certificates

orta:
OK let's look at before:

screen shot 2016-01-18 at 10 14 19

There's a bit of everything in there.

After:
screen shot 2016-01-18 at 10 17 09

All of the provisioning profiles are set, but the certs are left automatic. Which leaves us free to use our own certs to sign against.

All work is done in the project: 
screen shot 2016-01-18 at 10 17 14

No suprises


orta:
My automatic resolution: 
screen shot 2016-01-18 at 10 21 30
screen shot 2016-01-18 at 10 21 19

alloy:
Like I said, there shouldn’t be a need for a shared dev provisioning profile and thus no need to have a shared cert for development. I think it only makes things more difficult.

Can you try this:

1. set the dev provisioning profile to automatic
2. change your cert resolution to ‘iPhone Developer’

alloy:
(And to be clear, more difficult for us, but especially for OSS.)

orta:
It makes me a bit nervous ( you don't get to see the resolution of certs/prov profiles ) however, you're totally correct - works and is running :+1: on my device.

screen shot 2016-01-18 at 10 38 29

Artsy.xcodeproj/project.pbxproj
```diff
@@ -5032,7 +5032,7 @@
 				CLANG_ENABLE_OBJC_ARC = YES;
 				CLANG_WARN_EMPTY_BODY = YES;
 				CLANG_WARN__DUPLICATE_METHOD_MATCH = YES;
-				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
+				CODE_SIGN_IDENTITY = "iPhone Developer";
```
alloy: It only needs to be signed for device builds, though, so this can be reverted.

Artsy.xcodeproj/project.pbxproj
```diff
@@ -5486,7 +5485,7 @@
 				CLANG_ENABLE_OBJC_ARC = YES;
 				CLANG_WARN_EMPTY_BODY = YES;
 				CLANG_WARN__DUPLICATE_METHOD_MATCH = YES;
-				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Distribution";
+				CODE_SIGN_IDENTITY = "iPhone Distribution";
```
alloy: While this is for the store, there’s no need to sign for simulator, just in case you ever needed to test it on the sim.

orta:
screen shot 2016-01-18 at 10 43 10
better?

alloy:
So, I think that basically none of the changes to the project are really necessary, right?

orta:
haha - looking back, only one thing, which we can live without ( ensuring demo gets built with app store ) happy to close

lololol

alloy:
:smile:

alloy:
I get the nervous part, btw, the :sparkles: part of it is probably a reason why this aspect is so hard for most users.

PR #1040, opened 2016-01-18
Use size classes instead of isPad now that we're post iOS7

orta:
In terms of trying to keep the seams smaller between swift/objc - we can remove the isPad check with a size class check. This makes life much easier if we ever decide to do real work surrounding the split screen stuff that we do support already.

### DIFF
--- a/Artsy/Resources/Images.xcassets/AppIcon.appiconset/Contents.json
+++ b/Artsy/Resources/Images.xcassets/AppIcon.appiconset/Contents.json
@@ -65,6 +65,11 @@
       "idiom" : "ipad",
       "filename" : "Icon-76@2x.png",
       "scale" : "2x"
+    },
+    {
+      "idiom" : "ipad",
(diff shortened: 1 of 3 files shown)


Artsy/Resources/Images.xcassets/AppIcon.appiconset/Contents.json
```diff
@@ -65,6 +65,11 @@
       "idiom" : "ipad",
       "filename" : "Icon-76@2x.png",
       "scale" : "2x"
+    },
+    {
+      "idiom" : "ipad",
+      "size" : "83.5x83.5",
+      "scale" : "2x"
```
orta: every commit I've done - Xcode has been begging me to let this go in, so I figured I'd let it happen.

ashfurrow:
Looks good, re-buidling CI from the mysterious intermittent danger failure.

PR #354, opened 2015-04-03
@orta => X callback url

1aurabrown:
This adds a pod for handling x-callback-url according to the protocol's conventions. right now, it just handles opening a url as a view and pushing it, which is nothing new. next i'll wire up the back button to take you back to whatever place you came from.

the incoming url would be structured like this:
`artsy-x-callback-url://x-callback-url/open?url=some/artsy/url`

this is how you'er supposed to do it according to the x-callback-url specs. `open` is an action that takes some parameters, in the case, `url`. this should also work with a path.

PR #1046, opened 2016-01-20
Remove X-Callback support from our app - introduced in #354

orta:
Now that Eigen controls the artsy.net domain system wide, there's no need to faff with x-callback stuff. Everyone will just treat URLs as something they can route rather than making special cases.


orta:
Hm, the fail here is Danger. Tests are :+1: 
Gonna jump into Danger now.

orta:
I've fixed Danger - https://github.com/KrauseFx/danger/pull/35

ashfurrow:
Looks good, will merge on :green_apple:

orta:
ok

PR #1054, opened 2016-01-22
Updates Podfile.lock

ashfurrow:
Looks like it was accidentally left uncommited from https://github.com/artsy/eigen/pull/1046/


alloy:
Nuhuh, stealing this from @orta.

PR #1055, opened 2016-01-22
[Push Notification] Fix blank view issue with ‘works for you’.

alloy:
The previous code was presenting the top menu VC during routing, whereas it should’ve only returned it. This was leading to a blank web view being shown when tapping a “works for you” push notification.

Why the blank web view is being shown I’m not certain of, but as the initial cause for it being shown is faulty logic, I didn’t spend more time chasing that.


ashfurrow:
Looks great! :100:

PR #1056, opened 2016-01-22
[iPad] Fix crash when presenting an alert regarding external URL.

alloy:
https://rink.hockeyapp.net/manage/apps/37029/app_versions/49/crash_reasons/102830998?type=overview

simulator screen shot 22 jan 2016 23 05 52

simulator screen shot 22 jan 2016 23 06 55

----

If you really do prefer a sheet on iPhone, we’ll have to add a device conditional, because a sheet in a popover on iPad has to point to a location and there really is none to point to:

simulator screen shot 22 jan 2016 23 03 53


orta:
Modal is :+1: for me. Good to merge on :green_apple:

PR #1058, opened 2016-01-22
Use AROptions to make native auctionview an option 

orta:
Before
screen shot 2016-01-22 at 22 41 56

In-between
screen shot 2016-01-22 at 22 42 11

After
screen shot 2016-01-22 at 22 42 07


orta:
@ashfurrow use `ARTopMenuViewController+DeveloperExtras.m` to make sure that this default is on for you locally

ashfurrow:
Makes sense, thanks champ :+1: 

Looks good, will merge on :green_apple:

PR #1066, opened 2016-01-25
Improve how we get feedback

orta:
Adds some default text, and lets the analytics data be optional and show as text so we can click in the website and look.


orta:
screen shot 2016-01-25 at 16 06 30

Issue #985, opened 2015-12-14
CI locks?


orta:
It looks like it stopped here: https://circle-artifacts.com/gh/artsy/eigen/1078/artifacts/0/tmp/circle-artifacts.KgPk57x/xcode_test_raw.log

```
Test Case '-[ARTopMenuNavigationDataSourceSpec test_notifications__resets_the_app_icon_badge_if_there_are_0_notifications]' started.
2015-12-14 14:43:05.928 Artsy[6054:11667] Rewriting /shows as https://m.artsy.net/shows
```

ashfurrow:
Saw this today, can replicate it locally.

ashfurrow:
See this still, will put some time into it when I get a chance.

orta:
fixed by #1068

PR #1068, opened 2016-01-26
Fixes deadlocks.

ashfurrow:
This is a possible fix for #985. According to [other SDWebImage users](https://github.com/rs/SDWebImage/issues/507), there's a problem with deadlocking which is related to a race condition checking if a request is cancelled/finished. 

This solution cancels all downloads whenever we stub the user (which is where we've been deadlocking). Fairly straightforward – ran it a bunch of times locally and it was all good.


orta:
Oh nice :+1:

PR #1073, opened 2016-01-27
Get Gene pages working again

orta:
The API calls for filter come in as a dictionary, with the Artworks being inside `hits` rather than at the root of the array. 

This PR will mean that we need to deploy another build of our current beta 2.3.5.


ashfurrow:
:+1:

PR #1075, opened 2016-01-27
Finishes price range selection 📱

### LINKED ISSUE #971: Refine Tool
The refine tool will act as a catch all for both filtering and sorting that we can hopefully use wherever a large collection of works is displayed in the app. For now though, let's limit the scope to the auction sale page. It will act as a modal on top of the sale page. By default the works will show up in the "default" sort order (either lot order or artist a-z) displayed in a masonry layout. When a particular sort item is selected from the refine modal, this will switch the view to display the works in a list. We use this same logic in Eidolon and on the web. I'm going to assume we don't have the backend to support price filtering at the moment but I plan on putting a request into the auctions team to support it. I imagine for a V1 we can plan on excluding the price filter altogether. 

image

image

ashfurrow:
Fixes #971.

On iPhone:

screen shot 2016-01-27 at 3 59 55 pm

And iPad, shown working:


ashfurrow:
I've removed one test because it is inconsistent with the use of the view controller. It was testing what happened when a user changes a setting (it enables the clear/apply buttons), but I can't think of a way to simulate user interaction in a realistic way so the test is actually _useful_.

ashfurrow:
Never mind, got a (bit hacky) way implemented, adding it back (with a new name).

ArtsyMobile:
&nbsp; | 1 Error
    ------------- | ------------
        :no_entry_sign: | No CHANGELOG changes made
    

  

  
    &nbsp; | 1 Warning
    ------------- | ------------
        :warning: | Needs testing on a Phone if change is non-trivial
    

  

  


<p align="right" meta="generated_by_danger">
</p>

ashfurrow:
No changelog entry as it already exists for this change.

Artsy/View_Controllers/Auction/AuctionRefineViewController+Private.swift
```diff
@@ -83,6 +104,70 @@ private extension AuctionRefineViewController {
...
+                return
+            }
+            labelContainer.addSubview(maxLabel)
+
+            maxLabel.alignCenterYWithView(labelContainer, predicate: "0") // Center vertically in container.
+            maxLabel.alignCenterXWithView(slider.rightThumbView, predicate: "0").forEach(setConstraintPriority(.StayCenteredOverThumb))
+            maxLabel.alignAttribute(.Trailing, toAttribute: .Trailing, ofView: labelContainer, predicate: "<= 0").forEach(setConstraintPriority(.StayWithinFrame))
+
+            // Make sure they don't touch! Shouldn't be necessary since they'll be 10% appart, but this is "just in case" make sure the labels never overlap.
+            minLabel.constrainTrailingSpaceToView(maxLabel, predicate: "<= -10").forEach(setConstraintPriority(.DoNotOverlap))
```
orta: `appart`->`apart`

orta: instead of `forEach(setConstraintPriority(.DoNotOverlap))` you could use the `"<= -10@750"` or whatever?

ashfurrow: :grimacing: I forgot that I could do that.

I can change to that, honestly I kind of like this because it centralizes the priorities in an enum, but we could find a middle ground. Or whatever, I don't feel strongly.

orta: me neither for now, happy to let this settle out, but it's a global UIView technique that's unique to just this view controller, it should be everywhere or nowhere

ashfurrow: Fair enough!

orta:
:+1:

ashfurrow:
:bow:

Issue #972, opened 2015-12-08
List vs Masonry grid formatting and metadata styling


katarinabatina:
View changes occur when using the refine tool
![image](https://cloud.githubusercontent.com/assets/6482926/11725502/0b0ec48a-9f48-11e5-85a4-92f5869a2680.png)

ashfurrow:
*Estimate*: 2 days

PR #1097, opened 2016-02-01
Registration Status

### LINKED ISSUE #975: Registration status of user
We are moving away from requiring a user to register on a per sale basis. Instead we want to move to a simpler model where an artsy user can bid if they are qualified. 

That means, for users who have bid with us before will automatically be Approve to Bid, those who haven't will be required to register if they want to bid. When an auction places more strict vetting i.e. asks users to opt in to separate terms and conditions, asks user to send along a bank statement, etc. the user will be again asked to register. We don't anticipate more strict vetting in the short term but we know this will be a feature that partners will want so I figure it's good to build for the case now. 

Cases outlined:

1. Reg required
2. Reg pending (we shouldn't see this for some time, this would only be seen in the scenario that the user has given additional info during registration that needs to be reviewed)
3. Approved to bid
4. Outdated payment info (expired cc) on file

image

Measurements for Information Area

image

ashfurrow:
This is a WIP to make the register button work appropriately. It's a WIP because I still need to do:

- [x] title view layout
- [x] tests for title view
- [x] tests for network models (see below)

This PR makes significant changes to the networking done in the auctions' network models, and I'd like to explain my reasoning. I've broken each request-type (fetch the sale, the sale artworks, and the registration status) into their own network models. They can all fetch exactly one thing, and then store that thing locally. That's it. I want to keep these super-lean and follow the single-responsibility principle.

I think that's easy to get behind, but I have each fetch function return a `Signal`, and then the `AuctionNetworkModel` just calls the others' fetch methods and assembles the results. It's doing this with Interstellar, so I'm cautious about going too outside eigen's reference frame. 

If the general idea of isolating network models sounds good, I'll add some tests. If you have feedback on the use of Interstellar, please let me know :+1: 

Fixes #975.


Artsy/View_Controllers/Auction/AuctionNetworkModel.swift
```diff
@@ -1,78 +1,55 @@
...
+                return self.saleNetworkModel.fetchSale().subscribe(callback)
+            }
+            .merge(saleArtworksNetworkModel.fetchSaleArtworks())
+            .map { tuple in
+                // Tuple has the Sale and [SaleArtwork] from previous network requests.
+                return SaleViewModel(sale: tuple.0, saleArtworks: tuple.1)
+            }
+            .next { saleViewModel in
+                // Store the SaleViewModel
+                self.saleViewModel = saleViewModel
```
ashfurrow: These group of lines is what I'm most cautious about introducing into eigen. I'm trying to keep from introducing esoteric FP/FRP use, and I've tried to comment this to make it clear what it's doing and why. I'd like feedback to know if this is adequately explained, and if it's just too out there for eigen.

Artsy/Categories/InterstellarExtensions.swift
```diff
@@ -0,0 +1,21 @@
...
+func passOnFailure<T>(callback: Result<T> -> Void) -> (NSError!) -> Void {
+    return { error in
+        callback(.Error(error))
+    }
+}
+
+extension Signal {
+    /// Injects a signal into the receiver, but ignores the injected signals values.
+    /// Useful for injecting side-effects into a signal chain.
+    func inject<OtherType>(signal: Signal<OtherType>) -> Signal<T> {
```
orta: could this be called `then` or `tap` rather than introduce new terminology?

Artsy/Networking/API_Modules/ArtsyAPI+CurrentUserFunctions.m
```diff
@@ -13,5 +14,23 @@ + (void)updateCurrentUserProperty:(NSString *)property toValue:(id)value success
...
+    }
+
+    NSURLRequest *request = [ARRouter biddersRequestForSale:saleID];
+    [self getRequest:request parseIntoAnArrayOfClass:[Bidder class] success:^(id bidders) {
+        ArtsyAPISaleRegistrationStatus status = ArtsyAPISaleRegistrationStatusNotRegistered;
+
+        if ([bidders count] > 0) {
+            status = ArtsyAPISaleRegistrationStatusRegistered;
+        }
+
```
orta: One liner?
`ArtsyAPISaleRegistrationStatus status = bidders.count > 0 ? ArtsyAPISaleRegistrationStatusRegistered : ArtsyAPISaleRegistrationStatusNotRegistered`

Artsy/View_Controllers/Auction/AuctionSaleArtworksNetworkModel.swift
```diff
@@ -0,0 +1,63 @@
...
+        switch result {
+        case .Success(let saleArtworks):
+            callback(.Success(saleArtworks))
+        case .Error(let error):
+            callback(.Error(error))
+        }
+    }
+}
+
+/// Recursively calls itself with page+1 until the count of the returned array is < pageSize.
```
orta: non-ideal aspect of this is that we have to wait for all data before showing something, we can probably live with it here, but it's a bit of an Eigen anti-pattern

Artsy/Views/Auction/SaleViewModel.swift
```diff
@@ -27,6 +27,10 @@ extension SaleViewModel {
         return saleArtworks.count
     }
 
+    var displayName: String {
+        return sale.name
```
orta: think I've also exposed this, think I named it differently, happy with whichever

Artsy/View_Controllers/Auction/AuctionNetworkModel.swift
alloy: It’s too much imo. I’d prefer code that does not require the comments. The previous switch with a success and error case branch is much clearer in that regard.

alloy: (Not saying that such comments aren’t good, I’m saying that in this case it’s a signal that it went overboard.)

Artsy/Categories/InterstellarExtensions.swift
ashfurrow: Yeah, I struggled to name it. `tap` is a good idea.

Artsy/Networking/API_Modules/ArtsyAPI+CurrentUserFunctions.m
ashfurrow: Wasn't sure if I'd trigger Space Commander for having lines so long :speak_no_evil:

Artsy/View_Controllers/Auction/AuctionSaleArtworksNetworkModel.swift
ashfurrow: This is true, but it greatly simplifies filtering by price and re-ordering by other criteria. The networking is done once, up front. 

A way to make this better would be to increase the page size we fetch. It's currently ten, smaller than most auctions. The max is a hundred. Using that would greatly reduce the number of roundtrips, and most auctions would only need one. Thoughts?

Artsy/Views/Auction/SaleViewModel.swift
ashfurrow: I'll rebase and see. Thanks for the head's up!

Artsy/View_Controllers/Auction/AuctionNetworkModel.swift
ashfurrow: Yup yup yup. Going to go back to a `flatMap`-only style. It's be a little more verbose, but I think that's preferable to being arcane.

ashfurrow:
OK, merged latest in from master. Going to continue to work on the three remaining TODO items.

ArtsyMobile:
&nbsp; | 1 Error
    ------------- | ------------
        :no_entry_sign: | No CHANGELOG changes made
    

  

  
    &nbsp; | 2 Warnings
    ------------- | ------------
        :warning: | PR is classed as Work in Progress
        :warning: | Needs testing on a Phone if change is non-trivial
    

  

  


<p align="right" meta="generated_by_danger">
</p>

Artsy/Networking/API_Modules/ArtsyAPI+CurrentUserFunctions.m
orta: don't think space commander cares about line length - it's not rubocop :cop:

ashfurrow:
Alrighty! Managed to get _some_ coding in today. Likely won't get a chance to look at this until Friday, there's a showstopping Kiosk bug that needs to be fixed before the Friday Mocktion. 

This "works" but is untested. Oh, and it doesn't actually work:

screen shot 2016-02-03 at 6 35 14 pm

Haven't had a chance to really dig into it. It looks like the header stack view is getting covered up by the sticky header. @orta if you have a chance, I'd love some suggestions on this.

ashfurrow:
Oh, and there's a missing "info" button asset I need to add >.<

orta:
> . It looks like the header stack view is getting covered up by the sticky header. @orta if you have a chance, I'd love some suggestions on this. I tried rotating to force a re-layout but no dice.

You need to make sure the stack is fully re-laid out before triggering to the collection view that the height has changed

ashfurrow:
@orta the problem was on my end, forgot the `-` on a bottom-edge constraint.

I've decided to punt on the view tests, because the other constituent views aren't tested either, and I think the tests more properly belong in the view controller snapshot tests in #972  (instead of testing each view individually + aggregated in the view controller anyway, no need to redundantly test, but this might warrant a bigger discussion).

iPhone:

screen shot 2016-02-05 at 5 13 11 pm

iPad: 

screen shot 2016-02-05 at 5 14 44 pm

ashfurrow:
CI was failing comparing two (ObjC) enums in my Swift unit tests. I've updated Circle's config to Xcode 7.1 to try and fix.

ashfurrow:
Danger is complaining of a missing changelog entry, but there is one there... Maybe we're behind on versions.

orta:
Will look on Monday, stuck with a phone for a bit

orta:
Danger thing is known, been wondering I'd the commit range is wrong

ashfurrow:
Cool cool – commit range sounds plausible, this is a pretty long-lived PR.

orta:
I disagree with respect to not doing the tests for the view, ideally we want to be creating re-usable UI components, testing at the view level encourages that. Moving this up to VC level means we get a lot of "AuctionXXXView" views which are now only really useful within one context, and no-one will think to re-use them.

Artsy/Networking/API_Modules/ArtsyAPI+Sales.h
```diff
@@ -18,7 +18,7 @@
               success:(void (^)(Sale *sale))success
               failure:(void (^)(NSError *error))failure;
 
-+ (void)getSaleArtworksWithSale:(Sale *)sale
++ (void)getSaleArtworksWithSale:(NSString *)saleID
```
orta: :+1:

Artsy/Views/Auction/AuctionTitleView.swift
```diff
@@ -18,11 +39,157 @@ class AuctionTitleView: UIView {
...
+            $0.font = UIFont.serifFontWithSize(16)
+            $0.textColor = .artsyHeavyGrey()
+        }
+        container.addSubview(registrationLabel)
+
+        // Centre both horizontally
+        registerButton.alignCenterXWithView(container, predicate: "0")
+        registrationLabel.alignCenterXWithView(container, predicate: "0")
+
+        // Staack the register button on top of the label, and constrain them to the container.
```
orta: minor: typo

Artsy_Tests/View_Controller_Tests/Auction/AuctionNetworkModelTests.swift
```diff
@@ -0,0 +1,162 @@
...
+
+    init(result: Result<ArtsyAPISaleRegistrationStatus>) {
+        self.result = result
+    }
+    
+    func fetchRegistrationStatus(saleID: String, callback: Result<ArtsyAPISaleRegistrationStatus> -> Void) {
+        called = true
+        callback(result)
+    }
+}
```
orta: cool tests bro

Artsy_Tests/View_Controller_Tests/Auction/AuctionReglstrationStatusNetworkModelTests.swift
```diff
@@ -0,0 +1,90 @@
...
+import OHHTTPStubs
+@testable
+import Artsy
+
+class AuctionRegistrationStatusNetworkModelSpec: QuickSpec {
+    override func spec() {
+        let bidderJSON: NSArray = [["id": "bidder", "sale": ["id": "sale"]]]
+
+        afterEach {
+            OHHTTPStubs.removeAllStubs()
```
orta: I think this should be done for you

Artsy/Categories/InterstellarExtensions.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+import Interstellar
+
+/// Convenience function for invoking callback with the given error.
+func passOnFailure<T>(callback: Result<T> -> Void) -> (NSError!) -> Void {
```
orta: this might make more sense as `raiseError` ?

orta:
Looks good, only `this might make more sense as raiseError ?` is a blocker :+1:

Artsy/Categories/InterstellarExtensions.swift
ashfurrow: Hmm, interesting. It's not strictly raising/throwing an error, it's just passing it into the callback block. But I agree that "passOnFailure" doesn't really convey that either.

Artsy_Tests/View_Controller_Tests/Auction/AuctionReglstrationStatusNetworkModelTests.swift
ashfurrow: 🎉

ashfurrow:
> ideally we want to be creating re-usable UI components 

Agreed, will add some tests. Probably will involve moving testing traits to Forgeries.

ashfurrow:
OK, so we're using a branch of Forgeries pending https://github.com/ashfurrow/Forgeries/pull/15 getting merged. I added tests for the title view as requested, using both iPad and iPhone. I know we have something like this already for device-independent snapshotting but that's mainly around `UIDevice`, I think? So it may be worth thinking of a better abstraction for trait classes. But not right now >.<

Artsy_Tests/View_Controller_Tests/Auction/AuctionNetworkModelTests.swift
ashfurrow: protocol-oriented-blahblahblah

ashfurrow:
OK! Got that other PR merged, and moved this to Forgeries 0.6.1.

> "I think this is good for final review", he said trepidatiously.

ashfurrow:
But in the past tense, yay!

CHANGELOG.yml
```diff
@@ -14,8 +14,10 @@ releases:
     - Use Hockey for feedback, take a screenshot in app or use the admin menu to trigger - orta
     - View Controllers can now add sticky headers to collections of Artworks/Models - orta
     - Use Hockey for feedback, take a screenshot in app or use the admin menu to trigger the feedback mechanism - orta
+    - Registration status on auction page - ash
     - Removes rotation support on refine auction listings view for iPhone - ash
     - Remove prefix header from app target. - alloy
+    - Additional work on the auction title view - ash
```
orta: these changes have gone into an older release

orta: fixed in https://github.com/artsy/eigen/pull/1129

PR #1126, opened 2016-02-08
Add support for showing the bell tapped analytics events

orta:
Allow us to differentiate between push / taps WRT the bell.

### DIFF
--- a/Artsy/App/ARAnalyticsConstants.m
+++ b/Artsy/App/ARAnalyticsConstants.m
@@ -122,4 +122,4 @@ NSString *const ARAdjustSentArtworkInquiry = @"duxtlx";
 NSString *const ARAdjustFirstUserInstall = @"kju96h";
 NSString *const ARAdjustTappedBidButton = @"w5e24e";
 
-
+NSInteger const ARNavButtonNotificationsTag = (NSInteger)&ARNavButtonNotificationsTag;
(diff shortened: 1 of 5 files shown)


Artsy/App/ARAnalyticsConstants.m
```diff
@@ -122,4 +122,4 @@
 NSString *const ARAdjustFirstUserInstall = @"kju96h";
 NSString *const ARAdjustTappedBidButton = @"w5e24e";
 
-
+NSInteger const ARNavButtonNotificationsTag = 63634646;
```
alloy: Since a `NSInteger` should have the same size as a pointer, maybe we should standardise on something like the following, instead of making numbers up? (Assuming that that’s what you did ;))

```objc
NSInteger const ARNavButtonNotificationsTag = (NSInteger)&ARNavButtonNotificationsTag;
```

https://developer.apple.com/library/ios/documentation/General/Conceptual/CocoaTouch64BitGuide/Major64-BitChanges/Major64-BitChanges.html#//apple_ref/doc/uid/TP40013501-CH2-CHDGGBDA

orta: Works for me, yeah, will amend

orta:
Done

alloy:
Ace. Merging on green.

orta:
Failure was danger, I should start finding a way to deal with that properly, for now, this is :+1:

PR #1130, opened 2016-02-10
Use the latest Danger / pilot

orta:
This uses the new `--base` option which lets you specify a specific commitish to base against.


alloy:
So this always makes it compare the PR branch against master?

orta:
Yes

ashfurrow:
👍

PR #1139, opened 2016-02-11
Fix the ARSearchFieldButton tap delegate

orta:
ARSearchFieldButton accidentally used `self.delegate` before it existed, now it creates the gesture on `setDelegate` - :phone:


alloy:
Not sure how those snapshots were affected, but they seem related, at least in the ‘fair’ context sense.

orta:
Ah, nah, these have been flakey for a bit, I bet if I rebuild they'll go away.

alloy:
Ahhh, those font issues. Maybe the tests get ran without the custom fonts being registered?

orta:
OK, I have a fix attempt PR coming in

PR #1141, opened 2016-02-11
Attempt to cache fonts before loading the testing environment #trivial

orta:
I can't guarantee this will help, so if this fails with those errors, we can def not merge. Will run theses tests 3-4 times, if it's always green I'll merge.


orta:
This has succeeded 3 times in a row so far, doing it one more time.

alloy:
I think this was also a major reason for focussed tests to fail.

orta:
OK, 4th time, I'm happy.

alloy:
Ace :+1:

PR #1143, opened 2016-02-11
Make the hero units the same size as they are on Martsy

orta:
We have smaller hero units, I'm not sure if the spec changed since we first built out hero units, but admins expected to be able to add more information than we could comfortably show. This brings us back inline with martsy.


Artsy/View_Controllers/Core/ARHeroUnitViewController.m
```diff
@@ -31,7 +31,7 @@ @implementation ARHeroUnitViewController
 
 + (CGFloat)heroUnitHeight
 {
-    return [UIDevice isPad] ? 380 : 232;
+    return [UIDevice isPad] ? 380 : 252;
```
alloy: Is it working as expected on iPad?

orta:
Before:
screen shot 2016-02-11 at 12 40 20

After:
screen shot 2016-02-11 at 12 47 55

Artsy/View_Controllers/Core/ARHeroUnitViewController.m
orta: Spacing wise? I think so, will give it a quick look

orta: Yeah, they're gonna be fine.

alloy:
Groovy

PR #1146, opened 2016-02-11
use build trains for the tag

orta:
So previously we were using a fastlane action to grab the last build number, then increment that to do a tag. This doesn't work, cause out build number it `2016.01.30` or whatever, so it would always return `2017`. Now we count the builds via spaceship.


orta:
screen shot 2016-02-11 at 18 57 28

works as expected now

alloy:
:+1:

ArtsyMobile:
&nbsp; | 1 Error
    ------------- | ------------
        :no_entry_sign: | Developer Specific file shouldn't be changed
    

  

  

  


</p>

Artsy/App/ARAppStatus.m
```diff
@@ -1,11 +1,19 @@
...
 
+
 @implementation ARAppStatus
 
 // http://stackoverflow.com/questions/26081543/how-to-tell-at-runtime-whether-an-ios-app-is-running-through-a-testflight-beta-i
 
++ (BOOL)isBetaDevOrAdmin
+{
+    NSString *email = [User currentUser].email;
+    BOOL isArtsyEmail = [email containsString:@"@artsymail."] || [email containsString:@"@artsy.net"];
```
alloy: Why is this not checking for `artsymail.com` ? Now I can register a `artsymail.cheapo` domain and get access?

Also, I think `hasSuffix:` is slightly better than `containsString:`.

Artsy/App/ARAppStatus.m
```diff
@@ -1,11 +1,19 @@
...
+    return [self isBetaOrDev] || isArtsyEmail;
```
alloy: Nitpick, seeing as you’re already performing the work to check for the email above, move that first into the conditional so that we don’t need to perform the `isBetaOrDev` work if it is an admin.

alloy:
ObjC code looks good to me beside those logic issues.

So to be clear, the goal is that we (admins) will get the dev VC in App Store builds as well?

orta:
Yeah, this makes it easy to see useful details ( git sha, push notification ID, dev A/B options, etc )

orta:
The builds will always fail now, this is cause I have a merge commit - meaning danger takes the entire merge also into account. 
I wonder if there's a way to say I want the merge between these two branches, excluding things that already exist in the firs branch?

alloy:
I’m not 100% sure I understand the scenario (replicatable git commands for me to locally see what this is about and how Danger knows what code to check would be helpful), but if I do get it right, maybe you just want the real diff, e.g. `git diff branch1 branch2` ?

orta:
so, here's the code: https://github.com/danger/danger/blob/master/lib/danger/scm_source/git_repo.rb#L9-L12 - which I think should be the same as what you're talking about - but I need to do some danger PR merging to make a lot of this right

alloy:
Yeah, looks right, I’m not sure then what happens with a merge, does this happen with a rebase as well?

orta:
Hrm, good q. Unsure, will put it on my TODO sometime.

orta:
This worked.

alloy:
:+1:

PR #1149, opened 2016-02-12
Applies refine settings to filter artworks

ashfurrow:
Related to #972.

@orta I had some questions about how to use the embedded model view controller. Specifically, what are you thoughts on an approach to display not _just_ artworks? We'll need two different cells, one for each masonry/list layout. Could we chat tomorrow sometime?

In the mean time, filtering by price works :tada: and I shored up the tests around the sale view model.


orta:
The embedded models VC powers the favouritesVC, so it's being used for Artists/Shows/Artworks, SaleLots is just another model to be embedded :+1:

orta:
Cool, looks good :+1:

ashfurrow:
Thanks Orta, will look there :bow:

Issue #1061, opened 2016-01-23
Fix NSInvalidArgumentException in - [ARSiteHeroUnitViewController tappedUnit:], line 258


artsyit:
### Stacktrace ###

<pre>ARSiteHeroUnitViewController;tappedUnit:;ARHeroUnitViewController.m;258
ARWindow;sendEvent:;ARAppDelegate.m;450</pre>

### Reason ###

*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: 'Pushing the same view controller instance more than once is not supported (<ARSimpleShowFeedViewController: 0x13dd49fd0>)'

### Link to HockeyApp ###

* [https://rink.hockeyapp.net/manage/apps/37029/crash_reasons/105686395](https://rink.hockeyapp.net/manage/apps/37029/crash_reasons/105686395)

alloy:
@orta Any thoughts from the top of your head on this one?

orta:
I wonder if you can double tap the hero units, if so we'd need to support disabling on the first tap, then re-nabling when it comes back.

alloy:
Hmm, but I’d guess that that would push a new instance of the VC onto the stack, or are those VCs cached?

orta:
looking through some logs it's always a `ARSimpleShowFeedViewController` - perhaps someone had a `/` link in the HeroUnit. Might have to add a case for this.

https://rink.hockeyapp.net/manage/apps/37029/app_versions/49/crash_reasons/105686395.html?page=2&per_page=30&type=crashes

orta:
Seems to have been updated and fixed too, but we can stop regressions

![screen shot 2016-02-08 at 11 44 41](https://cloud.githubusercontent.com/assets/49038/12884918/61f79ef8-ce59-11e5-9818-c5f1216b4e6f.png)

alloy:
Indeed, good call :+1: 

I wonder if it lies in _how_ these get pushed onto the stack, because `-[ARTopViewController pushViewController:]` should already be taking care to not actually re-add a root view controller _again_.

orta:
aye,  could look in the childVCs and pop to the same one ( if you access any root VC via a URL this code  _could_ happen)

alloy:
https://github.com/artsy/eigen/blob/adaa5d857c16df7b411db268d7af9a7d53dc3774/Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m#L344-L346

orta:
[...]

PR #1150, opened 2016-02-12
Don't crash on hero units with a '/' url

orta:
Added comments to the HeoUnitVC as this is not a common pattern


alloy:
Fixes #1061

PR #1174, opened 2016-02-17
Native WorksForYou VC

mennenia:
In order for me and @sarahscott to PR our changes as we go forward, I propose merging a switch between the working martsy view and our WIP native view. 

I've created a dev options (shake phone / enter ` in the simulator) to get the native works for you view (basically copied what was done for the native auctions view). So you can choose to continue with the webview or see what we're up to in the native view. 

This view is still empty at the moment and only has a title. I've started on the layout and Sarah is working on the network part, but that's not part of this PR.

PR #1176, opened 2016-02-18
Mc nativeworksforyouvc switch

mennenia:
The sane version of [this PR](https://github.com/artsy/eigen/pull/1174)

A switch between native and web VC of "works for you"


alloy:
:ok_hand:

PR #1177, opened 2016-02-18
Sorting lot listings

ashfurrow:
Related to #972.

I've not included snapshot tests for the actual sorting yet because I need to implement the collection view cells first, so those will be included in my next PR.


Artsy/Categories/Apple/Int+Metrification.swift
```diff
@@ -0,0 +1,34 @@
...
+        guard self > 1000_00 else { return self }
+        // currencyStringForDollarCents will round something like 1500_00 up to $2000, but we want to round _down_.
+        // So we divide by 1000_00 to turn us into dollars momentarily, then back into cents to remove their significant digits.
+        let dollarsK = Int(floor(Float(self) / Float(1000_00)))
+        let cents = Int(dollarsK) * 1000_00
+        return cents
+    }
+
+    // Removes three significant digits from number.
+    func removeKSignificantDigits() -> Int {
```
orta: make private?

ashfurrow: Makes sense.

orta:
Cool, all in all feels good, you can merge or make the `removeKSignificantDigits` private if you want

PR #1178, opened 2016-02-18
@alloy=>Network model for 'Works For You' view controller

sarahscott:
:dizzy:


Artsy/Networking/Network_Models/ARWorksForYouNetworkModel.m
```diff
@@ -0,0 +1,106 @@
...
+    self.currentPage = 1;
+    return self;
+}
+
+- (void)getWorksForYou:(void (^)(NSArray<ARWorksForYouNotificationItem *> *))success failure:(void (^)(NSError *error))failure
+{
+
+    [self getArtworks:^(NSArray *artworks) {
+
+        NSMutableDictionary *artistDict = [[NSMutableDictionary alloc] init];
```
sarahscott: @alloy @ashfurrow feeeedback on this part?

Artsy/Models/ARWorksForYouNotificationItem.h
```diff
@@ -0,0 +1,15 @@
...
+
+@class Artist, Artwork;
+
+@interface ARWorksForYouNotificationItem : NSObject
+
+@property (nonatomic, copy, readonly) NSDate *date;
+@property (nonatomic, strong, readonly) Artist *artist;
+@property (nonatomic, strong, readonly) NSArray<Artwork *> *artworks;
+
+- (NSString *)numberOfWorksString;
```
alloy: I’d prefer something like `formattedNumberOfWorks` over Hungarian notation.

Artsy/Networking/Network_Models/ARWorksForYouNetworkModel.m
```diff
@@ -0,0 +1,106 @@
...
+@end
+
+@implementation ARWorksForYouNetworkModel
+
+- (instancetype)init
+{
+    self = [super init];
+    if (!self) return nil;
+
+    self.currentPage = 1;
```
alloy: Don’t use accessor methods in `init` or `dealloc`, just to ensure that no side-effects occur that messes with the state of the object at that point in time.

Artsy/Models/API_Models/Partner_Metadata/Artwork.m
```diff
@@ -67,7 +67,8 @@ + (NSDictionary *)JSONKeyPathsByPropertyKey
         ar_keypath(Artwork.new, published) : @"published",
         ar_keypath(Artwork.new, saleMessage) : @"sale_message",
         ar_keypath(Artwork.new, sold) : @"sold",
-        ar_keypath(Artwork.new, isPriceHidden) : @"price_hidden"
+        ar_keypath(Artwork.new, isPriceHidden) : @"price_hidden",
+        ar_keypath(Artwork.new, updatedAt) : @"published_changed_at"
```
alloy: How about `publishedAt` ?

Artsy/Networking/Network_Models/ARWorksForYouNetworkModel.m
```diff
@@ -0,0 +1,106 @@
...
+
+            ARWorksForYouNotificationItem *item = [[ARWorksForYouNotificationItem alloc] initWithArtist:artist artworks:artworks date:artworks.firstObject.updatedAt];
+
+            [notificationItems addObject:item];
+        }];
+
+        // sort notification items by date
+        NSSortDescriptor *descriptor = [[NSSortDescriptor alloc] initWithKey:@"date" ascending:NO];
+        if (success) {
+            success([notificationItems sortedArrayUsingDescriptors:@[descriptor]]);
```
alloy: Is there a point in doing _any_ of this work if no `success` block were given?

Maybe add the `_Nonnull` modified to he `success` parameter?

Artsy/Networking/Network_Models/ARWorksForYouNetworkModel.m
alloy: This looks good to me :+1: 

Only possible improvement, which would yield very slight performance increase, is to create the collections with the size that you know upfront. E.g.

```objc
NSMutableDictionary *artistDict = [NSMutableDictionary dictionaryWithCapacity:artworks.count];
// ...
NSMutableArray *notificationItems = [NSMutableArray arrayWithCapacity:artistDict.count];
```

Artsy/Networking/Network_Models/ARWorksForYouNetworkModel.m
```diff
@@ -0,0 +1,106 @@
...
+        if (!sself) return;
+
+        sself.currentPage++;
+        sself.downloadLock = NO;
+
+        if (artworks.count == 0) {
+            sself.allDownloaded = YES;
+        }
+
+        if (success) success(artworks);
```
alloy: Same question as above, is there any reason for there ever to not be a `success` callback?

Artsy/Networking/Network_Models/ARWorksForYouNetworkModel.m
```diff
@@ -0,0 +1,106 @@
...
+    } failure:nil];
+}
+
+- (void)getArtworks:(void (^)(NSArray<Artwork *> *))success failure:(void (^)(NSError *))failure
+{
+    if (self.downloadLock) {
+        return;
+    }
+
+    _downloadLock = YES;
```
alloy: We don’t need to fix this now, but this lock that’s not a threading lock is really making me itch.

I bet the reason is to ensure that we don’t have multiple requests _from the main thread_, so in that case we should maybe add an assertion that this is the main thread (this change can be made now), but in that case I wonder if we can’t also just simplify it by having a `weak` reference to the network request that’s made below and check for that on entry, assuming it gets deallocated as soon as it’s done. E.g.

```
@interface ARWorksForYouNetworkModel ()
@property (atomic, weak) AFHTTPRequestOperation *currentRequest;
@end

@implementation ARWorksForYouNetworkModel
- (void)getArtworks:(void (^)(NSArray<Artwork *> *))success failure:(void (^)(NSError *))failure
{
  NSAssert([NSThread isMainThread], @"This should only be called from the main thread.");

  if (self.currentRequest) {
    return;
  }

  // Assuming that AFNetworking releases the request operation as soon as it is done (be it success or failure),
  // then `self.currentRequest` will be set to `nil` automatically by the objc runtime (because it is `weak`).
  self.currentRequest = [self performNetworkRequestAtPage:self.currentPage success:^(NSArray<Artwork *> *artworks) {
    // ...
  }];
}
@end
```

What do you think, easier to understand or more complex?

alloy: _(Obviously you should do some testing to ensure that AFNetworking does indeed release the request operation as soon as it is done :wink:)_

Artsy/Networking/Network_Models/ARWorksForYouNetworkModel.m
```diff
@@ -0,0 +1,106 @@
...
+            sself.allDownloaded = YES;
+        }
+
+        if (success) success(artworks);
+
+    } failure:^(NSError *error) {
+        __strong typeof (wself) sself = wself;
+        if (!sself) return;
+
+        sself.allDownloaded = YES;
```
alloy: What’s the rationale for not allowing more downloads after a failure?

Artsy/Networking/Network_Models/ARWorksForYouNetworkModel.m
sarahscott: very C of you. will do!

Artsy/Models/ARWorksForYouNotificationItem.h
sarahscott: :+1: this is very helpful - I don't want to deviate from eigen's naming conventions

Artsy/Networking/Network_Models/ARWorksForYouNetworkModel.m
alloy: Hah :older_man: 

I don’t go out of my way to do this myself, but you were asking about performance yesterday, so I guess coding this way when you have the info readily available ends up being a lot of small improvements :)

Artsy/Networking/Network_Models/ARWorksForYouNetworkModel.m
sarahscott: :+1:

Artsy/Networking/Network_Models/ARWorksForYouNetworkModel.m
sarahscott: I can see what you mean but I'd rather make this into a new PR and apply the same fix to the FavoritesNetworkModel while I'm at it. Does that make sense?

Artsy/Networking/Network_Models/ARWorksForYouNetworkModel.m
sarahscott: Not sure - this part was lifted from the FavoritesNetworkModel as well. I think you're right - it shouldn't inhibit subsequent downloads.

Artsy/Networking/Network_Models/ARWorksForYouNetworkModel.m
alloy: Totes :+1:

Artsy/Networking/Network_Models/ARWorksForYouNetworkModel.m
alloy: Maybe the favourites endpoint returns a 404 or something for pages that don’t exist? But in that case it should probably explicitly check for those status codes before deciding it’s all done.

As with the one above, I’m fine with this being tackled in another PR.

sarahscott:
@mennenia and I are going to make all of our tests in a separate PR if that's ok. We won't forget :wink: !

ArtsyMobile:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%">1 Warning</th>
     </tr>
  </thead>
  <tbody>
        <tr>
      <td>:warning:</td>
      <td><p>Needs testing on a Phone if change is non-trivial</p>
</td>
    </tr>
    
  </tbody>
</table>
  

  


</p>

Artsy/Networking/Network_Models/ARWorksForYouNetworkModel.m
orta: This is a common pattern in most network models I think, it might be worth turning that issue into a larger "think about how we can add an abstraction to our network models" that could let us write less code in the specific network models while supporting the range of things we use them for ( paging, parsing, chaining, sometimes tiny amounts of logic)

PR #1180, opened 2016-02-18
Tapping an auction lot takes you to that lot's artwork page

ashfurrow:
... where you can then bid on the artwork 🎉

So the bridging head changers are a bit odd, eh? What's going on?

Well we have our switchboard, right? And it has methods on it that return various `UIViewController` subclasses. In order to call those methods, Swift needs to have access to the return types of the functions as well the types of any parameters. So even though I'm passing in a `nil` fair, Swift still needs `Fair.h` in the bridging header.


sarahscott:
:+1:

PR #1191, opened 2016-02-22
Fix the massive pods download by using Analytic's from a tag

orta:
Thus: using shallow clones in CP. This it's safe to assume anything we wanted in 3.0.3 is in 3.0.7


alloy:
Aye :+1:

Issue #1242, opened 2016-03-09
Lower sign-up numbers


alloy:
From @alexanderfabry:

> Seems like there are changes introduced on Jan 18, and Feb 17. Both correspond to app launches. These changes have different effects in China vs the US which utterly blows my mind.
>
> The three events i'm looking at are "first user install" "tapped sign up" and "created account"
>
> Mixpanel account creation funnels:
>
> Jan 18th: drop in tapped sign up => created account funnel breaks (possibly due to user ids overwriting anonymous ids so mixpanel can't put the events into a funnel?) (China AND US)
>
> Since we can't use funnels during account creation post Jan 18th, using mixpanel formulas:
>
> Feb 17th: ratio of created account to tap sign up drops (United States only NOT China)
> Feb 17th: ratio of tapped sign up to first user install drops (China only NOT United States)
>
> We do know that this isn't just a tracking issue though: the gravity database data tells us that at least as of Feb 17th, users of iOS apps are not successfully creating accounts in the same way they were previously.

- [x] Real new account numbers has gone down in Gravity’s DB, so that’s not just an analytics problem.
- [x] There might be some analytics issues, which need more figuring out as to which they are and what’s happening, if I understand it correctly.

alloy:
With regards to the actual account numbers, we were no longer showing the launch screen (sign-up/on-boarding) on application launch. Meaning that all new users were using the app as trial users and only some might convert to signed-up users.

This is fixed with #1243.

alloy:
I think one of the weird things we were seeing in the analytics data was people still tapping sign-up, but rarely converting to signed-up users. My current hypothesis is that:

1. because people were no longer signing-up at initial launch (#1243)
2. they would use the app and eventually run into a sign-up view (e.g. the ‘YOU’ tab)
3. then be more inclined to cancel sign-up

If we are measuring a ‘cancel’ event, then these numbers should have gone up, if my theory is correct.

anipetrov:
Don't think there are any outstanding analytics-related bugs related to this.

katarinabatina:
@alloy would be great if we could summarize this in just a few paragraphs to email to some team members. We're we able to identify why/how on-boarding was dropped?

alloy:
@katarinabatina It was disabled in order to test analytics changes, but was left in by accident and unfortunately not caught during review. We are looking into ways to stop that from happening again in the future.

PR #1243, opened 2016-03-09
[Onboarding] Re-enable on launch, which was accidentally disabled.

alloy:
Related to #1242

We were no longer showing the on-boarding screen when launching the app.


orta:
:+1:

PR #1254, opened 2016-03-12
[WIP] [Notification] Wake app when a notification is received.

### LINKED ISSUE #1119: [Push Notifications] Make app launch in background, so we can record ‘received’ event
Not sure if this is allowed or not, but we can use the `content-available` flag to indicate that the app should launch in the background (to prefetch data).

Right now the ‘received’ vs ‘tapped’ events are without distinction. Receiving a notification does not necessarily trigger the app to load, so all of those ‘received’ notifications were people that either tapped, or had the app open at the time they received it.

alloy:
Fixes #1119

Depends on https://github.com/artsy/gravity/pull/9835 to work.

### DIFF
--- a/Artsy/App/ARAnalyticsPapertrail.m
+++ b/Artsy/App/ARAnalyticsPapertrail.m
@@ -40,7 +40,7 @@
                                          options:0
                                            error:&error];
             NSAssert(error == nil, @"Unable to save analytics papertrail: %@", error);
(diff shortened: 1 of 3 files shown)


alloy:
The logic in this PR is not good enough yet, I just realised.

sarahscott:
Should I review this now or hang tight?

alloy:
Going to create a new PR.

PR #1276, opened 2016-03-15
Don't make danger use master, let the API give the commit ranges #trivial

orta:


alloy:
Sounds good to me :+1:

PR #1279, opened 2016-03-16
Add the plist option for 'does not contain crypto' #trivial

orta:
There's a way to skip that question in iTunes connect, by doing this. So less is less.


alloy:
:+1:

PR #1291, opened 2016-03-17
[Analytics] Include tab labels and rename ‘bell’ to ‘notifications’.

alloy:
Also makes all tabs have accessibility labels.

### DIFF
--- a/Artsy/App/ARAppDelegate+Analytics.m
+++ b/Artsy/App/ARAppDelegate+Analytics.m
@@ -803,9 +803,10 @@
                             ARAnalyticsProperties:^NSDictionary*(ARTabContentView *view, NSArray *parameters) {
                                 NSInteger index = [parameters.firstObject integerValue];
                                 UIButton *button = view.buttons[index];
-                                NSString *title = button.titleLabel.text.lowercaseString;
+                                NSString *title = button.accessibilityLabel ?: button.titleLabel.text;
+                                NSParameterAssert(title);
                                 return @{
-                                    @"tab name" : title ?: @""
+                                    @"tab name" : (title ?: @"").lowercaseString
                                 };
                             },
                             ARAnalyticsShouldFire: ^BOOL(ARTabContentView *view, NSArray *parameters) {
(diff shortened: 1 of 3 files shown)


Artsy/App/ARAppDelegate+Analytics.m
```diff
@@ -1243,13 +1244,13 @@ - (void)setupAnalytics
...
                         },@{
-                            ARAnalyticsPageName: @"Bell",
+                            ARAnalyticsPageName: @"Notifications",
                             ARAnalyticsSelectorName: @"forceSetCurrentViewIndex:animated:",
                             ARAnalyticsShouldFire: ^BOOL(ARTabContentView *view, NSArray *parameters) {
                                 return [parameters.firstObject integerValue] == ARTopTabControllerIndexNotifications;
                             },
                         },@{
-                            ARAnalyticsPageName: @"Bell Tapped",
+                            ARAnalyticsPageName: @"Notifications Tapped",
```
anipetrov: Isn't that the event name for tapping on a notification, too?

alloy: Hmm, it could get confusing yeah. The other is called ‘notification tapped’ (singular) and is not a ‘screen view’ event, but maybe it’s better to rename this to ‘Notifications Tab Tapped’ ?

anipetrov: Oh, right, this is a screen view... The only other confusing moment I forsee is that you can get to the notifications screen (screen view 'Notifications Tapped') by tapping on a notification (`notification_tapped`) or tapping the bell (`notifications`)

alloy: No this ‘screen view’ (‘Notifications Tapped’) is specifically meant to differentiate between those two. This one is _only_ triggered if the user taps the tab. I think we should rename this screen view event to ‘Notifications Tab Tapped’.

anipetrov: Oh got it! Okay, makes sense. Still have a hunch that we should make more use of properties to differentiate between the two, eg we have a single screen view for viewing the screen and we have a property that takes values `notification` and `bell`?

Issue #1123, opened 2016-02-04
[Analytics] Events around sign up/account creation


anipetrov:
I used the [Segment debugger](https://segment.com/artsy-engineering/eigen-dev/debugger0) tab for this troubleshooting but it is not exactly real-time + it is challenging if other people are on the beta at the moment. Let me know if you think there are better ways to go about figuring out which event fires when exactly. Here are some initial observations

1. `Sign up with email` event fires even if there was an error signing up; if consecutively you do sign up successfully, the sign up event is fired again
2. After a successful sign up, the  `Sign up with email` event is fired more than once
3. `Created account` event does not fire if onboarding is skipped

cc @katarinabatina

alloy:
@anipetrov Want to dip your toes in some play time with Xcode and the debugger (specifically breakpoints) to be able to track these down yourself?

anipetrov:
That would be amazing @alloy, thank you so much for offering!!

On Monday, February 8, 2016, Eloy Durán <notifications@github.com> wrote:

> @anipetrov <https://github.com/anipetrov> Want to dip your toes in some
> play time with Xcode and the debugger (specifically breakpoints) to be able
> to track these down yourself?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/artsy/eigen/issues/1123#issuecomment-181310907>.
>


-- 
[A.]      Ani Petrov

           Artsy <https://artsy.net/>
           ani@artsy.net
           +1 609 216 4575

alloy:
We’re gonna take it one step further and write some code: https://github.com/artsy/eigen/issues/1127

Can you make a cal event for when you’d have half a day in your morning (my afternoon) to spend on this? If you can wait till the first week of March, then we can even do it at the office when I’m over.

anipetrov:
I could do either -- whatever you think would be the best use of your time. If you'd rather not waist half a day of your time here on this, I will put a cal event for some time next week. 

Thank you again for offering to take the time!!

alloy:
Au contraire, I imagine that pairing whilst I’m there would be more efficient! (And let’s be honest, otherwise m time is filled with coffee drinking anyways.) So let’s do it then :+1:

anipetrov:
Amazing, thanks a million! I will put something on your cal!

alloy:
:bow: :+1:

orta:
Gonna take this I'm afraid, you two should pair on the analytics visual mode :P

[...]

PR #1301, opened 2016-03-18
[Analytics] Make ‘sign-up/connect with email/facebook/twitter’ -> ‘created account’ consistent.

alloy:
Fixes https://github.com/artsy/eigen/issues/1123#issuecomment-198306933.

### DIFF
--- a/Artsy/Networking/API_Modules/ARUserManager.m
+++ b/Artsy/Networking/API_Modules/ARUserManager.m
@@ -376,15 +376,13 @@ static BOOL ARUserManagerDisableSharedWebCredentials = NO;
                    failure:(void (^)(NSError *error, id JSON))failure
   saveSharedWebCredentials:(BOOL)saveSharedWebCredentials;
 {
+    [ARAnalytics event:ARAnalyticsSignUpEmail];
+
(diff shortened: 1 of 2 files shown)


anipetrov:
Makes perfect sense, thank you!

orta:
Should these provide a context dictionary?

anipetrov:
Ideally we would want to have a single `tapped sign up` event that has properties around how the sign up was created (email, facebook, twitter). This works, too, for now -- it is on my team's radar to revisit all major flows (eg account creation, inquiry etc.) and make sure eventing is instrumented with a clear pattern that translates between apps, so we can do that pass later, too. If not too much trouble, we could do this at this pass.

Additionally, we could add context_type: [sign up splash, follow, bell etc.] but I think this might require more work on our end to sketch out all possible contexts.

anipetrov:
Also not sure if more thorough event revamp is necessary for this round if we are moving through with obligatory sign ups.

alloy:
Good point, in which case I'd opt for deferring more elaborate changes until the new onboarding is in place.

anipetrov:
:+1:

PR #1308, opened 2016-03-18
[Search] Auto-selects the search query when you go back to a search view

orta:
This makes it so when you're trying to use the search more than once per session, you can easily do another query as the text is highlighted for you. 

As this is probably more the use case than the opposite, of wanting to expand on the current query,  I've added it.

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -23,8 +23,7 @@ upcoming:
     - Add accessibility labels for the Home and Bell tab. - alloy
     - Auctions are now native views by default - orta
     - Added dude in portrait view in room - dblock
-
-  notes:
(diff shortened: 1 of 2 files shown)


alloy:
Seems a fair assumption :+1:

CHANGELOG.yml
```diff
@@ -23,8 +23,7 @@ upcoming:
     - Add accessibility labels for the Home and Bell tab. - alloy
     - Auctions are now native views by default - orta
     - Added dude in portrait view in room - dblock
-
-  notes:
+    - Search will auto-select the old query, meaning when you go back and type again it will start a new search  - orta
```
orta: notes was in here twice

PR #1327, opened 2016-03-21
Deploy shenanigans

alloy:


orta:
the fastlane thing means we get out of date reallllllly quickly, are you sure about that?

alloy:
The purist in me doesn’t see the point of having Fastlane in the gem bundle and then not using the actual bundled versions. We could just as well leave it out of the Gemfile completely and install the gems as part of the CI distribution process?

alloy:
(I’m fine to revert btw, this is all more a philosophical issue.)

alloy:
My comment on https://forums.developer.apple.com/message/125506:

> In my 4th attempt I removed the `ITSAppUsesNonExemptEncryption` key from my Info.plist file and it succceeded, so I thought that that was the issue. However, after reading this thread I’m getting the feeling that it might have just been that the first 3 builds fail for those affected and the 4th succeeds.

So, maybe we shouldn’t remove it just yet.

orta:
> We could just as well leave it out of the Gemfile completely and install the gems as part of the CI distribution process? 

is a good idea, I'm up for that, remember you need some of the tooling for app store metadata deployment

alloy:
Deploying works again from CI including the info plist key, so punting on this for now.

PR #1346, opened 2016-03-22
Removes nav bar shadow for serif nav bar

### LINKED ISSUE #1340: get rid of rule on auction sale modal
image

ashfurrow:
I think this is likely what @orta was [talking about](https://github.com/artsy/eigen/issues/1340#issuecomment-199942980). @katarinabatina from the snapshot changes you can see the difference – we can chat about this whenever you're free. 

Fixes #1340.


ashfurrow:
Discussed with Katarina, and _instead_ we're just going to change the nav bar shadow to match the light grey used by the separators.

PR #1406, opened 2016-04-12
Remove swiftDeveloperExtras code

sarahscott:
Otherwise you're going to see a random auction every time you start eigen....


orta:
This file will need all the same kinds of documentation that exists for the other developerExtras, I didn't even know this existed. 

https://github.com/artsy/eigen/blob/master/Artsy/View_Controllers/App_Navigation/ARTopMenuViewController+DeveloperExtras.m#L4-L18

and 

https://github.com/artsy/eigen/blob/master/Dangerfile#L20

sarahscott:
:+1:

alloy:
Aye. I will merge this as-is so that others don’t run into this, but please add those other changes in a new PR.

PR #1411, opened 2016-04-13
[wip] Lot list (with sticky cells 🎉)

### LINKED ISSUE #1112: Lot List
Implement the lot list, presented in a modal on iPhone and presented in split view controller on iPad. This issue is only concerned with implementing the list, not the presentation of it.

### Implementation Idea

Create the following class hierarchy

- UITableViewCell
  - LotCell (creates subviews to be rendered)
    - CompactLotCell (organizes subviews based on compact size)
    - RegularLotCell (organizes subviews based on regular size)

The view controller displaying the table view would register both the compact and regular classes with corresponding reuse identifiers, and dequeue the appropriate class depending on the size class of the view controller.

### Estimate

1 Day

### LINKED ISSUE #1114: Investigate possible implementations for sticky cells in lot list
In #1112 we have designs for the currently on-the-block lot to "stick" to the top/bottom of the table view. This applies for lists in a master view controller as well as lists presented modally.

We don't know how we'll implement this yet, this ticket is to investigate possibilities. Looking into Apple's own apps for ideas would be helpful (do they do this anywhere? Music.app or anything?).

ashfurrow:
So I was thinking that we should use a collection view for the lot list. Why? Because we get _super_ easy customization of the builtin flow layout. We can say "give me the default behaviour, but for the current lot cell, make it sticky", which is exactly what we want. The [relevant code](https://github.com/artsy/eigen/pull/1411/files#diff-a0ad99162a9be75c7b116aa10757115fR40) is 4 lines long.

Pros:

- Really easy sticky headers, which were going to be a challenge.
- Performant.
- Cool.

Cons:

- We're using a collection view, so some table-view stuff like slide-to-delete won't work (we don't need any)
- Reusing the cells may have extra hassle down the road, since they're collection view cells instead of the (more common) table view cells.

What's left to do:

- [x] Complete [design](https://cloud.githubusercontent.com/assets/6482926/12704433/05ac9ca6-c829-11e5-8ab8-13a9e17455b4.png).
- [ ] Test layout/controller.
- [x] Implement callbacks for selected lots.

Fixes #1112. Fixes #1114.


ArtsyMobile:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%">1 Error</th>
     </tr>
  </thead>
  <tbody>
        <tr>
      <td>:no_entry_sign:</td>
      <td>No CHANGELOG changes made</td>
    </tr>
    
  </tbody>
</table>
  

  
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%">1 Warning</th>
     </tr>
  </thead>
  <tbody>
        <tr>
      <td>:warning:</td>
      <td>Needs testing on a Phone if change is non-trivial</td>
    </tr>
    
  </tbody>
</table>
  

  


  Generated by :no_entry_sign: <a href="https://github.com/danger/danger/">danger</a>
</p>

briancroom:
To reduce the pain of cell reuse, I've had a lot of success with designing such views as plain UIView subclasses and embedding them inside of a collection view cell or table view cell as necessary. Deriving from the cell subclasses kills composability 😔

orta:
This technique is considerably simpler than what I had in mind :100:

ashfurrow:
@briancroom that's a good idea! I don't think we currently show lots anywhere else in the UI in this list format, but I think building the cell subclass with that in mind would be a good idea :+1:

ashfurrow:
The current implementation looks like the following:

simulator screen shot apr 14 2016 9 11 05 pm

Once we get live events from the server, I'll want to double-check that things look ok when the current lot changes. I'm going to move on to testing then callbacks for selecting a lot.

Artsy/View_Controllers/Live_Auctions/LiveAuctionLotListStickyCellCollectionViewLayout.swift
```diff
@@ -0,0 +1,96 @@
...
+    }
+    
+    required init?(coder aDecoder: NSCoder) {
+        return nil
+    }
+
+}
+
+
+private typealias PublicFunctions = LiveAuctionLotListStickyCellCollectionViewLayout
```
orta: `private [..] PublicFunctions` ?

Artsy/View_Controllers/Live_Auctions/LotListCollectionViewCell+Private.swift
```diff
@@ -0,0 +1,150 @@
...
+
+    class func _currentAskingPriceLabel() -> UILabel {
+        return ARSansSerifLabel().then {
+            $0.font = UIFont.sansSerifFontWithSize(16)
+            $0.backgroundColor = .clearColor()
+        }
+    }
+
+    class func _labelContainerView() -> UIView {
+        // Container needs a reasonable-ish size when we add the artistsNamesLabel so that its preferredMaxLayoutWidth gets set to something non-zero.
```
orta: with #1416 this won't matter :+1:

Artsy/View_Controllers/Live_Auctions/LiveAuctionLotListStickyCellCollectionViewLayout.swift
ashfurrow: The typealias is private so that other files can have their own PublicFunctions typealias. Does that make sense?

orta: heh, it makes sense in that context, but reads like a contradiction

Artsy/View_Controllers/Live_Auctions/LotListCollectionViewCell+Private.swift
ashfurrow: Cool, I removed this and a similar hack in the auction title view in https://github.com/artsy/eigen/pull/1411/commits/5115a1c9c68c5585a1b7087a3809db2555c099ec

ashfurrow:
Merged as per slack convo.

PR #1420, opened 2016-04-15
Fixes overflow issue for 32-bit compilation

### LINKED ISSUE #1405: Mysterious overflow error in LiveAuctionBidViewController
Not easily reproducible, and inhibits builds but doesn't seem to be an issue at runtime if the compiler allows me to build.

screen shot 2016-04-12 at 12 01 26 pm

ashfurrow:
I tried switching to Int64 throughout the codebase but it quickly became a rabbit hole. A hundred billion dollars (the previous high value of the range) is a... pretty unlikely bid amount, so I used `Int.max` instead (which will work on either architecture). This introduces a problem if a bid exceeds $21 474 836.47 on an older device, but I don't think that's likely in the short term. @alloy it may be worth considering opening an issue to make sure this gets addressed eventually.

Fixes #1405.

### DIFF
--- a/Artsy/View_Controllers/Live_Auctions/LiveAuctionBidViewController.swift
+++ b/Artsy/View_Controllers/Live_Auctions/LiveAuctionBidViewController.swift
@@ -62,7 +62,7 @@ class LiveAuctionBidViewModel: NSObject {
         case 10_000_00...19_999_99: return 100_000
         case 20_000_00...49_999_99: return 200_000
         case 50_000_00...99_999_99: return 500_000
-        case 100_000_00...100_000_000_000_00: return 10_000_00
+        case 100_000_00...Int.max: return 10_000_00
         default: return bid
         }
     }


alloy:
> This introduces a problem if a bid exceeds $21 474 836.47 on an older device, but I don't think that's likely in the short term. @alloy it may be worth considering opening an issue to make sure this gets addressed eventually.

Note that I’m fine with whatever technical solution for the build failure.

My issues with `Int64` vs `Int32` are purely that it’s an auction team decision, from my pov. i.e. what is the max you want to support ?, what does Gravity support ?, what do other clients support ?, etc

So if you say this is fine for now, then I’ll merge on green 👍

ashfurrow:
@acjay We could use your input here – do we expect bid amount (expressed in cents) to exceed the value stored in a 32-bit signed integer?

ashfurrow:
Closing in favour of #1421, a more technically sound solution.

acjay:
That's a good point. Let's model prices as `Long`. In a dollars & cents auction, using 32-bit signed integers means the biggest price we can represent is ~$20M. Even though it may be a long time before we hit that point, it would be smart to not back ourselves into a corner, when the cost of going 64-bit and ensuring we never have a problem is negligible.

alloy:
@acjay Thanks

PR #1424, opened 2016-04-15
Adds tests for lot list

ashfurrow:
This adds tests absent from the #1411 PR, as well as un-`pending`'d the live auction view controller test (it was pending because of a problem I had to fix to test the lot list view controller anyway).

Danger will likely complain of no changelog entry, but as it only adds tests I didn't see the need.


ashfurrow:
Awwww yeah 

screen shot 2016-04-15 at 2 22 57 pm

I think that's because CI was already well-underway by the time I created the PR #lifehack

orta:
Cool, looks good

orta:
> Danger will likely complain of no changelog entry, but as it only adds tests I didn't see the need.

You can add `#trivial` to the title to make Danger ignore that check 👍

PR #1428, opened 2016-04-15
Updates Interstellar to v2

ashfurrow:
A work-in-progress; we need to switch from `Signal` to `Observable`. Then I need to add unsubscribing where necessary.


ashfurrow:
I'd say this can be merged as-is – it's backwards compatible with v1 anyway. Up to you @orta.

orta:
Fine by me 👍

Issue #1083, opened 2016-01-31
Lot Information


katarinabatina:
- [x] At-hand lot info under the artwork #1115 
- [ ] Expanded lot info #1116

![image](https://cloud.githubusercontent.com/assets/6482926/12705313/251ba2b2-c83d-11e5-89b7-add1a6110737.png)

![image](https://cloud.githubusercontent.com/assets/6482926/12704426/dd8d46b2-c828-11e5-8445-e2c195aab77f.png)

katarinabatina:
assets

auction status bar

2x grey http://cl.ly/0h421n32392K
2x purple http://cl.ly/061D1G2C252J

katarinabatina:
@devangt @ashfurrow we should keep an eye on this one, it could qualify as show shopper considering we're really burying any other way to access this info at the point that someone is bidding.

devangt:
Let's mark it that so its on our radar 

-devang




On Thu, May 26, 2016 at 5:28 PM -0700, "Katarina Batina" <notifications@github.com> wrote:












@devangt @ashfurrow we should keep an eye on this one, it could qualify as show shopper considering we're really burying any other way to access this info at the point that someone is bidding.

ashfurrow:
As far as I know, the data is being displayed, it's just missing from the server. I'll have a look today, if the data is there, then this might be an easy fix on eigen.

orta:
Yeah, just missing on lots, it should show if it can

katarinabatina:
@ashfurrow @orta if you look at prediction on the web, on a mobile viewport you'll see the data is being pulled in, not sure if that's helpful!

![image](https://cloud.githubusercontent.com/assets/6482926/15612395/9891bdd8-23fc-11e6-8536-de54e596aef3.png)

PR #1433, opened 2016-04-18
Updates the toolbar to look closer to the docs, and uses data from the View Model

orta:
Simplifies some of the logic around having a countdown too


orta:
Part of #1083

Artsy/View_Controllers/Live_Auctions/Views/LiveAuctionToolbarView.swift
```diff
@@ -85,8 +85,8 @@ class LiveAuctionToolbarView : UIView {
             label.font = UIFont.sansSerifFontWithSize(12)
             view.addSubview(label)
 
-            if key == "time" {
-                clockClosure(label)
+            if key == "time" && clockClosure != nil {
```
ashfurrow: Could be `if let clockClosure = clockClosure where key == "time"` but that's just a preference, this is fine.

PR #1435, opened 2016-04-18
Use Top Menu for view controllers

orta:
This does not actually add the routing for Live Auctions, but provides the foundations. If `ARSwitchBoard` now returns a `UINavigationController` then the topVC will present it modally instead of pushing it on a anv. `UINavigationController` doesn't support pushing a `UINavigationController` anyway.

I didn't add this _everywhere_ because there are a lot of places where we're guaranteed to return a `UIViewController` because we use an ARSwitchBoard+Artsy method that returns a known class.

I did consolidate the way we access `ARSwitchBoard sharedSwitchboard` specifically so it's easy to look over all of them with a single find.

### DIFF
--- a/Artsy/Categories/Apple/UIViewController+TopMenuViewController.m
+++ b/Artsy/Categories/Apple/UIViewController+TopMenuViewController.m
@@ -1,0 +1,20 @@
+#import "UIViewController+TopMenuViewController.h"
+#import "ARTopMenuViewController.h"
+
+
+@implementation UIViewController (TopMenuViewController)
+
+- (ARTopMenuViewController *_Nullable)ar_TopMenuViewController
+{
+    if ([self.parentViewController isKindOfClass:ARTopMenuViewController.class]) {
+        return (id)self.parentViewController;
+    }
+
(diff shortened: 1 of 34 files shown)


Artsy/View_Controllers/Fair/ARFairViewController.m
```diff
@@ -1,3 +1,5 @@
+// MARK: Formatter Exempt
```
orta: I was done with seeing those ugly `ARNavigationButtonHandlerKey` block things. 👯

Artsy/Categories/Apple/UIViewController+TopMenuViewController.m
```diff
@@ -0,0 +1,20 @@
...
+{
+    if ([self.parentViewController isKindOfClass:ARTopMenuViewController.class]) {
+        return (id)self.parentViewController;
+    }
+
+    if ([self.navigationController.parentViewController isKindOfClass:ARTopMenuViewController.class]) {
+        return (id)self.navigationController.parentViewController;
+    }
+
+    return [self.parentViewController ar_TopMenuViewController];
```
alloy: Ooooh, smart way around referencing the singleton 👍

Issue #415, opened 2015-05-04
Show page design updates


nsimmons206:
These updates include layout changes, changes to the installs carousel, a new "follow artist" section,  and the addition of new data like full gallery hours, map location, opening reception, additional info and press releases.
![showpage_eigenmartsy_finalspec_02](https://cloud.githubusercontent.com/assets/8950911/7523026/e3060e16-f4c6-11e4-8057-883df73359b5.jpg)

![showpage_eigenmartsy_finaldesign_02](https://cloud.githubusercontent.com/assets/8950911/7523027/e306f0ce-f4c6-11e4-8bad-eac31e06ccf2.jpg)

![showpage_ipad_finalspec_02](https://cloud.githubusercontent.com/assets/8950911/7523028/e307bd92-f4c6-11e4-9610-661a492d4a05.jpg)

![showpage_ipad_finaldesign_02](https://cloud.githubusercontent.com/assets/8950911/7523025/e3049662-f4c6-11e4-8f42-3ba8086ddad6.jpg)

nsimmons206:
Is there a way to copy this issue into microgravity as well? Or should i just make it twice?

alloy:
@nsimmons206 You’ll just have to create it twice.

I assume this is meant for the next sprint?

/cc @katarinabatina

katarinabatina:
@alloy I'm unsure at this point whether or not it will make it into the next sprint we'll have to see how it stacks up against other priorities.

PR #1436, opened 2016-04-18
Use AspectFill for Show Image Previews

orta:
The tests don't really do justice to the point as most show preview images are landscape, but now they will fill edge to edge

screen shot 2016-04-18 at 18 09 18

/cc @katarinabatina ( pretty sure you mentioned this to me offhand at some point in the past year or two )


alloy:
Does this not need to fetch higher-res images in cases like the portrait work in the snapshots?

alloy:
Needs a rebase.

ashfurrow:
Merge had no conflicts, running CI now.

alloy:
@orta I would still like an answer to my question, though.

orta:
Hrm, it _probably_ does, especially on iPads where that could be a long one

orta:
I was less worried about it ( it was already an improvement ) given that most shows have install shots, and most install shots are landscape

katarinabatina:
@orta I think this brings us full circle to a problem we had before where install shots that are more vertically oriented get cut off. Is this logic and improvement on what we had before? We specifically moved away from initial implementation when we launched the ipad. Nick had some very specific ideas for what an improvement would look like in this issue https://github.com/artsy/eigen/issues/415

Could you post a few other examples? From both ipad and iphone before we merge?

alloy:
Filed #1460.

PR #1486, opened 2016-05-10
Beautify README

ReadmeCritic:
😄


orta:
I get where you're coming from for sure, it would be a great choice if we had more metadata, but for only 3 items - I think the table is overkill - thanks though!

PR #1489, opened 2016-05-10
Cleanup surrounding non-native live support

ashfurrow:
## Problem

I'd like to route https://live-staging.artsy.net/54c7e8fa7261692b5acd0600 to use an internal web browser. However, it gets routed as a profile. 

## Change

I've changed the switchboard to only route URLs with known Artsy hosts (so an unexpected Artsy subdomain will get routed to an internal web view instead of as a profile, which would remove the custom subdomain and replace it with either staging or production (m)artsy).

Paired with @orta. I'll add tests tomorrow.


ashfurrow:
This commit should probably be applied against master, actually...

PR #1490, opened 2016-05-11
Cleanup surrounding non-native live support

ashfurrow:
In favour of #1489.

## Problem

I'd like to route https://live-staging.artsy.net/54c7e8fa7261692b5acd0600 to use an internal web browser. However, it gets routed as a profile. 

## Change

I've changed the switchboard to only route URLs with known Artsy hosts (so an unexpected Artsy subdomain will get routed to an internal web view instead of as a profile, which would remove the custom subdomain and replace it with either staging or production (m)artsy).

Paired with @orta.

### DIFF
--- a/Artsy/App/ARSwitchBoard.m
+++ b/Artsy/App/ARSwitchBoard.m
@@ -367,10 +367,13 @@ NSString *const AREscapeSandboxQueryString = @"eigen_escape_sandbox";
 
 - (UIViewController *)routeInternalURL:(NSURL *)url fair:(Fair *)fair
 {
-    // Use the internal JLRouter for the actual routing
-    id routedViewController = [self.routes routeURL:url withParameters:(fair ? @{ @"fair" : fair } : nil)];
-    if (routedViewController) {
-        return routedViewController;
+    BOOL isTrustedHostForPredictableRouting = ([[ARRouter artsyHosts] containsObject:url.host] || url.host == nil);
+    if (isTrustedHostForPredictableRouting) {
+        // Use the internal JLRouter for the actual routing
+        id routedViewController = [self.routes routeURL:url withParameters:(fair ? @{ @"fair" : fair } : nil)];
+        if (routedViewController) {
+            return routedViewController;
+        }
     }
 
     // We couldn't find one? Well, then we should present it as a martsy view
--- a/Artsy/View_Controllers/Debug/Quicksilver/ARQuicksilverViewController.m
+++ b/Artsy/View_Controllers/Debug/Quicksilver/ARQuicksilverViewController.m
@@ -222,7 +222,13 @@
 {
     [self.presentedViewController dismissViewControllerAnimated:NO completion:nil];
 
-    SearchResult *result = self.contentArray[indexPath.row];
+    SearchResult *result;
+    if (!self.contentArray || self.contentArray.count == 0) {
+        result = (id)self.searchController.searchBar.text;
+    } else {
+        result = self.contentArray[indexPath.row];
+    }
+
     [self addObjectToRecents:result];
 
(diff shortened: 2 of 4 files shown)


Artsy/App/ARSwitchBoard.m
```diff
@@ -367,10 +367,13 @@ - (void)openURLInExternalService:(NSURL *)url
 
 - (UIViewController *)routeInternalURL:(NSURL *)url fair:(Fair *)fair
 {
-    // Use the internal JLRouter for the actual routing
-    id routedViewController = [self.routes routeURL:url withParameters:(fair ? @{ @"fair" : fair } : nil)];
-    if (routedViewController) {
-        return routedViewController;
+    BOOL isTrustedHostForPredictableRouting = ([[ARRouter artsyHosts] containsObject:url.host] || url.host == nil);
+    if (isTrustedHostForPredictableRouting) {
```
alloy: Seeing as previously there was no condition on the host at all I don’t understand how adding one here _adds_ the ability to route `live-staging.artsy.net` internally. Or was that done elsewhere and are now all `*.artsy.net` domains being routed internally and is that what you’re trying to prevent?

ashfurrow: Certainly, that's a good question. `live.artsy.net` and `live-staging.artsy.net` are Artsy subdomains, but are _not_ included in `artsyHosts`. Instead, we rely on Echo to add domains to route these URLs. In the absence of Echo routing, the URL `live.artsy.net/abcd1234` _was_ being handled by `self.routes` and handled as a [profile](https://github.com/artsy/eigen/blob/master/Artsy/App/ARSwitchBoard.m#L217-L222). This would try to load `artsy.net/abcd1234`, which would 404. 

To fix this, the added code here checks to make sure that a URL's domain is trusted to have predictable routing. If it doesn't (in the case of `live.artsy.net`) then it's loaded as an internal web view. If a live domain had already been added by Echo, it would have been [routed earlier](https://github.com/artsy/eigen/blob/ff7e750ba38bc716cf6e68f49f678c32adc560e2/Artsy/App/ARSwitchBoard.m#L296-L299).

Let me know if this makes sense, happy to add documentation, clarify, or rewrite.

ArtsyMobile:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-kind="Warning">
        
            1 Warning
        
      </th>
     </tr>
  </thead>
  <tbody>
        <tr>
      <td>:warning:</td>
      <td data-sticky="true">Detected some Swift building time outliers</td>
    </tr>
    
    
  </tbody>
</table>
  

  



Time | Class | Function |
| --- | ----- | ----- |252.1ms | [AuctionViewController.swift](/artsy/eigen/tree/ashfurrow-new-logic/Artsy/View_Controllers/Auction/AuctionViewController.swift#L176) | @objc dynamic func showRefineTappedAnimated(animated: Bool)
183.7ms | [WebSocket.swift](/artsy/eigen/tree/ashfurrow-new-logic/Pods/Starscream/Source/WebSocket.swift#L528) | private func processRawMessage(buffer: UnsafePointer<UInt8>, bufferLen: Int)
57.9ms | [RefinementOptionsViewController+Private.swift](/artsy/eigen/tree/ashfurrow-new-logic/Artsy/View_Controllers/Auction/RefinementOptionsViewController+Private.swift#L63) | func stackView() -> ORStackView




  Generated by :no_entry_sign: <a href="https://github.com/danger/danger/">danger</a>
</p>

Artsy/App/ARSwitchBoard.m
alloy: Aha, yes, total sense! 👍

PR #1503, opened 2016-05-16
Improve analytics when an inquiry fails

orta:
It's looking like some of the shared logic seems to be failing on the artwork view's inquiry button occasionally, adding some more breadcrumbs till  I can devote [considerable time](https://artsy.slack.com/archives/gmv-ios/p1463403163000804) to it.


alloy:
Although the issue has already been fixed, this doesn’t seem like a bad idea to me 👍

PR #1563, opened 2016-05-27
[Pods] Update Emission and React Native.

alloy:
As described on the tin.

This branch has been sent off for a beta.

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -43,7 +43,6 @@ upcoming:
       - QuickSilver now supports direct URLs. - ash
     
     gmv:
-      - Integrate the new artist view (Emission). - alloy
       - Removed artwork auction results - sarah
       - WorksForYouVC reloads content upon network failure if you leave and then return to its tab - sarah
       - Removed updateConstraintsIfNeeded from ARNavigationButton to solve autolayout bugs in home screen tableview - sarah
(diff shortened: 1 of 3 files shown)


CHANGELOG.yml
```diff
@@ -57,6 +56,7 @@ upcoming:
       - Changed the paragraph padding on all ARTextView instances - orta
       - Moved ARSwitchView and ARSpinner into new Extraction pod. - alloy
       - Buyer's Premiunm is shown on AuctionInformationViewControllers - orta
+      - Update Emission since first QA. - alloy
```
orta: might be worth starting to think about a CHANGELOG over there?

Podfile
```diff
@@ -83,6 +83,7 @@ target 'Artsy' do
   # TODO Actually release to our spec repo.
   pod 'Extraction', :git => 'https://github.com/artsy/extraction.git'
   pod 'Emission', :git => 'https://github.com/artsy/emission.git'
+  pod 'React/Core', :git => 'https://github.com/facebook/react-native.git', :tag => 'v0.25.1'
```
orta: my feedback on "shouldn't emission handle this" from the last one of these still stands ( I disagreed with the comments in the react repo, a `Podfile.lock` is meant to be trusted. It's not like it was 2 years ago.

orta:
👍 on green

Podfile
alloy: That is not the point, though, it's that there is no spec for this version in the master spec repo. The comment on the React repo was about not just using master branch.

I'm still on the fence about copying specs for React to our spec repo.

orta: OK, cool, I dig.

CHANGELOG.yml
alloy: Agreed

PR #1607, opened 2016-06-03
Move React Artists to be the default, and add the ability to stop loading React views

orta:
Removes the AROption
Uses Echo to look for a flag that will disable React views. 

_minor note_ only one bad thing happens with this ( and I don't think it matters) we used to pass in a "full" artist object when searching, now it's only the ID, but given hat I think the react one ignores the model object completely - it's not really a problem.


orta:
screen shot 2016-06-03 at 6 11 50 pm

PR #1621, opened 2016-06-06
[Pods] Point Podfile to updated Interstellar project

mennenia:
To fix my podfile.lock / build issues :D (which ... others are having too...? I would assume?)

( edit: #trivial )

PR #1624, opened 2016-06-06
[Infra] Refactor the blurred overlay view #trivial 

orta:
Converts the view that was being done manually into something that can be used in many places if needed. Already covered by tests in LiveAuctionVCTests - so no changes means everything is OK.


ashfurrow:
CI failing because it was submitted before https://github.com/artsy/eigen/pull/1621 was merged, merging.

Issue #1594, opened 2016-06-02
Duplicate entires in the bid history event stream


orta:
Left = iPad, Right = Prediction

![screen shot 2016-06-02 at 2 07 29 pm](https://cloud.githubusercontent.com/assets/49038/15755597/7ed63ad0-28cb-11e6-8d48-e35ad48922b7.png)

orta:
It's the `CompositeOnlineBidConfirmed` events

```
2016-06-02 13:27:08.713 Artsy[76945:5896700] Live Event: {
    amountCents = 1480000;
    createdAtString = "2016-06-01T21:36:34.200Z";
    eventID = "22b84ac3-d538-406c-85cf-b5fd3fc25fb9";
    type = CompositeOnlineBidConfirmed;
}
2016-06-02 13:27:08.713 Artsy[76945:5896700] Live Event: {
    amountCents = 1530000;
    bidder = "<LiveBidder: 0x7f96f2df8c20> {\n    bidderID = \"<null>\";\n    type = OfflineBidder;\n}";
    createdAtString = "2016-06-01T21:36:42.754Z";
    eventID = "49776dca-2fec-498d-9005-ba0cd2349c6c";
    type = FirstPriceBidPlaced;
}

```

ashfurrow:
This is a problem with the table view insertion code, related to the stuff @orta commented out already, I'll fix it.

devangt:
Verified. Not seeing any duplicates.

PR #1640, opened 2016-06-07
Removes duplicates from event history list.

ashfurrow:
This is work for #1594.


Artsy/View_Controllers/Live_Auctions/LiveActionStateReconciler.swift
```diff
@@ -65,8 +65,8 @@ extension PublicFunctions: LiveAuctionStateReconcilerType {
             guard let derivedLotState = json["derivedLotState"] as? [String: AnyObject] else { continue }
             guard let eventHistory = json["eventHistory"] as? [[String: AnyObject]] else { continue } // TODO move to events
 
-            updateLotDerivedState(lot, derivedState: derivedLotState)
             updateLotWithEvents(lot, lotEvents: eventHistory)
+            updateLotDerivedState(lot, derivedState: derivedLotState)
```
ashfurrow: This change seemed like a good idea, since we do things in the same order on lot event updates, so we should be consistent.

Artsy/View_Controllers/Live_Auctions/LiveAuctionBidHistoryViewController.swift
```diff
@@ -119,3 +94,21 @@ class LiveAuctionBidHistoryViewController: UITableViewController {
...
+enum TableViewUpdateCondition {
+    case Reload
+    case Update
+}
+
+extension Array where Element: LiveAuctionEventViewModelType {
+
+    // If an event was removed from the derived list (maybe it was cancelled) then we can't rely
+    // on insertRowsAtIndexPaths(), we need to reload. Alternatively, sending in _empty_
+    // new events indicates that a previously user-facing event has been modified.
```
ashfurrow: An empty array is a bit of a leaky abstraction, but the proper solution is to construct another signal for deleting events, and a further new signal for updating events. Since Causality doesn't have deleting events quite under wraps, I think this is a good compromise for now.

Artsy/View_Controllers/Live_Auctions/LiveActionStateReconciler.swift
orta: cool

PR #1642, opened 2016-06-07
[Live] Use Bid Composite Events to determine the lot event history

orta:
Two major things:

* Bid confirmations
* Bid coloring according to state


orta:
and green on the first go, what's the chances

PR #1656, opened 2016-06-08
[Live] Don't scroll sideways when we are showing the full history of a work

orta:
Not sure how this regressed, but it's back in now 👍


ashfurrow:
👍

PR #1677, opened 2016-06-10
Make bid history larger

orta:
On devices that support it (anything but the iPhone 4S basically).

(From Ash)


Artsy/View_Controllers/Live_Auctions/LiveAuctionLotSetViewController.swift
```diff
@@ -44,6 +48,12 @@ class LiveAuctionLotSetViewController: UIViewController {
             let layout = LiveAuctionPlainLotCollectionViewLayout(delegate: dataSource)
             collectionViewLayout = layout
             lotCollectionViewLayout = layout
+            adjustConstraintsForLargeScreens = true
```
orta: ah, yeah, nice touch

ashfurrow: Tried to make it a closure that could be executed, but it accessed `self` before `super.init`, so had to fall back to this.

PR #1700, opened 2016-06-11
[Live] Handle out of order bids by ensuring the top bid is always at the top index

orta:
Wrote on a plane, not looked at your PRs yet, dunno how they'll fit yet.


ashfurrow:
I'll take a look 👍

Artsy/View_Controllers/Live_Auctions/LiveAuctionBidHistoryViewController.swift
```diff
@@ -78,8 +79,18 @@ class LiveAuctionBidHistoryViewController: UITableViewController {
         tableView.backgroundColor = .clearColor()
     }
 
+    override func numberOfSectionsInTableView(tableView: UITableView) -> Int {
```
ashfurrow: Ah, interesting approach.

Artsy/View_Controllers/Live_Auctions/ViewModels/LiveAuctionLotViewModel.swift
```diff
@@ -179,8 +181,12 @@ class LiveAuctionLotViewModel: NSObject, LiveAuctionLotViewModelType {
         return LiveAuctionBidViewModel.nextBidCents(model.askingPriceCents)
     }
 
-    var topBidEvent: LiveAuctionEventViewModel? {
-        return fullEventList.filter({ $0.isBid }).last
+    var topBidEvent: LiveAuctionEventViewModel? = nil
```
ashfurrow: This is one of the big changes I made in my PRs, to refer to the winning/selling-to bid events instead of computing it locally.

Artsy_Tests/View_Controller_Tests/Live_Auction/LiveAuctionBidHistoryViewControllerTests.swift
```diff
@@ -210,5 +210,35 @@ class LiveAuctionBidHistoryViewControllerTests: QuickSpec {
...
+            let e3 = bid(15000, bidder: ["type": "ArtsyBidder", "bidderId": "23653"])
+            let e4 = confirm(e3, lotID: lotID)
+            let e5 = bid(14000, bidder: ["type": "OfflineBidder"])
+
+            let subject = LiveAuctionBidHistoryViewController(lotViewModel: lotVM)
+            let events = [e1, e2, e3, e4, e5]
+            lotVM.addEvents(events as! [LiveEvent])
+
+            expect(subject) == snapshot()
+        }
```
ashfurrow: Good test, should've had one like this in my PR.

orta: worked entirely offline, so TDD was the only option. 👯

orta:
So is this something we want, can deal with merging etc if so?

ashfurrow:
I think we're okay without it, except for https://github.com/artsy/eigen/pull/1700/files#diff-cdbea92cdbacb87e160cef2621ada0c3L70 (not sure I follow the distinction there). I can add the tests from this in another PR, but that's not critical.

ashfurrow:
Ah I see, it's to make `findBidWithValue` more accurate? We use it when hooking up confirmations to bids, but confirmations come with `event.eventId` which would be a more precise way of looking up a corresponding bid.

orta:
The amountCents is only for confirmations, which was confusing, cause I thought I could get it on a bid.

orta:
Sounds like we should do a `make deploy` then, from current master

ashfurrow:
Done'd. Let's leave this open as a reference to port those tests over.

orta:
Cool beans

PR #1748, opened 2016-06-23
Changed loadUnknownPathWithID back to loadProfileWithID in Switchboard

sarahscott:


ashfurrow:
@orta Once https://github.com/artsy/eigen/pull/1750 is merged, we should revert this?

PR #1747, opened 2016-06-22
Fixes leaks 💦📱

ashfurrow:
This fixes a bunch of reference cycles we have around live auctions. It's a WIP because a few tests for one class are broken (puzzling  going to leave it to tomorrow). Also, I want to test it on-device thoroughly, since it's a fairly wide-reaching PR.

PR #1775, opened 2016-06-28
Fixes problems with unretained observables

### LINKED ISSUE #1774: Live interface hangs when opening on latest master
This isn't present in the App Store build or any betas. I suspect it's related to https://github.com/artsy/eigen/pull/1747

ashfurrow:
Fixes #1774, related to #1747. The problem is that calling `subscribe()` or `flatMap()` on an `Observable` doesn't strongly reference that observable. The code we had assumed it was, and didn't take responsibility for referencing the observables strongly, so they were deallocated before the network call completed. Since Interstellar lacks the clarity of a dispose bag or similar structure to explicitly handle "who owns this observable?", sometimes I feel like I'm just guessing 😞 

I take responsibility for this, I tested on device for the live interface but didn't test the existing non-live auction view controller even though I made changes to its network stack, too.

### DIFF
--- a/Artsy/View_Controllers/Auction/AuctionBiddersNetworkModel.swift
+++ b/Artsy/View_Controllers/Auction/AuctionBiddersNetworkModel.swift
@@ -16,11 +16,11 @@ class AuctionBiddersNetworkModel: AuctionBiddersNetworkModelType {
 
         // Based on the saleID signal, fetch the sale registration status.
         ArtsyAPI.getCurrentUserBiddersForSale(saleID,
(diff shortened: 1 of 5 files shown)


orta:
> sometimes I feel like I'm just guessing 😞

Aside from the issue, are there ways that we can improve interstellar to handle this?

ashfurrow:
That's a good question. Better documentation would help, but not everyone reads the docs. Interstellar is meant to be lightweight and I'm not sure an abstraction like a dispose bag fits in with @JensRavens 's vision for the library. I'll have a think.

PR #1780, opened 2016-06-29
Adds SwiftLint to Danger.

ashfurrow:
Fixes https://github.com/ashfurrow/danger-swiftlint/issues/2


Dangerfile
```diff
@@ -95,3 +95,8 @@ if outliers.any?
   markdown(([headings] + warnings).join)
 end
 
+swiftlint.config_file = 'Artsy/.swiftlint.yml'
+swiftlint.lint_files
+
+swiftlint.config_file = 'Artsy_Tests/.swiftlint.yml'
+swiftlint.lint_files
```
orta: Is running this twice fast? ( does it need to compile an AST twice, I guess? )

ashfurrow:
Hmm, thought the failure was something I was only getting on `danger local`, will need to look into this tomorrow with @orta.

Dangerfile
ashfurrow: As far as I understand, SwiftLint doesn't touch the compiler or ASTs or anything, it just parses text. For each file, the plugin makes a different `swiftlint` invocation, which are all pretty quick. Doing this twice, with two config files, is okay since one config file excludes the other's files, and vice versa.

ashfurrow:
Okay, so danger is still classing this not as a pull request, so I can't test if the lint warning in https://github.com/artsy/eigen/pull/1780/commits/9cafc338ee37b27ada09bfd0d1b11297e978ee8a is actually working. However, it works when I run Danger locally, so it should be fine, right? I'll remove that commit and fpush.

Dangerfile
orta: 👍

orta:
👍

orta:
skipping is odd though

PR #1872, opened 2016-09-28
[Home] Get rid of old background fetch related code.

### LINKED ISSUE #1842: Delete background fetch work from eigen
in an ideal world, that'd be replaced with a new version in the future that grabs your home modules and puts it in relay's cache, but for now you need to:

* remove https://github.com/artsy/eigen/blob/master/Artsy/App/ARAppBackgroundFetchDelegate.m
* and remove the fetch attribute from the plist, see this PR - https://github.com/artsy/eigen/commit/cd55459785c4ee379206122406ca5ddfb3bfcfd8

alloy:
Closes #1842


alloy:
The test failures are the same as before this PR. Merging.

PR #1883, opened 2016-09-30
Removes iOS 9-specific workaround for iOS 10

ashfurrow:
This bug was fixed in iOS 10, so compiling for iOS 10 on my device would throw an exception. This PR fixes that.


alloy:
Swift 💫

PR #1887, opened 2016-10-04
[WIP] Updates for status bar styling

### LINKED ISSUE #1876: [Home] Eliminate black rectangle behind status bar
image

image

sarahscott:
@ashfurrow, I noticed you've done a lot of things with the status bar for auctions and I want to make sure I don't mess any of that up. Do you foresee any issues with these changes?

Fixes #1876 

simulator screen shot oct 4 2016 5 29 27 pm

simulator screen shot oct 4 2016 4 54 13 pm

### DIFF
--- a/Artsy/View_Controllers/SerifModalWebNavigationController.swift
+++ b/Artsy/View_Controllers/SerifModalWebNavigationController.swift
@@ -10,7 +10,7 @@ class SerifModalWebNavigationController: UINavigationController, UINavigationCon
 
     override init(rootViewController: UIViewController) {
         super.init(rootViewController: rootViewController)
-        (rootViewController as? ARExternalWebBrowserViewController)?.ignoreStatusBar = true
+//        (rootViewController as? ARExternalWebBrowserViewController)?.ignoreStatusBar = true
     }
 
     required init?(coder aDecoder: NSCoder) {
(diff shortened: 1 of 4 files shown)


Artsy/View_Controllers/SerifModalWebNavigationController.swift
```diff
@@ -10,7 +10,7 @@ class SerifModalWebNavigationController: UINavigationController, UINavigationCon
 
     override init(rootViewController: UIViewController) {
         super.init(rootViewController: rootViewController)
-        (rootViewController as? ARExternalWebBrowserViewController)?.ignoreStatusBar = true
+//        (rootViewController as? ARExternalWebBrowserViewController)?.ignoreStatusBar = true
```
sarahscott: This vc throws a `fatalError` if I remove all the `ignoreStatusBar`-related code. Should we keep `ignoreStatusBar` in `ARExternalWebBrowserViewController` and change its functionality a bit?

sarahscott:
@katarinabatina do we want the status bar to have a white background app-wide or only for Home?

katarinabatina:
@sarahscott we don't actually want it to have a background at all. I suggest not changing this app wide at the moment though as we'll have to adjust where the top of the artwork VC begins, etc. For now the actual status text should be black with no background.

alloy:
> we don't actually want it to have a background at all

@katarinabatina Hmm, I’ve been trying to think about this a lot, but I’m a bit lost. For instance, in the case of the status bar not having a background at all (alpha = 0), then wouldn’t the text in the status bar completely clash with the content of our view when scrolling down?

katarinabatina:
@alloy I thought originally the clash wouldn't be bothersome but I took a second look given your comment and I think you're right, it's worth having a white background to the status bar. One thing I did like is keeping the bar at 98% opacity, pinterest does this and it's a nice effect. If that's possible great otherwise let's just make it solid.

sarahscott:
Closing in favor of other status-bar-related prs

PR #1904, opened 2016-10-11
[Onboarding] Fix capitalisation on search button asset

### LINKED ISSUE #1860: [Onboarding] Missing search icon
image

mennenia:
Not sure why this would work on simulator but not device. Think this fixes #1860.


orta:
wonder if the phone filesystem runs on a different case sensitivity?

PR #1911, opened 2016-10-12
[Quicksilver] Fixes adding a URL to the history #trivial

orta:
You can throw a URL in Quicksilver and it will go there and add it into history.


alloy:
Aye, this is good :ok_hand:

PR #1919, opened 2016-10-14
[Gene] Support running the Gene VC from React Native as a lab option

orta:
This currently doesn't work due to Objective-C API incompatabilities, I am assuming that a deploy of Eigen requires a deploy of Emission - if so, then this is cool 👍


alloy:
Yup, will have a deploy 👍

What's the default lab value, enabled?

orta:
off

alloy:
I’m going to turn it on by default, don’t want anyone to miss it in the beta.

PR #1935, opened 2016-10-18
[blocked] Update Gene to use the new ARGeneComponentVC

orta:
This switches the default to on, and uses echo to support turning it off.


orta:
Relies on a n Emission deploy and https://github.com/artsy/emission/pull/362

alloy:
Tapping the REFINE button [won’t do anything yet](https://github.com/artsy/eigen/blob/3703dee2ccf15548f0010e2c53a9e95e8660daaf/Artsy/App/ARAppDelegate+Emission.m#L147-L150), though, and I’m not entirely sure right now how it needs to be hooked-up.

alloy:
Oh, judging by the current state of Emission, it seems like the REFINE button is just not available at all? i.e. you decided to not make that work for now?

orta:
There's no refine button in the UI for this run

Issue #1926, opened 2016-10-15
nav rule separator should be same as bucket separator  


katarinabatina:
![image](https://cloud.githubusercontent.com/assets/6482926/19410726/3f5b803a-92c1-11e6-8334-06f77339425a.png)

the bucket separator rule (top one shown below) is correct, should be applied to the nav cc @sarahscott 
![image](https://cloud.githubusercontent.com/assets/6482926/19410729/6325eece-92c1-11e6-866d-2e5b00242edc.png)

alloy:
Hah, so 1pt thickness vs 1px after all. My bad 🤗

orta:
@katarinabatina can you confirm this is what you have in mind? 

![screen shot 2016-10-19 at 16 42 25](https://cloud.githubusercontent.com/assets/49038/19526071/1038cd8c-961b-11e6-9f60-fa6745c76231.png)

katarinabatina:
@orta rock on, looks right 👍

orta:
Fixed by #1944

PR #1944, opened 2016-10-19
[Nav] Use a thick separator for the nav

orta:
I think this is what is being asked for in #1926


orta:
Cool, got the confirmation 👍

PR #1949, opened 2016-10-19
[Onboarding] Add default placeholders for when no search results 

### LINKED ISSUE #1881: [Onboarding] When we have no artist suggestions, show top artists list again
right now we're showing no search results which is inaccurate, we should instead show the top artists list.

mennenia:
Fix #1881 

blah


orta:
This feels weird, maybes it's the GIF - does it not say that we've not found any results anywhere? ( maybe it's the GIF ) if it feels fine to you I'm 👍

mennenia:
It's not supposed to. 

It will when there are no artists matching your query, however, if you select an artist that has no suggested / related artist (e.g. H. Call in this case), then instead of just showing no results (after all you didn't search), we show popular artists again.

simulator screen shot 19 oct 2016 17 50 52

mennenia:
Good shout tho :).

orta:
Rockin - yep, that's great for me

Issue #1468, opened 2016-05-04
Re-instate auction results for select artists.


alloy:
See https://github.com/artsy/force/issues/4925

As I understand it, this means we’ll only ever show auction results _on an artist view_, **not** from an artwork page as was previously the case.

- [x] Add flag to artist payload in Gravity API: https://github.com/artsy/gravity/issues/9990
- [ ] Add flag to artist schema in metaphysics
- [ ] Update design of artist view to include auction results

alloy:
The Force ticket mentions:

> Not to be released publically before Tuesday May 10th.

Is there a due date or is just anywhere after May 10th fine?

briansw:
Here's the design. We just append the auction results to the accordion module in the last slot and make the arrow point to the right to indicate a page change. In terms of due date Matt would probably be best to answer that one.

![screen shot 2016-05-04 at 12 58 41 pm](https://cloud.githubusercontent.com/assets/1022172/15022101/0be0f3e2-11f8-11e6-85ed-146ce9a6ec42.png)

katarinabatina:
@alloy afraid this ticket got lost in the shuffle. Devang and I discovered a separate issue of how we handle users coming from google search results. We end up dumping them on the artist page versus taking them through to the auctions result page. this will be a priority for our next release.

alloy:
I don’t follow what the separate issue is. Is the thing that needs to be done still the exact same thing as initially described above?

alloy:
@katarinabatina bump

katarinabatina:
@alloy i will make a separate ticket for the other issue i was mentioning otherwise everything that needs to be done here is the same.

katarinabatina:
link to microgravity view for auction results

orta:
@katarinabatina in lieu of a design - is this alright? It sits under the bio

![screen shot 2016-10-19 at 18 22 56](https://cloud.githubusercontent.com/assets/49038/19529743/1b1f60cc-9629-11e6-8832-f66d192378c9.png)

katarinabatina:
👍

alloy:
Fixed by #1950

PR #1950, opened 2016-10-19
Artist Auction  Results

orta:
For https://github.com/artsy/eigen/issues/1468 and https://github.com/artsy/emission/pull/367


alloy:
👍

Pods
```diff
@@ -1 +1 @@
-Subproject commit 32c0c40daf850448ae99cebee42b52d4bd6e50bd
+Subproject commit 78cd2a4fb2a3b7a9e3efc7f9fcf6c580b5ee8df9
```
orta: hrm, I feel like this wasn't meant to happen, but we should be fine, CI would use the Podfile anyway

PR #1951, opened 2016-10-19
Removes ARAnalyticsPapertrail provider

### LINKED ISSUE #1852: [Onboarding] What to do with local cache of trial users
@alloy commented before:

> Since 2.4.0 we have also been keeping track of all the artists/artworks/categories a trial user visits in a local cache file, we can use that data to improve suggestions.
1 Need to determine what the best way is to do that. E.g. sorts artists by number of artworks viewed of that artist?
2 The code that does the collection of that data needs to be removed once the new on-boarding is released.

I'm assuming I may have deleted the data collection bit of the code in the trial code 🔥, but want to double check and in general, make sure we're firm on how we want to proceed.

ashfurrow:
Fixes #1852.


alloy:
So I take it that the code that initialised the `ARAnalyticsPapertrail` instance was already gone?

ashfurrow:
> the code that initialised the ARAnalyticsPapertrail instance was already gone?

Yup!

PR #1952, opened 2016-10-19
Status bar background color for Home #trivial

sarahscott:
This changes the status bar color to white at 98% opacity on the Home view and leaves in unchanged elsewhere in eigen.

### DIFF
--- a/Artsy/View_Controllers/Util/ARNavigationController.h
+++ b/Artsy/View_Controllers/Util/ARNavigationController.h
@@ -15,7 +15,7 @@
 
 - (void)showBackButton:(BOOL)visible animated:(BOOL)animated;
 - (void)showStatusBar:(BOOL)visible animated:(BOOL)animated;
-- (void)showStatusBarBackground:(BOOL)visible animated:(BOOL)animated;
+- (void)showStatusBarBackground:(BOOL)visible animated:(BOOL)animated isHome:(BOOL)isHome;
 
 - (IBAction)back:(id)sender;
 
--- a/Artsy/View_Controllers/Util/ARNavigationController.m
+++ b/Artsy/View_Controllers/Util/ARNavigationController.m
@@ -24,6 +24,7 @@
 #import <FLKAutoLayout/UIView+FLKAutoLayout.h>
 #import <ObjectiveSugar/ObjectiveSugar.h>
 #import <MultiDelegate/AIMultiDelegate.h>
+#import <Emission/ARHomeComponentViewController.h>
 
 static void *ARNavigationControllerButtonStateContext = &ARNavigationControllerButtonStateContext;
 static void *ARNavigationControllerScrollingChiefContext = &ARNavigationControllerScrollingChiefContext;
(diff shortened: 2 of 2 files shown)


sarahscott:
Greens locally 💚

Artsy/View_Controllers/Util/ARNavigationController.h
```diff
@@ -15,7 +15,7 @@
 
 - (void)showBackButton:(BOOL)visible animated:(BOOL)animated;
 - (void)showStatusBar:(BOOL)visible animated:(BOOL)animated;
-- (void)showStatusBarBackground:(BOOL)visible animated:(BOOL)animated;
+- (void)showStatusBarBackground:(BOOL)visible animated:(BOOL)animated isHome:(BOOL)isHome;
```
orta: This should probably be about being white instead of being home-view specific, it's not really the responsibility of the nav to know the view controllers it's showing

Artsy/View_Controllers/Util/ARNavigationController.m
```diff
@@ -316,10 +321,15 @@ - (void)showStatusBar:(BOOL)visible animated:(BOOL)animated
...
+    CGFloat visibleAlpha = isHome ? 0.98 : 1;
+    UIStatusBarStyle statusBarStyle = isHome ? UIStatusBarStyleDefault : UIStatusBarStyleLightContent;
+    [[UIApplication sharedApplication] setStatusBarStyle:statusBarStyle];
+
     [UIView animateIf:animated duration:ARAnimationDuration:^{
-        self.statusBarView.alpha = visible ? 1 : 0;
+        self.statusBarView.backgroundColor = isHome ? UIColor.whiteColor : UIColor.blackColor;
+        self.statusBarView.alpha = visible ? visibleAlpha : 0;
     }];
 }
```
orta: I worry about this PR, if it's done for speed that's fine, but this really should be an extension on the [ARMenuAware protocol](https://github.com/artsy/eigen/blob/master/Artsy/View_Controllers/Util/ARMenuAwareViewController.h#L1) like all the other customisations to a menu - perhaps a property like `@property (readonly, nonatomic, assign) BOOL useOpaqueStatusBarBackground;` which you can add to the `ARHomeComponentViewController` via a category

alloy: This is only done for pragmatic reasons. The real end goal is to have the same white status bar everywhere (minus e.g. the search VC), but doing so now was making things too complicated and would have required more time to get right, so best to not cause unintended side-effects.

Artsy/View_Controllers/Util/ARNavigationController.h
alloy: Neither ‘home’ or ‘white’ should be in the eventual final code-base, it should just be white everywhere.

Artsy/View_Controllers/Util/ARNavigationController.m
sarahscott: Yes. Ideally in the next version, I (or someone else) can overhaul status bar styling and make it easier to digest and make changes to. I'm a big fan of @ashfurrow's `ARSearchViewControllerStyling` class and think that would make a lot more sense for this sort of thing than appending a bunch of methods to `ARNavigationController`.

orta: Alright, I just don't want to see this sticking around too long, you know how these decisions can end up lasting far longer than anticipated

alloy: Aye, agreed #1955

PR #1963, opened 2016-10-21
Bump bundle version.

alloy:
Currently doing a deploy of this version. Will merge into master if we go through with replacing the release by this.


orta:
We did 👍

PR #1972, opened 2016-10-26
[Fix] Onboarding crashes

mennenia:
My first merge into master in _months_!!

Fix #1971  #1968

PR #1981, opened 2016-10-28
[Fix] for nil crash in Gene Image State Reconciler

### LINKED ISSUE #1980: Fix NSInvalidArgumentException in - [AROnboardingPersonalizationGeneImageStateReconciler addReplacedGene:], line 37
### Stacktrace (first of multiple) ###

<pre>AROnboardingPersonalizationGeneImageStateReconciler;addReplacedGene:;AROnboardingPersonalizationGeneImageStateReconciler.m;37
AROnboardingPersonalizeTableViewController;updateTableContentsFor:replaceContents:animated:;AROnboardingPersonalizeTableViewController.m;73
ARPersonalizeViewController;categoryFollowed:;ARPersonalizeViewController.m;316</pre>

### Reason ###

*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** -[__NSSetM addObject:]: object cannot be nil'

### Link to HockeyApp ###

* [https://rink.hockeyapp.net/manage/apps/37029/crash_reasons/142117442](https://rink.hockeyapp.net/manage/apps/37029/crash_reasons/142117442)

mennenia:
Fix #1980

### DIFF
--- a/Artsy/View_Controllers/Login_and_Onboarding/Onboarding_stages/5-_Onboarding_Personalization/AROnboardingPersonalizeTableViewController.m
+++ b/Artsy/View_Controllers/Login_and_Onboarding/Onboarding_stages/5-_Onboarding_Personalization/AROnboardingPersonalizeTableViewController.m
@@ -66,12 +66,11 @@
         case ARSearchResultsReplaceSingle:
             if (searchResults) {
                 [self.searchResults replaceObjectAtIndex:self.tableView.indexPathForSelectedRow.row withObject:searchResults[0]];
+                self.selectedRowToReplace = self.tableView.indexPathForSelectedRow;
+                [self.geneImageReconciler addReplacedGene:self.selectedRowToReplace];
             } else {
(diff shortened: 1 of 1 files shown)


orta:
Looks legit from my perspective 👍

alloy:
Wondering if it’s wise to also add some nullability type info to these cases as we run into them? E.g. label the `selectedRowToReplace` property to be nullable and the reconciler API to not accept `nil`?

mennenia:
I was thinking of modifying the reconciler API, but essentially the issue here was that sometimes there's no new row to replace. 

So I had already fixed this in an earlier PR (#1972) and it made sense to move the setting of the new image up into that block. No need to use it if you're just removing a row, and it could still be nil at that point. 

This felt like the right approach but yes we could add stricter checks all around.

alloy:
Let‘s get this merged, not considering the nullability types a blocker.

mennenia:
Lol was just about to add them (as the spinner PR is now done), but sure. Let's merge for now to ship. I've put them on my todo list!

alloy:
Ok, taking this one

PR #1999, opened 2016-11-04
Fix for searchVC routing to old geneVC #tiny

### LINKED ISSUE #1995: SearchVC routes to old gene VC instead of RN version
Selecting a gene from Home will push a RN GeneVC, but searching for one will push the old version.

sarahscott:
Is this part still necessary?

https://github.com/artsy/eigen/blob/master/Artsy/App/ARSwitchboard+Eigen.m#L140

Is there ever going to be a case where we want to be using the old version of this VC?

Fixes #1995


alloy:
I think we should leave it in at least until we have enabled the refine features _and_ are happy with it. The point of this blacklisting is to be able to workaround the case where the new VC appears to be massively breaking in production.

PR #2006, opened 2016-11-09
[Onboarding] Fix crash that occurs when de-selecting in a tableview.

### LINKED ISSUE #1994: Fix NSInvalidArgumentException in - [AROnboardingPersonalizationGeneImageStateReconciler addReplacedGene:], line 37
### Stacktrace ###

<pre>AROnboardingPersonalizationGeneImageStateReconciler;addReplacedGene:;AROnboardingPersonalizationGeneImageStateReconciler.m;37
AROnboardingPersonalizeTableViewController;updateTableContentsFor:replaceContents:animated:;AROnboardingPersonalizeTableViewController.m;70
ARPersonalizeViewController;artistFollowed:;ARPersonalizeViewController.m;279</pre>

### Reason ###

*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** -[__NSSetM addObject:]: object cannot be nil'

### Link to HockeyApp ###

* [https://rink.hockeyapp.net/manage/apps/37029/crash_reasons/143301624](https://rink.hockeyapp.net/manage/apps/37029/crash_reasons/143301624)

alloy:
Fixes #1994.

Alas, even with nullability info Xcode did not flag that line as being a risk, not sure what’s up with that, so doing this based purely on the crash log.

### DIFF
--- a/Artsy/View_Controllers/Login_and_Onboarding/Onboarding_stages/5-_Onboarding_Personalization/AROnboardingPersonalizationGeneImageStateReconciler.h
+++ b/Artsy/View_Controllers/Login_and_Onboarding/Onboarding_stages/5-_Onboarding_Personalization/AROnboardingPersonalizationGeneImageStateReconciler.h
@@ -2,7 +2,7 @@
 
 @class Gene;
 
-
+NS_ASSUME_NONNULL_BEGIN
 @interface AROnboardingPersonalizationGeneImageStateReconciler : NSObject
 
 - (NSURL *)imageURLForGene:(Gene *)gene atIndexPath:(NSIndexPath *)indexPath;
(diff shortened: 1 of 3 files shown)


mennenia:
I've added one more change - the reason `self.selectedRowToReplace` _can_ be nil, is because in the animation where the cell fades out, it's set to nil because there's nothing selected anymore. This was outside the completion block though, so I think that accounts for why the situation can happen in the first place.

Thanks @alloy for adding the guards!

sarahscott:
Seeing as the circle fail is a Danger thing and not the actual tests, I'm going to merge this so it can definitely be in our release tomorrow. I will make a ticket for the Danger issue.

alloy:
@mennenia Nice catch 👏

PR #2014, opened 2016-11-15
Fixes problem with users registered for too many sales

ashfurrow:
Problem: user sees "Register to Bid" instead of a "Bid" button.

Cause: we were asking for _all_ bidder models for all sales, instead of limiting to our current sale. The models appears to be ordered by oldest first, so after the first page we don't get the model we need to recognize that the user is registered for the sale.

Fix for https://github.com/artsy/auctions/issues/218.


alloy:
Looks good 👍 CI fail is due to danger dep + ruby.

PR #2015, opened 2016-11-15
Add support for fb app install tracking

orta:
I've not considered all the ramifications, but Danger was in the shower and I had a bit of time, so if this greens - I would recommend maybe putting it in after the next release so that it gets some time to simmer.


alloy:
Yeah not merging this now, the adding code to Eigen part is just a very small change, checking if it all works as expected and is properly setup probably takes most time.

alloy:
This was not needed at all, it appears that Adjust could pass that info on to FB.

PR #2016, opened 2016-11-16
Removes old subscriptions to improve lot list scrolling performance

ashfurrow:
Fix for https://github.com/artsy/auctions/issues/181.

Explanation: a lot's state is a mapping of its bidding status, which gets [updated](https://github.com/artsy/eigen/blob/master/Artsy/View_Controllers/Live_Auctions/LiveActionStateReconciler.swift#L139) every time there's a new derived lot data (which happens on every related lot broadcast, which happens very very often). We weren't unsubscribing from old subscriptions, leading to hundreds of callbacks _per lot_ in a list of hundreds of lots. 

This issue is difficult to debug because it only occurs on sales with many lots, so I'll have to monitor an auction later to QA.

### DIFF
--- a/Artsy/View_Controllers/Live_Auctions/LiveAuctionLotListViewController.swift
+++ b/Artsy/View_Controllers/Live_Auctions/LiveAuctionLotListViewController.swift
@@ -44,6 +44,8 @@ class LiveAuctionLotListViewController: UICollectionViewController {
                 return sSelf.stickyCollectionViewLayout.setActiveIndex(nil)
             }
 
+            sSelf.unsubscribeCurrentLotState()
+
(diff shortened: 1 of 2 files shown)


orta:
Conceptually looks good, would definitely require manual Q - aye 👍  (merge conflict BTW )

alloy:
> would definitely require manual QA

@ashfurrow I’m unsure how impactful this is and @orta’s comment makes it seem as though it might be. Are you confident enough in the change to include this in the 2nd beta we’re doing today/tomorrow?

ashfurrow:
I am, I tested locally and it works 👍

alloy:
Works for me 👍

PR #2020, opened 2016-11-17
[!] [Onboarding] Fix code using old delegate method, resulting in crash

mennenia:
Sorry everyone... this needs to go in for 3.0.2! 

In the clean up around Facebook's signup process, I removed methods we weren't using anymore / didn't do anything, but the email signup method still called it. 

Not sure if I should changelog this, as it hasn't officially gone out yet?

### DIFF
--- a/Artsy/View_Controllers/Login_and_Onboarding/AROnboardingViewController.h
+++ b/Artsy/View_Controllers/Login_and_Onboarding/AROnboardingViewController.h
@@ -17,7 +17,6 @@
 @end
 
 @protocol ARLoginSignupDelegate <NSObject>
-- (void)didSignUpAndLogin;
 - (void)dismissOnboardingWithVoidAnimation:(BOOL)animated;
 @end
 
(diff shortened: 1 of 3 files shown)


orta:
I'd put it in the changelog for master ?

Artsy/View_Controllers/Login_and_Onboarding/AROnboardingViewController.h
```diff
@@ -17,7 +17,6 @@
 @end
 
 @protocol ARLoginSignupDelegate <NSObject>
-- (void)didSignUpAndLogin;
```
alloy: Ah, I was wondering why this issues wasn’t caught at compilation time, I guess this being left in the protocol is why?

alloy:
@mennenia Yeah this is going to go in 3.0.2 still 👍 

@orta What does that mean? I’m confused about how the changelog works.

orta:
I meant upcoming, as this won't be the real 3.0.2 build

mennenia:
@alloy because it was still in the header >_< my bad

alloy:
I’m going to merge this as-is.

PR #2029, opened 2016-11-21
Fixes auction countdown timer issue on iPad

ashfurrow:
Fix for https://github.com/artsy/auctions/issues/185 . There just wasn't enough space on either side of the countdown timer, autolayout was clipping individual labels.


orta:
👍

PR #2031, opened 2016-11-21
Fixes auctions with sale artworks that are missing high estimates.

ashfurrow:
Fix for https://github.com/artsy/auctions/issues/178. Mantle was choking on `null` JSON fields and primitive `UInt64` types. This uses `NSNumber` instead to avoid future problems.

### DIFF
--- a/Artsy/Models/API_Models/LiveAuctionLot.h
+++ b/Artsy/Models/API_Models/LiveAuctionLot.h
@@ -45,8 +45,8 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, copy, readonly) NSString *currency;
 @property (nonatomic, copy, readonly) NSString *currencySymbol;
 
-@property (nonatomic, assign, readonly) UInt64 lowEstimateCents;
-@property (nonatomic, assign, readonly) UInt64 highEstimateCents;
+@property (nonatomic, assign, readonly) NSNumber *_Nullable lowEstimateCents;
+@property (nonatomic, assign, readonly) NSNumber *_Nullable highEstimateCents;
 @property (nonatomic, copy, readonly) NSString *_Nullable estimate;
 @property (nonatomic, assign, readonly) UInt64 askingPriceCents;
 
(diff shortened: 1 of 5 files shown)


Artsy_Tests/View_Controller_Tests/Live_Auction/LiveAuctionLotViewControllerTests.swift
```diff
@@ -144,7 +144,7 @@ class Test_LiveAuctionLotViewModel: LiveAuctionLotViewModelType {
     var numberOfDerivedEvents = 1
     var lotIndex = "2"
     var lotNumber = "2"
-    var highEstimateCents: UInt64 = 2_000_00
+    var highEstimateCents: UInt64? = 2_000_00
```
ashfurrow: There's no test, though the compiler will produces errors if it's not adhered to (which is why we had to make this change: https://github.com/artsy/eigen/pull/2031/files#diff-276f6d21e2b6eb6da7a795cca0282777R63).

sweir27: 👍

Artsy/Models/API_Models/LiveAuctionLot.h
```diff
@@ -45,8 +45,8 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, copy, readonly) NSString *currency;
 @property (nonatomic, copy, readonly) NSString *currencySymbol;
 
-@property (nonatomic, assign, readonly) UInt64 lowEstimateCents;
-@property (nonatomic, assign, readonly) UInt64 highEstimateCents;
+@property (nonatomic, assign, readonly) NSNumber *_Nullable lowEstimateCents;
+@property (nonatomic, assign, readonly) NSNumber *_Nullable highEstimateCents;
```
ashfurrow: Neat! Though I just noticed these are still `assign`, I'll update them to be strong references.

alloy: It’s all just sugar afaik, just an FYI.

PR #2043, opened 2016-11-29
Updated project and target settings for new automatic code signing

sarahscott:
This lets Xcode handle all signing and removes provisioning profile settings because they are now deprecated.


orta:
merging because I think this might conflict with my Apple Watch PR

PR #2049, opened 2016-11-30
Fixes issue where links in auction descriptions would lead nowhere wh…

ashfurrow:
Fixes https://github.com/artsy/auctions/issues/259. @orta you've worked with `ARTextView` more than I have, can you double-check my work here?


orta:
Yep, that is how I remember the API being 👍

PR #2052, opened 2016-12-01
Fixes assign/strong mixup #trivial

ashfurrow:
When I added `nullable`, I got a compile warning about duplicate nullability specifiers, so it looks like the behaviour is the same. Is this something we want to standardize on?

### DIFF
--- a/Artsy/Models/API_Models/LiveAuctionLot.h
+++ b/Artsy/Models/API_Models/LiveAuctionLot.h
@@ -45,8 +45,8 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, copy, readonly) NSString *currency;
 @property (nonatomic, copy, readonly) NSString *currencySymbol;
 
-@property (nonatomic, assign, readonly) NSNumber *_Nullable lowEstimateCents;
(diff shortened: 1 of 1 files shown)


alloy:
The same as what specifically?

ashfurrow:
I just meant that the following lines are semantically equivalent:

```objc
@property (nonatomic, assign, readonly) NSNumber *_Nullable lowEstimateCents;
@property (nonatomic, assign, readonly, nullable) NSNumber *lowEstimateCents;
```

And we currently have both throughout eigen.

alloy:
Ah yup yup, that’s correct. The `nullable` attribute was added more recent, I think. I think we should standardise on it, yeah.

ashfurrow:
Do you want that in a distinct PR? Or want it added to this one?

alloy:
If you’re planning to do a sweep now, then let’s add it to this PR; if not, that’s perfectly fine and I’ll merge now.

ashfurrow:
Thanks, I wasn't planning on it now, let's merge 👍

alloy:
Aiiiiiight 👍

PR #2067, opened 2016-12-07
Fixes flash of text when pushing auction view controllers.

ashfurrow:
Pretty simple fix, we were just disregarding the `animated` property.

Fix for https://github.com/artsy/auctions/issues/260.


mennenia:
Coool 👍 

Waiting for 🍏, then 💃

PR #2112, opened 2017-01-17
Adds APNS entitlements and privacy Info.plist specifiers

ashfurrow:
Tried to be as vague as possible without being meaningless.


orta:
Looks good to go on green 👍

PR #2116, opened 2017-01-18
Fixes live auction / contact buttons problem.

### LINKED ISSUE #2115: Feedback 'From: Ash Need to hide "contact gallery" if the...'
### Message from Ash Furrow sent on 2017-01-17 22:02:36 UTC###

> From: Ash
> 
> Need to hide "contact gallery" if the auction is live. This is related to the changes that retire end_at from eigen.

### Link to HockeyApp ###

* [https://rink.hockeyapp.net/manage/apps/37020/feedback/706807](https://rink.hockeyapp.net/manage/apps/37020/feedback/706807)

ashfurrow:
Fixes #2115. /cc @sweir27 

Problem caused as a result of an oversight when initially implementing this logic. The file to manage all this is quite large, I think a factoring the logic to decide which button(s) to show into its own class might be a good idea the next time any substantial work needs to be done here. Happy to take that on.

### DIFF
--- a/Artsy/Views/Artwork/ARArtworkActionsView.m
+++ b/Artsy/Views/Artwork/ARArtworkActionsView.m
@@ -345,7 +345,7 @@ return [navigationButtons copy];
 // Show if artwork is for sale and inquireable, but not acquireable and not in an auction
 - (BOOL)showContactButton
 {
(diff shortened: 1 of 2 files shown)


mennenia:
Cool!

PR #2120, opened 2017-01-20
Fix for double sliding animation in artworkVC

### LINKED ISSUE #2117: Fix double animation when pushing artwork VC
This seems to happen when a user tries to push an artworkVC from any view (RN, cocoa, or martsy)
To repro:
- Go to Home tab
- Tap any artwork
or
- Go to Explore tab
- Tap Artworks
- Tap any artwork
or 
- Go to You tab
- Tap any artwork

sarahscott:
I haven't done any work on the ArtworkVC animations so I'm not sure what the ideal is supposed to be - is it supposed to slide up from the bottom? In any case, the current app store version has the VC slide in from right with a gray artwork placeholder before loading the image. For some reason in the iOS10/Xcode8-compiled version, the transition was adding an extra slide. 

Was the slide supposed to be there all along and got lost in the shuffle? If so, is it worth it to try and make it work again, or leave this as-is? 

With redundant animation:

broken

With this PR, and also what the App Store version currently does:

fixed

Fixes #2117

### DIFF
--- a/Artsy/View_Controllers/Artwork/ARArtworkViewController.m
+++ b/Artsy/View_Controllers/Artwork/ARArtworkViewController.m
@@ -17,6 +17,7 @@
 
 #import <UIView_BooleanAnimations/UIView+BooleanAnimations.h>
 
+
 @interface ARArtworkViewController () <UIScrollViewDelegate, ARArtworkRelatedArtworksViewParentViewController, ARArtworkBlurbViewDelegate, ARPostsViewControllerDelegate>
 
 @property (nonatomic, strong) ARArtworkView *view;
@@ -83,7 +84,7 @@
         [self ar_removeIndeterminateLoadingIndicatorAnimated:ARPerformWorkAsynchronously];
     }
 
-    __weak typeof (self) wself = self;
+    __weak typeof(self) wself = self;
 
     void (^completion)(void) = ^{
         __strong typeof (wself) sself = wself;
@@ -147,7 +148,7 @@
 
 - (void)getRelatedPosts
 {
-    __weak typeof (self) wself = self;
+    __weak typeof(self) wself = self;
     [self.artwork getRelatedPosts:^(NSArray *posts) {
         __strong typeof (wself) sself = wself;
         [sself updateWithRelatedPosts:posts];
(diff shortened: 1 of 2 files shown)


mennenia:
It should indeed be from the side, because it's part of a navigation controller hierarchy stack :) So as soon as there is a top left back button, often the animation for the next VC will slide in from the right. 

An animation coming from the bottom up is more consistent when the way to dismiss it is in the top right, often an X or "close" or "dismiss" label.

mennenia:
Cool! This is awesome :D.

Gave it a go on device, looks great! IMHO it feels slightly slow, as if i'm expecting a speed up in the beginning of the animation curve, but it's the exact same on the store version 😅. 

@sarahscott Do you want to add a change log entry?

sarahscott:
@mennenia ah yeah that makes a lot of sense. Any ideas about the reason for the `animateTwoStep`? I'm sort of scratching my head wondering if I've removed anything interesting.

sarahscott:
I guess it would make the related artworks section fade in sort of nicely but if you try to animated `setNeedsLayout` and `layoutAsNeeded` at all, the weird slide-from-bottom animation bug happens.

mennenia:
hmm, well I guess that's an underlying change with the animation and layout cycles now or something, or a genuine Apple bug? We could check radars. 

I'm not sure why the two step is there, but it seems to be about first animating the space changes, and then fading in the section, which is nice. Now it would just show you the section. From what I can see, these are about additional bits of info further down? Not the primary image and information.. so I'm thinking that it's good to go if it fixes our bug. 

@orta @alloy any ideas? Might be worth for you to have a check too, before merging. 

(other than double checking the legacy, the PR is 👍, @sarahscott )

alloy:
> I guess it would make the related artworks section fade in sort of nicely

Yes, that was the purpose.

Maybe it has to do with the two combined animations? Maybe it should be something like:

1. The VC starts loading related works
2. If the works request comes back:
  * and the VC is **not** yet shown, do it without an animation as your PR does.
  * if the VC is already shown, use the animation.

alloy:
@sarahscott did you capture the above in a new ticket?

sarahscott:
Aye: https://github.com/artsy/eigen/issues/2127

PR #2121, opened 2017-01-20
Live messaging fix 📱

ashfurrow:
This is the eigen side of the work for https://github.com/artsy/auctions/issues/289 . Here's a breakdown of the problem:

- We were showing "sold to you for X" but that's not the correct messaging. The correct messaging is "sold for X" and only to show it if the user is the winning bidder.
- We were showing the "current price" aka the next increment above the highest bid. The nomenclature around this in eigen could use a revamp, but for now I've added a `winningLotValueString` that returns the bid amount of the winning bid event, if it exists.

I also added a test for the case that the user won the lot, which we didn't have for this particular view yet.

This is a WIP because I'm currently fighting with Xcode to get this tested on my device first.


ashfurrow:
Tested this on device, looks 👍

screen shot 2017-01-20 at 4 06 33 pm

PR #2130, opened 2017-01-24
Fixes problem with current lot CTA in LAI 📱

ashfurrow:
Fixes https://github.com/artsy/auctions/issues/294

The gist is: we had an object taking responsibility for managing the position of the CTA but when the current lot and/or lot the user is looking at change, the order isn't guaranteed and some weird race conditions were happening. I had to give up the observable abstraction within `LiveAuctionCurrentLotCTAPositionManager` and instead rely on functions called directly from the lot set view controller.

I also deleted two functions that were dead code.


orta:
👍

orta:
fails are released though

ashfurrow:
Aye, will get to those tomorrow 👍

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">You need to set the App’s plist version to 3.1.1</td>
    </tr>
  </tbody>
</table>

<p align="right" data-meta="generated_by_danger">
  Generated by :no_entry_sign: <a href="http://danger.systems/">danger</a>
</p>

ashfurrow:
That danger error will be fixed in https://github.com/artsy/eigen/pull/2131 . This should be good to review whenever (it's been deployed already).

PR #2198, opened 2017-03-03
Fixes "Artist at Artsy" and "Show link" buttons

### LINKED ISSUE #2195: Navigation buttons not showing properly in Fair -> Artist details
In Fair -> Artists -> Artist detail, buttons linking to artwork and "Artist at Artsy" look like this:

The `ARNavigationButton`s are directly added to the `stackView`, which is different from all other places that use `ARNavigationButton`. Seems like the common usage is to add the button in a `ARNavigationButtonsViewController` and then add the view to `stackView`. (For example, the ["More Info" button in Artwork](https://github.com/artsy/eigen/blob/master/Artsy/Views/Artwork/ARArtworkActionsView.m#L244)). In fact, the only other place I could find that is using `ARNavigationButton` directly is `ARArtistViewController`, which was replaced by an Emission component. 

As a test, I changed the Artist at Artsy button to use `ARNavigationButtonsViewController`:

```objective-c
- (void)addArtistOnArtsyButton
{
    NSString *title = NSStringWithFormat(@"%@ on Artsy", self.artist.name);

    NSArray *buttonDescription = @[ @{
                               ARNavigationButtonClassKey : ARNavigationButton.class,
                               ARNavigationButtonPropertiesKey : @{
                                       ar_keypath(ARNavigationButton.new, title) : title
                                       },
                               ARNavigationButtonHandlerKey : ^(UIButton *sender) {
                                   UIViewController *viewController = [ARSwitchBoard.sharedInstance loadArtistWithID:self.artist.artistID inFair:nil];
                                   [self.navigationController pushViewController:viewController animated:ARPerformWorkAsynchronously];
                               }
                            }];

    self.artistOnArtsyVC = [[ARNavigationButtonsViewController alloc] init];
    [self.artistOnArtsyVC addButtonDescriptions:buttonDescription unique:YES];
    self.artistOnArtsyVC.view.tag = ARFairArtistOnArtsy;

    [self.view.stackView addSubview:self.artistOnArtsyVC.view withTopMargin:@"20" sideMargin:@"40"];
}
```

which fixed the issue (and the button still works :smile:):

For the button that links to an artwork however, I am not sure how to proceed. It would require having an `ARNavigationButtonsViewController` for each artwork, which makes me think that maybe the whole thing should live in its own view controller (button, image and description). This is a bit of extra work, so I'd like to confirm the approach is good before starting on it.

On a side note, I tried a few things to make the buttons to show properly without needing `ARNavigationButtonsViewController`, but no success. Curious to know if I missed an easier solution 🤔

pedrovereza:
Fixes #2195 

As discussed in the issue, both buttons were wrapped in an `ARNavigationButtonsViewController`. Not sure if the buttons could have been added in a cleaner way 🤔 (feedback is welcome).

End result:

Also, @orta noticed that the snapshot test for this screen [is commented out](https://github.com/artsy/eigen/issues/2195#issuecomment-282342629). I was not able to run the tests locally, so I wasn't confident bringing the test back myself.

### DIFF
--- a/Artsy/View_Controllers/Fair/ARFairArtistViewController.m
+++ b/Artsy/View_Controllers/Fair/ARFairArtistViewController.m
@@ -131,13 +131,22 @@ AR_VC_OVERRIDE_SUPER_DESIGNATED_INITIALIZERS;
 - (void)addArtistOnArtsyButton
 {
     NSString *title = NSStringWithFormat(@"%@ on Artsy", self.artist.name);
-    ARNavigationButton *button = [[ARNavigationButton alloc] initWithTitle:title];
-    button.tag = ARFairArtistOnArtsy;
-    button.onTap = ^(UIButton *tappedButton) {
-        UIViewController *viewController = [ARSwitchBoard.sharedInstance loadArtistWithID:self.artist.artistID inFair:nil];
-        [self.navigationController pushViewController:viewController animated:ARPerformWorkAsynchronously];
-    };
-    [self.view.stackView addSubview:button withTopMargin:@"20" sideMargin:@"40"];
+
+    NSArray *artistOnArtsyButtonDescription = @[@{ ARNavigationButtonClassKey : ARNavigationButton.class,
+                                                   ARNavigationButtonPropertiesKey : @{
+                                                           ar_keypath(ARNavigationButton.new, title) : title
+                                                           },
+                                                   ARNavigationButtonHandlerKey : ^(UIButton *sender) {
+                                                       UIViewController *viewController = [ARSwitchBoard.sharedInstance loadArtistWithID:self.artist.artistID inFair:nil];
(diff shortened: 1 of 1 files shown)


pedrovereza:
cc. @orta @alloy

alloy:
Thanks a whole lot for taking the time @pedrovereza! 🙌 

Unfortunately we’ve been dealing with some other issues that make master unstable, so I’m waiting till we’ve had the time to deal with that (hopefully this week) before merging.

alloy:
@pedrovereza Thanks so much for the work! You are my personal hero of the day 👏 

I’ve updated your PR slightly in #2232, so closing this one in favour of that.

orta:
Nice one  - sorry, been on my honeymoon 👯‍♂️

Issue #2224, opened 2017-03-20
Feedback 'From: Ash Bottom 2 pixels of refine slider is c...'


artsyit:
### Message from Ash Furrow sent on 2017-03-20 17:34:07 UTC###

> From: Ash
> 
> Bottom 2 pixels of refine slider is cut off.

### Link to HockeyApp ###

* [https://rink.hockeyapp.net/manage/apps/37020/feedback/764963](https://rink.hockeyapp.net/manage/apps/37020/feedback/764963)

ashfurrow:
![asset png-2](https://cloud.githubusercontent.com/assets/498212/24113098/0c163e6e-0d72-11e7-9223-1fa343ee557d.jpeg)

PR #2225, opened 2017-03-20
Fixes when the refine price range sliders get cut off

ashfurrow:
They were getting cut off by 1 point, so I added a spacer. 

Fixes https://github.com/artsy/eigen/issues/2224


l2succes:
👍 LGTM!

PR #2229, opened 2017-03-22
[Adjust] Send `anonymous_id` to partner integrations (Segment). #trivial

alloy:
For https://github.com/artsy/collector-experience/issues/124

From the email thread by an Adjust employee:

> In case a customer wants to use the Segment module within Adjust, they must implement a Session Partner Parameter whose key is `anonymous_id`. 

So if I understood it correctly, this is the only required change on our side.


wrgoldstein:
👍

PR #2238, opened 2017-03-23
[Search] Open fair VC.

### LINKED ISSUE #2236: Fair Search Results doesn't link to fair microsite
Fairs are now appearing in the search results, however, as a user when you search for a fair and go to click on one of the search results it does not take you to the fair's microsite.

alloy:
Fixes #2236.


xtina-starr:
Looks like you just added a conditional to handle when the result is the fair class. 

LGTM.

PR #1049, opened 2016-01-21
Support iOS again, and support gracefully deprecating the platform when we choose 

orta:
fixes #1048 
fixes #1047 

Adds the same support we added for iOS7 back in #198

PR #2240, opened 2017-03-24
Eliminate semantic issue

y310:
It seems that there are several semantic issue warnings.
In this PR, I eliminated those issues as much as possible.

I just left following warning which was introduced by https://github.com/artsy/eigen/pull/1049 since I'm not sure which is better just removing or updating to something `checkForiOS9Deprecation`.

```
eigen/Artsy/App/ARAppDelegate.m:440:10: Comparison of address of 'UIApplicationOpenURLOptionsAnnotationKey' not equal to a null pointer is always true
```

### DIFF
--- a/Artsy/App/ARAppActivityContinuationDelegate.m
+++ b/Artsy/App/ARAppActivityContinuationDelegate.m
@@ -8,27 +8,22 @@
 
 #import <CoreSpotlight/CoreSpotlight.h>
 
-// Only available on iOS 9.
-static BOOL IsSpotlightActionTypeAvailable = NO;
-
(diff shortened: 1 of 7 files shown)


orta:
Started running this locally, I needed a tea anyway 🍵

orta:
OK, `Executed 1067 tests, with 0 failures (0 unexpected) in 50.123 (52.149) seconds` 👍 

Looks good to me - thanks @y310

orta:
We can live with the `UIApplicationOpenURLOptionsAnnotationKey` warning

PR #2245, opened 2017-03-28
Fixes Danger sticker version check

ashfurrow:
Popped the check into irb it was returning `Error Reading File: Artsy` and failing on all PRs. Spaces-in-file-paths strikes again!


orta:
Hah - yah

PR #2246, opened 2017-03-28
Switched from strings to enums

ashfurrow:
Small improvement, I'll see if others come to mind.


ashfurrow:
Failures are legit, will investigate tomorrow.

ashfurrow:
Great question! This mainly solves my own desire for ideal type safety. Using strings works fine but isn't as idiomatic as it could be.

PR #2247, opened 2017-03-28
Fixes a problem formatting small numbers of currencies

ashfurrow:
As discussed in Slack: https://artsy.slack.com/archives/C02BAQ5K7/p1490726395658278


orta:
@ashfurrow fail is legit

alloy:
I don’t get it, the number formatter is configured with `currencyFormatter.maximumFractionDigits = 0;`

alloy:
So yeah, no clue why the `currencyConverter` still adds decimals to $5. If you aren’t able to figure out why that happens, then I’m fine with merging the code as is and track down the issue afterwards. I’d really like to release tomorrow.

ashfurrow:
Yeah I'm confused as well. I'll try to get this wrapped up tonight.

ashfurrow:
For folks following along online, the number formatter is configured here: https://github.com/artsy/eigen/blob/759124707a92d577a62de221756838a0f60ed716/Artsy/Models/API_Models/Auctions/SaleArtwork.m#L17-L27 .

ashfurrow:
Okay, so I've moved the number formatter static variable into the method, and the same thing still happens. Messing around with the tests and test order, I believe that changing `currencySymbol` on the instance causes it to behave weirdly. After changing from the pound to dollar signs, this happens:

screen shot 2017-03-28 at 6 40 41 pm

I expect this could happen during the app lifecycle, but hasn't because we con't change currencies very often.

So instead I'm going to cache the number formatter (they're expensive to initialize, for whatever reason) until the symbol we're formatting changes.

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">There were <a href="https://eigen-ci.s3.amazonaws.com/3960/index.html">snapshot errors</a></td>
    </tr>
  </tbody>
</table>

<p align="right" data-meta="generated_by_danger">
  Generated by :no_entry_sign: <a href="http://danger.systems/">danger</a>
</p>

ashfurrow:
So here's a before/after of the snapshot test failure:

screen shot 2017-03-28 at 7 30 52 pm

I don't know why the reference image isn't displaying a price, I'm guess it had something to do with the old strange behaviour? Since the "failed" image looks correct, and matches the pattern established by similar tests' reference images, I'm going to replace the reference image. 

@orta @alloy I'm signing off for the night, feel free to merge or cherry-pick commits in the Euro-morning.

Artsy/Models/API_Models/Auctions/SaleArtwork.m
```diff
@@ -138,19 +124,22 @@ - (BidderPosition *)userMaxBidderPosition
...
+        // Changing currencySymbol on NSNumberFormatter causes it to behave weirdly...
+        // So we recreate it if the instance is nil or if the symbol has changed from the last invocations.
+        // See https://github.com/artsy/eigen/pull/2247#issuecomment-289928233
+        currencyFormatter = [[NSNumberFormatter alloc] init];
+        currencyFormatter.currencySymbol = symbol;
+        currencyFormatter.numberStyle = NSNumberFormatterCurrencyStyle;
+        currencyFormatter.currencyGroupingSeparator = @",";
+        currencyFormatter.maximumFractionDigits = 0;
+        currencyFormatter.minimumFractionDigits = 0;
     }
```
alloy: Are we sure that this method is only ever invoked from the same thread? If it should indeed only be invoked from the main thread, then adding an assertion would be a good idea.

alloy:
Yup, the new reference image looks like what we really want 👍

alloy:
This is the bug http://www.openradar.me/18034852 and it mentions the following workaround:

> If we set `currencyFormatter.internationalCurrencySymbol` as well, the `maximumFractionDigits` performs as expected.

Gonna try that now.

Artsy/Models/API_Models/Auctions/SaleArtwork.m
```diff
@@ -138,19 +124,22 @@ - (BidderPosition *)userMaxBidderPosition
 
 + (NSString *)dollarsFromCents:(NSNumber *)cents currencySymbol:(NSString *)symbol
 {
-    NSNumber *amount = @(roundf(cents.floatValue / 100));
-    if ([amount integerValue] == 0) {
-        return [symbol stringByAppendingString:@"0"];
```
alloy: So this 0 guard is really not needed?

alloy:
Yup, that works. Pushed a commit that reverts to the previous single definition of a formatter and uses the workaround. Do you have a preference for which fix we use?

Artsy/Models/API_Models/Auctions/SaleArtwork.m
alloy: Btw, even by just changing the currency symbols this isn’t really thread safe (although at least we’re not allocating/freeing memory).

ashfurrow:
No preference, thanks for following up 👍 I'd say this is good to go now.

PR #2256, opened 2017-04-05
Updates Aerodramus 📱

ashfurrow:
We want this fix: https://github.com/artsy/Aerodramus/pull/6 to hopefully resolve this issue: https://github.com/artsy/auctions/issues/262


ashfurrow:
Tests failed, what do you want to bet the failures are legit because they previously relied on the bug with date parsing?

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">There were <a href="https://eigen-ci.s3.amazonaws.com/4003/index.html">snapshot errors</a></td>
    </tr>
  </tbody>
</table>

<p align="right" data-meta="generated_by_circle">
</p>

orta:
That's possible, but it will be with fake data, so you probably won' be passing in bad data?

ashfurrow:
Yeah, I'm going to take a closer look tomorrow.

ashfurrow:
Okay, so tests were _crashing_ for me locally, something to do with ReactiveCocoa. Could be related to Xcode 8.3. I've updated to the latest pre-Swift version of the library and not only do tests complete but they also _pass_, let's see how that affects CI.

ashfurrow:
Tests passing locally, but not on CI. I've tried disabling wifi locally but they still pass. Hmm. This should be fun.

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

ashfurrow:
Okay, so I couldn't reproduce the test failures locally so I thought I'd use a stubbed subclass of ArtsyEcho/Aerodramus to never fetch new settings. However, I couldn't subclass the Swift class in my Objective-C tests. I then ran into problems importing a Swift subclass into the Objective-C tests, and the easiest solution to test the hypothesis was to rewrite the 4-line Swift file in Objective-C. So here we are :smile: Fingers crossed that this works.

Podfile
```diff
@@ -51,7 +51,7 @@ target 'Artsy' do
   pod 'Mantle', '~> 1.5.6'
   pod 'MMMarkdown'
   pod 'NPKeyboardLayoutGuide'
-  pod 'ReactiveCocoa'
+  pod 'ReactiveCocoa', '< 3'
```
orta: lols

ashfurrow: ❤️

ashfurrow:
Okay, so the test failures are all ARSwitchboard-related. Things like, returning `ARInternalMobileWebViewController` instead of the correct view controller. Like it's not loading the routes from the local file. @orta any ideas on why, in tests, the local file wouldn't be loading? Or any other reason this might be happening? I'm going to run `make update_echo` just to be safe.

ashfurrow:
I've rearranged a few things here: https://github.com/artsy/eigen/pull/2256/commits/4d1ced9c6e25b83e7f71f9743c5da186e9964a65 to see if we can get different test failures on CI.

ashfurrow:
Interesting, took a look at the raw log and found this:

```
You have to have the same named route in Echo in order to use dynamic routing
```

orta:
@ashfurrow I'll be in NYC next week, wanna pair on it then?

ashfurrow:
Might be nice, I'm going to peck away at it in the mean time. Thanks!

ashfurrow:
So the routes are empty, hmm. I've added some more debug logging to see why.

ashfurrow:
Think I figured it out, re-verifying now.

orta:
Ah yeah, because it's not using ArtsyAPI - it won't fail on non-stubbed networking.

ashfurrow:
Yup. Also the `ArtsyEchoProductionToken` environment variable was missing on CI, so echo has been broken for a bit, but we didn't notice because the payload that cam back `{"error": "unauthorized"}` didn't contain an `updated_at` field, failed to parse, and Aerodramus _didn't_ update. This PR makes it. So what was happening was, ArtsyEcho tried to update, update with bad routes with the error JSON, and that error JSON was persisted to disk. So when ARSwitchBoard tests happened, they were reading bad JSON. 

I'll push a fix, remove the debugging logging, and prevent ArtsyEcho from updating during tests.

Artsy/App/ArtsyEcho.m
```diff
@@ -0,0 +1,26 @@
...
+#import <Keys/ArtsyKeys.h>
+#import <Foundation/Foundation.h>
+#import "ARAppStatus.h"
+
+
+@implementation ArtsyEcho
+
+- (instancetype)init {
+    ArtsyKeys *keys = [ArtsyKeys new];
+    NSURL *url = [[NSURL alloc] initWithString:@"https://echo-api-production.herokuapp.com/"];
```
orta: so weird seeing a NSURL being created this way

PR #2271, opened 2017-04-11
[Onboarding] Fix issue when searching for artists to scroll to top

### LINKED ISSUE #2270: Onboarding artists to follow: always jump to top of artist list when searching
When you scroll down the list of artists and then search for another, it doesn't bump you back to the top of the list which makes it look like the artist isn't in the db.

mennenia:
Fix #2270


alloy:
Code looks good, but maybe add a CHANGELOG entry?

mennenia:
@alloy done ^_^

alloy:
Yay!

PR #2283, opened 2017-04-18
Fixes crash due to IUO

ashfurrow:
Sales with no closing date could cause issues due to Swift implicitly unwrapping a nil optional.

Related to https://github.com/artsy/emission/issues/486


orta:
👍

orta:
No need to wait on depCI

PR #2301, opened 2017-04-25
Fix push notifications not being shown outside account creation

### LINKED ISSUE #2294: No prompt to allow push notifications
I recently joined to cool club and started using a [loaner] ipad (mini, iOS 9). My first act was to install the Artsy app so I can get timely notifications.

However, when I check my user's `devices` with the Artsy API, there are none. Thinking back, I don't recall ever being prompted to allow push notifications. @alloy confirmed that he also didn't get prompted upon re-installing the app.

mennenia:
Fix #2294


alloy:
Ace, thanks! 👌

PR #2313, opened 2017-04-28
[ARUserManager] Re-enable save creds in iCloud Keychain.

alloy:
@mennenia I don’t understand the comment, re-enabling it doesn’t show me an alert or something. Was that the reason you had disabled it?

### DIFF
--- a/Artsy/Networking/API_Modules/ARUserManager.m
+++ b/Artsy/Networking/API_Modules/ARUserManager.m
@@ -157,7 +157,7 @@ static BOOL ARUserManagerDisableSharedWebCredentials = NO;
                          gotUser:gotUser
            authenticationFailure:authenticationFailure
                   networkFailure:networkFailure
-        saveSharedWebCredentials:NO]; // This has been changed to NO because in the current onboarding flow we do not request this
+        saveSharedWebCredentials:YES];
 }
 
 - (void)loginWithUsername:(NSString *)username
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -16,6 +16,7 @@ upcoming:
       - Adds support for lot_label in auctions - ash
       - Fixes a crash in opening sales with no end date - ash
       - Spruced up artworks in auctio view - ash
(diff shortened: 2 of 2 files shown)


orta:
I just installed from scratch (on master, not this PR) and I saw the alert twice (even after accepting it the first time) - perhaps this call here is not needed as it's somewhere else?

alloy:
When you sign in I don't believe there's any other way this could get triggered, maybe there is if you create a new account? What did you do?

orta:
I uninstalled the app, then hit run:

alloy:
Oh I see, this isn't about showing the saved creds, it's about saving them after entering creds.

orta:
Doh 🍡

mennenia:
@alloy it was taken out because @katarinabatina and I couldn't find a way for the pop up to appear gracefully mid onboarding. 

The problem:
- We now create an account first (before personalisation)
- As soon as you create the account in `ARUserManager`, you have the option to save creds. This results in the sudden popup between account creation and personalisation.
- We can't delay account creation, because if it fails, we'd need to display the error to the user straight away (e.g. no internet connection, faulty password, actually tied to Facebook, etc).


So the popup in the beginning that Orta refers to, is for existing people to log in easily. We kept that one. 

What I've also noticed is that it seems we save them anyway, because when I test multiple onboarding runs, they have been added to the keychain. Try creating an account, logging out, and seeing if your created account shows up.

alloy:
Hmm, yeah I see that happening when creating a new account indeed. I was only signing-in with an existing account.

I would really like this re-enabled, though, can you have a think about how else to solve this? e.g. delay saving the credentials, like we do with the push notifications alert?

mennenia:
Yeah I'll have a think. I don't want to bombard the user with too many popups either... but placing keychain outside the context of log in / signup doesn't make sense, nor does moving push notification permission elsewhere, because we want it as soon as possible.

🤔

alloy:
Bump, would really like this back, I keep having to fill out creds :)

orta:
OK, this is now back with https://github.com/artsy/eigen/pull/2709

Issue #2311, opened 2017-04-28
Update logic for "sale has ended" state


ashfurrow:
We [currently use a sale's `ended_at`](https://github.com/artsy/eigen/blob/b468ff22007ecd78c2d64614725ef9cb052fa5b4/Artsy/Models/API_Models/Auctions/SaleArtwork.m#L53) to determine if the sale has closed. But live sales typically don't have this value, so we should rely on the `auction_state` property of a sale instead, like force does.

PR #2315, opened 2017-04-29
[Auctions] Improve SaleArtworks state for auction closed

orta:
The definition is now based on gravity's sale data - fixes https://github.com/artsy/eigen/issues/2311

As all of the production examples from yesterday I could use are now "fixed" from a data perspective, I couldn't test this against a direct reproduction.


orta:
Mainly just need to get tests green, need to head out in a second - so if critical you can fix, otherwise I'll get around to it

ashfurrow:
Confirmed this works, thanks @orta 👍

PR #2327, opened 2017-05-05
WIP - Adding sale on hold alert banner

erikdstock:


Artsy/View_Controllers/ARSerifNavigationViewController.m
```diff
@@ -222,6 +225,10 @@ - (void)layoutSubviews
             object.backgroundColor = [UIColor artsyGraySemibold];
         }
     }];
+
+    if (self.alertView.superview == nil) {
+        [self addSubview: self.alertView];
+    }
 }
```
orta: I don't know if you're pairing with someone, but this probably isn't the right abstraction to really work at ( these nav controllers are used in many places, so adding auctions alerts stuff doesn't make sense for the rest) - we have ARNotificationView which should provide an API to do this kind of thing inside the view controller owning the alert

https://github.com/artsy/eigen/blob/master/Artsy/Views/Notifications/ARNotificationView.m

ashfurrow: Yup, we're working through it together. This is just a proof of concept so far, a kind of MVP banner. I'll take a look into ARNotificationView and see what it looks like, thanks 👍

ashfurrow: Looks like almost what we want, there could be some issues with a persistent notice though. I'll take a deeper look.

ashfurrow: Yeah, I'm a bit wary of using ARNotificationView for this kind of persistent UI. We could show it with a zero duration (which would show it forever) but it would queue up all other notifications behind it, which wouldn't be ideal.

Adding to the nav bar might not be the best solution (I was going for something similar to UINavigationBar's `prompt` property) so maybe adding to the sale VC might be a better approach. I'll chat with Erik about it.

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

erikdstock:
@katarinabatina - here is what we have so far. Any feedback?

simulator screen shot may 16 2017 5 18 39 pm
simulator screen shot may 16 2017 5 27 10 pm
simulator screen shot may 16 2017 5 26 13 pm

katarinabatina:
yes lots of feedback please let me submit spec'd design for this implementation

ashfurrow:
@katarinabatina any follow-up on the design for this?

katarinabatina:
I was ooo I will follow up today

katarinabatina:
There are some real estate challenges with getting this banner in. I propose slightly different implementation for ipad and iphone below 

image

for the phone, we should show this overlay and once the user taps ok we show the abbreviated banner 
image

ashfurrow:
Thanks for following up Katarina, the designs look good. One concern I have is that the header on iPad is taller than the container, whose height is fixed by the operating system. It's not impossible to customize, but would take a lot of work for a feature we don't often use. Would it be okay to have the same wording as on iPhone and use a similar click-through to get context? Happy to discuss further.

katarinabatina:
Not sure what you mean when you say "the header is taller than the container" for ipad

katarinabatina:
this container is closer to 90px tall no? 
image

ashfurrow:
Just took this screenshot from the iPad:

simulator screen shot may 23 2017 11 46 52 am

katarinabatina:
got it ok, thought that area was taller vertically. That's fine then we can pursue the same treatment 

image

PR #2331, opened 2017-05-09
Allows users to place higher bids in advance in LAI

ashfurrow:
/cc @devangt 

See https://github.com/artsy/auctions/issues/453


orta:
Cool 👍

PR #2346, opened 2017-05-16
[Emission] Update to 1.3.4

l2succes:
- Fixes loading state UI for artwork rail - luc


CHANGELOG.yml
```diff
@@ -32,6 +32,7 @@ upcoming:
       - Add popular artists fallback url for onboarding - maxim
       - Fixes display of artworks in non-expanding grids in artwork rails - ash
       - Adds inline bid info to artist grid elements - ash
+      - Fixes loading state UI for artwork rail - luc
```
alloy: This isn’t really needed because it’s not a change since Eigen 3.2.0. Not a blocker, though, can filter that out when cutting the release.

alloy:
Ok, dunno what’s happening there, but the important CIs passed 👍

PR #2350, opened 2017-05-18
[WFU] Fixes selected artist not showing when view is already presented

l2succes:

### DIFF
--- a/Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
+++ b/Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
@@ -477,6 +477,8 @@ static const CGFloat ARMenuButtonDimension = 50;
     
     // If there is an existing instance at that index, use it. Otherwise use the instance passed in as viewController.
     // If for some reason something went wrong, default to Home
+    BOOL alreadySelectedTab = self.selectedTabIndex == index;
+    
     switch (index) {
         case ARTopTabControllerIndexFeed:
         case ARTopTabControllerIndexBrowse:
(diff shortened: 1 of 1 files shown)


Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
```diff
@@ -477,6 +477,8 @@ - (void)presentRootViewController:(UIViewController *)viewController animated:(B
     
     // If there is an existing instance at that index, use it. Otherwise use the instance passed in as viewController.
     // If for some reason something went wrong, default to Home
+    BOOL alreadySelectedTab = self.selectedTabIndex == index;
```
l2succes: I moved that up so we can override the boolean based on the controller being presented

Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
```diff
@@ -485,12 +487,13 @@ - (void)presentRootViewController:(UIViewController *)viewController animated:(B
         case ARTopTabControllerIndexFavorites:
         case ARTopTabControllerIndexNotifications:
             presentableController = [[ARNavigationController alloc] initWithRootViewController:viewController];
+            alreadySelectedTab = NO;
```
l2succes: This makes sure we always present the new view controller instance

alloy: Can you add that as a comment to the code? Without it it's going to be very confusing to the next reader of the code.

l2succes: Sure

alloy:
👍

PR #2355, opened 2017-05-24
[Analytics] Fix double fire in artwork pages

### LINKED ISSUE #2265: [Analytics] Troubleshooting
## Outstanding issues 

- [ ] Double fire, specifically on artwork screens
- [x] Mobile views (e.g. editorial articles) - do we send any analytics there?

mennenia:
Fix #2265 

I understood from @katarinabatina that the only remaining issue in there was the artwork double fire, so this should complete that ticket! :)

The issue was RAC related - multiple fires of the artwork being updated, so I've just changed to the ArtworkVC's viewDidAppear event, with grabbing details from its artwork view still.

### DIFF
--- a/Artsy/App/ARAppDelegate+Analytics.m
+++ b/Artsy/App/ARAppDelegate+Analytics.m
@@ -1183,12 +1183,12 @@
                 },
                 // ========== CORE CONTENT SCREENS ==========
                 @{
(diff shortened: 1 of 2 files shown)


orta:
Yep - that makes sense from the double fire to me 👍

PR #2356, opened 2017-05-25
[Fix] Crash where facebook email could be nil 

mennenia:
Fix for [this Sentry crash report](https://sentry.io/artsynet/eigen/issues/279982705/)

We know FB emails can be nil (namely, you can have your phone number as unique ID), and I _thought_ we guarded against that, but alas, invalid logic. 

Thanks for spotting @alloy !


alloy:
Is this branch up-to-date with master?

mennenia:
It is now! :)

PR #2360, opened 2017-05-29
[Dev] Fix Sentry versioning and make dSYMs available for download from CircleCI. #trivial

alloy:


alloy:
CI failure is because of Fastlane not being able to parse the iTC response, which is probably a temporary situation.

PR #2368, opened 2017-06-12
Adopts iOS7+ status bar APIs

ashfurrow:
This touches a lot of code and needs thorough testing. So far I've got it to compile, but you've got to start somewhere! Will update this PR description when it's ready for review.


orta:
Bravo, brave move 🥇

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

ashfurrow:
I was complaining about iOS SDK adoption on twitter and realized I was being a Bad Role Model 🤷‍♂️

orta:
Heh, the problems of OSS by default - you underwear is always out on show

ashfurrow:
Okay, this is 🍏 now. @orta thoughts on merging as-is or waiting for an Emission update?

orta:
I think there's enough bug fixes in the Eigen changes to warrant a patch release (more is more) 

I think it might be a bit tricker to do an Emission release to coincide, but maybe not - as you're only change the Obj-C, then you can ship the same JS and just change those root VCs

ashfurrow:
Okay, I'm erring on the side of merging then. We can do a separate PR for the Emission pod update.

Artsy/View_Controllers/Core/UIViewController+BlurredStatusView.swift
```diff
@@ -100,8 +92,5 @@ extension UIViewController {
         guard let blurredStatusOverlayView = blurredStatusOverlayView else { return }
         blurredStatusOverlayView.removeFromSuperview()
         self.blurredStatusOverlayView = nil
-
-        UIApplication.shared.statusBarStyle = statusViewPreviousStatusBarStyle
     }
-
```
orta: This is something we may want to look into, the rm'd code was dynamically changing a view controller's status bar style base based on an addition to the view hierarchy. May need to do the same but for the preferred one

Artsy/View_Controllers/Live_Auctions/LiveAuctionViewController.swift
```diff
@@ -77,6 +73,14 @@ class LiveAuctionViewController: UIViewController {
             registrationStatusChanged = false
         }
     }
+    
+    override var preferredStatusBarStyle: UIStatusBarStyle {
+        if showingOverlay {
+            return .lightContent
```
orta: unless this is what I'm talking about above - in which case can you add a note at the top of the file that it's the responsilbility of the VC to change this

ashfurrow: Correct. I'll add a note, though this use is congruent with the new APIs.

PR #2369, opened 2017-06-12
Removes remaining deprecated UIKit API uses

ashfurrow:
~I was trawling for deprecated things to remove from Eigen and this seemed like an easy one. Next up is our use of popovers, but it won't be as easy or straightforward.~

Edit: I did popover controllers too 🙈 

- Removes UIAlertView and associated dependency
- Removes use of deprecated UIPopoverController API

### DIFF
--- a/Artsy/View_Controllers/Login_and_Onboarding/Onboarding_stages/5-_Onboarding_Personalization/ARPersonalizeViewController.m
+++ b/Artsy/View_Controllers/Login_and_Onboarding/Onboarding_stages/5-_Onboarding_Personalization/ARPersonalizeViewController.m
@@ -219,11 +219,8 @@
         default:
             break;
     }
-    
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wdeprecated-declarations"
-    [self finaliseValuesForiPadWithInterfaceOrientation:self.interfaceOrientation];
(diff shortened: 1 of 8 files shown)


Artsy/View_Controllers/Login_and_Onboarding/Onboarding_stages/5-_Onboarding_Personalization/ARPersonalizeViewController.m
```diff
@@ -219,11 +219,8 @@ - (void)showViews
         default:
             break;
     }
-    
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wdeprecated-declarations"
```
mennenia: Are these what pragma marks look like or something else?

ashfurrow: Yeah, these were being used before to ignore the deprecation warnings of the `self.interfaceOrientation` on `UIViewController`. Since moving to the status bar orientation, we don't need them. I tested to make sure the layout still works, too.

ashfurrow:
Okay, given no more discussion here I'm going to merge. Thanks again!

PR #2370, opened 2017-06-13
Updates Emission #trivial

ashfurrow:
To 1.3.9. See PR here: https://github.com/artsy/emission/pull/612


orta:
FWIW, this is probably worth a CHANGELOG - but other than that, looks 👍

ashfurrow:
I thought about a changelog entry, but felt it was already covered by the status bar API updates entry. I'll add one just in case.

orta:
Yeah, I agree that you've got the "what changed" covered, but knowing when the JS has changed is useful too (yeah - it's literally 3 characters, but it's got to change the reporting in sentry etc)

PR #2373, opened 2017-06-19
Fixes an issue with the status bar appearence returning from LAI modals.

ashfurrow:
Still not sure why this is necessary, going to look into it more. In the mean time, this works.


orta:
Yep - this looks legit to me, I can kinda see why it would need to do it this way. maybe UIWindow tries to find the top VC by looking at presentations/parents etc, and the modal still wins until completely gone.

Issue #2384, opened 2017-07-10
Status Bar issue on Refine VC


ashfurrow:
Status bar is visible in the refine VC.

![image uploaded from ios 1](https://user-images.githubusercontent.com/498212/28024807-b6cdd110-655f-11e7-9ad2-e9e3155a10f8.jpg)

It should be hidden. Reported by @katarinabatina here: https://artsy.slack.com/archives/C02BAQ5K7/p1499640459241641

PR #2386, opened 2017-07-10
WIP: Fixes status bar problem on Refine view controller

ashfurrow:
This problem was introduced [here](https://github.com/artsy/eigen/pull/2368/files#diff-49c86a13feb606822a3e4f37eae08989L42) in #2368. I accidentally switched from status bar _hidden_ to status bar _style_. But, since this VC isn't shown full screen on an iPad, we can simply always say it prefers the status bar hidden.

Investigating the iPad status bar issues led me further down a rabbit hole until I figured out what was causing our problems with status bars having to be manually changed back in `viewDidAppear`. The problem was our top menu VC was overriding `childViewControllerForStatusBarHidden` and `childViewControllerForStatusBarStyle`, returning the _presented_ view controller in all cases. iOS is smart enough to decide (based on presentation contexts and trait collections) which VC to use. So I removed that code and deferred only to our root nav controllers, which use the default iOS behaviour. Everything seems to work, and there's less code :tada: 

Fixes https://github.com/artsy/eigen/issues/2384

This is a WIP because I need to test the live VC's but our staging environment is currently under the weather. Will un-WIP once I've done that.

### DIFF
--- a/Artsy/View_Controllers/ARSerifNavigationViewController.m
+++ b/Artsy/View_Controllers/ARSerifNavigationViewController.m
@@ -89,11 +89,7 @@
 
 - (void)closeModal
 {
-    // For reasons yet unknown, when re-appearing from some (eg live auctions) view controllers, we need to call this manually.
-    id presentingViewController = self.presentingViewController;
-    [presentingViewController dismissViewControllerAnimated:YES completion:^{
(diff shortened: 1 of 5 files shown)


orta:
Cool - the concept sounds 👍 to me

Issue #2389, opened 2017-07-12
MenuAwareViewController protocol and Emission's switchboard


mennenia:
<img width="235" alt="screen shot 2017-07-12 at 12 01 03" src="https://user-images.githubusercontent.com/373860/28114931-cac584fa-66fa-11e7-8f5e-533591d3d631.png">

So in order to hide the floating back button at will, a class needs to conform to [`ARMenuAwareViewController`](https://github.com/artsy/eigen/blob/master/Artsy/View_Controllers/Util/ARMenuAwareViewController.h) OR be a root view controller (which is why there's no issue on home. Can't go any further back). 

Moving conversations into its own tab at the bottom, means subclassing `ARInboxComponentViewController` would no longer be necessary. (which has been done, temporarily, [here](https://github.com/artsy/eigen/pull/2390))

But there's still a problem when you press a message preview and go into a conversation. We have a new, chevron back button there, but the old Eigen style one still floats atop. Because it's routed from within Emission, I can't subclass the `conversation component`. Even if we were to take out the back button from that entire navigation stack, we'd still need it for the `ARArtworkViewController`, because that's still a native view.

<img width="829" alt="screen shot 2017-07-12 at 12 04 51" src="https://user-images.githubusercontent.com/373860/28114927-c4cec43a-66fa-11e7-8907-d7195bbdb079.png">

PR #2390, opened 2017-07-12
[MSG] Remove back button from inbox VC

mennenia:
Fix [Emission#666](https://github.com/artsy/emission/issues/666)

I only managed to get it out of the main inbox view, the button still hangs around on a conversation view, issue futher explained in #2389

screen shot 2017-07-12 at 12 04 51


alloy:
I don’t think the button should be completely removed, or at least not special-cased. For instance, currently we show the inbox in a nav controller from the admin menu, how would you navigate back?

Additionally, I don’t think we need to subclass to implement the menu aware protocol, iirc we do it through a class category for other component VCs.

In retrospect, maybe we should’ve waited to fix this until we got to reworking the locations of where these VCs will live so that we know for sure what is and is not needed. Sorry 😑

mennenia:
@alloy ahh so my understand was that this was supposed to be a quick fix (seeing as it's for the upcoming beta), because this implementation is definitely temporary (and hence I've created #2389). It needs a more thorough solution.

I couldn't quite get the category in `ARAppDelegate+Emission` to work myself, and because I assumed we'd delete this later, I opted for the subclass.

As for losing the behaviour of the back button, that would've been a given, considering you can only launch it from the admin list at the moment. I figured one would kill the app, shake again, or just use the other tab buttons to navigate around.

Feel free to just close this PR though if you don't think it's helpful for this beta.

alloy:
Yeah alas I think we should punt on it for now. Let’s keep the branch around for now, though.

alloy:
Aaargh, damn Safari extension beat me to it. Can you re-push your branch?

mennenia:
Branch has been restored 👍

alloy:
Aaaaaarggggggggghhhh… Can you do it one more time? It happened again, I’m not gonna click the notification for this PR again, I promise.

orta:
Lol, I just deleted it

orta:
Want to push it with a different name?

PR #2392, opened 2017-07-12
Fixes user login for 6-character passwords

ashfurrow:
The form still submits if you hit "Enter", but the "Next" button is disabled.


mennenia:
Good spot! You could've made an account when password creation wasn't as strict. 

What I would do however is be more forgiving in the login stage, seeing as any newly created passwords must still be over 7 characters. This is the primary use of the `validPassword` function, it just happens to be applied to the password field in both cases.

For example, in [`passwordTextChanged`](https://github.com/artsy/eigen/blob/master/Artsy/View_Controllers/Login_and_Onboarding/Onboarding_stages/5-_Onboarding_Personalization/ARPersonalizeViewController.m#L472), you could change the if statement that enables the button to either adhere to the validation (`validPassword`) or enable if it's in the login state anyway:

```
- (void)passwordTextChanged
{
    NSString *password = self.onboardingTextFields.passwordField.text;
    
    if ([self validPassword:password] || self.state == AROnboardingStagePersonalizeLogin) {
        [self.onboardingTextFields disableErrorState];
        [self.headerView disableErrorHelpText];
        [self.onboardingNavigationItems hideError];
        [self.onboardingNavigationItems enableNextStep];
    } else {
        [self.onboardingNavigationItems disableNextStep];
    }
}
```

What do you think?

ashfurrow:
That makes sense to me – is there a push to require passwords to be at least 7 chars long? We should change the Kiosk if that's the case.

mennenia:
It was a [change]((https://github.com/artsy/eigen/blob/master/Artsy/View_Controllers/Login_and_Onboarding/Onboarding_stages/5-_Onboarding_Personalization/ARPersonalizeViewController.m#L168)) introduced in the last update to Eigen onboarding. It's not necessarily an API requirement, so I'm not sure if it's necessarily rolled out to all apps. @katarinabatina any thoughts? Should Kiosk also update?

ashfurrow:
Awesome, thanks for following up @mennenia – I can amend this PR.

ashfurrow:
Cool, I've made those changes. I put them in the `validPassword:` function, let me know what you think.

mennenia:
Awesome! 🎉

PR #2398, opened 2017-07-16
Fix OSS builds

### LINKED ISSUE #2397: Sentry key build/run issue
Hi esteemed Artsy crew! I'm an iOS engineer wanting to get familiar with your codebase and how things work as part of our research into React Native. 

Following the build instructions for Eigen, I get stuck on the `bundle exec pod install` step, because `cocoapods-keys` prompts me to enter the keys for `SentryProductionDSN` and `SentryStagingDSN`. I don't know what I should be entering here.  

If I enter a random string, each as abcdefgh, then the `pod install` completes successfully, but at startup the code throws an exception in `[ARSentryAnalyticsProvider initWithDSN:]` with the message *'Unable to initialize a SentryClient SDK: Error Domain=SentryErrorDomain Code=100 "Project ID path component of DSN is missing"*

Am I missing something?

orta:
Fixes #2397 

* Needed to add primed keys for the Sentry stuff in the makefile.
* OSS builds don't need sentry, so we can skip it at runtime.


Artsy/App/ARSentryAnalyticsProvider.m
```diff
@@ -11,7 +11,10 @@ - (id)initWithDSN:(NSString *)DSN
 {
     self = [super init];
     if (!self) { return nil; }
-    
+
+    // For OSS builds
+    if ([DSN isEqualToString:@"-"]) { return self; }
```
alloy: I’m sightly worried that this may lead to somebody accidentally making a build with `-` set and is submitting that to the store. Is there another more reliable to differentiate an OSS build from others?

orta: We have checks for this in the Fastfile 👍 - https://docs.fastlane.tools/actions/#verify_pod_keys

Our build will fail if they are 1 char long

orta:
Updates the reference for ARMasonry -  https://github.com/ashfurrow/ARCollectionViewMasonryLayout/compare/a9dc52b13e32d064c49528ff9a3bd5aa65c02cd5...f22180d0267567fd3db62f903eaef1119b5c1f2d

While the version number is bumped, the only real changes are documentation ^

Artsy/App/ARSentryAnalyticsProvider.m
orta: We obviously have done this in the past - and so we built it into Fastlane ;)

alloy: Hahaha. Cool 👍

PR #2407, opened 2017-07-28
[Dev] Adds psych to the gemfile

orta:
No more changing the `Podfile.lock` WRT strings 🥇 - thanks @keith

Also re-orders the gems in Gemfile, removes commented out ones.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,19 +1,17 @@
 source 'https://rubygems.org'
 
-gem 'cocoapods', '>= 1.1.0'
-gem 'cocoapods-keys'
-gem 'cocoapods-check'
+gem 'cocoapods', '>= 1.1.0' # Deps
+gem 'cocoapods-check' # So we know if we need to run `pod install`
+gem 'cocoapods-keys' # So we don't leak ENV vars
+gem 'psych' # So our Podfile.lock is consistent
 
 group :development do
-  gem 'lowdown'
-  # Depends on older xcodeproj gem that is incompatible with current CocoaPods.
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -1,19 +1,17 @@
 source 'https://rubygems.org'
 
 gem 'cocoapods', '>= 1.1.0'
-gem 'cocoapods-keys'
 gem 'cocoapods-check'
+gem 'cocoapods-keys'
+gem 'psych'
```
ashfurrow: Could we add a comment to this explaining its purpose?

Podfile.lock
```diff
@@ -385,7 +385,7 @@ SPEC CHECKSUMS:
   AFOAuth1Client: 07ccc935ba06ac8d023c16d39a5bdf04bc6f92e1
   ALPValidator: c74ea0d49bbbff0f8a4228f1eb6589b992255cfe
   Analytics: b342fb4f43fa4f97ca6f4358b44d3295217ba294
-  AppHub: 03788112dd48c42b60d51321c0ffff4ef15a4432
+  AppHub: '03788112dd48c42b60d51321c0ffff4ef15a4432'
```
ashfurrow: I thought Psyche was supposed to remove these?

orta: No - it standardises it - so the standard is that it is in

Gemfile
orta: added comments to _all_ gems

keith:
❤️

PR #2408, opened 2017-08-03
[Auctions] Adds haptic feedback for any live auctions

orta:
This will only work on iPhone 7s, and it is behind and admin flag.

Our tests only run on an iOS 9 device, so I can't do a 'does this call happen' kind of test, but even then I can't verify it's called without something like OCMock. So, no tests.

### DIFF
--- a/Artsy/View_Controllers/Live_Auctions/LiveAuctionBidHistoryViewController.swift
+++ b/Artsy/View_Controllers/Live_Auctions/LiveAuctionBidHistoryViewController.swift
@@ -1,7 +1,6 @@
 import UIKit
 import Interstellar
 
-
 class LiveAuctionHistoryCell: UITableViewCell {
 
     override init(style: UITableViewCellStyle, reuseIdentifier: String?) {
@@ -22,10 +21,10 @@ class LiveAuctionHistoryCell: UITableViewCell {
     }
 }
 
+
 class LiveAuctionBidHistoryViewController: UITableViewController {
 
     let lotViewModel: LiveAuctionLotViewModelType
-
     var newEventsSubscription: ObserverToken<[LiveAuctionEventViewModel]>!
 
     init(lotViewModel: LiveAuctionLotViewModelType) {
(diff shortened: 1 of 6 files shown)


Artsy/View_Controllers/Live_Auctions/LiveAuctionBidHistoryViewController.swift
```diff
@@ -95,6 +100,28 @@ class LiveAuctionBidHistoryViewController: UITableViewController {
...
+                }
+                if event.isFairWarning {
+                    feedbackWeight = .medium
+                }
+                if event.isFinalCall {
+                    feedbackWeight = .heavy
+                }
+            }
+
+            if(feedbackWeight != nil) {
```
ashfurrow: This could be in a `guard let` statement, to avoid the force-unwrap on the next time.

Artsy/View_Controllers/Live_Auctions/LiveAuctionBidHistoryViewController.swift
```diff
@@ -95,6 +100,28 @@ class LiveAuctionBidHistoryViewController: UITableViewController {
         cell.updateWithEventViewModel(event)
         cell.drawBottomDottedBorder(with: UIColor.artsyGrayMedium())
     }
+
+    func provideHapticFeedbackForEvents(events: [LiveAuctionEventViewModel]) {
+        if #available(iOS 10.0, *) {
+            var feedbackWeight: UIImpactFeedbackStyle? = nil
+            events.forEach { event in
```
ashfurrow: There's a subtle bug here but I don't think it's worth fixing – wanted to mention it, might be worth a comment. So if several events come in at once, the last one gets to decide the feedback weight. So if a fair warning came in first, and a bid second (in the array), you'd only get the `.light` weight. I don't think it's likely (possible?) that causality would send over those two events in a single web socket message though, so this code should work as-is 👍

orta: Yeah, I think this is worth the time to get right 👍

orta: I've made it only assign the lowest feedback if there's no feedback currently, so at worst, you could get a medium instead of a heavy. I'm alright with that.

orta:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true">Messages</th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
      <td>

  Just verifying this continues to work.
  </td>
    </tr>
  </tbody>
</table>


<p align="right">
</p>

orta:
:D - err, just me doing some work - hey @devangt - this just got shipped to the App Store BTW, test it by shaking and turning it on

ashfurrow:
This wasn't actually included in the 3.2.4 version shipped to the store, but will be in the 3.3.0 beta coming out today.

Issue #2113, opened 2017-01-17
Don’t intercept `tel:` links


alloy:
From https://github.com/artsy/collector-experience/issues/74

Currently we end up showing this ugly alert when a user taps a `tel:` link in a webview:

![](https://cloud.githubusercontent.com/assets/23346526/21871412/ec54c23e-d82f-11e6-88e5-91e0fc7c1729.PNG)

izakp:
@alloy what is the new behavior to be?  I don't imagine we want to get rid of handling the link all together? Also I don't believe we can bypass the alert modal, as stated in the documentation here https://developer.apple.com/library/content/featuredarticles/iPhoneURLScheme_Reference/PhoneLinks/PhoneLinks.html

ArtsyPrivateDanger:
We shouldnt show our alert, just let it be the system one you linked to IMO

alloy:
Yes, like @ArtsyPrivateDanger says, we shouldn’t be showing our alert, but rather let iOS show the default system one that takes you to Phone.app.

PR #2412, opened 2017-08-18
Fea dont intercept tel urls

izakp:
My first pass at Eigen.  Closes https://github.com/artsy/eigen/issues/2113

### DIFF
--- a/Artsy/Networking/ARRouter.m
+++ b/Artsy/Networking/ARRouter.m
@@ -183,6 +183,12 @@ static NSString *hostFromString(NSString *string)
     return (!url.scheme || ([url.scheme isEqual:@"http"] || [url.scheme isEqual:@"https"]));
 }
 
++ (BOOL)isTelURL:(NSURL *)url
+{
+
+    return (!url.scheme || [url.scheme isEqual:@"tel"]);
+}
+
 /// TODO: don't allow Special Apple URLs? iTMS / maps.apple?
 
 + (BOOL)isInternalURL:(NSURL *)url
--- a/Artsy/Networking/ARRouter.m
+++ b/Artsy/Networking/ARRouter.m
@@ -185,8 +185,7 @@ static NSString *hostFromString(NSString *string)
 
 + (BOOL)isTelURL:(NSURL *)url
 {
-
-    return (!url.scheme || [url.scheme isEqual:@"tel"]);
+    return (url.scheme && [url.scheme isEqual:@"tel"]);
 }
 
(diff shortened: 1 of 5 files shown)


Artsy_Tests/App_Tests/ARSwitchBoardTests.m
```diff
@@ -119,11 +120,13 @@ @interface ARTopMenuNavigationDataSource (Tests)
...
-                [[switchboardMock expect] openURLInExternalService:OCMOCK_ANY];
+                id sharedAppMock = [OCMockObject partialMockForObject:[UIApplication sharedApplication]];
+                [[sharedAppMock expect] openURL:OCMOCK_ANY];
 
-                [switchboard loadURL:internalURL];
-                [switchboardMock verify];
+                NSURL *telephoneURL = [[NSURL alloc] initWithString:@"tel:111111"];
+
+                [switchboard loadURL:telephoneURL];
+                [sharedAppMock verify];
```
orta: I think it's worth making a new test for this as `tel` should be the only one with different behaviour

alloy:
Sweet! Thanks for plowing through :clap:

Artsy/Networking/ARRouter.m
```diff
@@ -183,6 +183,12 @@ + (BOOL)isWebURL:(NSURL *)url
     return (!url.scheme || ([url.scheme isEqual:@"http"] || [url.scheme isEqual:@"https"]));
 }
 
++ (BOOL)isTelURL:(NSURL *)url
+{
+
+    return (!url.scheme || [url.scheme isEqual:@"tel"]);
```
alloy: Should this really return `YES` if the URL has no `scheme`? I think it’s fair to assume a URL is a http URL in case the scheme is missing (above) but in the case of a telephone number I think it’s fair to require the scheme.

Artsy_Tests/App_Tests/ARSwitchBoardTests.m
alloy: I concur :+1:

izakp: Broke out into its own test

Artsy/Networking/ARRouter.m
izakp: Now returns NO if `url.scheme` is falsy

alloy:
Awesome, great job @izakp 👏 👏 👏

PR #2418, opened 2017-09-08
Updates bid count UI after model has been updated

ashfurrow:
Fixes https://github.com/artsy/auctions/issues/606 . What was happening was we were using the new events signal to update the # of bids UI, but the number of bids on the model wasn't being updated until _after_ the events were added (so we would typically number of bids from 1 event ago). This PR adds an observable for the number of bids specifically.


sweir27:
Aha, cool!

PR #2419, opened 2017-09-11
Fixes crash with new users without any favourites

ashfurrow:
This fixes [a crash](https://sentry.io/artsynet/eigen/issues/347208068) caused by a misunderstanding in how GraphQL errors are reported (they return a 200 status code, Eigen was relying on a non-200 to indicate failure). This needs to be applied to everywhere in Eigen we're calling metaphysics, and I will follow up with a PR to master that refactors that. In the mean time, this patch should be deployed as soon as possible.

### DIFF
--- a/Artsy Stickers/Info.plist
+++ b/Artsy Stickers/Info.plist
@@ -17,7 +17,7 @@
 	<key>CFBundlePackageType</key>
 	<string>XPC!</string>
 	<key>CFBundleShortVersionString</key>
-	<string>3.2.4</string>
+	<string>3.2.5</string>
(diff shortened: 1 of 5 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">You need to set the App’s plist version to 3.2.5</td>
    </tr>
  </tbody>
</table>

</p>

PR #2425, opened 2017-09-19
[Merge For Next Beta][Analytics/Emission] Update analytics hook for incoming Emission events

mennenia:
Don't merge yet (until 3.2.5 is safely in the store + no hotfixes in the pipeline)

🚨 DO merge if we're putting out a new messaging beta 🚨 

This work goes hand in hand with an update in Emission, so for the next Emission release this should be in 👌. 

Assignment is on an either/or basis, so that you're aware of this PR if you happen to be the one creating the next messaging beta 😄


mennenia:
Also this is pretty much guaranteed going to fail, as it's pointing to an Emission release that's out of sync. Will point to the correct one once it's published.

mennenia:
FYI this is outdated now, pushing a fix shortly

Artsy/App/ARAppDelegate+Emission.m
```diff
@@ -201,13 +201,21 @@ - (void)setupSharedEmissionWithPackagerURL:(NSURL *)packagerURL;
...
+        } else {
+            // Screen event
+            [properties removeObjectForKey:@"context_screen"];
+            [ARAnalytics pageView:info[@"context_screen"]  withProperties:[properties copy]];
+        }
+
         
         dispatch_async(dispatch_get_main_queue(), ^{
-            if ([info[@"name"] isEqual:@"Follow artist"] && [fromViewController isKindOfClass:[ARArtistComponentViewController class]]) {
+            if ([info[@"action_name"] isEqual:@"artistFollow"] && [info[@"action_type"] isEqual:@"success"]  && [info[@"context_screen"] isEqualToString:@"Artist"]) {
```
mennenia: This is a point that could break so easily 😞 

After this is out, I'll create a way of triggering prompt for push notifications from Emission

orta:
I updated the PR - should be good to go on green

PR #2426, opened 2017-09-19
Adds -u to `make pr` command #trivial

ashfurrow:
So I've had the following scenario happen a few times:

- Do work and `make pr` at work.
- Pull that branch at home, commit+push more work.
- Next day, try to pull that work.

Git doesn't know where to pull from because we didn't set our local branch to track the remote one when we initially pushed in `make pr`. I can't think of any downsides or inadvertent side-effects.


orta:
works for me, yeah!

PR #2429, opened 2017-09-25
Sends user agent info to causality.

ashfurrow:
This PR factors out the logic for generating Eigen's user agent to a class function on `ARRouter`, and then sends that user agent to Causality in the `clientMetadata` field. I think I've got the structure of the JSON right, but I haven't tested on staging yet. 

Work for: https://github.com/artsy/auctions/issues/629


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

ashfurrow:
Just tested this by placing a bid in Eigen and then using Postman to retrieve lot events:

screen shot 2017-09-25 at 10 46 47 pm

Anyone else working on the prediction side of this? Happy to take that on.

Podfile.lock
```diff
@@ -496,6 +496,6 @@ SPEC CHECKSUMS:
   XCTest+OHHTTPStubSuiteCleanUp: 4469ec8863c6bc022c5089a9b94233eb3416c5ee
   Yoga: 77f4e44e919b274504599a3536cdc90140fb58d8
 
-PODFILE CHECKSUM: f4423138bd4e02c345bfa003ac932fcbe469195d
+PODFILE CHECKSUM: 3d6786857a9d4e425e9e5b578658a8f99a389f5f
```
orta: This is worrying, can you rm `-rf Pods` and re-install - http://artsy.github.io/blog/2016/05/03/podspec-checksums/

ashfurrow: Hmm, no change. I'll try on my Mac at work.

orta: It's possible that this is a change that wasn't shipped previously ( it's the MD5 for the `Podfile` text contents) so realistically it's not a strong worry, but we shouldn't be seeing this without the `Podfile` also inside the PR diff

ashfurrow: You're right – I re-ran `bundle exec pod install` at work and I got the same checksum. Must've been an uncommitted change previously. I'll keep an eye out for Podfile.lock churn on the `3.2.5-release` branch.

Artsy/Networking/ARRouter.m
```diff
@@ -227,6 +214,24 @@ + (NSURLRequest *)requestForURL:(NSURL *)url
     return request;
 }
 
++ (NSString *)userAgent
```
acjay: Ah, so some setup code needs to access this value and then we also want `LiveAuctionSocketCommunicator` to be able to later?

Artsy/Networking/ARRouter.m
```diff
@@ -227,6 +214,24 @@ + (NSURLRequest *)requestForURL:(NSURL *)url
...
+    // * version information
+    // * include big browser engine names so that scripts like that of fast.fonts.net WKWebView will load
+
+    AFHTTPRequestSerializer *serializer = [[AFHTTPRequestSerializer alloc] init];
+    NSString *userAgent = serializer.HTTPRequestHeaders[@"User-Agent"];
+    NSString *agentString = [NSString stringWithFormat:@"Mozilla/5.0 Artsy-Mobile/%@ Eigen/%@", version, build];
+    userAgent = [userAgent stringByReplacingOccurrencesOfString:@"Artsy" withString:agentString];
+    userAgent = [userAgent stringByAppendingString:@" AppleWebKit/601.1.46 (KHTML, like Gecko)"];
+    return userAgent;
+}
```
alloy: I assume the auction code doesn’t need to call this often, but, nonetheless, how about caching the user agent into the user defaults _here_ and return that if it already exists, seeing as that’s being done already anyways?

Artsy/Networking/ARRouter.m
ashfurrow: Yeah, it's factoring existing logic into a new method and exposing it to the rest of the app.

Artsy/Networking/ARRouter.m
ashfurrow: Okay this is done, used `static` because I like to be close to the metal 🤘

ashfurrow:
Does 3fea733 look 👍 to you @alloy?

Artsy/Networking/ARRouter.m
alloy: As the value shouldn't ever change, I’m ok with not locking access. Even in the case of a race condition this would only lead to 1 extra computation. If you do feel strongly about it then you can add a `dispatch_once`, otherwise feel free to merge 👍

PR #2430, opened 2017-10-02
Fixes Deploys

ashfurrow:
The actual fix is 82f9f8d – I think it has to do with an implementation detail of Fastlane that changed, but I can't be sure. Took a look but can't determine the cause. However, this has fixed [the most recent beta build](https://circleci.com/gh/artsy/eigen/4492). The `DEVELOPMENT_TEAM` addition is scoped to Store builds, so this should not affect day-to-day development. 

The other commit was to remove `update_project_team` since `automatic_code_signing` now accepts a `team_id` param, so we can just use the one action.


Artsy.xcodeproj/project.pbxproj
```diff
@@ -6378,6 +6378,7 @@
 				);
 				GCC_TREAT_INCOMPATIBLE_POINTER_TYPE_WARNINGS_AS_ERRORS = YES;
 				GCC_TREAT_WARNINGS_AS_ERRORS = NO;
+				DEVELOPMENT_TEAM = 23KMWZ572J;
```
orta: We should add a danger rule that this string always exists inside this file

ashfurrow: This setting is repeated seven times throughout the project file, so it won't be as easy as checking for this string. Unless you just wanna check that it exists seven times :)

orta: Ergh

alloy: Is it repeated because it’s being explicitly set on multiple levels? If so, we should just set it at the highest level and have all other levels inherit?

ashfurrow: I moved it up and verified `gym` still works, but it's still repeated :\ I'm not _that_ familiar with the internals of an Xcode project file though.

orta: Off-hand without reading the xcodeproj it should probably be in there maybe twice, once for each app target ( Eigen and the Stickers App)

alloy:
Ace 👍

PR #2434, opened 2017-10-17
Properly present modal emission views #trivial

alloy:
Eigen’s side of https://github.com/artsy/emission/pull/790

### DIFF
--- a/Artsy/App/ARSwitchBoard.m
+++ b/Artsy/App/ARSwitchBoard.m
@@ -29,6 +29,7 @@
 #import "ARMutableLinkViewController.h"
 #import "ARTopMenuNavigationDataSource.h"
 #import "ARPaymentRequestWebViewController.h"
+#import "ARSerifNavigationViewController.h"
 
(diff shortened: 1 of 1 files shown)


Artsy/App/ARSwitchBoard.m
```diff
@@ -413,7 +414,8 @@ - (UIViewController *)routeInternalURL:(NSURL *)url fair:(Fair *)fair
     }
 
     if ([ARRouter isPaymentRequestURL:url]) {
-        return [[ARPaymentRequestWebViewController alloc] initWithURL:url];
+        UIViewController *paymentRequestViewController = [[ARPaymentRequestWebViewController alloc] initWithURL:url];
+        return [[ARSerifNavigationViewController alloc] initWithRootViewController:paymentRequestViewController];
```
alloy: A view controller that gets shown modally needs to be contained inside a navigation controller so that it gets a ‘navigation bar’, which in this case just means it will have a ‘dismiss’ button.

PR #2436, opened 2017-10-17
Fixes a problem with number of bids not updating 📱

ashfurrow:
While testing the beta to submit 3.2.6 to the App Store, I opened the Phillips sale we had on and noticed that the label for the number of bids in the LAI interface never updated, even as new bids came in.

The bug was that if we had a previous subscription to the new events signal, we would never replace it until we were removed from our superview. That's fine as long as we are removed from the superview before being set up again. But if you had your view set up twice, for whatever reason, the subscription would still be the first one pointing to the first label (which is now an orphaned view). The solution is to unsubscribe existing subscriptions and always create a new subscription.

I couldn't reproduce the build in the simulator, only on device. I suspect that something about the view setup process on devices differs from the simulator.

Probably related to https://github.com/artsy/auctions/issues/606

### DIFF
--- a/Artsy/View_Controllers/Live_Auctions/Views/LiveAuctionToolbarView.swift
+++ b/Artsy/View_Controllers/Live_Auctions/Views/LiveAuctionToolbarView.swift
@@ -43,6 +43,7 @@ class LiveAuctionToolbarView: UIView {
         if newSuperview == nil {
             lotStateObserver?.unsubscribe()
             numberOfBidsObserver?.unsubscribe()
+            numberOfBidsObserver = nil
         }
     }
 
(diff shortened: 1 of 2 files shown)


ashfurrow:
Yeah I agree that would be better. If we were using a more fully-featured library like RxSwift, automated teardown is supported directly by the framework. We purposefully chose the library we did so that anyone would be able to understand the abstraction (in all it's ~150 lines of code). I fear that if we were to build our own abstractions on top of the existing library, we would defeating the purpose of using a simple library in the first place.

FWIW, in RxSwift this could look something like:

```swift
observable
  .takeUntil(...) // takeUntil terminates the observable when triggered (for example, by moving to a new superview)
  .map { ... }
  .subscribe(onNext: ...)
  .disposedBy(self.rx.disposeBag)
```

ashfurrow:
Mmm, so CI did pass in this run: https://circleci.com/gh/artsy/eigen/4517 but Circle/GitHub are confused because I sent the same commit ( ff4ba96 ) to be deployed and the deploy (while successful) produced a non-zero exit code due to an intermittent problem with our CI provider. Since the build did pass, I'm going to merge.

PR #2443, opened 2017-10-26
Adds note about checking for blank screen to deploy instructions.

ashfurrow:
This bug has shipped twice, we should check for it before releasing.


orta:
Yeah, sure, seems legit 👍

PR #2462, opened 2017-11-15
[Onboarding] Fix bug where next button was stuck behind keyboard

mennenia:
Aaaand we're back in business!

Fix [#CE695](https://github.com/artsy/collector-experience/issues/695)

simulator screen shot - iphone 7 - 2017-11-15 at 17 54 48

### DIFF
--- a/Artsy/View_Controllers/Login_and_Onboarding/AROnboardingViewController.m
+++ b/Artsy/View_Controllers/Login_and_Onboarding/AROnboardingViewController.m
@@ -142,7 +142,7 @@
 
 - (void)keyboardWillShow:(NSNotification *)notification
 {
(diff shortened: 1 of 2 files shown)


Artsy/View_Controllers/Login_and_Onboarding/AROnboardingViewController.m
```diff
@@ -142,7 +142,7 @@ - (BOOL)prefersStatusBarHidden
 
 - (void)keyboardWillShow:(NSNotification *)notification
 {
-    _keyboardFrame = [[[notification userInfo] objectForKey:UIKeyboardFrameBeginUserInfoKey] CGRectValue];
```
mennenia: This no longer gives back the correct height in iOS 11. I checked, the new methodology is still compatible with previous iOS versions (I tested 9.0).

PR #2466, opened 2017-11-20
[Dev] Fix small bug where you can't turn off analytics popovers

mennenia:
Currently you need to be showing red dots for martsy views in order to turn off analytics popovers 😄


orta:
👍

PR #2467, opened 2017-11-21
Removes .eigen file docs #trivial

ashfurrow:
Specifically, the docs around login using the `.eigen` file, as it's no longer used and not planned for un-obsolescence.


mennenia:
🎉

PR #2468, opened 2017-11-21
Modifies types to handle missing image URLs.

ashfurrow:
Almost entirely type changes, didn't feel the need to test. Let me know what you think. 

Addresses https://github.com/artsy/auctions/issues/699 .

### DIFF
--- a/Artsy/Models/API_Models/LiveAuctionLot.h
+++ b/Artsy/Models/API_Models/LiveAuctionLot.h
@@ -18,8 +18,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface LiveAuctionLot : MTLModel <MTLJSONSerializing>
 
-- (NSURL *)urlForThumbnail;
-- (NSURL *)urlForProfile;
+- (NSURL * _Nullable)urlForThumbnail;
+- (NSURL * _Nullable)urlForProfile;
 - (CGFloat)imageAspectRatio;
 
 @property (nonatomic, copy, readonly) NSString *artworkTitle;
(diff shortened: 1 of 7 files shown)


Artsy/Models/API_Models/LiveAuctionLot.h
```diff
@@ -18,8 +18,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface LiveAuctionLot : MTLModel <MTLJSONSerializing>
 
-- (NSURL *)urlForThumbnail;
-- (NSURL *)urlForProfile;
+- (NSURL * _Nullable)urlForThumbnail;
+- (NSURL * _Nullable)urlForProfile;
```
erikdstock: the combination of the `+` and the `-` for the interface... odd. this just says the property may be null?

orta: Yeah, the lack of annotation in Obj-C means that in Swift it's assumed to be non-null. this lets Swift know it can sometimes not be there.

ashfurrow: This specific file opts into the "assume things aren't null" thing with `NS_ASSUME_NONNULL_BEGIN` on line 16. The missing `_Nullable` specifier here was an oversight.

ashfurrow:
It's a good question – if the URLs were null, things get weird. [This line](https://github.com/artsy/eigen/pull/2468/files#diff-8bc0e878dab735795d10bb6f37d91f01R166) returns a "non-null" URL from a function that _can_ return `nil`. When Swift tries to unwrap that `nil` value to return a non-null value, then the app crashes on – I guess – philosophical grounds :)

PR #2469, opened 2017-11-22
Moves to use the new iOS 10 open url syntax

### LINKED ISSUE #2460: First click on iOS Universal Link goes to /home instead of destination
Steps to reproduce:
1. Close the Artsy iOS app
2. Click on a link that should get redirected to the App
3. Opens app, but goes to home
4. Click on link again and see that it goes to the destination url

orta:
Apple announced an API switcheroo with iOS 10, then forced it on iOS 11 - should fix #2460


mennenia:
🎉 awesome

PR #2471, opened 2017-11-28
[WIP] works for you email routing for selected artist

sarahscott:
This depends on https://github.com/artsy/emission/pull/845/; will require an Emission version bump.

Fixes https://github.com/artsy/collector-experience/issues/723

### DIFF
--- a/Artsy/View_Controllers/App_Navigation/ARTopMenuNavigationDataSource.m
+++ b/Artsy/View_Controllers/App_Navigation/ARTopMenuNavigationDataSource.m
@@ -49,7 +49,7 @@
         _badgeCounts[i] = 0;
     }
 
-    ARHomeComponentViewController *homeVC = [[ARHomeComponentViewController alloc] initWithEmission:nil];
+    ARHomeComponentViewController *homeVC = [[ARHomeComponentViewController alloc] initWithSelectedArtist:nil emission:nil];
     _feedNavigationController = [[ARNavigationController alloc] initWithRootViewController:homeVC];
 
     return self;
@@ -66,6 +66,12 @@
     return _messagingNavigationController;
 }
 
+- (ARNavigationController *)getHomeViewControllerWithArtist:(NSString *)artistID
+{
+    ARHomeComponentViewController *homeVC = [[ARHomeComponentViewController alloc] initWithSelectedArtist:artistID emission:nil];
+    _feedNavigationController = [[ARNavigationController alloc] initWithRootViewController:homeVC];
+    return _feedNavigationController;
+}
 
 - (ARNavigationController *)getProfileNavigationController
 {
@@ -104,7 +110,11 @@
 {
     switch (index) {
         case ARTopTabControllerIndexHome:
-            return [self feedNavigationController];
+            if (params && params[@"artist_id"]) {
+                return [self getHomeViewControllerWithArtist:params[@"artist_id"]];
+            } else {
+                return [self feedNavigationController];
+            }
 
         case ARTopTabControllerIndexMessaging:
             return [self getMessagingNavigationController];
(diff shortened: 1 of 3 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">PR is classed as Work in Progress</td>
    </tr>
  </tbody>
</table>

</p>

Artsy/View_Controllers/App_Navigation/ARTopMenuNavigationDataSource.m
```diff
@@ -66,6 +66,12 @@ - (ARNavigationController *)getMessagingNavigationController
     return _messagingNavigationController;
 }
 
+- (ARNavigationController *)getHomeViewControllerWithArtist:(NSString *)artistID
+{
+    ARHomeComponentViewController *homeVC = [[ARHomeComponentViewController alloc] initWithSelectedArtist:artistID emission:nil];
+    _feedNavigationController = [[ARNavigationController alloc] initWithRootViewController:homeVC];
+    return _feedNavigationController;
```
alloy: Can you access this ivar through a property setter/getter instead?

sarahscott: Do you mean there would be two calls in `navigationControllerAtIndex`: one to instantiate a new `ARHomeComponentViewController` with the selected artist and set the property and a subsequent one to get the new `feedNavigationController`?

alloy: No sorry, I meant that as a general rule we don’t want to directly access instance variables such as `_feedNavigationController`, but instead use a property definition to generate `self.feedNavigationController = …` and `return self.feedNavigationController` accessor methods.

sarahscott: Ok, so should this also be changed to reflect that pattern? https://github.com/artsy/eigen/pull/2471/files#diff-244b3303c916ff096156dcf2356615b6R65

I'm a bit confused by what you mean exactly w/r/t to adding a selected artist (something which only could happen from `navigationControllerAtIndex` and would not be known when the property is first instantiated.

alloy: Ideally it should yeah.

Everything will remain identical to what you have now, except you set and get the feed nav controller through `self.feedNavigationController` rather than `_feedNavigationController`.

Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
```diff
@@ -244,6 +244,10 @@ - (void)registerWithSwitchBoard:(ARSwitchBoard *)switchboard
             return [self rootNavigationControllerAtIndex:tabIndex.integerValue parameters:parameters].rootViewController;
         }];
     }
+    
+    [switchboard registerPathCallbackAtPath:@"/works-for-you" callback:^id _Nullable(NSDictionary * _Nullable parameters) {
+        return [self rootNavigationControllerAtIndex:0 parameters:parameters].rootViewController;
```
orta: I think we have a constant for that index somewhere

alloy: https://github.com/artsy/eigen/blob/f7d01baabba38e1543745e86ef8a80473c4d5035/Artsy/View_Controllers/App_Navigation/ARTopMenuNavigationDataSource.h#L6

Artsy/View_Controllers/App_Navigation/ARTopMenuNavigationDataSource.m
alloy: This file needs a bunch more cleaning. Gonna merge as-is for now.

PR #2474, opened 2017-12-01
fix Deploying a beta doc link

ldehai:


orta:
Thanks! I'm alright with the Perils fails here 👍

PR #2484, opened 2017-12-11
[Dev] Fixes a crash bug when you exit the app #trivial

orta:
Fixes https://github.com/artsy/collector-experience/issues/793

Full explanation is in ^ - but the lowdown is that events we were sending to Segment, would actively cause a hard deadlock in our app when going coming back from the home screen. Fixed with an update to the Segment lib, which adds protection against the type of data.


sarahscott:
👏

PR #2485, opened 2017-12-12
Update the API routes used for setting the user's min/max price ranges

orta:
Fixes https://github.com/artsy/collector-experience/issues/780

`price_range` isn't supported on the route we use for the min/max - this does now cause two API requests, but I think that's OK.

### DIFF
--- a/Artsy/Models/API_Models/Social/User.m
+++ b/Artsy/Models/API_Models/Social/User.m
@@ -90,11 +90,12 @@
 
 - (void)setRemoteUpdatePriceRange:(NSInteger)maximumRange success:(void (^)(User *user))success failure:(void (^)(NSError *error))failure
 {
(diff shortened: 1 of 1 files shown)


mennenia:
Not sure what's up with the dependency CI, but from a code review, this is 🎉 yay!

Bitcoin for everyone 💰

orta:
Yeah, that's fine 
I also validated the data on staging:

```
gravity:staging> User.last
=> #<User _id: BSON::ObjectId('5a3005e18b3b813d7060a134'), price_range_max: 50000, price_range_min: -1, ...>
```

mennenia:
This would've been good to add to changelog btw @orta

orta:
Agree - just realized it's a bug in production code. Will add.

PR #2489, opened 2017-12-13
Adds some more logs for the Gene Refine settings

orta:
The refine settings from JSON is super safe about the data it gets, but it fails silently, now it fails more loudly.


orta:
Made it easier to figure out https://github.com/artsy/emission/pull/891

PR #2492, opened 2017-12-14
Moves cancel button and title label down 20px to accomodate status bar.

ashfurrow:
Tracked in https://github.com/artsy/auctions/issues/777 . This PR moves the cancel button and title label of the refine view controller down each twenty pixels. 

Note that this PR contributes to the technical debt we will need to pay off when adding support for the iPhone X (and moving away from "top edges" and towards "top layout anchors").


ashfurrow:
Yeah it's a Dependency-CI issue related to caching a commit that I've since force-pushed over. It happens occasionally; since this PR doesn't change any dependencies, I'm comfortable merging as-is.

PR #2493, opened 2017-12-14
Adds supports for point estimates

ashfurrow:
Okay so this PR adds support for point estimates (in lieu of estimate _ranges_). There are two important components:

- a generic lot model that adds support for the estimate field for display in the UI, and
- a live-specific lot model that needs the estimate field for capping the max bid to be placed by the user (display for estimates in live sales is already computed for us by metaphysics)

The second point is kind of interesting since we want to move away from Eigen calculating appropriate bid values altogether and only rely on Causality/Metaphysics. But in the mean time, I've replaced the unused `lowEstimateCents` with just `estimateCents`. 

~This PR is a WIP because it has no tests yet. In fact, I believe it will fail to compile on CI.~

Here's what it looks like:

simulator screen shot dec 14 2017 at 3 43 34 pm
simulator screen shot dec 14 2017 at 3 42 27 pm

Tracked in https://github.com/artsy/auctions/issues/177.

### DIFF
--- a/Artsy/Models/API_Models/Auctions/SaleArtwork.h
+++ b/Artsy/Models/API_Models/Auctions/SaleArtwork.h
@@ -41,6 +41,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic, strong) NSNumber *_Nullable openingBidCents;
 @property (nonatomic, strong) NSNumber *minimumNextBidCents;
+@property (nonatomic, strong) NSNumber *_Nullable estimateCents;
 @property (nonatomic, strong) NSNumber *_Nullable lowEstimateCents;
 @property (nonatomic, strong) NSNumber *_Nullable highEstimateCents;
 @property (nonatomic, strong) NSNumber *_Nullable bidCount;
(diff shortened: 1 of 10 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Artsy/Models/API_Models/Auctions/SaleArtwork.h
```diff
@@ -54,7 +55,7 @@ NS_ASSUME_NONNULL_BEGIN
 // This is shared behavior between SaleArtwork and LiveLot
 
 + (NSValueTransformer *)reserveStatusJSONTransformer;
-+ (NSString *)estimateStringForLowEstimate:(NSNumber *_Nullable)lowEstimateCents highEstimateCents:(NSNumber *_Nullable)highEstimateCents currencySymbol:(NSString *)symbol currency:(NSString *)currency;
++ (NSString *)estimateStringForEstimate:(NSNumber *_Nullable)estimateCents lowEstimate:(NSNumber *_Nullable)lowEstimateCents highEstimateCents:(NSNumber *_Nullable)highEstimateCents currencySymbol:(NSString *)symbol currency:(NSString *)currency;
```
erikdstock: Why all one one line like this? This is declaring multiple members of the interface, right?

ashfurrow: Not quite, this adds a new parameter to an existing method. It _is_ a long line – we should probably move to autoformatting Eigen code like with our modern codebases.

orta: > we should probably move to autoformatting Eigen code like with our modern codebases.

We tried this, it was pretty terrible with https://github.com/square/spacecommander which still is the best we have I think

ashfurrow: Oh yeah, it _was_ really bad wasn't it.

PR #2495, opened 2017-12-15
[home] Onboarding users land on For You tab of Home

sarahscott:
If a new user signs up for Artsy, they will be directed to the For You of tab instead of Artists. All other users will land on Artists.

onboarding


orta:
👍

PR #2503, opened 2018-01-04
Make sure modals cannot change back button visibility of presenters #trivial

sarahscott:
Without making any sweeping structural changes, this fix makes sure the payment request modal (or any visible view controller that uses the scroll chief) doesn't change the back button visibility of its presenter.

Fixes https://github.com/artsy/collector-experience/issues/791

#skip_new_tests

### DIFF
--- a/Artsy/View_Controllers/Web_Browsing/ARExternalWebBrowserViewController.m
+++ b/Artsy/View_Controllers/Web_Browsing/ARExternalWebBrowserViewController.m
@@ -138,7 +138,7 @@
 
 - (void)scrollViewDidScroll:(UIScrollView *)scrollView
 {
-    if (self.navigationController && [self.navigationController isKindOfClass:ARNavigationController.class]) {
+    if ([self.navigationController isKindOfClass:ARNavigationController.class]) {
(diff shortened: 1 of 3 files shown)


alloy:
Have you done a pass yet through the app to test if all the VCs that implement the menu aware protocol still work as expected?

sarahscott:
I'll do that now - this would only affect VCs that implement the scroll chief though, not the menu aware protocol.

Artsy/View_Controllers/Web_Browsing/ARExternalWebBrowserViewController.m
```diff
@@ -138,7 +138,9 @@ - (UIScrollView *)scrollView
 
 - (void)scrollViewDidScroll:(UIScrollView *)scrollView
 {
-    [[ARScrollNavigationChief chief] scrollViewDidScroll:scrollView];
+    if (self.navigationController && [self.navigationController isKindOfClass:ARNavigationController.class]) {
```
alloy: The first `nil` check isn’t really needed, because if it were `nil` then objc returns `nil` for any message you send it.

alloy:
All green after a few Travis restarts 👌

PR #2504, opened 2018-01-04
Updates to Emission b12

orta:
Took a while because I needed to find a new flag for pilot

( /cc @KrauseFx have you seen this anywhere else? It kinda came out of nowhere, our last build was two weeks ago and deploys always get the latest fastlane ) (also this is Xcode 8 build, not the Xcode 9 stuff in the [auto_sign](https://github.com/artsy/eigen/tree/orta-auto_sign) branch )
  
@sarahscott you can ignore the dependency CI yellow - I just disabled it for this repo

PR #2505, opened 2018-01-04
Adds support for using RNP in Eigen without Pod changes

### LINKED ISSUE #2497: Eigen's Podfile could use RCTWebSocket, but have it removed on deploy
In the guide: https://github.com/artsy/emission/blob/master/docs/running-emission-in-eigen.md

The biggest faff - and specifically the only thing stopping Eigen from being able to use Emission dev mode any-time is the lack of `RCTWebSocket`. 

We could add `RCTWebSocket` to our Podfile and then remove it in the Fastfile before a `pod install` on circle?

orta:
#trivial 

fixes #2497

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -85,8 +85,15 @@ target 'Artsy' do
   pod 'Extraction'
 
   pod 'Emission', '~> 1.4.0-beta.11'
-  pod 'React/Core'
 
+  # Allow easily running Emission from Metro inside Eigen (see issue #2497)
+  if ENV["CIRCLE_BUILD_NUM"]
+    # Production:
+    pod 'React', :subspecs => %w(Core)
+  else
+    # Development
+    pod 'React', :subspecs => %w(Core DevSupport)
+  end
   # Facebook
   pod 'FBSDKCoreKit', '~> 4.9'
   pod 'FBSDKLoginKit', '~> 4.9'
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -185,8 +185,13 @@ PODS:
     - Yoga (= 0.48.4.React)
   - React/cxxreact_legacy (0.48.4):
     - React/jschelpers_legacy
+  - React/DevSupport (0.48.4):
+    - React/Core
+    - React/RCTWebSocket
   - React/jschelpers_legacy (0.48.4)
   - React/RCTAnimation (0.48.4):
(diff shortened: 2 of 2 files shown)


fastlane/Fastfile
```diff
@@ -102,7 +102,7 @@ lane :ship_beta do
   FileUtils.mv(dsym_archive, ENV['CIRCLE_ARTIFACTS'])
 
   # # Send to the app store
-  pilot changelog: beta_readme
+  pilot changelog: beta_readme, itc_provider: "ArtsyInc"
```
alloy: What’s this change sneaking in?

alloy:
Alright, let’s give it a try!

Podfile
```diff
@@ -85,8 +85,15 @@ target 'Artsy' do
   pod 'Extraction'
 
   pod 'Emission', '~> 1.4.0-beta.11'
-  pod 'React/Core'
 
+  # Allow easily running Emission from Metro inside Eigen (see issue #2497)
+  if ENV["CIRCLE_BUILD_NUM"]
+    # Production:
+    pod 'React', :subspecs => %w(DevSupport)
```
alloy: I’m not sure I understand why we’re adding the dev support submodule in production.

* Didn’t we only need the web socket support?
* Why is Core being added in development but not production?
* This will probably break our shake device gesture, as RN’s dev support tries to use it as well.

fastlane/Fastfile
orta: That's from the now merged #2504

Podfile
orta: You're right, this is backwards - it just needs core

orta:
OK, that's fixed it up

alloy:
The question about enabling RN’s DevSupport still remains. I would have thought that this would break the shake gesture with both Eigen’s and RN’s dev tools fighting for it. Have you tested that?

orta:
Yeah I did get a bit surprised by that. 

Locally my `~` shortcut for Admin wasn't working, but `ctrl + alt + z` triggered our admin menu. I'll give it a fresh look when I have some more CI downtime.

orta:
Gonna call this as not gonna do it. Considering how often the prod version of RN might not compile right compared to the dev version, I don't want to move that kind of fail into deployment.

PR #2511, opened 2018-01-08
[Dev] push notification delegate fixes

l2succes:
This PR:

- Fixes crash when notification payload title is object
- Adds checks to see if our conversation view controller is presented with the same `conversationID` as the one in the notification payload. if yes, just fetch the latest data.
 - Switches to the messaging tab before presenting conversation VC

### DIFF
--- a/Artsy/App/ARAppNotificationsDelegate.m
+++ b/Artsy/App/ARAppNotificationsDelegate.m
@@ -189,7 +189,7 @@
     [notificationInfo setObject:uiApplicationState forKey:@"UIApplicationState"];
 
     NSString *url = userInfo[@"url"];
-    NSString *message = userInfo[@"aps"][@"alert"] ?: url;
+    id message = userInfo[@"aps"][@"alert"] ?: url;
     BOOL isConversation = url && [[[NSURL URLWithString:url] path] hasPrefix:@"/conversation/"];
     
     if (isConversation) {
@@ -213,10 +213,11 @@
             ARConversationComponentViewController * controller = [[[[ARTopMenuViewController sharedController] rootNavigationController] viewControllers] lastObject];
             NSString *conversationID = [notificationInfo[@"conversation_id"] stringValue];
             if ([controller isKindOfClass:ARConversationComponentViewController.class] && [controller.conversationID isEqualToString:conversationID]) {
-                [controller refreshView];
+                [[NSNotificationCenter defaultCenter] postNotificationName:@"notification_received" object:notificationInfo];
             } else {
+                NSString *title = [message isKindOfClass:[NSString class]] ? message : message[@"title"];
                 [ARNotificationView showNoticeInView:[self findVisibleWindow]
-                                               title:message
+                                               title:title
                                             response:^{
                                                 [self tappedNotification:notificationInfo viewController:viewController];
                                             }];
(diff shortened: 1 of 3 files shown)


Artsy/Views/Notifications/ARNotificationView.m
```diff
@@ -69,11 +69,16 @@ + (ARNotificationView *)showNoticeInView:(UIView *)view title:(NSString *)title
     return [self showNoticeInView:view title:title time:6.5 response:response];
 }
 
-+ (ARNotificationView *)showNoticeInView:(UIView *)view title:(NSString *)title time:(CGFloat)time response:(void (^)(void))response
++ (ARNotificationView *)showNoticeInView:(UIView *)view title:(id)title time:(CGFloat)time response:(void (^)(void))response
```
alloy: Why can this be an object at all?

l2succes: true, I can set it to `NSObject *` instead

Artsy/App/ARAppNotificationsDelegate.m
```diff
@@ -209,11 +209,19 @@ - (void)applicationDidReceiveRemoteNotification:(NSDictionary *)userInfo inAppli
...
-            [ARNotificationView showNoticeInView:[self findVisibleWindow]
-                                           title:message
-                                        response:^{
-                                            [self tappedNotification:notificationInfo viewController:viewController];
-                                        }];
+            
+            id controller = [[[[ARTopMenuViewController sharedController] rootNavigationController] viewControllers] lastObject];
+            NSString *conversationID = [notificationInfo[@"conversation_id"] stringValue];
+            if ([controller isKindOfClass:ARConversationComponentViewController.class] && [((ARConversationComponentViewController *) controller).conversationID isEqualToString:conversationID]) {
+                [[NSNotificationCenter defaultCenter] postNotificationName:@"notification_received" object:notificationInfo];
```
alloy: Why do you use a notification here rather than directly messaging the controller?

Artsy/Views/Notifications/ARNotificationView.m
alloy: No sorry, I didn't mean just the param type, I meant where does this object come in from and why is that not just the title?

l2succes: https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/PayloadKeyReference.html#//apple_ref/doc/uid/TP40008194-CH17-SW5

From Apple docs, `alert` can either be a String or Dictionary. In our case, it's a dictionary with `title` and `body`.

```ruby
ApnWrapper.delay.deliver(
        device.id, device.token,
        alert: {
          title: text,
          body: snippet
        },
        sound: sound,
        badge: total_unread_count,
        custom_payload: {
          url: "/conversation/#{conversation_id}",
          conversation_id: conversation_id,
          from_id: conversation_from_user_id,
          item: item
        }
      )
```

Artsy/App/ARAppNotificationsDelegate.m
l2succes: I agree, it'd be more straightforward to send a message directly to the controller. 

The thought process was to try to decouple the modules but there's no way around it in this case.

Artsy/Views/Notifications/ARNotificationView.m
alloy: Ahhh, I see. I don’t think it should be the `ARNotificationView`’s responsibility to deal with this, though. Imo the push notification handler should take out the right string and just pass that.

alloy:
Other than my comment about not making `ARNotificationView` deal with details of push notifications, this looks good to me 👍 Please merge when addressed and green.

orta:
👍

PR #2512, opened 2018-01-09
[home] force set homeVC with selected artist if using universal link

sarahscott:
Not the _most_ elegant way to do this, but it ensures that the selected artist param from a deep link definitely doesn't get lost in the shuffle.

I think I want to spend some more time thinking about how to do this more elegantly (maybe with some work in Emission?) but given time constraints, this at least fixes the bug I found and works in all cases on my physical iPad.

### DIFF
--- a/Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
+++ b/Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
@@ -509,9 +509,16 @@ static const CGFloat ARMenuButtonDimension = 50;
     // If there is an existing instance at that index, use it. Otherwise use the instance passed in as viewController.
     // If for some reason something went wrong, default to Home
     BOOL alreadySelectedTab = self.selectedTabIndex == index;
-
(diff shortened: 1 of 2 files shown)


Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
```diff
@@ -535,6 +542,10 @@ - (void)presentRootViewController:(UIViewController *)viewController animated:(B
     if (!alreadySelectedTab) {
         [self.tabContentView forceSetViewController:presentableController atIndex:index animated:animated];
     }
+    
+    if (showSelectedArtistFromUniversalLink) {
```
sarahscott: `animated` set to `NO` here because this would mean that we are on the `alreadySelectedTab` of Home, and should not do a slide-in animation.

alloy:
Yeah definitely should get some holistic attention after shipping 4.0. Be sure to file a ticket for that!

Have you tested every app/tab state permutation you can think of? If so, go ahead and merge 👍

sarahscott:
I've tested an email universal link with the following states:
- App is not active
- App is backgrounded on a different Home tab
- App is backgrounded on the Artist tab of Home
- App is backgrounded and on a non-Home tab

alloy:
Ace 👍

PR #2531, opened 2018-01-22
[iPhoneX] Converts the status bar to work with iPhone X

orta:
Moves the status bar from each individual ARNAvigationController (e.g. per tab) to the root view controller. 

This means the nav controller's view is now accurately set to include the status bar, which should kill a bunch of hard to diagnose "why doesn't the title look right" bugs we've had with random view controllers and especially artsy views.

#trivial because I already have a entry for this


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
```diff
@@ -131,6 +134,16 @@ - (void)viewDidLoad
     self.view.backgroundColor = [UIColor whiteColor];
     self.selectedTabIndex = -1;
 
+    _statusBarView = [[UIView alloc] init];
+    _statusBarView.backgroundColor = UIColor.blackColor;
+
+    [self.view addSubview:_statusBarView];
+
+    _statusBarVerticalConstraint = [_statusBarView constrainHeight:@"20"];
```
ashfurrow: Should this be `statusBarHeight` instead of a hard coded 20?

orta: No, we can't know that at init time, this is just a generic default

ashfurrow: Makes sense! Want to add a comment to that effect? LGTM.

PR #2535, opened 2018-01-29
[WFU] Do not call forceSetViewController more than once

sarahscott:
The blank bug was caused by `forceSetViewController` being called twice. This is a small fix to make sure it can't happen, but ideally in the long run we can revisit [this tab view](https://github.com/artsy/eigen/blob/master/Artsy/Views/Utilities/ARTabContentView.m#L147)'s responsibilities, because its relationship with ARTopMenuViewController and ARTopMenuNavigationDataSource is a bit confusing.

#skip_new_tests

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -3,6 +3,7 @@ upcoming:
     details: Bug Fixes / iPhone X / ARVIR
     emission_version: 1.4.x
     dev:
+      - Fixes blank view when coming from a deep link to Works For You tab - sarah
       - Updates to Xcode 9.1 - orta
       - Deletes un-used View Controllers - orta
       - Adds a 1024 app icon image - orta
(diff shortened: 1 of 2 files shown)


alloy:
Can you add a CHANGELOG?

I’m going to rebuild CI without cache to see if that fixes the build failure.

CHANGELOG.yml
```diff
@@ -3,6 +3,7 @@ upcoming:
     details: Messaging / Consignments / Tabbed Home
     emission_version: 1.4.0
     dev:
+      - Fixes blank view when coming from a deep link to Works For You tab - sarah
```
orta: I'm not sure this is the right place for it, this is the shipped release from a two weeks ago. You may need to rebase.

sarahscott: Good catch

PR #2536, opened 2018-01-29
Fix the old VIR animations, and remove the status bar

orta:
Will explain inline

### DIFF
--- a/Artsy/View_Controllers/Core/ARViewInRoomViewController.m
+++ b/Artsy/View_Controllers/Core/ARViewInRoomViewController.m
@@ -96,6 +96,11 @@ static const CGFloat DistanceToTopOfBenchPortrait = 90;
     return YES;
 }
 
+- (BOOL)prefersStatusBarHidden
+{
+    return YES;
+}
+
 #pragma mark - UIViewController
 
 - (void)viewDidAppear:(BOOL)animated
@@ -143,7 +148,7 @@ static const CGFloat DistanceToTopOfBenchPortrait = 90;
     UIView *inFrontOfArtworkView = self.chairImageView ?: self.artworkImageView;
     [self.view insertSubview:self.artworkImageView belowSubview:inFrontOfArtworkView];
 
-    UITapGestureRecognizer *exitTapGesture = [[UITapGestureRecognizer alloc] initWithTarget:self.navigationController action:@selector(popViewControllerAnimated:)];
+    UITapGestureRecognizer *exitTapGesture = [[UITapGestureRecognizer alloc] initWithTarget:self action:@selector(tappedArtwork)];
     self.artworkImageView.userInteractionEnabled = YES;
     [self.artworkImageView addGestureRecognizer:exitTapGesture];
 }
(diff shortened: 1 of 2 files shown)


Artsy/View_Controllers/Core/ARViewInRoomViewController.m
```diff
@@ -96,6 +96,11 @@ - (BOOL)hidesStatusBarBackground
     return YES;
 }
 
+- (BOOL)prefersStatusBarHidden
+{
+    return YES;
+}
+
```
orta: Each screen in eigen ( a View Controller ) declares things like `hidesStatusBarBackground`, and as we route them, Eigen change the global settings based on the view controller. This means when we show VIR we hide the status bar text.

Artsy/View_Controllers/Core/ARViewInRoomViewController.m
```diff
@@ -143,7 +148,7 @@ - (void)setArtworkImageView:(UIImageView *)artworkImageView
     UIView *inFrontOfArtworkView = self.chairImageView ?: self.artworkImageView;
     [self.view insertSubview:self.artworkImageView belowSubview:inFrontOfArtworkView];
 
-    UITapGestureRecognizer *exitTapGesture = [[UITapGestureRecognizer alloc] initWithTarget:self.navigationController action:@selector(popViewControllerAnimated:)];
+    UITapGestureRecognizer *exitTapGesture = [[UITapGestureRecognizer alloc] initWithTarget:self action:@selector(tappedArtwork)];
```
orta: We were working on a fun assumption, which is looked like it broke in Xcode 9. Previously, when this tap gesture was triggered code like this was run:

```objc
[self.navigationController popViewControllerAnimated:[the tap gesture]];
```

and because animated got _anything_ it was assumed to be true. The last parameter being an bool, which has always worked. However, now it doesn't ;) - maybe they did something to tighten the type system at runtime for Swift, I dunno.

So this now calls a function that calls it properly 

```objc
 - (void)tappedArtwork
 {
     [self.navigationController popViewControllerAnimated:YES];
 }
```

orta:
Fixes https://artsyproduct.atlassian.net/browse/IN-55

Artsy/View_Controllers/Core/ARViewInRoomViewController.m
ashkan18: oh yes, i do remember looking at this 👓  👍

CHANGELOG.yml
```diff
@@ -10,6 +10,7 @@ upcoming:
     user_facing:
       - iPhone X support - orta
       - Augmented Reality View in Room lab feature, defaults to off - orta/lazerwalker
+      - Removes the status bar in the old VIR - orta
```
ashkan18: This CHANGELOG format is really nice, seems easier to modify compared to Volt's

Artsy/View_Controllers/Core/ARViewInRoomViewController.m
alloy: Could it be that the gesture recogniser gets released _before_ the message is delivered? That would make it `nil`/`0`

PR #2537, opened 2018-01-29
Get both whether a vanity url is a martsy page, or a fair at the same time as loading the webview

orta:
Fixes https://artsyproduct.atlassian.net/browse/IN-13

## Currently, when we hit a vanity URL

* We see if the profile metadata, then we start showing martsy

## Now, when we hit a vanity URL

* We load martsy
* If on a phone: we see if the profile metadata say it's a fair, and we kill the martsy view and replace it with a FairVC

Should make martsy views overall faster by maybe a second.


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Artsy/View_Controllers/Fair/ARProfileViewController.m
```diff
@@ -60,29 +60,40 @@ - (UIStatusBarStyle)preferredStatusBarStyle
...
+        // We need to figure out if it's a fair URL or not
+        //
+        // So let's load the martsy view now, and at the same time
+        // make a request to find whether the vanity URL represents a
+        // fair, so that we can show the native VCs on iPhone
+        [self loadMartsyView];
+        [self ar_presentIndeterminateLoadingIndicatorAnimated:YES];
 
-    [ArtsyAPI getProfileForProfileID:self.profileID success:^(Profile *profile) {
+        [ArtsyAPI getProfileForProfileID:self.profileID success:^(Profile *profile) {
```
ashfurrow: This might be an edge case not worth addressing, but what would happen if the martsy view was loaded before the API finished? I worry about martsy appearing briefly and then being replaced with a spinner.

orta: The line above will present a spinner _over_ martsy, so it'll be 🆗

ashfurrow: Mmm roger that, missed that. Looks good!

PR #2543, opened 2018-02-13
Updates Nokogiri

ashfurrow:
It has a security vulnerability and it probably doesn't matter for our purposes but I'm tired of dismissing this warning:
screen shot 2018-02-13 at 12 23 15 pm


orta:
Works for me

PR #2547, opened 2018-02-23
[dev] Removes redundant call to ARAppNotificationsDelegate

sarahscott:
This makes sure the ARAppNotificationsDelegate is called only once when a push notification is received.


alloy:
Let’s wait with merging this until you’ve had the chance to check that everything works as expected on iOS 9 and 11.

orta:
I'm gonna close this for now as we've not done the testing for it, if it comes back up we should cherry-pick this commit back into a future PR.

PR #2553, opened 2018-03-02
Take another stab at connecting our Adjust tracking of URLs

orta:
@anipetrov mentioned that the anon sessions weren't hooking up to the same ids. 

What this PR does is ensure that the session has anon user stored before making _any_ adjust calls, rather than waiting until a user has completed creating a user.


alloy:
Sure, I guess that’s the best way to unsure we never regress again 👍

PR #2564, opened 2018-03-12
Update MP's staging URL to https

ashkan18:
# Problem
We are now forcing SSL on staging to match production. 

# Solution
Update default staging url to `https`.


alloy:
🤘

PR #2574, opened 2018-03-21
minor change to display string on AttributionClassification

oxaudo:
https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=36&projectKey=SELL&modal=detail&selectedIssue=SELL-240&sprint=49


orta:
CHANGELOG entry please 👍

oxaudo:
@orta - updated. Is that considered a user facing change? I though it's minor enough to not require changelog entry...

orta:
It's a bit of both, it's a user facing change so we try include all of those in there so people can test on betas - even if we don't present those in the changelog to the app store

PR #2581, opened 2018-04-10
EV-56 - Ensure we have a black BG for the ARVIR

orta:
Fixes https://artsyproduct.atlassian.net/browse/EV-56

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -3,7 +3,8 @@ upcoming:
   date: TBA
   emission_version: 1.4.6
   user_facing:
-    -  Buy now works show in an Auction screen - Chris
+    -  [WIP] Buy now works show in an Auction screen - Chris
+    -  ARVIR fixes & improvements - orta
   
 releases:
   - version: 4.1.0
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -3,7 +3,7 @@ upcoming:
   date: TBA
   emission_version: 1.4.6
   user_facing:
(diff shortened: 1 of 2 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">CHANGELOG isn’t valid YAML</td>
    </tr>
  </tbody>
</table>

</p>

CHANGELOG.yml
```diff
@@ -3,7 +3,8 @@ upcoming:
   date: TBA
   emission_version: 1.4.6
   user_facing:
-    -  Buy now works show in an Auction screen - Chris
+    -  WIP Buy now works show in an Auction screen - Chris
+    -  ARVIR fixes & improvements - orta
```
oxaudo: Not that I have anything to add here but just reading the change-log I'm wondering:
- why the first comment marked as WIP now?
- would not it be helpful to mark the second change something like `ARVIR background color chage`? Since I assume next commit will have it's own changelog entry so it might potentially be a lot of `ARVIR fixes & improvements` from Orta?

orta: 1. it went through QA, and another team are going to take the time to ship it - I added WIP so that it isn't accidentally deployed
2. I don't think they're important enough to warrant individual changelog entries, so I just made a generic one

PR #2588, opened 2018-04-17
EV-52 - Add shadows to the action buttons on ARVIR

orta:
Fixes - https://artsyproduct.atlassian.net/browse/EV-52


mennenia:
I guess this adds a shadow :P 

👍

PR #2592, opened 2018-04-18
EV-55 - Fixes the Artwork not being able to scroll correctly when moving into VIR

orta:
Re: https://artsyproduct.atlassian.net/browse/EV-55

I decoupled the setting of the new toolbar state from when the artwork is still the top view controller for the navigation stack. This means that the changes in the layout don't affect the ArtworkVC and it becomes immune to the bug.


mennenia:
Great 👍

PR #2593, opened 2018-04-18
[ARVIR] After placing an artwork, don't update the positions of walls

orta:
With this merged ARVIR does not update the wall positioning after an Artwork has been placed, [see slack](https://artsy.slack.com/archives/C9ZHUJXEZ/p1524062272000754) for full context.


mennenia:
Seems legit!

PR #2595, opened 2018-04-24
Fixes crash attaching photos to a consignment submission.

ashfurrow:
A friend reported that the app crashed while attaching a photo to their consignment submission. I found [the corresponding crash report](https://sentry.io/artsynet/eigen/issues/452245164/events/18647931286/) and took a very quick look, and it appears that as of iOS 11, the `NSPhotoLibraryAddUsageDescription` key is required to be included in an app's `Info.plist`, or else the following SIGABRT is raised:

```
abort | NSPhotoLibraryAddUsageDescription | This app has crashed because it attempted to access privacy-sensitive data without a usage description.  The app's Info.plist must contain an NSPhotoLibraryAddUsageDescription key with a string value explaining to the user how the app uses this data.
```

This PR adds the key, which should fix the crashes.


mennenia:
Looks great, nice spot! Resolved the merge conflict in the changelog. Just waiting for 🍏, and then I'll merge :)

PR #2597, opened 2018-04-27
ARVIR bug fixes

orta:
Handles:

* https://artsyproduct.atlassian.net/browse/EV-88
* https://artsyproduct.atlassian.net/browse/EV-89
* https://artsyproduct.atlassian.net/browse/EV-87
* https://artsyproduct.atlassian.net/browse/EV-90

### DIFF
--- a/Artsy/View_Controllers/ARVIR/AR/ARAugmentedVIRInteractionController.m
+++ b/Artsy/View_Controllers/ARVIR/AR/ARAugmentedVIRInteractionController.m
@@ -110,7 +110,7 @@ NSInteger wallHeightMeters = 5;
             if (!self.ghostWork) {
                 SCNArtworkNode *ghostBox = [SCNArtworkNode nodeWithConfig:self.config];
                 SCNNode *ghostWork = [SCNNode nodeWithGeometry:ghostBox];
-                ghostWork.opacity = 0.5;
+                ghostWork.opacity = 0.55;
 
                 [result.node addChildNode:ghostWork];
                 self.ghostWork = ghostWork;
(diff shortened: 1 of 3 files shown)


mennenia:
LGTM. I know Erik is assigned, but happy to merge if you want @orta

Artsy/View_Controllers/ARVIR/AR/ARAugmentedVIRInteractionController.m
```diff
@@ -159,11 +158,11 @@ - (void)renderer:(id<SCNSceneRenderer>)renderer didUpdateNode:(SCNNode *)node fo
 - (void)renderer:(id <SCNSceneRenderer>)renderer didAddNode:(SCNNode *)node forAnchor:(ARAnchor *)anchor API_AVAILABLE(ios(11.0));
 {
     // Only handle adding plane nodes
-    if(!anchor) { return; }
-    if(![anchor isKindOfClass:ARPlaneAnchor.class]) { return; }
+    if (!anchor) { return; }
+    if (![anchor isKindOfClass:ARPlaneAnchor.class]) { return; }
```
erikdstock: General swift question- these conditionals cannot be chained?

ashfurrow: In Swift, this could be simplified to something like `if (anchor?.isKindOfClass(ARPlaneAnchor.class)` or something, but this is Objective-C and it doesn't have the `?` operator. That said, I think the first `if` here might be superfluous, since if `anchor` is `nil`, then calling `isKindOfClass:` on `nil` will return `NO`, which gets negated and the `return` gets executed. No harm in keeping it in, I do like that it's clear what's going on.

ashfurrow: But yeah, the conditionals as-is could be chained using `&&` as normal. Short-circuiting of the boolean expression will work as you would expect it to.

PR #2600, opened 2018-05-02
Fixes unit tests from Emission upgrade.

ashfurrow:
This is succeeding locally:

screen shot 2018-05-02 at 3 07 03 pm


orta:
Cool, merge on green 👍

ashfurrow:
Mmm it failed with an error I saw locally (but disappeared when I re-ran `pod install`, I think it fails the first time). Going to try rebuilding not-from-cache.

PR #2601, opened 2018-05-03
Fix Emission VCs not being considered Root 

mennenia:
Fix [PURCHASE-103](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=41&projectKey=PURCHASE&modal=detail&selectedIssue=PURCHASE-103)

This is far from an elegant solution. Another option is amending the Emission View Controllers to have the `isRootNavViewController` property (which is currently considered `false`, which caused the bug).  That would mean an Emission release though, and I opted in favour for this, for the time being.

### DIFF
--- a/Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
+++ b/Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
@@ -531,17 +531,10 @@ static const CGFloat ARMenuButtonDimension = 50;
     }
 
     if ([viewController respondsToSelector:@selector(isRootNavViewController)] && [(id<ARRootViewController>)viewController isRootNavViewController]) {
-        [self presentRootViewController:viewController animated:animated];
-    } else if ([self isEmissionRootViewController:viewController]){
-        // Override animation, because tab switching should never be animated
         [self presentRootViewController:viewController animated:NO];
     } else {
         [self.rootNavigationController pushViewController:viewController animated:animated];
     }
-}
-
-- (BOOL)isEmissionRootViewController:(UIViewController *)viewController {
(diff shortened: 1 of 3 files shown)


Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
```diff
@@ -532,11 +532,18 @@ - (void)pushViewController:(UIViewController *)viewController animated:(BOOL)ani
 
     if ([viewController respondsToSelector:@selector(isRootNavViewController)] && [(id<ARRootViewController>)viewController isRootNavViewController]) {
```
mennenia: This was considered false, when dealing with Emission VCs (as they are not `ARRootViewControllers`).

mennenia:
Definitely 👍

alloy:
I would have expected that to work through this class extension, e.g. https://github.com/artsy/eigen/blob/41a0e35850ff4792ae3a0d7c700642749649ef2f/Artsy/App/ARAppDelegate%2BEmission.m#L238-L241. Maybe it did work for home but we just need more extensions for the other now RN root VCs?

mennenia:
Ha, it's already there. 

I tested it with Home, Inbox, and Favorites. All are being pushed, rather than "tabbed" if you will, and skip that `if` block. 

I can step through and test a bit more though! 👍

mennenia:
@alloy I might have been using the wrong route for home, but yes this works! Thanks 👍 

I did keep the `animated` set to `NO`, because if we're swapping root, it should never be animated from what I can think of. Without changing that, it animates a "push" currently.

alloy:
Makes sense 👍

PR #2609, opened 2018-05-10
BidFlow integration

ashfurrow:
This is a WIP because it's not done yet, but it integrates the new BidFlow component from Emission into Eigen.

There is a problem with the contract that Emission's BidFlow component exposes. Currently, it requires a sale artwork ID. However, Eigen needs to be able to handle Force-like URLs for bidding (from emails, push notifications, etc). These are in the format of `/auction/:id/bid/:artwork_id`, which is missing the sale artwork ID altogether. So, I will be opening a PR on Emission next.

EDIT:

Okay things are working! I've added an option to the debug menu, accessible through shaking your phone or pressing the ``` key in the simulator:

simulator screen shot - s - 2018-05-15 at 10 16 27

*Note: you _must_ register for the sale using the old bid flow.* 

It doesn't require an app restart. It's not pixel-perfect, but we can refine later:

simulator screen shot - s - 2018-05-15 at 14 37 24

### DIFF
--- a/Artsy/Models/API_Models/Auctions/SaleArtwork.m
+++ b/Artsy/Models/API_Models/Auctions/SaleArtwork.m
@@ -37,7 +37,6 @@ static NSMutableDictionary <NSString *, NSNumberFormatter *> *formattersPerCurre
     return @{
              ar_keypath(SaleArtwork.new, saleArtworkID) : @"id",
              ar_keypath(SaleArtwork.new, currencySymbol) : @"symbol",
-             ar_keypath(SaleArtwork.new, internalSaleArtworkID) : @"_id",
              ar_keypath(SaleArtwork.new, openingBidCents) : @"opening_bid_cents",
              ar_keypath(SaleArtwork.new, minimumNextBidCents) : @"minimum_next_bid_cents",
              ar_keypath(SaleArtwork.new, saleHighestBid) : @"highest_bid",
--- a/Artsy/Models/API_Models/Auctions/SaleArtwork.m
+++ b/Artsy/Models/API_Models/Auctions/SaleArtwork.m
@@ -37,7 +37,7 @@ static NSMutableDictionary <NSString *, NSNumberFormatter *> *formattersPerCurre
     return @{
              ar_keypath(SaleArtwork.new, saleArtworkID) : @"id",
              ar_keypath(SaleArtwork.new, currencySymbol) : @"symbol",
-             ar_keypath(SaleArtwork.new, saleArtworkID) : @"id",
+             ar_keypath(SaleArtwork.new, internalSaleArtworkID) : @"_id",
              ar_keypath(SaleArtwork.new, openingBidCents) : @"opening_bid_cents",
              ar_keypath(SaleArtwork.new, minimumNextBidCents) : @"minimum_next_bid_cents",
              ar_keypath(SaleArtwork.new, saleHighestBid) : @"highest_bid",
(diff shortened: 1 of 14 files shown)


Artsy/View_Controllers/Artwork/ARArtworkViewController+ButtonActions.m
```diff
@@ -189,27 +190,20 @@ - (void)tappedConditionsOfSale
...
-- (void)tappedBidButton:(UIButton *)button
-{
-    button.enabled = false;
-    [self.artwork onSaleArtworkUpdate:^(SaleArtwork *saleArtwork) {
-        button.enabled = true;
-        [self bidCompleted:saleArtwork];
-    } failure:^(NSError *error) {
-        ARErrorLog(@"Can't get sale to bid for artwork %@. Error: %@", self.artwork.artworkID, error.localizedDescription);
-    }];
-}
```
ashfurrow: This extra method, with its network call to update the sale artwork, was completely unnecessary. This method is only ever called from an object that already _has_ the sale artwork model. So many round trips! I've removed the extra call, even if the whole flow will be replaced soon.

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">PR is classed as Work in Progress</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

ashfurrow:
So the way that auctions are currently created in Artsy, the Batch Import tool (artsy/currents), no artwork is ever in two sales. In fact, this assumption already exists in Metaphysics' schema: https://github.com/artsy/metaphysics/blob/33eac8c55c1d0fd811872aaf541133683d4b3101/src/schema/artwork/index.js#L578-L596

So I'm pretty sure we can just _ignore_ the saleID in the route, since it's impossible to even use it to query Metaphysics for a specific sale artwork.

sweir27:
I think it's reasonable that an artwork won't be _biddable_ in two sales at one time, but an artwork _can_ technically have multiple sale artworks, and in general I think it's a "bug" that we've always defaulted to the first one (since it makes it impossible for us to represent an artwork even being sold in multiple sales by the same partner). You're right that there are many forces at play (i.e. currents 😛, client implementations) that prevent this association from actually being utilized, but it's very possible.

Also note that when we place bids, we do [require](https://github.com/artsy/metaphysics/blob/master/src/schema/me/bidder_position_mutation.js#L73-L83) both the `artwork_id` and the `sale_id`, but we can probably get that from the sale artwork? I'm not sure if any of this info affects this PR, just mentioning!

ashfurrow:
Okay. Hmm. Okay what if we did this: give the `sale_arwork` field on `Artwork` an `:id` string parameter with an empty default. We can use the given parameter when returning the sale artwork, or default to the first. Given the discussions around data modelling we've had lately, I think it makes sense to keep the requests for models as specific as possible.

ashfurrow:
Okay, the code for this is done. Just waiting for https://github.com/artsy/emission/pull/1048 to be merged, then we can release Emission 1.5.1 with the changes needed for this PR. CI will fail in the mean time.

ashfurrow:
Okay, just waiting on this MP PR: https://github.com/artsy/metaphysics/pull/1045

Artsy/Models/API_Models/Auctions/SaleArtwork.m
```diff
@@ -37,7 +37,6 @@ + (NSDictionary *)JSONKeyPathsByPropertyKey
     return @{
              ar_keypath(SaleArtwork.new, saleArtworkID) : @"id",
              ar_keypath(SaleArtwork.new, currencySymbol) : @"symbol",
-             ar_keypath(SaleArtwork.new, saleArtworkID) : @"id",
```
ashfurrow: This was a duplicate of line 38, so I took it out.

mennenia:
Looks awesome! Nice clean up also 💕

Finally merging ✨

PR #2611, opened 2018-05-15
Deploys merge commits for 4.1.2

l2succes:
For the current beta


orta:
Nice, you found a Danger ruby bug. 

I'm gonna ignore that cause we're pretty far behind on that.

PR #2612, opened 2018-05-17
[Auctions] Add support for showing auctions home tab on /auctions push notification

mennenia:
So, this will be corresponding with [Emission 1052](https://github.com/artsy/emission/pull/1052).

The switchboard registration and `rootNavigationControllerHomeWithTab` are legit, I think.

The code written in `presentRootViewController` is just... I don't know, I did what _I_ thought made sense, but I'm not 100% convinced. 

Basically, there are a few things that are odd:
1) There is no final `else`. There are cases in which nothing is presented, because nothing is being force set, as some of the variables/flags don't apply. Why is this a present VC method then? -> So I added a final clause
2) Favourites was missing, and would end up as a Home VC
3) (from my understanding) it essentially overwrites the passed through View Controller in the case of home, because it gets the rootNavigationVC from its `rootNavigationControllerAtIndex` method. That's fine if switching tabs to an existing RootNav, but not sufficient for needing to show a new home tab, as the information you needed to switch in the first place is now lost. If the index is 0 (meaning, home), it will default to Artists (the first tab within home). So I hope my change makes sense here.

I also deleted some code from an old TODO 😄 

Oh, and please test with a new Emission 1.4.x before merging 💕

### DIFF
--- a/Artsy Stickers/Info.plist
+++ b/Artsy Stickers/Info.plist
@@ -17,7 +17,7 @@
 	<key>CFBundlePackageType</key>
 	<string>XPC!</string>
 	<key>CFBundleShortVersionString</key>
-	<string>4.1.2</string>
+	<string>4.1.3</string>
 	<key>CFBundleVersion</key>
 	<string>2018.04.18.12</string>
 	<key>NSExtension</key>
--- a/Artsy/App_Resources/Artsy-Info.plist
+++ b/Artsy/App_Resources/Artsy-Info.plist
@@ -36,7 +36,7 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>4.1.2</string>
+	<string>4.1.3</string>
 	<key>CFBundleSignature</key>
 	<string>????</string>
 	<key>CFBundleURLTypes</key>
--- a/Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
+++ b/Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
@@ -297,7 +297,7 @@ static const CGFloat ARMenuButtonDimension = 50;
     }
     
     [switchboard registerPathCallbackAtPath:@"/works-for-you" callback:^id _Nullable(NSDictionary * _Nullable parameters) {
-        return [self rootNavigationControllerAtIndex:ARTopTabControllerIndexHome parameters:parameters].rootViewController;
+        return [self rootNavigationControllerHomeWithTab:ARHomeTabArtists].rootViewController;
     }];
     
     [switchboard registerPathCallbackAtPath:@"/auctions" callback:^id _Nullable(NSDictionary * _Nullable parameters) {
@@ -611,8 +611,6 @@ static const CGFloat ARMenuButtonDimension = 50;
     } else if (showWorksForYouWithSelectedArtistFromUniversalLink) {
         // We are already on Home, and need to force the tab view to show the new Home with its selected artist & without animation
         [self.tabContentView forceSetViewController:presentableController atIndex:ARTopTabControllerIndexHome animated:NO];
-    } else {
-        [self.tabContentView forceSetViewController:presentableController atIndex:index animated:NO];
     }
 }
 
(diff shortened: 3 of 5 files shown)


ashfurrow:
Cool, so I pulled down this code and ran it against a local, up-to-date version of Emission, and we found a problem when opening the app and switching to the Auctions tab:

simulator screen shot - s - 2018-05-17 at 16 53 01

Division by zero, a classic 😄 

Some notes:

- Couldn't reproduce the bug in the Emission Example app, so it'll need to be debugged from Eigen against the local packager.
- Despite being run in Dev mode, the React Native Developer Menu (⌘D) isn't showing up, something to take a look at.
- After fixing this, we still need to run through an end-to-end testing scenario when pushing a notification from the staging console, in foreground, background, and closed app states.

alloy:
@ashfurrow I think I’ve seen this error locally before, but have yet to see it happen in a production build or in Sentry crash reports. If you happen to easily find the cause then by all means do fix it, but I would time box heavily. Regarding the RN dev tools, these are not included in (production) builds of Emission.

ashfurrow:
Sounds good. I was running Eigen against a local packager with `dev=true` in the query string, which I think should work, right?

ashfurrow:
Okay so just documenting my progress, I've only had a short time to look at this, but here's by rubber-ducking: 

- Running against local packager, we can trick the app into launching with a remote notification by adding `remoteNotification = @{@"url": @"https://staging.artsy.net/auctions"};` to the app delegate
- Doing this produces the error I mention above. It's unclear if this will impact users in production builds.
- Error appears to be caused by _any_ push notification routing to a home page tab, such as `/works-for-you`.

ashfurrow:
Getting around the division by zero error still gives a blank screen, this need further investigation:

screen shot 2018-05-18 at 11 57 39 am

I put `forceSetViewController:` in a dispatch async to get around the error but it's insufficient.

ashfurrow:
Okay so I've made progress in understanding what's going on, because this is a bit more complicated than it appears. 

<details>

<summary>I'm stating the obvious a bit, but just so we're on the same page...</summary>

- Eigen's routing is centred around "get a route, return a view controller, present that controller"
- But handling a push notification is centred around _changing the current controller_ to show the auctions tab.

We get around this by changing the view controller's selected tab in `rootNavigationControllerHomeWithTab:` and then returning the existing on-screen controller. This leads to blank screens because `forceSetViewController:` doesn't handle force-setting the _already visible_ view controller. Additionally, setting the tab of an on-screen controller isn't taking effect.
</details>

---

So there are a two problems we need to address:

- The `changeHomeTabTo:` function on `ARHomeComponentViewController` isn't working as we expect it to. It changes the tab at the top but doesn't load the contained components content.
- `forceSetViewController:` needs to handle controllers that are already set, and ignore those situations.

ashfurrow:
Okay! A lot of progress today. We're waiting on [a final Emission analytics tweak](https://github.com/artsy/emission/pull/1056) but in the meantime, I've tested this by archiving a production build locally and installing on my phone, and sending notifications through staging. Things look good! Mostly...

Ran into a small problem. Steps to reproduce:

- Close app completely
- Send a push notification to /auctions
- Open the notification
- Tap favourites tab
- Tap back to home tab

Expected results: Auctions tab is still selected.
Actual results: Artists tab is still selected.

Happens also when opening the multitasking tray. Hmm!

I reproduced the problem in the simulator. Looks like it's the `page` prop of the react-native-scrollable-tab-view component is causing problems, but it was only [added recently](https://github.com/artsy/emission/commit/8819e30f60615fd8243da9b6bcf417d802accd8b#diff-9820b1884af89490c91f1af390092b35R93). Looks like ultimately [this code](https://github.com/skv-headless/react-native-scrollable-tab-view/blob/a50e3f07e07898099f6fd3d88ba9835979d4185a/index.js#L115-L117) will change to the `page` tab when it receives new props, such as `isVisible` from our `ARComponentViewController`'s `viewWillDisappear:`. 

@mennenia Do you remember why we needed to supply a `page` prop to the `ScrollableTabView`? Looks like the [project's readme](https://github.com/skv-headless/react-native-scrollable-tab-view#props) specifies it can be buggy 😅 From what I can gather, the component should be set up with `initialTab` and then changed with `goToPage`, thoughts on removing the `page` prop?

I tested opening the notification from the foreground, background, and closed states. Everything looks good – I'd want to test one more time on the beta before submitting, but I'm confident it'll work (once we make the change above).

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">You need to set the App’s plist version to 4.1.3 - use <code>make update_bundle_version</code></td>
    </tr>
  </tbody>
</table>

</p>

mennenia:
I was wondering why it was green on circle and not here! That's great 🎉

PR #2617, opened 2018-05-23
Updates pre-deploy docs

ashfurrow:
This does a few things:

- Moves testing changes specified in the changelog to the top since it seems like it's the most important and also most likely to spoil a beta.
- Adds detail on how to test certain things, for example where to open a mocktion on staging.
- Removes the inquiry flow test, which is covered by the new conversation push notification test, and replaces it with a smoke test of a populated conversation view.
- Adds steps for our new auctions push notification testing.


orta:
👍

PR #2618, opened 2018-05-24
ARVIR tweaks

orta:
* Makes the iPhone app compile when compiling for a device (not a sim)
* Uses the new image assets for the beta/close buttons https://artsyproduct.atlassian.net/browse/EV-144
* Updates the type on the setup page https://artsyproduct.atlassian.net/browse/EV-142

#trivial because ARVIR is in already


orta:
Merge on green.

PR #2620, opened 2018-05-25
Exposes admin menu to non-admin users via URL

ashfurrow:
This was actually already done! Eigen handles the `/admin` route already so this PR just removes the check when instantiating that view controller. I looked around and I don't see anywhere else that calls that function, so we're good.

**One note**: In Safari I had to use triple-slash to get the route recognized by JRRoutes (I think because the URL is missing a host name). So the URL to visit in Safari is `artsy:///admin`.


orta:
Good ideas ey?

PR #2628, opened 2018-06-01
Fixes some typos, adds Jira ticket to docs

ashfurrow:
Fixed two typos and adds a link to a Jira ticket around something that's not working.


ashfurrow:
The new works-for-you push code in Gravity doesn't appear to be idempotent, it can only be invoked once successfully every so often. So the docs aren't clear on that either, not sure next steps on that.

mennenia:
LGTM 👍

PR #2632, opened 2018-06-05
Switches the back and beta positions on the floor based ARVIR

orta:
Brings it in line with how having it work on a push instead of the current design for modals

#trivial


alloy:
👍

PR #2635, opened 2018-06-06
[ARVIR] Use a higher point in the screen for projecting the artwork vs the wall

orta:
Re-aligns the artwork to use a different screen point of reference in comparison to the floor. This fixes a weird UX issue. EV-209


zephraph:
Looks good to me!

Merge on green. 🤞

PR #2645, opened 2018-06-19
Updates Emission to 1.5.5

ashfurrow:
⚠️ Do not merge! ⚠️ 

We're trying to get an Eigen release out the door, would rather not complicate things with an Emission update. Also I can't find where to add the `pod keys set` command for our deploys.


orta:
Merge on green

PR #2658, opened 2018-07-03
increase modal close hit area

erikdstock:
This PR makes the hit area for the modal exit button 10 bigger.
image


sweir27:
Coool! Merge on green 🌮

PR #2663, opened 2018-07-10
Fix tiny white box on bidding form

erikdstock:
Tracked in [purchase-220](https://artsyproduct.atlassian.net/browse/PURCHASE-220)
cc @sweir27 @yuki24 @zephraph 

This PR is an attempt to address the issue mentioned above- a tiny white box in the upper left corner of the screen which becomes visible when you scroll the Billing Address form (here covering the 'F' in 'Full Name':
image

This seems to address the issue but I'm not sure if it is the best way or what other implications might come along- I just shoehorned in a single check for presence of `nav.title` and `nav.length`.

View debugger shows it is an ARSerifLabel:
image

Updated debugger view hierarchy:
image

### DIFF
--- a/Artsy/View_Controllers/ARSerifNavigationViewController.m
+++ b/Artsy/View_Controllers/ARSerifNavigationViewController.m
@@ -109,29 +109,30 @@
         [navBar hideNavigationBarShadow:false];
 
     } else {
-        // On the root view, we want a left aligned title.
-        UILabel *label = [ARSerifLabel new];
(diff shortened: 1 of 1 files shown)


erikdstock:
This was right on the first go? I was sure there would be more to it. 😅

erikdstock:
Talked to @ashfurrow about this and the only concern was whether this title might change from `nil/""` to something else and not be able to update later. I'm unable to find anywhere in the code where this happens so I'm going to try to smoke any weirdness out and then merge.

PR #2665, opened 2018-07-13
Removes duplicate nav bar from BidFlow.

ashfurrow:
This addresses [PURCHASE-239](https://artsyproduct.atlassian.net/browse/PURCHASE-239), which was caused by a double navigation bar. Basically:

- Eigen puts the `ARBidFlowViewController` inside a navigation controller with a navigation bar.
- Emission uses `NavigatorIOS` inside its `BidFlow` component, which is itself a navigation controller with its own navigation bar.

Normally, iOS forbids nesting navigation controllers within one another, but our use of React Native circumvents the normal checks that iOS does. This was leading to other problems, so this PR should also fix [PURCHASE-234](https://artsyproduct.atlassian.net/browse/PURCHASE-234).

Here's the before/after on iPhone:

before after

And here's what it looks like on iPad:

simulator screen shot - ipad air 2 - 2018-07-13 at 14 46 23

I just kind of guessed on the iPad, since it looked pretty gnarly as wasn't a good reference.


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

sweir27:
AMAZING!!! Thank you @ashfurrow ! 🎉 🎉

erikdstock:
AWESOME

PR #2666, opened 2018-07-16
Updates the check in supressing notification views to match new format

orta:
in https://github.com/artsy/pulse/pull/215

#trivial

### DIFF
--- a/Artsy/App/ARAppNotificationsDelegate.m
+++ b/Artsy/App/ARAppNotificationsDelegate.m
@@ -218,12 +218,15 @@
             NSString *artworkID = notificationInfo[@"artwork_id"];
             NSString *action = notificationInfo[@"action"];
 
+            // We check whether a notification coming through has this as its action
+            NSString *outbidNotificationLabel = @"bid outbid";
+
             ARConversationComponentViewController *conversationVC = (id)controller;
             ARBidFlowViewController *bidFlowVC = (id)controller;
 
             if ([controller isKindOfClass:ARConversationComponentViewController.class] && [conversationVC.conversationID isEqualToString:conversationID]) {
                 [[NSNotificationCenter defaultCenter] postNotificationName:@"notification_received" object:notificationInfo];
-            } else if ([controller isKindOfClass:ARBidFlowViewController.class] && [action isEqualToString:@"outbid"] && [bidFlowVC.artworkID isEqualToString:artworkID] && [bidFlowVC.saleID isEqualToString:saleID]) {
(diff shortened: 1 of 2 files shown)


Artsy/App/ARAppNotificationsDelegate.m
```diff
@@ -218,12 +218,15 @@ - (void)applicationDidReceiveRemoteNotification:(NSDictionary *)userInfo inAppli
...
+            // We check whether a notification coming through has this as its action
+            NSString *outbidNotificationLabel = @"bid outbid";
+
             ARConversationComponentViewController *conversationVC = (id)controller;
             ARBidFlowViewController *bidFlowVC = (id)controller;
 
             if ([controller isKindOfClass:ARConversationComponentViewController.class] && [conversationVC.conversationID isEqualToString:conversationID]) {
                 [[NSNotificationCenter defaultCenter] postNotificationName:@"notification_received" object:notificationInfo];
-            } else if ([controller isKindOfClass:ARBidFlowViewController.class] && [action isEqualToString:@"outbid"] && [bidFlowVC.artworkID isEqualToString:artworkID] && [bidFlowVC.saleID isEqualToString:saleID]) {
+            } else if ([controller isKindOfClass:ARBidFlowViewController.class] && [action isEqualToString:outbidNotificationLabel] && [bidFlowVC.artworkID isEqualToString:artworkID] && [bidFlowVC.saleID isEqualToString:saleID]) {
```
ashfurrow: We're still seeing the notifications. I took a look at this and I think the problem is this line. Specifically I think it's the `[controller isKindOfClass:ARBidFlowViewController.class]` clause of the `if` statement. This `controller` comes from `[self getGlobalTopViewController]` on line 214. However, when the app is open to the bid flow, I can pause the app and write the following in the debugger:

```objc
(lldb) po [[[JSDecoupledAppDelegate sharedAppDelegate] remoteNotificationsDelegate] getGlobalTopViewController]
<ARArtworkSetViewController: 0x7fb624144600>
```

Which is kind of what I'd expect, since that's the artwork view controller. The BidFlow view controller is presented as a modal, so we need this `if` statement to check for the _presented_ view controller (which is a navigation controller, so we need to check it's view controller stack – this is getting complicated!). In the debugger again:

```objc
(lldb) po [[[[[[JSDecoupledAppDelegate sharedAppDelegate] remoteNotificationsDelegate] getGlobalTopViewController] presentedViewController] viewControllers] first]
<ARBidFlowViewController: 0x7fb623dcbb00>
```

Perfect, that's what we're looking for. It's not as easy as replacing the `if` statement, because of types and stuff, but it shouldn't be a difficult fix. I'll get on it 👍

PR #2667, opened 2018-07-16
Adds default 'bid' button label behaviour.

ashfurrow:
This is a fix for [Purchase-240](https://artsyproduct.atlassian.net/browse/PURCHASE-240). The main changes are to the code in `ARBidButton.m`. I've added a new, more specific method for setting the auction state on the bid button that let's us define if the button should said "Bid" by default, or "Register". "Register" was the old default value, so I used `ARBidButtonIntentRegister` as a default value for the more specific function. This is a common pattern in Objective-C, but it can be confusing at first, so let me know if I can clarify.

The "bid button" class would more accurately be described as an "auction button" because it contains centralized logic for the CTA we show the user at any point in a sale. It could say "Bid" or "Register" or "Registration Pending" or "Registration Closed", etc. Complicating things, this button gets used in the following contexts, so any change we make needs to be backwards-compatible.

- AuctionViewController (at the top of the list of lots)
- ARArtworkActionsView (presented in an ArtworkView)
- AuctionInformationViewController (accessible via AuctionViewController as well as LAI)

I added a new snapshot test for the non-default intent of bidding, as well as updated the existing snapshot for tests that contain this button.

Continues work from https://github.com/artsy/eigen/pull/2659


sweir27:
Ah, cool!

PR #2673, opened 2018-07-18
Fixes bid status not updating

ashfurrow:
This is for [Purchase-290](
https://artsyproduct.atlassian.net/browse/PURCHASE-290) but it needs more testing before merging. 

Here's the cause of the bug:

- `ARArtworkView` fetches API results when new bids are placed.
- We were relying on `ARArtworkView` to notify subviews when there is new bid status (from API fetch).
- We expected the model in the nested `ARArtworkActionsView` to be updated in-place based on the fetch **but** they weren't.

The solution is to pass down the new sale artwork model as a parameter to the `updateUI` function.

### DIFF
--- a/Artsy/Views/Artwork/ARArtworkActionsView.h
+++ b/Artsy/Views/Artwork/ARArtworkActionsView.h
@@ -2,7 +2,7 @@
 
 #import <ORStackView/ORStackView.h>
 
-@class Artwork, ARArtworkActionsView;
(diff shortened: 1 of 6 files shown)


erikdstock:
Ah, this makes much more sense to me. Tested and working in the xcode simulator.

ashfurrow:
Thanks Erik! Appreciate the help testing – I've also verified that it works. Feel free to merge when ready 👍

PR #2676, opened 2018-07-20
Fixes outbid notification suppression.

ashfurrow:
I've got a more detailed description of the bug [here](
https://github.com/artsy/eigen/pull/2666/files#r204059063) but the gist is, since BidFlow is presented modally, we can't rely on a simple class check of the global top view controller. 

Testing push notifications is really tricky since our staging environment uses Apple's production servers and they'll only push to properly deployed builds of the app, so I've updated the tests and read the code until I'm confident it'll probably work. More eyes on this would be helpful to double-check. I'll definitely test on the next beta.

### DIFF
--- a/Artsy/App/ARAppNotificationsDelegate.m
+++ b/Artsy/App/ARAppNotificationsDelegate.m
@@ -10,6 +10,7 @@
 #import "ARNotificationView.h"
 #import "ARSwitchBoard.h"
 #import "ARTopMenuViewController.h"
+#import "ARSerifNavigationViewController.h"
 #import "ARLogger.h"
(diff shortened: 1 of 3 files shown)


l2succes:
👍  looks good. 

I was trying to track down the instructions I wrote up a while back on testing push notifications using a device id and pulse, if it's of any use to you let me know.

ashfurrow:
I might take you up on that – hopefully the beta works and I won't need them 😉 Eigen's existing docs are barebones and a bit outdated, any notes you could add would be helpful: https://github.com/artsy/eigen/blob/master/docs/push_notifications.md

PR #2681, opened 2018-08-27
Fixes bug where LAI registration would invoke BidFlow

ashfurrow:
Took a look with @erikdstock on this – just an oversight from the BidFlow project. It fixes this crash: https://sentry.io/artsynet/eigen/issues/626942969/?query=is:unresolved# And addresses [PURCHASE-411](https://artsyproduct.atlassian.net/browse/PURCHASE-411#).

### DIFF
--- a/Artsy_Tests/App_Tests/ARSwitchBoardTests.m
+++ b/Artsy_Tests/App_Tests/ARSwitchBoardTests.m
@@ -25,6 +25,7 @@
 
 #import <Emission/ARArtistComponentViewController.h>
 #import <Emission/ARFavoritesComponentViewController.h>
+#import <Emission/ARBidFlowViewController.h>
 
 #import "Artsy-Swift.h"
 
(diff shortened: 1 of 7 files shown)


Artsy_Tests/App_Tests/ARSwitchBoardTests.m
```diff
@@ -421,6 +422,24 @@ @interface ARTopMenuNavigationDataSource (Tests)
...
+            switchboard = [[ARSwitchBoard alloc] init];
+            [switchboard updateRoutes];
+
+            ARSerifNavigationViewController *subjectNav = (id)[switchboard loadPath:@"/auction-registration/the-sale"];
+            ARBidFlowViewController *subject = (id)subjectNav.topViewController;
+
+            expect(subject.saleID).to.equal(@"the-sale");
+        });
+
+        it(@"falls back to force bid registration when skip_bid_flow=true", ^{
```
sweir27: Question about this-- why wouldn't we want to just kick them to our new registration flow even if they're in a live sale? Theoretically that should accomplish the same thing as falling back to the force flow.

sweir27: Okay, that makes sense (those darn nested navigation controllers 😳). 

Thanks for creating the follow-up ticket! Since this fixes the crasher will :marge:.

PR #2685, opened 2018-09-14
Updates the mutation name to use the prefix

orta:
We want to have all stitched in mutations [prefixed with the domains](https://github.com/artsy/README/blob/master/best-practices/graphql-schema-design.md#schema-stitching) to handle the issue of what "if we need order in two contexts" in the future.

This got added with https://github.com/artsy/metaphysics/pull/1281 (check the `_schema.graphql`) but isn't yet shipped to prod.

### DIFF
--- a/Artsy/Networking/create_order.graphql
+++ b/Artsy/Networking/create_order.graphql
@@ -1,7 +1,7 @@
 mutation createOrder(
   $input: CreateOrderWithArtworkInput!
 ) {
-  createOrderWithArtwork(
+  ecommerceCreateOrderWithArtwork(
     input: $input
   ) {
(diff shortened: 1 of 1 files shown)


ashfurrow:
Merge on green.

orta:
Yes for staging

ashfurrow:
My bad, but we'll need to update the following code that extracts data from the response:

https://github.com/artsy/eigen/blob/91ba0618607768a3180e057bf853aa7f88064846/Artsy/View_Controllers/Artwork/ARArtworkViewController%2BButtonActions.m#L233

I'll send a follow-up PR.

PR #2686, opened 2018-09-14
Fixes mutation reference to include ecommerce prefix

ashfurrow:
Follow-up from #2685. Also adds a changelog entry for that PR.


ashfurrow:
Hmm weird, I wonder why the tests failed _now_ but not on Orta's original PR?

ashfurrow:
Oh I see, hmm. Our testing setup is a little brittle here. The Artwork view is next on the React Native rewrite list, so probably not worth it to invest into fixing it.

orta:
weirrrrdddd

orta:
could there be dummy data that needs the same change?

PR #2689, opened 2018-09-17
Fixes a problem with order creation input #trivial

ashfurrow:
Fixes [PURCHASE-463](https://artsyproduct.atlassian.net/browse/PURCHASE-463), which was caused by sending an artwork slug instead of its proper ID. Something something string data types 😉


ashfurrow:
Gonna self-merge so we can submit a beta.

Issue #2691, opened 2018-09-20
Re-auth when you get a 401


orta:
metaphysics can send it in the `errors`

orta:
This is in

PR #2699, opened 2018-10-09
Re-authenticate if we have the username and password in keychain and get a 401 on launch

orta:
This fixes a long-running problem where the auth token is a 401 from gravity - 
https://github.com/artsy/eigen/issues/2691

Will explain how it works inline


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Artsy/App/ARAuthValidator.h
```diff
@@ -0,0 +1,19 @@
...
+ * process with stored creds and re-creates the Emission object
+ * with the new access token.
+ */
+@interface ARAuthValidator : NSObject
+
+/// Check for the validity of the current auth token
++ ( void)validateAuthCredentialsAreCorrect;
+
+@end
+
```
orta: I added my comments inline instead here ^

Artsy/App/ARAuthValidator.m
```diff
@@ -0,0 +1,25 @@
...
+        NSString *serverTDLR = [error userInfo][@"NSLocalizedDescription"];
+        if ([serverTDLR containsString:@"401"]){
+            // We hit a 401 so we need to be a bit worried about this
+            ARUserManager *manager = [ARUserManager sharedManager];
+            [manager tryReLoginWithKeychainCredentials:^(User *currentUser) {
+                [[ARAppDelegate sharedInstance] setupEmission];
+            }];
+        }
+    }];
+}
```
orta: We don't get a HTTP status code from the NSError, (it's some weird foundation number) so instead grab the error description coming from AFNetworking and check for a 401 in there.

Artsy/Constants/ARDefaults.m
```diff
@@ -9,6 +9,8 @@
 NSString *const ARStagingMetaphysicsURLDefault = @"ARStagingMetaphysicsURLDefault";
 NSString *const ARStagingLiveAuctionSocketURLDefault = @"ARStagingLiveAuctionSocketURLDefault";
 
+NSString *const ARUsernameKeychainKey = @"ARUsernameKeychainKey";
+NSString *const ARPasswordKeychainKey = @"ARPasswordKeychainKey";
 NSString *const AROAuthTokenDefault = @"AROAuthToken";
```
orta: We weren't storing username/password at all, except in safari keychain, but that's a special case

Artsy/Networking/API_Modules/ARUserManager.m
```diff
@@ -179,6 +179,7 @@ - (void)loginWithUsername:(NSString *)username
         NSString *expiryDateString = JSON[AROExpiryDateKey];
 
         [ARRouter setAuthToken:token];
+        [self storeUsername:username password:password];
```
orta: Makes sure we save on login, this won't work for facebook auth'd folks

Artsy/Networking/API_Modules/ARUserManager.m
```diff
@@ -286,8 +286,8 @@ - (void)loginWithFacebookToken:(NSString *)token
                 networkFailure(error);
             }
         }
+    }];
 
-        }];
     [op start];
 }
```
orta: formatting ^

Artsy/Networking/API_Modules/ARUserManager.m
```diff
@@ -442,6 +442,8 @@ + (void)clearUserData:(ARUserManager *)manager useStaging:(id)useStaging
 
     [manager.keychain removeKeychainStringForKey:AROAuthTokenDefault];
     [manager.keychain removeKeychainStringForKey:ARXAppTokenKeychainKey];
+    [manager.keychain removeKeychainStringForKey:ARUsernameKeychainKey];
+    [manager.keychain removeKeychainStringForKey:ARPasswordKeychainKey];
```
orta: making sure logout works

Artsy/Networking/API_Modules/ARUserManager.m
```diff
@@ -565,6 +567,33 @@ - (void)saveSharedWebCredentialsWithEmail:(NSString *)email
...
+        // NOOP
+    }
+    gotUser:success
+    authenticationFailure:^(NSError *error) {
+        // NOOP
+    } networkFailure:^(NSError *error) {
+        // NOOP
+    }];
+}
+
```
orta: Mostly `NOOP`s but it'll accurately emulate going through the login button process (which will set up the ARRouter etc)

Artsy/Networking/API_Modules/ArtsyAPI+CurrentUserFunctions.m
```diff
@@ -8,6 +8,16 @@
...
 
++ (void)getMeHEADWithSuccess:(void (^)(void))success failure:(void (^)(NSError *error))failure;
+{
+    NSURLRequest *request = [ARRouter newMeHEADRequest];
+    [self performRequest:request success:^(id thing) {
+        success();
+    } failure:^(NSURLRequest *request, NSHTTPURLResponse *response, NSError *error) {
+        failure(error);
+    }];
+}
```
orta: Just exposing a way to get the HEAD for the me route

Artsy/App/ARAuthValidator.m
alloy: Could it be a wrapped error? (Check `userInfo[NSUnderlyingError]`)

alloy:
Yo, thanks for attempting to deal with this!

The code looks good, although it would be nice to get rid of the string matching of the HTTP status.

However, I’m wondering if it wouldn’t make more sense to have a user re-sign in from the splash screen instead. The reason I say this is because I assume the issue can come up just as well with FB authentication, web credentials, or when a user has changed their password on artsy.net. I think that in all these cases the solution wouldn’t suffice, whereas throwing them back to the splash screen, while slightly less smooth of an experience, would address all of these. Do you have thoughts on this?

orta:
I think I know a reasonable answer, which is to log you out if the re-auth fails or we really can't log you in because of no username/password. 

I've made some changes to do that - it's not elegant because it's an alert and a log out, but the onboarding is meant to be set up at the start and not thrown up async.

alloy:
Merge on green

PR #2700, opened 2018-10-09
Buy Now presents in a modal

ashfurrow:
This PR addresses [PURCHASE-595](https://artsyproduct.atlassian.net/browse/PURCHASE-545), to present the Buy Now flow in a modal instead of as part of Eigen's navigation stack. This leads to problems because the Buy Now flow maintains its own navigation state, and the user might hit the "back" button expecting to back to an earlier Buy Now step, but instead their order is discarded and they are sent back to the artwork view.

This also involved fixing some Swift IUO-inference faff, as well as fixes a bug that [I introduced](https://github.com/artsy/eigen/commit/556935c53e351799c1acffb4d435d26e46bbd456#diff-15d035ea7cd028d3e2dfc3f84333b785R102) when removing pre-iOS 7's status bar API use (we were always returning `UIStatusBarStyleLightContent` instead of `self.statusBar`, which was also a infinite recursive loop – the bug is split across [two commits](https://github.com/artsy/eigen/commit/5b5c7cc98bcf3c5136687cc2212af35f5f9c58f4#diff-15d035ea7cd028d3e2dfc3f84333b785R95)). Now our web views will show whatever status bar content they are specified to, with a default of `.Default`.

Also has work for [PURCHASE-392](
https://artsyproduct.atlassian.net/browse/PURCHASE-392), to update echo config.


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Artsy/View_Controllers/Artwork/ARArtworkViewController+ButtonActions.m
```diff
@@ -238,11 +239,12 @@ - (void)tappedBuyButton
             NSString *path = self.echo.routes[@"ARBuyNowRoute"].path;
             if (!path) {
                 // path should never be nil, but I'd rather not crash the app if it is.
-                path = @"/order/:id";
+                path = @"/orders/:id";
```
orta: this feels incompatible with how rails normally does rest-y routes

ashfurrow: Yeah, I agree. This was a change in Force, maybe @sweir27 has more context?

sweir27: Yep! Since `/order` is an existing route, in order to not have overlap and ensure a smooth transition we're using `/orders` for now.

We'll likely want to update to `/order` in the future (post-launch/deprecation), so if you want to leave this how it is then it's probably fine.

orta:
OK, cool, yep - merge on green

PR #2701, opened 2018-10-10
Adds support for showing multiple artists in live lots

orta:
The MVVM model, Mantle's keypathing and our approach to GraphQL's flexibility really paid off well for this PR

screen shot 2018-10-10 at 1 15 37 pm

screen shot 2018-10-10 at 1 15 33 pm

### DIFF
--- a/Artsy/Models/API_Models/LiveAuctionLot.m
+++ b/Artsy/Models/API_Models/LiveAuctionLot.m
@@ -33,7 +33,7 @@
         ar_keypath(LiveAuctionLot.new, artworkTitle) : @"artwork.title",
         ar_keypath(LiveAuctionLot.new, artworkDate) : @"artwork.date",
         ar_keypath(LiveAuctionLot.new, artwork) : @"artwork",
-        ar_keypath(LiveAuctionLot.new, artistName) : @"artwork.artist.name",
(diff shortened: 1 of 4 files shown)


Artsy/Models/API_Models/LiveAuctionLot.m
```diff
@@ -33,7 +33,7 @@ + (NSDictionary *)JSONKeyPathsByPropertyKey
         ar_keypath(LiveAuctionLot.new, artworkTitle) : @"artwork.title",
         ar_keypath(LiveAuctionLot.new, artworkDate) : @"artwork.date",
         ar_keypath(LiveAuctionLot.new, artwork) : @"artwork",
-        ar_keypath(LiveAuctionLot.new, artistName) : @"artwork.artist.name",
+        ar_keypath(LiveAuctionLot.new, artistName) : @"artwork.artist_names",
```
l2succes: Nice 👍is that a preprocessor macro or a helper c function?

orta: It's a macro, it compiles down to just an @"string" of the final key path and never executes the `.new` on that class.

```
#define ar_keypath(anObject, aKeyPath) ((void)(NO && ((void)anObject.aKeyPath, NO)), @ #aKeyPath)
```

PR #2705, opened 2018-10-11
Handle links that come through the FB app in an unusual format

orta:
OK, so I went through getting started docs and I'm alright with saying this is just how they send URLs to us occasionally. I verified that it doesn't break FB auth in the app.


l2succes:
Yup! Here's the JIRA ticket https://artsyproduct.atlassian.net/browse/GROW-838

l2succes:
merge on green

PR #2706, opened 2018-10-11
Migrates Adjust to use universal links instead of deep links

orta:
This is the new recommendation from Adjust, so it seems like a good time to give it a shot. I've verified that links like https://kqwq.adj.st/?adjust_t=esa8xn would correctly open Eigen and seen in the logs that Adjust has reported the link.


l2succes:
Amazing! Thanks @orta

PR #2709, opened 2018-10-11
Re-enables the safari web keychain, and only shows it on the 3rd launch

orta:
Now that we store the email/password we can try save the creds at a later stage than just onboarding. This adds it on the third run.

### DIFF
--- a/Artsy/App/ARAppDelegate.m
+++ b/Artsy/App/ARAppDelegate.m
@@ -346,7 +346,7 @@ static ARAppDelegate *_sharedInstance = nil;
   sourceApplication:(NSString *)sourceApplication
          annotation:(id)annotation
 {
-    return [self application:application openURL:url options:@{
+    return [self application:application openURL:url options: @{
         UIApplicationOpenURLOptionsSourceApplicationKey: sourceApplication ?: @"",
         UIApplicationOpenURLOptionsAnnotationKey: annotation  ?: @""
     }];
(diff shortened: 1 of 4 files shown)


Artsy/App/ARAppDelegate.m
```diff
@@ -471,6 +471,10 @@ - (void)countNumberOfRuns
         [Adjust trackEvent:[ADJEvent eventWithEventToken:ARAdjustFirstUserInstall]];
     }
 
+    if (numberOfRuns == 3) {
+        [[ARUserManager sharedManager] tryStoreSavedCredentialsToWebKeychain];
+    }
```
alloy: Why specifically on the third run?

orta: Just felt like a reasonable enough number, the first 2 times we show the push notifications alert.

alloy: I guess it could just feel a bit weird for an app to suddenly (not fresh out of onboarding) ask you stuff about creds

alloy:
LGTM 👍 

Merge on green

CHANGELOG.yml
```diff
@@ -19,6 +19,8 @@ upcoming:
     - Fixes to loading Facebook links coming from the FB app - orta
     - If you wait on an Artwork zoom view, the back button is hidden - orta
     - When zooming into an artwork you can wait 2 seconds and the back button will disappear - orta
+    - Adds a 5s timeout to the shared web credentials - orta
```
alloy: Oops, this isn’t in yet :D

Artsy/App/ARAppDelegate.m
orta: I think that's reasonable - yep, lets see what the message says and maybe we can prime people with a warning alert, it's cheap to do that

PR #2714, opened 2018-10-16
Fixes crash in Fair analytics

ashfurrow:
I noticed the crash [here](https://sentry.io/artsynet/eigen/issues/294042266/?query=is:unresolved), caused by creating a dictionary with a `nil` value entry. It's a one-line fix so I just did it. WIP because we're currently in-review and I don't want to changelog-faff right now.


ashfurrow:
Looking for the string `id,` in the file showed a bunch of others, so I fixed those too.

orta:
👍  merge on green

PR #2688, opened 2018-09-14
Migrate the main Artwork information API to use MP instead of gravity

orta:
The aim being to almost recreate the `v1/artwork/:id` api in metaphysics, *then* we can use some of the new metaphysics fields we need for BNMO.

Also includes #2687 which will result in a smaller PR when merged.

#known for the schema changes from Xcode updates

PR #2716, opened 2018-10-18
Fixes problem with attribution classes from Gravity

ashfurrow:
This PR is to remedy [an ongoing incident](https://trello.com/c/UTDSwXwm/181-ios-app-not-showing-lots-in-auction-views).

[Recent changes](https://github.com/artsy/eigen/pull/2688) to the artwork modelling were deployed today, causing a long series of events that led users to see the auction view with few or zero lots. Only artworks with no attribution class were visible.

tl;dr

- We changed the JSON Mantle mapping for `attributionClass` to be `attributionClass.name`, since it's nested in Metaphysics.
- However, Gravity models don't return that nesting. When Mantle calls `valueForKeyPath` with `attribution_class.name` on Gravity models, it has an `NSString` for the `attribution_class` key path, but string is not key value coding compliant for the `name` key, so [this](https://github.com/Mantle/Mantle/blob/e41221ee948e00707bd5651465e1c0d1c3cb2b1a/Mantle/MTLJSONAdapter.m#L166-L180) try/catch re-throws and is later [silently ignored](https://github.com/Mantle/Mantle/blob/e41221ee948e00707bd5651465e1c0d1c3cb2b1a/Mantle/MTLJSONAdapter.m#L211).

The solution was to extend the existing pattern for handling collisions. I had to prefix `attribution_class` with `mp_` since the type are different.

### DIFF
--- a/Artsy/Networking/artwork.graphql
+++ b/Artsy/Networking/artwork.graphql
@@ -46,7 +46,7 @@ query OldArtworkQuery($artworkID: String!) {
       in
     }
 
-    attribution_class {
+    mp_attribution_class {
       name
     }
     
--- a/Artsy/Networking/artwork.graphql
+++ b/Artsy/Networking/artwork.graphql
@@ -2,7 +2,7 @@ query OldArtworkQuery($artworkID: String!) {
   artwork(id: $artworkID) {
     id
     _id
- 
+
     artist {
       id
       _id
@@ -46,10 +46,10 @@ query OldArtworkQuery($artworkID: String!) {
       in
     }
 
-    mp_attribution_class {
+    mp_attribution_class: attribution_class {
       name
     }
-    
+
     edition_sets {
       id
       sale_message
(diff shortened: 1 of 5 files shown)


ashfurrow:
CI passed here, but Circle's being weird about reporting it: https://circleci.com/gh/artsy/eigen/5611

Artsy/Networking/artwork.graphql
```diff
@@ -46,7 +46,7 @@ query OldArtworkQuery($artworkID: String!) {
       in
     }
 
-    attribution_class {
+    mp_attribution_class {
```
alloy: I don’t understand this change. Has MP’s schema changed to add this field? Or was the intention to alias the field?

alloy:
Pushed a change that I have tested in GraphiQL, but not yet inside the app as I didn’t have the time to get my local env up-to-date. I’ve also ran `make deploy` again.

alloy:
i.e. please test the change in the app before releasing!!

ashfurrow:
It’s weird how as soon as I saw your commit message, I knew exactly what was wrong 😅 I guess this is why we get people to review our sleepy code! 💤

alloy:
Yup! 😄

ashfurrow:
Okay, found one more slight problem that's keeping attribution classes from being displayed on the artwork view. I don't think I want to give up our place in queue for it, but will PR a fix separately.

PR #2717, opened 2018-10-18
Makes attribution classes case-insensitive

ashfurrow:
This fixes a bug mentioned here:
https://github.com/artsy/eigen/pull/2716#issuecomment-431046347 It prevents the artwork view from displaying the attribution class information for many (all?) artworks, whose `attribution_class` has a capital letter at the beginning.

### DIFF
--- a/Artsy/Models/API_Models/Partner_Metadata/Artwork+AttributionClassStrings.m
+++ b/Artsy/Models/API_Models/Partner_Metadata/Artwork+AttributionClassStrings.m
@@ -6,25 +6,25 @@
 
 - (NSString *)shortDescriptionForAttributionClass
 {
-    if ([self.attributionClass isEqualToString:@"unique"]) {
+    if ([self.attributionClass.lowercaseString isEqualToString:@"unique"]) {
         return @"This is a unique work.";
 
-    } else if ([self.attributionClass isEqualToString:@"limited edition"]) {
+    } else if ([self.attributionClass.lowercaseString isEqualToString:@"limited edition"]) {
         return @"This is part of a limited edition set.";
 
-    } else if ([self.attributionClass isEqualToString:@"made-to-order"]) {
+    } else if ([self.attributionClass.lowercaseString isEqualToString:@"made-to-order"]) {
         return @"This is a made-to-order piece.";
 
-    } else if ([self.attributionClass isEqualToString:@"reproduction"]) {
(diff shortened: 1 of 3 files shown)


orta:
Weird, I wouldn't have expected this change to happen?

orta:
merge on green

ashfurrow:
Ah, that's why:

screen shot 2018-10-18 at 11 23 24 am

Artsy/Models/API_Models/Partner_Metadata/Artwork+AttributionClassStrings.m
```diff
@@ -6,25 +6,25 @@ @implementation Artwork (AttributionClassStrings)
...
-    if ([self.attributionClass isEqualToString:@"unique"]) {
+    if ([self.attributionClass.lowercaseString isEqualToString:@"unique"]) {
         return @"This is a unique work.";
 
-    } else if ([self.attributionClass isEqualToString:@"limited edition"]) {
+    } else if ([self.attributionClass.lowercaseString isEqualToString:@"limited edition"]) {
         return @"This is part of a limited edition set.";
 
-    } else if ([self.attributionClass isEqualToString:@"made-to-order"]) {
+    } else if ([self.attributionClass.lowercaseString isEqualToString:@"made-to-order"]) {
```
alloy: Does this match the `name` field value? Asking because the hyphens look very much like a `id` field value

ashfurrow: I will check immediately.

ashfurrow: CMS has the hyphens, I'm guessing we're okay but still checking:

screen shot 2018-10-18 at 11 46 16 am

ashfurrow: Yup, we're okay 👍

alloy: Ok, and not to be pedantic, but did you check the other hyphenated values too? Although I guess it looks like we can always fix it in CMS?

ashfurrow: I did check the other hyphenated value, `Non-editioned multiple`. Being pedantic is very acceptable in this case 👍

PR #2721, opened 2018-10-19
Updates Emission and prepares for the next version

orta:
See title


orta:
gonna also ship a beta of this, but it's good to go

ashfurrow:
@orta sounds good. Can you please pull in the commit on #2718?

ashfurrow:
Oh, we've already submitted a beta. I know this is annoying but I'd like to submit another one with that fix in – artwork blurbs aren't visible right now in the app.

PR #2722, opened 2018-10-22
Updates deploy docs

ashfurrow:
This updates the deploy docs in a few important ways:

- Splits our pre-deploy QA into one of three groups:
  - Tests that need to be run on staging.
  - Tests that need to be run on production.
  - Tests that currently don't work (with Jira links).
- I added a test that would have caught [Eigen's incident last week](https://trello.com/c/UTDSwXwm/181-ios-app-not-showing-lots-in-auction-views)
- I added notes for a recent problem with view-in-room as well.
- Adds note about our use of IDFA.

I also updated the Markdown headers so the page has an `H1`.


alloy:
Awesome 👌

PR #2728, opened 2018-11-09
Removes the position field from a LiveAuctionLot 

orta:
As it's not used, and caused a problem. Position is used in MP/ Grab to order, and we don't do custom ordering inline in the app.


orta:
Sibling to 
https://github.com/artsy/metaphysics/pull/1377

PR #2732, opened 2018-11-21
Add switchboard routing for ARFairSearchViewController

javamonn:
As part of [LD-28](https://artsyproduct.atlassian.net/browse/LD-28), we're adding a "Find Exhibitors and Artists" link button to the new Fair view that links to a fair-scoped search. This mirrors functionality that exists in Eigen today as part of `ARFairSearchViewController`. This PR aims to expose this screen via SwitchBoard.

PR #2733, opened 2018-11-21
Bumps everything to 4.3.4 #trivial

ashfurrow:
Now that we've released 4.3.3.

I have also removed some extraneous spacing (so that now, _versions_ are divided by newlines but sections of the notes, like `dev` and `user_facing` aren't separated). I also added a `signed_of_by` field to the release to note who did the final QA and release to the App Store.


ashfurrow:
Gonna self-merge to unblock #2732.

PR #2734, opened 2018-11-27
Make Offer

ashfurrow:
This PR addresses [PURCHASE-582](https://artsyproduct.atlassian.net/browse/PURCHASE-582), which adds the Make Offer flow to Eigen. The PR has the following technical details:

- Introduces new `ExchangeCurrentVersion` [Echo](https://github.com/artsy/echo) flag to disable BNMO remotely (in the case of breaking Exchange changes).
- Introduces new `AREnableMakeOfferFlow` feature to Echo to remote-enable Make Offer for users.
- Adds admin option to enable Make Offer (see screenshot below).
- Adds Make Offer button to `offerable`/`is_offerable` artworks, which will call the `ecommerceCreateOfferOrderWithArtwork` mutation on Exchange-via-Metaphysics, and will redirect to `/order/:order_id` on success.

screen shot 2018-11-27 at 2 36 22 pm

screen shot 2018-11-27 at 2 47 25 pm

screen shot 2018-11-27 at 2 38 27 pm


ashfurrow:
Looks like the vertical margin might be too large between the two buttons, I'll check with Brian.

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

ashfurrow:
Okay, fixed that in 
https://github.com/artsy/eigen/pull/2734/commits/b80053f9de6a8bf618a173109d7812076ef40821

Artsy/Networking/ARRouter.m
```diff
@@ -1246,7 +1246,12 @@ + (NSURLRequest *)newSystemTimeRequest
 
 + (NSURLRequest *)newBuyNowRequestWithArtworkID:(NSString *)artworkID
 {
-    return [self graphQLRequestForQuery:[self graphQueryToCreateBuyNowOrderForArtwork:artworkID] variables:@{ @"input" : @{ @"artworkId": artworkID, @"quantity": @(1) } }];
+    return [self graphQLRequestForQuery:[self graphQueryToCreateBuyNowOrder] variables:@{ @"input" : @{ @"artworkId": artworkID, @"quantity": @(1) } }];
+}
+
++ (NSURLRequest *)newOfferRequestWithArtworkID:(NSString *)artworkID
+{
+    return [self graphQLRequestForQuery:[self graphQueryToCreateOffer] variables:@{ @"artworkId": artworkID, @"quantity": @(1) }];
```
l2succes: Does the shape of the NSDictionary here depend on the mutation input type?

Because its different from `[self graphQueryToCreateBuyNowOrder]`

Artsy/View_Controllers/Artwork/ARArtworkViewController+ButtonActions.m
```diff
@@ -220,35 +221,48 @@ - (void)tappedBuyersPremium:(UIButton *)button
     }];
 }
 
+- (void)tappedMakeOfferButton
+{
+    [ArtsyAPI createOfferOrderWithArtworkID:self.artwork.artworkUUID success:^(id results) {
```
l2succes: Are you using `artworkUUID` instead of `artworkID` because MP needs the globally unique entity ID for this mutation or there's another reason?

Artsy/View_Controllers/Artwork/ARArtworkViewController+ButtonActions.m
```diff
@@ -220,35 +221,48 @@ - (void)tappedBuyersPremium:(UIButton *)button
...
 
+- (void)handleOrderCreationOrderOrError:(id)orderOrError
+{
+    NSString *orderID = orderOrError[@"orderOrError"][@"order"][@"id"];
+    if (!orderID) {
+        [self presentErrorMessage:@"Something went wrong. Please try again or contact support@artsy.net."];
+        return;
+    }
+    NSString *path = self.echo.routes[@"ARBuyNowRoute"].path;
+    if (!path) {
```
l2succes: I've had `echo.routes` return an empty array before. Cause the key was invalid but still 👍

Artsy/Networking/ARRouter.m
ashfurrow: It does, I noticed that initially (in `graphQueryToCreateBuyNowOrder`) we wrapped everything into an `input` object, but it seems like we aren't doing that anymore in Insomnia, etc, so I removed them. Either works.

Artsy/View_Controllers/Artwork/ARArtworkViewController+ButtonActions.m
ashfurrow: Yeah exactly, `artworkID` in Eigen is more accurately the artwork _slug_.

Artsy/Views/Artwork/ARArtworkActionsView.h
```diff
@@ -16,18 +16,20 @@
...
 - (void)tappedBuyButton;
+- (void)tappedMakeOfferButton;
 - (void)tappedMoreInfo;
 @end
 
 
 @interface ARArtworkActionsView : ORStackView
 
-- (instancetype)initWithArtwork:(Artwork *)artwork;
+- (instancetype)initWithArtwork:(Artwork *)artwork echo:(ArtsyEcho *)echo;
```
l2succes: Couldn't we just add `[ArtsyEcho sharedInstance]` instead of passing it down to Views in this case?
Since it should get instantiated / changed once per session.

ashfurrow: Yeah totally. I just don't want to use a singleton, and injecting via an initializer was an easy choice. Thoughts @orta ?

l2succes: Yeah for sure, I totally don't like overusing singletons either but in this case I think it makes sense

ashfurrow: It would be a nontrivial refactor, since there are lots of places where view controllers/etc are creating their own `ArtsyEcho` instances. Updating those classes isn't difficult, but updating all their tests _is_.

ashfurrow:
I've just submitted a beta with these changes in it, for tomorrow's cross-team demo.

Artsy/Views/Artwork/ARArtworkActionsView.h
l2succes: Totally, we could just make it optional to use the singleton like `[ArtsyEcho defaultInstance]` which is a common pattern. In case you don't need/want to create your own instance. 

Then again not sure it's worth it considering how often we touch eigen code.

Artsy/App/Echo.json
```diff
@@ -1 +1 @@
...
...,"value":true},{"name":"ARDisableReactNativeBidFlow","value":false},{"name":"AREnableBuyNowFlow","value":true},{"name":"AREnableMakeOfferFlow","value":false}],"messages":[{"name":"LiveAuctionsCurrentWebSocketVersion","content":"3"},{"name":"ARVIRVideo","content":"https://files.artsy.net/videos/artsy-eigen-arvir-iphone.mp4"},{"name":"ExchangeCurrentVersion","content":"1"}],"routes":[{"name":"ARArtistRoute","path":"/artist/:id"},{"name":"ARProfileArtistRoute","path":"/:profile_id/artist/:id"},{"name":"ARArtworkRoute","path":"/artwork/:id"},{"name":"ARGeneRoute","path":"/gene/:id"},{"name":"ARShowRoute","path":"/show/:id"},{"name":"ARAuctionRegistrationRoute","path":"/auction-registration/:id"},{"name":"ARAuctionRoute","path":"/auction/:id"},{"name":"ARAuctionBidArtworkRoute","path":"/auction/:id/bid/:artwork_id"},{"name":"ARFairProfileForYouRoute","path":"/:profile_id/for-you"},{"name":"ARFairBrowseArtistRoute","path":"/:profile_id/browse/artist/:id"},{"name":"ARBrowseCategoriesRoute","path":"/categories"},{"name":"ARLiveAuctionsURLDomain","path":"live.artsy.net"},{"name":"ARLiveAuctionsStagingURLDomain","path":"live-staging.artsy.net"},{"name":"ARBuyNowRoute","path":"/orders/:id"}]}
```
ashkan18: really cool to have buy now route in echo!

Artsy/Models/API_Models/Partner_Metadata/Artwork.m
```diff
@@ -279,6 +284,11 @@ - (NSNumber *)isAcquireable
     return self.gravIsAcquirable.boolValue ? self.gravIsAcquirable : self.mpIsAcquirable;
 }
 
+- (NSNumber *)isOfferable
+{
+    return self.gravIsOfferable.boolValue ? self.gravIsOfferable : self.mpIsOfferable;
```
ashkan18: is this because this method can get gravity data or MP data?

Artsy/Networking/ARRouter.m
```diff
@@ -1246,7 +1246,12 @@ + (NSURLRequest *)newSystemTimeRequest
 
 + (NSURLRequest *)newBuyNowRequestWithArtworkID:(NSString *)artworkID
 {
-    return [self graphQLRequestForQuery:[self graphQueryToCreateBuyNowOrderForArtwork:artworkID] variables:@{ @"input" : @{ @"artworkId": artworkID, @"quantity": @(1) } }];
+    return [self graphQLRequestForQuery:[self graphQueryToCreateBuyNowOrder] variables:@{ @"input" : @{ @"artworkId": artworkID, @"quantity": @(1) } }];
```
ashkan18: 👍 in sending `quantity`, its _optional_ but i think its a good reminder to have it here for later on support :)

Artsy/Networking/ARRouter.m
ashkan18: yes, we can either make the whole input a _variable_ or we can have fields inside `input` to be specific variables, both should work, but I feel the second one being more explicit.

Artsy/Models/API_Models/Partner_Metadata/Artwork.m
ashfurrow: That's right. Realistically, we'll only get this from MP, but this is just to make sure other parts of the app that still use Gravity can rely on this.

Artsy/Views/Artwork/ARArtworkActionsView.h
orta: I wonder if it might be a good case for Echo having a [config class](http://artsy.github.io/blog/2015/09/01/Cocoa-Architecture-Dropped-Design-Patterns/#Configuration.Classes.over.Inter-Class.Communication) which could also exist in Emission (which could sorta replace the weird dict object representation today) which could be the singleton? 

This *should* be an edge case overall, as I'd much rather we can move that kind of stuff into the metaphysics query for the related view ( that was the aim of [Mondrian](https://github.com/orta/mondrian) ) so then the features data lives with the query for the data (removing this dance at all)

That said, in this case, I'd prefer that we DI this kind of stuff in because of testing simplicity, it's just real faff trying to write native tests so making your life easier there can be more valuable than elegant code :D

ashfurrow: Okay, cool. I agree, I'm a fan of DI but I think in the long-term we might want to set up Emission to accept in an `ArtsyEcho`-shaped object for its own config (with `ArtsyKeys`, too). With that said, I'm going to merge because 🤞 all this code will be moved into Emission soon anyway.

PR #2735, opened 2018-11-30
Fix Contact/Make offer bug, and a few others

ashfurrow:
The other two bugs are so small that they don't really warrant individual PRs, so I rolled them into this one.

## The Main Bug fix

This PR is a follow-up to #2734 to address [PURCHASE-582](https://artsyproduct.atlassian.net/browse/PURCHASE-582). There was a bug in the original implementation, which wasn't producible due to a limitation in our backend. The bug is: when an artwork is contactable for price, _and_ offerable, but _not_ Buy Now-able, we should only show the Make offer button. I fixed this, simplified some boolean logic, and added a new UI test case.

## Other Bugs

One issue was all the extraneous Xcode logs, which is fixed with the `OS_ACTIVITY_MODE` environment variable:

screen shot 2018-11-30 at 12 26 56 pm

The other issue is user-facing. The scenario is:

- user signs in after installing the app
- user accepts our prompt for push notifications
- user then _rejects_ the OS prompt for push notifications
- user leaves, comes back a time later

On app launch, we ask the user to enable push notifications in the Settings app. That's cool, lots of apps do that. But we _weren't_ giving them an option to say "no", _forcing_ them to leave our app and go to Settings. So I added a cancel button:

| Before  | After  |
|---|---|
| before  | after  |

#known


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Errors
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Developer Specific file shouldn’t be changed, you can skip by adding #known to the PR body</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">CHANGELOG isn’t valid YAML</td>
    </tr>
  </tbody>
</table>

</p>

Artsy/App/ARAppNotificationsDelegate.m
```diff
@@ -70,6 +70,8 @@ - (void)displayPushNotificationSettingsPrompt
                                                               [self presentSettings];
                                                            }];
     [alert addAction:settingsAction];
+    alert.preferredAction = settingsAction;
+    [alert addAction:[UIAlertAction actionWithTitle:@"No thanks" style:UIAlertActionStyleCancel handler:nil]];
```
jonallured: Just to confirm: passing `nil` here would dismiss the alert, right?

CHANGELOG.yml
```diff
@@ -6,8 +6,11 @@ upcoming:
     - Adds AREnableMakeOfferFlow flag to Echo - ash&dlevenson
     - Adds ExchangeCurrentVersion to gate BNMO - ash
     - Add SwitchBoard path for ARFairSearchViewController - javamonn
+    - Removes extraneous OS debugging logs from Xcode - ash
   user_facing:
     - Make Offer now accessible - ash
+    - "Contact Gallery" button doens't show up with "Make offer" button anymore - ash
```
jonallured:
```suggestion
    - Contact Gallery button doesn't show up with Make Offer button anymore - ash
```

Maybe the quote marks were blowing things up?

orta:
Cool, looks good then 👍

PR #2754, opened 2019-01-08
Fix build by using UILabel directly

javamonn:
Fixes a compilation error causing a failing build introduced in: #2752 

Instead of adding a special-cased uncapitalized sans-serif font to `Artsy+UILabels`, we use `UILabel` directly to the same result.

screen shot 2019-01-08 at 4 47 38 pm


javamonn:
Hmm, this fixes the compilation error but causes a separate issue with the auction padding indicator I think:

screen shot 2019-01-08 at 4 52 42 pm

javamonn:
Fixed above by setting the label background color, similar to the `setup` done in `Artsy+UILabels`.

alloy:
Can you add some context about the build failure?

The code change itself looks good to me, it doesn't seem like that specialised label was doing much of anything 👍

alloy:
Ah, I bet it passed CI because you’re PR lives on a fork and we’re not letting CI access secret env vars to PRs from dorks.

PR #2761, opened 2019-01-17
Bump version

l2succes:
4.3.5 -> 4.3.6


CHANGELOG.yml
```diff
@@ -1,21 +1,29 @@
...
-    - Skips running tests on beta deploys - ash
-    - Updates Related Artworks grid item metadata styling - javamonn
-    
-  user_facing:
-    - Uses new search API for the app search - orta
-    - AR feature allows tapping on the done state to hide the UI - orta
-    - Fix for favourites pagination via Emission update - orta
-    - Removes the buy now promo video from the app - orta
-    - Adds Local Discovery tab behind lab option - luc
+  - version: 4.3.6
```
ashfurrow: CI might complain because `upcoming` is an array instead of a dictionary.

l2succes: Ah ok! Let me fix that

ashfurrow:
The CI failure was Error 65, which we periodically get. I've restarted the build.

orta:
Just pushed in the changes from #2762

orta:
I've also moved this to point to master, so that we don't forget these commits exist after deploying

orta:
Worth noting: 

> [!] The platform of the target `Artsy` (iOS 8.0) may not be compatible with `Pulley (2.6.2)` which has a minimum requirement of iOS 9.0.

ashfurrow:
Hmm, yeah. Pulley probably would work fine and they just used iOS 9 as a minimum because it's a pretty reasonable minimum at this point. I think it'll be fine.

l2succes:
Yeah I saw that too but wasn't sure if we should change the minimum version

orta:
It might be using iOS 9 APIs which would crash at runtime, I'll install an iOS 8 sim and check

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  danger-swift` failed.
  </td>
    </tr>
  </tbody>
</table>



### Log


```sh
[31mERROR: Failed to parse JSON: valueNotFound(Swift.String, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "danger", intValue: nil), CodingKeys(stringValue: "git", intValue: nil), CodingKeys(stringValue: "modified_files", intValue: nil), _JSONKey(stringValue: "Index 3", intValue: 3)], debugDescription: "Expected String but found null value instead.", underlyingError: nil))
[0;0m[31mERROR: Dangerfile eval failed at Dangerfile.swift
[0;0m[31mERROR: Could not get the results JSON file at /var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/danger-response.json
[0;0m
```

<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against fb7c29a9521b79b9f9813b26103b660dbb1f33ba
</p>

orta:
hah, I love how everything just broke all at once

On the other hand Xcode passes, just adding a last feature now

orta:
Our Podfile is wrong, we only support iOS 9 in the app.

orta:
OK, I'm gonna deploy this

l2succes:
@orta How do we get this merged in? We need to fix that Peril/Danger error first right?

PR #2769, opened 2019-02-15
Fixes some fair routing issues

ashfurrow:
The PR fixes [LD-199](https://artsyproduct.atlassian.net/browse/LD-199) by wrapping all not-already-wrapping view controllers in navigation controllers (providing them with a close button, so the user doesn't become trapped in a modal). I still need to test that this works with the following invocations routed from Emission (see the comment in `ARAppDelegate+Emission.m`):

- [x] Bid flow's live bidding exception case
- [x] Bid flow's conditions of sale 
- [x] Bid flow's Registration's conditions of sale
- [x] Search
- [x] Conversation's opening invoices
- [x] Generic markdown links
- [x] Fair's "view site" and "buy tickets" links

This PR also fixes a problem where Eigen would route `/conditions-of-sale` to the new fair component (if enabled, so this bug only presents for our beta testers) with a `fairID` set to `conditions-of-sale`. I described the problem in more detail [in Slack](https://artsy.slack.com/archives/C02BAQ5K7/p1550266403153100).

### DIFF
--- a/Artsy/View_Controllers/Fair/ARProfileViewController.m
+++ b/Artsy/View_Controllers/Fair/ARProfileViewController.m
@@ -3,6 +3,7 @@
 #import "ArtsyAPI+Profiles.h"
 #import "Fair.h"
 #import "Profile.h"
+#import "AROptions.h"
 #import "ARSwitchBoard+Eigen.h"
 #import "ARLogger.h"
 
@@ -15,6 +16,7 @@
 
 #import <ReactiveObjC/ReactiveObjC.h>
 #import <FLKAutoLayout/UIView+FLKAutoLayout.h>
+#import <Emission/ARFairComponentViewController.h>
 
 
 @interface ARProfileViewController () <ARMenuAwareViewController>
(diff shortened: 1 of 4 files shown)


Artsy/View_Controllers/Fair/ARProfileViewController.m
```diff
@@ -81,11 +83,14 @@ - (void)loadProfile
...
-
-                RAC(self, hidesNavigationButtons) = RACObserve(viewController, hidesNavigationButtons);
-
-                [self showViewController:viewController];
+                if ([AROptions boolForOption:AROptionsModernShowFairPages]) {
+                    ARFairComponentViewController *viewController = [[ARFairComponentViewController alloc] initWithFairID:fairID];
+                       [self showViewController:viewController];
+                } else {
+                    ARFairViewController *viewController = [[ARFairViewController alloc] initWithFair:fair andProfile:profile];
+                    RAC(self, hidesNavigationButtons) = RACObserve(viewController, hidesNavigationButtons);
```
ashfurrow: I double-checked and this isn't required by our new fair component (`hidesNavigationButtons` is only used because the old fair view controller has its own funky search).

l2succes:
Thanks @ashfurrow, this will also fix Gallery view opening in the new fair page as well

PR #2771, opened 2019-02-19
Fixes hard link simulator installation failure #trivial

ashfurrow:
I've been seeing [this error](https://twitter.com/ashfurrow/status/1068618230426730502) for months, and @kierangillen began seeing it on his machine, too:

screen shot 2019-02-19 at 1 27 29 pm

The solution appears to be the following check box in the main target's build phases:

untitled

I found the solution in one of the many answer is [this StackOverflow post](https://stackoverflow.com/questions/27167020/ios-simulator-could-not-hardlink-copy-wrong-path-in-my-filesystem).


l2succes:
👍

PR #2776, opened 2019-02-22
Fixes crash with opening search on Fair/Show components

ashfurrow:
Opening search while looking at a fair/show view was crashing because we were sending the `pushViewController:animated:` message to the `NavigatorIOS` that the fair/show components are using. This was due to our use of `self.ar_innermostTopViewController.navigationController`, which looks from the git history like a copy/paste thing and not a conscious decision. I've look for other uses of `ar_innermostTopViewController` that might cause trouble, but I think we're good 👍 

In the long run, we really need to get those components doing their routing through Eigen's switchboard instead of having their own navigation controller. I wouldn't be surprised if there are more edge case bugs like this one that we might run into.

Fixes [LD-242](https://artsyproduct.atlassian.net/browse/LD-242).

### DIFF
--- a/Artsy/View_Controllers/Util/ARNavigationController.m
+++ b/Artsy/View_Controllers/Util/ARNavigationController.m
@@ -495,7 +495,7 @@ ShouldHideItem(UIViewController *viewController, SEL itemSelector, ...)
 
 - (BOOL)isShowingSearch
 {
-    return self.ar_innermostTopViewController.navigationController.topViewController == self.searchViewController;
+    return [[[ARTopMenuViewController sharedController] rootNavigationController] topViewController] == self.searchViewController;
 }
 
 - (void)showSearch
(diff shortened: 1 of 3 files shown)


Artsy/View_Controllers/Util/ARNavigationController.m
```diff
@@ -504,9 +504,7 @@ - (void)showSearch
         self.searchViewController = [ARAppSearchViewController sharedSearchViewController];
     }
 
-    UINavigationController *navigationController = self.ar_innermostTopViewController.navigationController;
-    [navigationController pushViewController:self.searchViewController
-                                    animated:ARPerformWorkAsynchronously];
+    [[[ARTopMenuViewController sharedController] rootNavigationController] pushViewController:self.searchViewController animated:ARPerformWorkAsynchronously];
```
orta: OK cool - so as long as we're definitely deleting ARFairViewController from Eigen then I'm +1 on this, maybe I should try get that done today

ashfurrow: Ah good point, yeah. We're definitely deleting it, the new component controller from Emission will handle all our Fairs.

orta: OK

PR #2779, opened 2019-02-26
Removes native Fair & Shows completely

orta:
This one, is a tad more intense. I think this is more or less everything related to fairs and shows inside Eigen. There's still some models like MapPoint, and PartnerShowFairLocation etc to work through also.


orta:
OK, merge on green

PR #2781, opened 2019-02-26
Removes some unused networking code

orta:
@joeyAghion  spotted some logs from this API, but it doesn't look like it's being used anywhere in the app. So I removed it while waiting on CI for something else.


alloy:
Yup, I remember the count being necessary for the badge count for Eigen < v4, but we don’t use that any longer.

PR #2782, opened 2019-02-26
Fix back button hiding, routing improvements

ashfurrow:
This PR is for [LD-250](https://artsyproduct.atlassian.net/browse/LD-250) and [LD-233](https://artsyproduct.atlassian.net/browse/LD-233); it requires https://github.com/artsy/emission/pull/1351 to be merged _and released_ before it can be merged.

The PR adds the ability for Eigen to recognize query strings sent from Emission (`entity=fair`) and bypass a complex `ARProfileViewController` -> `ARMutableLinkViewController` -> `ARFairBoothComponentViewController` controller hierarchy. It also makes our scroll chief conforming code into a C macro that can be re-used across view controllers.

This PR is a WIP because:

- [x] It needs [the same changes](https://github.com/artsy/emission/pull/1351/files#diff-d00d369a092249f7c8764b15b71abfd5R329) from Emission's app delegate made to Eigen's switchboard (to handle the new fairs/shows routes).
- [x] It needs the new Emission controllers to conform to the `ARMenuAwareViewController` protocol.
- [x] Wait for Emission deploy with the above changes, upgrade the pod here.

Both of these require a new version of Emission to be cut, so we'll hold off until v0 is released.


ashfurrow:
I'm going to go ahead and add the new routes from the Emission PR to our Echo config.

ashfurrow:
Actually, hmm. We probably don't want to do that, since if we _do_ have to update the routes, they'll 404 (since they don't exist on Force). Eigen's binary is tightly coupled the Emission version's routes, so I'm going to add the routes to Eigen as string constants.

ashfurrow:
Fair search looks weird, probably a layout difference with the new navigation controller setup:

screen shot 2019-02-26 at 5 00 44 pm

Artsy/App/ARAppDelegate+Emission.m
```diff
@@ -357,27 +358,31 @@ @implementation ARArtistComponentViewController (ARMenuAwareViewController)
...
-    }
-}
-
-- (UIScrollView *)menuAwareScrollView;
-{
-    return objc_getAssociatedObject(self, &menuAwareScrollViewKey);
-}
-
+/// Macro to extend view controller classes to conform to ARMenuAwareViewController.
+#define MakeMenuAware(ControllerClass) @interface ControllerClass (ARMenuAwareViewController) <ARMenuAwareViewController>\
```
ashfurrow: Before we merge this, we should open a ticket to move from this macro to an Objective-C runtime-based solution ([see @alloy's suggestion](https://twitter.com/alloy/status/1100714728333234176)).

ashfurrow: I've opened a ticket here: https://artsyproduct.atlassian.net/browse/PLATFORM-1263 Wasn't sure where to put it (the universal bug backlog didn't seem right) so I put it under Platform with the client infrastructure tag.

ashfurrow:
Gonna self-merge; this went out in a beta.

PR #2783, opened 2019-02-28
Updates the status bar / nav buttons for full-bleed UIViewControllers

orta:
This should handle the Eigen side of correctly showing a full screen UIVC


Artsy.xcodeproj/project.pbxproj
```diff
@@ -5085,8 +5105,8 @@
 				"${PODS_ROOT}/Artsy+UIFonts/Pod/Assets/Unica77LL-Regular.otf",
 				"${PODS_ROOT}/Artsy+UILabels/Pod/Assets/Chevron_Black.png",
 				"${PODS_ROOT}/Artsy+UILabels/Pod/Assets/Chevron_Black@2x.png",
-				"${PODS_ROOT}/Emission/Pod/Assets/Emission.js",
-				"${PODS_ROOT}/Emission/Pod/Assets/assets",
+				"${PODS_ROOT}/../../Emission/Pod/Assets/Emission.js",
+				"${PODS_ROOT}/../../Emission/Pod/Assets/assets",
```
ashfurrow: I think you need to re-run `pod install`

Artsy/View_Controllers/Util/ARNavigationController.m
```diff
@@ -318,9 +354,16 @@ - (void)makeScrollViewReportToTheChief:(UIViewController<ARMenuAwareViewControll
...
+//{
+//    ChangeButtonVisibility(self.backButton, visible, animated);
+//}
+
 - (void)showBackButton:(BOOL)visible animated:(BOOL)animated
 {
-    ChangeButtonVisibility(self.backButton, visible, animated);
+    [UIView animateIf:animated duration:ARAnimationDuration :^{
+        self.backButton.alpha = visible ? 1 : 0;
+    }];
```
ashfurrow: Any reason we need to use `UIView` animations instead of the `CAAnimation` in `ChangeButtonVisibility`?

orta: Yeah, I was gonna try take it back to that

l2succes: Yeah it isn't needed I was just testing something. But actually why are we using `CAAnimation` since it's so much more verbose?

orta: kinda looks like everything did, so it stayed [that way](https://github.com/artsy/eigen-private/blame/3963bfd12ea91202cdcdd8d3429bc02af6d74ff1/Artsy/Classes/View%20Controllers/ARNavigationController.m)

Artsy/App/ARSwitchBoard.m
```diff
@@ -287,10 +289,15 @@ - (void)updateRoutes
         [self registerPathCallbackForDomain:route.path callback:presentNativeAuctionsViewControllerBlock];
         [self registerPathCallbackForDomain:stagingDomain callback:presentNativeAuctionsViewControllerBlock];
     }
+    
 
     // This route will match any single path component and thus should be added last.
     // It doesn't need to run through echo, as it's pretty much here to stay forever.
     [self.routes addRoute:@"/:slug" priority:0 handler:JLRouteParams {
+        if ([parameters[@"entity"] isEqualToString:@"fair"]) {
```
l2succes: We should remove the check for fair below

l2succes: I'll do it

l2succes:
merge on green

PR #2784, opened 2019-02-28
Back button hiding bug fix

ashfurrow:
This PR fixes a problem where the back button wasn't hiding in some Emission component controllers. Mostly this fix is in this commit: 
https://github.com/artsy/eigen/commit/5e4b265ee480abd7b9d36f9368e61e329112c13f However, this wasn't sufficient.

Our code relies on `viewDidLayoutSubviews`, which _should_ get called when the component controller lays out its subviews. Unfortunately, it's not working for some of our component controllers; it gets called while the view hierarchy is still just the spinner, and isn't called again after the full UI is laid out. This was unexpected. A quick look at the different components doesn't point to any obvious differences why the fair component works but the fair artists component doesn't.

We [already have a ticket](https://artsyproduct.atlassian.net/browse/PLATFORM-1263) to revisit this mechanism in general so, while I dislike the `dispatch_after` hack, it does work for now. Open to discussion.

### DIFF
--- a/Artsy/App/ARAppDelegate+Emission.m
+++ b/Artsy/App/ARAppDelegate+Emission.m
@@ -28,17 +28,25 @@
 #import <Emission/AREventsModule.h>
 #import <Emission/ARTakeCameraPhotoModule.h>
 #import <Emission/ARRefineOptionsModule.h>
-#import <Emission/ARFairComponentViewController.h>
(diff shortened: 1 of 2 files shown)


ashfurrow:
Okay, talked to @orta in-person and this is good enough to ship for now.

PR #2794, opened 2019-03-10
Expose the Emission podspec to the app at runtime

orta:
And notes it in the admin panel.

Screen Shot 2019-03-10 at 5 53 13 PM

#trivial


orta:
Interesting fail, build caching maybe?

PR #2795, opened 2019-03-11
Use Echo to set up the local disco option

orta:
It has the same structure as the BNMO options - but also forces all admins into using it.


Artsy/App/ArtsyEcho+LocalDisco.m
```diff
@@ -0,0 +1,26 @@
...
+/// To be kept in lock-step with the corresponding echo value, and updated when there is a breaking Maps change.
+/// https://echo-web-production.herokuapp.com/accounts/1/features
+///
+NSInteger const ARLocalDiscoCurrentVersionCompatibility = 0;
+
+@implementation ArtsyEcho (LocalDiscovery)
+
+- (BOOL)shouldShowLocalDiscovery
+{
+    return (self.features[@"AREnableLocalDiscovery"].state && self.isLocalDiscoCompatible) || [ARAppStatus isBetaDevOrAdmin];
```
ashfurrow: Can we add a check for `is iPad?` here? We don't support LD on iPad.

ashfurrow: Or maybe in the other function.

orta: yeah, that makes sense

orta:
OK, that's in and it should green

PR #2796, opened 2019-03-11
Adds new routes

kierangillen:
- Adds CityListSection route and CityFairList route

### DIFF
--- a/Artsy/App/ARSwitchBoard.m
+++ b/Artsy/App/ARSwitchBoard.m
@@ -256,15 +256,12 @@ NSInteger const ARLiveAuctionsCurrentWebSocketVersionCompatibility = 4;
         return [[ARFairMoreInfoComponentViewController alloc] initWithFairID:parameters[@"id"]];
     }];
     
-    [self.routes addRoute:@"/city/" handler:JLRouteParams {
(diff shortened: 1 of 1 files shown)


Artsy/App/ARSwitchBoard.m
```diff
@@ -253,7 +253,18 @@ - (void)updateRoutes
     [self.routes addRoute:@"/fair/:id/info" handler:JLRouteParams {
         return [[ARFairMoreInfoComponentViewController alloc] initWithFairID:parameters[@"id"]];
     }];
+    
+    [self.routes addRoute:@"/city/" handler:JLRouteParams {
```
ashfurrow: We can use `/city/:city_slug/:section` for the route string, and then the `parameters` object in the callback will have `parameters[@"city_slug"]` and `parameters[@"section"]` values. We should use those instead, and similar for the city fair below.

ashfurrow: This is kind of tricky because the block looks weird (`JLRouteParams` is a C macro that expands to include the `parameters` param). So it is a little confusing!

ashfurrow:
We're going to merge as-is and add the routes for the other VC's once they're ready.

PR #2798, opened 2019-03-12
Emission updates and new routes #trivial

ashfurrow:
Updates Emission. Follow-up to #2796 for new routes. Also includes Pulley change for LD-398. Merge on green.


ashfurrow:
This is already been queued for deploy.

orta:
merge on green

PR #2805, opened 2019-03-13
Hooks up Echo to make API calls

orta:
Our current betas don't update echo, which is a bit strange. This fixes that.


orta:
I broke this in #2802

Artsy_Tests/App_Tests/ARSwitchboard+EchoSpec.m
```diff
@@ -11,6 +11,11 @@ @interface ARStubbedEcho : Aerodramus
 
 @implementation ARStubbedEcho
 
+- (NSString *)name
+{
+    return @"ARStubbedEcho";
```
ashfurrow: Cute.

PR #2810, opened 2019-03-19
Only updates echo once per switchboard call.

ashfurrow:
This is kind of a follow-up to https://github.com/artsy/Aerodramus/pull/9. I saw some logs that made me suspect that Echo might be getting updated a lot. In fact, if the JSON parsing for the updated Echo config fails (or you comment out that line), you'll see an infinite loop:

Screen Shot 2019-03-19 at 7 06 56 PM

That screenshot shows an echo `update:` call invoking its callback, which calls `updateRoutes` on the switchboard, which checks for updates again, and which (if the parsing fails) will repeat forever. 

(The problem where `updateRoutes` is called on a background thread should be fixed by https://github.com/artsy/Aerodramus/pull/9 )

This PR introduces a guard against these repeated calls. I'm still verifying for sure this is the cause of the crash [here](https://sentry.io/organizations/artsynet/issues/942429014/?project=166784&referrer=slack&statsPeriod=14d) but this is a good change to make anyway.

### DIFF
--- a/Artsy/App/ARSwitchBoard.m
+++ b/Artsy/App/ARSwitchBoard.m
@@ -74,6 +74,8 @@ NSInteger const ARLiveAuctionsCurrentWebSocketVersionCompatibility = 4;
 @property (nonatomic, readwrite, strong) Aerodramus *echo;
 @property (nonatomic, strong) NSArray<ARSwitchBoardDomain *> *domains;
 
+@property (nonatomic, assign) BOOL isEchoSetup;
+
 @end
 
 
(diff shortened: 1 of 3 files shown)


ashfurrow:
Looks like it's not the JSON saving that's failing, but rather, the date parsing. See: https://github.com/artsy/Aerodramus/issues/3 We should still merge this (it will fix the _apparent_ bug of echo looping) but we _really_ need to fix that Aerodramus problem :sweat_smile:

PR #2811, opened 2019-03-20
Updates to Aerodramus 1.0 #trivial

ashfurrow:
There's an API change (moving from an array to a dictionary) so lots of little changes throughout. I've tested the app locally and can't find any issues, and I've looked at every callsite of `messages`. 

Fixes LD-511


ashfurrow:
Failure looks intermittent; I've restarted.

PR #2812, opened 2019-03-20
Increments ARLocalDiscoCurrentVersionCompatibility #trivial

ashfurrow:
This will allow us to separate out the v1 and v1.1 users (ie: to only allow v1.1 to access LD). We don't _need_ to, but we might _want_ to, so why not ¯\_(ツ)_/¯


ashfurrow:
Chatted in-person. Going to merge.

PR #2815, opened 2019-03-26
Echo fix for Local Discovery

ashfurrow:
Also fixes a compiler error with Xcode 10.2.


ashfurrow:
Going to merge this to unblock subsequent Xcode 10.2 PRs. @orta FYI this fixes the `<` vs `>` bug I mentioned in Slack on Monday.

PR #2816, opened 2019-03-26
Fixes email autocomplete during login/signup

ashfurrow:
When iOS inserts an autocomplete email address, it adds a space at the end. Which isn't good, and other engineers have filed radars, but alas it breaks our current signup/signin flow (the "next" button gets disabled):

ScreenRecording_03-26-2019 13-27-55 gifcask 2019-03-26 13_30_43

The solution is to trim whitespace characters. It's not clear how data flows through the UI flow, so I made the change in the top-level text field notification-handler.


kierangillen:
Merge on green.

PR #2817, opened 2019-03-27
 Upgrade React Native 0.57 & Emission

l2succes:
I published a new version of Emission (1.9.2) which includes React Native 0.57 and its sub dependencies which I also added to our Specs repo:
- `yoga 0.57.7.React`
- `DoubleConversion 1.1.6`
- `Folly 2016.10.31.00`
- `glog 0.3.5`


orta:
These glog issues look legit, interesting

l2succes:
Yeah :(.... I'm looking into it now. There's some proposed solutions on this stack overflow https://stackoverflow.com/a/51374479. I'll see if one of those solves the issue.

PR #2820, opened 2019-04-02
Fix fastlane promote

orta:
CI is failing on the promote to App Store release - https://circleci.com/gh/artsy/eigen/6067

PR #2822, opened 2019-04-12
Fixes store promotion

ashfurrow:
This commit had a typo that renamed `username` to `apple_id` (`Deliverfile` uses one and `Appfile` uses another):

https://github.com/artsy/eigen/commit/d865463a49b7d3306a777865ab1294d780581c10#diff-915e65323d2a5bb17e964d92abce9497

#2820 tried to fix this but I was getting the same error as it got. I'll keep this as a draft until I verify it works.


ashfurrow:
This did fix the `deliver` problem, though `deliver` failed for another reason that I'll follow-up on Slack about.

ashfurrow:
Okay, this worked. Gonna merge.

PR #2823, opened 2019-04-15
Update to RN 0.59

orta:
This works for a lot of screens, but there are a few queries which fail on staging. Shouldn't be a blocker for merging and making a beta IMO.

Screen Shot 2019-04-15 at 4 08 35 PM

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -96,7 +96,7 @@ target 'Artsy' do
   # use `cat ~/.cocoapods/repos/artsy/Emission/1.x.x/Emission.podspec.json` to see the Podspec
 
   # For Stripe integration with Emission
-  pod 'tipsi-stripe', git: 'https://github.com/tipsi/tipsi-stripe.git', tag: '5.2.4'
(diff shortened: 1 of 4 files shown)


Podfile
```diff
@@ -96,7 +96,7 @@ target 'Artsy' do
   # use `cat ~/.cocoapods/repos/artsy/Emission/1.x.x/Emission.podspec.json` to see the Podspec
 
   # For Stripe integration with Emission
-  pod 'tipsi-stripe', git: 'https://github.com/tipsi/tipsi-stripe.git', tag: '5.2.4'
+  pod 'tipsi-stripe', git: 'https://github.com/tipsi/tipsi-stripe.git', tag: '7.4.0'
```
ashfurrow: Once a beta is submitted, we'll want to double-check that the bid flow credit card entry UI still works (it relies on `tipsi-stripe`).

ashfurrow:
I'll submit a beta now.

PR #2826, opened 2019-05-13
[PLATFORM-1399] Reconfigures Echo

ashfurrow:
This PR moves Stripe publishable keys from our CocoaPods-Keys plugin to Echo. Because checking keys into git is a bad idea, we introduced a new mechanism for ignoring `Echo.json` from the git index, but downloading a fresh copy on `pod install`. It introduces an extra delay on running `pod install`, but I think it's worth it for the simplicity of the solution. We've documented everything in the Readme.

Once this is merged, we should remove the `StripeStagingPublishableKey` and `StripeProductionPublishableKey` env vars from Circle CI. We'll also want to test credit card inputs upon auction registration with the next beta (since that's the only spot we use the Stripe key).

(Note: when we _do_ rotate the Stripe keys, the old ones stay valid for 12 hours. During this time, we expect most users will get the updated Echo config and the key will get passed to Emission on next app launch. This does leave room for users to try to use the old keys, but the problem would be fixed by restarting the app, so I think it's reasonable.)

Fixes [PLATFORM-1399](https://artsyproduct.atlassian.net/browse/PLATFORM-1399).


Makefile
```diff
@@ -51,6 +51,7 @@ oss:
 	bundle exec pod keys set "ArtsyEchoProductionToken" "-"
 	bundle exec pod keys set "SentryProductionDSN" "-"
 	bundle exec pod keys set "SentryStagingDSN" "-"
+	cp Artsy/App/Echo.json.example Artsy/App/Echo.json
```
alloy: 👍

alloy:
Great stuff 👌

PR #2829, opened 2019-05-15
Points react-native-navigator-ios in Podspec to Ash's fork #trivial

ashfurrow:
Follow-up from eb2761a3584d16b4b68350c31c2a43b2540d11a3 and https://github.com/artsy/emission/pull/1586.

This PR will unify Emission+Eigen to point to my fork of `react-native-navigator-ios`. The only functional difference between my fork and upstream is some types that Orta added [in this PR](https://github.com/michaelknoch/react-native-navigator-ios/pull/2).


alloy:
Still think we should probably have these on the Artsy org, but don’t want to block your current work on that 👍

PR #2835, opened 2019-05-21
Fix availability warnings

ashfurrow:
Xcode has a warning (enabled by default for new projects) to guard against APIs requiring iOS versions. This PR enables the new, more aggressive setting, and cleans up our availability checks in general.

I felt  some ambiguity between some existing checks for `@available(iOS 11.0, *)` and `@available(iOS 11.3, *)`. Since `canOpenARView` already uses 11.3, we're not changing any _actual_ app behaviour, and the compiler is now happy, so what more could you want?

(There are several unit tests that don't make sense to include because our CI/tests require iOS 10 and ARKit needs 11.3. Indeed, the weren't really testing much. I'm happy to trade a few unit tests for better static analysis + warnings.)

Thank you to @bjhomer for the suggestion :bow:


alloy:
Very nice 🎀

PR #2836, opened 2019-05-21
Removes unused if statement #trivial

ashfurrow:
This is a follow-up from #2830. The `else if` statement will never be reached because an earlier `if` catches it (with the `prefixless` array, which is likely why the static analyzer didn't catch this).

### DIFF
--- a/Artsy/Models/API_Models/Application/SearchSuggestion.m
+++ b/Artsy/Models/API_Models/Application/SearchSuggestion.m
@@ -46,8 +46,6 @@
         return NSStringWithFormat(@"/artwork/%@", self.modelID);
     } else if ([self.model isEqualToString:@"article"]) {
         return NSStringWithFormat(@"/article/%@", self.modelID);
-    } else if ([self.model isEqualToString:@"page"]) {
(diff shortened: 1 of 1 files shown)


Artsy/Models/API_Models/Application/SearchSuggestion.m
```diff
@@ -46,8 +46,6 @@ - (NSString *)href
         return NSStringWithFormat(@"/artwork/%@", self.modelID);
     } else if ([self.model isEqualToString:@"article"]) {
         return NSStringWithFormat(@"/article/%@", self.modelID);
-    } else if ([self.model isEqualToString:@"page"]) {
-        return NSStringWithFormat(@"/artwork/%@", self.modelID);
```
anandaroop: Thanks for this follow-up! 👍

anandaroop:
Looks like there was a CI fail on the master branch build. Not sure if it's spurious but it does seem search related...

https://circleci.com/gh/artsy/eigen/6148

ashfurrow:
Looks like it was spurious – I restarted it and it worked :)

PR #2837, opened 2019-05-23
Updates Artsy fonts pod #trivial

ashfurrow:
Includes the change in this PR: https://github.com/artsy/Artsy-OSSUIFonts/pull/18 This catches a common problem that Artsy staff run into when setting up this repo – that PR discusses it in more detail, but let me know if I can clarify anything!


ashkan18:
nice!! 🙌 thanks

PR #2838, opened 2019-05-23
Fixes issue with tipsi-stripe

ashfurrow:
This PR moves `tipsi-stripe` to my fork, which includes [this commit](https://github.com/ashfurrow/tipsi-stripe/commit/bfebb733fb44d3e1d0522a38a317993f9c686422) (copied from [this PR](https://github.com/tipsi/tipsi-stripe/pull/418) that was closed without resolution) which fixes [this bug](https://github.com/tipsi/tipsi-stripe/issues/408). The bug _appears_ related to our recent React Native upgrade (I ruled out the `tipsi-stripe` [upgrade from 7.4.0 -> 7.5.0](https://github.com/tipsi/tipsi-stripe/compare/7.4.0...7.5.0)).

The fix was to remove a callback for when the keyboard appears; the callback that was being executed was this:

```objc
- (void)keyboardWillShow:(NSNotification *)n {
    if (!_jsRequestingFirstResponder && !_isFirstResponder) {
        [_paymentCardTextField resignFirstResponder];
    }
}
```

Which reads: "if the request to become a first respond (focus the text field) doesn't come from JS and if I am not the first responder (I have the keyboard focus) then make sure that the payment card text field subview loses focus." 

I've looked through the git history to figure out the purpose of this code, but it's unclear. My _guess_ is it has to do with supporting credit card inputs in a larger form with other inputs. Our UI has _just_ the credit card input on screen at a time (no other text inputs) so the _only_ way the keyboard shows is if if the `_paymentCardTextField` has been tapped by the user and is becoming the responder. Because the code only ever makes the payment field resign first responder and is only ever invoked when the payment field becomes the first responder, I am comfortable shipping this. (This also explains [the crash](https://sentry.io/organizations/artsynet/issues/1033339009/?project=1403105&referrer=slack) we're seeing: stack overflow from infinite show-and-hide-and-show-etc the keyboard.)

Please let me know what I can clarify. I've verified this works as expected on device by going through the complete bid -> register flow successfully (well, I didn't meet the reserve, but the bid did go through).

I will follow-up on the PR to ask why it was closed and will follow-up here.

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -5,7 +5,7 @@ upcoming:
   dev:
     - Adds new availability checks and enables new compiler warning for better checking of API availabilities - ash
   user_facing:
-    - 
+    - Fixes a problem entering credit cards for auction registration - ash
 
 releases:
   - version: 5.0.3
--- a/Podfile
+++ b/Podfile
@@ -93,8 +93,8 @@ target 'Artsy' do
   # Emission's dependencies
   # use `cat ~/.cocoapods/repos/artsy/Emission/1.x.x/Emission.podspec.json` to see the Podspec
 
-  # For Stripe integration with Emission
-  pod 'tipsi-stripe', git: 'https://github.com/tipsi/tipsi-stripe.git', tag: '7.5.0'
+  # For Stripe integration with Emission. Using Ash's fork for this issue: https://github.com/tipsi/tipsi-stripe/issues/408
+  pod 'tipsi-stripe', git: 'https://github.com/ashfurrow/tipsi-stripe.git', branch: 'fix-infinite-loop'
   pod 'react-native-mapbox-gl', git: 'https://github.com/l2succes/react-native-mapbox-gl', branch: 'fix-gesture-recognizer'
   pod 'SentryReactNative', git: 'https://github.com/getsentry/react-native-sentry.git', tag: 'v0.30.3'
   pod 'Pulley', :git => 'https://github.com/l2succes/Pulley.git', :branch => 'master'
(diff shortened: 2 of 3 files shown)


alloy:
LGTM, thanks for the fast fix! 🙌

ashfurrow:
I've heard back from the author of the PR with the changes I've borrowed – they closed the PR because some end-to-end tests weren't working anymore, but that's it. We're going to investigate and compare notes.

DarkKnight1992:
I understand the e2e test was broke..... BUT YOU CAN"T JUST CLOSE A PR WITHOUT FIXING THE ISSUE

ashfurrow:
@DarkKnight1992 We didn't close the PR, that's a different repo. I would be happy to provide further clarification on why this solution works for us, but you are unlikely to get assistance from the open source community without a respectful tone.

DarkKnight1992:
firstly, i haven't disrespected anyone and sorry for not realizing this was a different repo. And i have a point.

alloy:
Even if this were the right repo _and_ besides your all-caps tone, which usually doesn’t inspire a fruitful debate–you’re not respecting the prerogative that the author has to close their own PR, they owe you absolutely nothing.

If you wish to get to a place where your problem is solved, which I assume you want, then I would suggest you try a different approach when interacting with the author of the PR. _Or_ you can exercise your right that the MIT license gives–which the authors were so friendly to give to you _for free_–and apply the change on a fork, like we did in this PR.

Either way, please don’t continue down this line here 🙏

DarkKnight1992:
My apologizes didn't mean to offend anybody.

PR #2846, opened 2019-06-13
Moved artwork set analytics into artwork set view controller

ashfurrow:
We saw that the analytics for swiping between view controllers weren't firing for artwork set view controllers _opened from an auction_. The others would work no problem, just not the ones presented from the Auction view controller. Super-weird. 

I dug through the ARAnalytics code, and then the ReactiveCocoa code its built on top of, but couldn't find any reason for `rac_signalForSelect:` not to fire for certain instances. I'm a bit worried we might be missing analytics events in other places, though with more and more of our codebase moving to React Native, it might be worth ripping out ARAnalytics+DSL altogether soon. This would also let us remove ReactiveCocoa and get a smaller app size.

### DIFF
--- a/Artsy/App/ARAppDelegate+Analytics.m
+++ b/Artsy/App/ARAppDelegate+Analytics.m
@@ -136,45 +136,6 @@
                             ]
                     },
                 @{
-                    ARAnalyticsClass: ARArtworkSetViewController.class,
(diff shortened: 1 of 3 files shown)


alloy:
That sounds like a nice prospect ✨

PR #2850, opened 2019-06-25
PURCHASE-1171: If an artist is consignable, a user can click to consign their work through Artsy

lilyfromseattle:
Adds a link to the consignments flow information page.

### DIFF
--- a/Artsy/App/ARSwitchBoard.m
+++ b/Artsy/App/ARSwitchBoard.m
@@ -234,6 +234,10 @@ NSInteger const ARLiveAuctionsCurrentWebSocketVersionCompatibility = 4;
         return [[ARNavigationController alloc] initWithRootViewController:submissionVC];
     }];
 
+    [self.routes addRoute:@"/consign/info" handler:JLRouteParams {
+        return [[ARShowConsignmentsFlowViewController alloc] init];
+    }];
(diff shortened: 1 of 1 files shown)


Artsy/App/ARSwitchBoard.m
```diff
@@ -234,6 +234,10 @@ - (void)updateRoutes
         return [[ARNavigationController alloc] initWithRootViewController:submissionVC];
     }];
 
+    [self.routes addRoute:@"/consign/info" handler:JLRouteParams {
```
sweir27: It looks like both this and `/consign/submission` are instantiating `ARShowConsignmentsFlowViewController` (and based on the screenshot in the ticket it looks like that's what we want anyways?)

I'm curious why we need this extra route.

lilyfromseattle: `/consign/submission` links to the second step of the consignments flow where the user enters the information about the work. Both routes initialize ARShowConsignmentsFlowViewController but the route above overrides the default behavior which is to show the info page first and initializes it with the submission page as the root, which is not what we want for this story.

sweir27: Thanks for the explanation!

PR #2852, opened 2019-06-28
Cleans up unused lab options

ashfurrow:
I noticed that we have a few lab options that we aren't using anymore, so I went through and pruned them down. Now the labs screen looks like this:

Screen Shot 2019-06-28 at 16 58 20


sweir27:
Niiiice thanks for this!

merge on green

PR #2853, opened 2019-07-01
Updates version number

ashfurrow:
After 5.0.5 was released, we need to do two things:

- Add a new version (5.0.6 in this case) to Apple's AppStoreConnect dashboard, and
- Edit our changelog and Xcode projects to match that new version.

More details are [in the docs](https://github.com/artsy/eigen/blob/master/docs/deploy_to_app_store.md#prepare-for-the-next-release).

We will also need to pull the changes from the `5.0.5-develop` branch into `master`, which I'll take care of later today.

### DIFF
--- a/Artsy Stickers/Info.plist
+++ b/Artsy Stickers/Info.plist
@@ -17,7 +17,7 @@
 	<key>CFBundlePackageType</key>
 	<string>XPC!</string>
 	<key>CFBundleShortVersionString</key>
-	<string>5.0.5</string>
+	<string>5.0.6</string>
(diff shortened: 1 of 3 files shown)


sweir27:
Cool! Looks good.

And thank you for the link to docs! 🙏 

merge on green

ashfurrow:
The failure is caused by a problem with a downstream dependency of Danger's (see [the discussion here](https://github.com/danger/danger-js/issues/884)) so I'm going to merge as-is with the 🔴

PR #2855, opened 2019-07-01
[PURCHASE-1236] Add edition info to LAI interface

ashfurrow:
This adds an artwork's edition info to the LAI interface under "Lot Info":

Screen Shot 2019-07-01 at 3 37 15 PM

This PR is a work-in-progress because we're waiting for a finalized design.

Lily and I also changed how the dimensions are displayed: users with their locale set to use the metric system will now see `cm` values, and others see `in` values. This is a change from before, when _everyone_ saw `cm` values 😛 /cc @alloy


alloy:
I’m in favour of that change, regardless of how superior the metric system is 😛

ashfurrow:
Here's what the finalized design looks like:

Screen Shot 2019-07-02 at 3 22 11 PM

Artsy/View_Controllers/Live_Auctions/ViewModels/LiveAuctionLotViewModel.swift
```diff
@@ -151,13 +155,21 @@ class LiveAuctionLotViewModel: NSObject, LiveAuctionLotViewModelType {
     }
 
     var lotArtworkDimensions: String? {
-        return model.artwork.dimensionsCM
+        if self.usersLocale.usesMetricSystem {
+            return model.artwork.dimensionsCM
+        } else {
+            return model.artwork.dimensionsInches
+        }
```
alloy: Not something we should do now, but ideally we wouldn’t even fetch unneeded details–i.e. use the locale defaults at query time instead.

Artsy_Tests/Networking_Tests/LiveAuctionLotViewModelSpecs.swift
```diff
@@ -94,5 +94,31 @@ class LiveAuctionLotViewModelSpec: QuickSpec {
...
+                subject = LiveAuctionLotViewModel(lot: lot!, bidderCredentials: creds)
+            }
+
+            it("works for inches") {
+                subject.usersLocale = Locale(identifier: "en-US")
+                expect(subject.lotArtworkDimensions) == "2x2in"
+            }
+
+            it("works for centimeters") {
+                subject.usersLocale = Locale(identifier: "en-CA")
```
alloy: 🇨🇦 🤘

Artsy/View_Controllers/Live_Auctions/ViewModels/LiveAuctionLotViewModel.swift
ashfurrow: Oh yeah, that'd be cool!

PR #2856, opened 2019-07-02
React Native Artwork view AR VIR invocation

ashfurrow:
This PR, which depends on https://github.com/artsy/emission/pull/1698 to function properly, adds a callback hook for Emission to show the Augmented Reality View-in-Room experience. I've tried to document the different parts of the code.

I haven't added unit tests for this because we have almost zero testing infrastructure around either our app delegate/Emission integration, or our Augmented Reality code. This will be extra difficult to test because our unit tests run on iOS 10, so `if (@available(iOS 11.3, *)) {` just won't work. I don't think that the extra benefit of having unit tests is outweighed by the effort it would take to set this infrastructure up, though I'm happy to keep discussing it!


Artsy/App/ARAppDelegate+Emission.m
```diff
@@ -194,6 +198,14 @@ - (void)setupSharedEmissionWithPackagerURL:(NSURL *)packagerURL;
         }];
     };
 
+    emission.APIModule.augmentedRealityVIRPresenter = ^(NSString *imgUrl, CGFloat width, CGFloat height, NSString *artworkSlug, NSString *artworkId) {
```
sweir27: I know this would be more back-and-forth, but given that it's a requirement that these dimensions are in inches, it would be nice if they were named accordingly (`widthIn` and `heightIn`). Curious if you think that would be valuable and/or if you have thoughts on whether to do this now or later?

Artsy/App/ARAppDelegate+Emission.m
```diff
@@ -295,6 +307,75 @@ - (NSDictionary *)getOptionsForEmission:(NSDictionary *)echoFeatures labOptions:
     return options;
 }
 
+#pragma mark - AR View-in-Room Experience
+
+// This is largely copied from ARArtworkViewController+ButtonActions.m. Normally we would
```
sweir27: These comments are helpful, but they too will be unnecessary once the rollout is complete.

Can we add a TODO to remove it?

ashfurrow: Yeah that's a good idea!

Artsy/App/ARAppDelegate+Emission.m
ashfurrow: There's no cost to change them now, or later. But we might as well do it now 👍 /cc @saolsen

ashfurrow: Okay I've opened https://github.com/artsy/emission/pull/1699 to address this. Since this is only a change of the parameter name (used only at the implementation site, and not the call site), I've already updated this PR to use the new names. We can merge when ready without waiting for the Emission PR 👍

Artsy/App/ARAppDelegate+Emission.m
ashfurrow: I've added a TODO above this comment 👍

Artsy/App/ARAppDelegate+Emission.m
```diff
@@ -194,6 +198,15 @@ - (void)setupSharedEmissionWithPackagerURL:(NSURL *)packagerURL;
         }];
     };
 
+    emission.APIModule.augmentedRealityVIRPresenter = ^(NSString *imgUrl, CGFloat widthIn, CGFloat heightIn, NSString *artworkSlug, NSString *artworkId) {
+        // A bit weird, eh? Normally CGSize stores width+height in terms of pixels, but this one is stored in inches instead.
```
sweir27: 😆

PR #2858, opened 2019-07-09
Correctly reference isInquireable.boolValue for artwork view switching logic

sweir27:
My bad!

Thought I had tested this in the simulator after adding this piece of logic but clearly I had not. 😞Works now!

#trivial


alloy:
Hah, yeah makes sense 👍

PR #2866, opened 2019-07-12
[PURCHASE-1117, PURCHASE-1318] Users are no longer required to bid one increment above asking price on upcoming LAI lots

ashfurrow:
Assigning this to @dleve123 because it's been a while since I've been in-the-weeds of auctions logic and want to make sure we get this right.

This PR changes the behaviour of bidding on **upcoming** LAI lots. The client calculates its own bidding increments, and it now starts _at_ the asking price (instead of one increment above it). Let me know if that makes sense 👍

Screen Shot 2019-07-12 at 14 59 38


Artsy/View_Controllers/Live_Auctions/LiveAuctionBidViewModel.swift
```diff
@@ -104,6 +102,11 @@ class LiveAuctionBidViewModel: NSObject {
         return value.convertToDollarString(lotViewModel.currencySymbol)
     }
 
+    var currentAskingPrice: UInt64 {
```
erikdstock: is `var x: UInt64 { block that returns x }` just a lazy initializer?

Artsy/View_Controllers/Live_Auctions/LiveAuctionBidViewModel.swift
```diff
@@ -104,6 +102,11 @@ class LiveAuctionBidViewModel: NSObject {
...
+        // Should not return 0, this is just to satisfy the compiler.
+        return lotViewModel.askingPriceSignal.peek() ?? 0
```
erikdstock: what is `??` ? `||` ?

ashfurrow: Yup, exactly!

Artsy/View_Controllers/Live_Auctions/LiveAuctionBidViewModel.swift
ashfurrow: It's a computed property, actually. So every time someone accesses the `.currentAskingPrice` property, the code in the `{ block }` gets executed.

You _can_ do what you suggest by using Swift's `lazy` keyword, and also adding some indirection (make the property equal to a self-evaluating closure, [which I discuss here](https://ashfurrow.com/blog/lazy-property-setup-in-swift/)). But then the closure is only ever executed once.

dleve123:
@ashfurrow Can you post a screenshot attesting to what happens when there's an _existing_ max bid on a lot, for a live auction, but before the lot has hit the block. I _believe_ (but could be wrong) that we _should_ see the next increment above the prior max bid in this case.

ashfurrow:
@dleve123 sure, good idea. I'll get to that – in the mean time, could you remind me what "asking price" means from a Causality context? Is it the current highest bid amount? Or the bid amount of the next bid?

dleve123:
@ashfurrow 

> @dleve123 sure, good idea. I'll get to that – in the mean time, could you remind me what "asking price" means from a Causality context? Is it the current highest bid amount? Or the bid amount of the next bid?

`askingPriceCents` is the value that we're asking from online bidders on Artsy at any given time. During live, its 1 increment above the **minimum needed for Artsy to win**  - so in the attached screenshot, there's competition from Artsy bidders such that Artsy **has to** bid at least 17, so the `currentAsk` is 18. Note that floor can be much behind the current ask, in this case, the floor is currently asking for 16.

askingPriceCents-is-what-online-is-asking-for

Hope this helps, let's talk synchronously tomorrow if you have any more questions.

ashfurrow:
@dleve123 thanks, that's really helpful!

In the below screenshots, I have two different users bidding on an upcoming lot. Through Prediction, the left user has bid 3k and is now being prompted to increase their bid to 3250. Through Eigen, the right user has opened the bidding interface for the same lot and is now being asked to bid the (same) asking price:

Screen Shot 2019-07-16 at 14 33 13

So in both cases, the user is being asked to bid the increment above the current highest bid. (Note however that Prediction initially had 4k _selected_ as the next bid (with 3250 also available), but when I navigated away and back to that lot, 3250 was selected, so I think it's a Prediction quirk). So I think we're 👍 to go on this.

dleve123:
@ashfurrow I have some questions about this screenshot, particularly about how:

> Through Prediction, the left user has bid 3k and is now being prompted to increase their bid to 3250. 

But prediction is displaying a "current bid" of 3250...

Can we find a few mins to chat about this later today or tomorrow? Want to let me know some times that work for you (over Slack).

ashfurrow:
Hmm yeah, let's follow-up in-person.

ashfurrow:
Holding on this for now, see details here: https://artsyproduct.atlassian.net/browse/PURCHASE-1318

Artsy_Tests/View_Controller_Tests/Live_Auction/LiveAuctionBidViewControllerTests.swift
```diff
@@ -63,9 +63,10 @@ class LiveAuctionPlaceMaxBidViewControllerSpecs: QuickSpec {
 
             it("ensures the bid is moved above the current max bid") {
```
dleve123: I would recommend elaborating the docstring here: "it ensures that the minimum possible bid is at the asking price" considering that this is pretty tricky stuff

Artsy_Tests/View_Controller_Tests/Live_Auction/LiveAuctionBidViewControllerTests.swift
```diff
@@ -63,9 +63,10 @@ class LiveAuctionPlaceMaxBidViewControllerSpecs: QuickSpec {
...
 
-                expect(subject.bidViewModel.currentBid) == 5_500_00
+                expect(subject.bidViewModel.currentBid) == 5_000_00
 
-                salesPerson.currentLotValue = 6_000_00
+                lotVM.askingPrice = 6_000_00
+                lotVM.askingPriceSignal.update(lotVM.askingPrice)
```
dleve123: perhaps this is some swift-iOS specific thing, but don't really understand why we need to do this update...

ashfurrow:
@dleve123 gotcha. I'll apply the fix from https://artsyproduct.atlassian.net/browse/PURCHASE-1318 in this PR as well, since it's very minor. (We tested this change locally yesterday but I haven't committed it.)

Artsy_Tests/View_Controller_Tests/Live_Auction/LiveAuctionBidViewControllerTests.swift
ashfurrow: Sure, it's a good question. The lot view model maintains internal state (it's `askingPrice` property) as well as a signal that emits changes to this property. Due to constraints (which I'm happy to discuss if you're curious), the update is not automatic. So we first update the property, and then update the signal to reflect that property's change. 

Adding the update for the signal is necessary here because, previously, the test was relying on default values provided by the salesperson (but now relies directly on the lot view model).

dleve123:
> @dleve123 gotcha. I'll apply the fix from https://artsyproduct.atlassian.net/browse/PURCHASE-1318 in this PR as well, since it's very minor. (We tested this change locally yesterday but I haven't committed it.)

Cool! Perhaps you are already planning on doing so, but can you upload a screencapture/movie of that bug being fixed once it is?

ashfurrow:
@dleve123 yeah definitely! Here's the app showing the correct, and updating, value in the purple bottom view:



(You can see me placing floor bids from Operator while looking at the current lot.)

dleve123:
hey @ashfurrow do we fix described in https://github.com/artsy/eigen/pull/2866#issuecomment-512948913 to be in the latest testflight (one released today)? We are doing some auctions QA but didn't see this fixed...

PR #2869, opened 2019-07-17
Fixes problem not being able to open LAI sales.

ashfurrow:
Introduced in #2867, it was caused by a mutex lock. The solution here is, instead of merging signals, we no longer filter non-bidding status updates from our signal. That is to say: we were only letting observers know about new bidding status updates when a subset of fields got updated, and now we send all updates from Causality through the signal.


yuki24:
Sweet, LGTM!

PR #2871, opened 2019-07-18
[PURCHASE-1277] Improves UX and TX for new artwork view

ashfurrow:
This PR improves the user experience and tester experience of the new artwork view. It does a bunch of little things, so here's the summary:

- The artwork view, which appears on-screen while we do an MP call to determine which _actual_ artwork view to present, is now opaque white. It shows the spinner immediately instead of fading it in (the fade animation was taking longer than the MP request).
- There are new, more gradual lab options for testing the view. The lab options will now always appear in the same order, too (alphabetized). See the screenshot below.
- Applies consistent indentation to the `ARArtworkViewController` class and its tests (looks like we were mixing-and-matching 2 and 4 tab widths). For this reason, I recommend reviewing this PR with GitHub's "ignore whitespace" option.

Screen Shot 2019-07-18 at 11 28 04

### DIFF
--- a/Artsy/View_Controllers/Artwork/ARArtworkViewController.m
+++ b/Artsy/View_Controllers/Artwork/ARArtworkViewController.m
@@ -21,7 +21,24 @@
 
 - (BOOL)shouldShowNewVersion;
 {
-    return [AROptions boolForOption:AROptionsRNArtwork] && self.artwork.availability != ARArtworkAvailabilityForSale && !self.artwork.isInquireable.boolValue;
+    if ([AROptions boolForOption:AROptionsRNArtworkAlways]) {
+        return YES;
+    }
+
+    BOOL isArtworkNonCommerical = self.artwork.availability != ARArtworkAvailabilityForSale && !self.artwork.isInquireable.boolValue;
+    BOOL isArtworkNSOInquiry = self.artwork.isOfferable || self.artwork.isAcquireable || self.artwork.isInquireable;
+    BOOL isArtworkAuctions = self.artwork.saleArtwork;
+
+    if ([AROptions boolForOption:AROptionsRNArtworkNonCommerical] && isArtworkNonCommerical) {
+        return YES;
+    } else if ([AROptions boolForOption:AROptionsRNArtworkNSOInquiry] && isArtworkNSOInquiry) {
+        return YES;
+    } else if ([AROptions boolForOption:AROptionsRNArtworkAuctions] && isArtworkAuctions) {
+        return YES;
+    }
+
+    // Should never get here, but let's return NO to satisfy the compiler.
+    return NO;
 }
 
 - (instancetype)initWithArtwork:(Artwork *)artwork fair:(Fair *)fair;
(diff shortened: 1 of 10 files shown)


Artsy/View_Controllers/Artwork/ARArtworkViewController.m
```diff
@@ -47,42 +64,39 @@ - (void)loadView;
 - (void)viewDidLoad
 {
     [super viewDidLoad];
+    self.view.backgroundColor = [UIColor whiteColor];
```
ashfurrow: @alloy I want you to know that I initially wrote this as `UIColor.whiteColor` but thought better of it.

alloy: 🍪 👍

Artsy/View_Controllers/Artwork/ARArtworkViewController.m
```diff
@@ -47,42 +64,39 @@ - (void)loadView;
...
+                return;
+            }
+            __strong typeof(wself) sself = wself;
+
+            [sself.spinner stopAnimating];
+            [sself.spinner removeFromSuperview];
+
+            if (sself.shouldShowNewVersion) {
+                ARArtworkComponentViewController *vc = [[ARArtworkComponentViewController alloc] initWithArtworkID:sself.artwork.artworkID];
+                [sself ar_addAlignedModernChildViewController:vc];
```
alloy: 👍

Artsy/View_Controllers/Artwork/ARArtworkViewController.m
```diff
@@ -20,7 +21,23 @@ @implementation ARArtworkViewController
...
 - (BOOL)shouldShowNewVersion;
 {
-    return [AROptions boolForOption:AROptionsRNArtwork] && self.artwork.availability != ARArtworkAvailabilityForSale && !self.artwork.isInquireable.boolValue;
+    if ([AROptions boolForOption:AROptionsRNArtworkAlways]) {
+        return YES;
+    }
+
+    BOOL isArtworkNonCommerical = self.artwork.availability != ARArtworkAvailabilityForSale && !self.artwork.isInquireable.boolValue;
+    BOOL isArtworkNSOInquiry = self.artwork.isOfferable || self.artwork.isAcquireable || self.artwork.isInquireable;
+    BOOL isArtworkAuctions = self.artwork.saleArtwork;
```
sweir27: I don't think we want this specific logic, since past-auction works will still have a `saleArtwork`. I believe we should use the `is_in_auction` field returned from metaphysics here.

ds300:
Looks great! Just waiting for sarah's comment to be addressed and then I think we're good to go.

What is TX? (T)ab-width e(X)perience?

Artsy/View_Controllers/Artwork/ARArtworkViewController.m
ashfurrow: Ah! Yes excellent point. I'd originally thought that we were fetching the Artwork model from Gravity and had to use this, and only later learned that it's coming from metaphysics. This change makes a lot of sense 👍

ashfurrow:
@ds300 tab-width experience makes sense! I'm trying to coin a new abbreviation – how does "TX: Tester Experience" sound?

PR #2878, opened 2019-08-01
Updates Documentation

ashfurrow:
I reviewed our Eigen+Emission docs for outdated or duplicated content – expect an Emission PR shortly. I also ran the docs through Prettier for the first time in a while, which are in two separate commits here.


docs/updating_emission.md
```diff
@@ -2,25 +2,25 @@
...
-## Get Emission runtime 
+## Get Emission runtime
 
-Does it crash on launch? 
+Does it crash on launch?
 
-* Might need to add new settings to [the `AREmissionConfiguration` object](https://github.com/artsy/emission/blob/master/Pod/Classes/Core/AREmission.h) in Eigen's side. Ideally these should get raised by compiler warnings, but you never know.
-
-* A native module may have new exposed callbacks which are required to be set up on launch.
+- Sometimes the simulator will crash after you've run unit tests in it. Uninstall the app from the simulator to fix the crash.
```
alloy: Huh, never had that happen to me.

alloy:
Niiiice ✨

docs/updating_emission.md
ashfurrow: Next time it happens to me, I want to dig in a little deeper to figure out the problem. If I recall correctly, it was this parameter assert that was failing:

https://github.com/artsy/emission/blob/2f73a9ea7f0f0d799c31a4456e5e7053de5dd39e/Pod/Classes/Core/AREmission.m#L85

alloy: Hmm. Ok, yeah please do 👍

PR #2884, opened 2019-08-12
[PURCHASE-1143] Removes ARArtworkSetViewController

ashfurrow:
This PR removes the `ARArtworkSetViewController` class. Previously, we had a three-layer view controller hierarchy:

```
ARArtworkSetViewController (provides left-and-right swiping)
👇 
ARArtworkViewController (provides layer of indirection between legacy and RN artwork view controllers)
👇
ARArtworkComponentViewController or ARLegacyArtworkViewController (shows the artwork)
```

This left-to-right swiping conflicts with our new image carousel, and the extra layer of indirection was causing a lot of headaches while debugging. So we're going to remove it.

This PR is a work-in-progress because it depends on a forthcoming Emission PR.


ds300:
Looks great  🎉 #mergeOnGreen

PR #2886, opened 2019-08-12
Fixes rotating to open non-AR VIR view

ashfurrow:
Follow-up from #2884. I copied this from previous implementation: 

https://github.com/artsy/eigen/blob/a3c05dc21bd5bb6af2a95d2c38a99e48cd4102fc/Artsy/View_Controllers/Artwork/ARArtworkSetViewController.m#L93-L117

(Though I did remove the TODO, since it worked without it.)

Old artwork view:

old

New artwork view:

new


ashfurrow:
Talked with Sam and we're going to remove the non-AR VIR altogether. I'll open another PR that disables landscape on this view.

PR #2887, opened 2019-08-13
[PURCHASE-1366] Adds echo flag support for enabling various types of RN Artwork view.

ashfurrow:
This adds new flags. Note that they are **enabled** on Echo, so if you need to see the old view, you will have to disable them.


ashkan18:
Nice! and thanks for knowledge share! #mergeOnGreen

PR #2889, opened 2019-08-21
Switch to CDN version of CocoaPods' master spec repo

alloy:
🚀


alloy:
Gah, these tests all pass locally 🤔

alloy:
Well, I give up for now, this is taking up too much time.

@ashfurrow [for when you’re back] Any ideas why this might fail on CI only?

ashfurrow:
@alloy The failing tests are largely switchboard tests, which I've seen behave differently based on test order. My initial thought was that the new changes would prevent Echo's JSON from being updated, but I checked the logs for `pod install` and everything looks fine.

Some ideas for next steps:

- Bust the CocoaPods cache (ie: change `v3-pods-{{ checksum "Podfile.lock" }}` to `v4-pods-{{ checksum "Podfile.lock" }}`)
- Re-run the CI job with SSH enabled, and download the CI machine's code with `scp` to see if tests fail locally.

ashfurrow:
This was failing because `make update_echo` failed because `$(shell bundle exec pod keys get ArtsyEchoProductionToken Artsy)` included extra output used in the `Http-Authorization` header, causing the downloaded JSON to be `{"error": "Unauthorized"}`. We saw similar behaviour in #2962 and fixed it by looking at the invocation used while evaluating the Podfile.

My hope is that these changes will also fix this PR.

ashfurrow:
Shoot, still failing.

ashfurrow:
I ssh'd into the build machine and was getting this error from `pod keys get`:

```
[!] Could not find a project for this folder
```

Which was causing `make update_echo` to fail. I've modified the Makefile to refer to the `ArtsyEchoProductionToken` env var, if present, instead of the `pod keys get`. I don't like that our CI and dev machine behaviour diverges in this way, but I think it's fairly minor, and I made a comment.

ds300:
hoooray! Thanks @ashfurrow 👍

PR #2896, opened 2019-09-03
[PURCHASE-1414] Fixes Analytics

ashfurrow:
A few things in this PR:

- Fixed a problem where we were showing users the new Artwork page for Auctions artworks /cc @sweir27 
- The UI for analytics debugging was being presented from a background thread, likely causing the delays and other UI bugs. I moved it to the main thread with `dispatch_async`.
- Presenting the detail view on iPad required some additional UIAlertController configuration. I added a link for further reading in a comment.

I also investigated the errant `"tap"` events being tracked:

Screen Shot 2019-09-03 at 15 26 29

One of the most used components in Emission, `GenericGrid`, fires this event _on layout_:

https://github.com/artsy/emission/blob/ce9ce5eaa143216fc9cd8a6fc41bee2eba3e2bf7/src/lib/Components/ArtworkGrids/GenericGrid.tsx#L42-L48

This seems like a bug to me, but I need to do more digging. I've verified that duplicate events are _not_ being generated, and that these extra events are coming from bugs in how we use tracking in Emission. I'll keep the Jira ticket open, but will follow-up with an Emission PR.

### DIFF
--- a/Artsy/App/ARAnalyticsVisualizer.m
+++ b/Artsy/App/ARAnalyticsVisualizer.m
@@ -11,27 +11,38 @@
 
 - (void)event:(NSString *)event withProperties:(NSDictionary *)properties
 {
-    NSString *title = [self alertTitleForEvent:event withProperties:properties];
+    dispatch_async(dispatch_get_main_queue(), ^{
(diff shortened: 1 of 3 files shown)


sweir27:
Thank you!!

PR #2902, opened 2019-09-13
Fixes crash when AuctionViewController re-appears without a populated view model

ashfurrow:
I was looking through Sentry and found this crash: https://sentry.io/organizations/artsynet/issues/739658774 I'd seen the crash before, but the solution just kind of clicked so I thought I'd get a fix in quickly. 

The problem is that the `viewWillAppear` callback assumes that the `saleViewModel` is already loaded _if_ the view is re-appearing. If the network is slow, this might not be the case, so I've added a guard to check that the `saleViewModel` is non-nil before trying to access any of its properties. (Note: this would normally be caught by the Swift compiler, but we're using an implicitly-unwrapped optional here. Ideally we'd avoid this and use a view controller containment to deal with the optionality of the view model, but that's a much larger refactor.)

### DIFF
--- a/Artsy/App/SwiftExtensions.swift
+++ b/Artsy/App/SwiftExtensions.swift
@@ -34,6 +34,12 @@ extension Occupiable {
     var isNotEmpty: Bool { return !isEmpty }
 }
 
+extension Optional: Occupiable {
+    var isEmpty: Bool {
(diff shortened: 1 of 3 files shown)


ds300:
#mergeOnGreen

PR #2903, opened 2019-09-13
Fixes show the new RN artwork view for sale artworks (that are inquirable) #trivial

ashfurrow:
@lilyfromseattle and I found this today – basically if an artwork is inquirable _and_ in an auction, we want to show the old view.


sweir27:
#mergeongreen

PR #2910, opened 2019-09-24
Updates setup instructions

ashfurrow:
Our setup instructions are quite out of date, including references to Spacecommander and Swiftlint that we simply don't use in practice. Maybe once the Mobile Experience team is up and running, we can look into automated formatters again. Having reliable setup instructions in our README is an important step to making sure engineers are comfortable contributing to our iOS software.


alloy:
Agreed, thanks 👍

PR #2912, opened 2019-09-25
[ME-3] Fixes opening Terms/Privacy Policy links during onboarding

ashfurrow:
At one point, we supported customizing the navigation bar during onboarding to selectively show a back button, but that stopped working at some point (likely when we added support for the iPhone X notch). Now, when users tap the Terms of Service or Privacy Policy links, they are shown a web view that they can't back out of; the only solution is to force close the app and start the onboarding process again (a terrible UX).

This PR changes the behaviour of these links; users will now be shown the ToS or PP in Safari, rather than the app. This isn't a _great_ UX either, since it takes the user out of the app, but is _significantly_ less work than fixing a navigation bar that is _only_ used during onboarding for these specific links (which we don't expect will get tapped often, anyway). I'm going to confirm with @samchieng that this changes makes sense before merging.

### DIFF
--- a/Artsy/View_Controllers/Login_and_Onboarding/AROnboardingViewController.m
+++ b/Artsy/View_Controllers/Login_and_Onboarding/AROnboardingViewController.m
@@ -244,14 +244,12 @@
 
 - (void)showTermsAndConditions
 {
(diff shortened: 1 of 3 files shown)


ashfurrow:
Okay! Chatted with Sam and we're 👍 on this change.

PR #2917, opened 2019-10-02
Dependency cleanup #trivial

ashfurrow:
This PR does two things:

- Removes `cocoapods-fix-react-native`, to avoid the warning of unsupported version at the end of `pod install`. We can add it back in (and PR any changes to the plugin) if we need to, but since CocoaPods is officially supported by React Native now, I doubt we'll need to.
- Updates Nokogiri to address a security vulnerability. It's not applicable to our project, since we only use it for local development kinda, but I'm tired of those GitHub warnings.


ashfurrow:
@sepans hit that "merge" button and you'll have a commit landed in the repo 🚀

alloy:
@ashfurrow FYI You can mark those security warnings as “not applicable”.

ashfurrow:
Mmmm good point, I'll do that next time 👍

PR #2922, opened 2019-10-04
[PURCHASE-1537] Fix search transition animation

ashfurrow:
This reverts the change from #2904 because it introduced the problem for PURCHASE-1537. We re-opened ticket 1482 (the extra white padding above the artwork page) and moved it to Mobile Experience's backlog. It's a relatively minor bug compared to what this fixes.

We also tried to clean up code indentation and commenting, I'll comment on the import changes below.

### DIFF
--- a/Artsy/Navigation_Transitions/ARAppSearchTransition.m
+++ b/Artsy/Navigation_Transitions/ARAppSearchTransition.m
@@ -1,8 +1,6 @@
 #import "ARAppSearchTransition.h"
 #import "ARAppSearchViewController.h"
-#import "ARTopMenuViewController.h"
 
-#import <FLKAutoLayout/FLKAutoLayout.h>
 
 @implementation ARAppSearchTransition
 
@@ -24,16 +22,15 @@
 
     [transitionContext.containerView addSubview:fromVC.view];
     [transitionContext.containerView addSubview:toVC.view];
-    [toVC.view alignToView:transitionContext.containerView];
 
     [UIView animateWithDuration:[self transitionDuration:transitionContext]
         delay:0.0
         options:UIViewAnimationOptionCurveEaseOut
         animations:^{
-            toVC.view.alpha = 1;
+                         toVC.view.alpha = 1;
         }
         completion:^(BOOL finished) {
-            [transitionContext completeTransition:YES];
+                        [transitionContext completeTransition:YES];
         }];
 }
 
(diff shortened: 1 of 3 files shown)


Artsy/Navigation_Transitions/ARAppSearchTransition.m
```diff
@@ -19,18 +17,23 @@ - (void)pushTransitionFrom:(UIViewController *)fromVC
     UIGraphicsEndImageContext();
 
     if ([toVC isKindOfClass:[ARAppSearchViewController class]]) {
+        // Showing search
         toVC.backgroundImage = viewImage;
+    } else {
+        // Selecting search result
+        toVC.view.frame = [transitionContext finalFrameForViewController:toVC];
+        toVC.view.transform = CGAffineTransformMakeTranslation(toVC.view.frame.size.width, 0); // Push it offscreen
```
ashfurrow: Search transition wasn't working well on iOS 10 so I added an explicit `frame` here. I'm not really sure how this was ever working, to be honest, but it works now 👍

Artsy/Navigation_Transitions/ARDefaultNavigationTransition.m
```diff
@@ -97,33 +97,34 @@ - (void)popTransitionFrom:(UIViewController *)fromVC to:(UIViewController *)toVC
     }
 
     toVC.view.alpha = 0.2;
+    toVC.view.frame = [context finalFrameForViewController:toVC];
```
ashfurrow: This is the key change, it makes the pop from the view controller (presented through search) work.

ashfurrow:
Ah! Of course. Sorry for not including that on Friday.

So the bug was when you selected an auction from search, tapped into a sale artwork, and then tapped back, the auction view would be blank (specifically, it had `CGRectZero` geometry). The last PR (#2904) added some autolayout constraints to the transition animation, but apparently that's not the recommended approach, so I think it introduced this bug. Here's what it looked like:

bad gifcask 2019-10-07 09_23_11

The solution was to remove the autolayout code, accept the minor UI bug that #2904 fixed, and shore up the code a bit so it would be easier to read+maintain going forward. Here's what it looks like with this PR:

fixed gifcask 2019-10-07 09_20_55

alloy:
Hah, is the search -> vc animation bug, where it animated from the top corner, now mirrored? (It feels like it grows from top-left in first GIF and top-right in second GIF.) A bug in either direction btw, so not a blocker.

ashfurrow:
@alloy ha, yes, though that's unrelated to this fix. _That_ bug is intermittent and I think depends on the iOS version. I've tried to fix it before, but have come away frustrated. Someday soon!

PR #2925, opened 2019-10-07
ME-43 - Adds killSwitch

kierangillen:
- Adds killswitch based on Eigen version
- https://artsyproduct.atlassian.net/browse/ME-43

Screen Shot 2019-10-08 at 1 41 18 PM

### DIFF
--- a/Artsy/App/ARAppDelegate.m
+++ b/Artsy/App/ARAppDelegate.m
@@ -42,6 +42,7 @@
 #import "ARAuthValidator.h"
 
 #import "UIDevice-Hardware.h"
+#import "ArtsyEcho.h"
 
 #import <Keys/ArtsyKeys.h>
 #import "AREndOfLineInternalMobileWebViewController.h"
@@ -212,6 +213,9 @@ static ARAppDelegate *_sharedInstance = nil;
 
     [ARWebViewCacheHost startup];
     [self registerNewSessionOpened];
+    ar_dispatch_after(1, ^{
+        [self killSwitch];
+    });
 }
 
 /// This is called when the app is almost done launching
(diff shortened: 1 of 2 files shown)


Artsy/App/ARAppDelegate.m
```diff
@@ -212,6 +213,9 @@ - (void)setupForAppLaunch
 
     [ARWebViewCacheHost startup];
     [self registerNewSessionOpened];
+    ar_dispatch_after(1, ^{
```
ashfurrow:
```suggestion
    // We wait 1 second to make sure the view controller hierarchy has been set up.
    ar_dispatch_after(1, ^{
```

Artsy/App/ARAppDelegate.m
```diff
@@ -264,6 +268,38 @@ - (void)showOnboarding;
     [self showOnboardingWithState:ARInitialOnboardingStateSlideShow];
 }
 
+- (void)killSwitch;
+{
+    Message *killSwitchVersion = ARSwitchBoard.sharedInstance.echo.messages[@"KillSwitchBuildMinimum"];
+    if(killSwitchVersion.content != nil){
+        NSMutableDictionary *infoDictionary = [[[NSBundle mainBundle] infoDictionary] mutableCopy];
```
ashfurrow: Since we're not mutating this dictionary, we don't need the `mutableCopy`. 

```suggestion
        NSDictionary *infoDictionary = [[NSBundle mainBundle] infoDictionary];
```

Artsy/App/ARAppDelegate.m
```diff
@@ -264,6 +268,38 @@ - (void)showOnboarding;
...
+        NSString *buildVersion = infoDictionary[@"CFBundleShortVersionString"];
+        NSString *echoMinimumBuild = killSwitchVersion.content;
```
ashfurrow: Since we're comparing `killSwitchVersion.content` in the `if` statement, we could move this line outside the `if` statement and replace it with `if (echoMinimumBuild != nil)` (or even, to be more succinct, `if (echoMinimumBuild)`).

Artsy/App/ARAppDelegate.m
```diff
@@ -264,6 +268,38 @@ - (void)showOnboarding;
...
+        
+        if ([buildVersion compare:echoMinimumBuild options:NSNumericSearch] == NSOrderedDescending) {
+            UIAlertController * alert = [UIAlertController
```
ashfurrow: It sucks that we don't have a code formatter for this repo, but generally in Objective-C the `*` goes right next to the variable name.

```suggestion
            UIAlertController *alert = [UIAlertController
```

Artsy/App/ARAppDelegate.m
```diff
@@ -264,6 +268,38 @@ - (void)showOnboarding;
...
+                                                  [[UIApplication sharedApplication] openURL:[NSURL URLWithString:iTunesLink]];
+                                                  ar_dispatch_after(1, ^{
+                                                      exit(0);
+                                                  });
+                                              }];
+            
+            [alert addAction:linkToAppButton];
+            
+            [self.window.rootViewController presentViewController:alert animated:YES completion:nil];
+            return;
```
ashfurrow: The `return` here is unnecessary.

```suggestion
```

Artsy/App/ARAppDelegate.m
kierangillen: Hm, I was seeing a crash if `killSwitchVersion` is `null` though if we assume its available and try to use `killSwitchVersion.content`, does your suggest prevent that?

kierangillen:
@ashfurrow Okay this has been updated with changes regarding your comments and also with updated copy from @samchieng

ashfurrow:
#mergeongreen

PR #2927, opened 2019-10-09
Prepares for 6.0.0 betas

ashfurrow:
Using `make update_bundle_version`.


kierangillen:
Are we waiting for the Auctions view of the new artwork view before release `6.0.0`?

ashfurrow:
I'll follow up with Purchase to see if they're ready, but I _think_ the showstoppers are mostly in. We can merge this for now, which will unblock Purchase to make betas 👍 Otherwise, we can release whenever it makes sense for Mobile Experience.

PR #2931, opened 2019-10-16
ME-15 Fix ARSerifNavigationBar element positions

ds300:
https://artsyproduct.atlassian.net/browse/ME-15

## Before

bad-nav-bar

## After

good-nav-bar


Artsy/View_Controllers/ARSerifNavigationViewController.m
```diff
@@ -274,21 +284,24 @@ - (void)verticallyCenterView:(id)viewOrArray
...
-    CGFloat barMidpoint = roundf(self.frame.size.height / 2);
-    CGFloat viewMidpoint = roundf(viewToCenter.frame.size.height / 2);
+    CGFloat yOffset = 0;
+    if ([viewToCenter isDescendantOfView:self]) {
+        UIView *child = viewToCenter.superview;
+        int safety = 0;
+        while (safety++ < 10 && child != self) {
+            yOffset += child.frame.origin.y;
+            child = child.superview;
+        }
```
ashfurrow: It seems like what this loop is trying to find the vertical difference in frames between `self` and `viewToCenter`. We can use a few methods in `UIView` to accomplish this as well:

```objc
// Grab the bounds, which is viewToCenter's CGRect within its own coordinate space, and covert to self's space
CGRect viewToCenterFrameInSelf = [viewToCenter convertRect:viewToCenter.bounds toCoordinateSpace:self];
CGFloat yOffset = viewToCenterFrameInSelf.origin.y;
```

I _think_ that should do it.

alloy:
> It's not an ideal workflow, maybe there's a reason we haven't documented it yet 🙈 I'm very open to suggestions here. /cc @alloy

I don’t know of another solution. I suspect not much time was put into it because the need for it doesn’t arise often.

Artsy/View_Controllers/ARSerifNavigationViewController.m
ds300: Hmm this is not working. Looks like this:

image

Also tried converting the bounds to window space before calculating the offset, same result. 🤔 

Any other approaches you think might work? I'm not crazy about this while look either.

ds300: could it be that the height of the text label is scaled down somewhere and needs to be converted too?

ds300: it could not

ashfurrow: Mmm, [`locationInView:`](https://developer.apple.com/documentation/uikit/uitouch/1618116-locationinview?language=objc) might be helpful here? Agreed the loop isn't ideal, but if it works (and has `safety`!) then I'm okay shipping it. Maybe add a comment explaining what we tried instead, for posterity?

ds300: Unless I missed something, locationInView is a method of UITouch, which I don't have an instance of?

I think let's stick with the while loop. I added a comment, thanks for the suggestion :+1:

ds300:
This should be 🍏

PR #2934, opened 2019-10-25
Add a feature flag for Price Transparency

yuki24:
addresses https://artsyproduct.atlassian.net/browse/AUCT-697

This adds a new feature flag `AROptionsPriceTransparency` to Eigen so we can toggle the entire feature either globally or per admin basis. I've already [added a new flag that has the consistent name in Echo](https://echo-web-production.herokuapp.com/accounts/1/features/13), so in theory it should work fine. Though this is my first Eigen PR for the first time in a while, so let me know if there's anything wrong.

## Todos

 * [x] Merge https://github.com/artsy/emission/pull/1936
 * [x] ~Make sure Echo's feature flag is properly loaded?~ We will do this in a separate PR: https://github.com/artsy/eigen/pull/2937

### DIFF
--- a/Artsy/App/ARAppDelegate+Emission.m
+++ b/Artsy/App/ARAppDelegate+Emission.m
@@ -309,6 +309,11 @@ FollowRequestFailure(RCTResponseSenderBlock block, BOOL following, NSError *erro
 
     // Lab options come last (as they are admin/dev controlled, giving them a chance to override)
     [options addEntriesFromDictionary:labOptions];
+
+    if (labOptions[AROptionsPriceTransparency] != nil) {
+        options[@"AROptionsPriceTransparency"] = labOptions[AROptionsPriceTransparency];
+    }
+
     return options;
 }
 
--- a/Artsy/App/ARAppDelegate+Emission.m
+++ b/Artsy/App/ARAppDelegate+Emission.m
@@ -310,7 +310,7 @@ FollowRequestFailure(RCTResponseSenderBlock block, BOOL following, NSError *erro
     // Lab options come last (as they are admin/dev controlled, giving them a chance to override)
     [options addEntriesFromDictionary:labOptions];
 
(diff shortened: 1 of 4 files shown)


Artsy/App/ARAppDelegate+Emission.m
```diff
@@ -309,6 +309,11 @@ - (NSDictionary *)getOptionsForEmission:(NSDictionary *)echoFeatures labOptions:
 
     // Lab options come last (as they are admin/dev controlled, giving them a chance to override)
     [options addEntriesFromDictionary:labOptions];
+
+    if (labOptions[AROptionsPriceTransparency] != nil) {
```
yuki24: I assumed it is important to check the existence of the key to avoid overriding the option by accident (e.g. if `labOptions[AROptionsPriceTransparency]` is not set at all, the line below could end up assigning `nil` all the time regardless of what is set in Echo), but it might be an unnecessary guard.

Artsy/App/AROptions.m
```diff
@@ -24,6 +24,7 @@
 NSString *const AROptionsRNArtworkNonCommerical = @"New RN Artwork view (Non-commerical)";
 NSString *const AROptionsRNArtworkNSOInquiry = @"New RN Artwork view (NSO&Inquiry)";
 NSString *const AROptionsRNArtworkAuctions = @"New RN Artwork view (Auctions)";
+NSString *const AROptionsPriceTransparency = @"Price Transpareny";
```
erikdstock: Typo- Transparen**c**y

ashfurrow:
We talked about splitting this into two PRs:

- One has the lab options, which we can merge now.
- The other has the Echo option, which we will only merge once we want users to see it.

Echo will act in this case as a safety that we can use to switch off the feature if we have to, post-release.

yuki24:
I have removed the Echo flag both from Eigen https://github.com/artsy/eigen/pull/2934/commits/452d08dee7b6a99abb9d33c97c756cbaa2fcac90 and Echo's admin UI. I'll send a separate PR once this is merged.

PR #2937, opened 2019-10-28
⚠️ Do not merge: Add a new flag for Price Transparency to Echo

yuki24:
Following up on https://github.com/artsy/eigen/pull/2934#issuecomment-547111199. This update Eigen to also look at the Echo flag for the new Price Transparecy as well as the lab feature flag.

## Todos

 * [x] Rebase once https://github.com/artsy/eigen/pull/2934 is merged
 * [x] **Do not do this until the feature is ready**: Add `AROptionsPriceTransparency` to [Echo](https://echo-web-production.herokuapp.com/accounts/1/features)

### DIFF
--- a/Artsy/App/ARAppDelegate+Emission.m
+++ b/Artsy/App/ARAppDelegate+Emission.m
@@ -307,7 +307,7 @@ FollowRequestFailure(RCTResponseSenderBlock block, BOOL following, NSError *erro
     // Lab options come last (as they are admin/dev controlled, giving them a chance to override)
     [options addEntriesFromDictionary:labOptions];
 
-    options[@"AROptionsPriceTransparency"] = @([labOptions[AROptionsPriceTransparency] boolValue]);
+    options[@"AROptionsPriceTransparency"] = @([options[@"AROptionsPriceTransparency"] boolValue] || [labOptions[AROptionsPriceTransparency] boolValue]);
 
     return options;
 }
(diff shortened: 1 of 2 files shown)


Artsy/App/ARAppDelegate+Emission.m
```diff
@@ -307,7 +307,7 @@ - (NSDictionary *)getOptionsForEmission:(NSDictionary *)echoFeatures labOptions:
     // Lab options come last (as they are admin/dev controlled, giving them a chance to override)
     [options addEntriesFromDictionary:labOptions];
 
-    options[@"AROptionsPriceTransparency"] = @([labOptions[AROptionsPriceTransparency] boolValue]);
+    options[@"AROptionsPriceTransparency"] = @([options[@"AROptionsPriceTransparency"] boolValue] || [labOptions[AROptionsPriceTransparency] boolValue]);
```
dleve123: @yuki24 I know that we discussed this a bit IRL, but can we confirm the order of boolean precedence here? Assuming this is Ruby-ish, then a truthy value for `AROptionsPriceTransparency` will _always_ override the value of the lab feature. Is that the case here?

yuki24: > a truthy value for AROptionsPriceTransparency will always override the value of the lab feature

Correct.

yuki24:
#mergeongreen

PR #2942, opened 2019-11-01
[ME-1] Fixes ARKit point cloud bug

ashfurrow:
I've left a comment explaining the change and how the fix works. Here's the tl;dr

- We use ARKit's `rawFeaturePoints` (which is intended for debugging) to show the user that we're loading the ARKit scene (ie: that ARKit is mapping the world).
- In iOS 13, this "point cloud" started including points that were really close to the camera. [I asked around](https://twitter.com/ashfurrow/status/1189910896920989700) but no one else has been seeing this behaviour. Here's what it looked like:

<details><summary>screenshot of ARKit session start</summary>

Unknown

</details>

- The docs say that this point cloud isn't stable between versions, so I think iOS is just adding extra points really early in the AR session (they go away after a few seconds, by which time we no longer want to show the point cloud anyway).
- The best solution is to compute the distance from each point to the camera and filter out any that are <20cm away. There are still a few points that seem further from the floor than they should be, but they're far enough away from the camera to not look bad or obscure the actual point cloud.

Open to feedback on this!

### DIFF
--- a/Artsy/View_Controllers/ARVIR/AR/ARVIRHorizontalPlaneInteractionController.m
+++ b/Artsy/View_Controllers/ARVIR/AR/ARVIRHorizontalPlaneInteractionController.m
@@ -300,7 +300,20 @@ NSInteger attempt = 0;
             SCNNode *pointNode = [SCNNode nodeWithGeometry:pointSphere];
             pointNode.position = vector;
 
-            [self.pointCloudNode addChildNode:pointNode];
+            // Grab the 4th column of the transform, which indicates its scale factor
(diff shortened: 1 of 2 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>The CHANGELOG is not valid YML</td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/swift">Danger Swift</a> against 37e5975f8012b7bc9ef0c1131063e096e01a6780
</p>

PR #2962, opened 2019-12-02
[MX-113] React Native upgrade

ashfurrow:
This is the companion PR to https://github.com/artsy/emission/pull/1970

Not sure if it'll pass on CI right away, since the CI machines will require Yarn for the `pod install` process. But we'll see!


ashfurrow:
Oh weird! Same build failures as #2889. Once we fix these, hopefully we can merge that too!

PR #2969, opened 2019-12-12
[MX-127] Hide tool bar when in search nav stack

ds300:
https://artsyproduct.atlassian.net/browse/MX-127

Now that search is it's own page we had a problem in deciding _where_ tapping a search result takes you in our nav system.

This PR implements the decision to copy the twitter app by keeping all search-related activity in the search page. It accomplishes this by hiding the bottom toolbar nav when viewing search results.

Here's how that looks.

Kapture 2019-12-12 at 14 37 56

Where I've put this logic feels like the wrong place conceptually but also the simplest place practically. I'm curious if you're concerned about that @ashfurrow or maybe you have a suggestion for where else this could go?

### DIFF
--- a/Artsy/View_Controllers/Util/ARNavigationController.m
+++ b/Artsy/View_Controllers/Util/ARNavigationController.m
@@ -16,6 +16,7 @@
 #import "ARTopMenuViewController.h"
 #import "ARScrollNavigationChief.h"
 #import "AREigenMapContainerViewController.h"
+#import "AROptions.h"
 
 #import "ARMacros.h"
 #import "UIDevice-Hardware.h"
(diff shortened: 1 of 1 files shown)


Artsy/View_Controllers/Util/ARNavigationController.m
```diff
@@ -393,7 +394,17 @@ - (BOOL)shouldShowStatusBarBackgroundForViewController:(UIViewController *)viewC
 
 - (BOOL)shouldHideToolbarMenuForViewController:(UIViewController *)viewController
 {
-    return ShouldHideItem(viewController, @selector(hidesToolbarMenu), NULL);
+  if (ShouldHideItem(viewController, @selector(hidesToolbarMenu), NULL)) {
+    return YES;
+  }
+  
+  // also hide if in search view and other VCs are pushed above root search view
+  for (int i = 0; i < self.viewControllers.count; i++) {
```
ashfurrow: I think the search view is the first view controller on the stack, isn't it? So this could be replaced with:

```objc
return [self.viewControllers.firstObject isKindOfClass:ARSearchComponentViewController.class] && self.viewControllers.count > 1;
```

I _think_.

ds300:
> That said, other apps do this, so there's probably a good reason that I'm not seeing. Could you fill me in on the advantages that this interaction gives us/the user?

I think the main advantage over other options is that it leaves it up to the user to be in control of their rabbit hole. Otherwise we have to either clear it on their behalf when they leave the search page, which is not easy to get right, or let it get in their way when they come back.

ashfurrow:
Okay, that makes sense. Let's try this – I think it'll feel right once I use it on-device. My only other recommendation is to put the logic behind the lab option, similar to here:

https://github.com/artsy/eigen/blob/7716765ee8f0ea5e362cfd29ab8fbaa636f7cc96/Artsy/View_Controllers/App_Navigation/ARTopMenuNavigationDataSource.m#L152

ds300:
Good suggestions, thanks! 🍏

ds300:
turns out that twitter only does this on android, not iOS, which leaves me wondering whether it is indeed the best thing or whether it even matters much. 🤷‍♂

ashfurrow:
Let's try it out and see! We can highlight this with Sam+the rest of the team to get feedback.

PR #2985, opened 2020-01-13
Moves back to old-style search nav bar behaviour #trivial

ashfurrow:
Sam, Brittney and I were discussing the behaviour of hiding the bottom tab bar when tapping in a search. In the abstract, the idea made sense, but seeing it play out on-device felt weird. If a user is searching, they might want to tap from a search result back to their Home (or Conversations, or Favourites, etc) and then quickly back to their search result again (see the below GIF).

<details><summary>GIF of the original (now new again) behaviour</summary>

</details> 

This PR changes back the behaviour introduced #2969. I think it was worth trying out, and we learned from it 👍

### DIFF
--- a/Artsy/View_Controllers/Util/ARNavigationController.m
+++ b/Artsy/View_Controllers/Util/ARNavigationController.m
@@ -16,7 +16,6 @@
 #import "ARTopMenuViewController.h"
 #import "ARScrollNavigationChief.h"
 #import "AREigenMapContainerViewController.h"
-#import "AROptions.h"
 
 #import "ARMacros.h"
 #import "UIDevice-Hardware.h"
@@ -29,7 +28,6 @@
 #import <MultiDelegate/AIMultiDelegate.h>
 #import <Emission/ARMapContainerViewController.h>
 #import <Emission/ARComponentViewController.h>
-#import <Emission/ARSearchComponentViewController.h>
 
 static void *ARNavigationControllerButtonStateContext = &ARNavigationControllerButtonStateContext;
 static void *ARNavigationControllerScrollingChiefContext = &ARNavigationControllerScrollingChiefContext;
@@ -395,12 +393,7 @@ ShouldHideItem(UIViewController *viewController, SEL itemSelector, ...)
 
 - (BOOL)shouldHideToolbarMenuForViewController:(UIViewController *)viewController
 {
-  if (ShouldHideItem(viewController, @selector(hidesToolbarMenu), NULL)) {
-      return YES;
-  }
-
-  // also hide if in search view and other VCs are pushed above root search view
-  return [AROptions boolForOption:AROptionsNewSearch] && [self.viewControllers.firstObject isKindOfClass:ARSearchComponentViewController.class] && self.viewControllers.count > 1;
+    return ShouldHideItem(viewController, @selector(hidesToolbarMenu), NULL);
 }
 
 


ds300:
I'm personally not comfortable switching back to this behaviour until we come up with a strategy to avoid the situation where

1. a user is deep in a search rabbit hole
2. they tab to home or saves and lose their mental search context
3. they decide they want to search again
4. they switch back to the search tab only to discover that they first have to tap the back button a dozen times
5. after which do we show the keyboard automatically or make them tap into it?

This is IMO much worse UX than being in a search rabbit hole and having to tap back a dozen times to get 'back to the main app' .

Was this conversation had? I think we'd need to do some version of 'clear the search navigation stack on the user's behalf when they don't need it any more' which could be tricky.

ashfurrow:
That’s a good point! Let me lay out my thinking. By convention, the app’s tab bar lets users pop all the way back to the root view by tapping a tab button while it’s already selected. This is quite nice, since a user who taps the Search tab and is confused by not seeing search (because it’s at the bottom of a navigation stack) will probably tap the Search tab icon again, triggering the pop back to Search. 

I think that you’re right about bringing up the keyboard, that this behaviour would need to take different user paths into account. While it might be helpful to show a keyboard when first tapping Search, returning to a list of Search results and having the keyboard animate in and obscure those results (and the tab bar) isn’t good. We could implement the text input to take focus only in the situation where the user has opened the Search tab and the text input is empty, or maybe just the first time it’s made visible.

What do you think?

ds300:
Yeah I like that :+1: The 'tap search again to clear the nav stack' pattern never occurred to me and it's not something I've used consciously in other apps but I think you're right that users could figure it out naturally.

Also this morning I've been wondering whether it would be simpler for users if we just clear the nav stack whenever they leave/tap into search? I mean, is this deliberate switching-back-and-forth behaviour something we should be optimising for? My thinking is we could keep the last 1-2 things the user has _navigated to_ at the top of the 'Recent' list on the search page, so if they want to switch back-and-forth, or continue a previous rabbit hole, it's only an extra tap. The big benefit of this approach is that we maintain a consistent behaviour for the search button (+ keyboard) at all times, and searching is definitely a behaviour we want to optimise for!

ashfurrow:
Hmm! That's an interesting idea – my concern with altering the nav stack when tapping away from the Search tab is that it would break users' expectations. If I'm on the NYC subway where there's limited LTE for example, I would expect to be able to go from one tab to another and back without needing to re-fetch anything from the network.

I re-read the [Tab Bars section of the iOS Human Interface Guidelines](https://developer.apple.com/design/human-interface-guidelines/ios/bars/tab-bars/) to refresh my understanding of what tab bars are supposed to be and how they should behave. Since we're moving away from the pseudo-modal Search tab we have now (which _definitely_ doesn't adhere to the HIG or user expectations), I think we should be really deliberate about any deviations from those guidelines/expectations. 

Let me know what you think – I think the best way to make these decisions sometimes is to prototype an interaction and play with it on a device.

ds300:
Hmm the only app I could find that clears the search nav stack when you tab away is Spotify. All the other big apps with a search tab keep the nav stack until you close the app or dismiss it manually.

I'm happy to merge this if we include the change to clear the nav stack if the search tab button is tapped when the search tab is already open. Would the right place to do that be in here: https://github.com/artsy/eigen/blob/d8d7c83ed605b9415e6031a7769c21aac465c69c/Artsy/Views/Utilities/ARTabContentView.m#L148 ?

ashfurrow:
@ds300 good call looking to other apps for cues here 👍 

Eigen already adheres to the convention of tapping a tab bar button while it's select to navigate to its root view controller, though I had to do some digging to find the code that's responsible:

https://github.com/artsy/eigen/blob/d8d7c83ed605b9415e6031a7769c21aac465c69c/Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m#L726-L735

Here's how the code in this PR behaves when the user drills down a bit from a search result, and then taps the Search icon:

PR #2987, opened 2020-01-16
Change default collection + upgrade emission

ashleyjelks:
Changes the default collection slug + upgrades to Emission v 1.21.8

#trivial


ashleyjelks:
#mergeongreen

PR #2991, opened 2020-01-17
Enables url schemes for Collections routes

ashleyjelks:
Submitting this PR as [a part of this spike](https://artsyproduct.atlassian.net/browse/FX-1600) for testing iOS Collections.

Kapture 2020-01-17 at 16 31 07

#trivial


ashfurrow:
Cool! You can now also use Emission's `SwitchBoard` module to use this routing, too! eg:

```tsx
SwitchBoard.presentNavigationViewController(this, "/collection/collection-id")
```

This is how we'll eventually add entry points from within the app 👍

PR #2994, opened 2020-01-23
[MX-151] clear search history on log out

ds300:
https://artsyproduct.atlassian.net/browse/MX-151

This covers both admin logout and user logout. Do you think it's worth adding a test for this @ashfurrow ? I couldn't figure out how to run the tests for `ARUserManager` locally.

### DIFF
--- a/Artsy/Networking/API_Modules/ARUserManager.m
+++ b/Artsy/Networking/API_Modules/ARUserManager.m
@@ -28,6 +28,7 @@
 #import <Emission/AREmission.h>
 #import <Emission/ARGraphQLQueryCache.h>
 
+#import "RNCAsyncStorage.h"
 
 NSString *const ARUserSessionStartedNotification = @"ARUserSessionStarted";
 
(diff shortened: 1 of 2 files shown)


Artsy/Networking/API_Modules/ARUserManager.m
```diff
@@ -472,9 +473,14 @@ + (void)clearUserData:(ARUserManager *)manager useStaging:(id)useStaging
     [manager deleteHTTPCookies];
     [ARRouter setAuthToken:nil];
     manager.currentUser = nil;
-    
+
     [[[AREmission sharedInstance] graphQLQueryCacheModule] clearAll];
 
+    RNCAsyncStorage *asyncStorage = [[[AREmission sharedInstance] bridge] moduleForName:@"RNCAsyncStorage"];
+    if (asyncStorage) {
```
ashfurrow: As a quirk of Objective-C, the `if` statement here won't actually affect behaviour. if `asyncStorage` _is_ `nil`, then that's actually fine because Objective-C allows sending messages to `nil` (the return value is always `nil`). This is totally fine, just FYI!

ds300: Wow, TIL. That seems like a dangerous language design choice to me! 🙈is it notorious for causing hard-to-track-down bugs?

ds300:
Thanks for the super helpful testing instructions! I was able to run the tests. I think I'm with you being on the fence about whether it's needed here though, so will err on the side of less work and merge already.

#mergeOnGreen

Artsy/Networking/API_Modules/ARUserManager.m
ashfurrow: > is it notorious for causing hard-to-track-down bugs?

Kind of! Objective-C's weird relationship with `nil` goes a lot deeper, since `nil` can't be included in `NSArray` or `NSDictionary` instances. So sometimes you need to use the `[NSNull null]` singleton to represent a missing value, and calling unknown methods on `NSNull` _will_ definitely cause problems.

I think of it this way: early Objective-C was a wild west, with a Scheme-esque message-passing approach to method calls. But Objective-C 2.0 (circa 2008, I think?) introduced a lot more type safety. And Swift added even more, with features like generics and nullable types being ported back from Swift into Objective-C. What we end up with is a weird mismatch of different approaches, and no one is really that happy with it, haha.

(Well, I'm happy with it 😅 Objective-C feels like a comfy blanket to me, but only because I know its quirks.)

PR #2996, opened 2020-01-24
[MX-154] - Fix auction banner display on iOS 13

brainbicycle:
It seems like we are reliant on some behavior that changed in iOS 13 with regards to [traitCollectionDidChange](https://developer.apple.com/documentation/uikit/uitraitenvironment/1623516-traitcollectiondidchange) in some of the auction code. [This blog post](https://useyourloaf.com/blog/predicting-size-classes-in-ios-13/) has a good explanation but the tldr; is that traitCollectionDidChange is not guaranteed to be called when a view is added to the hierarchy after iOS 13 so we need to explicitly set up views wherever we were reliant on this.


ashfurrow:
(I cancelled the redundant build for 63e3499.)

PR #2997, opened 2020-01-24
[MX-152] Opts out of new iOS 13 modal presentation style on iPhone

ashfurrow:
The underlying issue is that React Native's `KeyboardAvoidingView` isn't yet taking into account the additional space added by the iOS 13 modal presentation style ([there is an open pull request](https://github.com/facebook/react-native/pull/27607)). We could adopt this approach, but I'd rather not duplicate logic that will be included in the view itself pretty soon.

Instead, I read more about the transition changes ([this blog post was really helpful](https://medium.com/@hacknicity/view-controller-presentation-changes-in-ios-13-ac8c901ebc4e)) and have opted out of the new transition style on iPhone only. Here's a before-and-after of the behaviour on iOS 13 devices in the different contexts that we present modal view controllers (not including uncommonly-seen views, like the Privacy Policy or Conditions of Sale).

This fixes the bug, though we had to modify the iPad consignments flow to be full screen for now as well (as far as I can tell, the bug was already present in this flow). I don't love all the `isKindOfClass:` code we have here, but until we move routing into Emission, I don't want to rock the boat too much with how this is structured. 

I will also follow-up with an Emission PR that fixes the status bar contents not being legible on the consignments flow, which fell out of https://github.com/artsy/emission/pull/1993

## iPhone

|   | LAI | Bid Flow | Consignments | BNMO  |
|---|---|---|---|---|
| Before |  Old_Phone_LAI |  Old_Phone_Bid Flow | Old_Phone_Consignments  | Old_Phone_BNMO  |
| After | New_Phone_LAI  | New_Phone_Bid Flow  | New_Phone_Consignments  | New_Phone_BNMO  |

## iPad

|   | LAI | Bid Flow | Consignments | BNMO  |
|---|---|---|---|---|
| Before | Old_Pad_LAI | Old_Pad_Bid Flow |  Old_Pad_Consignments |  Old_Pad_BNMO |
| After | New_Pad_LAI  | Screen Shot 2020-01-24 at 13 25 33 | New_Pad_Consignments  |  Screen Shot 2020-01-24 at 13 34 29 |

### DIFF
--- a/Artsy/App/ARAppDelegate+Emission.m
+++ b/Artsy/App/ARAppDelegate+Emission.m
@@ -16,6 +16,7 @@
 #import "ARAugmentedVIRSetupViewController.h"
 #import "ARAugmentedRealityConfig.h"
 #import "ARAugmentedFloorBasedVIRViewController.h"
+#import "ARInternalMobileWebViewController.h"
 #import "ARDefaults.h"
 #import "ARNavigationController.h"
 #import "ARTopMenuViewController.h"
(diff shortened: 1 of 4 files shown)


brainbicycle:
LGTM 👍

PR #3003, opened 2020-01-28
[MX-159] clear APN token on logout

ds300:
This won't work until artsy/gravity#12751 gets deployed, but it can be deployed first without disrupting the current behaviour. Worth noting that I haven't tested the happy path yet! Only the sad path.

https://artsyproduct.atlassian.net/browse/MX-159

### DIFF
--- a/Artsy/Networking/API_Modules/ArtsyAPI+DeviceTokens.m
+++ b/Artsy/Networking/API_Modules/ArtsyAPI+DeviceTokens.m
@@ -2,6 +2,7 @@
 
 #import "ArtsyAPI+Private.h"
 #import "ARRouter.h"
+#import "ARAppConstants.h"
 
 
 @implementation ArtsyAPI (DeviceTokens)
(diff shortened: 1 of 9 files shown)


Artsy/Networking/API_Modules/ArtsyAPI+DeviceTokens.m
```diff
@@ -18,8 +19,23 @@ + (AFHTTPRequestOperation *)setAPNTokenForCurrentDevice:(NSString *)token succes
...
-
+    
     return nil;
 }
 
++ (AFHTTPRequestOperation *)deleteAPNTokenForCurrentDeviceWithCompletion:(void (^)(void))completion
+{
+    NSString *token = [[NSUserDefaults standardUserDefaults] stringForKey:ARAPNSDeviceTokenKey];
+    if (!token) {
+        completion();
```
ashfurrow: Yes! Excellent catch 🙌

PR #3008, opened 2020-02-03
Move token deletion into logout method

ds300:
Tiny followup from #3007, moving the push notification token deletion into the `logout` method so it gets called after session expiration too. Perhaps in gravity we should clear all device tokens associated with a user when they reset their password?

#trivial


ds300:
Sending PNs to devices that aren't logged in feels like leaking secure information to me 🤔admittedly in this case users don't realise that they've been logged out, so I'm not too concerned from a UX perspective.

PR #3013, opened 2020-02-05
Set react packager url automatically

ds300:
When running eigen linked with emission on a device you need to manually edit `Artsy/App/ARAppDelegate+Emission.m` to set the packager url to point to your local machine. This is a pain if, like me, you change wifi networks a lot.

This PR causes the packager URL to be set automatically when in dev mode.

During the build phase we write an obj-c file containing your macbook's hostname. This PR adds that file to the project index but not the git index, and git will ignore it going forward.

I think this is a robust way to do it. Definitely try it out before merging though!

#known

### DIFF
--- a/scripts/set_packager_host.sh
+++ b/scripts/set_packager_host.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+
+OUTPUT=$(cat << EOF
+// this file is generated automatically in a build phase
+#import "ARReactPackagerHost.h"
+#import <Foundation/Foundation.h>
+
+@implementation ARReactPackagerHost
+
++ (NSString *)hostname
+{
+#if defined(DEBUG)
+    return @"$(hostname)";
+#else
+    return @"localhost";
+#endif
+}
+
+@end
+EOF)
+
+echo "$OUTPUT" > Artsy/Networking/ARReactPackagerHost.m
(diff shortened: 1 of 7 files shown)


scripts/set_packager_host.sh
```diff
@@ -0,0 +1,22 @@
...
+    return @"$(hostname)";
+#else
+    return @"localhost";
+#endif
+}
+
+@end
+EOF)
+
+echo "$OUTPUT" > Artsy/Networking/ARReactPackagerHost.m
```
ashfurrow: This path might be giving us trouble on CI?

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>Developer Specific file shouldn't be changed, you can skip by adding #known to the PR body and re-runnning CI</td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/swift">Danger Swift</a> against af90e71e33c971c97e5b9c295f6df5bf43965446
</p>

ds300:
hmm couldn't figure out why, but it seemed like xcodebuild wasn't running the build phase I added. So I moved the script execution to a 'pre-action' on the build step of the Artsy scheme. It did the trick.

PR #3014, opened 2020-02-05
Upgrades Emission to 1.21.35

ashleyjelks:
- Upgrades Emission to 1.21.35
- Updates default collection to "Old Masters"


ashleyjelks:
#mergeongreen

PR #3016, opened 2020-02-06
Adds a lab feature for collections artworks

ashleyjelks:
Adds a lab feature for collections artworks.

https://artsyproduct.atlassian.net/browse/FX-1796

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -1,10 +1,11 @@
 upcoming:
   version: 6.3.0
   date: TBD
-  emission_version: 1.21.32
+  emission_version: 1.21.39
   dev:
     - Fixes full-bleed collections image rendering - ash & MX Knowledge Share Feb 4 2020
     - Adds lab option for lot condition report - brian
+    - Adds lab option for filtering collections artworks - ashley
   user_facing:
     -
 
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -1,4 +1,13 @@
 upcoming:
+  version: 6.3.1
+  date: TBD
+  emission_version: 1.21.39
+  dev:
+    - Adds lab option for filtering collections artworks - ashley
+  user_facing:
+    -
+
+upcoming:
   version: 6.3.0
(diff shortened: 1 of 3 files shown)


CHANGELOG.yml
```diff
@@ -1,3 +1,12 @@
+upcoming:
+  version: 6.3.1
```
ashleyjelks: @brainbicycle I did have a question about Eigen versions, they are automatically generated after each merge like in Emission? The current version of Eigen is 6.3.0 so I assumed the next would be 6.3.1 but please correct me if I am wrong here?

brainbicycle: Hey Ashley good question! You should not have to change this. We do manually change this version number but only after the previous version has been released to app store, see here: https://github.com/artsy/eigen/commit/b7abef23b5ff03e8af092558f6f42769f61af7ed. When you merge to beta branch our script updates the build number to the current date automatically: https://github.com/artsy/eigen/blob/master/Makefile#L138 which allows the beta to be uploaded under the same version number. So in your case (assuming this gets merged today) Version number: 6.3.0 build number: 2020-02-07-<HH>

ashleyjelks: Got it @brainbicycle, thanks for the clarification! I just removed changes to that file, so I think it is ready for merge!

brainbicycle: Thanks Ashley 👍 Sorry we do still want changelog entry for the change just under 6.3.0 e.g.:
```upcoming:
  version: 6.3.0
  date: TBD
  emission_version: 1.21.32
  dev:
    - Fixes full-bleed collections image rendering - ash & MX Knowledge Share Feb 4 2020
    - Adds lab option for lot condition report - brian
    - Adds lab option for filtering collections artworks - ashley
  user_facing:
    -
```
Should have been more clear about that.

brainbicycle: Added the update below 👍 All else LGTM

brainbicycle:
👍 #mergeongreen

PR #3017, opened 2020-02-07
[MX-2] Deletes unused code, cleans up deprecation warnings, removes old Echo flags

ashfurrow:
This pull request is housecleaning for Eigen. Sorry for the huge PR, but since it's mostly deletions, it didn't make sense to me to split it up.

Overall, **the number of native unit tests dropped from 911 to 801**. With far less code, I expect to see significantly improved compile times locally for developers, as well as on CI. Users will also see a smaller app download size. A win-win-win 🎉 

Here's a high-level summary:

- Removes a lot of classes that weren't being used anymore (examples: old Search, old Artwork utility views, old (pre-2016) home feed views, quicksilver integration for devs to navigate around that was unused, etc)
- Cleans up Echo flags, see [MX-2](https://artsyproduct.atlassian.net/browse/MX-2). This includes the Search flag – it's been three weeks since launch and if something _did_ go wrong at this point, I'd expect we'd submit an expedited review request rather than go back to the old native search.
- Cleans up _some_ deprecation warnings, see [MX-119](https://artsyproduct.atlassian.net/browse/MX-119) (the remaining ones will be tough). I ignored  a few, since that logic will be moved into React Native soon enough anyway (and migrating to the new APIs was nontrivial).


ashfurrow:
Nice, this shaved more than a minute off of Eigen's CI time!

PR #3021, opened 2020-02-12
[MX-165] - Fix Bid flow close button alignment

brainbicycle:
https://artsyproduct.atlassian.net/browse/MX-165

- Uses safe area layout guide rather than device specific check for iPhone X
- Adds some more padding to be consistent with other close buttons

**iPhone 11 Pro Max after fix**:

iPhone11ProMaxExit

**iPad Pro 11 inch after fix**:

iPadPro11in

**iPhone 5s after fix**:

image

Filed a bug for the pointy corner on iPad (see screenshot above top right corner) think this is on Emission side: https://artsyproduct.atlassian.net/browse/MX-189


brainbicycle:
> 🙏 Looks great, thanks!
> 
> Would it break consistency with other views if we push the button down another pixel or two? Would be nice to align the x with the title text.

Thanks! I can push it down a bit, to be honest from what I have seen padding for close buttons are not consistent in a few places in app probably warrants a sweep through and clean up at some point.

> 
> Also I noticed that the top right corner of the modal is not rounded on iPad 🙈 It's unrelated to this PR, but seems like the kind of thing that might take only a minute or two to fix. What do you think? Otherwise we can add a ticket to the backlog.

Yep I noticed that when I was putting in my screenshots, I looked into it briefly but seems to be on the Emission side so created a ticket for it here: https://artsyproduct.atlassian.net/browse/MX-189

ds300:
cool, let's nudge it down then. Looks like 2px to me but whatever creates the best feeling of alignment with the text. And thanks for adding the ipad JIRA ticket already 🙌 🚀

PR #3027, opened 2020-02-21
[MX-196] New Home Artwork cards

ashfurrow:
This is a WIP to get feedback on these changes early. The code wasn't great so I'm cleaning it up a bit first. ~I've left TODOs where things will need follow-up.~ Here's a tl;dr

- Updated lodash typing so we can `zip` arrays of two different types ([see PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/23232)).
- Updated Home `index.tsx` to remove state computation and replace imperative data computation with strongly-typed alternative.
- Converted `ArtistRail` to use styled-components and async/await instead of Promises everywhere (these were especially bad because the result of the promises were ignored anyway).

Screen Shot 2020-02-24 at 10 55 25


emission/src/lib/Scenes/Home/Components/ForYou/index.tsx
```diff
@@ -101,17 +95,17 @@ export class ForYou extends React.Component<Props, State> {
             />
           )
         }}
-        renderItem={({ item: { data, type } }) => {
```
ashfurrow: TypeScript wasn't recognizing how the `type` string property discriminated the `data` type when restructuring `item` like this, so I broken it out.

emission/src/lib/Scenes/Home/Components/ForYou/index.tsx
```diff
@@ -82,9 +42,43 @@ export class ForYou extends React.Component<Props, State> {
...
+                type: "artist",
+                data: module,
+              } as const)
+          ),
+          artworkModules.map(
+            module =>
+              ({
+                type: "artwork",
+                data: module,
+              } as const)
```
ashfurrow: First use of `as const` for me, pretty nifty!

emission/src/lib/Components/Home/ArtistRails/ArtistRail.tsx
```diff
@@ -3,20 +3,21 @@
 //       let Relay re-render the cards.
```
ashfurrow: I've decided to leave this for another day, especially since the designs say we should iterate on the interaction, so this might be a non-issue with a new design.

emission/src/lib/Scenes/Home/Components/ForYou/index.tsx
```diff
@@ -82,9 +42,43 @@ export class ForYou extends React.Component<Props, State> {
...
 
   render() {
+    const { forYou } = this.props
+    const artworkModules = forYou.artwork_modules || []
+    const artistModules = forYou.artist_modules && forYou.artist_modules.concat()
+    const fairsModule = forYou.fairs_module
+
+    const interleavedArtworkArtists = compact(
+      flatten(
+        zip(
```
ds300: ✨

emission/src/lib/Components/Home/ArtistRails/ArtistRail.tsx
```diff
@@ -83,81 +90,81 @@ export class ArtistRail extends Component<Props, State> {
...
+      completionHandler(true)
+      if (suggestion) {
+        await this.followedArtistAnimation(followArtist)
+        await this.replaceFollowedArtist(followArtist, suggestion)
+        await this.suggestedArtistAnimation(suggestion)
+      }
+    } catch (error) {
+      console.warn(error)
+      completionHandler(false)
+    }
```
ds300: ✨ 💆

ashfurrow:
Okay, this is mostly done, just some cleanup to do. The artist name is currently overflowing its parents container in a way that I don't quite understand, but I'll look into it on Monday. 

Screen Shot 2020-02-21 at 16 53 03
Screen Shot 2020-02-21 at 16 53 50

ashfurrow:
@ds300 okay, this is ready for review 👍

ArtsyOpenSource:
## TSLint Issues:

  * :no_entry_sign: <a href="https://github.com/artsy/eigen/blob/artwork-cards/src/lib/Scenes/Home/Components/ForYou/index.tsx">index.tsx</a> - ordered-imports -Named imports must be alphabetized.
  
<p align="right">
</p>

src/lib/Components/Home/ArtistRails/ArtistRail.tsx
```diff
@@ -184,57 +190,62 @@ export class ArtistRail extends Component<Props, State> {
...
-    if (!this.state.artists.length) {
-      return null
-    }
-
     return (
-      <View>
-        <View style={styles.title}>
-          <SectionTitle>{this.title()}</SectionTitle>
-          <SectionTitle style={styles.subtitle}>{this.subtitle()}</SectionTitle>
+      this.state.artists.length && (
```
ds300: this render method could return the value `0` which react-native doesn't like. You end up with one of these 
image

should probably be `this.state.artists.length ? (...jsx...) : null`

src/lib/Components/Home/ArtistRails/ArtistRail.tsx
```diff
@@ -184,57 +190,62 @@ export class ArtistRail extends Component<Props, State> {
...
-        {this.renderModuleResults()}
-        <Separator />
-      </View>
+      )
     )
   }
 }
 
+// TODO: Specs say this should be Sans, but let's update all Home titles at once
+// and convert to styled-components
```
ds300: good call, consistency is more important 👍

src/lib/Components/Home/ArtistRails/ArtistCard.tsx
```diff
@@ -36,133 +39,88 @@ export class ArtistCard extends React.Component<Props, State> {
...
-
-    return lines
-  }
-
   render() {
     const artist = this.props.artist
-    const imageURL = artist.image && artist.image.url
+    const avatarImageURL = artist.avatar && artist.avatar.url
+    const artworkImages = compact(artist.artworksConnection.edges.map(edge => edge.node.image?.url))
+    const artworkImageWidth = (CARD_WIDTH - (artworkImages.length - 1)) / artworkImages.length
```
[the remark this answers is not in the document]
ashfurrow: The images appear to have a 1px border between them, but since `ImageView` requires either an aspect ratio or explicit dimensions, I thought that reducing the width by the number of spaces between images (`artworkImages.length - 1`) would be good. If that makes sense, I can add a comment.

package.json
```diff
@@ -118,7 +118,7 @@
     "@types/chalk": "2.2.0",
     "@types/dedent": "0.7.0",
     "@types/jest": "24.0.11",
-    "@types/lodash": "4.14.91",
+    "@types/lodash": "^4.14.149",
```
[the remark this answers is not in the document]
ashfurrow: Haha, yeah 😅

src/lib/Components/Home/ArtistRails/ArtistRail.tsx
ashfurrow: Oh this ol' bug eh? Hmm, good catch.

src/lib/Components/Home/ArtistRails/ArtistCard.tsx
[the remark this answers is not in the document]
ds300: ah I se! Before I was looking at the image in the PR description which doesn't appear to have the 1px space.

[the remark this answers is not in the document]
ashfurrow: Huh! Now that you mention it, I'm not seeing the division that I'd expect. I'll look into it further to match the designs 👍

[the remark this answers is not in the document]
ashfurrow: I fixed this, and added a comment, in https://github.com/artsy/eigen/pull/3027/commits/1865e43ace453f5968f3c5926648e3c4901a9f82

PR #3033, opened 2020-02-24
Sort options filter

ashleyjelks:
Enables selecting sort options in artwork filter nav stack.

https://artsyproduct.atlassian.net/browse/FX-1782

Kapture 2020-02-25 at 12 14 18

#trivial

### DIFF
--- a/src/lib/Components/ArtworkFilterOptions/SortOptions.tsx
+++ b/src/lib/Components/ArtworkFilterOptions/SortOptions.tsx
@@ -1,4 +1,4 @@
-import { ArrowLeftIcon, Box, Flex, Sans, Serif, space } from "@artsy/palette"
+import { ArrowLeftIcon, Box, CheckIcon, Flex, Sans, Serif, space } from "@artsy/palette"
 import { BackgroundFill, OptionListItem } from "lib/Components/FilterModal"
 import React from "react"
 import { FlatList, TouchableOpacity } from "react-native"
@@ -7,11 +7,35 @@ import styled from "styled-components/native"
 
 interface SortOptionsScreenProps {
   navigator: NavigatorIOS
+  updatedSortOption: (string: SortTypes) => void
 }
 
-export class SortOptionsScreen extends React.Component<SortOptionsScreenProps> {
+interface SortOptionsScreenState {
+  currentSelection: SortTypes
+}
+
+export class SortOptionsScreen extends React.Component<SortOptionsScreenProps, SortOptionsScreenState> {
+  state: SortOptionsScreenState = {
+    currentSelection: "Default",
+  }
+
   handleBackNavigation() {
     this.props.navigator.pop()
+  }
+
+  getCheckboxSelection(selectedOption) {
+    return (
+      selectedOption === this.state.currentSelection && (
+        <Box mb={0.1}>
+          <CheckIcon fill="black100" />
+        </Box>
+      )
+    )
+  }
+
+  selectSortOption(selectedOption) {
+    this.setState({ currentSelection: selectedOption })
+    this.props.updatedSortOption(selectedOption) // callback to set the current sort option on the Filter home screen
   }
 
   render() {
@@ -35,11 +59,14 @@ export class SortOptionsScreen extends React.Component<SortOptionsScreenProps> {
             renderItem={({ item }) => (
               <Box>
                 {
-                  <OptionListItem>
-                    <Flex p={2} flexDirection="row" justifyContent="space-between" flexGrow={1}>
-                      <Serif size="3">{item}</Serif>
-                    </Flex>
-                  </OptionListItem>
+                  <TouchableOpacity onPress={() => this.selectSortOption(item)}>
+                    <OptionListItem>
+                      <Flex p={2} flexDirection="row" justifyContent="space-between" flexGrow={1} alignItems="flex-end">
+                        <Serif size="3">{item}</Serif>
+                        {this.getCheckboxSelection(item)}
+                      </Flex>
+                    </OptionListItem>
+                  </TouchableOpacity>
                 }
               </Box>
             )}
@@ -51,7 +78,7 @@ export class SortOptionsScreen extends React.Component<SortOptionsScreenProps> {
   }
 }
 
-const SortOptions = [
+const SortOptions: SortTypes[] = [
   "Default",
   "Price (low to high)",
   "Price (high to low)",
@@ -71,3 +98,12 @@ export const ArrowLeftIconContainer = styled(TouchableOpacity)`
   margin-top: ${space(2)};
   margin-left: ${space(2)};
 `
+
+export type SortTypes =
+  | "Default"
+  | "Price (low to high)"
+  | "Price (high to low)"
+  | "Recently Updated"
+  | "Recently Added"
+  | "Artwork year (descending)"
+  | "Artwork year (ascending)"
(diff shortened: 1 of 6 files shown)


src/lib/Components/ArtworkFilterOptions/SortOptions.tsx
```diff
@@ -7,13 +7,37 @@ import styled from "styled-components/native"
...
+export class SortOptionsScreen extends React.Component<SortOptionsScreenProps, SortOptionsScreenState> {
+  state: SortOptionsScreenState = {
+    currentSelection: "Default",
+  }
+
   handleBackNavigation() {
     this.props.navigator.pop()
   }
 
+  getCheckboxSelection(selectedOption) {
```
sweir27: #minor the "Checkbox" naming here is confusing. It looks like this method is returning a check mark if the option is selected, so maybe something like `renderCheckmarkIfSelected`?

Side note-- this logic is small enough that you might be able to inline it.

src/lib/Components/ArtworkFilterOptions/SortOptions.tsx
```diff
@@ -7,13 +7,37 @@ import styled from "styled-components/native"
...
+        <Box mb={0.1}>
+          <CheckIcon fill="black100" />
+        </Box>
+      )
+    )
+  }
+
+  selectSortOption(selectedOption) {
+    this.setState({ currentSelection: selectedOption })
+    this.props.updatedSortOption(selectedOption) // callback to set the current sort option on the Filter home screen
```
sweir27: Also #minor but since this method is _updating_ the sort option, I think it would be clearer to call it `updateSortOption`.

src/lib/Components/__tests__/FilterModal-tests.tsx
```diff
@@ -62,10 +74,92 @@ describe("Filter modal navigation flow", () => {
...
   })
+
+  it("only displays a check mark next to the currently selected sort option on the sort screen", () => {
+    const sortScreen = create(
+      <Theme>
+        <SortOptionsScreen navigator={mockNavigator as any} updatedSortOption={jest.fn()} />
+      </Theme>
+    )
+
+    const instance = sortScreen.root.findAllByType(SortOptionListItemRow)[2]
```
sweir27: This test would be a little stronger if we were asserting based on the actual sort option (i.e. `Artwork year (descending)`) instead of just the position in the list! (Only because it's possible for the order to change on the different screens, even though we don't expect them to).

So something like `const row = sortScreen.root.filterWhere(x => x.text().includes("Price (high to low)"))`. I think if you do it that way you can remove the assertion on line 100.

src/lib/Components/__tests__/FilterModal-tests.tsx
```diff
@@ -62,10 +74,92 @@ describe("Filter modal navigation flow", () => {
...
+    act(() => instance.props.onPress())
+
+    const sortOption = sortScreen.root.findAllByType(InnerOptionListItem)[2]
+    const selectedOption = sortOption.findByType(SortSelection).props.children
+    const selectedOptionIcon = sortOption.findByType(CheckIcon)
+    const checkMark = sortScreen.root.findAllByType(CheckIcon)
+
+    expect(selectedOptionIcon.props).toEqual({ fill: "black100" })
```
sweir27: Because the checkmark is only rendered if the item is selected, I would expect us to be able to assert based on the presence (something like `expect(selectedOption.findByType(CheckIcon).length).toEqual(1)`). Does that work?

src/lib/Components/__tests__/FilterModal-tests.tsx
```diff
@@ -62,10 +74,92 @@ describe("Filter modal navigation flow", () => {
...
+  it("displays the currently selected sort option on the filter screen", () => {
+    const filterScreen = create(
+      <Theme>
+        <FilterOptions closeModal={jest.fn()} navigator={mockNavigator as any} />
+      </Theme>
+    )
+
+    const instance = filterScreen.root.findByType(FilterOptions).instance
+
+    instance.getSortSelection("Price (low to high)")
```
sweir27: Is it possible to instead have this test actually navigate between the screens and assert that the filter screen updates after an `onPress()` on the sort screen?

No need to spend a lot of time on that now, we can always add as we include more functionality!

src/lib/Components/__tests__/FilterModal-tests.tsx
```diff
@@ -62,10 +74,92 @@ describe("Filter modal navigation flow", () => {
...
+
+    const currentSortSelection = filterScreen.root.findByType(CurrentOption).props.children
+
+    expect(currentSortSelection).toEqual("Price (low to high)")
+  })
+
+  it("displays the currently selected sort option on the sort screen", () => {
```
sweir27: This test seems like it duplicates the one starting on line 84! I think you can remove (unless I'm missing something).

src/lib/Components/__tests__/FilterModal-tests.tsx
```diff
@@ -62,10 +74,92 @@ describe("Filter modal navigation flow", () => {
...
+      <Theme>
+        <FilterOptions closeModal={jest.fn()} navigator={mockNavigator as any} />
+      </Theme>
+    )
+
+    const instance = filterScreen.root.findByType(CurrentOption).props.children
+    expect(instance).toEqual("Default")
+  })
+
+  it("displays the default selection when the modal is closed", () => {
```
sweir27: This test doesn't seem to do too much-- it would be stronger if we could verify that, after visiting the sort screen + selecting an item and _then_ closing out of the modal, re-rendering the filter modal shows the "Default" option (and not the selected one). Aside from that, I'm not sure if closing out the modal presents a different case than the above test.

I'm not sure if that ☝️ is possible from within this context! We can always revisit.

src/lib/Components/__tests__/FilterModal-tests.tsx
ashleyjelks: I spent quite a bit of time on this and it is still not clear how to keep the context between the filter and sort screen components. I paired with David to confirm, so the suggestion was just to mock the state change as though the onPress func had been called/to treat this as a unit rather than an integration test.

src/lib/Components/__tests__/FilterModal-tests.tsx
ashleyjelks: Same here wrt to context.

src/lib/Components/__tests__/FilterModal-tests.tsx
ashleyjelks: No this didn't work, but we can find the check mark component by its props since they are unique to this screen. `sortScreen.root.findAllByProps({ fill: "black100" })` The problem with this is if another component with that property is later added to the screen then this assertion fails. 

`expect(selectedOption.findByType(CheckIcon).length).toEqual(1)` doesn't work because the test render returns a single object instance with the findByType method, and there's no length property available on object.

sweir27: Ah! Good call on `findByType` vs. `findAllByType`. (I clearly didn't run the code that I suggested. 😄) In any case, I think asserting based on the presence of the element (via its type) vs. its prop makes more sense here. We could either assert that the returned array from `findAllByType` has a length of 1, or that `findByType` returns anything at all.

The current assertion just verifies that the checkmark _has_ that prop, but I think we just care whether or not it's visible on the page.

ashleyjelks: Make sense!

PR #3034, opened 2020-02-24
[MX-205] Show password error details, increase minimum password length

ashfurrow:
This PR does two things:

- Show the user the error from Gravity when new account creation fails.
- Increases minimum allowed new password length to 8.

The UX is not... ideal. Modifying the error-handling is difficult because of how the code is organized, and tightly coupled. Additionally, while the user _can_ swipe to go back, there is no affordance for this and no back button. So in the case of an error, I decided to take the user back to the first screen where they can take action, where we ask for email. They'll have to go through the process from there. 

Screen Shot 2020-02-24 at 12 35 57

When it comes to migrating code from Objective-C to React Native, I'd say that the login/signup flow would be a great candidate. Maybe even higher priority than our routing logic, since it's so user-facing and high-impact. Open to feedback on that, though. /cc @ds300

### DIFF
--- a/Artsy/View_Controllers/Login_and_Onboarding/AROnboardingViewController.m
+++ b/Artsy/View_Controllers/Login_and_Onboarding/AROnboardingViewController.m
@@ -486,10 +486,8 @@
         }];
     } failure:^(NSError *error, id JSON) {
         __strong typeof (wself) sself = wself;
-        if (JSON) {
-            if (sself.state == AROnboardingStagePersonalizeName) {
-                [sself displayError:@"Could not create account"];
-            }
+        if ([JSON[@"type"] isEqualToString:@"param_error"]) {
+            [self displayError:JSON[@"message"]];
         } else {
             [sself displayNetworkFailureError];
         }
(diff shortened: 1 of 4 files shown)


Artsy/View_Controllers/Login_and_Onboarding/AROnboardingViewController.m
```diff
@@ -486,10 +486,8 @@ - (void)createUserWithName:(NSString *)name email:(NSString *)email password:(NS
         }];
     } failure:^(NSError *error, id JSON) {
         __strong typeof (wself) sself = wself;
-        if (JSON) {
-            if (sself.state == AROnboardingStagePersonalizeName) {
-                [sself displayError:@"Could not create account"];
-            }
+        if ([JSON[@"type"] isEqualToString:@"param_error"]) {
+            [self displayError:JSON[@"message"]];
```
brainbicycle: Curious as to reasoning behind not using strong self here?

Artsy/View_Controllers/Login_and_Onboarding/AROnboardingViewController.m
```diff
@@ -671,7 +669,15 @@ - (void)displayError:(NSString *)errorMessage
 {
     if (self.state == AROnboardingStateAcceptConditions) {
         self.shouldPresentFacebook = NO;
-        [self popViewControllerAnimated:YES];
+        // This is hacky. But the top bar isn't visible so the user doesn't know they can pop back with swipe.
+        // We pop back to the email view controller because the user can tap next from here to the
+        // actual problem.
+        // Can't wait for this to be moved to React Native.
```
brainbicycle: 👍

Artsy/View_Controllers/Login_and_Onboarding/AROnboardingViewController.m
ashfurrow: Ah, that would be... a mistake 🙈 I will follow-up.

ashfurrow: Pushed a fix: https://github.com/artsy/eigen/commit/84f615bac9eb3e78b3afd3e0ccf9d1f0dba7d09d

PR #3036, opened 2020-02-25
Intelligent fallbacks for packager hostname

ds300:
@anandaroop [ran into an issue](https://artsy.slack.com/archives/C02BAQ5K7/p1582576904013500) where his macbook's local hostname was not resolvable because he didn't have any sharing options enabled in his system settings. This meant the app couldn't connect to the react packager to load the JS bundle. 

To prevent people running into the same issue in the future, this PR makes the host string fall back to your local wifi IP, or `localhost` if you're not connected to wifi.

#trivial

### DIFF
--- a/scripts/set_packager_host.sh
+++ b/scripts/set_packager_host.sh
@@ -1,5 +1,34 @@
 #!/usr/bin/env bash
 
+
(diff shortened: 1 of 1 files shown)


anandaroop:
🙏

PR #3040, opened 2020-02-25
[MX-184] Allows some mutations to bypass client-side cache clearing

ashfurrow:
This fixes an issue where marking an artwork as recently viewed would clear the entire client-side cache. We should re-evaluate this strategy altogether sometime, since we still do a lot of cache clearing that we don't need to be. 

I also added some types to `cacheMiddleware.ts`.

### DIFF
--- a/src/lib/relay/middlewares/__tests__/cacheMiddleware-tests.ts
+++ b/src/lib/relay/middlewares/__tests__/cacheMiddleware-tests.ts
@@ -4,7 +4,7 @@ import * as _cache from "../../../NativeModules/GraphQLQueryCache"
 const cache: jest.Mocked<typeof _cache> = _cache as any
 
 import { NetworkError } from "lib/utils/errors"
-import { cacheMiddleware } from "../cacheMiddleware"
+import { cacheMiddleware, GraphQLRequest } from "../cacheMiddleware"
 
 describe("cacheMiddleware", () => {
   const operation = {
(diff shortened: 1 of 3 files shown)


src/lib/relay/middlewares/cacheMiddleware.ts
```diff
@@ -1,9 +1,26 @@
 import { NetworkError } from "lib/utils/errors"
 import { Sentry } from "react-native-sentry"
+import { CacheConfig as RelayCacheConfig, RequestParameters } from "relay-runtime"
 import * as cache from "../../NativeModules/GraphQLQueryCache"
 
+type Mutable<T> = { -readonly [P in keyof T]: T[P] } // Remove readonly
```
ds300: ✨

src/lib/relay/middlewares/__tests__/cacheMiddleware-tests.ts
```diff
@@ -173,17 +173,27 @@ describe("cacheMiddleware", () => {
...
+      await cacheMiddleware()(mockedNext)(mutation)
+      expect(cache.clearAll).toHaveBeenCalled()
+    })
+
+    it("ignores clearing the cache for allowed mutations", async () => {
+      const mutation = {
+        operation: { id: "ArtworkMarkAsRecentlyViewedQuery", operationKind: "mutation" },
+        variables,
+        cacheConfig,
+      } as GraphQLRequest
```
ds300: #superminor if possible it's a good idea to move the type to the annotation position rather than the assertion position.

i.e. 

```ts
      const mutation: GraphQLRequest = {
        operation: { id: "ArtworkMarkAsRecentlyViewedQuery", operationKind: "mutation" },
        variables,
        cacheConfig,
      }
```

This way we get stricter checking for object literals.

There's a few instances of the same pattern above too.

PR #3042, opened 2020-02-27
[MX-230] Grid bugfix

ds300:
https://artsyproduct.atlassian.net/browse/MX-230

Accidentally stumbled on this error while working on other things. No idea why it's only just started happening, but it seems to happen *sometimes* when you navigate *away* from an artwork page. Either by tapping a link or hitting the back button.

I could reliably get it to happen after tapping between an artist and their artworks madly for a few seconds, and after this change the same kind of mad tapping does not produce errors.


src/lib/Scenes/Artwork/Components/OtherWorks/OtherWorks.tsx
```diff
@@ -51,6 +51,8 @@ export const OtherWorksFragmentContainer = createFragmentContainer<{ artwork: Ot
           ))}
         </Join>
       )
+    } else {
+      return null
     }
```
ds300: this is unrelated, but I think react warns if a component returns `undefined`, so it's best to explicitly return null.

PR #3048, opened 2020-02-28
[MX-183] Untrack derived files

ds300:
https://artsyproduct.atlassian.net/browse/MX-183

This PR

- Removes the query map from the git index so that we have fewer conflicts there.
- Removes the derived `emission/Pod/Assets/assets` folder from the git index.
- Updates the CI scripts to compile the emission bundle + assets before `pod install` so that they get added to the cocoapods resource copy script without us having to `touch` anything.
- Updates the paths in the fastlane script section where we upload source maps to Sentry. I missed this one when combining the repos!

I have verified that from a clean checkout:

- users can run `bundle exec pod install && yarn start`, hit play in xcode, and everything works.
- the steps in the CI script produce a functioning app bundle including `Emission.js` and the `assets` folder and all of its contents.


.gitignore
```diff
@@ -90,4 +90,7 @@ tslint-errors.json
 tsc_raw.log
 
 # Emission bundle
-emission/Pod/Assets/Emission.js*
+emission/Pod/Assets/*
```
ashfurrow: 🙌

Artsy.xcodeproj/project.pbxproj
```diff
@@ -4649,7 +4649,6 @@
 				"${PODS_ROOT}/Artsy+UIFonts/Pod/Assets/Unica77LL-Regular.otf",
 				"${PODS_ROOT}/Artsy+UILabels/Pod/Assets/Chevron_Black.png",
 				"${PODS_ROOT}/Artsy+UILabels/Pod/Assets/Chevron_Black@2x.png",
-				"${PODS_ROOT}/../emission/Pod/Assets/assets",
```
ashfurrow: Let's see what happens when we merge this PR, but I wouldn't be surprised to see this re-added to the project upon `pod install`.

ds300: It will be on CI, but not on people's local machines unless they run one of the `yarn bundle` commands before running `pod install`

ashfurrow: Ahhhhh, right right 👍

PR #3041, opened 2020-02-26
[MX-197] New Fair cards

ashfurrow:
This is looking good so far!

| Screen Shot 2020-02-26 at 15 40 20  |  Screen Shot 2020-02-26 at 15 40 10 |
|---|---|

Spent some time refactoring out the work from #3027 to be more generally applicable. I also reordered the modules to have recently saved near the top – we are planning to do reorder things generally later, but I moved this now so that we didn't have two similar-looking shelves sitting directly atop one another.

This is a WIP because:

- [x] I need to write tests, for existing and new components. 
- [x] I still have some general cleanup to do.
- [x] @samchieng and I are meeting with Devon H on Monday to go over the designs – Artsy has some constraints here, and we want to make sure we're on the same page before shipping.

For simple components like `CardScrollView.tsx`, I don't think tests are really warranted. But then again, maybe the file is _too_ simple – open to feedback. Our Danger rule assuming every component file needs a corresponding unit test file was definitely not written to contend with this, but I have to admit that I prefer the simplicity of very small component files.

#skip_new_tests

PR #3061, opened 2020-03-03
[MX-199] For you horizontal artwork rail

ds300:
https://artsyproduct.atlassian.net/browse/MX-199

horizontal-rail

Also updated the heading style

new-headings

Also:

- Deleted some unused code and assets
- Optimized the initial render of the for-you page
   I noticed that with a cold cache the images above the fold seemed to take forever to load. Upon investigation I found that 350+ images were being fetched as soon as I landed on the for-you page. So I did some FlatList tweaking and now it's more like 50.

#skip_new_tests


src/lib/Components/Home/ArtistRails/ArtistRail.tsx
```diff
@@ -167,37 +168,42 @@ export class ArtistRail extends Component<Props, State> {
...
-        <CardScrollView>
-          {cards}
-          <View style={{ width: 15 }} />
-        </CardScrollView>
-      )
-    } else {
-      return <Spinner style={{ flex: 1, marginBottom: 20 }} />
-    }
+    return (
+      <FlatList<SuggestedArtist>
```
ds300: refactored the ArtistRail to use a flat list so we could avoid rendering too many things at once

src/lib/Components/Home/ArtistRails/ArtistRail.tsx
```diff
@@ -167,37 +168,42 @@ export class ArtistRail extends Component<Props, State> {
...
+        ListFooterComponent={() => <Spacer mr={2} />}
+        ItemSeparatorComponent={() => <Spacer mr="15px" />}
+        ListEmptyComponent={() => <Spinner style={{ flex: 1, marginBottom: 20 }} />}
+        horizontal
+        showsHorizontalScrollIndicator={false}
+        scrollsToTop={false}
+        keyExtractor={artist => artist.id}
+        initialNumToRender={2}
+        windowSize={this.state.userHasScrolled ? 4 : 1}
+        onScrollBeginDrag={() => this.setState({ userHasScrolled: true })}
```
ds300: this is the trick to minimize rendering.

`initialNumToRender` should be just enough to fit on screen, then `windowSize` (measured in screen widths) should be `1` on mount. Then as soon as the user starts to scroll `windowSize` is increased to a normal level.

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>Missing Test Files:

- `src/lib/Scenes/Home/Components/ForYou/Components/__tests__/ArtworkRail-tests.tsx`

    </tr>
  </tbody>
</table>




<p align="right">
</p>

ds300:
@ashfurrow just realised this is going to create some significant conflicts with #3041 as they both refactored the artist rail — shall we get that one in first and then I can resolve the conflicts? It would feel so rude to inflict that work on you 😅

ashfurrow:
@ds300 ah yes, that's very kind of you! I'll get that merged right away. We received some feedback from the fairs team about the design, but I think we can merge #3041 now and then apply that feedback afterward.

src/lib/Components/SectionTitle.tsx
```diff
@@ -0,0 +1,38 @@
...
+      )
+    : React.Fragment
+  return (
+    <Wrapper>
+      <Flex mt="3" mb="1" flexDirection="row" alignItems="center">
+        <View style={{ overflow: "hidden", flex: 1 }}>
+          <Sans size="4" ellipsizeMode="tail" numberOfLines={1} data-test-id="title">
+            {title}
+          </Sans>
+          {subtitle && (
```
ashfurrow:
```suggestion
          {Boolean(subtitle) && (
```

ds300:
@ashfurrow conflicts resolved

can't believe how good this looks with the new fairs cards too ❤️ 

image

PR #3064, opened 2020-03-03
[MX-246] Remove beta label, fix artist follow

ashfurrow:
This is the commit we're deploying for the new beta. This PR merges changes back into master. I fixed MX-246 since it was a regression introduced in 6.3.2, and the fix was simple. /cc @ds300


ashfurrow:
For the record, be8f257 is the commit we're trying to deploy to Apple.

src/lib/Components/Home/ArtistRails/ArtistRail.tsx
```diff
@@ -111,7 +111,7 @@ export class ArtistRail extends Component<Props, State> {
                   ) {
                     edges {
                       node {
-                        ...ArtistCard_artist
+                        ...ArtistCard_artist @relay(mask: false)
```
brainbicycle: Datamasking is a new concept for me, curious how you determined this was the issue?

ds300:
@ashfurrow  looks like we need to update some native snapshots here

src/lib/Components/Home/ArtistRails/ArtistRail.tsx
ashfurrow: [Data masking](https://relay.dev/docs/en/thinking-in-relay#data-masking) is a really weird concept, it took me a while to get my head around it. Relay wants fragment containers to only have access to the data that the fragment container is asking for (even if more data was actually fetched from the server). The consequence is that the data we get back from this line is also masked. Since the `ArtistRail` fragment container doesn't ask for `internalID`, then `internalID` is "missing" in the model we get back. That's what Relay uses the ` $fragmentRefs` property for.

I'm not sure why this was working before and stopped working, but my thinking was that since there's logic to use either the `ArtistCardContainer` fragment container _or_ the plain `ArtistCard` plain React component, then the data we fetch for that plain component needs to be unmasked.

It's pretty tricky, so let me know what I can clarify 👍

PR #3065, opened 2020-03-04
Alphabetize package.json scripts

damassi:
This alphabetizes `package.json` scripts so that discovery is easier.


damassi:
Oops, closing as this is from a fork

PR #3067, opened 2020-03-04
[MX-248] Fix works-for-you scrolling

ds300:
https://artsyproduct.atlassian.net/browse/MX-248

Thanks to @alloy for discovering this bug  🙌 

The problem was

1. the `GenericGrid` component initially renders with a height of 0, then measures it's own width before populating the grid with artworks.
2. in `WorksForYou` (Artists tab on home screen) `GenericGrid`s were being rendered in a `FlatList`
3. When you scroll far enough in a `FlatList` things that are off-screen will be unmounted and removed. When you scroll back they will be mounted and shown again.
4. When a component re-mounts with a different height `FlatList` rightly adjusts the top offset of succeeding items to match.
5. Scrolling far down in `WorksForYou` and then scrolling back up was a disco funky janky mess.

This has been a problem since the recent react-native upgrade when all of our `ListView` components were refactored to use `FlatList`. `WorksForYou` was one of those.

The solution: calculate the width at a higher level and pass it down to the `GenericGrid` instances so they can do their layout synchronously on first render.

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -12,6 +12,7 @@ upcoming:
     - Fix WSODs related to markdown elements not being supported in ReadMore - david
     - Fix grid.artworks null pointer bug - david
     - Fixes issue with create password screen saying you needed 6 character but you need 8 - ash
+    - Fix bug where the `Artists` tab scroll view would jump around while scrolling
 
 releases:
   - version: 6.3.1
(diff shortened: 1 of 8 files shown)


ds300:
> we'll need to apply this fix to other areas where we have paging GenericGrids, like the Partner page, right?

Yeah I'll do a proper audit now.

ashfurrow:
@alloy That Street Sharks gif is giving me _strong_ 90's nostalgia 🤩

PR #3076, opened 2020-03-06
Lots of cleanup to the repo

ashfurrow:
This cleans up a few things:

- Removes old docs.
- Removes references to Spacecommander (a project like Prettier but just for Objective-C, which was never really used widely in Eigen).
- Removes Storybooks. (Note: I had to add an explicit `events` module to our dependencies, since it was only transitively being pulled in by storybooks. Not sure why `react-native-navigator-ios` doesn't depend on it explicitly.)

Some of the consignment tests were just looping through the Consignments storybooks and making snapshots, which weren't very useful tests IMO. Also, the Consignments flow is probably going to get some love from @damassi and the rest of the Consignments team, so I didn't feel it was necessary to keep these low-value snapshot tests.


package.json
```diff
@@ -28,14 +28,12 @@
     "open-sim": "open -a Simulator",
     "postinstall": "prettier --write package.json && node verify-npm-podspecs.js",
     "prepare": "patch-package",
-    "prestorybook": "rnstl --silent",
     "prettier-project": "yarn run prettier-write 'src/**/*.{ts,tsx}'",
     "prettier-write": "prettier --write",
     "relay": "relay-compiler",
     "start-packager": "rm -rf ./node_modules/.cache/babel-loader && env REACT_DEBUGGER=\"$([ -d '/Applications/React Native Debugger.app' ] && echo \"open 'rndebugger://set-debugger-loc?host=localhost&port=8081' --args\")\" node node_modules/react-native/local-cli/cli.js start",
-    "start-servers": "concurrently --kill-others 'yarn start-packager' 'yarn relay --watch' 'yarn storybook'",
```
ds300: yasss bye bye webpack 👋

PR #3078, opened 2020-03-06
User does not see lot condition report request if it's disabled for a sale

yuki24:
addresses https://artsyproduct.atlassian.net/browse/AUCT-928

This is basically the same change as https://github.com/artsy/reaction/pull/3223 but for our iOS app. This replaces the `isBiddable` flag with the new `canRequestLotCOnditionsReport` flag to toggle the request button.

### DIFF
--- a/src/lib/Scenes/Artwork/__tests__/Artwork-tests.tsx
+++ b/src/lib/Scenes/Artwork/__tests__/Artwork-tests.tsx
@@ -99,6 +99,31 @@ describe("Artwork", () => {
     expect(tree.root.findAllByType(ArtworkDetails)).toHaveLength(1)
   })
 
+  it("renders the ArtworkDetails component when conditionDescription is null but canRequestLotConditionsReport is true", async () => {
+    const tree = ReactTestRenderer.create(<TestRenderer />)
+    mockMostRecentOperation("ArtworkTestsQuery")
+    mockMostRecentOperation("ArtworkFullQuery", {
+      Artwork() {
+        return {
+          category: null,
+          canRequestLotConditionsReport: true,
+          conditionDescription: null,
+          signature: null,
+          signatureInfo: null,
+          certificateOfAuthenticity: null,
+          framed: null,
+          series: null,
+          publisher: null,
+          manufacturer: null,
+          image_rights: null,
+        }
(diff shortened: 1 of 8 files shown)


src/lib/Scenes/Artwork/__tests__/Artwork-tests.tsx
```diff
@@ -99,6 +99,31 @@ describe("Artwork", () => {
...
+          series: null,
+          publisher: null,
+          manufacturer: null,
+          image_rights: null,
+        }
+      }
+    })
+    mockMostRecentOperation("ArtworkMarkAsRecentlyViewedQuery")
+    await flushPromiseQueue()
+    expect(tree.root.findAllByType(ArtworkDetails)).toHaveLength(1)
```
dleve123: Is it possible to assert upon a component that's a bit more "close" to the actual lot condition request button? Like the `RequestConditionReport` component (or it's corresponding wrapped component)?

dleve123: Perhaps I don't understand the purpose of this test relative to corresponding spec in [ArtworkDetails-test](https://github.com/artsy/eigen/pull/3078/files#diff-a27d3021ebaac1cf475edd4295f15fedR148)?

ashfurrow: This makes sense to me. The `ArtworkDetails` component is only rendered if one of the conditions [here](https://github.com/artsy/eigen/pull/3078/files#diff-2ffc28e6bb7cee7e95f356f7c4ecd8d9R78) are met. The `ArtworkDetails` tests cover the individual cases for what's rendered and not within that component, but this test is only testing if the `ArtworkDetails` itself is included in the component tree.

yuki24:
The CI failure might be legit, looking into it now...

ashfurrow:
@yuki24 This test failure is a known issue – it intermittently fails. We've been restarted CI jobs as needed. Sorry for that!

yuki24:
Ah, I didn't know about that. I've restarted the job twice and the last one seems to have worked fine. Thanks for the heads-up!

PR #3093, opened 2020-03-17
[MX-240] Consignments ui fixes

ds300:
https://artsyproduct.atlassian.net/browse/MX-240

- Fix a bug where navigator push transitions were duplicated, by making sure `.focus` is called on a TextInput synchronously after mounting rather than during the transition animation (though not sure why the animation was being restarted as a result of that. The animation was handled by CA so I didn't feel it was worth digging into.)
- Fix another bug I found where the submit button on the Metadata step of the consignments flow was being hidden behind the keyboard.

Before:

image

After:

image


ashfurrow:
Travis is still having issues, going to merge.

ashfurrow:
Mmm, my mistake! Seems like there's no issue on bid flow 👍

PR #3097, opened 2020-03-18
Update snapshots

ds300:
we're getting failures in travis because we forgot to update some snapshots in #3093 

this is trivial, so gonna self-merge


ashfurrow:
My bad – Travis was having issues so I merged without waiting. Thanks for following up!

PR #3099, opened 2020-03-19
Update schema

jonallured:
This PR updates the MP schema to what I have rolling out over on this one: https://github.com/artsy/metaphysics/pull/2246. Note that there are breaking changes to parts of the stitched convection schema, but only to the query operations, not the mutations. In order to have more confidence that I didn't screw this up, I set things up locally so that my build of Eigen was pointed to my local MP which was pointed at my local convection. I ran through the consignment flow and saw the requests hitting both MP and convection and saw the database get new submission records added and updated throughout the process.

Here's some visual proof of that:

Screen Shot 2020-03-18 at 4 13 07 PM

So I feel good about these changes and that we've ensured backwards compatibility but plan to QA again once this is merged because I'm paranoid, haha.


ashfurrow:
Not sure why CI failed, so I restarted the build.

PR #3100, opened 2020-03-19
[MX-202] Updates tracking for tapping recent search results #trivial

ashfurrow:
The [tracking schema](https://github.com/artsy/eigen/blob/master/src/lib/utils/track/schema.ts) is kind of all over the place – some events `have spaces`, some `use_snake_case`, and some `UseCamelCase`. @mikehrom we've previously discussed having the Data team take ownership of this schema, which would let use use the schema y'all define with less back-and-forth. Maybe we can chat about this in Tuesday's Knowledge Share?

### DIFF
--- a/src/lib/utils/track/schema.ts
+++ b/src/lib/utils/track/schema.ts
@@ -310,6 +310,7 @@ export enum ActionNames {
   // dispatch this on search input focus and again any time they type
   ARAnalyticsSearchStartedQuery = "Searched",
   ARAnalyticsSearchItemSelected = "Selected result from search screen",
+  ARAnalyticsSearchRecentItemSelected = "selected_recent_item_from_search",
 
   /**
    * Collection page events
--- a/src/lib/Components/Consignments/Components/__tests__/__snapshots__/TextInput-tests.tsx.snap
+++ b/src/lib/Components/Consignments/Components/__tests__/__snapshots__/TextInput-tests.tsx.snap
@@ -23,6 +23,7 @@ exports[`does not have an activity when searching  1`] = `
     <TextInput
       allowFontScaling={true}
       autoCorrect={false}
+      autoFocus={false}
       clearButtonMode="while-editing"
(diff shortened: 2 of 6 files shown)


mikehrom:
@ashfurrow sounds good. we should always be using snake_case but we've never really exerted enough pressure/ownership over that, especially when it comes to the app. i have work in sprint now to audit/create tracking schema for recommendation rails so this can be a great example of how it should be done from here on out.

src/lib/utils/track/schema.ts
```diff
@@ -310,6 +310,7 @@ export enum ActionNames {
   // dispatch this on search input focus and again any time they type
   ARAnalyticsSearchStartedQuery = "Searched",
   ARAnalyticsSearchItemSelected = "Selected result from search screen",
+  ARAnalyticsSearchRecentItemSelected = "selected_recent_item_from_search",
```
mikehrom: this is good! i think the reason we maybe haven't minded using spaces/snake case/camel case etc. is because no matter what, these events show up in a table like `eigen_production.selected_recent_item_from_search` or `eigen_production.selected_result_from_search_screen` -- so Segment is enforcing snake case at some point.

ashfurrow: Gotcha gotcha. I thought there was some tooling around this!

mikehrom: That said, I'm not sure if Segment parses camel case the same way. We have always avoided camel case the most in data team practice.

mikehrom:
@ashfurrow my bad for not bringing this up earlier, but data team had internally discussed a different setup for this specific event. rather than create a new, independent event for selecting recent searches - can we instead put a descriptive property on the existing `selected_result_from_search_screen` event, specifically a boolean `recent` that is true if the item is a recent search selection and false if not? 

in the future we might even be able to add a boolean for `saved` or other types of interesting search selections we implement. let me know!

PR #3102, opened 2020-03-19
Adds a test for applying filters

ashleyjelks:
Adds a test for the apply filters refetch query.

#trivial


ashleyjelks:
#mergeongreen

PR #3105, opened 2020-03-20
Improve CI setup

ds300:
- Aggressively cache step results
- Do minimum amount of work on MacOS containers
  - Only use macos for building the app bundle (not js bundle), running native tests, deploying beta, and promoting beta.
  - On promote branch, don't run any checks, just the promotion fastlane script. 
- Move JS tests to circle from travis

- [x] Test all the caches
  - [x] Node modules cache (restoring from cache takes about 20 secs usually, vs 50 secs on macos, and 2mins to install from scratch)
  - [x] JS Tests cache (tests on circle take about 2 mins vs 4 mins on travis)
  - [x] JS Bundle cache
    - [x] metro transform cache (builds with transform cache are about 20s vs 2mins without cache)
  - [x] Gems cache
  - [x] Pods cache
  - [x] App build cache
  - [x] Native test cache
- [ ] Test the metaphysics deployment
- [ ] Test the beta deployment
- [ ] Test the app store promotion


ds300:
@ashfurrow this is ready for review 🍏

scripts/generate-manifest.js
```diff
@@ -0,0 +1,65 @@
...
+ * @param {string} file
+ */
+function hash(file) {
+  const md5 = crypto.createHash("md5")
+  md5.update(fs.readFileSync(file))
+  return md5.digest("hex")
+}
+
+function run() {
+  const [_, __, outFileName, ...stringPatterns] = process.argv
```
ashfurrow: Oh neat, hadn't seen `_, __` before, but that makes sense 👍

alloy: I recently learned you can even use only commas

```js
const [,, outFileName, ...stringPatterns] = process.argv
```

ashfurrow: Whoa! That's really cool – feels very _JavaScript-y_ 😄

alloy: Yeah, like a broken record 🙊

ds300: haha i had no idea this was a thing

PR #3106, opened 2020-03-20
[MX-202] - Use additional property for recent search distinction #trivial

brainbicycle:
Update recent search tracking based on feedback here:
https://github.com/artsy/eigen/pull/3100

- Add Boolean property rather than separate event


mikehrom:
Apologies for the back and forth on this one, but we've ultimately decided that the changes in #3100 are sufficient. Tracking separate events for each unique search-related event will be easier for analysis in the long run when we roll these interactions up to sessions. Thanks! I'll close now.

brainbicycle:
Data team decided to keep it as 2 separate events, closing.

PR #3107, opened 2020-03-21
Selected sort option

sweir27:
This PR fixes https://artsyproduct.atlassian.net/browse/FX-1851 and makes a few minor refactors to the store (will explain inline)!

To summarize:
- Removed the `selectedSortOption` item from the global state object since this value can be derived from existing properties (the `selectedFilters` and `appliedFilters` arrays).
- Refactored the shape of the `FilterArray`s slightly.
- Fixed the bug described in the ticket!

filterstates

#trivial

### DIFF
--- a/src/lib/Components/ArtworkFilterOptions/SortOptions.tsx
+++ b/src/lib/Components/ArtworkFilterOptions/SortOptions.tsx
@@ -3,7 +3,7 @@ import React, { useContext } from "react"
 import { FlatList, TouchableOpacity } from "react-native"
 import NavigatorIOS from "react-native-navigator-ios"
 import styled from "styled-components/native"
-import { ArtworkFilterContext, SortOption } from "../../utils/ArtworkFiltersStore"
+import { ArtworkFilterContext, selectedOptionsDisplay, SortOption } from "../../utils/ArtworkFiltersStore"
 import { BackgroundFill, OptionListItem } from "../FilterModal"
 
 interface SortOptionsScreenProps {
@@ -11,11 +11,14 @@ interface SortOptionsScreenProps {
 }
 
 export const SortOptionsScreen: React.SFC<SortOptionsScreenProps> = ({ navigator }) => {
-  const { dispatch, state } = useContext(ArtworkFilterContext)
+  const { dispatch } = useContext(ArtworkFilterContext)
 
   const handleBackNavigation = () => {
     navigator.pop()
   }
+
+  const selectedOptions = selectedOptionsDisplay()
+  const selectedSortOption = selectedOptions.find(option => option.filter === "sort")?.type
 
   const selectSortOption = (selectedOption: SortOption) => {
     dispatch({ type: "selectFilters", payload: { type: selectedOption, filter: "sort" } })
@@ -47,7 +50,7 @@ export const SortOptionsScreen: React.SFC<SortOptionsScreenProps> = ({ navigator
                       <SortSelection color="black100" size="3t">
                         {item}
                       </SortSelection>
-                      {item === state.selectedSortOption && (
+                      {item === selectedSortOption && (
                         <Box mb={0.1}>
                           <CheckIcon fill="black100" />
                         </Box>
--- a/src/lib/Components/ArtworkFilterOptions/SortOptions.tsx
+++ b/src/lib/Components/ArtworkFilterOptions/SortOptions.tsx
@@ -18,10 +18,10 @@ export const SortOptionsScreen: React.SFC<SortOptionsScreenProps> = ({ navigator
   }
 
   const selectedOptions = selectedOptionsDisplay()
-  const selectedSortOption = selectedOptions.find(option => option.filter === "sort")?.type
+  const selectedSortOption = selectedOptions.find(option => option.filterType === "sort")?.value
 
   const selectSortOption = (selectedOption: SortOption) => {
-    dispatch({ type: "selectFilters", payload: { type: selectedOption, filter: "sort" } })
+    dispatch({ type: "selectFilters", payload: { value: selectedOption, filterType: "sort" } })
   }
 
   return (
--- a/src/lib/Components/ArtworkFilterOptions/SortOptions.tsx
+++ b/src/lib/Components/ArtworkFilterOptions/SortOptions.tsx
@@ -3,7 +3,7 @@ import React, { useContext } from "react"
 import { FlatList, TouchableOpacity } from "react-native"
 import NavigatorIOS from "react-native-navigator-ios"
 import styled from "styled-components/native"
-import { ArtworkFilterContext, selectedOptionsDisplay, SortOption } from "../../utils/ArtworkFiltersStore"
+import { ArtworkFilterContext, SortOption, useSelectedOptionsDisplay } from "../../utils/ArtworkFiltersStore"
 import { BackgroundFill, OptionListItem } from "../FilterModal"
 
 interface SortOptionsScreenProps {
@@ -17,7 +17,7 @@ export const SortOptionsScreen: React.SFC<SortOptionsScreenProps> = ({ navigator
     navigator.pop()
   }
 
-  const selectedOptions = selectedOptionsDisplay()
+  const selectedOptions = useSelectedOptionsDisplay()
   const selectedSortOption = selectedOptions.find(option => option.filterType === "sort")?.value
 
   const selectSortOption = (selectedOption: SortOption) => {
--- a/src/lib/Components/ArtworkFilterOptions/__tests__/SortOptions-tests.tsx
+++ b/src/lib/Components/ArtworkFilterOptions/__tests__/SortOptions-tests.tsx
@@ -1,39 +1,93 @@
-import { Theme } from "@artsy/palette"
+import { Box, Theme } from "@artsy/palette"
+import { mount } from "enzyme"
 import React from "react"
-import * as renderer from "react-test-renderer"
 import { FakeNavigator as MockNavigator } from "../../../../lib/Components/Bidding/__tests__/Helpers/FakeNavigator"
 import { OptionListItem } from "../../../../lib/Components/FilterModal"
-import { ArtworkFilterContext, ArtworkFilterContextState } from "../../../utils/ArtworkFiltersStore"
-import { SortOptionsScreen as SortOptions } from "../SortOptions"
+import { ArtworkFilterContext } from "../../../utils/ArtworkFiltersStore"
+import { InnerOptionListItem, SortOptionsScreen as SortOptions } from "../SortOptions"
 
 describe("Sort Options Screen", () => {
   let mockNavigator: MockNavigator
-  let state: ArtworkFilterContextState
 
   beforeEach(() => {
     mockNavigator = new MockNavigator()
-    state = {
-      selectedSortOption: "Default",
-      selectedFilters: [],
-      appliedFilters: [],
-      applyFilters: false,
-    }
   })
 
-  it("renders the correct number of sort options", () => {
-    const root = renderer.create(
+  const MockSortScreen = ({ initialState }) => {
+    return (
       <Theme>
         <ArtworkFilterContext.Provider
           value={{
-            state,
+            state: initialState,
             dispatch: null,
           }}
         >
           <SortOptions navigator={mockNavigator as any} />
         </ArtworkFilterContext.Provider>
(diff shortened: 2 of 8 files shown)


src/lib/Components/ArtworkFilterOptions/SortOptions.tsx
```diff
@@ -3,22 +3,25 @@ import React, { useContext } from "react"
...
   const handleBackNavigation = () => {
     navigator.pop()
   }
 
+  const selectedOptions = selectedOptionsDisplay()
+  const selectedSortOption = selectedOptions.find(option => option.filterType === "sort")?.value
+
   const selectSortOption = (selectedOption: SortOption) => {
-    dispatch({ type: "selectFilters", payload: { type: selectedOption, filter: "sort" } })
+    dispatch({ type: "selectFilters", payload: { value: selectedOption, filterType: "sort" } })
```
sweir27: Just a minor naming refactor! I kept thinking that `type` referred to the "type" of filter (i.e. "sort"), so I tried to make this more explicit with `filterType` and `value`. Happy to revisit!

src/lib/Components/ArtworkFilterOptions/__tests__/SortOptions-tests.tsx
```diff
@@ -1,39 +1,93 @@
...
-      selectedSortOption: "Default",
-      selectedFilters: [],
-      appliedFilters: [],
-      applyFilters: false,
-    }
   })
 
-  it("renders the correct number of sort options", () => {
-    const root = renderer.create(
+  const MockSortScreen = ({ initialState }) => {
```
sweir27: Extracted this out since I added a few more tests. 😄

src/lib/Components/FilterModal.tsx
```diff
@@ -77,18 +75,19 @@ interface FilterOptionsProps {
...
   const handleNavigationToSortScreen = () => {
     navigator.push({
       component: SortOptions,
     })
   }
   const [filterOptions] = useState<FilterOptions>([
     {
-      type: "Sort by",
+      filterText: "Sort by",
+      filterType: "sort",
```
sweir27: Added this extra property to make it easier for us to choose the selected option.

src/lib/Scenes/Collection/Screens/CollectionArtworks.tsx
```diff
@@ -3,7 +3,36 @@ import { InfiniteScrollArtworksGridContainer as InfiniteScrollArtworksGrid } fro
 import { get } from "lib/utils/get"
 import React, { useContext, useEffect } from "react"
 import { createPaginationContainer, graphql, RelayPaginationProp } from "react-relay"
-import { ArtworkFilterContext, ArtworkSorts } from "../../../utils/ArtworkFiltersStore"
+import { ArtworkFilterContext, FilterArray } from "../../../utils/ArtworkFiltersStore"
+
+enum ArtworkSorts {
```
sweir27: I moved this here and un-exported it since it was only used in one place.

src/lib/Scenes/Collection/Screens/CollectionArtworks.tsx
```diff
@@ -3,7 +3,36 @@ import { InfiniteScrollArtworksGridContainer as InfiniteScrollArtworksGrid } fro
...
+  "Recently added" = "-published_at",
+  "Artwork year (descending)" = "-year",
+  "Artwork year (ascending)" = "year",
+}
+
+const filterTypeToParam = {
+  sort: ArtworkSorts,
+}
+
+export const filterArtworksParams = (appliedFilters: FilterArray) => {
```
sweir27: I pulled this logic out into its own function in slight anticipation of the additional filter, which won't always have a selected value.

Sorting is a special case, because it's the only one where we will _always_ have a value to send in our request.

However, in order to make the UI correct (see: bug ticket), I opted to achieve that ☝️ here by explicitly passing `sort: "-decayed_merch"` if there is no other sort value applied.

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>TypeScript hasn't passed, see below for full logs</td>
    </tr>
  </tbody>
</table>



### TypeScript Fails

```src/lib/Components/__tests__/FilterModal-tests.tsx(312,27): error TS2322: Type '{ type: string; filter: string; }' is not assignable to type 'FilterData'.
  Object literal may only specify known properties, and 'type' does not exist in type 'FilterData'.
src/lib/Components/__tests__/FilterModal-tests.tsx(313,26): error TS2322: Type '{ type: string; filter: string; }' is not assignable to type 'FilterData'.
  Object literal may only specify known properties, and 'type' does not exist in type 'FilterData'.
```
<p align="right">
</p>

src/lib/utils/ArtworkFiltersStore.tsx
```diff
@@ -1,76 +1,77 @@
...
-      // if a filter is applied for the first time update state with the newly applied filter
-      if (previouslyAppliedFilters.length < 1) {
-        return {
-          applyFilters: true,
-          appliedFilters: action.payload,
-          selectedFilters: [],
-          selectedSortOption: artworkFilterState.selectedSortOption,
-        }
-      }
+      const filtersToApply = unionBy(
```
sweir27: This logic is slightly different. Like we talked about in-person, I think we can treat the "Ways to Buy" (which allows for multiple options selected) as the special case.

I discovered `unionBy` 😄-- this logic basically forms an array based on the `selected` and `applied` filters, favoring the `selected` ones (since that's what we'll ultimately want to "apply").

Through this I was also able to remove the `payload` from this event, since I believe when we apply filters, we always want to apply everything that's been selected.

src/lib/utils/ArtworkFiltersStore.tsx
```diff
@@ -1,76 +1,77 @@
...
+        artworkFilterState.selectedFilters,
+        artworkFilterState.appliedFilters,
+        "filterType"
+      )
 
-      // otherwise remove filters that can only be selected once, e.g. "sort" and merge those with newly applied filters
-      const payloadFilters = action.payload
-      sanitizeFilters = filter(previouslyAppliedFilters, prop => prop.filter === "sort")
-      const appliedFilters = union(differenceWith(previouslyAppliedFilters, sanitizeFilters), payloadFilters)
+      // Remove default values as those are accounted for when we make the API request.
```
sweir27: I _think_ this will work for other cases as well... but basically, when we've selected the "Default" option for any of the potential filters, it means we _don't_ want to filter by that at all (and so it's essentially... not applied anymore).

Like I mention above, sorting is a special case since when it's in Default-mode, we still need to send a value in our API request! But that's handled in the function above.

src/lib/utils/ArtworkFiltersStore.tsx
```diff
@@ -1,76 +1,77 @@
...
-
-      if (isFilterAlreadySelected) {
-        const mergedFilters = [...previouslySelectedFilters, currentlySelectedFilter]
-
-        return {
-          applyFilters: false,
-          selectedFilters: mergedFilters,
-          appliedFilters: filtersApplied,
-          selectedSortOption: currentlySelectedFilter.type,
+      const filtersToSelect = unionBy([action.payload], artworkFilterState.selectedFilters, "filterType")
```
sweir27: This updated logic is what fixes the bug in the ticket. I'm very thankful for all the tests! Makes it a lot less scary to play around here.

The trick here is we have to:
1. Update our potential "Selected filters" based on the one that was just tapped
2. Decide whether that filter was actually "valid", and remove it if not.

A filter _cannot_ be selected if:
- It has already been applied with that exact value. (So, if I have already applied "Recently added" and I try to select that option again).
- It has _not_ been applied, and you're trying to select the default. (So, if I haven't applied the "sort" filter at all and I try to select "Default").

Note that the selectedFilters array is meant to represent the _change from what's already been applied_.

src/lib/utils/ArtworkFiltersStore.tsx
```diff
@@ -1,76 +1,77 @@
...
-      return { applyFilters: false, appliedFilters: [], selectedFilters: [], selectedSortOption: "Default" }
+      return { applyFilters: false, appliedFilters: [], selectedFilters: [] }
   }
 }
 
+const defaultFilterOptions = {
+  sort: "Default",
+}
+
+export const selectedOptionsDisplay = (): FilterArray => {
```
sweir27: I'm not super familiar with hooks!

This function depends on `useContext`, but it could also just accept the necessary pieces of state as arguments to the function. I'm not sure which is better (this way is certainly more frustrating to test). Happy to change if it's clearer (or if this is less performant).

src/lib/utils/ArtworkFiltersStore.tsx
```diff
@@ -1,76 +1,77 @@
...
+  const { state } = useContext(ArtworkFilterContext)
+
+  const defaultFilters: FilterArray = [{ filterType: "sort", value: "Default" }]
+
+  return unionBy(state.selectedFilters, state.appliedFilters, defaultFilters, "filterType")
```
sweir27: This function returns the "selected" state that we want to display in our UI. Note that it's actually a combination of a  few things!

If you've selected a filter option in the filter modal UI, that should take precedence.
If you haven't selected anything, but you have _applied_ an option in a previous interaction, we will show that.
If you haven't selected or applied anything, we show the default.

(We may not have to keep track of the defaults for the other filters as they're more straightforward).

src/lib/utils/ArtworkFiltersStore.tsx
```diff
@@ -116,14 +121,4 @@ export type SortOption =
...
-export enum ArtworkSorts {
-  "Default" = "-decayed_merch",
-  "Price (high to low)" = "sold,-has_price,-prices",
-  "Price (low to high)" = "sold,-has_price,prices",
-  "Recently updated" = "-partner_updated_at",
-  "Recently added" = "-published_at",
-  "Artwork year (descending)" = "-year",
-  "Artwork year (ascending)" = "year",
-}
+export type FilterOption = "sort"
```
sweir27: Just cleaned this up a little bit since we only have the one filter for now. 😄

src/lib/utils/ArtworkFiltersStore.tsx
ds300: I don't think it matters much either way unless, like you say, you want to unit test this function. In which case passing the state in makes more sense to me too.

I would also note that it's a very very strong idiom (practically a law) to begin the names of hooks with the word 'use'. Any function (except a functional component) which uses a hook is also a hook. e.g. I would call this `useSelectedOptionsDisplay` if you leave it as a hook.

src/lib/utils/ArtworkFiltersStore.tsx
ashleyjelks: This looks good, any chance we can put this in a comment above the new `filtersToSelect` function? I think it will help anyone that's new to the code understand what this is supposed to do.

src/lib/Scenes/Collection/Screens/CollectionArtworks.tsx
ashleyjelks: I wonder if we can set a default value for the `appliedFilters` function parameter, and when ClearAll is applied, just call this function, which should make a request to Relay passing in the default parameters.

ashleyjelks:
Just a few comments/questions but I think is good to merge once the CI tests pass!

src/lib/utils/ArtworkFiltersStore.tsx
sweir27: Ah! Yes-- I put a comment below to explain but I can move it up.

src/lib/Scenes/Collection/Screens/CollectionArtworks.tsx
sweir27: For "Clear All", we'll need to make sure that we're properly handling the UI treatment within the filter modal (everything visually gets cleared) and also handling the ability to Apply _or_ cancel out.

So I think we'll have to do something a little more involved. But totally-- when we `Apply` the result of a "Clear All", we should be able to pass an empty array and the default params here will make sure we're sending the right thing.

sweir27:
Per our conversation on slack-- this appears to be failing due to an issue on travis' side. Will merge in and monitor master.

PR #3116, opened 2020-03-25
[AUCT-964] User who needs IDV sees a button to register which sends user to typical sale registration form and does not see max bid dropdown

yuki24:
addresses https://artsyproduct.atlassian.net/browse/AUCT-964

## Todos

 * [x] Merge and deploy https://github.com/artsy/metaphysics/pull/2262
 * [x] Fix type errors
 * [x] Discuss the FAQ link with @deliciousnachos 
 * [x] Back-fill tests

## Screenshot

Screen Shot 2020-03-25 at 5 25 56 PM

### FAQ pages

Screen Shot 2020-03-30 at 4 03 49 PM

### DIFF
--- a/src/lib/Scenes/Artwork/Components/CommercialButtons/__tests__/BidButton-tests.tsx
+++ b/src/lib/Scenes/Artwork/Components/CommercialButtons/__tests__/BidButton-tests.tsx
@@ -17,7 +17,9 @@ import { renderRelayTree } from "lib/tests/renderRelayTree"
 import { merge as _merge } from "lodash"
 import { Settings } from "luxon"
 import React from "react"
+import { View } from "react-native"
 import { graphql } from "react-relay"
+
 import { BidButtonFragmentContainer as BidButton } from "../BidButton"
 
 jest.unmock("react-relay")
@@ -26,6 +28,11 @@ const merge: (...args: object[]) => any = _merge
 
 const realNow = Settings.now
 const realDefaultZone = Settings.defaultZoneName
+
+const meFixture: BidButtonTestsQueryRawResponse['me'] = {
+  id: 'id',
+  identityVerified: false,
+}
 
 describe("BidButton", () => {
   beforeAll(() => {
@@ -38,11 +45,13 @@ describe("BidButton", () => {
     Settings.defaultZoneName = realDefaultZone
   })
 
-  const getWrapper = async (response, auctionState) => {
+  const getWrapper = async (artwork, me, auctionState) => {
     return await renderRelayTree({
       Component: (props: any) => (
         <Theme>
-          <BidButton {...props} auctionState={auctionState} />
+          <View>
+            <BidButton {...props} auctionState={auctionState} />
+          </View>
         </Theme>
       ),
       query: graphql`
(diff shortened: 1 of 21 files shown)


yuki24:
This PR is ready for review now.

src/lib/Scenes/Artwork/Components/CommercialButtons/__tests__/BidButton-tests.tsx
```diff
@@ -91,103 +149,151 @@ describe("BidButton", () => {
...
-      const wrapper = await getWrapper(artwork, AuctionTimerState.LIVE_INTEGRATION_UPCOMING)
+      const wrapper = await getWrapper(artwork, meFixture, AuctionTimerState.LIVE_INTEGRATION_UPCOMING)
 
       expect(wrapper.text()).toContain("Registration closed")
     })
 
     it("with bidder registration pending approval", async () => {
       const artwork = merge({}, ArtworkFromTimedAuctionRegistrationOpen, BidderPendingApproval)
-      const wrapper = await getWrapper(artwork, AuctionTimerState.LIVE_INTEGRATION_UPCOMING)
+      const wrapper = await getWrapper(artwork, meFixture, AuctionTimerState.LIVE_INTEGRATION_UPCOMING)
```
dleve123: thoughts on adding a spec / extending this spec to assert that "Identity verification required to bid." is **not** displayed in the "context" of "timed, but not IDV-required auction"

src/lib/Scenes/Artwork/Components/__tests__/CommercialInformation-tests.tsx
```diff
@@ -20,7 +20,10 @@ describe("CommercialInformation", () => {
   it("renders all information when the data is present", () => {
     const component = mount(
       <Theme>
-        <CommercialInformationTimerWrapper artwork={CommercialInformationArtwork as any} />
+        <CommercialInformationTimerWrapper
+          artwork={CommercialInformationArtwork as any}
+          me={{ identityVerified: false } as any}
```
dleve123: Pardon the perhaps newbie comment, I realize it's consistent with the above, but why would we need to cast `{ identityVerified: false }` to `any`? It seems like this is the type signature for `CommericalInformationTimerWrapper`: https://github.com/artsy/eigen/blob/1cd0ba6d4f67e9d9bbde70d61e9663fb2cc8e840/src/lib/Scenes/Artwork/Components/CommercialInformation.tsx#L24-L30

Would it make sense to extend the type signature instead to handle this new prop?

yuki24: There is a new `me` type added to `CommercialInformationProps `:

https://github.com/artsy/eigen/pull/3116/files#diff-4eb345a5f1c2d53e59806a3bff373250R27

It needs to be casted to `any` because the actual type of `CommercialInformation_me` looks like this:

```ts
export type CommercialInformation_me = {
    readonly " $fragmentRefs": FragmentRefs<"CommercialButtons_me">;
    readonly " $refType": "CommercialInformation_me";
};
```

We don't want to specify `$fragmentRefs` or `$refType` otherwise fixtures could be overloaded with verbose data that is not necessarily related to the actual tests.

PR #3111, opened 2020-03-24
Galleries viewing rooms shell

ashfurrow:
This is just an empty shell behind a lab option (so users don't see it until we're ready). #skip_new_tests

PR #3122, opened 2020-03-27
FX-1873: Allows user to "Clear All" and apply changes in collections filter

sweir27:
Addresses https://artsyproduct.atlassian.net/browse/FX-1873

When a user clicks "Clear All", we want to re-set their current UI state in the filter modal to the default, but also allow them to:
- X out of the modal entirely, in which case if they re-open it's as if they'd never clicked "Clear all" at all
- Apply the changes, in which case everything is returned to the default upon re-opening the modal.

In order to do this, I added an additional piece of state (currently named `previouslyAppliedFilters`). The updated definition of our `ArtworkContextFilterState` is:
- `selectedFilters`: Options that are selected within the filter modal UI. Gets re-set to `[]` when the modal is closed.
- `appliedFilters`: Always represents the filters that are applied to the artwork grid. Only changes when new filters are applied.
- `previouslyAppliedFilters`: Always equal to `appliedFilters` when the modal is opened. Gets re-set to `[]` when "Clear All" is selected.

I don't _think_ there's a way to achieve what we want without this added piece of state (but could be wrong!!). `appliedFilters` is our source of truth for what is _actually_ "applied", so it can only be updated when a change is _actually_ applied. `selectedFilters` represents the diff between what is currently applied and what the user has selected in the UI. In the case of "Clear All", it's possible you won't have selected _anything_ at the time when you click it, so I don't think we can use that.

Let me know if this is confusing! 

clearall

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -3,6 +3,7 @@ upcoming:
   date: TBD
   dev:
     - Added shell for Galleries development of Viewing Rooms - ash
+    - Fixes Clear All functionality in collections filtering - sweir27
   user_facing:
     -
 
--- a/src/lib/Components/FilterModal.tsx
+++ b/src/lib/Components/FilterModal.tsx
@@ -16,9 +16,7 @@ export const FilterModalNavigator: React.SFC<FilterModalProps> = ({ closeModal, 
   const { dispatch, state } = useContext(ArtworkFilterContext)
 
   const handleClosingModal = () => {
-    if (!state.appliedFilters) {
-      dispatch({ type: "resetFilters" })
-    }
+    dispatch({ type: "resetFilters" })
     closeModal()
   }
 
@@ -31,6 +29,9 @@ export const FilterModalNavigator: React.SFC<FilterModalProps> = ({ closeModal, 
     const selectedFiltersSum = state.selectedFilters.length
     return selectedFiltersSum > 0 ? `Apply (${selectedFiltersSum})` : "Apply"
   }
+
+  const isApplyButtonEnabled =
+    state.selectedFilters.length > 0 || (state.previouslyAppliedFilters.length === 0 && state.appliedFilters.length > 0)
 
   return (
     <>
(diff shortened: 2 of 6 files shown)


src/lib/Components/FilterModal.tsx
```diff
@@ -16,9 +16,7 @@ export const FilterModalNavigator: React.SFC<FilterModalProps> = ({ closeModal,
   const { dispatch, state } = useContext(ArtworkFilterContext)
 
   const handleClosingModal = () => {
-    if (!state.appliedFilters) {
```
sweir27: I don't think we need this check as I can't think of a reason why we _wouldn't_ want to reset the filters when you close the modal, especially as we're calling `applyFilters` first.

src/lib/Components/FilterModal.tsx
```diff
@@ -32,6 +30,9 @@ export const FilterModalNavigator: React.SFC<FilterModalProps> = ({ closeModal,
     return selectedFiltersSum > 0 ? `Apply (${selectedFiltersSum})` : "Apply"
   }
 
+  const isApplyButtonEnabled =
+    state.selectedFilters.length > 0 || (state.previouslyAppliedFilters.length === 0 && state.appliedFilters.length > 0)
```
sweir27: The second bit of logic enables the button if you've clicked "Clear all" and there's a diff between `appliedFilters` and `previouslyAppliedFilters`.

CHANGELOG.yml
```diff
@@ -3,6 +3,7 @@ upcoming:
   date: TBD
   dev:
     - Added shell for Galleries development of Viewing Rooms - ash
+    - Fixes Clear All functionality in collections filtering - sweir27
```
ashleyjelks: Are we updating these again?  I thought this was an automated process now?

sweir27: I think that we are! Per changes like https://github.com/artsy/eigen/pull/3111/files#diff-96801928eca93eea4a5b44f359b868b5R5

src/lib/Components/FilterModal.tsx
ashleyjelks: I was to save the selected but un-applied filters (so if I user selected some filters then closed the modal, but didn't apply them they would see those same filters selected next time the modal is opened.  I think it was one of the acceptance criteria but I'll have to confirm.

sweir27: If that's part of the acceptance criteria it strikes me as kind of confusing, as it would mean there's no way to clear _just_ your selected filters. ("Clear all" will clear both selected + applied filters).

If we _do_ end up wanting to persist the selected filters, I think we can just do that as part of the `resetFilters` action.

PR #3130, opened 2020-03-31
Upgrades CocoaPods-Keys

ashfurrow:
See: https://github.com/orta/cocoapods-keys/pull/202


ds300:
One more thing before this gets merged, I think we should add the `Gemfile.lock` to the cocoapods manifest since the ruby gems can (and, in this case, will) change the outcome of running `pod install`. I didn't realise that before, apologies.

ashfurrow:
That makes sense – no worries! We're figuring this out as we go together. I'll follow-up with that.

PR #3132, opened 2020-03-31
[AUCT-941] - Identity verification text

brainbicycle:
Show identity verification text on sale page with user without identity verified and sale requiring identity verification.

https://artsyproduct.atlassian.net/browse/AUCT-941

Paired with @yuki24 @bhoggard @dleve123 

See text under 'Register to Bid'

auctions-identity-verify-text


Artsy/Views/Auction/AuctionTitleView.swift
```diff
@@ -171,13 +176,27 @@ private extension AuctionTitleView {
         }
         container.addSubview(registerButton)
 
+        let identityVerificationAttributedString = NSMutableAttributedString(string: "Identity verification required to bid. FAQ")
+        let faqRange = (identityVerificationAttributedString.string as NSString).range(of: "FAQ")
```
ds300, line 180: nonblocking: this is UX feedback rather than code feedback, this 'FAQ' link at the end seemed weird to me. It might be better to have the link applied to the words `Identity verification`

image

This is also a bigger hit target.

Artsy_Tests/View_Tests/Auctions/AuctionTitleViewTests.swift
```diff
@@ -58,6 +59,17 @@ class AuctionTitleViewSpec: QuickSpec {
 
                 expect(subject).to( haveValidSnapshot() )
             }
+
+            it("looks good with a sale and user requiring identity verification") {
```
dleve123, line 63: just a thought – would it make sense to test the case where a sale requires identity verification and the user is already verified? Would a snapshot test for that provide any value / protect against any crashes?

Artsy/Views/Auction/AuctionTitleView.swift
dleve123: The underlined FAQ text is consistent throughout web and iOS. Also, we don't necessarily want to optimize for a bigger hit target as this is the first step in a 5ish minute process.

ds300: > The underlined FAQ text is consistent throughout web and iOS

I'd argue it would be better UX to change it on web too.

> Also, we don't necessarily want to optimize for a bigger hit target as this is the first step in a 5ish minute process.

Are you saying you'd prefer that people don't tap this link? If so, making things harder to tap is not a good way to get people to not tap them. A better alternative is de-emphasis through copy or style adjustments.

ashfurrow: We should always make tap targets reasonably big enough for users to easily tap – as an accessibility rule.

dleve123: Good points all around! 

> Are you saying you'd prefer that people don't tap this link? If so, making things harder to tap is not a good way to get people to not tap them. A better alternative is de-emphasis through copy or style adjustments.

Sorry, should have been more clear. We definitely want people to read the information linked if it's a blocker to registration. With that said, it's not the primary call to action on that view (completing sale registration is).

----

For the sake of making progress while acknowledging the possibility of a more optimal UX, I'd recommend that we keep the UX as is for this PR, but start a conversation with @deliciousnachos on whether this is the most appropriate UX.

If decide we want to change the design, we could follow up in a subsequent PR.

How does that sound?

ds300: Sure. Like I said, nonblocking :) and oh my glob I love nicole's github handle 🙌

Artsy_Tests/View_Tests/Auctions/AuctionTitleViewTests.swift
ashfurrow: Yeah good idea, it would be worthwhile to have as a test case. We don't have any near-term plans to move this into React Native – we'll be living with this code for at least a little while so it's worth testing. I'll review the test of the PR now.

Artsy/View_Controllers/Auction/AuctionViewController.swift
```diff
@@ -386,6 +386,19 @@ extension TitleCallbacks: AuctionTitleViewDelegate {
...
         }
     }
+
+    func userDidPressIdentityFAQ(_ titleView: AuctionTitleView) {
+        let showIdentityFAQ = {
+            let identityFAQPath = "/identity-verification-faq"
+            let viewController = ARSwitchBoard.sharedInstance().loadPath(identityFAQPath)
+            self.ar_TopMenuViewController()?.push(viewController, animated: true)
+        }
+        if let _ = presentedViewController {
```
ashfurrow, line 396: Nice catch 👍

Artsy/View_Controllers/Live_Auctions/LiveAuctionLotSetViewController.swift
```diff
@@ -347,6 +347,10 @@ extension LiveAuctionLotSetViewController: AuctionTitleViewDelegate {
         // NO-OP, button for this cannot be seen in this context
     }
 
+    func userDidPressIdentityFAQ(_ titleView: AuctionTitleView) {
+        // NO-OP, button for this cannot be seen in this context
```
ashfurrow, line 351: Thanks for the comment – reusing this class for Live Auctions is a big regret of mine 😅

Artsy_Tests/View_Tests/Auctions/AuctionTitleViewTests.swift
ashfurrow: Added the test case here: https://github.com/artsy/eigen/pull/3132/commits/c00fa98f3e5cb8580696fbbe1ca19ab07927d934

PR #3133, opened 2020-03-31
[Consignments] Add ArtistConsignButton to Artist page

damassi:
Adds a new ConsignButton to the Artist page, follow up on [recent](https://github.com/artsy/reaction/pull/3329) [work](https://github.com/artsy/reaction/pull/3317) in reaction. 

artist


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>Missing Test Files:

- `src/lib/Components/Artist/__tests__/ArtistConsignButton-tests.tsx`

    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/Components/Artist/ArtistConsignButton.tsx
```diff
@@ -0,0 +1,109 @@
...
+import { Schema } from "lib/utils/track"
+import { TouchableWithoutFeedback } from "react-native"
+
+export interface ArtistConsignButtonProps {
+  artist: ArtistConsignButton_artist
+  context: React.Component<any>
+}
+
+export const ArtistConsignButton: React.FC<ArtistConsignButtonProps> = props => {
+  const { isInMicrofunnel, imageURL, headline } = getData(props)
```
ashfurrow, line 19: Any reason to put this into `getData` instead of inlining that logic in `render()`?

ds300:
:ceiling-cat:

I have some design feedback. This widget immediately jumped out at me as not fitting with the rest of the app.

image

It's all tight and jumbled on the left there, needs more breathing room.

Also not sure about the copy "Sell your {Artist Name}", seems a little awkward to me, but then again I'm not an art collector.

ashfurrow:
Good catch David. The design doesn’t quite match the balance we have in, for example, the EntityHeader component. Maybe we could float the “get started” button to the right?

ds300:
yeah and also the border seems unnecessary in this context

src/lib/Components/Artist/ArtistConsignButton.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const tracking = useTracking()
+
+  return (
+    <TouchableWithoutFeedback
```
ds300, line 23: We should be using `TouchableOpacity` here. `TouchableWithoutFeedback` should be used rarely in situations where the user doesn't need extra feedback about whether or not they might trigger an action on press.

src/lib/Components/Artist/ArtistConsignButton.tsx
```diff
@@ -0,0 +1,109 @@
...
+              </Button>
+            </Box>
+          </Flex>
+        </Flex>
+      </BorderBox>
+    </TouchableWithoutFeedback>
+  )
+}
+
+function getData(props) {
```
ds300, line 66: we're going to be enabling `noImplicitAny` soon so would be good to make sure all functions have appropriate parameter types.

data/schema.graphql
```diff
@@ -5407,89 +5375,42 @@ type Mutation {
   ): RequestCredentialsForAssetUploadPayload
 
   # Attach an gemini asset to a consignment submission
-  createGeminiEntryForAsset(
-    input: CreateGeminiEntryForAssetInput!
-  ): CreateGeminiEntryForAssetPayload
+  createGeminiEntryForAsset(input: CreateGeminiEntryForAssetInput!): CreateGeminiEntryForAssetPayload
```
ds300, line 5379: all these formatting changes strike me as odd. any idea what's going on here?

CHANGELOG.yml
```diff
@@ -4,6 +4,7 @@ upcoming:
   dev:
     - Added shell for Galleries development of Viewing Rooms - ash
     - Fixes Clear All functionality in collections filtering - sweir27
+    - Adds <ArtistConsignButton> to Artist page - chris
```
ashfurrow, line 7: We'll meed to move this up to `upcoming` to merge in changelog changes in #3109.

src/lib/Components/Artist/ArtistConsignButton.tsx
jonallured: FWIW I like that we have extracted some logic to `getData` as has better encapsulation but totally defer to @ashfurrow in terms of following the conventions of Eigen. Small functions like this are also really easy to get specs around!

ds300: I'd vote for inlining. `getData` is hardly doing anything so having it separated out actually makes it harder to follow the logic of this render function IMO.

damassi: Ah, this was copy/paste from reaction. Will fix.

damassi:
@ds300 - these are the designs. They're trying to surface consignments more. Feel free to raise in #consignments channel however and we can iterate.

ds300:
@damassi I'm offline for the evening but I've asked Sam to talk to Nicole B about it. This design work great on the web but on mobile it's awkward and not QWOA so let's iterate before merging :+1:

src/lib/Components/Artist/ArtistConsignButton.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+import { ArtistConsignButton_artist } from "__generated__/ArtistConsignButton_artist.graphql"
+import SwitchBoard from "lib/NativeModules/SwitchBoard"
+import { Router } from "lib/utils/router"
+import { Schema } from "lib/utils/track"
+import { TouchableWithoutFeedback } from "react-native"
+
+export interface ArtistConsignButtonProps {
+  artist: ArtistConsignButton_artist
+  context: React.Component<any>
```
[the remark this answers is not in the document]
damassi: Cool, that worked good. Was a bit confusing in the sense that the `ref` needed to be assigned to the container but in retrospect that makes sense.

data/schema.graphql
damassi: Prettier? Not too sure.

damassi:
@ds300 - as discussed in [this thread](https://artsy.slack.com/archives/CS3H1K4BD/p1585768533202600) we're going to merge and iterate. Thank you for raising as well as the detailed review.

PR #3134, opened 2020-04-01
[MX-203] Clean up query renderers

ds300:
https://artsyproduct.atlassian.net/browse/MX-203

In this I took all the query renders in QueryRenderers.tsx and moved them to colocate them with the fragment containers they wrap, avoiding a useless layer of indirection.

We also had some native code around displaying `WorksForYou` in its own view controller, but it wasn't actually being used so I removed it.

#trivial


Artsy/App/ARAppDelegate+Analytics.m
```diff
@@ -44,7 +44,6 @@
 #import "ARAugmentedVIRSetupViewController.h"
 #import "ARAugmentedFloorBasedVIRViewController.h"
 
-#import <Emission/ARWorksForYouComponentViewController.h>
```
ashfurrow, line 47: Cool, that works 👍 [MX-192](https://artsyproduct.atlassian.net/browse/MX-192) will probably end up re-creating this controller, but I think it's better to start from scratch anyways.

src/lib/AppRegistry.tsx
```diff
@@ -198,41 +142,13 @@ interface BidderFlowProps {
...
-  return <FairRenderer fairID={fairID} render={renderWithLoadProgress(Containers.Fair, { fairID })} />
-}
-
-interface ShowProps {
-  showID: string
-}
-
-const Show: React.SFC<ShowProps> = ({ showID }) => {
-  return <ShowRenderer showID={showID} render={renderWithLoadProgress(Containers.Show, { showID })} />
+  switch (intent) {
```
ashfurrow, line 146: Ha, much simpler 😄

src/lib/Containers/WorksForYou.tsx
```diff
@@ -123,29 +126,27 @@ export class WorksForYou extends React.Component<Props, State> {
...
 }
 
-const WorksForYouContainer = createPaginationContainer(
+export const WorksForYouContainer = createPaginationContainer(
   WorksForYou,
   {
-    query: graphql`
-      fragment WorksForYou_query on Query
+    me: graphql`
+      fragment WorksForYou_me on Me
```
ashfurrow, line 133: Nice!

PR #3141, opened 2020-04-02
Non-IDV bidder sees copy in the "registration pending" modals about identity verification

yuki24:
addresses https://artsyproduct.atlassian.net/browse/AUCT-951 and https://artsyproduct.atlassian.net/browse/AUCT-953.

## Todos

 * [x] Fix type errors
 * [x] Add more test coverage
    * [x] `src/lib/Components/Bidding/Screens/Registration.tsx`
    * [x] `src/lib/Components/Bidding/Screens/RegistrationResult.tsx`
   
### Registration Pending on iPhone X

AUCT-953

### Other screens on iPhone SE

Screen Shot 2020-04-03 at 12 10 40 PM

### DIFF
--- a/src/lib/Components/Bidding/Screens/RegistrationResult.tsx
+++ b/src/lib/Components/Bidding/Screens/RegistrationResult.tsx
@@ -1,5 +1,7 @@
+import { Button, Sans } from "@artsy/palette"
 import React from "react"
 import { View } from "react-native"
+import { blockRegex } from "simple-markdown"
 
 import { Icon20 } from "../Components/Icon"
 import { Flex } from "../Elements/Flex"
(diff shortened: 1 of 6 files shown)


src/lib/utils/renderMarkdown.tsx
```diff
@@ -209,6 +203,7 @@ export function defaultRules(modal: boolean = false): ParserRules {
     hr: {
       react: () => <Separator mb={2}></Separator>,
     },
+    ...roleOverrides
```
yuki24, line 206: @ds300 I wasn't sure if this is a good way to override the default Markdown rules. Do you have any other solutions? Let me know if there's anything better.

ds300: This seems good to me, nice job :+1:

src/lib/Components/Bidding/Screens/RegistrationResult.tsx
```diff
@@ -53,6 +67,19 @@ const registrationNetworkErrorMessage = {
   description: "Please\ncheck your internet connection\nand try again.",
 }
 
+const markdownRules = defaultRules(false, {
+  paragraph: {
+    match: blockRegex(/^((?:[^\n]|\n(?! *\n))+)(?:\n *)/),
```
erikdstock, line 72: is this matching a double-newline?

yuki24: I think this is matching a single-newline.

yuki24:
I have seen a bunch of test failures locally due to the RN version upgrade. Not sure what the CI thinks but marking this as WIP for now.

src/lib/Components/Bidding/Screens/RegistrationResult.tsx
yuki24: Well, actually, I was wrong. This is probably matching a double-newline.

PR #3144, opened 2020-04-03
[MX-258] Fix conversations styles

ds300:
https://artsyproduct.atlassian.net/browse/MX-258

After fixing the line height restrictions I discovered that it just plain looked bad with large text so I paired with @samchieng to update the components to palette.

I also checked the active bids inbox items and they look fine with large text so I left them alone for now.

## normal text size

image

## big text

image

Also did some general spring cleaning  🌱


src/lib/Components/Inbox/Conversations/ConversationSnippet.tsx
```diff
@@ -153,23 +81,32 @@ export class ConversationSnippet extends React.Component<Props> {
     const conversationText = conversation.lastMessage && conversation.lastMessage.replace(/\n/g, " ")
     const date = moment(conversation.lastMessageAt).fromNow(true) + " ago"
     return (
-      <TouchableWithoutFeedback onPress={() => this.conversationSelected()}>
-        <Card>
-          <CardContent>
+      <TouchableHighlight onPress={() => this.conversationSelected()} underlayColor={color("black5")}>
```
ds300, line 84: added some touch feedback for these inbox, in line with new motion guidelines.

src/lib/Components/Inbox/Conversations/Conversations.tsx
```diff
@@ -98,15 +93,20 @@ export class Conversations extends Component<Props, State> {
               />
             )
           }}
-          onScroll={isCloseToBottom(this.fetchData)} // TODO: Investiate why onEndReached fires erroniously
+          onEndReached={this.fetchData}
+          onEndReachedThreshold={2}
         />
-        {this.state.isLoading && <Spinner style={{ marginTop: 20, marginBottom: 20 }} />}
+        {this.props.relay.hasMore() && this.state.isLoading && (
```
ds300, line 99: the spinner was flickering whenever you scrolled before so I added an extra check for `.hasMore()` to prevent it

PR #3142, opened 2020-04-02
[MX-192] [MX-194] [MX-198] Collapse home subtabs into one

ashfurrow:
Still lots to do:

- [x] Need to show Auctions and Works for You modules in home feed (see: MX-194).
- [x] Build Auctions card (see: MX-198).
- [x] Verify changes to analytics with Mike.
- [x] Update home feed design (I think there's a new header at the top).
- [x] Update tests.
- [x] Works-for-you and Auctions screens have bottom cut off by tab bar.

Screen Shot 2020-04-02 at 17 06 42

Screen Shot 2020-04-06 at 14 30 22
Screen Shot 2020-04-06 at 14 30 19
Screen Shot 2020-04-06 at 14 30 30

#skip_new_tests

PR #3150, opened 2020-04-06
[MX-247] Upgrade Jest and React Native

ds300:
https://artsyproduct.atlassian.net/browse/MX-247

- Upgraded Jest (relatively painless, only one problem with `waitUntil` behaviour changing)
- Upgraded React-Native (Ran into trouble adding native [Flipper](https://fbflipper.com) support but reactintegration works, will leave native for another day)
- Moved from built-in Picker to @react-native-community picker (required fixing their typings. see react-native-community/react-native-picker#63) we have a patch for now
- Upgraded Swift (in an effort to get flipper sdk working, but it didn't seem to help)
- Converted all snapshot tests to 'renders without failing' tests, unless I could improve the tests in reasonable time, in which cases I did.
- Deleted some unused test files and useless tests.
- Enabled the new experimental LogBox

I've smoke-tested the app and everything seems normal. The new LogBox is ✨gorgeous ✨ 

@ashfurrow I realise you've got some pretty big changes brewing and that takes precedence IMO so I'm happy to let those go in and then deal with any conflicts later.

After this gets merged everyone will need to **`(╯°□°)╯︵ ┻━┻`**


Makefile
```diff
@@ -185,7 +185,7 @@ flip_table:
 	rm -rf emission/Pod/Assets/Emission*
 	rm -rf emission/Pod/Assets/assets
 	@echo 'Reinstall dependencies ┬─┬ノ( º _ ºノ)'
-	bundle exec pod install
+	bundle exec pod install --repo-update
```
ashfurrow, line 188: Nice 👍

src/lib/Components/Bidding/Components/Animation/CssTransition.tsx
```diff
@@ -30,6 +30,7 @@ export class CssTransition extends React.Component<CssTransitionProps, CssTransi
     Animated.timing(this.animatedValue, {
       toValue: 1,
       duration: nextProps.duration,
+      useNativeDriver: false,
```
ashfurrow, line 33: I _think_ I know why this is `false` and elsewhere it's `true`, but a comment would be helpful 👍

src/lib/Scenes/Artwork/__tests__/Artwork-tests.tsx
```diff
@@ -37,7 +37,7 @@ jest.unmock("react-relay")
 
 jest.mock("lib/relay/createEnvironment", () => ({
   defaultEnvironment: require("relay-test-utils").createMockEnvironment(),
-  reset() {
+  reset(this: { defaultEnvironment: any }) {
```
ashfurrow, line 40: TIL!

src/lib/Scenes/Map/Components/PinsShapeLayer.tsx
```diff
@@ -79,24 +81,19 @@ export class ShapeLayer extends Component<Props, State> {
...
         pinOpacity: new Animated.Value(0),
         clusterOpacity: new Animated.Value(0),
-        clusterRadius: new Animated.Value(0),
         rendered: true,
       },
       () => {
         Animated.timing(this.state.pinOpacity, {
           toValue: 1,
           duration: this.props.duration,
+          useNativeDriver: false,
```
ashfurrow, line 90: Curious about these two as well.

ashfurrow:
Okay, #3142 is merged so I think we should be able to go ahead with this PR at your convenience 👍

PR #3153, opened 2020-04-07
[MX-193] Home re-ordering and skeleton loader

ashfurrow:
Here's what it looks like:

Screen Shot 2020-04-07 at 15 13 12

We don't have a tonne of control over the ordering, so I called that out in a comment.


brainbicycle:
LGTM 👍

PR #3155, opened 2020-04-07
Add documentation for updating screenshots

yuki24:
Following up on the slack thread: https://artsy.slack.com/archives/CN8S32FJR/p1586288493195200?thread_ts=1586288257.193600&cid=CN8S32FJR

I'm not sure if this is all correct, but let me know if there should be anything added.


docs/getting_started.md
```diff
@@ -40,6 +40,21 @@ You can run tests via the CLI using:
...
 
+#### Updating snapshots
+
+We use [Nimble-Snapshots](https://github.com/ashfurrow/Nimble-Snapshots) to take screenshots while running tests and
+these screenshots are checked in to the source control. When you change e.g. the background color of a particular
+button, it calculates the diff between two screenshots and makes the test fail if the delta is above a certain
+threshold.
+
+In order to update existing screenshots, replace individual calls to `haveValidSnapshot()` with `recordSnapshot()`,
+or you can change `switchChecksWithRecords` to `true` in `HaveValidSnapshot.swift` (this is a CocoaPod, Xcode will
```
brainbicycle, line 51: @ashfurrow is it the case that setting `switchChecksWithRecords` flags updates all snapshots? Would think in most cases we just want to update individual failing snapshots and then compare the changes to make sure they make sense?

ashfurrow: That is true, but it's typically how I do it (very lazy, I am). The snapshots do all get re-generated, but they are deterministic, so only the changed snapshot files are shown as changed in the git index.

docs/getting_started.md
```diff
@@ -40,6 +40,21 @@ You can run tests via the CLI using:
...
+warn you about changing a locked file). The location of the file could be looked up by:
+
+```
+tree -f Pods | grep HaveValidSnapshot
+```
+
```
ashfurrow, line 57:
```suggestion

For updating Objective-C snapshots, you'll need to do the same work but on a file named `EXPMatchers+FBSnapshotTest.m`.
```

PR #3157, opened 2020-04-08
[Consignments] Update ArtistConsignButton with targetSupply check

damassi:
Addresses https://artsyproduct.atlassian.net/browse/CSGN-98
Depends on https://github.com/artsy/metaphysics/pull/2284 
Related https://github.com/artsy/reaction/pull/3364

Updates the Consign button with new `isTargetSupply` check. 

#trivial (changelog entry already has sufficient description)

### DIFF
--- a/src/lib/Components/Artist/ArtistConsignButton.tsx
+++ b/src/lib/Components/Artist/ArtistConsignButton.tsx
@@ -20,13 +20,14 @@ export const ArtistConsignButton: React.FC<ArtistConsignButtonProps> = props => 
 
   const {
     artist: {
-      targetSupply: { isInMicrofunnel },
+      targetSupply: { isInMicrofunnel, isTargetSupply },
       name,
       image,
     },
   } = props
   const imageURL = image?.cropped?.url
-  const headline = isInMicrofunnel ? `Sell your ${name}` : "Sell art from your collection"
+  const showImage = imageURL && (isInMicrofunnel || isTargetSupply)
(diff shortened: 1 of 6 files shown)


src/lib/Components/Artist/ArtistConsignButton.tsx
```diff
@@ -46,7 +46,7 @@ export const ArtistConsignButton: React.FC<ArtistConsignButtonProps> = props =>
     <TouchableOpacity ref={buttonRef} onPress={onConsignButtonPress}>
       <BorderBox p={1}>
         <Flex alignItems="center" flexDirection="row">
-          {isInMicrofunnel && imageURL && (
+          {(isInMicrofunnel || isTargetSupply) && imageURL && (
```
jonallured, line 49: I like extracting locals for conditions like this because in my experience they tend to change and here's a good example of that! I think it would be cool to see the call site look more like this:

```ts
{showImage && (
  <Box pr={2}>
    <Image source={{ uri: imageURL }} />
  </Box>
)}
```

And then up top we could do this:

```ts
const showImage = (isInMicrofunnel || isTargetSupply) && imageURL
```

And that would separate the computation from the shape and structure. Is it complicated computation? No, but this approach supports future change by encapsulating the conditions and giving them a name. I think this is worth doing, what do you think?

damassi: Will do 👍

PR #3162, opened 2020-04-09
Makes corrections to collections filters analytics tracking

ashleyjelks:
Adds changes to tracking based on QA by @abhitip.

- Updates missing action types
- De-deduplicates previously selected, but currently applied filters.
- Adds additional unit tests

Screen Shot 2020-04-08 at 2 14 05 PM
Screen Shot 2020-04-08 at 2 14 21 PM
Screen Shot 2020-04-08 at 2 15 31 PM
Screen Shot 2020-04-09 at 12 44 24 AM

https://artsyproduct.atlassian.net/browse/FX-1896

#trivial

### DIFF
--- a/src/lib/Scenes/Collection/Helpers/FilterArtworksHelpers.ts
+++ b/src/lib/Scenes/Collection/Helpers/FilterArtworksHelpers.ts
@@ -1,15 +1,11 @@
 import { FilterArray } from "lib/utils/ArtworkFiltersStore"
+import { forOwn, omit } from "lodash"
 
 export const filterArtworksParams = (appliedFilters: FilterArray) => {
   // Default params
   const filterParams = {
     sort: "-decayed_merch",
     medium: "*",
-  }
-
-  const filterTypeToParam = {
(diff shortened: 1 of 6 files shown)


src/lib/Scenes/Collection/Helpers/FilterArtworksHelpers.ts
```diff
@@ -1,4 +1,5 @@
 import { FilterArray } from "lib/utils/ArtworkFiltersStore"
+import { forOwn, omit } from "lodash"
 
 export const filterArtworksParams = (appliedFilters: FilterArray) => {
```
sweir27, line 4: #superminor This function looks very similar to the one below (but I could be missing some nuance!).

If we have to make changes in the future, it might be worth extracting out the common bits into a separate function.

ashleyjelks: Good catch, I can make that change in the branch I'm working on now!

PR #3166, opened 2020-04-09
Add cron job to deploy beta at 6am UTC

ds300:
6am UTC is currently 2am in NY and 7am in London so it should be finished by the time I start working, and it shouldn't be started while y'all are still awake.

Paired on this with @brainbicycle during knowledge share 🎓

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -153,6 +153,15 @@ commands:
             - derived_data
 
 jobs:
+  deploy-nightly-beta:
+    executor:
+      name: node/default
+      tag: "12.13.1"
+    steps:
+      - checkout
+      - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -153,6 +153,15 @@ commands:
...
 jobs:
+  deploy-nightly-beta:
+    executor:
+      name: node/default
+      tag: "12.13.1"
+    steps:
+      - checkout
+      - run:
+          name: Deploy beta
+          command: make deploy
```
ashfurrow, line 164: Mmmm this makes sense. I was initially thinking we'd have the cron job itself deploy the beta, but I like this approach more 👍

PR #3167, opened 2020-04-09
Add typings for AROptions & Echo flags

ds300:
Paired on this with @brainbicycle during knowledge share 🎓


src/lib/Components/Bidding/Screens/BidResult.tsx
```diff
@@ -61,9 +61,8 @@ export class BidResult extends React.Component<BidResultProps> {
 
   exitBidFlow = async () => {
     if (this.props.bidderPositionResult.status === "LIVE_BIDDING_STARTED") {
-      const Emission = NativeModules.Emission || {}
```
ds300, line 64: this pattern was being used in a bunch of places. Might have been a remnant of the storybooks era, or an era before we had a jest mock for this.

src/lib/options.ts
```diff
@@ -1,18 +0,0 @@
-import { NativeModules } from "react-native"
-const { Emission } = NativeModules
```
ds300, line 2: this file wasn't referenced anywhere

typings/emission.d.ts
```diff
@@ -0,0 +1,41 @@
+import { NativeModulesStatic } from "react-native"
+declare module "react-native" {
+  interface NativeModulesStatic {
+    Emission: {
+      userID: string
```
ds300, line 5: this is called "module augmentation" in case you want to google the TS jargon :)

brainbicycle: 📚 🧠 Thanks!

ashfurrow:
Awesome stuff, thanks to you both!

PR #3182, opened 2020-04-13
Defers Sign In with Apple until next release

ashfurrow:
I agree that flagging it off is better than reverting the whole PR 👍

### DIFF
--- a/Artsy/View_Controllers/Login_and_Onboarding/Onboarding_stages/5-_Onboarding_Personalization/ARPersonalizeViewController.m
+++ b/Artsy/View_Controllers/Login_and_Onboarding/Onboarding_stages/5-_Onboarding_Personalization/ARPersonalizeViewController.m
@@ -326,13 +326,16 @@
 {
     [self addButtons];
 
-    if (@available(iOS 13.0, *)) {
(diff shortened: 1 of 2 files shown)


Artsy/View_Controllers/Login_and_Onboarding/Onboarding_stages/5-_Onboarding_Personalization/ARPersonalizeViewController.m
```diff
@@ -326,13 +326,16 @@ - (void)addThirdPartySignInButtons
 {
     [self addButtons];
 
+    BOOL signInWithApple = false;
     if (@available(iOS 13.0, *)) {
```
ashfurrow, line 330: I found it interesting that the compiler requires any availability checks like this one to include _only_ the availability check. When I wrote this:

```objc
if (signInWithApple && @available(iOS 13.0, *)) {
```

The compiler complained and refused to recognize it as an availability check.

ashfurrow: Oh good catch, I'll reverse the order of the `if` statements to get that behaviour 👍

PR #3183, opened 2020-04-13
[MX-283] - Home design tweaks

brainbicycle:
https://artsyproduct.atlassian.net/browse/MX-283

- Makes header sticky in home
- Bumps font size of rail subtitles

sticky-nav

### DIFF
--- a/src/lib/Scenes/Home/Home.tsx
+++ b/src/lib/Scenes/Home/Home.tsx
@@ -108,6 +108,12 @@ export class Home extends React.Component<Props, State> {
     return (
       <Theme>
         <View style={{ flex: 1 }}>
+          <Box mb={1} mt={2}>
+            <Flex alignItems="center">
+              <ArtsyLogoIcon scale={0.75} />
+            </Flex>
+          </Box>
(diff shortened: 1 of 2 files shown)


src/lib/Scenes/Home/Home.tsx
```diff
@@ -124,19 +130,8 @@ export class Home extends React.Component<Props, State> {
...
-                  <Flex alignItems="center">
-                    <ArtsyLogoIcon scale={0.75} />
-                  </Flex>
-                </Box>
-                <Separator />
-              </View>
-            }
             ListFooterComponent={() => <Spacer mb={3} />}
             keyExtractor={(_item, index) => String(index)}
-            style={{ overflow: "visible" }}
```
brainbicycle, line 139: Had to remove this to prevent content scrolling over nav bar, wanted to call out because it looks like you explicitly set.

ashfurrow: It was only copy/pasted from the previous implementation ¯\\_(ツ)_/¯

PR #3185, opened 2020-04-13
[MX-287] - Testflight distribute external

brainbicycle:
Explicitly provides beta test info on testflight uploads, attempted workaround for this issue:
https://github.com/fastlane/fastlane/issues/16294

Build 6.4.1 - 2020.4.14.10 was successfully uploaded and distributed to external testers from my local machine, there was a problem with the build that seems more likely related to something in CI not being present in my local set up than these changes but still would hold off merging until after submission to app store just in case.


shepting:
@brainbicycle Did this work for you to distribute to testers a previously uploaded build? Or does this upload a "just built" build and wait on Apple until it's ready to distribute?

We are running into the same issues and were wondering if this actually fixes the situation where we _only_ want to distribute to testers (and _not_ upload a build).

brainbicycle:
> @brainbicycle Did this work for you to distribute to testers a previously uploaded build? Or does this upload a "just built" build and wait on Apple until it's ready to distribute?
> 
> We are running into the same issues and were wondering if this actually fixes the situation where we _only_ want to distribute to testers (and _not_ upload a build).

@shepting I haven't tried the updating test info on an existing build and distributing, only uploading a just built build sorry to say.

ashfurrow:
I've kicked off a beta just to see if this works and will report back. Here's the build: https://circleci.com/gh/artsy/eigen/8267

PR #3187, opened 2020-04-14
[MX-276] - Update eigen side copy for home rails

brainbicycle:
https://artsyproduct.atlassian.net/browse/MX-276

Updates copy based on Sam's feedback for home rails.

Corresponding Metaphysics PR here:
https://github.com/artsy/metaphysics/pull/2311

### DIFF
--- a/src/lib/Scenes/Home/Components/ArtworkRail.tsx
+++ b/src/lib/Scenes/Home/Components/ArtworkRail.tsx
@@ -40,6 +40,8 @@ const ArtworkRail: React.FC<{ rail: ArtworkRail_rail }> = ({ rail }) => {
   let subtitle: React.ReactChild = null
   if (context?.__typename === "HomePageRelatedArtistArtworkModule") {
     subtitle = `Based on ${context.basedOn.name}`
+  } else if (rail.key === "recommended_works") {
+    subtitle = `Based on your activity on Artsy`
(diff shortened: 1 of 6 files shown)


src/lib/Scenes/Home/Components/ArtworkRail.tsx
```diff
@@ -40,6 +40,8 @@ const ArtworkRail: React.FC<{ rail: ArtworkRail_rail }> = ({ rail }) => {
   let subtitle: React.ReactChild = null
   if (context?.__typename === "HomePageRelatedArtistArtworkModule") {
     subtitle = `Based on ${context.basedOn.name}`
+  } else if (rail.key === "recommended_works") {
+    subtitle = `Based on your activity on Artsy`
```
brainbicycle, line 44: Wasn't sure about this, maybe should be served by Metaphysics?

ashfurrow: Yeah, I'm split here. Seems like this is specific to the app, otherwise it would use ["Recommended Works for You"](https://github.com/artsy/metaphysics/pull/2311/files#diff-1ba396aac04760abc3ec221529813ec8R48), right? 

@samchieng do you have a sense of if we want this to be identical on the site and app?

PR #3189, opened 2020-04-14
Trending artists collections rails DRAFT -- DO NOT MERGE

ashleyjelks:
#trivial


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>Missing Test Files:

- `src/lib/Scenes/Collection/Components/CollectionHubsRails/ArtistSeries/__tests__/ArtistSeriesRail-tests.tsx`
- `src/lib/Scenes/Collection/Components/CollectionHubsRails/__tests__/index-tests.tsx`

    </tr>
  </tbody>
</table>




<p align="right">
</p>

Artsy/View_Controllers/Login_and_Onboarding/Onboarding_stages/5-_Onboarding_Personalization/ARPersonalizeViewController.m
```diff
@@ -211,7 +211,7 @@ - (void)showViews
             self.searchResultsTable.headerPlaceholderText = @"TOP ARTISTS ON ARTSY";
             self.headerView.searchField.searchField.delegate = self;
             [self.headerView.searchField.searchField setPlaceholder:@"Search artist"];
-            [self populateTrendingArtistsAnimated:NO];
+            [self populateArtistSeriesAnimated:NO];
```
sweir27, line 214: Are these naming changes intentional?

data/schema.graphql
```diff
@@ -4507,7 +4507,7 @@ type Gene implements Node & Searchable {
     before: String
     last: Int
   ): GeneConnection
-  trendingArtists(sample: Int): [Artist]
+  ArtistSeries(sample: Int): [Artist]
```
sweir27, line 4510: Same here-- is this intentional?

src/lib/Components/FilterModal.tsx
```diff
@@ -212,7 +212,7 @@ export const FilterHeader = styled(Sans)`
 `
 
 export const BackgroundFill = styled(Flex)`
-  background-color: ${color("black10")};
+  background-color: white;
```
sweir27, line 215: Is this the default by any chance? Doesn't hurt to keep it here, but curious if we can remove.

src/lib/Components/Gene/About.tsx
```diff
@@ -54,7 +54,7 @@ export default createFragmentContainer(About, {
   gene: graphql`
     fragment About_gene on Gene {
       ...Biography_gene
-      trending_artists: trendingArtists {
+      trending_artists: ArtistSeries {
```
sweir27, line 57: I _think_ the metaphysics field is still called `trendingArtists` but it's possible I'm looking at something out-of-date. Is it possible this is unintentional?

src/lib/Scenes/Collection/Collection.tsx
```diff
@@ -62,19 +70,22 @@ export class Collection extends Component<CollectionProps, CollectionState> {
   }
 
   renderItem = ({ item: { type } }) => {
+    const { collection } = this.props
+    const { linkedCollections, isDepartment } = collection
+    const shouldShowDepartmentRails = NativeModules.Emission?.options?.AROptionsFilterCollectionsArtworks
```
sweir27, line 75: I believe we have the green light to ship the rails incrementally as they are finished, without needing to hide it behind a feature flag or wait for the filters work to be finished.

Was this added just so we have a little buffer on QA? If so, 👍 (but we should revisit this check soon 😄). If you're confident in shipping this change after this PR is merged, we could also just remove now! Will leave it up to you.

src/lib/Scenes/Collection/Collection.tsx
```diff
@@ -211,9 +222,14 @@ export const CollectionContainer = createFragmentContainer(Collection, {
       @argumentDefinitions(screenWidth: { type: "Int", defaultValue: 500 }) {
       id
       slug
+      isDepartment
```
sweir27, line 225: Woo! Finally using it!

src/lib/Scenes/Collection/Components/CollectionHubsRails/ArtistSeries/ArtistSeriesRail.tsx
```diff
@@ -0,0 +1,133 @@
...
+import SwitchBoard from "lib/NativeModules/SwitchBoard"
+import React, { useRef } from "react"
+import { FlatList, TouchableOpacity } from "react-native"
+import { createFragmentContainer, graphql } from "react-relay"
+import styled from "styled-components/native"
+
+interface ArtistSeriesRailProps {
+  collectionGroup: ArtistSeriesRail_collectionGroup
+}
+export const ArtistSeriesRail: React.SFC<ArtistSeriesRailProps> = props => {
```
sweir27, line 13: I think you mentioned you were going to look into this, but is it possible to re-use the Fairs/Auctions component from the home view here? I know you mentioned talking to the MX team so let me know if you've already explored.

src/lib/Scenes/Collection/Components/CollectionHubsRails/ArtistSeries/__tests__/ArtistSeriesRail-tests.tsx
```diff
@@ -0,0 +1,242 @@
...
+  CollectionName,
+} from "../ArtistSeriesRail"
+
+jest.unmock("react-relay")
+jest.mock("lib/NativeModules/SwitchBoard", () => ({
+  presentNavigationViewController: jest.fn(),
+}))
+
+it("renders without throwing an error", async () => {
+  await renderRelayTree({
```
sweir27, line 25: Will just calling `renderRelayTree` throw an error if it's broken?

src/lib/Scenes/Collection/Components/CollectionHubsRails/ArtistSeries/__tests__/ArtistSeriesRail-tests.tsx
```diff
@@ -0,0 +1,242 @@
...
+            ...ArtistSeriesRail_collectionGroup
+          }
+        }
+      }
+    `,
+    mockData: { marketingCollections: CollectionHubRailsArtistSeriesFixture },
+  })
+})
+
+describe("Trending Artists Rail", () => {
```
sweir27, line 45: These are awesome tests! ✨

src/lib/Components/FilterModal.tsx
ashfurrow: Palette uses the defaults for the idiom-specific component (`div` on web and `View` in RN), but I checked [the docs](https://reactnative.dev/docs/view-style-props) and they don't list any default. I would bet that it would use the `UIView` default from Apple's SDK, which is transparent.

src/lib/Scenes/Collection/Components/CollectionHubsRails/ArtistSeries/ArtistSeriesRail.tsx
```diff
@@ -0,0 +1,133 @@
...
+  const navRef = useRef<any>()
+  const { collectionGroup } = props
+
+  const renderItem = artworkPanelItem => {
```
ashfurrow, line 17: When possible, we'd like to keep all the rendering logic inline in the `return` statement. An issue here is that TypeScript can't infer the type of `artworkPanelItem`.

src/lib/Scenes/Collection/Components/CollectionHubsRails/ArtistSeries/ArtistSeriesRail.tsx
```diff
@@ -0,0 +1,133 @@
...
+        case 2:
+          return { style: { width: 89, height: 89 }, url: artwork?.node?.image?.url }
+      }
+    }
+
+    return (
+      <ArtworkImageContainer>
+        {
+          <Join separator={<Division />}>
+            <ImageView imageURL={imageMeta(artworks[0], 0).url} style={imageMeta(artworks[0], 0).style} />
```
ashfurrow, line 34: I think I see what `imageMeta` is doing here, but it took me a minute to figure it out. Since we only have 3 different images, it might be more straightforward to inline these attributes instead of defining styles ahead of time. Here is a similar component where we choose to repeat things for the sake of easier readability:

https://github.com/artsy/eigen/blob/810e9eddb887068f221595e92a348fef3c0d1f1e/src/lib/Scenes/Home/Components/SalesRail.tsx#L49-L65

src/lib/Scenes/Collection/Components/CollectionHubsRails/ArtistSeries/ArtistSeriesRail.tsx
```diff
@@ -0,0 +1,133 @@
...
+  const renderItem = artworkPanelItem => {
+    const artworks = artworkPanelItem?.artworksConnection?.edges.slice(0, 3)
+    const imageMeta = (artwork, index) => {
+      switch (index) {
+        case 0:
+          return { style: { width: 178, height: 181 }, url: artwork?.node?.image?.url }
+        case 1:
+          return { style: { width: 89, height: 90 }, url: artwork?.node?.image?.url }
+        case 2:
+          return { style: { width: 89, height: 89 }, url: artwork?.node?.image?.url }
```
ashfurrow, line 26: I'd also recommend turning these "magic numbers" into constants, like this one:

https://github.com/artsy/eigen/blob/810e9eddb887068f221595e92a348fef3c0d1f1e/src/lib/Scenes/Home/Components/SalesRail.tsx#L17

So we could have `IMAGE_HEIGHT = 178`, and then this line of code would be `IMAGE_HEIGHT / 2`, or above, `IMAGE_HEIGHT / 2 + 1`. This will give the numbers more significance and make it easier to reason about the layout.

src/lib/Scenes/Collection/Components/CollectionHubsRails/ArtistSeries/ArtistSeriesRail.tsx
ashfurrow: We chatted about this on Slack, but based on the designs we thought it'd be a good idea to start with a separate component, and then try to move to a common abstraction. MX can take that refactor work on, but we're happy to pair too 👍

sweir27: Awesome, thanks @ashfurrow !

Artsy/View_Controllers/Login_and_Onboarding/Onboarding_stages/5-_Onboarding_Personalization/ARPersonalizeViewController.m
ashleyjelks: It was an auto generated change.

data/schema.graphql
ashleyjelks: In the first iterations of this branch I used trendingArtists and then changed it to make the verbiage in the Relay response--Artist series. These changes were generated.

sweir27: This file is the actual graphQL schema from metaphysics! Is it possible that a find + replace all accidentally updated this file as well?

I wouldn't expect to see this file changed unless there was a corresponding change in metaphysics.

src/lib/Scenes/Collection/Components/CollectionHubsRails/ArtistSeries/ArtistSeriesRail.tsx
ashleyjelks: @ashfurrow since the ArtistSeriesRail on Collections and the FairRail and SalesRail on Home are using so many of the same components for styling I went ahead and moved the those to the CardRail component file. I confirmed the behavior and styling of the Home rail components remain the same, will post some screenshots in the MX channel as well.

src/lib/Scenes/Collection/Components/CollectionHubsRails/ArtistSeries/ArtistSeriesRail.tsx
```diff
@@ -0,0 +1,129 @@
...
+    <ArtistSeriesWrapper>
+      <CollectionName size="4" mb={2} ml={2}>
+        {collectionGroup.name}
+      </CollectionName>
+      <CardRailFlatList<ArtistSeriesItem>
+        data={collectionGroup?.members}
+        keyExtractor={(_item, index) => String(index)}
+        initialNumToRender={3}
+        renderItem={({ item: result, index }) => {
+          const artworkImageURLs = result?.artworksConnection?.edges?.map(edge => edge.node.image.url)
```
sweir27, line 37: As long as `ImageView` can handle a `null` `imageURL` (I think it shows a gray box in that case?), you can update this to `edge?.node?.image?.url`. I think it also makes sense to use the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) here (otherwise, we'll hit errors when trying to index into that array):
```
const artworkImageURLs = result?.artworksConnection?.edges?.map(edge => edge?.node?.image?.url) ?? []
```

☝️ this behavior can be verified with tests!

ashfurrow:
Arrrrgh, sorry for the duplicated comments. Looks like GitHub was having issues.

PR #3193, opened 2020-04-15
[MX-214] - Fix sign up on initial sign up with apple failure

brainbicycle:
- Reenable sign in with apple
- Workaround for a bug that occurs when first auth fails

Apple only shares user email and name on first authentication so if initial authentication fails to create an artsy user for whatever reason (previously linked account, network error) we need to store the user name and email to enable sign up later. This does that by storing in keychain and deleting once a user is created successfully.

### DIFF
--- a/Artsy/Networking/API_Modules/ARUserManager.h
+++ b/Artsy/Networking/API_Modules/ARUserManager.h
@@ -18,6 +18,12 @@ extern NSString *const ARUserSessionStartedNotification;
 - (User *)currentUser;
 - (void)storeUserData;
 
+@property (nonatomic, strong) NSString *appleEmail;
+@property (nonatomic, strong) NSString *appleDisplayName;
+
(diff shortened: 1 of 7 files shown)


Artsy/Networking/API_Modules/ARUserManager.h
```diff
@@ -18,6 +18,12 @@ extern NSString *const ARUserSessionStartedNotification;
 - (User *)currentUser;
 - (void)storeUserData;
 
+@property (nonatomic, strong) NSString *appleEmail;
+@property (nonatomic, strong) NSString *appleDisplayName;
```
ashfurrow, line 22: I would say these should be readonly, since the getters are computed anyway 👍 

```suggestion
@property (nonatomic, readonly) NSString *appleEmail;
@property (nonatomic, readonly) NSString *appleDisplayName;
```

PR #3196, opened 2020-04-16
Moves to Circle CI jobs to queue MP update

ashfurrow:
Instead of waiting for another job to complete _in_ a job ([which was failing due to a timeout](https://circleci.com/gh/artsy/eigen/8268)), we can use the `requires` on workflows to just not start the MP update job until the beta is deployed.


brainbicycle:
Makes sense to me 👍

PR #3198, opened 2020-04-16
Removes ARSwitchBoardModule requirement that fromViewController lookup succeeds

ashfurrow:
Yuki was running into a problem where the `RCTRootView`'s `reactViewController` property was `nil`. This was because the code was being invoked from the Bid Flow interface, which uses its own navigation controller (that's my intuition, anyway).

The thing is, `fromViewController` isn't even used in the `presentNavigationViewController` and `presentModalViewController` blocks. Removing the parameter altogether would be a big refactor, so I opted instead to remove the `_Nonnull` specifier and the `NSParameterAssert` that was causing the crash. I also prefixed the parameter names in the blocks in `ARAppDelegate+Emission.m` to indicate that `_fromViewController` isn't used.

### DIFF
--- a/Artsy/App/ARAppDelegate+Emission.m
+++ b/Artsy/App/ARAppDelegate+Emission.m
@@ -241,13 +241,13 @@ FollowRequestFailure(RCTResponseSenderBlock block, BOOL following, NSError *erro
         });
     };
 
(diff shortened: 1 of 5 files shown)


yuki24:
Also here is a screenshot of the behaviour:

faq

PR #3199, opened 2020-04-17
Add null check before attempting to store in keychain #trivial

brainbicycle:
Another silly one, need to check values are non-nil before storing in keychain otherwise 💥

### DIFF
--- a/Artsy/View_Controllers/Login_and_Onboarding/AROnboardingViewController.m
+++ b/Artsy/View_Controllers/Login_and_Onboarding/AROnboardingViewController.m
@@ -723,7 +723,7 @@
     NSString *authEmail = email ? email : ARUserManager.sharedManager.appleEmail;
     NSString *authName = name ? name : ARUserManager.sharedManager.appleDisplayName;
 
-    if (authName != nil && authEmail != nil) {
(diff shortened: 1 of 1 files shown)


Artsy/View_Controllers/Login_and_Onboarding/AROnboardingViewController.m
```diff
@@ -722,7 +722,10 @@ - (void)appleSuccessWithUID:(NSString * _Nonnull)appleUID email:(NSString * _Nul
     // app in settings, we need to store this info in case first auth fails to create an Artsy user
     NSString *authEmail = email ? email : ARUserManager.sharedManager.appleEmail;
     NSString *authName = name ? name : ARUserManager.sharedManager.appleDisplayName;
-    [ARUserManager.sharedManager storeAppleDisplayName:authName email:authEmail];
+
+    if (authName != nil && authEmail != nil) {
```
ashfurrow, line 726: This makes total sense – but this change would make it a bit more idiomatic Objective-C: 

```suggestion
    if (authName && authEmail) {
```

I still kind of wish Swift would add this back in 😅

PR #3200, opened 2020-04-17
[MX-138] - Auctions empty state fixes - Hold merging

brainbicycle:
https://artsyproduct.atlassian.net/browse/MX-138

There is actually an empty state defined for this case but it had a couple problems, 
1. It uses react-native-webview which was not included in our npm_podspecs 
2. The link handler was not bound to this so links weren't working
3. There was no header so text was being obscured by back button

Here is what it looks like now:
auctions-empty-state

Please hold on merging, waiting for @samchieng feedback to see if this is what we want to go with.


Artsy.xcodeproj/project.pbxproj
```diff
@@ -4676,6 +4678,8 @@
 				"${TARGET_BUILD_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}/Unica77LL-Regular.otf",
 				"${TARGET_BUILD_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}/Chevron_Black.png",
 				"${TARGET_BUILD_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}/Chevron_Black@2x.png",
+				"${TARGET_BUILD_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}/Emission.js",
+				"${TARGET_BUILD_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}/assets",
```
ashfurrow, line 4682: This isn't a huge deal, but these get added when you've run `yarn bundle` locally. The bundles themselves are ignored by git, but their presence modifies the Xcode project when `pod install` gets ran. I'd recommend removing these before they churn in someone else's PR.

brainbicycle: Sounds good, I'll remove, thanks for letting me know 👍

brainbicycle:
Closing in favor of https://github.com/artsy/eigen/pull/3201

PR #3206, opened 2020-04-20
Fixes a bug where the "Register To Bid" button displays that IDV is required after begin verified

yuki24:
fixes https://artsyproduct.atlassian.net/browse/AUCT-986

We have talked about this a couple times, but the current user object is locally stored in the phone and doesn't get updated unless the user explicitly logs out and back in. As a result of this, the identity verification status stays unchanged even after the user completes the verification. This PR adds another call to `/me` to get fresh data about the user to displays the latest information.

We may want to iterate on the `/me` call in general once we merge this PR, so we could use the up-to-date data everywhere else rather than throwing it away after the sale screen is gone. Let me know what you think.

## Todos

 * [x] Add more test coverage
 * [x] ~~Get error handling right: https://github.com/artsy/eigen/pull/3206/files#diff-cc7a410bdaac96a51f5a008b854d4788L169-R200~~ left a comment there
 * [x] ~~Fix the failing specs (they fail only locally due to snapshot mismatch, but they don't in CI)~~ couldn't find the root cause, but might be worth looking into later.

## Screenshots

AUCT-986-large

### DIFF
--- a/Artsy/View_Controllers/Auction/AuctionNetworkModel.swift
+++ b/Artsy/View_Controllers/Auction/AuctionNetworkModel.swift
@@ -5,6 +5,7 @@ protocol AuctionNetworkModelType {
     var bidders: [Bidder] { get }
 
     func fetch() -> Observable<Result<SaleViewModel>>
+    func fetchMe() -> Observable<Result<User>>
     func fetchSale(_ saleID: String) -> Observable<Result<Sale>>
     func fetchSaleArtworks(_ saleID: String) -> Observable<Result<[SaleArtwork]>>
     func fetchBidders(_ saleID: String) -> Observable<Result<[Bidder]>>
@@ -22,6 +23,8 @@ class AuctionNetworkModel {
 
     // Each one of these network models performs their request to fetch exactly one thing,
     // and then store it locally.
+    lazy var meNetworkModel: AuctionUserNetworkModel
+        = AuctionUserNetworkModel()
     lazy var saleNetworkModel: AuctionSaleNetworkModelType
         = AuctionSaleNetworkModel()
     lazy var saleArtworksNetworkModel: AuctionSaleArtworksNetworkModelType
@@ -39,6 +42,10 @@ class AuctionNetworkModel {
 extension AuctionNetworkModel: AuctionNetworkModelType {
     var bidders: [Bidder] {
         return bidderNetworkModel.bidders
+    }
+
+    func fetchMe() -> Observable<Result<User>> {
+        return self.meNetworkModel.fetchCurrentUser()
     }
 
     func fetchSale(_ saleID: String) -> Observable<Result<Sale>> {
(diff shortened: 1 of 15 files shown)


Artsy/View_Controllers/Auction/AuctionNetworkModel.swift
```diff
@@ -162,57 +176,66 @@ extension AuctionNetworkModel: AuctionNetworkModelType {
...
 
-        case (.success, .error(let error), _):
+        case (.success, .error(let error), _, _):
             return .error(error)
 
-        case (.success(_), .success(_), .error(let error)):
+        case (.success(_), .success(_), .error(let error), _):
+            return .error(error)
+
+        case (.success(_), .success(_), .success(_), .error(let error)):
```
yuki24, line 198: Is this good enough to handle errors? I basically just guessed what it should look like based on what the existing code looks like, so I could be very wrong here...

ashfurrow: Yeah, it is. Each one of the network results (either `.success` or `.error`) is captured here. It's like a Scala `case` statement where the `_` matches anything. The compiler would also warn about any missing cases, so we should be covered.

In an ideal world, we would somehow `flatMap` these instead of having this ridiculous `switch` statement (or better yet, we'd rewrite this in React Native) but unfortunately we don't live in an ideal world 😭

Artsy/View_Controllers/Auction/AuctionNetworkModel.swift
```diff
@@ -162,57 +176,66 @@ extension AuctionNetworkModel: AuctionNetworkModelType {
...
                     promotedSaleArtworks: promotedSaleArtworks,
                     bidders: bidders,
-                    lotStandings: lotStandings
+                    lotStandings: lotStandings,
+                    me: me
                 )
             )
 
-        case (.error(let error), .error, _):
+        // TODO: not sure if I understand this well, talk to MX.
```
ashfurrow, line 184: Tried to explain how this works a bit in the other comment, but happy to explain further.

Artsy/View_Controllers/Live_Auctions/LiveAuctionLotSetViewController.swift
```diff
@@ -259,7 +259,7 @@ class LiveAuctionLotSetViewController: UIViewController {
     @objc func moreInfo() {
         guard let sale = saleNetworkModel.sale else { return }
 
-        let saleVM = SaleViewModel(sale: sale, saleArtworks: [], bidders: biddersNetworkModel.bidders, lotStandings: [])
+        let saleVM = SaleViewModel(sale: sale, saleArtworks: [], bidders: biddersNetworkModel.bidders, lotStandings: [], me: User.current())
```
ashfurrow, line 262: 👍

ashfurrow:
Feel free to merge when ready, possibly removing that `TODO` comment 👍

Artsy/View_Controllers/Auction/AuctionNetworkModel.swift
yuki24: Got it, thanks for the explanation. Yeah I understand the circumstances. I also think that this isn't that bad. In fact, it was actually great that Xcode pointed it out at compile time. I don't have strong needs to re-write the whole screen in RN, so this should be god for the time being.

yuki24:
Thanks both for reviewing :+1: merging...

PR #3211, opened 2020-04-21
Extracts component for a SingleSelectOption to be used with collection filters

sweir27:
This PR does two things:
- Extracts the common visual treatment for the `MediumOptions` and `SortOptions` screens into a single component, `SingleSelectOption`. This can be used for other filters which only allow a single item to be selected.
- Tries to DRY up the places where we're defining the list of sort and medium options. I feel like we could do even better here, but not sure how worth it it is.

Proof that things still look OK:
single-option

#skip_new_tests
#trivial


src/lib/Components/FilterModal.tsx
```diff
@@ -104,33 +108,36 @@ interface FilterOptionsProps {
...
+
+  const navigateToNextFilterScreen = NextComponent => {
     navigator.push({
-      component: FilterScreenComponents[filterOption],
+      component: NextComponent,
     })
   }
-  const [filterOptions] = useState<FilterOptions>([
+
+  const filterOptions: FilterOptions[] = [
```
sweir27, line 129: I made this a little simpler-- it didn't seem necessary to use state here, as these values never change.

src/lib/Scenes/Collection/Helpers/FilterArtworksHelpers.ts
```diff
@@ -66,6 +67,19 @@ enum ArtworkSorts {
   "Artwork year (ascending)" = "year",
 }
 
+export type SortOption = keyof typeof ArtworkSorts
+
+export const OrderedArtworkSorts: SortOption[] = [
```
sweir27, line 72: I tried to move all of these definitions into one place. Now we define an object for each sorting option that maps the filter "names" to their underlying query params _and_ an array of filter "names" so they can be ordered correctly in the UI.

There's definitely a way to combine these data types, I just wasn't sure if it was worth the overhead. Let me know what you think.

src/lib/Components/ArtworkFilterOptions/SingleSelectOption.tsx
```diff
@@ -0,0 +1,107 @@
...
+      </FilterHeader>
+      <Flex mb={120}>
+        <FlatList<SingleSelectOptions>
+          initialNumToRender={12}
+          keyExtractor={(_item, index) => String(index)}
+          data={filterData}
+          renderItem={({ item }) => (
+            <Box>
+              {
+                <SingleSelectOptionListItemRow onPress={() => selectOption(item)}>
```
sweir27, line 63: cc @ds300 in case you know 😄 

I'm hitting a TS error on this line, I _think_ because the list that powers this `FlatList` is a union, but by the time we're selecting an item TS isn't sure which item in the union we're referring to.

I imagine there's a better way to structure this, just not sure what it is. Let me know what you think.

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>Missing Test Files:

- `src/lib/Components/ArtworkFilterOptions/__tests__/SingleSelectOption-tests.tsx`

    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/Components/ArtworkFilterOptions/SingleSelectOption.tsx
sweir27: 🙌 good point, thanks!

src/lib/Scenes/Collection/Helpers/FilterArtworksHelpers.ts
```diff
@@ -81,7 +95,31 @@ enum MediumFilters {
...
+  "Works on paper",
+  "Design",
+  "Drawing",
+  "Installation",
+  "Film & video",
+  "Jewelry",
+  "Performance art",
+]
+
+export type MediumOption = keyof typeof MediumFilters
```
ashleyjelks, line 113: `keyof` here is a cleaner way of declaring `type MediumOption = "All" | "Painting" | "Photography`... ??

src/lib/Components/ArtworkFilterOptions/SingleSelectOption.tsx
```diff
@@ -0,0 +1,107 @@
...
+  margin-top: ${space(2)};
+  margin-left: ${space(2)};
+`
+
+export const InnerOptionListItem = styled(Flex)`
+  flex-direction: row;
+  justify-content: space-between;
+  flex-grow: 1;
+  align-items: flex-end;
+  padding: ${space(2)}px;
```
ashleyjelks, line 103: Might be worth making all the padding and margin CSS values either by adding `px` where or removing all instances of `px`.

src/lib/Scenes/Collection/Helpers/FilterArtworksHelpers.ts
sweir27: Yep! Exactly. (Somewhat satisfying answer to that [here](https://stackoverflow.com/questions/55377365/what-does-keyof-typeof-mean-in-typescript))

src/lib/Components/ArtworkFilterOptions/SingleSelectOption.tsx
sweir27: Oh interesting-- will fix!

PR #3215, opened 2020-04-21
Viewing rooms tracking [GALL-2785]

mdole:
## Update 5/8

Ok! Been a longer road to get here than anticipated, but I think this one's ready for review.

This PR adds tracking + tracking tests for taps and impressions in viewing rooms. That's pretty much it!

---

This PR will implement tracking for Viewing Rooms. Still a bit of clarity needed around exactly what we're tracking and how we're doing it; I'll check in with Louis and continue to update here.

[Jira ticket](https://artsyproduct.atlassian.net/browse/GALL-2785)


src/lib/Scenes/ViewingRoom/Components/ViewingRoomArtworkRail.tsx
```diff
@@ -48,12 +58,20 @@ export const ViewingRoomArtworkRail: React.FC<ViewingRoomArtworkRailProps> = pro
...
           renderItem={({ item }) => (
             <ArtworkCard
-              onPress={() =>
+              onPress={() => {
+                tracking.trackEvent({
+                  action_name: Schema.ActionNames.TappedArtworkGroup,
+                  context_module: Schema.ContextModules.ViewingRoomArtworkRail,
+                  destination_screen: Schema.PageNames.ArtworkPage,
+                  destination_screen_owner_id: "TKTK",
+                  destination_screen_owner_slug: "TKTK",
```
mdole, line 67: @ds300 do we need to fetch these things from Metaphysics, or is it possible to get them from the destination page? I think this is a new pattern for Eigen tracking

src/lib/Scenes/ViewingRoom/ViewingRoom.tsx
```diff
@@ -156,20 +162,29 @@ export const ViewingRoomFragmentContainer = createFragmentContainer(ViewingRoom,
...
-      `}
-      cacheConfig={{ force: true }}
-      variables={{
-        viewingRoomID: "1489f6b2-39f2-449d-9cc2-6baa5782c756",
+    <ProvideScreenTracking
+      info={{
+        context_screen: Schema.PageNames.ViewingRoom,
+        context_screen_owner_type: Schema.OwnerEntityTypes.ViewingRoom,
+        context_screen_owner_id: "TKTK",
+        context_screen_owner_slug: "TKTK",
```
mdole, line 170: @ds300 I found that I couldn't set up `const tracking = useTracking()` inside the `ViewingRoom` component - I was getting an error saying that you can't call `useTracking` when it's not inside a tracking context. This was my workaround; is this kosher from a best-practices standpoint?

src/lib/Scenes/ViewingRoom/ViewingRoom.tsx
```diff
@@ -71,51 +73,53 @@ export const ViewingRoom: React.FC<ViewingRoomProps> = props => {
...
+    <Theme>
+      <View style={{ flex: 1 }} ref={navRef as any /* STRICTNESS_MIGRATION */}>
+        <FlatList<ViewingRoomSection>
+          onViewableItemsChanged={useCallback(({ viewableItems }) => {
+            if (viewableItems.find((viewableItem: ViewToken) => viewableItem.item.key === "pullQuote")) {
+              if (!hasViewedPullQuote.current) {
+                tracking.trackEvent({
+                  action_name: Schema.ActionNames.PullQuoteImpression,
+                })
+                hasViewedPullQuote.current = true
```
mdole, line 85: This is one way to not fire tracking every time we see the pull quote. I think it works OK, but open to suggestions if a better way exists.

While doing this, realized that we're running `setDisplayViewWorksButton(true)` every time `viewableItems` includes the pull quote. Is that a bad idea from a performance perspective? I played around with making `displayViewWorksButton` a `ref` instead, but then of course we didn't rerender after setting it to `true` so the button never showed up 😅

src/lib/Scenes/ViewingRoom/Components/ViewingRoomArtworkRail.tsx
mdole: discussed IRL; we get these from MP. Implemented in 0c1d97d

src/lib/Scenes/ViewingRoom/ViewingRoom.tsx
mdole: Discussed IRL; after #3236 was merged, this became a non-issue

src/lib/Scenes/ViewingRoom/ViewingRoom.tsx
mdole: Discussed IRL and refactored a bit! see 1740d40

src/lib/Scenes/ViewingRoom/ViewingRoom.tsx
```diff
@@ -26,8 +27,20 @@ interface ViewingRoomSection {
...
   const navRef = useRef()
+  const tracking = useTracking()
+  const trackBodyImpression = useCallback(
+    once(() =>
+      tracking.trackEvent({
+        action_name: Schema.ActionNames.BodyImpression,
+        context_screen: Schema.PageNames.ViewingRoom,
+        context_screen_owner_type: Schema.OwnerEntityTypes.ViewingRoom,
+        context_screen_owner_id: viewingRoom.internalID,
+        context_screen_owner_slug: viewingRoom.slug,
```
mdole, line 38: I felt like I shouldn't have to specify these on this event since this is wrapped in a `<ProvideScreenTracking>` component that includes these values in its `info` object - but without them, when I checked what was printed to the console when that event was fired, it only included `action_name`:

```
[Fri May 08 2020 13:03:42.163]  LOG      [Event tracked] {
  "action_name": "bodyImpression",
}
```
What am I missing here?

src/lib/Scenes/ViewingRoom/ViewingRoom.tsx
```diff
@@ -98,12 +116,22 @@ export const ViewingRoom: React.FC<ViewingRoomProps> = props => {
...
             <ViewWorksButtonContainer>
               <TouchableWithoutFeedback
-                onPress={() =>
+                onPress={() => {
+                  tracking.trackEvent({
+                    action_name: Schema.ActionNames.TappedViewWorksButton,
+                    context_screen: Schema.PageNames.ViewingRoom,
+                    context_screen_owner_type: Schema.OwnerEntityTypes.ViewingRoom,
+                    context_screen_owner_id: viewingRoom.internalID,
+                    context_screen_owner_slug: viewingRoom.slug,
```
mdole, line 125: Same question here as above

src/lib/Scenes/ViewingRoom/Components/__tests__/ViewingRoomArtworkRail-tests.tsx
```diff
@@ -42,7 +44,33 @@ describe("ViewingRoomSubsections", () => {
...
+      "/viewing-room/gallery-name-viewing-room-name/artworks"
+    )
+    expect(useTracking().trackEvent).toHaveBeenCalledWith({
+      action_name: Schema.ActionNames.TappedArtworkGroup,
+      context_module: Schema.ContextModules.ViewingRoomArtworkRail,
+      destination_screen: Schema.PageNames.ViewingRoomArtworks,
+      destination_screen_owner_id: "2955ab33-c205-44ea-93d2-514cd7ee2bcd",
+      destination_screen_owner_slug: "gallery-name-viewing-room-name",
+      type: "header",
+    })
```
mdole, line 73: Felt like I was doing a lot of copy/pasting for these navigation + tracking tests - is there a nicer pattern that I could be using?

pvinis: I was reading this and I was thinking "wow, that's an intense-looking test." Probably we could wrap this in a some kind of helper func if it is repeated all over the place. I don't know if there is something already available.

src/lib/Scenes/ViewingRoom/ViewingRoom.tsx
pvinis: Hm, this should work as you suggest. I'll take a look.

src/lib/Scenes/ViewingRoom/Components/ViewingRoomArtworkRail.tsx
```diff
@@ -46,12 +54,20 @@ export const ViewingRoomArtworkRail: React.FC<ViewingRoomArtworkRailProps> = pro
...
-              onPress={() =>
+              onPress={() => {
+                tracking.trackEvent({
+                  action_name: Schema.ActionNames.TappedArtworkGroup,
+                  context_module: Schema.ContextModules.ViewingRoomArtworkRail,
+                  destination_screen: Schema.PageNames.ArtworkPage,
+                  destination_screen_owner_id: item?.node?.internalID,
+                  destination_screen_owner_slug: item?.node?.slug,
+                  type: "thumbnail",
+                })
```
pvinis, line 65: I always feel these tracking objects are getting in the way of the actual code when reading. Would it make sense to keep them on the bottom of the file like the graphql stuff, something like
```
const tracks = {
tappedArtworkGroup: item => {
                  action_name: Schema.ActionNames.TappedArtworkGroup,
                  context_module: Schema.ContextModules.ViewingRoomArtworkRail,
                  destination_screen: Schema.PageNames.ArtworkPage,
                  destination_screen_owner_id: item?.node?.internalID,
                  destination_screen_owner_slug: item?.node?.slug,
                  type: "thumbnail",
                },
otherEvent: { ... },
}
```
and then in the place we use them, like here, we just do `tracking.trackEvent(tappedArtworkGroup(item))` and it's just one line? Maybe even `tracks.tappedArtworkGroup(item)` if we modify a bit more.

@artsy/mobile-experience tagging for opinions.

src/lib/Scenes/ViewingRoom/Components/__tests__/ViewingRoomArtworkRail-tests.tsx
```diff
@@ -42,7 +44,33 @@ describe("ViewingRoomSubsections", () => {
...
+      return result
+    })
+    tree.root.findByType(SectionTitle).props.onPress()
+
+    expect(SwitchBoard.presentNavigationViewController).toHaveBeenCalledWith(
+      expect.anything(),
+      "/viewing-room/gallery-name-viewing-room-name/artworks"
+    )
+    expect(useTracking().trackEvent).toHaveBeenCalledWith({
+      action_name: Schema.ActionNames.TappedArtworkGroup,
```
mikehrom, line 67: i believe all events need to have a `context_screen_owner_type` as a minimum, and then `context_screen_owner_id` and `context_screen_owner_slug` if applicable

src/lib/Scenes/ViewingRoom/Components/__tests__/ViewingRoomArtworkRail-tests.tsx
```diff
@@ -42,7 +44,33 @@ describe("ViewingRoomSubsections", () => {
...
+      context_module: Schema.ContextModules.ViewingRoomArtworkRail,
+      destination_screen: Schema.PageNames.ViewingRoomArtworks,
```
mikehrom, line 69: `destination_screen_owner_type` as well. we've started defining these in detail in cohesion: https://cohesion.artsy.net/interfaces/_schema_events_tap_.tappedartworkgroup.html

src/lib/Scenes/ViewingRoom/Components/ViewingRoomArtworkRail.tsx
brainbicycle: I think it makes sense, FWIW we are going to have similar events for all home rails so I was planning on having a helper function that just creates these events depending on the home rail type so hopefully that will clean things up a little. As a general pattern though I think defining all the tracking events at end of file or a separate file makes sense to me, I prefer to inline as well. Also we are planning on moving to use https://github.com/artsy/cohesion in the future so potentially can add any shared boilerplate analytics stuff there as well.

src/lib/Scenes/ViewingRoom/Components/__tests__/ViewingRoomArtworkRail-tests.tsx
louislecluse: hm yes you are right this should be `destination_screen_owner_type`

src/lib/Scenes/ViewingRoom/ViewingRoom.tsx
mdole: Pavlos and I looked at this together. From what we can tell, the issue is that if the `useTracking` hook is in the same component whose returned component has the  `<ProvideScreenTracking />` wrapper, the hook is called before the wrapper is rendered so it doesn't include any of the properties from the wrapper. @pvinis anything you'd add to that?

src/lib/Scenes/ViewingRoom/Components/__tests__/ViewingRoomArtworkRail-tests.tsx
mdole: yep! the actual events DO have those properties - they come from the `<ProvideScreenTracking />` that wraps the whole dealio. they're not reflected in the test because when we test we don't have that wrapper. does leave a bit of a hole in the tests though...not sure if there's a good way to handle that

Example tracking call:

Screen Shot 2020-05-08 at 4 43 39 PM

src/lib/Scenes/ViewingRoom/Components/__tests__/ViewingRoomArtworkRail-tests.tsx
mdole: Good catch! updated in 64bbc3a

src/lib/Scenes/ViewingRoom/Components/ViewingRoomArtworkRail.tsx
mdole: Good thought! Took a stab at this in f58e6c1; I'm sure we can figure out a helper function that works more broadly but it definitely feels a bit cleaner at least

src/lib/Scenes/ViewingRoom/Components/__tests__/ViewingRoomArtworkRail-tests.tsx
mdole: Updated to use the `tracks` functions in 4cb030b!

pvinis: Sweeeeeeeet!

src/lib/Scenes/ViewingRoom/ViewingRoom.tsx
pvinis: Yup. I'd like to have this documented somewhere, so I might make a quick PR for this. Hopefully, we will not fall on this edge case again/often. Nice debugging session though :D

PR #3217, opened 2020-04-22
Use full version on .nvmrc

pvinis:
I think this #trivial change can get some discussion, but I think nothing changes when using nvm. I use asdf, which unfortunately required a full version. If this doesn't cause problems for others, then that would be great to merge. If it does, then let's talk and I'll try to find a better way for all of us, or worst case I'll make some script or something just for me locally.

PS. I want to suggest/show off asdf at some point, as an option to replace nvm, rvm etc, after seeing the line https://github.com/artsy/potential/blob/master/scripts/setup#L134, maybe in some KS or something.


ds300:
I think this should be OK, except on CI it's configured for "12.13.1". Do you mind updating `config.yml`?

Also we'll need to broadcast this change to folks in the #front-end-ios app in case people run into trouble after this change. Maybe point people to this: https://stackoverflow.com/questions/23556330/run-nvm-use-automatically-every-time-theres-a-nvmrc-file-on-the-directory

ashfurrow:
KS works for sure – we also have a weekly Engineering "Lunch & Learn" meeting on Wednesday – [on days without a specific presentation](https://github.com/artsy/README/blob/master/events/show-and-tell.md), we show off smaller things we've learned and this would be great there too 👍

PR #3218, opened 2020-04-22
Scroll rails on refresh

pvinis:
(This is my first actual PR, and it touches a few places, so I'll try to do what would make sense for me, and please let me know if there is anything I can do differently to explain the changes, to show the changes, in the way of the code, or anything else. Comment with feedback or we can pair to review. Here we go!)

This PR makes the rails in the home screen scroll all the way to their beginning, whenever the user pulls to refresh.

To implement that in the func components was simple, like https://github.com/artsy/eigen/pull/3218/files#diff-d8f4c64d1e1aa4e2813c0e2b1eed8fc8R41. Adding the scroll handle, adding the ref from home to the comp and from the comp to the list and done. Also adding a ref from our "list components" to the actual FlatList inside.

For the class components it was a bit trickier. Here https://github.com/artsy/eigen/pull/3218/files#diff-9903b26c17949578a7c60b5ec0b20397R30 for example is the changes for the class comp, all good. Adding the function and then the refs.
One of the problems I faced was with naming the ref `scrollRef` like in the func comps. I tried and I could do it but with some forwarding and weird extra code, so I gave up on that. `ref` worked great but the relay container complained for types, so I found out there is `componentRef` that gives you a ref to the inside component. I used that and it worked great, but the types were still crying. I saw that the way `@types/react-relay` handles the type of `componentRef` is the "old way", like `ref={r => this.someRef = r}`, and not the "new way" like `ref={someRef}`. So I added a patch for that. That was the first "extra" change I had to do.

The other extra change was on the tracking. I found that `@track()` on top of `ArtistRail` was messing with the refs, I couldn't figure out why. I noticed that if I comment it out, the rail scrolling via ref was working. I decided to change that component from a class to a func, in order to make things easier. I turned the the class into a func, and then I had to add the tracking back. To add it back I used `useTracking` and I wrapped it so we `postEvent` as well. That also helped with the `Home` component, and the `@screenTrack` decorator as well.

Lastly, the two things I am not super happy with but I think are fine, are:
- that in `ArtistRail` there was `this.setState({ artists }, resolve)` which turned into ```
setArtists(nextArtists)
resolve()
```
which is definitely not the same, but I couldn't make it break, and
- the fact that in the `Home` component we have `scrollRef` and `componentRef` for the same thing, because of differences between class and func components.

I went through some ideas and that's the best I could come up with. I'm open to suggestions if there is a better way, and I'm open to accepting this until we have a more consistent set of rail components. 🤷‍♂️ 

To test this:
- Be in Home screen
- Scroll some rails
- Pull to refresh. Now all rails should be back to their beginning.
- (extra) Check the console or debugger to see that the track events for Home screenTracking and for ArtistRail handleFollowChange tracking are showing up. (Home was there, followChange was not printing, but now it should.)

### CODE REVIEW

**Comment by pvinis** [comment]
This is ready again. I had to merge and fix some conflicts with the strictness.

File: src/lib/utils/track/index.ts [hunk 1]
```diff
@@ -86,7 +86,12 @@ export interface Track<P = any, S = null, T extends Schema.Global = Schema.Entit
  *      }
  *      ```
  */
-export const track: Track = _track
+export const track: Track = (trackingInfo, options) => {
```
ds300, line 89: 🎉

src/lib/Scenes/Home/Home.tsx
```diff
@@ -19,96 +19,97 @@ import SwitchBoard from "lib/NativeModules/SwitchBoard"
...
+    } as const,
+    ...compact(flatten(zip(drop(artworkRails, 3), artistRails))),
+  ]
+
+  const scrollRefs = useRef<Array<RefObject<RailScrollRef>>>(rowData.map(_ => createRef()))
+  const scrollRailsToTop = () => scrollRefs.current.forEach(r => r.current?.scrollToTop())
+
+  const [isRefreshing, setIsRefreshing] = useState(false)
+  const handleRefresh = async () => {
+    scrollRailsToTop()
```
ds300, line 88: Let's do this after the refetch instead of before.

pvinis:
Scrolling after fetching, much nicer now indeed.

PR #3219, opened 2020-04-22
Support staging of partials/hunks of files

pvinis:
I stage hunks all the time, rather than full files. I tried that while working on another PR and the whole file got staged under my feet 😬! `git` betrayed me, I thought! Then I tried some more and another try committed my staged change but reverted my other one!! Even worse betrayal!

I did some research and apparently the versions of `lint-staged` and `husky` we had didn't have great support of partials. I updated both, and tried to do a small change in a file in two parts. Works great now with this #trivial PR.

How to test this:
- `yarn install` to make sure we have the latest versions
- Make two changes in a file
(You could try the same I did in this PR. Add a closing `</FlatList>` tag in the file, and also add or remove an empty line.
- Try to stage and commit only one of those changes
- Check your git status
The other change in the file should be still unstaged, and still there.


ds300:
looks great, I'll test this after lunch. Do you mind cleaning up the commit history?

pvinis:
In this one I actually left them so it's obvious what worked or not, but sure, I can clean them up, like put the updates together etc.

ds300:
Thank you for fixing this ❤️Partial commits are a common workflow for me, so this is a really nice to have :)

PR #3220, opened 2020-04-22
Small Collections refactors

ashfurrow:
Hi Ashley 👋 I've been following your PRs and wanted to say, really great work on the Collections UI! Emission currently uses a few patterns that we're trying to move away from, which have snuck into the Collections work, so I put this PR together to show what we're trying to do instead. All my feedback is open to discussion, too, so please let me know where I can clarify my thinking (or if I got something wrong).

What we're trying to do:

---

Move away from `renderItem()`, `renderSection()`, or `renderThing()` functions. We should put as much of the render logic in the `render()` function as possible. There are a few reasons, but the big one is that TypeScript is a lot less useful when breaking things into smaller, one-off functions. Take a look at the typings first, it's just `any`, which isn't helpful.

Screen Shot 2020-04-22 at 10 11 34

Now take a look:

Screen Shot 2020-04-22 at 10 49 49

Not only do we know it's an array of strings, but we know exactly which strings to expect (because of `as const`). 

(I also removed `data` from the sections because it wasn't being used, but we can add it back if it will be 👍)

---

Move away from setting up sections in `componentDidMount()`. We originally used this pattern to perform heavy calculations upfront, instead of in every render. What we've found over the years is that these computations aren't really that expensive, and would be better computed in `render`. This lets us avoid having to add explicit types to a lot of code, because TypeScript will infer the types for us. 

A lot of components still use `this.state.sections`, but we're hoping to get rid of most of them.

---

This isn't a coding refactor, but I also changed the horizontal card rail to go edge-to-edge, to match the home screen:

| Before | After |
|---|---|
| Screen Shot 2020-04-22 at 10 38 35 |  Screen Shot 2020-04-22 at 10 38 20 |

I did this using the `ListHeaderComponent`, `ListFooterComponent`, and `ItemSeparatorComponent` props. But I did have to remove the `<Box>` container in `Collection`. I moved the bottom margin from the `<Box>` to the `Collection`'s `ItemSeparatorComponent` prop.

---

Let me know again where I can clarify anything 👍

### DIFF
--- a/src/lib/Scenes/Collection/Collection.tsx
+++ b/src/lib/Scenes/Collection/Collection.tsx
@@ -20,7 +20,6 @@ interface CollectionProps {
 }
 
 interface CollectionState {
-  sections: Array<{ type: string; data: any }>
   isArtworkGridVisible: boolean
   isFilterArtworksModalVisible: boolean
 }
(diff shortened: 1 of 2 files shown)


ashleyjelks:
> Hi Ashley 👋 I've been following your PRs and wanted to say, really great work on the Collections UI! Emission currently uses a few patterns that we're trying to move away from, which have snuck into the Collections work, so I put this PR together to show what we're trying to do instead. All my feedback is open to discussion, too, so please let me know where I can clarify my thinking (or if I got something wrong).
> 
> What we're trying to do:
> 
> Move away from `renderItem()`, `renderSection()`, or `renderThing()` functions. We should put as much of the render logic in the `render()` function as possible. There are a few reasons, but the big one is that TypeScript is a lot less useful when breaking things into smaller, one-off functions. Take a look at the typings first, it's just `any`, which isn't helpful.
> 
> Screen Shot 2020-04-22 at 10 11 34
> 
> Now take a look:
> 
> Screen Shot 2020-04-22 at 10 49 49
> 
> Not only do we know it's an array of strings, but we know exactly which strings to expect (because of `as const`).
> 
> (I also removed `data` from the sections because it wasn't being used, but we can add it back if it will be 👍)
> 
> Move away from setting up sections in `componentDidMount()`. We originally used this pattern to perform heavy calculations upfront, instead of in every render. What we've found over the years is that these computations aren't really that expensive, and would be better computed in `render`. This lets us avoid having to add explicit types to a lot of code, because TypeScript will infer the types for us.
> 
> A lot of components still use `this.state.sections`, but we're hoping to get rid of most of them.
> 
> This isn't a coding refactor, but I also changed the horizontal card rail to go edge-to-edge, to match the home screen:
> 
> Before	After
> Screen Shot 2020-04-22 at 10 38 35	Screen Shot 2020-04-22 at 10 38 20
> I did this using the `ListHeaderComponent`, `ListFooterComponent`, and `ItemSeparatorComponent` props. But I did have to remove the `<Box>` container in `Collection`. I moved the bottom margin from the `<Box>` to the `Collection`'s `ItemSeparatorComponent` prop.
> 
> Let me know again where I can clarify anything 👍

This looks great Ash, thanks!

PR #3221, opened 2020-04-22
Refactor ViewingRoomHeader to use a fragment container (GALL-2784)

iskounen:
The purpose of this PR is to refactor `ViewingRoomHeader` to use a fragment container (instead of receiving its props directly from its parent) so that its test can be written with the `ReactTestRenderer` instead of `enzyme`.


iskounen:
Tests to write:

- [ ] ViewingRoom
- [ ] ViewingRoomArtworks
- [ ] ViewingRoomArtworkRail
- [ ] ViewingRoomHeader
- [ ] ViewingRoomStatement
- [ ] ViewingRoomSubsections

PR #3226, opened 2020-04-22
Add Artist Series/Iconic Collections Rail to the Artist Page in iOS

ashleyjelks:
The PR exposes an artist rail with Iconic Collections on Artist pages that have 2 or more collections associated with the artist.

Kapture 2020-04-22 at 17 28 26

Kapture 2020-04-22 at 17 37 14

Without Iconic Collection:
Screen Shot 2020-04-24 at 5 03 23 PM

https://artsyproduct.atlassian.net/browse/FX-1890

#skip_new_tests


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>Missing Test Files:

- `src/lib/Components/Artist/ArtistArtworks/__tests__/ArtistCollectionsRailRenderer-tests.tsx`
- `src/lib/Components/__tests__/ArtistSeriesRail-tests.tsx`

    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/Components/Artist/ArtistArtworks/index.tsx
```diff
@@ -14,6 +16,10 @@ interface Props extends InfiniteScrollGridProps {
 
 const ArtworksGrid: React.FC<Props> = ({ artist, relay, ...props }) => (
   <StickyTabPageScrollView>
+    <ArtistCollectionsRail artistID={artist.internalID} />
```
ds300, line 19: Adding this rail as a standalone query renderer degrades the loading experience on the artist page.

Before:

image


After:

image

We should add it as a fragment container instead, and include the data in the 'above the fold' query so it gets fetched straight away.

Also, was it intentional to keep the `All works` title visible even when there's no iconic collections rail?

ds300: remember to update the loading placeholder too!

src/lib/Components/Artist/ArtistArtworks/index.tsx
```diff
@@ -14,6 +16,10 @@ interface Props extends InfiniteScrollGridProps {
...
+    <Sans size="4" mb={1}>
+      All works
```
ds300, line 21:
```suggestion
      All Works
```

src/lib/Components/Artist/ArtistArtworks/ArtistCollectionsRailRenderer.tsx
```diff
@@ -0,0 +1,40 @@
+import { Box } from "@artsy/palette"
+import { ArtistArtworks_artist } from "__generated__/ArtistArtworks_artist.graphql"
+import { ArtistCollectionsRailRendererQuery } from "__generated__/ArtistCollectionsRailRendererQuery.graphql"
+import { defaultEnvironment } from "lib/relay/createEnvironment"
+import renderWithLoadProgress from "lib/utils/renderWithLoadProgress"
+import React from "react"
+import { graphql, QueryRenderer } from "react-relay"
+import { ArtistCollectionsRailFragmentContainer as ArtistCollectionsRail } from "./ArtistCollectionsRail"
```
ds300, line 8:
```suggestion
import { ArtistCollectionsRailFragmentContainer } from "./ArtistCollectionsRail"
```

renaming fragment containers in import statements has been a fairly common pattern at Artsy but going forward we're moving away from it so that components have consistent naming across different files.

src/lib/Components/Artist/ArtistArtworks/index.tsx
ashleyjelks: This is correct per the spec, but I just followed with @nicoleyeo to see what the expected case should be!

src/lib/Components/Artist/ArtistArtworks/index.tsx
ashleyjelks: > Also, was it intentional to keep the All works title visible even when there's no iconic collections rail?

Just followed up with @nicoleyeo  about this as well!

ashleyjelks:
> Looks great!
> 
> I noticed there's still an issue with janky loading, but it seems to be a problem with the sticky tab page code rather than your code. Still, I'd like to fix it before this gets released. I added a ticket to track it here: https://artsyproduct.atlassian.net/browse/MX-311

Sounds good! Depending on long that fix will take we can either move this work behind a flag so we can still test it a beta or just leave it in this state if the fix for MX-311 will go out soon. Thanks for looking into this!

sweir27:
@ashleyjelks this looks good to me! Looks like there are a few strictness failures-- the output doesn't specify where they are, do you see them on your end?

ashleyjelks:
#mergeongreen

ashleyjelks:
> @ashleyjelks this looks good to me! Looks like there are a few strictness failures-- the output doesn't specify where they are, do you see them on your end?

I just followed up with MX about this. So anytime a new PR introduces an increase in the number of instances of `STRICTNESS_MIGRATION` this check will fail. In this case the increases are valid/acceptable because they were introduced after resolving merge conflicts from rebasing against master. Going to go ahead and merge since I got :white_check_mark: from the MX team that its safe to merge in these instances.

PR #3227, opened 2020-04-22
Adds Price Range filter to collection view

sweir27:
Pretty straightforward now that this is the third filter we've built! Adds the ability to filter by price range from a collection view.

Addresses: https://artsyproduct.atlassian.net/browse/FX-1635

price-filter


sweir27:
@ashleyjelks uploaded a gif 😄

PR #3232, opened 2020-04-23
Viewing rooms visual cleanup

mdole:
Still a few things to go here, but it's a start! 

cc @iskounen @ds300 I'll work on wrapping this + #3215 tomorrow!

Here's the list I'm working off of:

- [x] Check with Ash about testing behavior for deeplinking
- [x] Remove "Viewing Room" text from header
- [x] Move flatlist from statement component to toplevel
- [x] Convert artworks flatlist so that each artwork is its own section + change type
- [x] Make # artworks correctly plural/singular for artworks button + rail
- [x] Subsection 30px top padding
- [x] Get hovering button to be sticky all the time
- [x] Test out deeplinking behavior - does back button take us home? (If you have the app open already, it takes you back to the screen you were already on. Seems good)
- [x] Get hovering button to only appear once we've scrolled to a certain point
- [ ] ~~Header image extends into the horns~~ (leaving this for a followup PR that focuses on a few header tweaks)

#trivial


src/lib/Scenes/ViewingRoom/__tests__/ViewingRoom-tests.tsx
```diff
@@ -79,14 +78,19 @@ describe("ViewingRoomStatement", () => {
...
     mockEnvironment.mock.resolveMostRecentOperation(operation => {
       const result = MockPayloadGenerator.generate(operation, {
         ViewingRoom: () => ({ artworksForCount: { totalCount: 42 } }),
       })
       return result
     })
+    // Matt: goal here is to scroll and trigger the button to pop, but I'm missing something :/
+    act(() => {
+      tree.root.findByType(FlatList).instance.scrollToEnd()
+    })
```
mdole, line 93: @ds300 any advice on how to handle this?

ds300: Hmm would it be feasible to call `tree.root.findByType(FlatList).props. onViewableItemsChanged `?

mdole: good thought - would that be `onViewableItemsChanged()` called as a method or `onViewableItemsChanged` as a prop? Tried both out; with the former I get an error `TypeError: Cannot read property 'viewableItems' of undefined` and the latter just fails and says `No instances found with props: {"data-test-id":"view-works"}`.

Mmm thinking through this more - we might be calling it correctly, but because we only call `setDisplayWorksButton(true)` when `itemKey == "pullQuote"`, I bet we're just not displaying the right section.

So seems like we should figure out a way to add `pullQuote` to `viewableItems`?

Screen Shot 2020-04-28 at 10 37 12 AM

Screen Shot 2020-04-28 at 10 40 48 AM

src/lib/Scenes/ViewingRoom/Components/ViewingRoomArtworkRail.tsx
```diff
@@ -21,12 +21,16 @@ const ArtworkCard = styled.TouchableHighlight`
 
 export const ViewingRoomArtworkRail: React.FC<ViewingRoomArtworkRailProps> = props => {
   const artworks = props.viewingRoomArtworks.artworks! /* STRICTNESS_MIGRATION */.edges! /* STRICTNESS_MIGRATION */
+  const totalCount = props.viewingRoomArtworks.artworks! /* STRICTNESS_MIGRATION */
+    .totalCount! /* STRICTNESS_MIGRATION */
```
ds300, line 25: as a follow-up it would be good to make these non-nullable in MP!

src/lib/Scenes/ViewingRoom/__tests__/ViewingRoom-tests.tsx
mdole: resolved in 6039626

src/lib/Scenes/ViewingRoom/ViewingRoom.tsx
```diff
@@ -1,22 +1,78 @@
...
+  const pluralizedArtworksCount = artworksCount === 1 ? "work" : "works"
+
+  sections.push({
+    key: "introStatement",
+    content: (
+      <Serif data-test-id="intro-statement" size="4" mt="2" mx="2">
+        {viewingRoom.introStatement}
+      </Serif>
+    ),
+  })
```
ds300, line 42: no need to be pushing these, just plop them directly in the array:


```ts
const sections = [{
  key: "introStatement",
  content: (
    <Serif data-test-id="intro-statement" size="4" mt="2" mx="2">
      {viewingRoom.introStatement}
    </Serif>
  ),
},
  ...
]
```

src/lib/Scenes/ViewingRoom/ViewingRoom.tsx
```diff
@@ -1,22 +1,78 @@
...
+  content: JSX.Element
+}
 
 export const ViewingRoom: React.FC<ViewingRoomProps> = props => {
   const viewingRoom = props.viewingRoom
+  const sections: ViewingRoomSection[] = []
+  const navRef = useRef()
+  const [displayViewWorksButton, setDisplayViewWorksButton] = useState(false)
+
+  const artworksCount = viewingRoom.artworksForCount! /* STRICTNESS_MIGRATION */.totalCount
```
ds300, line 32: you shouldn't be adding `STRICTNESS_MIGRATION` tags in new code. use `?` and `!` or whatever is safest.

src/lib/Scenes/ViewingRoom/ViewingRoom.tsx
```diff
@@ -27,23 +83,76 @@ export const ViewingRoom: React.FC<ViewingRoomProps> = props => {
...
+            renderItem={({ item }) => {
+              return item.content
+            }}
+          />
+          {displayViewWorksButton && (
+            <ViewWorksButtonContainer>
+              <TouchableWithoutFeedback
+                onPress={() =>
+                  SwitchBoard.presentNavigationViewController(
+                    // @ts-ignore STRICTNESS_MIGRATION
```
ds300, line 106: same same. It's safe to use `!` here, e.g. `navRef.current!`

src/lib/Scenes/ViewingRoom/Components/ViewingRoomArtworkRail.tsx
mdole: good thought! added to my to-do list

PR #3234, opened 2020-04-24
Suppress build log output from beta builds #trivial

ashfurrow:
Building the app for beta deploys produces a tonne of output, causing this to happen to our Circle logs:

```
Your output is too large to display in the browser. Only the last 400000 characters are displayed.
```

And we have to wait until the whole job completes before we can see the rest of the log. Most of the output comes from actually building the app, which should always succeed; let's silence the build log to get better CI logging on beta builds.


pvinis:
Oops. I missed that one.

#mergeongreen

PR #3235, opened 2020-04-24
Add more tests for viewing room components (GALL-2784)

iskounen:
This is a continuation of the work to write tests for the new viewing room components, which was started in #3221. That code was merged before its time so that we could merge the newly refactored `ViewingRoomHeader` functional component.

### DIFF
--- a/src/lib/Scenes/ViewingRoom/Components/ViewingRoomSubsections.tsx
+++ b/src/lib/Scenes/ViewingRoom/Components/ViewingRoomSubsections.tsx
@@ -1,18 +1,18 @@
 import { Box, Sans, Serif } from "@artsy/palette"
-import { ViewingRoomSubsections_viewingRoomSubsections } from "__generated__/ViewingRoomSubsections_viewingRoomSubsections.graphql"
+import { ViewingRoomSubsections_viewingRoom } from "__generated__/ViewingRoomSubsections_viewingRoom.graphql"
 import ImageView from "lib/Components/OpaqueImageView/OpaqueImageView"
 import React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
 
 interface ViewingRoomSubsectionProps {
-  viewingRoomSubsections: ViewingRoomSubsections_viewingRoomSubsections
+  viewingRoom: ViewingRoomSubsections_viewingRoom
 }
 
-export class ViewingRoomSubsections extends React.Component<ViewingRoomSubsectionProps> {
-  render() {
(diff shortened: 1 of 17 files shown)


src/lib/Scenes/ViewingRoom/Components/ViewingRoomSubsections.tsx
```diff
@@ -1,18 +1,18 @@
...
+  viewingRoom: ViewingRoomSubsections_viewingRoom
 }
 
-export class ViewingRoomSubsections extends React.Component<ViewingRoomSubsectionProps> {
-  render() {
-    const subsections = this.props.viewingRoomSubsections.subsections
-    return subsections.map((subsection, index) => {
-      return (
+export const ViewingRoomSubsections: React.FC<ViewingRoomSubsectionProps> = props => {
+  const subsections = props.viewingRoom.subsections
```
iskounen, line 12: I had to rename `ViewingRoomSubsections_viewingRoomSubsections` to `ViewingRoomSubsections_viewingRoom` to prevent this line from throwing an error when this component was populated by the `ReactTestRenderer`.

The error that was being thrown was "TypeError: cannot read property 'subsections' of null". Debugging this message made me realize that `props.vieiwngRoomSubsections` was always null when this component was being initialized during a test run. I am still not sure why the result of the fragment wasn't being mapped to this prop, but now the test passes.

**I have not tested this in a simulator yet!**

ds300: It was probably due to the fact that the fragment name needs to match the name/alias of the selection in the root query.

e.g.

```ts
const MyComponent = ({viewingRoom: MyComponent_viewingRoom}) => { ... }
const MyComponentContainer = createFragmentContainer<...>(MyComponent, {viewingRoom: graphql`...`})
const TestRenderer = () =>
  <QueryRenderer
    ...
    query={graphql`
      query Blah {
        viewingRoom {
          ...
        }
      }
    `}
    render={({props}) => <MyComponent {...props} />}
```

In the code above `viewingRoom` needs to match in all four places otherwise it won't work. if you want to change the name of the prop you need to make an alias for the field selection in the query
```ts
const MyComponent = ({blahBlah: MyComponent_blahBlah}) => { ... }
const MyComponentContainer = createFragmentContainer<...>(MyComponent, {blahBlah: graphql`...`})
const TestRenderer = () =>
  <QueryRenderer
    ...
    query={graphql`
      query Blah {
        blahBlah: viewingRoom {
          ...
        }
      }
    `}
    render={({props}) => <MyComponent {...props} />}
```

PR #3241, opened 2020-04-27
[MX-300] - Feature flag for consignments tab

brainbicycle:
- Adds a feature flag for adding sales tab and removing city guide tab
- Adds a navigation controller and component view controller for consignments
- Deletes unused ARTopMenuInternalWebViewController
- Refactors ARTopMenuDataSource to be only place tab ordering is defined or depended upon
- Updates tab deeplink logic to include Sales tab, fixes some bugs around deeplinking to tabs
- Updates docs for adding lab and echo options 

The thinking behind the refactor was that a few places were implicitly depending on ordering of tabs via `if ([UIDevice isiPhone])` checks to show cityGuide or passing a number and assuming it corresponded with a given tab type so changing ordering or removing a tab could potentially break these assumptions. 

The goal of the refactor was to make it so ordering is only defined once in the ARRTopMenuDataSource and everywhere else uses a tabType rather than an index where possible and everywhere else ask the dataSource for the index for a given tab type. 

**Flag Enabled**
For now just put the existing consignments flow here

flagEnabled

**Flag Disabled**

flagDisabled

**Flag**

moveCityFlag


Artsy/View_Controllers/App_Navigation/ARNavigationTabButtonWithBadge.h
```diff
@@ -0,0 +1,18 @@
+//
+//  ARNavigationTabButtonWithBadge.h
+//  Artsy
+//
+//  Created by Brian Beckerle on 4/23/20.
+//  Copyright © 2020 Artsy. All rights reserved.
+//
```
ashfurrow, line 7: Can we delete these Xcode-generated header comments? We used to have a Danger rule for detecting them, even – one of the previous engineers really didn't like them. But we should be consistent 👍

Artsy/View_Controllers/App_Navigation/ARNavigationTabButtonWithBadge.h
```diff
@@ -0,0 +1,18 @@
...
+
+#import "ARButtonSubclasses.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface ARNavigationTabButtonWithBadge : ARNavigationTabButton
+@property (nonatomic, strong) UIImage *icon;
+- (void)setBadgeCount:(NSUInteger)badgeCount;
+@end
```
ashfurrow, line 16: Minor suggestion for legibility: 

```suggestion
@interface ARNavigationTabButtonWithBadge : ARNavigationTabButton

@property (nonatomic, strong) UIImage *icon;
- (void)setBadgeCount:(NSUInteger)badgeCount;

@end
```

Artsy/View_Controllers/App_Navigation/ARNavigationTabButtonWithBadge.m
```diff
@@ -0,0 +1,70 @@
...
+@implementation ARNavigationTabButtonWithBadge
+
+- (void)setup;
+{
+    [super setup];
+    [self setTitleColor:[UIColor whiteColor] forState:UIControlStateNormal];
+    [self setTitleColor:[UIColor whiteColor] forState:UIControlStateSelected];
+
+    // TODO: Clears all on tap. This needs to get more sophisticated, as per
+    // https://github.com/artsy/collector-experience/issues/661
```
ashfurrow, line 21: plmusf5

Artsy/View_Controllers/App_Navigation/ARTopMenuNavigationDataSource.m
```diff
@@ -8,9 +8,9 @@
 #import <Emission/ARMyProfileComponentViewController.h>
 #import <Emission/ARMapContainerViewController.h>
 #import <Emission/ARSearchComponentViewController.h>
+#import <Emission/ARConsignmentsComponentViewController.h>
 
 #import "AREigenMapContainerViewController.h"
-#import "ARTopMenuInternalMobileWebViewController.h"
```
ashfurrow, line 13: Nice find!

Artsy/View_Controllers/App_Navigation/ARTopMenuNavigationDataSource.m
```diff
@@ -116,45 +128,97 @@ - (ARNavigationController *)favoritesNavigationController
...
+        case ARProfileTab:
+            return @"profile";
+        default:
+            return @"unknown";
+    }
+}
 
-        case ARTopTabControllerIndexFavorites:
-            return [self favoritesNavigationController];
+- (NSArray *)tabOrder
```
ashfurrow, line 181: Ahhhhh this is _such_ a better abstraction, nice job 👍

Artsy/View_Controllers/App_Navigation/ARTopMenuNavigationDataSource.m
```diff
@@ -37,10 +37,10 @@ @interface ARTopMenuNavigationDataSource ()
 @property (nonatomic, strong) ARNavigationController *localDiscoveryNavigationController;
 @property (nonatomic, strong) ARNavigationController *messagingNavigationController;
 @property (nonatomic, strong) ARNavigationController *profileNavigationController;
+@property (nonatomic, strong) ARNavigationController *consignmentsNavigationController;
```
ashfurrow, line 40: Maybe we could rename this to `sellNavigationController`? It might help differentiate it from the (modal) consignments flow itself.

Artsy/View_Controllers/App_Navigation/ARTopMenuNavigationDataSource.m
```diff
@@ -198,16 +262,19 @@ - (void)setBadgeNumber:(NSUInteger)number forTabAtIndex:(NSInteger)index;
...
     [[UIApplication sharedApplication] setApplicationIconBadgeNumber:total];
 }
 
 // Just an alias for the above, which keeps the ARTabViewDataSource and ARTopMenuViewController concerns seperated.
-- (void)setNotificationCount:(NSUInteger)number forControllerAtIndex:(ARTopTabControllerIndex)index;
+- (void)setNotificationCount:(NSUInteger)number forControllerAtTab:(ARTopTabControllerTabType)tabType;
 {
-    [self setBadgeNumber:number forTabAtIndex:index];
+    NSUInteger index = [self indexForTabType:tabType];
+    if (index != NSNotFound) {
```
ashfurrow, line 275: 👍

Artsy_Tests/View_Controller_Tests/App_Menu/ARTopMenuViewControllerSpec.m
```diff
@@ -183,7 +185,7 @@ @interface ARTopMenuViewController (ExposedForTesting) <ARTabViewDelegate>
            after(^{
                [ARUserManager clearUserData];
            });
-           itShouldBehaveLike(@"tab behavior", @{@"tab" : [NSNumber numberWithInt:ARTopTabControllerIndexFavorites]});
```
ashfurrow, line 186: `numberWithInt:` made me wonder how old this line of code was... [five years](https://github.com/artsy/eigen/commit/9767f006585a05c2160b504f2edff381fd0db097)!

docs/developing_a_feature.md
```diff
@@ -52,7 +55,7 @@ An lab option enabled in the Eigen admin screen will override whatever the Echo
 
 If you make Echo changes, you can update the local bundled copy of the echo settings by running `make update_echo` in Eigen. This is done automatically when running `pod install`.
 
-## Using an Emission Option
+## Using an option in Emission
 
 What do we mean when we say "a new feature should be **put behind** a lab option or Echo Feature"? It means that the behaviour of the app changes depending on if this option is set. This can look like adding a new view to the hierarchy _if_ the option is set:
```
ashfurrow, line 61: Could we update this next paragraph to reflect the updates from #3167? We would need a link to [where to add the new option](https://github.com/artsy/eigen/blob/c9a6a07cac8d9ce8f6018bb522204c7d6385478a/typings/emission.d.ts#L22-L39) and then modify `enableNewAndExcitingFeature` to read:

```ts
const enableNewAndExcitingFeature = NativeModules.Emission.options.AROptionsNewAndExcitingFeature
```

emission/Pod/Classes/ViewControllers/ARConsignmentsComponentViewController.m
```diff
@@ -0,0 +1,15 @@
...
+
+@interface ARConsignmentsComponentViewController ()
+@end
+
+@implementation ARConsignmentsComponentViewController
+
+- (instancetype)init;
+{
+    return [super initWithEmission:nil
+                        moduleName:@"Consignments"
```
ashfurrow, line 11: Same thing here, changing "Consignments" to "Sales" 👍

src/lib/Components/Consignments/Screens/Confirmation.tsx
```diff
@@ -72,7 +72,7 @@ export default class Confirmation extends React.Component<Props, State> {
   successContent = () => (
     <View>
       <Serif size="5" style={{ textAlign: "center" }}>
-        Succesfully submitted
+        Successfully submitted
```
ashfurrow, line 75: 🙌

Artsy/View_Controllers/App_Navigation/ARNavigationTabButtonWithBadge.h
brainbicycle: Sure thing 👍

Artsy/View_Controllers/App_Navigation/ARNavigationTabButtonWithBadge.h
brainbicycle: 👍

docs/developing_a_feature.md
brainbicycle: Good call 👍

Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
```diff
@@ -303,16 +239,18 @@ - (void)fadeInFromOnboarding
 - (void)registerWithSwitchBoard:(ARSwitchBoard *)switchboard
 {
     NSDictionary *menuToPaths = @{
-        @(ARTopTabControllerIndexHome) : @"/",
-        @(ARTopTabControllerIndexMessaging) : @"/inbox",
-        @(ARTopTabControllerIndexLocalDiscovery) : @"/local-discovery",
-        @(ARTopTabControllerIndexFavorites) : @"/favorites",
-        @(ARTopTabControllerIndexProfile) : @"/ios-settings", // A good argument is "user/edit", _but_ the app barely supports any of it's features
```
brainbicycle, line 310: We don't currently have a dedicated tab for profile so I got rid of this for the time being because the current logic doesn't seem to present the profile in the correct tab (favorites) and the workaround didn't seem worth it because we will have a dedicated profile tab pretty soon. We also have a deeplink already for this in switchboard which seems to behave the same as the logic and registration we had here: https://github.com/artsy/eigen/blob/master/Artsy/App/ARSwitchBoard.m#L266

Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
```diff
@@ -580,39 +539,32 @@ - (void)presentRootViewControllerAtIndex:(NSInteger)index animated:(BOOL)animate
...
+        case ARLocalDiscoveryTab:
+        case ARSearchTab:
+        case ARSalesTab:
+        case ARFavoritesTab:
             presentableController = [self rootNavigationControllerAtIndex:index];
             break;
-        case ARTopTabControllerIndexFavorites:
-            presentableController = [self rootNavigationControllerAtIndex:index];
-            break;
-        case ARTopTabControllerIndexProfile:
```
brainbicycle, line 606: other piece of profile tab presentation logic removed.

Artsy/App/ARAppDelegate+Emission.m
```diff
@@ -412,6 +415,42 @@ - (BOOL)isRootNavViewController
...
+
+@implementation ARSalesComponentViewController (ARRootViewController)
+
+- (BOOL)isRootNavViewController
+{
+    return YES;
+}
+
+@end
+
```
brainbicycle, line 453: There were a couple bugs around deeplinking to tabs, if the viewController presented is not registered as a rootNavViewController it will be presented in whatever tab is currently active rather than its correct tab: https://github.com/artsy/eigen/blob/5008396c39cbbb55fdf3058b13ef041599ef421f/Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m#L564, this becomes a bigger problem because we reuse these navigation controllers whenever possibly so if one ends up in the wrong tab it will not appear when you tap into the correct tab.

Artsy/App/ARAppDelegate+Emission.m
```diff
@@ -412,6 +415,42 @@ - (BOOL)isRootNavViewController
 
 @end
 
+@interface ARSearchComponentViewController (ARRootViewController) <ARRootViewController>
+@end
+
+@implementation ARSearchComponentViewController (ARRootViewController)
+
+- (BOOL)isRootNavViewController
```
brainbicycle, line 423: There were a couple bugs around deeplinking to tabs, if the viewController presented is not registered as a rootNavViewController it will be presented in whatever tab is currently active rather than its correct tab: https://github.com/artsy/eigen/blob/5008396c39cbbb55fdf3058b13ef041599ef421f/Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m#L564, this becomes a bigger problem because we reuse these navigation controllers whenever possibly so if one ends up in the wrong tab it will not appear when you tap into the correct tab.

I made fancy gifs to show the behavior but they are too big for github 😫

Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
```diff
@@ -303,16 +239,18 @@ - (void)fadeInFromOnboarding
...
     NSDictionary *menuToPaths = @{
-        @(ARTopTabControllerIndexHome) : @"/",
-        @(ARTopTabControllerIndexMessaging) : @"/inbox",
-        @(ARTopTabControllerIndexLocalDiscovery) : @"/local-discovery",
-        @(ARTopTabControllerIndexFavorites) : @"/favorites",
-        @(ARTopTabControllerIndexProfile) : @"/ios-settings", // A good argument is "user/edit", _but_ the app barely supports any of it's features
+        @(ARHomeTab) : @"/",
+        @(ARMessagingTab) : @"/inbox",
+        @(ARLocalDiscoveryTab) : @"/local-discovery",
+        @(ARSearchTab) : @"/search",
```
brainbicycle, line 245: Also added search to switchboard registration

PR #3243, opened 2020-04-27
Visual QA Launch Blocker Fixes for v2 Collections Filters

ashleyjelks:
This PR fixes the following bugs reported from QA-ing the v2 iOS Collections Filters:

- [Increasing the tappable area for the close and back icons](https://artsyproduct.atlassian.net/browse/FX-1910)
- [Decreasing the viewability config of the artworks grid so the filter button animates sooner](https://artsyproduct.atlassian.net/browse/FX-1911)

Kapture 2020-04-27 at 13 18 32

Kapture 2020-04-27 at 17 33 54

### DIFF
--- a/src/lib/Components/ArtworkFilterOptions/SingleSelectOption.tsx
+++ b/src/lib/Components/ArtworkFilterOptions/SingleSelectOption.tsx
@@ -41,7 +41,7 @@ export const SingleSelectOptionScreen: React.SFC<SingleSelectOptionScreenProps> 
         </Sans>
         <Box></Box>
       </FilterHeader>
-      <SingleSelectOptionsContainer mb={120}>
+      <Flex mb={120} mt={"-20px"}>
         <FlatList<SingleSelectOptions>
           initialNumToRender={12}
           keyExtractor={(_item, index) => String(index)}
@@ -67,15 +67,11 @@ export const SingleSelectOptionScreen: React.SFC<SingleSelectOptionScreenProps> 
             </Box>
           )}
         />
-      </SingleSelectOptionsContainer>
+      </Flex>
       <BackgroundFill />
     </Flex>
   )
 }
-
-export const SingleSelectOptionsContainer = styled(Flex)`
-  margin-top: -20px;
-`
 
 export const FilterHeader = styled(Flex)`
   flex-direction: row;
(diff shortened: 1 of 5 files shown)


src/lib/Components/ArtworkFilterOptions/SingleSelectOption.tsx
```diff
@@ -67,21 +67,24 @@ export const SingleSelectOptionScreen: React.SFC<SingleSelectOptionScreenProps>
...
         />
-      </Flex>
+      </SingleSelectOptionsContainer>
       <BackgroundFill />
     </Flex>
   )
 }
 
+export const SingleSelectOptionsContainer = styled(Flex)`
+  margin-top: -20px;
```
sweir27, line 77: Did it not work to just add `mt={"-20px"}` to the `Flex` itself? Curious if we can avoid adding this wrapper/

src/lib/Components/FilterModal.tsx
```diff
@@ -214,11 +214,14 @@ export const FilterOptions: React.SFC<FilterOptionsProps> = props => {
             </Box>
           )}
         />
-      </Flex>
+      </FilterOptionsContainer>
       <BackgroundFill />
     </Flex>
   )
 }
+export const FilterOptionsContainer = styled(Flex)`
```
sweir27, line 222: Same here re: question above-- can we add this to the `Flex` without this wrapper? Just makes it a little simpler.

ashleyjelks: It is being used in the test files and since `react-test-render` and `enzyme` will mount all of the Flex components in an array, we'd have to access it by grabbing its index, e.g. `findByType(Flex[8])....`  so it was clearer in the code to export an empty but re-named styled-component.

sweir27: That makes sense generally-- in this case I don't see it in the tests! (but could be missing something)

ashleyjelks:
> Couple of non-blocking questions. This looks good!

Cool, this works inline `mt={"-20px"}` so I removed the styled component declarations + responded to your question about the empty styled-component! Will add a merge on green to this!

src/lib/Components/FilterModal.tsx
ashleyjelks: Nope you're not missing anything, looks I copied that line from a styled component that was being exported for use in a test, but have since removed it since moving the css inline e.g. `mt={"-20px"}`. So its gone now!

ds300:
yo, to increase tappable area you can use the `hitSlop` property on `Touchable*` components, which lets you do it without messing with margins and padding.

PR #3245, opened 2020-04-28
Add styled-components/native typings

ds300:
I first just upgraded  `@types/styled-components` and fixed all the issues but it made TypeScript unusably slow in vscode, and it would have meant upgrading palette and reaction and who knows what else too, so I scrapped that work and just added this extra local typings file.

While fixing typings issues I cleaned up the Auctions page code, and then paired with @samchieng to clean up the design of that page too.

### Before
image

### After
image


src/lib/relay/middlewares/cacheMiddleware.ts
```diff
@@ -75,9 +75,12 @@ export const cacheMiddleware = () => {
...
 
-      // @ts-ignore STRICTNESS_MIGRATION
-      const error = new NetworkError(response.statusText)
-      // @ts-ignore STRICTNESS_MIGRATION
+      const errorMessage = `
+errors: ${JSON.stringify(response.json?.errors ?? response.statusText, null, "  ")}
+queryID: ${queryID}
+variables: ${JSON.stringify(req.variables, null, "  ")}
+`
+      const error = new NetworkError(errorMessage)
```
ds300, line 83: tapping around I found a view that triggered this error. `response.statusText` was `null` here, and I think this is the source of https://artsyproduct.atlassian.net/browse/MX-271 so I added the actual error messages from metaphysics.

The view I found was a past show from Alpha 137 gallery. I think we might have id/internalID mixed up from the MPv2 migration. Will add a note to that ticket.

ds300:
my PR to DefinitelyTyped just got merged so I can remove the patch file in this PR tomorrow morning https://github.com/DefinitelyTyped/DefinitelyTyped/pull/44296#issuecomment-620798794

src/lib/Scenes/Artwork/Components/ImageCarousel/FullScreen/ImageZoomView.tsx
```diff
@@ -130,8 +130,20 @@ export const ImageZoomView =
...
     const transform = useMemo(() => createTransform(transition, imageTransitionOffset), [imageTransitionOffset])
 
-    // @ts-ignore STRICTNESS_MIGRATION
-    const imageFittedWithinScreen = fitInside(screenDimensions, image.deepZoom.image.size)
+    const rawImageSize = image?.deepZoom?.image?.size
+    if (!rawImageSize) {
+      if (__DEV__) {
+        console.error("No image size info found")
+      } else {
+        captureMessage("No image size info found (see breadcrumbs for artowrk id)")
```
ashfurrow, line 138: Nice update 👍

src/lib/Scenes/Sales/Components/SaleListItem.tsx
```diff
@@ -89,35 +26,30 @@ export class SaleListItem extends React.Component<Props> {
   render() {
     const sale = this.props.sale
     const image = sale.coverImage
-    // @ts-ignore STRICTNESS_MIGRATION
-    const timestamp = sale.displayTimelyAt.toUpperCase()
+    const timestamp = capitalize(sale.displayTimelyAt?.replace(/M$/, "mo").toLowerCase())
```
ashfurrow, line 29: 💯

src/lib/Scenes/Sales/Components/SaleList.tsx
```diff
@@ -1,77 +1,29 @@
...
 
+import { SectionTitle } from "lib/Components/SectionTitle"
+import { useScreenDimensions } from "lib/utils/useScreenDimensions"
 import SaleListItem from "./SaleListItem"
-import { SectionHeader } from "./SectionHeader"
 
-interface Props {
-  item: {
-    data: any[]
+export const SaleList: React.FC<{ sales: any[]; title: string }> = ({ sales, title }) => {
```
ashfurrow, line 9: Would it be possible/prudent to type `sales` here to be `sales: SaleListItem_sale[]`?

src/lib/Scenes/Sales/index.tsx
```diff
@@ -47,60 +48,28 @@ class Sales extends React.Component<Props, State> {
...
-      {
-        data: [{ data: this.props.me }],
-        title: "Lots by Artists You Follow",
-        // @ts-ignore STRICTNESS_MIGRATION
-        renderItem: props => {
-          return <LotsByFollowedArtists title={props.section.title} me={props.item.data} />
-        },
-      },
-    ]
+    const sales = this.props.sales.edges?.map(edge => edge?.node!) ?? []
```
ashfurrow, line 51: The `!` in `eges?.node!` makes me nervous. I know Metaphysics won't return an `edges` array with nulls in it, so we're covered, but it feels dangerous 😅 Maybe it's the Swift developer in me?

src/lib/Components/Bidding/Screens/SelectCountry.tsx
```diff
@@ -73,10 +69,8 @@ export class SelectCountry extends React.Component<SelectCountryProps, SelectCou
     })
 
     const predictions = (results.predictions || [])
-      // @ts-ignore STRICTNESS_MIGRATION
-      .filter(prediction => prediction.types[0] === "country")
-      // @ts-ignore STRICTNESS_MIGRATION
-      .map(prediction => ({ id: prediction.place_id, name: prediction.description }))
+      .filter((prediction: any) => prediction.types[0] === "country")
+      .map((prediction: any) => ({ id: prediction.place_id, name: prediction.description }))
```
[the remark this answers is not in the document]
ds300: sure :+1: I bet someone has implemented these API typings somewhere so it might encourage someone to find them one day

src/lib/Scenes/Sales/Components/SaleList.tsx
ds300: That wouldn't be correct either, since that's the unmasked fragment type but these sales are the masked version, which we can get using the helper type `FragmentRef`. The correct type would therefore be `Array<FragmentRef<SaleListItem_sale>>` so I'll update it to that. Good catch 🙏

src/lib/Scenes/Sales/index.tsx
ds300: Yeah i feel you! once we update the connections typings in metaphysics we can safely get rid of these checks.

ds300:
@ashfurrow thanks for the helpful review! This is 🍏 again

PR #3247, opened 2020-04-28
Collections Filters Visual QA Fixes

ashleyjelks:
Collections Filters Visual QA Fixes

- Adds a border in the filter modals to indicate separation between list row items and the modal header
- Adds a top border to the apply button to indicate separation between Apply button and list row items 
- Removes "double" border between list row items
- [Increases the tappable area for the close and back icons with hitSlop property](https://artsyproduct.atlassian.net/browse/FX-1910)

Kapture 2020-04-29 at 12 50 35

#trivial

### DIFF
--- a/src/lib/Components/ArtworkFilterOptions/SingleSelectOption.tsx
+++ b/src/lib/Components/ArtworkFilterOptions/SingleSelectOption.tsx
@@ -94,7 +94,7 @@ export const InnerOptionListItem = styled(Flex)`
   justify-content: space-between;
   flex-grow: 1;
   align-items: flex-end;
-  padding: 20px;
+  padding: ${space(2)}px;
 `
 
 export const SingleSelectOptionListItemRow = styled(TouchableOpacity)``
--- a/src/lib/Components/ArtworkFilterOptions/SingleSelectOption.tsx
+++ b/src/lib/Components/ArtworkFilterOptions/SingleSelectOption.tsx
@@ -1,4 +1,4 @@
-import { ArrowLeftIcon, Box, CheckIcon, Flex, Sans, space } from "@artsy/palette"
+import { ArrowLeftIcon, Box, CheckIcon, color, Flex, Sans, space } from "@artsy/palette"
 import { MediumOption, PriceRangeOption, SortOption } from "lib/Scenes/Collection/Helpers/FilterArtworksHelpers"
 import React from "react"
 import { FlatList, TouchableOpacity } from "react-native"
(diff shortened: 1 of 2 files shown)


ds300:
:ceiling-cat: the fixed height looks a little too high, which is to say that the title + back button are no longer appear vertically centred.

Also, unrelated to this PR, but I just noticed that the borders on the filter list items are duplicated, so they're slightly too thick wherever two list items meet.

image

src/lib/Components/ArtworkFilterOptions/SingleSelectOption.tsx
```diff
@@ -77,18 +80,21 @@ export const FilterHeader = styled(Flex)`
...
   padding-right: ${space(2)}px;
+  border: solid 0.5px ${color("black10")};
+  border-right-width: 0;
+  border-left-width: 0;
+  border-top-width: 0;
 `
 export const NavigateBackIconContainer = styled(TouchableOpacity)`
-  margin: 10px 0px 10px 20px;
-  padding: 10px;
+  margin: 20px 0px 0px 20px;
```
ashleyjelks, line 89: This was always here, originally it was:

```
export const ArrowLeftIconContainer = styled(TouchableOpacity)
margin: 10px 0px 10px 20px;
  margin-top: ${space(2)}px;	  padding: 10px;
  margin-left: ${space(2)}px;
```
... just converted it to shorthand.

src/lib/Components/ArtworkFilterOptions/SingleSelectOption.tsx
```diff
@@ -77,18 +80,21 @@ export const FilterHeader = styled(Flex)`
...
 `
 
 export const InnerOptionListItem = styled(Flex)`
   flex-direction: row;
   justify-content: space-between;
   flex-grow: 1;
   align-items: flex-end;
-  padding: ${space(2)}px;
+  padding: 20px;
```
sweir27, line 97: Not a blocker, but why move away from `space(2)` here?

ashleyjelks: Seems unnecessarily verbose, but I can revert it.

ashleyjelks:
> It seems like these changes fix the issues-- but it does seem like a lot of custom CSS. Would it be possible to use `StackableBorderBox` (from palette) here to fix the border issues?

It throws some compile errors and I'm not sure why, posted in the MX channel to see if there's a known issue with it. I don't see any uses of it in the app when I search for it, so not sure if that is because devs aren't aware of it or if there's some other underlying issue.

ashleyjelks:
> > It seems like these changes fix the issues-- but it does seem like a lot of custom CSS. Would it be possible to use `StackableBorderBox` (from palette) here to fix the border issues?
> 
> It throws some compile errors and I'm not sure why, posted in the MX channel to see if there's a known issue with it. I don't see any uses of it in the app when I search for it, so not sure if that is because devs aren't aware of it or if there's some other underlying issue.


MX just followed up, the TLDR is that it doesn't work because it uses CSS pseudo-selectors under the hood! https://artsy.slack.com/archives/CN8S32FJR/p1588186364077400

PR #3250, opened 2020-04-29
Fix sticky tab page issues

ds300:
This does a few things:

- Fixes a problem that was blocking FX where any views on a sticky tab page that had explicit height would cause jumping content during loading. I'm not sure why this was happening, potentially a glitch with Yoga. The workaround was to make sure that we mount the header in the flat list (which acts as a spacer) before mounting the rest of the content.

- Brings in some refactoring that I did in emission back before christmas, to make the sticky tab page header height dynamic, and generally clean up the sticky tab page code by using react context.

- Makes the active tab bottom border animated and 2px high, which @samchieng has been waiting patiently for 🙈

  animated-border

- Fixes some strictness issues in the artist page and sticky header code.


src/lib/Components/StickyTabPage/StickyTabPage.tsx
```diff
@@ -40,15 +68,12 @@ export const StickyTabPage: React.FC<{
...
-    // we just set it now to avoid having to trigger an 'onLayout' in every test
-    __TEST__ ? new Animated.Value(300) : null
-  )
+  const activeTabIndexNative = useAnimatedValue(initialTabIndex)
+  const [activeTabIndex, setActiveTabIndex] = useGlobalState(initialTabIndex)
+  const { jsx: staticHeader, nativeHeight: staticHeaderHeight } = useAutoCollapsingMeasuredView(staticHeaderContent)
   const tracking = useTracking()
   const headerOffsetY = useAnimatedValue(0)
-  const railRef = useRef<SnappyHorizontalRail>()
+  const railRef = useRef<SnappyHorizontalRail>(null)
```
ashfurrow, line 76: This seems like an interesting story?

src/lib/Components/StickyTabPage/StickyTabPageFlatList.tsx
```diff
@@ -83,22 +84,11 @@ export const StickyTabPageFlatList: React.FC<StickyTabFlatListProps> = props =>
...
-            space(3)
-          ),
-        }}
-      />
-    ),
-    [headerHeight]
-  )
+  // We need to wait for the header to mount before showing the rest of the content
+  // to avoid jumping content in situations where certain items have fixed height.
+  // This doesn't make a lot of sense but sometimes you just go with what works ¯\_(ツ)_/¯
```
ashfurrow, line 89: 🙌

src/lib/Components/StickyTabPage/StickyTabPage.tsx
ds300: yeah... the typings for `useRef` right now are... interesting... there's three signature overloads:

```ts
    function useRef<T>(initialValue: T): MutableRefObject<T>;
    function useRef<T>(initialValue: T|null): RefObject<T>;
    function useRef<T = undefined>(): MutableRefObject<T | undefined>;
```

If you don't pass an initial value you end up with the last one, so in this case I'd get a `MutableRefObject<SnappyHorizontalRail | undefined>` but that's incompatible with passing the ref to an actual component, because when a component unmounts, react sets refs to `null` rather than `undefined`.

So when trying to do `<SnappyHorizontalRail ref={railRef} />` you end up with this error: 
```
Type 'MutableRefObject<SnappyHorizontalRail | undefined>' is not assignable to type '((instance: SnappyHorizontalRail | null) => void) | RefObject<SnappyHorizontalRail> | null | undefined'.
  Type 'MutableRefObject<SnappyHorizontalRail | undefined>' is not assignable to type 'RefObject<SnappyHorizontalRail>'.
    Types of property 'current' are incompatible.
      Type 'SnappyHorizontalRail | undefined' is not assignable to type 'SnappyHorizontalRail | null'.
        Type 'undefined' is not assignable to type 'SnappyHorizontalRail | null'.
```

So if you're using the ref with an actual component then you always want to pass an initial value of null to make sure you get that second overload. The other two are useful when you want to manually set the value of `ref.current` but in those cases you probably are not also passing the ref to a component.

PR #3257, opened 2020-04-30
[FX-1799] Implements small other collections category rail

dzucconi:
Okay let's try this again

-----

Re: [FX-1799](https://artsyproduct.atlassian.net/browse/FX-1799)

### DIFF
--- a/src/lib/Scenes/Collection/Components/CollectionHubsRails/OtherCollections/OtherCollectionsRail.tsx
+++ b/src/lib/Scenes/Collection/Components/CollectionHubsRails/OtherCollections/OtherCollectionsRail.tsx
@@ -1,0 +1,66 @@
+import { Sans, Spacer } from "@artsy/palette"
+import { OtherCollectionsRail_collectionGroup } from "__generated__/OtherCollectionsRail_collectionGroup.graphql"
+import { CardRailFlatList } from "lib/Components/Home/CardRailFlatList"
+import SwitchBoard from "lib/NativeModules/SwitchBoard"
+import React, { useRef } from "react"
+import { TouchableOpacity, View } from "react-native"
+import { createFragmentContainer, graphql } from "react-relay"
+import styled from "styled-components/native"
+
+export interface OtherCollectionsRailProps {
+  collectionGroup: OtherCollectionsRail_collectionGroup
+}
+
+export const CollectionGroupMemberPill = styled(Sans).attrs({
+  px: 3,
+  py: 2,
+  size: "3t",
+  bg: "black10",
+})`
+  overflow: hidden;
+  border-radius: 6px;
+`
+
+export const OtherCollectionsRail: React.FC<OtherCollectionsRailProps> = ({ collectionGroup: { name, members } }) => {
+  const ref = useRef<View | null>(null)
+
+  return (
+    <View ref={ref}>
+      <Sans size="4" m={2}>
+        {name}
+      </Sans>
+
+      <CardRailFlatList
+        data={members}
+        initialNumToRender={3}
+        keyExtractor={({ id }) => id}
+        ItemSeparatorComponent={() => <Spacer mr={0.5} />}
+        renderItem={({ item: { slug, title } }) => {
+          return (
+            <TouchableOpacity
+              onPress={() => {
+                SwitchBoard.presentNavigationViewController(ref.current!, `/collection/${slug}`)
+              }}
+            >
+              <CollectionGroupMemberPill>{title}</CollectionGroupMemberPill>
+            </TouchableOpacity>
+          )
+        }}
+      />
+    </View>
+  )
+}
+
+export const OtherCollectionsRailContainer = createFragmentContainer(OtherCollectionsRail, {
+  collectionGroup: graphql`
+    fragment OtherCollectionsRail_collectionGroup on MarketingCollectionGroup {
+      groupType
+      name
+      members {
+        id
+        slug
+        title
+      }
+    }
+  `,
+})
--- a/src/lib/Scenes/Collection/Components/CollectionHubsRails/OtherCollections/__tests__/OtherCollectionsRail-tests.tsx
+++ b/src/lib/Scenes/Collection/Components/CollectionHubsRails/OtherCollections/__tests__/OtherCollectionsRail-tests.tsx
@@ -1,0 +1,48 @@
+import { Sans, Theme } from "@artsy/palette"
+import SwitchBoard from "lib/NativeModules/SwitchBoard"
+import { CollectionHubRailsOtherCollectionsRailFixture } from "lib/Scenes/Collection/Components/__fixtures__/CollectionFixture"
+import React from "react"
+import { TouchableOpacity } from "react-native"
+import ReactTestRenderer from "react-test-renderer"
+import { CollectionGroupMemberPill, OtherCollectionsRail } from "../OtherCollectionsRail"
+
+jest.mock("lib/NativeModules/SwitchBoard", () => ({ presentNavigationViewController: jest.fn() }))
+
+describe("Other Collections Rail", () => {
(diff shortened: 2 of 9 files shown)


src/lib/Scenes/Collection/Components/CollectionHubsRails/OtherCollections/OtherCollectionsRail.tsx
```diff
@@ -0,0 +1,67 @@
...
+      <CardRailFlatList
+        data={members}
+        initialNumToRender={3}
+        keyExtractor={({ id }) => id}
+        ItemSeparatorComponent={() => <Spacer mr={0.5} />}
+        renderItem={({ item: { slug, title } }) => {
+          return (
+            <TouchableOpacity
+              onPress={() => {
+                SwitchBoard.presentNavigationViewController(ref.current!, `/collection/${slug}`)
```
dzucconi, line 43: It is not clear to me at all what ref is being used for. Any insight?

src/lib/Scenes/Collection/Components/CollectionHubsRails/index.tsx
```diff
@@ -3,27 +3,30 @@ import { CollectionHubsRails_linkedCollections } from "__generated__/CollectionH
...
   const { linkedCollections } = props
-  const railForGroupType = (collectionGroup: any) => {
-    const { groupType } = collectionGroup
-    switch (groupType) {
-      case "ArtistSeries":
-        return <TrendingArtistSeriesRail collectionGroup={collectionGroup} {...props} />
-      default:
-        return null
-    }
-  }
```
dzucconi, line 21: Inlined this so that we get the implicit typing in the switch statement.

src/lib/Scenes/Collection/Components/CollectionHubsRails/OtherCollections/OtherCollectionsRail.tsx
ashleyjelks: https://github.com/artsy/eigen/blob/0a89fde623f09802313753a575ea7119127ca1e0/src/lib/NativeModules/SwitchBoard.tsx#L13

`presentNavigationViewController` takes 2 params--the component being navigated from and the route being navigated to....we use ref here since its not a class based component and we therefore cannot use `this`.

dzucconi: So what's it actually doing with it though? Like I noticed that https://github.com/artsy/eigen/blob/master/src/lib/Components/GenericArtistSeriesRail.tsx#L54 here this is just overwriting the ref for every item in that collection

ashleyjelks: Not sure what you mean by "what's it actually doing with it though",  but you're right about the ref being overwritten in the GenericArtistSeriesRail, I'll make that change in follow up PR. It should only be on a single element in the component, not each of the cards in the Flatlist.

dzucconi: I guess, why does it need it in a reference to the current component in the first place? What's the function ultimately doing with that data?

src/lib/Scenes/Collection/Components/CollectionHubsRails/OtherCollections/OtherCollectionsRail.tsx
```diff
@@ -0,0 +1,67 @@
+import { Sans, Spacer } from "@artsy/palette"
+import { OtherCollectionsRail_collectionGroup } from "__generated__/OtherCollectionsRail_collectionGroup.graphql"
+import { CardRailFlatList } from "lib/Components/Home/CardRailFlatList"
+import SwitchBoard from "lib/NativeModules/SwitchBoard"
+import React, { useRef } from "react"
+import { TouchableOpacity, View } from "react-native"
+import { createFragmentContainer, graphql } from "react-relay"
+// @ts-ignore STRICTNESS_MIGRATION
```
ds300, line 8: you can delete this, we have typings for styled-components in master now

src/lib/Scenes/Collection/Components/CollectionHubsRails/OtherCollections/OtherCollectionsRail.tsx
ds300: @dzucconi our Obj-c routing code needs a reference to a component in the current react-native view to find out which tab's nav stack to push the new view to.

src/lib/Scenes/Collection/Components/CollectionHubsRails/OtherCollections/__tests__/OtherCollectionsRail-tests.tsx
```diff
@@ -0,0 +1,69 @@
+import { Theme } from "@artsy/palette"
+// @ts-ignore STRICTNESS_MIGRATION
+import { mount } from "enzyme"
```
ds300, line 3: we're migrating away from enzyme. Please check out tests that use the package `relay-test-utils` for future testing style :)

dzucconi: Ah cool — will update.

ashleyjelks: @ds300 what was the rationale for why we want to move away from enzyme? I got the memo, but never knew the reasoning behind it.

src/lib/Scenes/Collection/Components/CollectionHubsRails/OtherCollections/OtherCollectionsRail.tsx
ashleyjelks: Nice!

src/lib/Scenes/Collection/Components/CollectionHubsRails/OtherCollections/__tests__/OtherCollectionsRail-tests.tsx
ds300: It's slow and doesn't work that well with react-native (it is based on jsdom).

src/lib/Scenes/Collection/Components/CollectionHubsRails/OtherCollections/OtherCollectionsRail.tsx
```diff
@@ -0,0 +1,67 @@
...
+}
+
+export const CollectionGroupMemberPill = styled(Sans).attrs({
+  px: 3,
+  py: 2,
+  size: "3t",
+  bg: "black10",
+})`
+  overflow: hidden;
+  border-radius: 6px;
```
dzucconi, line 22: Oh, meant to ask. Do we have a token for this value? I don't want to use literal values where possible.

dzucconi:
Looks like related to strictness migration. @ds300 how do I check this locally? I'm not seeing any type errors/warnings locally?

ds300:
this check fails if you increase the total number of comments in the codebase which include the text `STRICTNESS_MIGRATION`. You can ignore it if you're in a rush, but we're trying to reduce the number of these issues by cleaning things up, so if you could help out on that front it would be appreciated.

dzucconi:
@ds300 Sure I get that — I'm happy to abide by strict — I'm just not seeing any warnings from Typescript locally.

ds300:
that's right, this check is separate from typescript. It only looks for the string `STRICTNESS_MIGRATION` and counts the number of occurrences.

dzucconi:
OH. Got it.

PR #3258, opened 2020-04-30
Removes incorrectly placed ref

ashleyjelks:
A ref needs to be attached to the GenericArtistSeriesRail component to support the use of Switchboard's `presentNavigationViewController` method. This PR moves that ref to a View wrapping the entire component.


dzucconi:
LGTM

PR #3268, opened 2020-05-04
Adds Ways to Buy Filter to Collections

ashleyjelks:
This PR introduces the latest filter--Ways to Buy, the first one that allows for multiple selections to Collections Artworks.

Kapture 2020-05-12 at 18 37 52
Kapture 2020-05-12 at 18 40 00

#skip_new_tests

https://artsyproduct.atlassian.net/browse/FX-1718


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>Missing Test Files:

- `src/lib/Components/ArtworkFilterOptions/__tests__/FilterHeader-tests.tsx`
- `src/lib/Components/ArtworkFilterOptions/__tests__/FilterToggleButton-tests.tsx`
- `src/lib/Components/ArtworkFilterOptions/__tests__/MultiSelectOption-tests.tsx`
- `src/lib/Components/ArtworkFilterOptions/__tests__/WaysToBuyOptions-tests.tsx`

    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/lib/Components/ArtworkFilterOptions/WaysToBuyOptions.tsx
```diff
@@ -0,0 +1,80 @@
...
+interface WaysToBuyOptionsScreenProps {
+  navigator: NavigatorIOS
+}
+
+export const WaysToBuyOptionsScreen: React.SFC<WaysToBuyOptionsScreenProps> = ({ navigator }) => {
+  const { dispatch, state } = useContext(ArtworkFilterContext)
+
+  const filterType = "waysToBuy"
+
+  const selectOption = (option: WaysToBuyOptions) => {
```
sweir27, line 17: In general, this seems complex! Trying to offer a simplification but you certainly know the intricacies better than me. 😄 

Here are some thoughts/ideas:
- Why do we need the `on` and `off` arrays? It seems like we should be able to deduce the state from the global store just based on `appliedFilters` and `selectedFilters`, similar to how we do the rest of the toggles. I'd try to tie the state of the toggles as much as possible to our existing state!
- If we are able to do that, then the `selectOption` function can be simple (just needs to send the filter type and its own filter option)! And we can move this logic in our reducers. 

Curious if you tried anything like that? Happy to pair or take a deeper look later!

ashleyjelks: There was no way to capture a de-selected filter state without changing how filter states are defined. 

For example, it's pretty clear that a selected or an applied filter should toggle `on`.  However, when you apply a filter and then de-select it, that action _should_ toggle `off`, but this action still adds the filter to the `selectedFilter` state, so it stays toggled `on`. 

One approach there could be adding more logic to the `selectedFilter` reducer to check if the filter is already applied, and if applied, then to skip adding this filter to the `selectedFilter` state,  and _then_ if a user selects "Apply" to deselect the action then dispatch some 'unApplyFilter` reducer to remove it from the appliedFilters state. But that's also pretty complex. So that's why I went the route of updating the toggle states based on the possible toggle actions, to get it working without bugs. But you're right, I think we should pair on it because I'm sure there's a better way to reduce the complexity here.

src/lib/Components/ArtworkFilterOptions/MultiSelectOption.tsx
```diff
@@ -0,0 +1,69 @@
...
+import React from "react"
+import { FlatList, TouchableWithoutFeedback } from "react-native"
+import NavigatorIOS from "react-native-navigator-ios"
+import styled from "styled-components/native"
+
+type MultiSelectOptions = WaysToBuyOptions
+
+interface MultiSelectOptionScreenProps {
+  navigator: NavigatorIOS
+  filterText: "Ways to Buy"
```
ashfurrow, line 14: Minor thing, but this `MultiSelectOptionScreen` seems to be totally generic (for any set of multiple selections) except for this line, which is specific to "Ways to Buy" – what do you think about moving this text to the parent component?

src/lib/Components/ArtworkFilterOptions/FilterHeader.tsx
```diff
@@ -0,0 +1,43 @@
...
+import React from "react"
+import { TouchableOpacity } from "react-native"
+import styled from "styled-components/native"
+
+interface ArtworkFilterHeaderProps {
+  filterName: "Sort" | "Medium" | "Price Range" | "Ways to Buy"
+  handleBackNavigation: () => void
+}
+
+export const ArtworkFilterHeader: React.FC<ArtworkFilterHeaderProps> = props => {
```
sweir27, line 11: Nice extraction!

src/lib/Components/ArtworkFilterOptions/WaysToBuyOptions.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export const WaysToBuyOptionsScreen: React.SFC<WaysToBuyOptionsScreenProps> = ({ navigator }) => {
+  const { dispatch } = useContext(ArtworkFilterContext)
+  const selectedOptions = useSelectedOptionsDisplay()
+
+  const selectOption = (option: WaysToBuyOptions) => {
+    const value = !selectedOptions.find(filter => filter.filterType === mapWaysToBuyTypesToFilterTypes[option])
+      ?.value as boolean
+
+    dispatch({
```
sweir27, line 24: 🙌 niiiice!

src/lib/Components/ArtworkFilterOptions/__tests__/WaysToBuyOptions-tests.tsx
```diff
@@ -0,0 +1,210 @@
...
+    mockNavigator = new MockNavigator()
+    state = {
+      selectedFilters: [],
+      appliedFilters: [],
+      previouslyAppliedFilters: [],
+      applyFilters: false,
+    }
+  })
+
+  // @ts-ignore STRICTNESS_MIGRATION
```
sweir27, line 27: Curious what the error is if you remove this?

src/lib/Components/ArtworkFilterOptions/__tests__/WaysToBuyOptions-tests.tsx
```diff
@@ -0,0 +1,210 @@
...
+        { value: true, filterType: "waysToBuyBid" },
+      ],
+      appliedFilters: [],
+      previouslyAppliedFilters: [],
+      applyFilters: false,
+    }
+
+    const filterModal = mount(<MockFilterScreen initialState={state} />)
+
+    expect(
```
sweir27, line 111: #minor but I would just verify the full text string here 

```
    expect(
      filterModal
        .find(FilterModalOptionListItem)
        .at(3)
        .text()
    ).toEqual("Buy now, Bid, Inquire")
```

Has the added benefit of verifying the order in which they show up + how they are joined together!

src/lib/Components/FilterModal.tsx
```diff
@@ -144,6 +154,11 @@ export const FilterOptions: React.SFC<FilterOptionsProps> = props => {
       filterType: "priceRange",
       FilterScreenComponent: PriceRangeOptionsScreen,
     },
+    {
+      filterText: "Ways to Buy",
+      filterType: "waysToBuyBuy",
```
sweir27, line 159: This doesn't seem quite right since it's only capturing one of the options-- if this is only used below, should it just be `waysToBuy`?

src/lib/Scenes/Collection/Helpers/FilterArtworksHelpers.ts
```diff
@@ -5,14 +5,23 @@ const defaultFilterParams = {
...
+const applyFilters = (appliedFilters: FilterArray, filterParams: FilterParams) => {
   appliedFilters.forEach(appliedFilterOption => {
     const paramMapping = filterTypeToParam[appliedFilterOption.filterType]
-    const paramFromFilterType = paramMapping[appliedFilterOption.value]
-    // @ts-ignore STRICTNESS_MIGRATION
-    filterParams[appliedFilterOption.filterType] = paramFromFilterType
+    const paramFromFilterType = paramMapping[appliedFilterOption.value as SortOption | MediumOption | PriceRangeOption]
+
+    if (appliedFilterOption.value === true) {
+      const mapParamToRelayValue = paramMapping[appliedFilterOption.filterType]
```
sweir27, line 20: #superminor `mapParamToRelayValue` is slightly confusing naming ("relay value" could mean many things, but I think here we're aiming to retrieve the ultimate query param that's sent to the API)-- I might call it `mapToRelayParam` or just `queryParam` ?

src/lib/Scenes/Collection/Helpers/FilterArtworksHelpers.ts
```diff
@@ -135,22 +144,106 @@ export const OrderedPriceRangeFilters: PriceRangeOption[] = [
...
   sort: any
   medium: any
   priceRange: any
+  waysToBuyBuy: any
+  waysToBuyBid: any
+  waysToBuyInquire: any
+  waysToBuyMakeOffer: any
+}
+
+enum WaysToBuyRelayFilters {
```
sweir27, line 198: This also appears to be redundant with `WaysToBuyFilters` and one of the other data types.

(I'd also avoid the `Relay-` naming as the query params are just about data fetching in general and aren't relay-specific.)

src/lib/Scenes/Collection/Helpers/FilterArtworksHelpers.ts
```diff
@@ -135,22 +144,106 @@ export const OrderedPriceRangeFilters: PriceRangeOption[] = [
...
 
+// Ways to Buy types
+enum WaysToBuyFilters {
+  "Buy now" = "acquireable",
+  "Make offer" = "offerable",
+  "Bid" = "atAuction",
+  "Inquire" = "inquireableOnly",
+}
+
+export enum WaysToBuyFilterTypes {
```
sweir27, line 155: `WaysToBuyFilterTypes`, `mapWaysToBuyTypesToFilterTypes`, and `mapWaysToBuyFilters` all seem to encode the same information. We should aim to have just one representation, even if it's a little extra data manipulation or TS finagling where they're currently used. Let me know if I can help!

src/lib/Scenes/Collection/Helpers/FilterArtworksHelpers.ts
```diff
@@ -135,22 +144,106 @@ export const OrderedPriceRangeFilters: PriceRangeOption[] = [
...
-export const filterTypeToOrderedOptionsList: FilterTypes = {
-  sort: OrderedArtworkSorts,
-  medium: OrderedMediumFilters,
-  priceRange: OrderedPriceRangeFilters,
+// Types for the parameters passed to Relay
+type MultiOptionRelayParams = "acquireable" | "inquireableOnly" | "atAuction" | "offerable"
+
+type SingleOptionRelayParams = "sort" | "medium" | "priceRange"
+
+interface FilterParams {
```
sweir27, line 222: Nice typing! 👏

src/lib/Components/ArtworkFilterOptions/MultiSelectOption.tsx
ashleyjelks: Making this a separate component was an effort to future-proof any new multiple selection filters so the styling/layout can be maintained in a single component. But I heard we might not be adding the color and size filters, in which case it makes sense to do this!

src/lib/Components/ArtworkFilterOptions/__tests__/WaysToBuyOptions-tests.tsx
ashleyjelks: `Binding element 'initialState' implicitly has an 'any' type.`

ashleyjelks: I could tried typing the `intialState` parameter to match the shape of it's data:

`export interface ArtworkFilterContextState {
  readonly appliedFilters: FilterArray
  readonly selectedFilters: FilterArray
  readonly previouslyAppliedFilters: FilterArray
  readonly applyFilters: boolean
}`

But then we get this error: `Property 'initialState' does not exist on type 'ArtworkFilterContextState'.`

I think the workaround is to pass each typed prop individually, but is extremely verbose (especially since this component is used all over our tests). 

If you have suggestions on a better workaround I'd love to pair/hear them!

ashleyjelks: Actually, I just investigated it a little further and we can create an Interface that matches the shape of the params we pass to `MockWaysToBuyScreen`!

src/lib/Components/FilterModal.tsx
ashleyjelks: If it's clearer for people reading the code, I can change all the instances of `waysToBuyBuy`-> `waysToBuy`. The reason I did this was to avoid this ts error: `Type '"waysToBuy"' is not assignable to type '"sort" | "medium" | "priceRange" | "waysToBuyBuy" | "waysToBuyBid" | "waysToBuyInquire" | "waysToBuyMakeOffer"'`  This array expects one of the filterTypes that corresponds to the screen that filterType would appear on.  So filterType `sort` shows the SortScreen, filterType `medium` shows the MediumScreen, etc. The issue with Ways to Buy filters is that there is no type `waysToBuy` hence my using one of the filterTypes that is used on the Ways to Buy Screen.

sweir27: I think it's just that the typing for `filterType` shouldn't be `FilterOption` in this case. On this screen, we _only_ want to allow `sort`, `medium`, `priceRange`, and `waysToBuy` because those are the only screens that we have.

PR #3269, opened 2020-05-05
Fastlane notify on beta deploy Sentry failures #trivial

ashfurrow:
Follow up from #3254. We worked on this in MX Knowledge Share. 

Basically, we want Sentry upload failures to be noisier but don't want them to totally fail the beta deploys. So we created a new Slack app with an incoming webhook URL and added that as `SLACK_URL` environment variable to Circle CI's config. Now we can use fastlane's [`slack` action](http://docs.fastlane.tools/actions/slack/#slack) to post messages. We played around with the options a bit too.

We also removed the auto-generated README file, since it was outdated and not actually useful.


pvinis:
#mergeongreen

PR #3272, opened 2020-05-05
[Refactor] Move consign app to Scenes folder

damassi:
Noticed that we've been moving apps to `Scenes` top-level folder. This brings the consign app over.

#trivial


jonallured:
Good to know!!

PR #3274, opened 2020-05-06
[Consignments] Build out static v2 landing

damassi:
This builds out a static layout for the v2 page. Next up will wire in dynamic data.

app

#trivial


jonallured:
Cool, nice prototyping! 😍

PR #3277, opened 2020-05-06
Adds tracking to collection rails on artist and collection pages

sweir27:
Addresses: https://artsyproduct.atlassian.net/browse/FX-1941

From collection page
Screen Shot 2020-05-06 at 9 32 48 AM

From artist page
Screen Shot 2020-05-06 at 9 33 07 AM


src/lib/Components/GenericArtistSeriesRail.tsx
```diff
@@ -54,6 +63,19 @@ export const GenericArtistSeriesRail: React.FC<GenericArtistSeriesRailProps> = (
             <CardRailCard
               key={index}
               onPress={() => {
+                tracking.trackEvent({
+                  context_module: Schema.ContextModules.ArtistSeriesRail,
+                  context_screen_owner_type: contextScreenOwnerType,
+                  context_screen_owner_id: contextScreenOwnerId,
```
sweir27, line 69: This and the `slug` need to be passed in explicitly because they're governed by the parent component.

src/lib/Scenes/Collection/Components/__fixtures__/CollectionFixture.ts
```diff
@@ -360,112 +360,6 @@ export const FullFeaturedArtistListCollectionFixture: FullFeaturedArtistListTest
   },
 }
 
-export const CollectionHubRailsArtistSeriesFixture = {
```
sweir27, line 363: If the fixtures are only used in one place, it's better to define them just within that file (otherwise refactors become more difficult as we don't analyze when exported vars aren't being used!)

src/lib/Scenes/Collection/Components/CollectionHubsRails/ArtistSeries/__tests__/CollectionArtistSeriesRail-tests.tsx
```diff
@@ -10,246 +11,438 @@ import {
...
 
-  it("renders three artist series in the Trending Artists Series", () => {
-    const wrapper = mount(
-      <Theme>
-        <CollectionArtistSeriesRail {...props} />
-      </Theme>
-    )
-
-    expect(wrapper.find(ArtworkImageContainer)).toHaveLength(3)
+  it("renders without throwing an error", async () => {
```
sweir27, line 70: Heads up @ashleyjelks for these kinds of tests, we need to both render the component + assert _something_ (otherwise it'll silently fail!).

src/lib/Scenes/Collection/Components/__fixtures__/CollectionFixture.ts
ashleyjelks: Makes sense!

PR #3279, opened 2020-05-06
Show number of full days in emission Ticker, not just number of days in month

erikdstock:
Completes 🔒 [[AUCT-999]](https://artsyproduct.atlassian.net/browse/AUCT-999)
Paired with @brainbicycle 

This PR updates the react native countdown timer to count number of full days left in the timer using [`Duration.asDays()`](https://momentjs.com/docs/#/durations/days/) - `days()` only returns the number in the current month.

_possible followup: should we ever include the year in the 'timer ends May 6, blah blah blah am EDT' part below the timer? I don't think this is very important since the year would be implied by anything except a multi-year auction - eg if it is november but this sale ends in january, i don't need to know what year, but if it ends in december of the following year that would be relevant._

_by the way: while investigating this I saw a similar bug in the main auction view, which uses the (Objective C) [ARCountdownView.m]() on the same auction, with the live start time set 4 years out- it was just showing `00 days`. Today pairing with Brian we could not reproduce this bug and it was suddenly working correctly._

Fixed version.
image

[AUCT-999]: https://artsyproduct.atlassian.net/browse/AUCT-999

### DIFF
--- a/src/lib/Components/Countdown/__tests__/Ticker-tests.tsx
+++ b/src/lib/Components/Countdown/__tests__/Ticker-tests.tsx
@@ -20,6 +20,13 @@ describe("SimpleTicker", () => {
     const comp = render(<SimpleTicker duration={zeroDuration} separator="  " size="5" />)
     expect(comp.text()).toEqual("00d  00h  00m  00s")
   })
+
+  it("renders properly with days overflowing a single", () => {
+    // 2 years
+    const farOutDuration = moment.duration(63113904000)
+    const comp = render(<SimpleTicker duration={farOutDuration} separator="  " size="5" />)
+    expect(comp.text()).toContain("730d")
+  })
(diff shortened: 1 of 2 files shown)


src/lib/Components/Countdown/__tests__/Ticker-tests.tsx
```diff
@@ -20,6 +20,13 @@ describe("SimpleTicker", () => {
     const comp = render(<SimpleTicker duration={zeroDuration} separator="  " size="5" />)
     expect(comp.text()).toEqual("00d  00h  00m  00s")
   })
+
+  it("renders properly with days overflowing a single", () => {
+    // 2 years
+    duration = moment.duration(63113904000)
```
ashfurrow, line 26: Nice test 👍

PR #3280, opened 2020-05-06
Removes Collection Artworks Filters Feature Flag

ashleyjelks:
Collections Artworks filters are ready to go out with the next app release. This PR removes the feature flag that had to be enabled in order to interact with the filters.

As we still have filters that in progress/not ready for release, new filter additions will use the `AROptionsFilterCollectionsArtworks` native module, so for now we will keep this flag active, and deprecate completely once the rest of the in-progress and/or upcoming filters don't need to be hidden.

https://artsyproduct.atlassian.net/browse/FX-1942

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -13,6 +13,7 @@ upcoming:
     - Fix for dark mode tracking - brian, david, pavlos, mike
     - Adds a toggle button component - ashley
     - Add city guide entry in search behind flag - brian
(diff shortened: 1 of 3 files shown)


sweir27:
Sounds good!! I'm a little more bullish 🙃 and think we could probably just 🔪 the feature flag entirely and ship things as they come! But not bad to keep around for now.

ashfurrow:
Yay! So exciting 🎉

PR #3283, opened 2020-05-07
Scroll to top of Collections Artworks Grid after Applying Filters

ashleyjelks:
**Standard Collection**
Kapture 2020-05-11 at 14 54 24

**Department Collection**
Kapture 2020-05-11 at 14 55 46

https://artsyproduct.atlassian.net/browse/FX-1921

#trivial

### DIFF
--- a/src/lib/Scenes/Collection/Collection.tsx
+++ b/src/lib/Scenes/Collection/Collection.tsx
@@ -2,7 +2,7 @@ import { Box, color, FilterIcon, Flex, Sans, Spacer, Theme } from "@artsy/palett
 import { CollectionQuery } from "__generated__/CollectionQuery.graphql"
 import { defaultEnvironment } from "lib/relay/createEnvironment"
 import renderWithLoadProgress from "lib/utils/renderWithLoadProgress"
-import React, { Component } from "react"
+import React, { Component, createRef } from "react"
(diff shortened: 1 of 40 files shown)


sweir27:
The implementation looks okay, but it feels a little jarring in the gifs! I wonder if we can make it slower? Nicole might have some animation specs/ideas. Maybe we can ease it?

ashleyjelks:
> The implementation looks okay, but it feels a little jarring in the gifs! I wonder if we can make it slower? Nicole might have some animation specs/ideas. Maybe we can ease it?

Actually just posted a video in Slack of what it looks like with and without the animation just as you noted this!

src/lib/Scenes/Collection/Collection.tsx
```diff
@@ -116,9 +134,14 @@ export class Collection extends Component<CollectionProps, CollectionState> {
                             <CollectionHubsRails linkedCollections={linkedCollections} {...this.props} />
                           ) : null
                         case "collectionArtworks":
+                          const scrollIndex = isDepartment ? 2 : 1
```
sweir27, line 137: If we're going to `- 1` above, can we just say `isDepartment ? 1 : 0` here? Nots sure if there's another use for the `scrollIndex` that I'm missing!

sweir27:
Nice!! Looks awesome 😍!

PR #3284, opened 2020-05-07
[MX-291] Adds collections to home page

ashfurrow:
This depends on this Metaphysics PR: https://github.com/artsy/metaphysics/pull/2361 ~We'll need to wait for that to merge and for the schema to get updated in Eigen first, but I have included the changes to the generated files in this PR (CI will fail).~

Screen Shot 2020-05-07 at 13 35 16

The `CollectionsRail` component is heavily based on the existing `SalesRail`, so I think it makes sense to create an abstraction for these cards. But we can follow-up with that. 

I'm _very_ excited to see such an important screen like collections finally get a well-earned entry point from the home screen of the app 🎉 /cc @ashleyjelks

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -23,6 +23,7 @@ upcoming:
     - Updated artwork tile design - ash & brian
     - Improve UX for Home Artist rail - david
     - Fixes missing artworks on partner page - ash
+    - Adds collections to home page - ash
 
 releases:
   - version: 6.4.3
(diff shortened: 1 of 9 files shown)


ashleyjelks:
This good for a merge after CI checks are green, lmk if you want to pair on these!

ashleyjelks:
Yay, so exciting!

ashfurrow:
Cool, thanks for the review @brainbicycle! I'm going to mark as #mergeongreen.

PR #3289, opened 2020-05-08
Scrolls rails back to starting position without animation #trivial

ashfurrow:
I'm 99% sure we don't want to animate these offset changes, since it draws attention to the user interface when pulling-to-refresh.

But I've left this in its own commit/PR so we can revert it if I'm wrong 👍


pvinis:
Oops. I guess I should have asked, my bad. I assumed it would be animated.

ashfurrow:
@pvinis no worries! In general, I _do_ think animating things is better than not, just not in this case specifically 👍

PR #3290, opened 2020-05-08
Updates the app map

ashfurrow:
This came up in a recent retro. I've tried to elevate the abstraction level a bit: less "this is how to code a specific thing" and more "these are our goals." I also included some historical notes on changes we've made since the map was originally written.

### DIFF
--- a/docs/map_to_the_app.md
+++ b/docs/map_to_the_app.md
@@ -1,8 +1,8 @@
 # :world_map:
 
-As we have built our React Native components, our understanding of how to build those components as evolved. As that understanding grows, newer components use newer techniques and older ones are often left un-updated. It can be difficult to orient oneself around what the current preferred practices are.
+As we have built our app in React Native, our understanding of how to build software has evolved. As our understanding grows, newer code uses newer techniques. Older code is often left un-updated. It can be difficult to orient oneself around what the current preferred practices are.
 
-This document is a map. Not of Eigen/Emission at a specific time, but a map of how we got here and where we want to go next. This is a living document, expected to be updated regularly, of links to:
+This document is a map. Not of Eigen at a specific time, but a map of how we got here and where we want to go next. This is a living document, expected to be updated regularly, of links to:
 
 - Example code.
 - Pull requests with interesting discussions.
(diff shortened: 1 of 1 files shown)


docs/map_to_the_app.md
```diff
@@ -13,17 +13,63 @@ Links should point to specific commits, and not a branch (in case the branch or
...
+
+```sh
+touch .i-am-helping-out-with-the-strictness-migration
+```
+
+### File Structure
+
+Everything in `src/` is React Native. Within this folder, things can be a bit of a mess. We are working on cleaning it up. Check back here for more later.
+
+**TODO**: Figure out what we want our directory structure to be and define it here.
```
ashfurrow, line 62: I remember seeing @damassi write up a really great description of what we want here, I think for some of the consignments work. I couldn't find it anywhere, though. I liked it. Chris, do you remember what I'm talking about?

docs/map_to_the_app.md
```diff
@@ -42,14 +94,13 @@ Unit testing on Emission is a bit all over the place. Here are some great exampl
...
+  - [`CollectionsRail` tests](https://github.com/artsy/eigen/blob/39644610eb2a5609d992f434a7b37b46e0953ff4/src/lib/Scenes/Home/Components/__tests__/CollectionsRail-tests.tsx) demonstrate `relay-test-utils`.
 
-### Eigen Interop
-
-Emission is used by Eigen, our main app. The structure is [described in this blog post](http://artsy.github.io/blog/2016/08/24/On-Emission/). Interop between the JavaScript (Emission) and the Objective-C/Swift (Eigen) can be tricky.
+### Interop
 
-_Most_ interactions are made through a "switchboard" to open links. A switchboard exists in Eigen, which users see, and a switchboard exists in the Emission test app for our internal testing. This is key: the switchboard in Emission isn't actually used in production, only the Eigen one is.
+Our React Native code ("Emission") is used by our Native code ("Eigen"). They used to be two repositories but were [combined in February 2020](https://github.com/artsy/eigen/pull/3030). Traces of the separation remain. The structure we originally took is [described in this blog post](http://artsy.github.io/blog/2016/08/24/On-Emission/). Interop between JavaScript and Native can be tricky.
```
brainbicycle, line 101: 🎉 I think this is a lot more clear.

docs/map_to_the_app.md
```diff
@@ -13,17 +13,63 @@ Links should point to specific commits, and not a branch (in case the branch or
...
+- The Auction view controller.
+- The SwitchBoard ([see below](#interop)) to navigate between view controllers.
+- The top-level tab bar, and each tab's navigation controller.
+- Deep-link and notification handling (via SwitchBoard).
+- Analytics for Native UI.
+- Initializing the React Native runtime.
+
+Everything else is written in React Native.
+
 ### React Native
```
brainbicycle, line 31: Since this is a list of preferred practices I wonder if it makes sense for these headers to be explicit summaries of the preferred practice described? Something like:

```
### Use React Native for new feature development
### Leverage typescript to prevent runtime bugs
### Keep the file structure organized (that might need some clarity)
### Use Relay for network requests
### Write unit tests for new components

Eigen interop title is not as obvious because it is also a description of current state of affairs and maybe not our long term preferred practice? 

Maybe something like:
### Use the Native Switchboard for navigation (for now)
```

ashfurrow: Yeah I like that a lot, I'll update 👍

PR #3293, opened 2020-05-11
Adding some docs for the tracking

pvinis:
After the work on the tracking rework PR #3236 and some problems @mdole had using the changes for his PR on viewing rooms #3215, I thought it would be a good idea to document these, mostly for the gotchas that appeared, and the `tracks` object.

(docs are not #trivial 😛)


docs/analytics_and_tracking.md
```diff
@@ -0,0 +1,115 @@
...
+export class MyClassComp extends React.Component<Props> {
+  render() {
+    return <View />
+  }
+}
+
+const tracks = {
+  context: () => ({
+    context_screen: Schema.PageNames.Home,
+    context_screen_owner_type: null as any /* STRICTNESS_MIGRATION */
```
ashfurrow, line 69: Feels weird to include a `STRICTNESS_MIGRATION` comment here – maybe we could replace it with a stubbed value?

pvinis: 😅 indeed. Copy-paste fingers.

pvinis:
#mergeongreen

PR #3295, opened 2020-05-11
Finishing touches for 6.4.4 release #trivial

ashfurrow:
Mobbed on the changelog update and home module updates with MX, so I'm going to self-assign. The beta is already underway.

### DIFF
--- a/src/lib/Scenes/Home/Home.tsx
+++ b/src/lib/Scenes/Home/Home.tsx
@@ -78,15 +78,15 @@ const Home = (props: Props) => {
         type: "sales",
         data: salesModule,
       } as const),
+    fairsModule &&
+      ({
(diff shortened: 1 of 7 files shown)


src/lib/Scenes/Home/Home.tsx
```diff
@@ -78,16 +78,16 @@ const Home = (props: Props) => {
...
     collectionsModule &&
       ({
         type: "collections",
         data: collectionsModule,
       } as const),
-    salesModule &&
-      ({
-        type: "fairs",
-        data: fairsModule,
-      } as const),
```
ashfurrow, line 90: This was interesting – looks like a copy+paste error. We were checking `salesModule` but using `fairsModule` 😛

PR #3296, opened 2020-05-11
Refine static layout of sell tab's artist list

pepopowitz:
Addresses https://artsyproduct.atlassian.net/browse/CSGN-185

This PR updates the static layout for the artist list on the new sell tab's landing page to:

* use react-native/eigen concepts like `FlatList` for virtualization and `TouchableHighlight` for user feedback
* extract components for displaying lists of data
* display data in the same shape as what will come from MP/relay.

My next step will be to connect the component to MP via relay, and with these changes I think that should be almost all "plumbing" work.

## What it looks like to interact with these changes

sell-artist-list

#trivial #skip_new_tests (I anticipate including tests & changelog entry when this component gets closer to shippable.)

### DIFF
--- a/src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/ArtistList.tsx
+++ b/src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/ArtistList.tsx
@@ -1,0 +1,320 @@
+import { Box, color, Flex, Join, Sans, Spacer } from "@artsy/palette"
+import OpaqueImageView from "lib/Components/OpaqueImageView/OpaqueImageView"
+import SwitchBoard from "lib/NativeModules/SwitchBoard"
+import { chunk } from "lodash"
+import React, { useRef } from "react"
+import { FlatList, TouchableHighlight } from "react-native"
+
+export const ArtistList = () => {
+  const data = fakeData()
+  const chunksOfArtists = chunk(data.artists, 3)
+
+  return (
+    <FlatList
+      horizontal
+      ItemSeparatorComponent={() => <Spacer mr={3} />}
+      data={chunksOfArtists}
+      initialNumToRender={2}
+      renderItem={({ item }) => <StackOfArtists artists={item} />}
+      keyExtractor={(_item, index) => String(index)}
+    />
+  )
+}
+
+const StackOfArtists: React.FC<{ artists: Artist[] }> = ({ artists }) => {
+  return (
+    <Flex>
+      <Join separator={<Spacer mb={2} />}>
+        {artists.map(artist => {
+          return <ArtistItem artist={artist} key={artist.name} />
+        })}
+      </Join>
+    </Flex>
+  )
+}
+
+const ArtistItem: React.FC<{ artist: Artist }> = ({ artist }) => {
+  const navRef = useRef<any>()
+  const imageUrl = artist.image?.cropped?.url
+  const aspectRatio = artist.image?.cropped?.width / (artist.image?.cropped?.height || 1)
+
+  const handlePress = () => {
+    SwitchBoard.presentNavigationViewController(navRef.current, artist.href)
+  }
+
+  return (
+    <TouchableHighlight underlayColor={color("white100")} activeOpacity={0.8} onPress={handlePress} ref={navRef}>
+      <Flex flexDirection="row" alignItems="center" width="300">
+        <Box width={76} height={70} mr={1}>
+          {imageUrl && <OpaqueImageView aspectRatio={aspectRatio} imageURL={imageUrl} useRawURL />}
+        </Box>
+        <Sans size="4">{artist.name}</Sans>
+      </Flex>
+    </TouchableHighlight>
+  )
+}
+
+// NOTE: anything below here will be gone once we hook this component up to Relay
+
+interface Artist {
+  name: string
+  href: string
+  image: {
+    cropped: {
+      url: string
+      width: number
+      height: number
+    }
+  }
+}
+
+interface Data {
+  artists: Artist[]
+}
+
+function fakeData(): Data {
+  return {
+    artists: [
+      {
+        name: "Alex Katz",
+        href: "/artist/alex-katz",
+        image: {
+          cropped: {
+            url:
+              "https://d7hftxdivxxvm.cloudfront.net?resize_to=fill&width=76&height=70&quality=80&src=https%3A%2F%2Fd32dm0rphc51dk.cloudfront.net%2FbrHdWfNxoereaVk2VOneuw%2Flarge.jpg",
+            width: 76,
+            height: 70,
+          },
+        },
+      },
+      {
+        name: "Banksy",
+        href: "/artist/banksy",
+        image: {
+          cropped: {
+            url:
+              "https://d7hftxdivxxvm.cloudfront.net?resize_to=fill&width=76&height=70&quality=80&src=https%3A%2F%2Fd32dm0rphc51dk.cloudfront.net%2FX9vVvod7QY73ZwLDSZzljw%2Flarge.jpg",
+            width: 76,
+            height: 70,
+          },
+        },
+      },
+      {
+        name: "Katherine Bernhardt",
+        href: "/artist/katherine-bernhardt",
+        image: {
+          cropped: {
+            url:
+              "https://d7hftxdivxxvm.cloudfront.net?resize_to=fill&width=76&height=70&quality=80&src=https%3A%2F%2Fd32dm0rphc51dk.cloudfront.net%2FdAf2ClwBT2EElifyIe5Y0w%2Flarge.jpg",
+            width: 76,
+            height: 70,
+          },
+        },
+      },
+      {
+        name: "KAWS",
+        href: "/artist/kaws",
+        image: {
+          cropped: {
+            url:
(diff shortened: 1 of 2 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>Missing Test Files:

- `src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/__tests__/ArtistList-tests.tsx`

    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/ArtistList.tsx
```diff
@@ -0,0 +1,320 @@
...
+      height: number
+    }
+  }
+}
+
+interface Data {
+  artists: Artist[]
+}
+
+function fakeData(): Data {
```
damassi, line 75: Nice 👍

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/ArtistList.tsx
```diff
@@ -0,0 +1,320 @@
...
+  )
+}
+
+const ArtistItem: React.FC<{ artist: Artist }> = ({ artist }) => {
+  const navRef = useRef<any>()
+  const imageUrl = artist.image?.cropped?.url
+  const aspectRatio = artist.image?.cropped?.width / (artist.image?.cropped?.height || 1)
+
+  const handlePress = () => {
+    SwitchBoard.presentNavigationViewController(navRef.current, artist.href)
```
pepopowitz, line 42: Is this the right action to take here? In the demo video you can see that after tapping the artist, you go to the artist's page but stay within the "sell" tab on the bottom. I'm not sure if this is what we want or not.

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/ArtistList.tsx
```diff
@@ -0,0 +1,320 @@
...
+  }
+
+  return (
+    <TouchableHighlight underlayColor={color("white100")} activeOpacity={0.8} onPress={handlePress} ref={navRef}>
```
pepopowitz, line 46: the `underlayColor` and `activeOpacity` are based on examples I found in the codebase for similar lists. I am not sure if there's some official style guide for underlays & opacities, or general best practices, but I'm open to suggestions if this isn't right!

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/ArtistList.tsx
```diff
@@ -0,0 +1,320 @@
+import { Box, color, Flex, Join, Sans, Spacer } from "@artsy/palette"
+import OpaqueImageView from "lib/Components/OpaqueImageView/OpaqueImageView"
+import SwitchBoard from "lib/NativeModules/SwitchBoard"
+import { chunk } from "lodash"
+import React, { useRef } from "react"
+import { FlatList, TouchableHighlight } from "react-native"
+
+export const ArtistList = () => {
+  const data = fakeData()
+  const chunksOfArtists = chunk(data.artists, 3)
```
damassi, line 10: 👍

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/ArtistList.tsx
```diff
@@ -0,0 +1,320 @@
...
+      renderItem={({ item }) => <StackOfArtists artists={item} />}
+      keyExtractor={(_item, index) => String(index)}
+    />
+  )
+}
+
+const StackOfArtists: React.FC<{ artists: Artist[] }> = ({ artists }) => {
+  return (
+    <Flex>
+      <Join separator={<Spacer mb={2} />}>
```
damassi, line 27: Unrelated thinking: @dzucconi  has been talking "stack" components lately, and have been thinking it would be good if we added an alias to `Join`:

```
export const Stack = Join
```
(or somethething.)
 
It reads a bit better, follows some common patterns. 

I wanna exploring this https://github.com/mobily/stacks (for palette) and the article that spawned it: 
- https://mxstbr.com/thoughts/margin
- https://every-layout.dev/layouts/stack/#the-problem

dzucconi: I think if we do that we can do without the spacer component and just manipulate the margins though. We also want  to just be able to do <Stack spacing={6} /> <Stack spacing={6} direction="horizontal" /> and it work correctly (vs having to think about how to configure your spacer)

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/ArtistList.tsx
jonallured: Yeah, I'm curious here too. I wonder if we want to encourage people to bounce off this screen. 🤔 I would have expected those artist to not link to another page so that the only action one can take here is to begin consigning works. 🤷‍♂️

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/ArtistList.tsx
jonallured: 🧀

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/ArtistList.tsx
ashfurrow: Yup, that's right. The iOS paradigm has users able to move between tabs with their own independent navigation stacks. You can see this now by using the Saves & Follows tab to go to artists/artworks/etc 👍

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/ArtistList.tsx
ashfurrow: We don't have any standards yet but I'd like to set them soon. In any case, the values here are what I would pick 👍

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/ArtistList.tsx
```diff
@@ -0,0 +1,320 @@
...
+
+  const handlePress = () => {
+    SwitchBoard.presentNavigationViewController(navRef.current, artist.href)
+  }
+
+  return (
+    <TouchableHighlight underlayColor={color("white100")} activeOpacity={0.8} onPress={handlePress} ref={navRef}>
+      <Flex flexDirection="row" alignItems="center" width="300">
+        <Box width={76} height={70} mr={1}>
+          {imageUrl && <OpaqueImageView aspectRatio={aspectRatio} imageURL={imageUrl} useRawURL />}
```
ashfurrow, line 49: `OpaqueImageView` can either work with `aspectRaio` _or_ explicit `width` and `height`. `aspectRaio` typically comes from Metaphysics directly. If you always expect square images, I would suggest supplying those expected values with `width` and `height` because `OpaqueImageView` is going to make a request to Gemini for a resized image based on its width and height anyway: 

https://github.com/artsy/eigen/blob/086d41bcbad3b5bd7b4a2a9a3487c307d42cc0de/src/lib/Components/OpaqueImageView/OpaqueImageView.tsx#L109-L116

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/ArtistList.tsx
pepopowitz: ❤️ for something like `<Stack gap={3}>...</Stack>`. 

Related, this was a recent podcast episode I liked where Mark Dalgleish talked a lot about layout components - http://www.fullstackradio.com/134

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/ArtistList.tsx
pepopowitz: Happy to use `width` and `height` instead of `aspectRatio`, but just to clarify, since I'm also specifying `useRawURL` [no Gemini calls will actually be made, right?

https://github.com/artsy/eigen/blob/086d41bcbad3b5bd7b4a2a9a3487c307d42cc0de/src/lib/Components/OpaqueImageView/OpaqueImageView.tsx#L105-L109

ashfurrow: Ahhhh I totally missed that prop 🙈 Yup you should be fine either way then.

pepopowitz: 👍 In d4b94cd I updated to use `width` & `height` instead of `aspectRatio`, to avoid the unnecessary math.

ashfurrow: tenor-2

PR #3297, opened 2020-05-11
Update Viewing Rooms artwork rail

mdole:
This PR adds artist name and price to the rail in viewing rooms. The code is mostly copy/pasted from `SmallTileRail`, with a few tweaks to make it work with this flatlist.

@ashfurrow I'd love your perspective on the copy/paste-ness of this since I believe it was you who worked on implementing the home rail version. My thinking is that I would ideally like every place we use the same rail to rely on the same component so we don't get out of sync. However, I feel like that gets tricky when the Relay fragments are slightly different for each one, unless we're comfortable passing a bunch of stuff as props.

One thought I had was around making the `ArtworkCard` an all-purpose component that could be reused by both rails as a step in that direction.

What do you think? 

Screen Shot 2020-05-11 at 1 33 48 PM

#trivial

### DIFF
--- a/src/lib/Components/__tests__/ArtworkTileRail-tests.tsx
+++ b/src/lib/Components/__tests__/ArtworkTileRail-tests.tsx
@@ -1,0 +1,89 @@
+import { Theme } from "@artsy/palette"
+import { ArtworkTileRailTestsQuery } from "__generated__/ArtworkTileRailTestsQuery.graphql"
+import SwitchBoard from "lib/NativeModules/SwitchBoard"
+import { Schema } from "lib/utils/track"
+import React from "react"
+import { graphql, QueryRenderer } from "react-relay"
+import ReactTestRenderer from "react-test-renderer"
+import { useTracking } from "react-tracking"
+import { createMockEnvironment, MockPayloadGenerator } from "relay-test-utils"
+import { ArtworkCard, ArtworkTileRail, tappedArtworkGroupThumbnail } from "../ArtworkTileRail"
+
+jest.unmock("react-relay")
+jest.mock("lib/NativeModules/SwitchBoard", () => ({
+  presentNavigationViewController: jest.fn(),
+}))
+
+describe("ArtworkTileRail", () => {
+  let mockEnvironment: ReturnType<typeof createMockEnvironment>
+  const TestRenderer = () => (
+    <Theme>
+      <QueryRenderer<ArtworkTileRailTestsQuery>
+        environment={mockEnvironment}
+        query={graphql`
+          query ArtworkTileRailTestsQuery {
+            viewingRoom(id: "whatever") {
+              artworksConnection {
+                ...ArtworkTileRail_artworksConnection
+              }
+            }
+          }
+        `}
+        render={({ props, error }) => {
+          if (props?.viewingRoom) {
+            return (
+              <Theme>
+                <ArtworkTileRail
+                  artworksConnection={props.viewingRoom.artworksConnection! /* STRICTNESS_MIGRATION */}
+                  contextModule={Schema.ContextModules.ViewingRoomArtworkRail}
+                />
+              </Theme>
+            )
+          } else if (error) {
+            console.log(error)
+          }
+        }}
+        variables={{}}
+      />
+    </Theme>
+  )
+  beforeEach(() => {
+    mockEnvironment = createMockEnvironment()
+  })
+
+  it("navigates to an artwork + calls tracking when a card is tapped", () => {
+    const tree = ReactTestRenderer.create(<TestRenderer />)
+    mockEnvironment.mock.resolveMostRecentOperation(operation => {
+      const result = MockPayloadGenerator.generate(operation, {
+        ViewingRoom: () => ({
+          artworksConnection: {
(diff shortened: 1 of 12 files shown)


src/lib/Scenes/ViewingRoom/Components/ViewingRoomArtworkRail.tsx
```diff
@@ -11,14 +11,17 @@ import { createFragmentContainer, graphql } from "react-relay"
...
 
+const SMALL_TILE_IMAGE_SIZE = 120
+
 interface ViewingRoomArtworkRailProps {
   viewingRoom: ViewingRoomArtworkRail_viewingRoom
 }
 
-export const ArtworkCard = styled.TouchableHighlight`
+export const ArtworkCard = styled.TouchableHighlight.attrs({ underlayColor: color("white100"), activeOpacity: 0.8 })`
   border-radius: 2px;
```
ashfurrow, line 21: You've raised a really important point around copy/pasting similar code, versus building an abstraction @mdole. I think this line is a really interesting case because it's actually a bug I wrote ([subsequently fixed here, pending merge](https://github.com/artsy/eigen/pull/3295/files#diff-dbed83a74a4b4d30797245c9f377fc5aR39)). The `border-radius` is applied only to the top corners of the artwork, leaving the bottom corners sharp and jagged. 

Let me do some out-loud thinking...

Relay fragments can get tricky, especially when the design diverges. I think given that @samchieng worked with the design team to standardize on these new tiles, we can rely on them being stable for a while (and, if not stable, then at least all changing in the same way). 

My initial idea was that we could re-use the existing `SmallTileRail` totally as-is. It defines a fragment on `Artwork` with the `@relay(plural: true)` GraphQL directive here:

https://github.com/artsy/eigen/blob/086d41bcbad3b5bd7b4a2a9a3487c307d42cc0de/src/lib/Scenes/Home/Components/SmallTileRail.tsx#L60

But the component here uses an `ArtworksConnection` rather than an array of `Artwork`s. That's a quirk of the Home page GraphQL schema – everything else should be using `ArtworksConnection`. Hmm.

https://github.com/artsy/eigen/blob/02e0c3c8eef8162850505e39dc446513d303c479/src/lib/Scenes/ViewingRoom/Components/ViewingRoomArtworkRail.tsx#L114-L117

So I don't think we can re-use the components as-is, and I don't see an obvious path forward to totally share code. I think copy/pasting what you have here is fine for now, and then the next person/team to use this code should move it to another folder and re-use it. @mdole you could make that easier for them by breaking this component into two sub-components: the existing component on the `ViewingRoom` fragment, and a new one on `ArtworksConnection`. Then move the `FlatList` code to that new component. Or, if you're short on time, feel free to leave a comment explaining what we'd like to do. Actually, probably best to leave a comment anyway since this is a bit complicated 😅 

Let me know what I can clarify there.

src/lib/Components/__tests__/ArtworkTileRail-tests.tsx
```diff
@@ -0,0 +1,89 @@
...
+                node: {
+                  href: "/artwork/nicolas-party-rocks-ii",
+                  internalID: "5deff4b96fz7e7000f36ce37",
+                  slug: "nicolas-party-rocks-ii",
+                },
+              },
+            ],
+          },
+        }),
+      })
```
mdole, line 71: @ashfurrow we were actually super close on this! Something you said while pairing tipped me off to what we had wrong - I had initially put in `ArtworkConnection: () => ({...})`, and you noted that what we were doing was specifying "what we should return next time we get a request for `Thing`". As soon as we hung up I realized that `MockPayloadGenerator` probably worked the same way, and we had to specify what we should return next time we got a request for `ViewingRoom`, not `ArtworkConnection`. Worked great!

For reference, this is what we had implemented:
```ts
act(() =>
  mockEnvironment.mock.resolveMostRecentOperation({
    errors: [],
    data: {
      viewingRoom: {
        artworksConnection: {
          edges: [
            {
              node: {
                href: "/artwork/nicolas-party-rocks-ii",
                internalID: "5deff4b96fz7e7000f36ce37",
                slug: "nicolas-party-rocks-ii",
              },
            },
          ],
        },
      },
    },
  })
);
```

src/lib/Scenes/ViewingRoom/Components/ViewingRoomArtworkRail.tsx
mdole: Update: @ashfurrow and I did a bit of pairing on this and came up with an implementation that we're reasonably happy with as a next step. Hopefully this will make it simple for the next person who wants to build an artwork rail that relies on `artworksConnection`! Thanks for the excellent feedback, Ash 🙏

mdole:
Ok, going to go ahead and get this out! Happy with how it turned out; I think it'll be easier to iterate on this in the future to get to a more standardized approach across the app

src/lib/Components/__tests__/ArtworkTileRail-tests.tsx
ashfurrow: Ah, this makes sense! Thanks for following up 🙏

PR #3299, opened 2020-05-11
[MX-277] - Home analytics

brainbicycle:
https://artsyproduct.atlassian.net/browse/MX-277

Implements analytics for tap events on home rails. 
Integrates [cohesion](https://github.com/artsy/cohesion) 🎉


src/lib/Components/ArtworkGrids/ArtworkGridItem.tsx
```diff
@@ -36,6 +34,17 @@ export class Artwork extends React.Component<Props, any> {
...
 
+  trackArtworkTap() {
+    const trackTap = this.props.trackTap
+    const genericTapEvent = {
+      action_name: Schema.ActionNames.GridArtwork,
+      action_type: Schema.ActionTypes.Tap,
+      flow: this.props.trackingFlow,
+      context_module: this.props.contextModule,
+    }
+    trackTap ? trackTap(this.props.artwork.slug) : this.props.trackingProp?.trackEvent(genericTapEvent)
```
brainbicycle, line 45: Wasn't sure best way to do this but didn't want to break our existing tracking since we use this artwork grid in multiple places. Currently passing a trackTap function from home which will be called when present otherwise call the old tap track event. Open to suggestions on a better way.

src/lib/Components/ArtworkGrids/ArtworkGridItem.tsx
```diff
@@ -14,19 +15,16 @@ interface Props {
   onPress?: (artworkID: string) => void
   trackingFlow?: string
   contextModule?: string
+  trackTap?: (artworkSlug: string) => void
+  trackingProp?: TrackingProp
```
brainbicycle, line 19: Add a comment here to describe behavior + why

src/lib/utils/track/schema.ts
```diff
@@ -120,6 +120,7 @@ export enum PageNames {
   WorksForYou = "WorksForYou",
   Home = "Home",
   Auctions = "Auctions",
+  Auction = "Auction",
```
eessex, line 123: I believe that we use `sale` to refer to an individual auction, cc @mikehrom

src/lib/Scenes/Home/homeAnalytics.ts
```diff
@@ -0,0 +1,220 @@
...
+    eventType,
+  }: HomeEventData): TappedEntityGroup {
+    return {
+      action,
+      context_module: contextModule,
+      context_screen_owner_type: Analytics.OwnerType.home,
+      destination_screen_owner_type: destinationScreenOwnerType,
+      destination_screen_owner_id: id,
+      destination_screen_owner_slug: slug,
+      module_height: moduleHeight,
```
eessex, line 153: It looks like we are missing one property here, `horizontalSlidePosition ` -
https://cohesion.artsy.net/interfaces/_events_tap_tappedentitygroup_.tappedentitygroupargs.html#horizontalslideposition

src/lib/Scenes/Home/homeAnalytics.ts
```diff
@@ -0,0 +1,220 @@
...
+  static tapEvent({
+    action,
+    id,
+    slug,
+    moduleHeight,
+    destinationScreenOwnerType,
+    contextModule,
+    eventType,
+  }: HomeEventData): TappedEntityGroup {
+    return {
```
eessex, line 146: We should use the `tappedEntityGroup` here, as we worked on pairing yesterday, to ensure that types are correct for the context. Where did it go?
https://cohesion.artsy.net/modules/_events_tap_tappedentitygroup_.html#tappedentitygroup

brainbicycle: I got rid of it while I was figuring out why my types were disagreeing with me 😬 added back now, I have a PR in cohesion to add some missing home screen destinations: https://github.com/artsy/cohesion/pull/36

src/lib/Containers/WorksForYou.tsx
```diff
@@ -44,7 +45,7 @@ export class WorksForYou extends React.Component<Props, State> {
       } else {
         postEvent({
           name: "Notifications read",
-          source_screen: "works for you page",
+          source_screen: Schema.PageNames.WorksForYou,
```
brainbicycle, line 48: should update to use cohesion schema instead

src/lib/Components/Home/ArtistRails/ArtistRail.tsx
```diff
@@ -147,13 +146,6 @@ const ArtistRail: React.FC<Props & RailScrollProps> = props => {
     followArtist: SuggestedArtist,
     completionHandler: (followStatus: boolean) => void
   ) => {
-    trackEvent({
-      action_name: Schema.ActionNames.HomeArtistRailFollow,
-      action_type: Schema.ActionTypes.Tap,
-      owner_id: followArtist.internalID,
-      owner_slug: followArtist.id,
-      owner_type: Schema.OwnerEntityTypes.Artist,
-    })
```
brainbicycle, line 156: 👀 I should add this back now that it was determined to not be part of home screen tappedEntity events.

brainbicycle: @mikehrom can you confirm this event is a separate schema from the rest of events in home modules?

CHANGELOG.yml
```diff
@@ -2,7 +2,7 @@ upcoming:
   version: 6.4.4
```
brainbicycle, line 2: Deliberately same version?

ashfurrow: Ah whoopsie!

ashfurrow: I'll take care of that when no one else is sending PRs – don't want to cause a roadblock with the "Update branch" requirement.

src/lib/Scenes/Home/homeAnalytics.ts
brainbicycle: Added!

src/lib/utils/track/schema.ts
brainbicycle: I added this before realizing there was a sale OwnerType in cohesion, removed in favor of that 👍

mikehrom: 👍

src/lib/Components/Home/ArtistRails/ArtistRail.tsx
mikehrom: this is an interesting case -- following an artist should definitely be its own event, but as it stands now, I think we have two versions of this: `HomeArtistRailFollow` and `ArtistFollow`. I think ideally, we would have just `ArtistFollow` with context on the follow-tap to describe where the user followed from. But I do not have this event set up in cohesion right now.

src/lib/Scenes/Home/Components/CollectionsRail.tsx
```diff
@@ -17,81 +17,87 @@ import {
...
 
 interface Props {
   collectionsModule: CollectionsRail_collectionsModule
 }
 
 type Collection = CollectionsRail_collectionsModule["results"][0]
 
-export class CollectionsRail extends Component<Props> implements RailScrollRef {
-  private listRef = createRef<FlatList<any>>()
+const CollectionsRail: React.FC<Props & RailScrollProps> = props => {
```
brainbicycle, line 30: Sorry for multiple review request @ashfurrow but figure worth another look at conversion from class to functional components for Collections, Sales, and Fairs rails

src/lib/Scenes/Home/Components/FairsRail.tsx
```diff
@@ -16,100 +17,96 @@ import {
...
+        listRef={listRef}
+        data={props.fairsModule.results}
+        renderItem={({ item: result, index }) => {
+          // Fairs are expected to always have >= 2 artworks and a hero image.
+          // We can make assumptions about this in UI layout, but should still
+          // be cautious to avoid crashes if this assumption is broken.
+          const artworkImageURLs = take(
+            concat(
+              [result?.image?.url!],
+              result?.followedArtistArtworks?.edges?.map?.(edge => edge?.node?.image?.url!)!,
```
ashfurrow, line 54: Nice work on the strictness migration – can't wait for some of these types to get cleaned up in Metaphysics so this will get even cleaner 👍

src/lib/Scenes/Home/Components/__tests__/CollectionsRail-tests.tsx
```diff
@@ -103,8 +109,32 @@ describe("CollectionsRailFragmentContainer", () => {
...
+        data: {
+          homePage: {
+            marketingCollectionsModule: collectionsModuleMock,
+          },
+        },
+      })
+    })
+    // @ts-ignore
+    first(tree.root.findAllByType(CardRailCard)).props.onPress()
+    expect(trackEvent).toHaveBeenCalledWith(HomeAnalytics.collectionThumbnailTapEvent("test-collection-one", 0))
```
ashfurrow, line 133: Nice work on these tests 👍

src/lib/Scenes/Home/Components/__tests__/SalesRail-tests.tsx
```diff
@@ -15,8 +15,14 @@ jest.mock("lib/NativeModules/SwitchBoard", () => ({
 import SwitchBoard from "lib/NativeModules/SwitchBoard"
 
 import { CardRailCard } from "lib/Components/Home/CardRailCard"
+import { SectionTitle } from "lib/Components/SectionTitle"
+import { useTracking } from "react-tracking"
+import HomeAnalytics from "../../homeAnalytics"
 import { SalesRailFragmentContainer } from "../SalesRail"
 
+const trackEvent = jest.fn()
+const mockScrollRef = jest.fn()
```
ashfurrow, line 24: This is minor, and not worth fixing before the merge, but stylistically I think it makes more sense to declare these as `let` variables and then use a `beforeEach` block to re-instantiate them each time. Because of how Jest evaluates test files, there isn't really a runtime difference – it's strictly style. So something like this:

```tsx
let trackEvent: Jest.Mock

beforeEach(() => {
  trackEvent = jest.fn()
})
```

src/lib/Scenes/Home/__tests__/homeAnalytics-tests.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    const headerTapEvent = HomeAnalytics.artworkHeaderTapEvent(artworkRail.key)
+    expect(headerTapEvent).toEqual({
+      action: Analytics.ActionType.tappedArtworkGroup,
+      context_module: Analytics.ContextModule.newWorksByArtistsYouFollowRail,
+      context_screen_owner_type: "home",
+      destination_screen_owner_type: "worksForYou",
+      module_height: "double",
+      type: "header",
+    })
```
ashfurrow, line 53: These expectations are fine as-is, but _if_ you only care about some properties, then you can get a little more succinct with [Jest partial matchers](https://codewithhugo.com/jest-array-object-match-contain/). So this expectation _could_ be:

```tsx
    expect(headerTapEvent).toEqual(expect.objectContaining({
       action: Analytics.ActionType.tappedArtworkGroup,
       context_module: Analytics.ContextModule.newWorksByArtistsYouFollowRail,
       context_screen_owner_type: "home",
       destination_screen_owner_type: "worksForYou"
     }))
```

Which lets you only match the object properties you really care about. Again, this is great as-is 👍

PR #3301, opened 2020-05-12
[dont-merge] Spike on routing/nav for sell tab

jonallured:
Hey pals, this is a spike PR that I intend to close without merging. My goal is to demonstrate the approach I'm pitching for routing and navigating for the new sell tab. I would then expect to refer back to this PR as a reference as we move through the actual implementation. Here's a little taste of what we've got going on in this PR:

<details>
<summary>recording of the flows</summary>

Kapture 2020-05-12 at 10 01 49

</details>

The comps I'm working off of are here:

https://www.figma.com/file/yDSHlbJVzBxvSUROD2bfRI/Sell-tab-new-file?node-id=1%3A17

## Routes, modules and folder structure

An important concern for me is keeping myself organized with all the moving pieces here. Sure, I may change my mind as things evolve but it's best to think more widely about how one will structure a line of work so that's why I made this spreadsheet:

https://docs.google.com/spreadsheets/d/10lHDjCsZ1YN8NWiJhUhcAiGvjI7qkuFi7f8Vlb_HYKo/edit?usp=sharing

The idea was to write down the routes, view controllers, modules and then component file paths so that doing the paperwork of making all of these things is easier and more organized. If you take a look at that sheet you'll see that the routes are "RESTful" and I think that's handy for this line of work. Something we have to manage is how the artwork form will behave so now the call sites can convey this by picking which route they are linking to. 😎 

## Making a scene

One thing that might jump out at people while reviewing this PR is the number of scenes that are added. My logic is like this:

* src/lib/Scenes/SellDashboard - this is where we make the choice about whether to show you the blank slate view or the actual dashboard. I expect this to evolve over time and we may want to show things that are not directly related to one's "My Collection" collection, so I chose the "Sell Dashboard" name because it feels more accurate and more flexible.

* src/lib/Scenes/MyCollection - this folder houses the various top-level components that control how the artwork form will behave. I expect this folder to pick up more views too later, things like the list of works in one's "My Collection" collection, a detail view of a work, etc.

* src/lib/Scenes/ArtworkForm - this folder is where we put the building blocks of the artwork form flows. The contract between callers and the ArtworkForm will probably end up looking a little different than what's seen here, but establishing this contract is the point of this spike. Note too that while the sell tab is currently the only place using this form, I'm building it with partners in the back of my mind too. Seems like we could re-use a LOT of this if we build without assumptions about who the user is.

## Call for feedback

This PR won't be merged, but is intended to serve as a reference for the team as we work through the new sell tab so now is a great time for feedback! Any red flags for people? I'll leave it open for a bit to give people some time to comment and then close this sucker at a certain point.

### DIFF
--- a/Artsy/App/ARSwitchBoard.m
+++ b/Artsy/App/ARSwitchBoard.m
@@ -26,6 +26,10 @@
 #import "AREigenCollectionComponentViewController.h"
 #import "AREigenMapContainerViewController.h"
 
+#import <Emission/AREditArtworkFormComponentViewController.h>
+#import <Emission/ARNewArtworkFormComponentViewController.h>
+#import <Emission/ARNewSubmissionFormComponentViewController.h>
+
 #import <Emission/ARShowConsignmentsFlowViewController.h>
 #import <Emission/ARFairComponentViewController.h>
 #import <Emission/ARFairBoothComponentViewController.h>
@@ -273,6 +277,30 @@ static ARSwitchBoard *sharedInstance = nil;
 
     [self.routes addRoute:@"/privacy-request" handler:JLRouteParams {
         return [[ARPrivacyRequestComponentViewController alloc] init];
+    }];
+
+    // this route opens a modal that allows the user to add a work to their collection
+    [self.routes addRoute:@"/collections/my-collection/artworks/new" handler:JLRouteParams {
+        UIViewController *controller = [[ARNewArtworkFormComponentViewController alloc] init];
+        return [[ARNavigationController alloc] initWithRootViewController:controller];
+    }];
+
+    // this route opens a modal that allows the user to edit the given work. there are buttons for this on the list page, but also the artwork detail page
+    [self.routes addRoute:@"/collections/my-collection/artworks/:id/edit" handler:JLRouteParams {
+        UIViewController *controller = [[AREditArtworkFormComponentViewController alloc] initWithArtworkID:parameters[@"id"]];
+        return [[ARNavigationController alloc] initWithRootViewController:controller];
+    }];
+
+    // this route opens a modal that allows the user to create an artwork and a submission for that artwork all in one go
+    [self.routes addRoute:@"/collections/my-collection/artworks/new/submissions/new" handler:JLRouteParams {
+        UIViewController *controller = [[ARNewSubmissionFormComponentViewController alloc] init];
+        return [[ARNavigationController alloc] initWithRootViewController:controller];
+    }];
+
+    // this route opens a modal that allows the user to create a submission from an existing artwork
+    [self.routes addRoute:@"/collections/my-collection/artworks/:id/submissions/new" handler:JLRouteParams {
+        UIViewController *controller = [[ARNewSubmissionFormComponentViewController alloc] initWithArtworkID:parameters[@"id"]];
+        return [[ARNavigationController alloc] initWithRootViewController:controller];
     }];
 
     [self.routes addRoute:@"/consign/submission" handler:JLRouteParams {
--- a/src/lib/Scenes/ArtworkForm/Summary/index.tsx
+++ b/src/lib/Scenes/ArtworkForm/Summary/index.tsx
@@ -1,0 +1,78 @@
+import React, { useState } from "react"
+import { BorderBox, Button, Sans, Flex } from "@artsy/palette"
+import { PhotosSection } from "../PhotosSection"
+import { ArtistSection } from "../ArtistSection"
+import { MediumSection } from "../MediumSection"
+
+const photosRoute = {
+  component: PhotosSection,
+  title: "Add photos",
+}
+
+const artistRoute = {
+  component: ArtistSection,
+  title: "Artist",
+}
+
+const mediumRoute = {
+  component: MediumSection,
+  title: "Select Medium",
+}
+
+interface SummaryProps {
+  pushRoute: (route: any) => void
+}
+
+export const Summary: React.FC<SummaryProps> = props => {
+  const [photosDone, setPhotosDone] = useState(false)
+  const [artistDone, setArtistDone] = useState(false)
+  const [mediumDone, setMediumDone] = useState(false)
+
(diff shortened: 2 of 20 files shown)


Artsy/App/ARSwitchBoard.m
```diff
@@ -275,6 +279,30 @@ - (void)updateRoutes
         return [[ARPrivacyRequestComponentViewController alloc] init];
     }];
 
+    // this route opens a modal that allows the user to add a work to their collection
+    [self.routes addRoute:@"/collections/my-collection/artworks/new" handler:JLRouteParams {
+        UIViewController *controller = [[ARNewArtworkFormComponentViewController alloc] init];
+        return [[ARNavigationController alloc] initWithRootViewController:controller];
```
ashfurrow, line 285: I noticed that the recording you posted had a standard iOS navigation bar. That's totally fine! But if your designs call for _no_ bar, you can also use `ARSerifNavigationController` to hide it.

src/lib/Scenes/ArtworkForm/Summary/index.tsx
```diff
@@ -0,0 +1,78 @@
+import React, { useState } from "react"
```
ashfurrow, line 1: Really minor nitpick, and definitely something we should document, but we're trying to move away from files named `index.tsx`. Where we do have them ([example](https://github.com/artsy/eigen/blob/master/src/lib/Scenes/Artwork/index.ts)) we tend to only re-export from more descriptively-named files. We're also trying to avoid default exports, but it seems like all of Artsy is on the same page about that already 👍

src/lib/Scenes/MyCollection/EditArtworkForm/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+interface EditArtworkFormProps {
+  artworkID: string
+}
+
+export class EditArtworkForm extends React.Component<EditArtworkFormProps> {
+  render() {
+    const { artworkID } = this.props
+
+    const handleClosePress = () => {
+      SwitchBoard.dismissModalViewController(this)
```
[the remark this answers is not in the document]
ashfurrow: Yeah, we don't have a great resource for finding out how something is presented. Modal presentations are definitely the exception rather than the rule, but yeah. Ideally, and definitely when we move to Android, we'll want to move all this navigation logic from Objective-C to TypeScript anyway.

src/lib/Scenes/ArtworkForm/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+  navigator: NavigatorIOS
+}
+
+export const ArtworkForm: React.FC<ArtworkFormProps> = props => {
+  const handleClosePress = () => {
+    props.closeForm()
+  }
+
+  const pushRoute = (route: any): void => {
+    props.navigator.push(route)
```
ashfurrow, line 17: Makes sense 👍 I think handling the navigation internally to this "modal scene" as you've described and implemented here is a good approach. Ideally we'd have fewer `any` types, but this is only a proof-of-concept PR 😉

src/lib/Scenes/SellDashboard/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+import NavigatorIOS from "react-native-navigator-ios"
+import { BlankSlate } from "./Screens/BlankSlate"
+import { Dashboard } from "./Screens/Dashboard"
+import SwitchBoard from "../../NativeModules/SwitchBoard"
+
+interface SellDashboardProps {}
+
+export class SellDashboard extends React.Component<SellDashboardProps, any> {
+  render() {
+    const routeTo = (route: string): void => {
```
ashfurrow, line 12: I'm not totally sold on abstracting out the call to `SwitchBoard.presentModalViewController` here. It's a bit ugly but I like how it calls its ugliness out so we can fix it in the longterm.

src/lib/Scenes/ArtworkForm/Summary/index.tsx
jonallured: Yeah! I would def move towards index files that re-export but wanted to keep moving on this spike. Sounds like we're aligned here. 👍

Artsy/App/ARSwitchBoard.m
jonallured: Good to know, thanks!! 🙌

src/lib/Scenes/SellDashboard/index.tsx
jonallured: Yeah, part of this was me not understanding how to use `presentModalViewController` without a `this`, but I think @pepopowitz found a way:

https://github.com/artsy/eigen/blob/423651df444e6593ff98bfed5630a29edaff28c7/src/lib/Scenes/Search/SearchResult.tsx#L20

So what I'd probably end up doing this is pushing this stuff down and moving this to a functional component. Or at least that's what I'd tinker around with.

src/lib/Scenes/ArtworkForm/index.tsx
jonallured: Yeah, I guess I could just import that type, not sure why I did `any` other than just wanting to move forward, haha.

src/lib/Scenes/MyCollection/EditArtworkForm/index.tsx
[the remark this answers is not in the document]
jonallured: Yeah I'm not sure I have a strong opinion here. I guess I'd let the future dictate a refactor like this. I'm not aware of any plans to present the artwork form any other way than modally so it seems like something we could let evolve.

src/lib/Scenes/SellDashboard/index.tsx
ashfurrow: For sure. You can do a project-wide search for `useRef<` to look at other places we use it (without `any` 🙈).

jonallured:
Closing this but thanks for all the feedback friends!! 🤗

PR #3304, opened 2020-05-12
[CSGN-186] Connect sell tab/landing page/artist list to MP via relay

pepopowitz:
Addresses https://artsyproduct.atlassian.net/browse/CSGN-186

(Note: this UI is only visible to users who toggle the AROptionsMoveCityGuideEnableSales feature.)

This PR attaches the artist list on the landing page of the new sell tab to real data from MP. 

It uses one QueryRenderer for the page, with the expectation that it will eventually also load an artwork rail elsewhere on the page. 

There is a temporary placeholder when the screen loads - you can see it below as I navigate to the Sell tab.

landing-with-temp-placeholder

Work that will follow in future PRs:

- The order of the artist IDs will be randomized
- A real placeholder will be built (but the UI is going to change, so I'll do that before building a real placeholder)
- Test coverage

#trivial #skip_new_tests (future PRs!)


src/lib/Scenes/Consignments/index.tsx
```diff
@@ -5,7 +5,7 @@ import { Theme } from "@artsy/palette"
 import { ConsignmentSubmissionCategoryAggregation } from "__generated__/createConsignmentSubmissionMutation.graphql"
 import { NativeModules, ViewProperties } from "react-native"
 import NavigatorIOS from "react-native-navigator-ios"
-import { ConsignmentsHome } from "./v2/Screens/ConsignmentsHome"
+import { ConsignmentsHomeRenderer as ConsignmentsHome } from "./v2/Screens/ConsignmentsHome"
```
damassi, line 8: We should update name from `ConsignmentsHomeRenderer` to  `ConsignmentsHomeQueryRenderer` which follows wider ecosystem patterns.

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/ArtistList.tsx
```diff
@@ -21,25 +26,44 @@ export const ArtistList = () => {
...
+})
+
+const StackOfArtists: React.FC<{ artists: Array<ArtistList_artists[0]> }> = ({ artists }) => {
   return (
     <Flex>
       <Join separator={<Spacer mb={2} />}>
-        {artists.map(artist => {
-          return <ArtistItem artist={artist} key={artist.name} />
+        {artists.map((artist, index) => {
+          return <ArtistItem artist={artist} key={artist.name || index} />
```
damassi, line 50: Would the artist name ever not be present for a "top 20" artist?

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/ArtistList.tsx
```diff
@@ -54,267 +78,27 @@ const ArtistItem: React.FC<{ artist: Artist }> = ({ artist }) => {
...
-      url: string
-      width: number
-      height: number
-    }
-  }
-}
-
-interface Data {
-  artists: Artist[]
+const top20Artists = {
```
damassi, line 81: #superminor - Might be worth turning this variable name into a louder CONSTANT form and adding a comment about moving this this to MP. Also, language-wise "top 20" => `microfunnel`.

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/index.tsx
```diff
@@ -100,3 +112,35 @@ export const ConsignmentsHome = () => {
...
+
+const ConsignmentsHomeContainer = createFragmentContainer(ConsignmentsHome, {
+  artists: graphql`
+    fragment ConsignmentsHome_artists on Artist @relay(plural: true) {
+      ...ArtistList_artists
+    }
+  `,
+})
+
+export const ConsignmentsHomeRenderer: React.FC = () => {
```
damassi, line 124: `Renderer` -> `QueryRenderer`

src/lib/Scenes/Consignments/index.tsx
pepopowitz: The majority of examples I found in the eigen codebase omit `Query`. @ashfurrow can you confirm that we want to move toward including `Query` in the name?

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/ArtistList.tsx
ashfurrow: It's probably not likely but this line of code does something else important: it makes the compiler happy 😎

pepopowitz: Probably not, but the type of `name` is `string | null`, and this seems as effective a way to eliminate the TypeScript error as anything else.

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/ArtistList.tsx
pepopowitz: Good call on the name. Nicole's most recent designs call them "focused 20" - I'll change to that name before this gets merged.

I'll CAPITALIZE IT too.

src/lib/Scenes/Consignments/index.tsx
ashfurrow: Hmmmm I actually thought we were more consistent here than we are. If the wider Relay community uses the full `QueryRenderer` suffix, then we should probably adopt it. 

In any case, we should be consistent so I'll add the cleanup work to consolidate them all to the MX backlog – sounds like a nice Friday afternoon task.

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/ArtistList.tsx
ashfurrow: Good point. I'd prefer not to hardcode these when we ship to users, but I'd be fine with leaving a TODO comment for now. 

Metaphysics should be the source of truth on these – maybe instead of using the Metaphysics top-level `artists(ids: )` resolver, we could add a new `microfunnelArtists` resolver? Or maybe, add an arg to the existing one: `artists(microfunnel: true)`. Changing these IDs through an env var on Metaphysics would be a lot easier than releasing an app update 👍

pepopowitz: Good suggestions Ash - I will make a subtask in the related story to make sure we move these hardcoded values to MP before this feature launches to real users.

src/lib/Scenes/Consignments/index.tsx
pepopowitz: I'm actually not clear on what you're recommending @ashfurrow. Should I roll with `ConsignmentsHomeRenderer` because it's consistent with the rest of the code _currently_,  or `ConsignmentsHomeQueryRenderer` because will _eventually_ align with the rest of the code?

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/ArtistList.tsx
```diff
@@ -21,25 +26,44 @@ export const ArtistList = () => {
...
+        {artists.map((artist, index) => {
+          return <ArtistItem artist={artist} key={artist.name || index} />
         })}
       </Join>
     </Flex>
   )
 }
 
-const ArtistItem: React.FC<{ artist: Artist }> = ({ artist }) => {
+const ArtistItem: React.FC<{ artist: ArtistList_artists[0] }> = ({ artist }) => {
```
jonallured, line 57: Trying to follow along here, this stuck out - so we're specifying a type here, but the type in question is a member of an array type? So I think the way to read this is that we have a `ArtistList_artists` type and when we call `[0]` on it, we get back the type of the members of that type? There's no name for that type because it's generated, so that's why we have to reference it like this?

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/ArtistList.tsx
```diff
@@ -21,25 +26,44 @@ export const ArtistList = () => {
...
   const navRef = useRef<any>()
   const imageUrl = artist.image?.cropped?.url
-  const { width, height } = artist.image?.cropped
+  const width = artist.image?.cropped?.width || 76
+  const height = artist.image?.cropped?.height || 70
```
jonallured, line 61: Maybe worth extracting locals here?

```
const fallbackWidth = 76
const width = artist.image?.cropped?.width || fallbackWidth
```

What do you think?

src/lib/Scenes/Consignments/index.tsx
ashfurrow: Oh sorry, I should have been specific. I'm actually okay with either of these approaches because we'll clean it all up soon. We will be going with the full `ConsignmentsHomeQueryRenderer` eventually, so we can make that change now (or you can leave it for us later). So specifically, either works 🤗

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/ArtistList.tsx
ashfurrow: That's right – it's a weird bit of TypeScript syntax at first but once you get used to it it's great 🎉

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/ArtistList.tsx
pepopowitz: makes sense but this width will be gone in a followup, where I switch this over to use `EntityHeader`. I'll leave it as-is.

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/index.tsx
pepopowitz: Done!

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/ArtistList.tsx
pepopowitz: Done!

src/lib/Scenes/Consignments/index.tsx
pepopowitz: I renamed it to `ConsignmentsHomeQueryRenderer`.

jonallured:
LGTM!! 🎉 #mergeongreen

PR #3305, opened 2020-05-12
Countdown refactor

mdole:
@pvinis and I took a look at countdown clocks in Eigen today and found ourselves a bit overwhelmed - the countdown structure is pretty wild. We eventually realized that what we wanted for Viewing Rooms was going to look almost exactly the same as the `Fairs` countdown, so we said "well, let's generalize it a little bit and use it for both." This PR implements that refactor.

**Do we foresee using a similar countdown clock in enough places for this to be worth generalizing**? Auction screens already make use of a [very similar timer](https://github.com/artsy/eigen/blob/master/src/lib/Components/Bidding/Components/Timer.tsx), but it diverges enough from this Fairs/VR version that I don't think it's worth trying to make it fit so I left it separate.

Input is appreciated! If folks feel like this refactor makes sense, we can merge and proceed, and if not I can simply copy/paste most of the code from the `Fair` countdown for `ViewingRoom`

Edit: gifs + screenshots!

Fair opening in the future: https://share.getcloudapp.com/P8uEDv6v
Fair open now: https://share.getcloudapp.com/RBuvz6Le
Fair already closed: https://share.getcloudapp.com/9ZuELqQ5

#trivial


ashleyjelks:
This looks great! Can you post gifs of the countdown in the places the refactored component is used? Always nice to have visual confirmation that a refactor still performs as expected!

mdole:
ah smart thinking, thanks @ashleyjelks! just added those to the PR description

PR #3309, opened 2020-05-13
Fix for fair booth display with null profile

brainbicycle:
This fixes a bug that was causing fair booths not to load when a partner profile was empty.

Reported in slack: https://artsy.slack.com/archives/CN8S32FJR/p1589375619280400

Co-authored with @ashfurrow

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -1,10 +1,10 @@
 upcoming:
-  version: 6.4.4
+  version: 6.4.5
   date: TBD
   dev:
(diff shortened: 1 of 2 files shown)


CHANGELOG.yml
```diff
@@ -1,10 +1,10 @@
 upcoming:
-  version: 6.4.4
+  version: 6.4.5
```
ashfurrow, line 2: Thank you! 🙌

PR #3312, opened 2020-05-13
[CSGN-188] Update sell landing page layout according to new specs

pepopowitz:
Addresses https://artsyproduct.atlassian.net/browse/CSGN-188

@deliciousnachos gave us a 🔥redesign of the sell tab landing page. This PR implements it.

sell-tab-redesign

#trivial #skip_new_tests (coming in followup work)


src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/index.tsx
```diff
@@ -18,101 +19,201 @@ interface Props {
...
-          </Sans>
-        </Box>
+        <Spacer mb={2} />
+
+        <Flex flexDirection="row">
+          <NumberBox pl={0.5} pr={1}>
+            <Sans size="4">2</Sans>
+          </NumberBox>
+
+          <FlexChildThatWontStretchOutsideOfParent>
```
damassi, line 172: 😄 - maybe something a bit more semantic for future devs?

ashfurrow: As long as it is _as_ verbose – we need to honour Eigen's Objective-C roots!

damassi: heh, now that you mention it, this is indeed very Objective-C

pepopowitz: I think this captures everything I wanted to capture for future devs - these styles are applied to prevent this element from stretching outside of the parent flex element.

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/index.tsx
```diff
@@ -18,101 +19,201 @@ interface Props {
...
-          <Sans size="3">Start today</Sans>
+          <Sans size="3">Start selling</Sans>
         </Button>
       </Box>
     </ScrollView>
   )
 }
 
+const FlexChildThatWontStretchOutsideOfParent = styled(Box)`
+  flex: 1;
```
jonallured, line 208: I would not have known that setting flex to a value of one means that it won't stretch outside of it's parent, haha. Is there some doc you share that explains this?

jonallured:
Nice one! Looks like it needs a rebase but otherwise I'm +1 on it! ❤️

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/index.tsx
damassi: I guess i'm wondering why this level of detail in a variable name is necessary, when all that's being applied is `flex: 1`, a few lines down? Versus, say, `SomethingContainer` - a thing which contains.

damassi: (note that ^ is obviously non-blocking, just for the sake of discussion.)

pepopowitz: - "why this level of detail in a variable name is necessary": The text in these elements was overflowing the parent flex element, but subtly, by only a few pixels. It was easy to miss the fact that something wasn't right, and I had to dig pretty deep to (a) prove to myself that it was overflowing, and (b) figure out how to prevent it from overflowing. By giving this variable a name, and a verbose name at that, it makes it clear to future devs that this is being done _for a reason_. 
- "This flex: 1 can also be applied inline": this is not true, unless I'm misunderstanding what you're saying. As I mentioned in [this slack thread](https://artsy.slack.com/archives/C9XJKPY9W/p1589377765113900), there are a handful of props we applied to `Flex` that should also be applied to elements that can be placed inside of a `flex` container. `flex` shorthand, `flexBasis`, `flexGrow`, and `flexShrink` are all properties that get set on the _child_ of a flex element, not the flex element itself. In this case, I've got a `Box`, which doesn't have `flexProps` applied to it in palette, so `<Box flex="1">` is not valid.

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/index.tsx
pepopowitz: I didn't know either. I also didn't know how to fix it. [This](https://stackoverflow.com/a/42537091) was I think the lead that pointed me in the right direction.

The issue I was seeing was very subtle. On an iphone 11 pro max, it almost didn't look like an issue, but I spotted a couple pixels of misalignment. I played with some other devices and text lengths to verify, and found that those elements were indeed overflowing beyond the parent.

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/index.tsx
damassi: 🏄 no prob, just discussion @pepopowitz 😄

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/index.tsx
ashfurrow: Huh! TIL.

PR #3316, opened 2020-05-14
Extract ArtworkTileRailCard for reuse with other shapes of data

pepopowitz:
I will be rendering an artwork rail on the sell tab landing page, and I want it to look like the other artwork rails in the app. The shape of my data is very different from what `ArtworkTileRail` uses, though, so this PR extracts an `ArtworkTileRailCard` that takes things like `artistNames` and `saleMessage` as props. This leaves it up to the parent component to decide how to get those bits of data.

I chose to do this at the card level instead of the rail level because there might be subtle differences between my upcoming rail and the other rails, specifically in the spacing between cards. Perhaps it's confirmation bias, but it also feels nice that the parent can decide if it's rendering a `FlatList`, an `AboveTheFoldFlatList`, or a `Join` to render this rail of cards.

The PR also includes a small fix with border-radius - moving it from the rail level to the card level. I'll point it out with a comment.

### DIFF
--- a/src/lib/Components/ArtworkTileRail/ArtworkTileRail.tsx
+++ b/src/lib/Components/ArtworkTileRail/ArtworkTileRail.tsx
@@ -1,4 +1,4 @@
-import { Box, color, Flex, Sans, Spacer } from "@artsy/palette"
+import { Spacer } from "@artsy/palette"
 import { ArtworkTileRail_artworksConnection } from "__generated__/ArtworkTileRail_artworksConnection.graphql"
 import SwitchBoard from "lib/NativeModules/SwitchBoard"
 import { Schema } from "lib/utils/track"
@@ -6,12 +6,7 @@ import React, { useRef } from "react"
 import { FlatList, View } from "react-native"
 import { createFragmentContainer, graphql } from "react-relay"
 import { useTracking } from "react-tracking"
-import styled from "styled-components/native"
-import OpaqueImageView from "./OpaqueImageView/OpaqueImageView"
-
-const SMALL_TILE_IMAGE_SIZE = 120
-
-export const ArtworkCard = styled.TouchableHighlight.attrs({ underlayColor: color("white100"), activeOpacity: 0.8 })``
+import { ArtworkTileRailCard } from "./ArtworkTileRailCard"
 
 export const ArtworkTileRailContainer: React.FC<{
   artworksConnection: ArtworkTileRail_artworksConnection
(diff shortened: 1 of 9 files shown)


src/lib/Components/ArtworkTileRail/ArtworkTileRail.tsx
```diff
@@ -25,35 +20,21 @@ export const ArtworkTileRailContainer: React.FC<{
     <View ref={navRef}>
       <FlatList
         horizontal
-        style={{ borderRadius: 2, overflow: "hidden" }}
```
pepopowitz, line 28: Here's the first half of the border-radius fix. It was accidentally applied at the rail level, instead of the card.

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>Missing Test Files:

- `src/lib/Components/ArtworkTileRail/__tests__/ArtworkTileRailCard-tests.tsx`
- `src/lib/Components/ArtworkTileRail/__tests__/index-tests.ts`

    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/Components/ArtworkTileRail/ArtworkTileRailCard.tsx
```diff
@@ -0,0 +1,63 @@
...
+  imageURL,
+  artistNames,
+  saleMessage,
+}) => {
+  const imageDisplay = imageURL ? (
+    <OpaqueImageView
+      imageURL={imageURL.replace(":version", "square")}
+      width={SMALL_TILE_IMAGE_SIZE}
+      height={SMALL_TILE_IMAGE_SIZE}
+      style={{ borderRadius: 2, overflow: "hidden" }}
```
pepopowitz, line 29: And here's the second half of the border-radius fix. It was accidentally applied at the rail level, instead of the card.

src/lib/Components/ArtworkTileRail/ArtworkTileRail.tsx
```diff
@@ -25,35 +20,21 @@ export const ArtworkTileRailContainer: React.FC<{
...
-                <Sans size="3t" weight="medium" numberOfLines={1}>
-                  {item?.node?.artistNames}
-                </Sans>
-                <Sans size="3t" color="black60" numberOfLines={1}>
-                  {item?.node?.saleMessage}
-                </Sans>
-              </Box>
-            </Flex>
-          </ArtworkCard>
+            imageURL={item?.node?.image?.imageURL ?? ""}
```
ashfurrow, line 34: `imageURL` is nullable so I think we can remove the null coalesce:
```suggestion
            imageURL={item?.node?.image?.imageURL}
```

src/lib/Components/ArtworkTileRail/ArtworkTileRailCard.tsx
```diff
@@ -0,0 +1,63 @@
...
+      <Flex>
+        {imageDisplay}
+        <Box mt={1} width={SMALL_TILE_IMAGE_SIZE}>
+          {artistNamesDisplay}
+          {saleMessageDisplay}
+        </Box>
+      </Flex>
+    </ArtworkCard>
+  )
+}
```
ashfurrow, line 63: I like the structure of this `render` method quite a bit – I think I'll try moving more of my `thing && <Component>{thing}</Component>` into consts instead of inlining them 👍

src/lib/Components/ArtworkTileRail/ArtworkTileRail.tsx
pepopowitz: Good call, I'll make this change!

pepopowitz:
@mdole I think this one is all cleaned up. There is a complaint from danger about no tests for an index.ts, but that file just exports other files. Maybe someone from @artsy/mobile-experience has opinions on if I should handle that differently, but for me it feels safe to ignore.

mdole:
heck yeah! onwards to staging!

PR #3322, opened 2020-05-19
Make viewing rooms dynamic [GALL-2794]

mdole:
This PR moves us from a static "test-viewing-room-id" slug to dynamic navigation.

For example, you can now open artsy:///viewing-room/david-zwirner-private-party-selected-works-by-nicolas-party in a simulator Safari and be taken to a VR. This PR also lays the groundwork for displaying VRs on the partner screen (which may or may not be used for a while since we're unclear how we're gating access to/from rooms). Still, should be handy for testing since you can go to a gallery page and tap through to a VR, assuming you have the feature flag.

@starsirius while working on this, I discovered that querying any partner for `vieiwngRooms` returns ALL viewing rooms in existence, regardless of whether or not they're associated with that particular partner. I'm wondering if you might have some insight since you've dealt with similar issues recently. Maybe we can pair on it tomorrow?

Screenshot example of the issue: 
Screen Shot 2020-05-18 at 10 25 27 PM

[Jira ticket](https://artsyproduct.atlassian.net/browse/GALL-2794)

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -4,6 +4,7 @@ upcoming:
   dev:
     - Add tracking for homescreen modules - brian
     - Make viewing room header full-bleed and incorporate countdown and partner info - mdole
+    - Make viewing room navigation dynamic + link to them on partner shows page - mdole
     - Extract common artwork tile independent of the shape of the data source - pepopowitz
   user_facing:
     - Fix for bug with fair booth display when partner profile was empty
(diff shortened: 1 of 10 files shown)


starsirius:
@mdole Nice catch! I think the arguments passed to Gravity schema should be `partnerId` rather than [`partner_id`](https://github.com/artsy/metaphysics/blob/1adbbbc454cddabe86985cb9dfc840478f8ea59b/src/lib/stitching/gravity/stitching.ts#L142). Will verify and fix it!

PR #3324, opened 2020-05-19
[CSGN-189] Placeholder for Sell Tab landing page

pepopowitz:
Addresses https://artsyproduct.atlassian.net/browse/CSGN-189

Adds a proper placeholder screen for the Sell Tab landing page.

Static content remains static, and only the dynamic pieces are grayed out. Here's what it looks like - note the very bottom, where the "Artists collectors are looking to buy" section fills in:

placeholder-full

To give a better idea of what that artist list would look like if it were to, say, move up the page some day, here's how that artist section loads when you can see it completely:

placeholder-artist-list

There is a slight difference in heights that causes the section below it to jump a couple pixels...I was unable to figure out what was causing that difference. I don't personally think it's a big deal because no one really sees the bottom of the artist section when the full view is loading, but let me know if you feel otherwise.

#trivial


jonallured:
LGTM 🏆

PR #3325, opened 2020-05-19
[AUCT-1011] - Add confirmation alert for email verification in homeVC

brainbicycle:
https://artsyproduct.atlassian.net/browse/AUCT-1011

This is an initial attempt to show a confirmation dialog when users tap an email verification link in Eigen @ds300 wanted to get your feedback on the approach and if this is something we want to handle more generally for teams that need to show a modal or banner in Eigen from a deep link.

Also the preferred UX here is to show a sash on bottom of feed wondering if you had any insight into difficulty there: 
https://www.figma.com/file/dymIntVYc075QVdlWR99UQ/Email-Verification?node-id=113%3A161

Co-authored with @bhoggard @erikdstock 

Edit: This is agreed upon approach and ready for review. 

See copy for additional states here: 
https://github.com/artsy/eigen/pull/3325/files#diff-2996077a298cdc9e41e9bf32fc2431d1R46

**Error state**:
error-state

**Success State**:
email-confirmed


ds300:
> if this is something we want to handle more generally for teams that need to show a modal or banner in Eigen from a deep link

Yeah that would be nice to have. Tricky problem though, given that modals can be handled both by switchboard and by react-native (e.g. collections filters modal). 

> Also the preferred UX here is to show a sash on bottom of feed wondering if you had any insight into difficulty there

Yeah we'd need to find a way of pushing the sash state into react-native land, most likely via a prop. There's plenty of precedent for this (e.g. artist id on artist page). Should be simple technically. The UX implications are much trickier though, given that there's already a permanent sash for consignments there. e.g. people might be used to ignoring it and not notice the confirmation/error message, how would interactions of showing/dismissing the sash work, etc.

This code looks good to me though!

dleve123:
@erikdstock @bhoggard @brainbicycle Can we add some screenshots to see what this looks like?

Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
```diff
@@ -248,8 +248,19 @@ - (void)registerWithSwitchBoard:(ARSwitchBoard *)switchboard
...
         [switchboard registerPathCallbackAtPath:menuToPaths[tabNum] callback:^id _Nullable(NSDictionary *_Nullable parameters) {
+
+            NSString *confirmationMessageCode = parameters[@"flashMessage"];
+
             ARTopTabControllerTabType tabType = [tabNum integerValue];
-            return [self rootNavigationControllerAtTab:tabType].rootViewController;
+            switch (tabType) {
+                case ARHomeTab:
+                    if (confirmationMessageCode != nil) {
+                        return [self homeWithEmailConfirmation:confirmationMessageCode];
```
erikdstock, line 258: Given that we’ve made this more general to any user message we want to present rather than specifically email should we make a similar naming change here?

emission/Pod/Classes/ViewControllers/ARHomeComponentViewController.m
```diff
@@ -14,4 +15,58 @@ - (instancetype)initWithEmission:(nullable AREmission*)emission;
...
+                                          message:message
+                                          preferredStyle:UIAlertControllerStyleAlert];
+    UIAlertAction *okAction = [UIAlertAction actionWithTitle:@"OK" style:UIAlertActionStyleDefault handler:^(UIAlertAction *_Nonnull action) {
+        [self dismissViewControllerAnimated:YES completion:nil];
+    }];
+    [alert addAction:okAction];
+    [self presentViewController:alert animated:YES completion:nil];
+}
+
+#pragma mark - Email Confirmation
```
erikdstock, line 31: Curious what these pragmas do

brainbicycle: They are just for code organization, adding them adds it as section in Xcode's source navigator which makes it easier to jump to, but honestly I barely ever use that feature and just use it to mark different blocks of responsibilities. 
https://nshipster.com/pragma/#organizing-your-code

Swift has similar with `// MARK` comments
https://nshipster.com/swift-documentation/#mark-todo-fixme

Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
brainbicycle: I can definitely make the naming more general but I do think that if we find we are using this for a lot of different user messages we should probably rethink the UX of showing a standard iOS alert.

brainbicycle:
> First off i really enjoyed working on this and felt like i learned a lot. My main question is- if we are adding a new interface to both force and eigen, basically a key in the url `flashMessage` that might be handled with a sash or alert or something else, should we tale care to make values we pass specific, eg confirmationMessage: confirmed becomes flashMessage: email_confirmed? My understanding is that these details are harder to iterate on later in native code.''

👍 

I think it is a good idea to be more specific here with the message codes and perhaps prefix with *email_* . That said although we are making the query param naming more generic the code as it stands still is tied to the pretty narrow use case of whatever message codes we have hardcoded in the app and any new messages with this approach will require a new app release. If our only current use case of this is email confirmation messages I would err on keeping it simple for the time being and revisiting if we have other use cases.

emission/Pod/Classes/ViewControllers/ARHomeComponentViewController.m
```diff
@@ -14,4 +15,58 @@ - (instancetype)initWithEmission:(nullable AREmission*)emission;
...
+           @"title": @"Email Confirmed",
+           @"message": @"Your email has been confirmed."
+       },
+       @"already_confirmed": @{
+           @"title": @"Already Confirmed",
+           @"message": @"You have already confirmed your email."
+       },
+       @"invalid_token": @{
+           @"title": @"Error",
+           @"message": @"Your token is invalid. Please try again."
```
dleve123, line 58: An error has occurred. Please contact support@artsy.net

emission/Pod/Classes/ViewControllers/ARHomeComponentViewController.m
```diff
@@ -14,4 +15,58 @@ - (instancetype)initWithEmission:(nullable AREmission*)emission;
...
+       },
+       @"blank_token": @{
+           @"title": @"Error",
+           @"message": @"No token found. Please try again."
```
dleve123, line 62: An error has occurred. Please contact support@artsy.net

emission/Pod/Classes/ViewControllers/ARHomeComponentViewController.m
```diff
@@ -14,4 +15,58 @@ - (instancetype)initWithEmission:(nullable AREmission*)emission;
...
+       },
+       @"expired_token": @{
+           @"title": @"Error",
+           @"message": @"An error has occurred. Please try again."
```
dleve123, line 66: Link expired. Please request a new verification email below.

emission/Pod/Classes/ViewControllers/ARHomeComponentViewController.m
```diff
@@ -14,4 +15,58 @@ - (instancetype)initWithEmission:(nullable AREmission*)emission;
...
+       @"invalid_token": @{
+           @"title": @"Error",
+           @"message": @"Your token is invalid. Please try again."
+       },
+       @"blank_token": @{
+           @"title": @"Error",
+           @"message": @"No token found. Please try again."
+       },
+       @"expired_token": @{
+           @"title": @"Error",
```
dleve123, line 65: Link Expired

dleve123:
hey @brainbicycle – we just took a look at the text for the error cases here. Would it be possible for you to tweak the working a bit inline with the comments I just left?

PR #3330, opened 2020-05-20
Move npm-vendored podspecs to podfile

ds300:
Not sure if you remember us creating this npm-podspecs file @brainbicycle but we don't need it anymore since we combined eigen and emission. To avoid confusion let's get rid of it and keep all our pod dependency declarations in the podfile.

#trivial


Podfile.lock
```diff
@@ -858,6 +858,6 @@ SPEC CHECKSUMS:
   "XCTest+OHHTTPStubSuiteCleanUp": 4469ec8863c6bc022c5089a9b94233eb3416c5ee
   Yoga: 50fb6eb13d2152e7363293ff603385db380815b1
 
-PODFILE CHECKSUM: 1db98f52ac11bd0b220f43b4666d25105c259932
+PODFILE CHECKSUM: 5a1d55bcad5c6ad066e297a54e38657b2f5396c2
```
ds300, line 861: weird that this changed but nothing else in this file changed.

PR #3339, opened 2020-05-20
Add artwork description to Viewing Room artworks view

mdole:
Missed this in the spec till now! This PR grabs the artwork description if one exists and puts it below the artwork in the list view.

Screen Shot 2020-05-20 at 5 32 00 PM


oxaudo:
@mdole - It's super confusing, but I'm 99.9% sure that you want `additionalInformation` here instead of `description`. Description is an admin only `blurb` field. `additional information` is entered by partner.

Sorry for the late note here. Just now catching up on the emails after Friday:)

mdole:
Ooo good catch @oxaudo, thank you! I'll confirm + make that change today

PR #3341, opened 2020-05-20
Add Featured Collections Rail to Collection Departments

ashleyjelks:
Kapture 2020-05-22 at 15 31 20

Kapture 2020-05-22 at 15 33 07

Kapture 2020-05-20 at 19 15 42

https://artsyproduct.atlassian.net/browse/FX-1798


src/lib/Scenes/Collection/Components/CollectionHubsRails/FeaturedCollections/__tests__/FeaturedCollections-tests.tsx
```diff
@@ -0,0 +1,334 @@
...
+      destination_screen_owner_id: "featured-collection-id-1",
+      destination_screen_owner_slug: "featured-collection-1",
+      destination_screen_owner_type: "Collection",
+      horizontal_slide_position: 1,
+      type: "thumbnail",
+    })
+  })
+
+  describe("Featured Collections Rail", () => {
+    let props: any /* STRICTNESS_MIGRATION */
```
ashleyjelks, line 117: This will cause CI to fail.

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>Missing Test Files:

- `src/lib/Scenes/Collection/Components/CollectionHubsRails/FeaturedCollections/__tests__/FeaturedCollectionsRail-tests.tsx`

    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/Scenes/Collection/Components/CollectionHubsRails/ArtistSeries/CollectionArtistSeriesRail.tsx
```diff
@@ -59,15 +59,6 @@ export const CollectionArtistSeriesRailContainer = createFragmentContainer(Colle
             }
           }
         }
-        defaultHeader: artworksConnection(sort: "-decayed_merch", first: 1) {
-          edges {
```
ashleyjelks, line 63: This isn't being used as there is no default header in iOS Collections--we always show the top most merchandisable artwork in the collection as the header.

src/lib/Scenes/Collection/Components/CollectionHubsRails/FeaturedCollections/__tests__/FeaturedCollectionsRail-tests.tsx
```diff
@@ -0,0 +1,334 @@
...
+
+      expect(
+        wrapper
+          .find(FeaturedCollectionPrice)
+          .at(2)
+          .text()
+      ).toBe("From $35,000")
+    })
+
+    xit("navigates to a new collection when tapped", () => {
```
ashleyjelks, line 218: This test is failing for an odd reason, I'm probably missing something very obvious, will see if someone on FX wants to pair on this!

src/__generated__/ViewingRoomHeader_viewingRoom.graphql.ts
```diff
@@ -5,9 +5,9 @@ import { ReaderFragment } from "relay-runtime";
 import { FragmentRefs } from "relay-runtime";
 export type ViewingRoomHeader_viewingRoom = {
     readonly title: string;
-    readonly startAt: unknown | null;
-    readonly endAt: unknown | null;
-    readonly heroImageURL: string | null;
```
anandaroop, line 10: It's ok that this unrelated diff is in here?

src/lib/Components/Markdown.tsx
```diff
@@ -32,7 +32,17 @@ export class Markdown extends React.Component<Props & FlexProps> {
...
+    const children = stringifyChildren(this.props.children)
 
-    return <Flex {...rest}>{renderMarkdown(stringifyChildren(this.props.children), rules)}</Flex>
+    if (rules?.truncationLimit) {
+      // an end limit of undefined on the slice method will return the full string
+      const truncationLimit: number | undefined =
+        children.length <= rules.truncationLimit ? undefined : rules.truncationLimit
+      const ellipsis: string = !truncationLimit ? "" : "..."
+
+      return <Flex {...rest}>{renderMarkdown(children.slice(0, truncationLimit) + ellipsis, rules)}</Flex>
```
ds300, line 43: This is not a reliable way to truncate markdown. e.g. the markdown `Hello *Banksy*` truncated at 9 chars would render as 'Hello *Ba...' instead of 'Hello *Ban*...'

We have some sophisticated code for doing proper markdown truncation in `ReadMore`, and I'd prefer if we could re-use that here.

But alternatively, looking at the use case in the collections rail, I'd suggest undoing the changes to this component and then using `renderMarkdown` directly in the rail cards. You can grab the first `Sans` component that it outputs, and then add the `numberOfLines` prop to let react-native take care of the truncation and ellipsis.

src/lib/Scenes/Collection/Components/CollectionHubsRails/FeaturedCollections/FeaturedCollectionsRail.tsx
```diff
@@ -0,0 +1,148 @@
...
+  const { collection, collectionGroup } = props
+  const collections = collectionGroup?.members ?? []
+  const basicRules = defaultRules(true)
+  const markdownRules = {
+    ...basicRules,
+    truncationLimit: 100,
+    paragraph: {
+      ...basicRules.paragraph,
+      // @ts-ignore STRICTNESS_MIGRATION
+      react: (node, output, state) => (
```
ds300, line 35: Let's avoid introducing new strictness error! :)

The problem here is that these parameters are untyped. We can let typescript infer them if we define these rules in a context with type information. The `defaultRules` helper has been set up for that:

```
const markdownRules = defaultRules(true, {
  paragraph: {
    ...basicRules.paragraph,
    react: (node, output, state) => ...
  }
})
```

src/lib/Scenes/Collection/Components/CollectionHubsRails/FeaturedCollections/__tests__/FeaturedCollectionsRail-tests.tsx
```diff
@@ -0,0 +1,334 @@
+import { Theme } from "@artsy/palette"
+import {
+  FeaturedCollectionsTestsQuery,
+  FeaturedCollectionsTestsQueryRawResponse,
+} from "__generated__/FeaturedCollectionsTestsQuery.graphql"
+// @ts-ignore STRICTNESS_MIGRATION
+import { mount } from "enzyme"
```
ds300, line 7: We're moving away from `enzyme` and towards `react-test-renderer` and `relay-test-utils`. Please avoid creating new tests with enzyme from now on.

src/lib/Scenes/Collection/Components/CollectionHubsRails/FeaturedCollections/FeaturedCollectionsRail.tsx
```diff
@@ -0,0 +1,148 @@
...
+export const ImageWrapper = styled(Flex)`
+  border: solid 1px ${color("black10")};
+  height: 385px;
+  width: 260px;
+  border-radius: 5px;
+`
+
+export const FeaturedCollectionTitle = styled(Sans)``
+export const FeaturedCollectionPrice = styled(Sans)``
+export const CollectionGroup = styled(Sans)``
```
ds300, line 117: nonblocking: this is an odd pattern. I think these are only being exported to help find the components in tests, right? Normally we'd use the `data-test-id` prop for that, which is a bit more explicit and conventional.

src/lib/Scenes/Collection/Components/CollectionHubsRails/FeaturedCollections/FeaturedCollectionsRail.tsx
```diff
@@ -0,0 +1,148 @@
...
+import ImageView from "lib/Components/OpaqueImageView/OpaqueImageView"
+import SwitchBoard from "lib/NativeModules/SwitchBoard"
+import { defaultRules } from "lib/utils/renderMarkdown"
+import { Schema } from "lib/utils/track"
+import React, { useRef } from "react"
+import { FlatList, TouchableHighlight } from "react-native"
+import { createFragmentContainer, graphql } from "react-relay"
+import { useTracking } from "react-tracking"
+// @ts-ignore
+import styled from "styled-components/native"
```
ds300, line 14: this is not required any more, we have typings for styled-components now

src/lib/Scenes/Collection/Components/CollectionHubsRails/FeaturedCollections/FeaturedCollectionsRail.tsx
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  return collections.length > 0 ? (
+    <>
+      <Flex ml={"-20px"} ref={navRef}>
+        <CollectionGroup size="4" my={2} ml={4}>
+          {collectionGroup.name}
+        </CollectionGroup>
+      </Flex>
+      <FlatList<FeaturedCollection>
```
ds300, line 54: It'd be cool if you could use an `AboveTheFoldFlatList` here, which more aggressively prevents off-screen things from being rendered until the user starts scrolling. This helps us prevent loading too many images at once when a user lands on a page.

src/lib/Components/Markdown.tsx
ashleyjelks: Nice suggestion, this works perfectly!

src/lib/Scenes/Collection/Components/CollectionHubsRails/FeaturedCollections/FeaturedCollectionsRail.tsx
ashleyjelks: Makes sense. I'd like to do this in a follow up PR because there are now 3 horizontally scrolling flatlists on department pages that I think can be similarly optimized!

ashleyjelks: Added a ticket to track this: https://artsyproduct.atlassian.net/browse/FX-1970

src/lib/Scenes/Collection/Components/CollectionHubsRails/FeaturedCollections/__tests__/FeaturedCollectionsRail-tests.tsx
```diff
@@ -0,0 +1,262 @@
...
+    })
+
+    it("renders each Featured Collection's title", () => {
+      const tree = ReactTestRenderer.create(
+        <Theme>
+          <FeaturedCollectionsRail {...props} />
+        </Theme>
+      ).root
+
+      const title1 = tree.findByProps({ "data-test-id": "title-0" })
```
ashleyjelks, line 146: I constructed these ReactTestRenderer instances like this because calling `findAllByProps({data-test-id: "title"})` returned more than 3 instances.  I saw others were having this issue as well https://github.com/artsy/eigen/pull/3341.

What's interesting was that when I used `findByProps({data-test-id: "title"})`, it throws an error (as expected) that it found 3 instances instead of 1. So it is unclear to me if this is some bug (finding more than 3 instances when using `findAllByProps({data-test-id: "title"})`) in the react-test-renderer library like others note in that Github issue, or something else.

PR #3350, opened 2020-05-21
Add more tracking for consign screen

jonallured:
This PR adds tracking to the consign button presses and then also for artist taps from the consign landing screen. I used cohesion, especially the helper from https://github.com/artsy/cohesion/pull/46 so I think this all lines up to the Schema that data has created for us. One thing to note is that I _think_ we should update the tracking on the recently sold to use cohesion too, right @damassi? I'm thinking about this part in particular:

https://github.com/artsy/eigen/blob/049b868bf4b1c56c444bac7aa1c8b11f6446243d/src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/Components/RecentlySold.tsx#L50-L59

Unless I'm missing something I believe that code is using Eigen's schema rather than Cohesion's version?


jonallured:
Oops, forgot an important note: this one is blocked by exposing the helpers over on https://github.com/artsy/cohesion/pull/48 so marking it as WIP for now. 😝

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/Components/Header.tsx
```diff
@@ -1,11 +1,22 @@
 import { Box, Button, Sans, Spacer } from "@artsy/palette"
 import React from "react"
+import { ContextModule, OwnerType, TappedConsignArgs } from "@artsy/cohesion"
 
 interface HeaderProps {
-  onCTAPress?: () => void
+  handleConsignPress: (tappedConsignArgs: TappedConsignArgs) => void
```
damassi, line 6: This an incorrect idiom in React -- callback functions that come into a component via props are prefixed with `onSomething`; internal handlers are prefixed with `handleSomething` to distinguish directional control flow.

jonallured: Hrm, I must be missing something, I thought it was supposed to look like this:

```ts
onSomething={handleSomething}
```

Have I got that wrong?

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/Components/Footer.tsx
```diff
@@ -2,12 +2,23 @@ import { Box, Button, Flex, Sans, Spacer } from "@artsy/palette"
...
-export const Footer: React.FC<FooterProps> = ({ onCTAPress }) => {
+export const Footer: React.FC<FooterProps> = ({ handleConsignPress }) => {
+  const event = {
+    contextModule: ContextModule.sellFooter,
+    contextScreenOwnerType: OwnerType.sell,
+    subject: "Start selling",
+  }
+
+  const handlePress = () => {
+    handleConsignPress(event)
```
damassi, line 19: Are these intermediate variables necessary? 

```
onPress={() => handleConsignPress({ ... })}
```

Also "event" as a variable name is easily confused with events flowing through the internal eventing system. Its somewhat uncommon to piggyback on that keyword outside of an event handler.

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/Components/Header.tsx
damassi: Oh no you're right! This just changes if the function is being passed in or calls out to the parent. In this case here, its totally cool to have an internal function called `handleSomething` that then invokes the `onSomething` callback that's passed in as props. The rule of thumb is:

onSomething -> out to in
handleSomething -> in to outer internal scope

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/Components/Footer.tsx
```diff
@@ -2,12 +2,23 @@ import { Box, Button, Flex, Sans, Spacer } from "@artsy/palette"
...
+import { ContextModule, OwnerType, TappedConsignArgs } from "@artsy/cohesion"
 
 interface FooterProps {
-  onCTAPress?: () => void
+  handleConsignPress: (tappedConsignArgs: TappedConsignArgs) => void
 }
 
-export const Footer: React.FC<FooterProps> = ({ onCTAPress }) => {
+export const Footer: React.FC<FooterProps> = ({ handleConsignPress }) => {
+  const event = {
```
eessex, line 12: In order to get type correction closer to these args, might be a good idea to type `event` as `TappedConsignArgs` or just move the helper `tappedConsign` to this level.

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/Components/ArtistList.tsx
```diff
@@ -65,12 +67,20 @@ export const ArtistListFragmentContainer = createFragmentContainer(ArtistList, {
   `,
 })
 
+const trackingEvent = {
```
eessex, line 70: This needs to use the `tappedEntityGroup` helper: https://cohesion.artsy.net/modules/_events_tap_tappedentitygroup_.html

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/Components/Header.tsx
jonallured: Ok, I think I got it, will update! 🙌

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/Components/ArtistList.tsx
jonallured: Gotcha, I was able to follow this nudge and get something like this going:

```ts
const trackingArgs: TappedEntityGroupArgs = {
  contextModule: ContextModule.artistHighDemandGrid,
  contextScreenOwnerType: OwnerType.sell,
  destinationScreenOwnerType: OwnerType.artist,
  moduleHeight: "single",
  type: "thumbnail",
}

tracking.trackEvent(tappedEntityGroup(trackingArgs))
```

And all that makes sense to me now, except for the `moduleHeight` and `type` properties. Those are required by the `TappedEntityGroupArgs` but do not appear in the tracking sheet that @louislecluse setup here:

https://docs.google.com/spreadsheets/d/1Jo8Ns8p0N-pEWKvsfMBGY-UuU7lTQnFnrUbSjFJO5sU/edit#gid=0

So maybe you can help me understand that part?

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/Components/Header.tsx
jonallured: Ok, updated with e170b26 - is that what you were thinking? Thanks for helping me understand this stuff!! 😅

jonallured:
Ok, I took for the WIP label as the cohesion update was merged in and this will be ready to go once feedback has been addressed. Thanks all!!

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/Components/ArtistList.tsx
louislecluse: I didn't add that to the tracking because it doesn't seem relevant to track that in this use case.
Are these properties absolutely required?

eessex: cc @mikehrom on properties

mikehrom: `module_height` is probably a property we should make optional at this point, but it was informed by designs for new rails on iOS -- for example, fairs and auctions card rails were 'double' height and recently saved/viewed rails are 'single' height, but I would agree this is not a top concern to track.

`type` indicates what is being tapped within the grouping. this might be a header (like the link to the generic auctions home page within the auctions rail) or a thumbnail (artwork bricks) among other options. In this case, I believe the only thing that can be tapped in this rail is a thumbnail, so 'thumbnail' will do

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/Components/Footer.tsx
```diff
@@ -1,13 +1,24 @@
...
+
 interface FooterProps {
-  onCTAPress?: () => void
+  onConsignPress: (tappedConsignArgs: TappedConsignArgs) => void
 }
 
-export const Footer: React.FC<FooterProps> = ({ onCTAPress }) => {
+export const Footer: React.FC<FooterProps> = ({ onConsignPress }) => {
+  const handlePress = () => {
+    onConsignPress(consignArgs)
```
damassi, line 19: Looks good @jonallured 👍

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/Components/ArtistList.tsx
jonallured: Cool, thanks pals! I have open https://github.com/artsy/cohesion/pull/50 to mark this as optional and then I'll yank module height from the args.

louislecluse: Added `type: "thumbnail"` to the tracking spreadsheet!

jonallured:
Alright, I finally got this all straightened out so going to head into the weekend with a merged PR. 😎

PR #3351, opened 2020-05-21
Add the ability to request an email confirmation

yuki24:
addresses https://artsyproduct.atlassian.net/browse/AUCT-994

This adds a new email confirmation banner to the `<Home>` screen. This doesn't directly replace the consignments banner as it's so much easier to add a feature flag and I think we could just hide it by default and enable for e.g. QA when we want to. This way we could safely merge this PR without affecting or blocking any other teams.

In order to enable this feature, both the `Email Confirmation Banner` and `Move City Guide, Enable Sales` should be on:

Screen Shot 2020-05-21 at 3 46 24 PM

### Happy Path:

AUCT-994-happy-path

### Already Confirmed Email:

AUCT-994-already-confirmed

### In Case of a Failure:

AUCT-994-failure


src/lib/Scenes/Home/Components/__tests__/EmailConfirmationBanner-tests.tsx
```diff
@@ -0,0 +1,203 @@
+import { Theme } from "@artsy/palette"
+import React from "react"
+import { graphql } from "react-relay"
+
+import { EmailConfirmationBannerTestsQuery } from "__generated__/EmailConfirmationBannerTestsQuery.graphql"
+import { renderRelayTree } from "lib/tests/renderRelayTree"
```
ds300, line 6: nonblocking: We're moving away from enzyme (which renderRelayTree is built on) so in the future it'd be great if you could use `react-test-renderer` and `relay-test-utils` for writing tests.

src/lib/Scenes/Home/Components/EmailConfirmationBanner.tsx
```diff
@@ -0,0 +1,132 @@
...
+import { commitMutation, createFragmentContainer, graphql, RelayProp } from "react-relay"
+import { Environment } from "relay-runtime"
+
+import { EmailConfirmationBanner_me } from "__generated__/EmailConfirmationBanner_me.graphql"
+import {
+  EmailConfirmationBannerMutation,
+  EmailConfirmationBannerMutationResponse,
+} from "__generated__/EmailConfirmationBannerMutation.graphql"
+
+type UnconfirmedEmail = string | null | undefined
```
dleve123, line 13: Mostly for my own education, why do we need to have both `null` and `undefined` in the union type?

src/lib/Scenes/Home/Components/EmailConfirmationBanner.tsx
```diff
@@ -0,0 +1,132 @@
...
+      variables: {},
+    })
+  })
+}
+
+export const EmailConfirmationBanner: React.FC<Props> = ({ me, relay }) => {
+  const [shouldDisplayBanner, toggleVisible] = useState<boolean>(me.canRequestEmailConfirmation)
+  const [isLoading, setLoading] = useState<boolean>(false)
+  const [unconfirmedEmail, setUnconfirmedEmail] = useState<UnconfirmedEmail>(null)
+  const [errorMessage, setErrorMessage] = useState<string | null>(null)
```
dleve123, line 53: minor: thoughts on extracting a type for `ErrorMessage`, similar to what you do for `UnconfirmedEmail`? Might be a bit more consistent that way.

src/lib/Scenes/Home/Components/EmailConfirmationBanner.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  const didTapResend = async () => {
+    try {
+      setLoading(true)
+
+      const { sendConfirmationEmail } = await submitMutation(relay.environment)
+      const confirmationOrError = sendConfirmationEmail?.confirmationOrError
```
dleve123, line 60: The stricter type checking in Eigen is definitely interesting! For instance, is it deliberate or incidental that it's possible for `sendConfirmationEmail` to be null? Taking a look at the MP schema:
```
  sendConfirmationEmail(
    input: SendConfirmationEmailMutationInput!
  ): SendConfirmationEmailMutationPayload
```
(https://github.com/artsy/metaphysics/blob/master/_schemaV2.graphql#L5872)

there's no `!` at the end of `SendConfirmationEmailMutationPayload`, so this all makes sense from a types perspective, but does make me wonder why isn't MP typed so that there's always a value of type `SendConfirmationEmailMutationPayload` returned -- aka `SendConfirmationEmailMutationPayload!`

Not suggesting any changes with this comment, but curious if you have any thoughts here.

src/lib/Scenes/Home/Components/EmailConfirmationBanner.tsx
```diff
@@ -0,0 +1,132 @@
...
+        switch (message) {
+          case "email is already confirmed":
+            setErrorMessage("Your email is already confirmed")
+            break
+          default:
+            setErrorMessage("Something went wrong. Try again?")
+            break
+        }
+      }
+    } catch (error) {
```
dleve123, line 78: So this catch helps ensure that if the underlying APIs change such that Eigen doesn't know how to properly a response from MP, the user is still able to use Eigen in a slightly degraded state, is that right?

How does this catch interplay with the [`onError` key-value](https://github.com/artsy/eigen/pull/3351/files#diff-868d2a90cb57dc630dfa7cc267ef1b80R26)?

src/lib/Scenes/Home/Components/EmailConfirmationBanner.tsx
```diff
@@ -0,0 +1,132 @@
...
+        {isLoading ? (
+          <>
+            <Text>Sending an confirmation email...</Text>
+
+            <Flex pr="1">
+              <Spinner size="small" color="white100" />
+            </Flex>
+          </>
+        ) : (
+          <TouchableWithoutFeedback onPress={unconfirmedEmail ? undefined : didTapResend}>
```
dleve123, line 105: minor / more opinion than anything else – seems like there are 2 modes that this banner can be in:

1. about to be clicked / pending (aka `unconfirmedEmail` is `null`)
2. clicked / requested (aka `unconfirmedEmail` is `string` OR there's an error message present -- generally there's a response from the mutation)

I wonder if there would be any value in creating `PendingConfirmation` & `ConfirmationRequested` private subcomponents to represent these states? This would introduce some duplication of palette components, but might be worth it?

Food for thought -- at the very least, I hope I'm at least understanding how this works.

src/lib/Scenes/Home/Components/__tests__/EmailConfirmationBanner-tests.tsx
```diff
@@ -0,0 +1,203 @@
...
+    await component
+      .find(TouchableWithoutFeedback)
+      .at(0)
+      .props()
+      .onPress()
+
+    expect(component.text()).toEqual("Email sent to i.am.unconfirmed@gmail.com")
+  })
+
+  it("should not be clickable after a successful response", async () => {
```
dleve123, line 117: should it be clickable after an error? Thinking out loud here: I don't think it would be clickable again if you are already confirmed, but on other errors (sporatic MP downtime) I think it should be clickable.

yuki24: > I don't think it would be clickable again if you are already confirmed, but on other errors (sporatic MP downtime) I think it should be clickable.

That's exactly how the banner behaves as it is right now.

dleve123: Awesome! I see you just extended some tests to document/verify this behavior as well 👌

src/lib/Scenes/Home/Components/EmailConfirmationBanner.tsx
yuki24: I've updated this to be just `message` so you could set an appropriate message depending on the situations: https://github.com/artsy/eigen/pull/3351/commits/67a71c59e06763ea6ed7010493eac0390f619bbf and that also helped remove a few conditionals . The `UnconfirmedEmail` is also gone now since we don't need a concrete type any more.

src/lib/Scenes/Home/Components/EmailConfirmationBanner.tsx
yuki24: I think the intention is that any nested object/field could be `null` so we tend to default to optionals by default, which leads to a type like this. The optional chain here is a bit odd, but at the same time it isn't too verbose.

src/lib/Scenes/Home/Components/EmailConfirmationBanner.tsx
yuki24: Generally this error could be an underlaying Network error that is thrown by the device or the client or an authentication error.

src/lib/Scenes/Home/Components/EmailConfirmationBanner.tsx
yuki24: Now that we have this refactor https://github.com/artsy/eigen/pull/3351/commits/9381e048ae8ec03a77aedd359a5023542ceb6270#diff-868d2a90cb57dc630dfa7cc267ef1b80R104 I'm not sure we have to be very sophisticated about the abstraction here.

src/lib/Scenes/Home/Components/EmailConfirmationBanner.tsx
yuki24: I've actually removed this type: 67a71c59e06763ea6ed7010493eac0390f619bbf

dleve123:
re-assigned to @yuki24 as this PR already has multiple approving reviews, so I don't think there's a need to wait on more reviews. Merge at your discretion @yuki24.

yuki24:
Just Copying [the Slack thread](https://artsy.slack.com/archives/CA3D02SDN/p1590594079128100), the tabs are implemented in Obj-C and at least it would be a bit more work to send e.g. tab a tap event all the way down to the banner. The test helper shouldn't block merging this, so I'll merge and send it to QA.

PR #3353, opened 2020-05-22
[MX-326] Fix Explore unfocused states

ds300:
https://artsyproduct.atlassian.net/browse/MX-326

spec: https://www.figma.com/file/2Im0qW2WpwQj2BowaFypxG/Explore?node-id=322%3A1143

1. Make sure the city guide call-to-action is visible when there are recent searches
2. Updated the 'Recent Searches' title copy to match the spec.
3. Updated the search page typography to match the spec (serif -> sans)
3. Fix an aspect ratio bug in OpaqueImageView which I introduced yesterday 😬 
4. Fix a crash that happened when artist pages have no tabs (data folks say this shouldn't happen but it does, see, e.g. Jillian Banks) – I used the empty state copy from the web.
5. Strictness migration cleanup

### Explore with recent searches
image

### Explore without recent searches
image

### DIFF
--- a/src/lib/Containers/Artist.tsx
+++ b/src/lib/Containers/Artist.tsx
@@ -1,4 +1,4 @@
-import { Flex, Separator, Spacer, Theme } from "@artsy/palette"
+import { Flex, Message, Separator, Spacer, Theme } from "@artsy/palette"
 import {
   ArtistAboveTheFoldQuery,
   ArtistAboveTheFoldQueryVariables,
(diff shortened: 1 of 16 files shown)


src/lib/Scenes/Search/__tests__/Search-tests.tsx
```diff
@@ -44,21 +43,44 @@ describe("The Search page", () => {
...
   it(`shows city guide entrance when flag is enabled and on iPhone`, async () => {
     const isPadMock = isPad as jest.Mock
     isPadMock.mockImplementationOnce(() => false)
     NativeModules.Emission.options.AROptionsMoveCityGuideEnableSales = true
     const tree = ReactTestRenderer.create(<TestWrapper />)
-    expect(extractText(tree.root.findByType(SearchEmptyState))).toContain("Explore Art on View by City")
+    expect(extractText(tree.root.findByType(CityGuideCTA))).toContain("Explore Art on View by City")
+  })
+
+  it(`shows city guide entrance when flag is enabled and on iPhone and there are recent searches`, async () => {
```
brainbicycle, line 57: 💥

src/lib/Containers/Artist.tsx
```diff
@@ -54,6 +55,20 @@ export const Artist: React.FC<{
     })
   }
 
+  if (tabs.length === 0) {
```
brainbicycle, line 58: Nice 👍

brainbicycle:
Looks great! Love all the strictness fixes 👍

PR #3357, opened 2020-05-26
Don't show 'sold for' message if no realized price

pepopowitz:
Addresses https://artsyproduct.atlassian.net/browse/CSGN-205

Artworks in our recently sold rail may or may not have a realized price. This PR protects against artworks that don't, by suppressing the "Sold for" message if there is no `realizedPrice` property.

## Before

image

## After

image

### DIFF
--- a/src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/Components/__tests__/RecentlySold-tests.tsx
+++ b/src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/Components/__tests__/RecentlySold-tests.tsx
@@ -1,0 +1,73 @@
+import { Theme } from "@artsy/palette"
+import React from "react"
+import "react-native"
+import { create } from "react-test-renderer"
+
+import { RecentlySold_artists } from "__generated__/RecentlySold_artists.graphql"
+import { extractText } from "lib/tests/extractText"
+import { RecentlySold } from "../RecentlySold"
+
+describe("RecentlySold", () => {
+  const defaultArtist: any = {
+    artistNames: "Andy Goldsworthy",
+    href: "/artist/andy-goldsworthy",
+    image: {
+      imageURL: "http://placekitten.com/200/200",
+    },
+    internalID: "1235",
+    realizedPrice: "$1,200",
+    slug: "andy-goldsworthy",
+  }
+
+  it("renders sale message if artwork has realized price", () => {
+    const artist = {
+      ...defaultArtist,
+      realizedPrice: "$1,200",
+    }
+    const artists = makeArtists(artist)
+
+    const tree = create(
+      <Theme>
+        <RecentlySold artists={artists} />
+      </Theme>
+    )
+
+    expect(extractText(tree.root)).toContain("Sold for $1,200")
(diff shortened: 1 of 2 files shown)


src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/Components/__tests__/RecentlySold-tests.tsx
```diff
@@ -0,0 +1,73 @@
+import { Theme } from "@artsy/palette"
+// @ts-ignore STRICTNESS_MIGRATION
```
jonallured, line 2: Looks like this is failing CI:

```
WARNING: The number of comments with the substring STRICTNESS_MIGRATION has risen from 1561 in master to 1562 in this PR.
 
These comments were added when we switched on TypeScript's strict mode, and their number should only ever go down.
 

 
It is safe to ignore this failure if you are in a hurry, but please consider refactoring to remove these tags.
```

https://app.circleci.com/pipelines/github/artsy/eigen/1882/workflows/2bcacd24-85fa-49e0-b383-9084eb957557/jobs/10611/steps

Is there an easy way to fix this one?

pepopowitz: I don't think there is an "easy" way. The type error that's disabled via `@ts-ignore STRICTNESS_MIGRATION` is due to a lack of types for enzyme - the error gives the suggestion to install `@types/enzyme` as a dependency. Unfortunately, installing `@types/enzyme` uncovers ~85 new type-checking errors.

I'm personally more comfortable adding this one ts-ignore than trying to clean up 85 unrelated type errors in this PR.

jonallured: Yikes! Yeah that sounds like a can of worms. So I guess we need to update that check and raise the number by one? Or is there some other way the @artsy/mobile-experience team would like us to handle this?

brainbicycle: Sorry for the delay just seeing this, we are preferring `react-test-renderer` over `enzyme` for testing components partly for this reason. There are some examples of tests in the preferred practices doc here: https://github.com/artsy/eigen/blob/master/docs/map_to_the_app.md#write-unit-tests-for-new-components

pepopowitz: Ok, I'll rewrite the tests!

brainbicycle: Thanks 🎉

pepopowitz: ✅Rewritten

PR #3359, opened 2020-05-26
enable full bleed on artist series rail

ashleyjelks:
UI Bug Fix where the Iconic Collections rail wasn't doing a full bleed across the screen:

https://artsyproduct.atlassian.net/browse/FX-1975

Kapture 2020-05-26 at 12 33 35

#trivial


sweir27:
Looks good! As we add more to this page we might want to remove the default margin around the content and have each component control their own, but this seems fine for now 👍 .

PR #3363, opened 2020-05-26
Viewing room QA [GALL-2865]

mdole:
**Edit:** going to remove WIP + mark this one for review/merge and follow up with other stuff in a separate PR in the interest of incremental progress + not leaving this stuck open for a while!

Addressing a bunch of issues that came up during QA! Copy/pasting the list I'm working through from the Jira ticket:

### Definitely need to investigate/fix:
- [ ] **Following up in a separate PR** Hero** image is stretched out on some phones (might need to wait until we know how we're storing + accessing images for VRs in general)
- [x] Artworks page doesn't load more artworks when you hit the bottom
- [x] Artworks on the list view flash when tapping - can we exclude the text from the flashing effect and make the opacity over the image lighter to match the artwork rail?
- [x] View works button should have a 150ms fade-in (possibly `duration-2`)
- [x] Top navbar should have a divider (`black10`, matching home/settings tab bar)
- [x] Tracking: should we be using `action_name` or something else to correctly register the action type? (banner that shows tracking events registering a view event when it seems like we want other types like tap)
- [x] Floating button should have drop shadow `box-shadow: 0px 3px 3px rgba(0, 0, 0, 0.12);`
- [ ] **Following up in a separate PR** Update countdown to match web: https://www.notion.so/artsy/Update-countdown-to-match-web-faaa9bc9ab894359b862e253f105a6e7
- [x] Use `additionalInformation` instead of `description` for additional artwork text: https://github.com/artsy/eigen/pull/3339

### Less critical/app-wide issues:

- [ ] Loading new screens causes a strange up/down shifting effect (also happens on home screen when tapping into collections, possibly other types) - may be due to different page headers Slack link
- [ ] Deep linking shows "unable to load" (staging issue?)
- [ ] Deep linking to a screen you're already on adds another copy of it to the stack (app-wide behavior)

[Jira ticket](https://artsyproduct.atlassian.net/browse/GALL-2865)


src/lib/Scenes/ViewingRoom/Components/ViewingRoomArtworkRail.tsx
```diff
@@ -27,7 +27,10 @@ export const ViewingRoomArtworkRail: React.FC<ViewingRoomArtworkRailProps> = pro
           title={`${totalCount} ${pluralizedArtworksCount}`}
           onPress={() => {
             tracking.trackEvent(tracks.tappedArtworkGroupHeader(viewingRoom.internalID, viewingRoom.slug))
-            SwitchBoard.presentNavigationViewController(navRef.current!, `/viewing-room/${viewingRoom.slug}/artworks`)
+            SwitchBoard.presentNavigationViewController(
+              navRef.current!,
+              `/viewing-room/${viewingRoom.internalID}/artworks`
+            )
```
mdole, line 33: @ds300 would love your help understanding this one - @starsirius noticed during QA that we weren't fetching more artworks when we hit the bottom of the page. I looked into it, and it seems like by passing the slug to the navigation controller, we ended up using the slug instead of the `internalID` when trying to fetch more artworks, which didn't work for some reason. Changing to using `internalID` like this ^ solves the issue, but strikes me as maybe an anti-pattern.

So what I'm not sure I understand here:
1. how the URI passed to the navigation controller is used in the initial fetch
2. why the slug was getting used instead of `internalID` when fetching more works (I think the answer to this might be "because it was used in the first fetch and was therefore kept in `fragmentVariables`, perhaps?)
3. why using the slug instead of the ID wouldn't work when fetching more

Any light you can shed would be appreciated! Also happy to chat if you don't want to write me an essay :)

mdole: Oh and here's some Slack back-and-forth that may be helpful as additional context: [link](https://artsy.slack.com/archives/CP9P4KR35/p1590523133367300)

ds300: > how the URI passed to the navigation controller is used in the initial fetch

Switchboard plucks out the id/slug from the path and passes it as a prop to the component https://github.com/artsy/eigen/blob/63f5edc3f51e8bc3660a4ccae8a31883d9a470fb/Artsy/App/ARSwitchBoard.m#L245-L248

As for the other two points, it looks like the situation is that before this PR the `slug` is used for the initial fetch in the `ViewingRoomArtworksRenderer` 

https://github.com/artsy/eigen/blob/63f5edc3f51e8bc3660a4ccae8a31883d9a470fb/src/lib/Scenes/ViewingRoom/ViewingRoomArtworks.tsx#L195

And then the `internalID` was used for subsequent fetches in the pagination container

https://github.com/artsy/eigen/blob/master/src/lib/Scenes/ViewingRoom/ViewingRoomArtworks.tsx#L174

and my best guess for why this doesn't work is that the cursors are generated based on the `id` parameter passed in to the query. So even though the `slug` and the `internalID` identify the same viewing room, because they're different strings they give different pagination cursors and that initial 'fetch more' operation fails.

ds300: (that was just a guess, I'm not certain that ids do factor into the cursors, would be happy to look into the source with you to figure it out)

mdole: Ahhh thanks @ds300, that makes a lot of sense. I switched back to using slugs for navigation + for pagination in f507d66 and that works just fine. I'm also a little unclear about the cursors part but that sounds like the probable cause to me

PR #3364, opened 2020-05-26
Point consignments landing page at targetSupply root field instead of artist-based targetSupply

pepopowitz:
Addresses https://artsyproduct.atlassian.net/browse/CSGN-190 and https://artsyproduct.atlassian.net/browse/CSGN-206

We added [a new root field in metaphysics named `targetSupply`](https://github.com/artsy/metaphysics/pull/2414). This PR switches our consignments landing page to use that instead of the `targetSupply` inside of artists. This allows us to remove the hardcoded list of artist IDs from here, since MP will now control that. 

Included in this work is shuffling the order of items in the ArtistList and RecentlySold rails.


src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/Components/ArtistList.tsx
```diff
@@ -74,7 +84,7 @@ const trackingArgs: TappedEntityGroupArgs = {
   type: "thumbnail",
 }
 
-const ArtistItem: React.FC<{ artist: ArtistList_artists[0] }> = ({ artist }) => {
+const ArtistItem: React.FC<{ artist: any }> = ({ artist }) => {
```
pepopowitz, line 87: I'm not happy about removing a type here, but I'm not sure how to access an inner type of a nullable array. `ArtistList_targetSupply['microfunnel'][0][artist]` makes TypeScript mad, because `microfunnel` can be an array _or_ null.

I asked in slack; if anyone here knows, I'd love a hint.

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/Components/__tests__/ArtistList-tests.tsx
```diff
@@ -0,0 +1,57 @@
...
+import React from "react"
+import "react-native"
+import { create } from "react-test-renderer"
+
+import { ArtistList_targetSupply } from "__generated__/ArtistList_targetSupply.graphql"
+import { extractText } from "lib/tests/extractText"
+import { ArtistList } from "../ArtistList"
+
+describe("ArtistList", () => {
+  const defaultArtist: any = {
```
pepopowitz, line 11: Typed as `any` because of the same problem described above (accessing inner types on nullable arrays).

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/index.tsx
```diff
@@ -58,41 +58,18 @@ export const ConsignmentsHomeQueryRenderer: React.FC = () => {
...
       `}
       render={renderWithPlaceholder({
         Container: ConsignmentsHomeContainer,
-        renderPlaceholder: () => <ConsignmentsHome isLoading={true} artists={null as any} />,
+        renderPlaceholder: () => <ConsignmentsHome isLoading={true} targetSupply={null as any} />,
       })}
     />
   )
 }
-
```
pepopowitz, line 76: This is the main point of this PR - get the hardcoded list out of eigen, for people who never update their app.

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/Components/ArtistList.tsx
damassi, line 87: No go with `ArtistList_targetSupply['microfunnel']['artist'][0]`?

pepopowitz: @ds300 you are my goto TypeScript resource, maybe you know the answer?

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/Components/RecentlySold.tsx
```diff
@@ -11,17 +11,24 @@ import { useTracking } from "react-tracking"
...
 
   const navRef = useRef<any>()
   const tracking = useTracking()
 
+  const microfunnelItems = targetSupply.microfunnel || []
+  if (microfunnelItems.length === 0) {
+    return null
+  }
+
+  const recentlySoldArtworks = microfunnelItems.map(x => x?.artworksConnection?.edges?.[0]?.node)
```
damassi, line 30: Did `shuffle` work here?

pepopowitz: I'm planning to follow that up separately.

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/Components/ArtistList.tsx
damassi: (Ah, i see comment up above)

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/Components/RecentlySold.tsx
damassi: cool 👍

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/Components/ArtistList.tsx
ds300: It's ugly but the only way I know how to do this is to wrap each nullable level with `NonNullable` e.g. https://github.com/artsy/eigen/blob/3100b5f4beba8b49d3811b2767bab590898d48df/src/lib/Components/Lists/SavedFairItemRow.tsx#L15-L17

jonallured: I might be muddying the waters here but are we able to adjust the underlying types here? I would love to see the nullable array lose it's nullability! 😝 

Empty arrays > nullable arrays

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/Components/RecentlySold.tsx
```diff
@@ -11,17 +11,24 @@ import { useTracking } from "react-tracking"
...
-export const RecentlySold: React.FC<RecentlySoldProps> = ({ artists, isLoading }) => {
+export const RecentlySold: React.FC<RecentlySoldProps> = ({ targetSupply, isLoading }) => {
   if (isLoading) {
     return <RecentlySoldPlaceholder />
   }
 
   const navRef = useRef<any>()
   const tracking = useTracking()
 
+  const microfunnelItems = targetSupply.microfunnel || []
```
jonallured, line 25: This drives me crazy - having to coerce a nullable into an empty points to woes higher in the stack, no?

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/index.tsx
jonallured: ❤️ ❤️ ❤️

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/Components/RecentlySold.tsx
pepopowitz: I'm not sure how to control the nullability of these properties. They're generated by Relay, so maybe there's something in the way they're defined in metaphysics that convinces relay to make them nullable?

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/Components/ArtistList.tsx
pepopowitz: As far as I understand, the nullability is there because all types in graphql are nullable by default. We do have the ability to override that in metaphysics with the `GraphQLNonNull` type. 

Whether that's the right thing to do, I can't say. As far as I can tell we only use `GraphQLNonNull` for scalar types, not object or list types. I can't speak to whether that's good or bad, but I've read a couple articles now that encourage thinking extra hard about when you make non-scalar types non-null. (https://www.apollographql.com/blog/using-nullability-in-graphql-2254f84c4ed7 and https://medium.com/@calebmer/when-to-use-graphql-non-null-fields-4059337f6fc8).

Regardless of where we fall on null/non-null at the schema level, I think those changes fall outside the scope of this PR. TBH I don't really have a problem with `|| []` and wrapping things in `NonNullable<T>`, so I'd prefer for you to drive that if you're interested @jonallured.

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/Components/RecentlySold.tsx
damassi: Yeah @jonallured - this come up a lot lately, a large number of our fields in MP are nullable, erroniously. Some of this stuff has started to be looked at. 

@pepopowitz - not suggesting this, but if you're _sure_ that data will not be null can use the `!` operator, eg `foo!.bar!.baz`. But our schema is scary in this way... prob better to be defensive 👍

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/index.tsx
damassi: Really nice work all around

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/Components/ArtistList.tsx
ds300: > I might be muddying the waters here but are we able to adjust the underlying types here? I would love to see the nullable array lose it's nullability! 😝

I have some in-progress work to do that for connections over at https://github.com/artsy/metaphysics/pull/2346 

SOON

PR #3365, opened 2020-05-27
[MX-327] Simplify Safe Area Handling

ds300:
https://artsyproduct.atlassian.net/browse/MX-327

🚨 The first few commits here are mostly-unrelated bonus fixes I made while QA-ing the change (I had to go through every single view controller in the app to make sure everything is still hunky dory)

The problem this fixes is that sometimes, when navigating between certain view controllers, content could jump around. This was caused by the fact that the navigation infrastructure had been given the responsibility of adding/removing page padding for the safe area insets (status bar in old ios parlance). Different view controllers with different preferences about whether or not they need extra padding in the same navigation stack? Jumpy content.

This PR removes that responsibility from the navigation infra, and pushes it into the view controllers themselves. In other words, all view controllers now take up the whole screen and have to make sure their own content is not obscured by hardware or status bar stuff outside of the safe area.


emission/Pod/Classes/ViewControllers/ARComponentViewController.m
```diff
@@ -136,20 +127,4 @@ - (void)setProperty:(id)value forKey:(NSString *)key
     }
 }
 
-
--(void)viewSafeAreaInsetsDidChange
```
brainbicycle, line 140: `@property (nonatomic) BOOL safeAreaInsetsWereUpdated;`

Believe we can remove this property now.

src/lib/AppRegistry.tsx
```diff
@@ -236,58 +237,77 @@ const SearchWithTracking: React.SFC<SearchWithTrackingProps> = screenTrack<Searc
...
+  render() {
+    return (
+      <ProvideScreenDimensions>
+        <InnerPageWrapper {...this.props} />
+      </ProvideScreenDimensions>
+    )
+  }
+}
+
+function register(screenName: string, Component: React.ComponentType<any>, options?: PageWrapperProps) {
```
brainbicycle, line 261: 🎉  a lot more explicit and centralized, love it!

Artsy/View_Controllers/Util/ARNavigationController.m
```diff
@@ -192,14 +192,12 @@ - (void)navigationController:(UINavigationController *)navigationController will
     // If it is a non-interactive transition, we fade the buttons in or out
     // ourselves. Otherwise, we'll leave it to the interactive transition.
     if (self.interactiveTransitionHandler == nil) {
-        [self showBackButton:[self shouldShowBackButtonForViewController:viewController] animated:animated];
-        [self showStatusBarBackground:[self shouldShowStatusBarBackgroundForViewController:viewController] animated:animated white:useWhite];
```
brainbicycle, line 196: It looks like we are no longer using this `useWhite` var which updates the statusBar background view and instead relying on react native views to extend beneath status bar. If that is the case we should be able to get rid of the var
here: https://github.com/artsy/eigen/blob/e981d91454b699b326670ea8b03047c9bdfc23ac/Artsy/View_Controllers/Util/ARNavigationController.m#L190
and the associated method for generating it for a viewController:
https://github.com/artsy/eigen/blob/e981d91454b699b326670ea8b03047c9bdfc23ac/Artsy/View_Controllers/Util/ARNavigationController.m#L183
and maybe even the statusBarBackground view in the ARTopMenuViewController?:
https://github.com/artsy/eigen/blob/e981d91454b699b326670ea8b03047c9bdfc23ac/Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m#L89

brainbicycle:
Looks great! There might be some more dead code with these changes we can remove but other than that LGTM!

Artsy/View_Controllers/Util/ARNavigationController.m
ds300: great catch! 🙇

PR #3372, opened 2020-05-27
Replace enzyme usages with react-test-renderer

yuki24:
Just to follow up on https://github.com/artsy/eigen/pull/3351 I tried to replace enzyme usages with `react-test-renderer`. This was mostly straight forward except I cuoldn't figure out how I could get the tests pass without `done` and `setTimeout`. Perhaps I'm missing something, so it would be great if you could review and guide me to the right direction.

#trivial


ds300:
Looks like your interaction-handling logic is async so you could use `flushPromiseQueue` instead. Also you shouldn't need to use `act` when mocking operations :)

image

ds300:
@yuki24 I noticed this test file is failing in master so to unblock other PRs I'm going to apply the suggested change and get this merged. Big thank you for going back and converting these tests 🙏

yuki24:
Oh thanks for taking care of that @ds300! TIL `flushPromiseQueue`.

PR #3377, opened 2020-05-28
Send users to the new sell tab when they tap on the artist consign CTA

pepopowitz:
Addresses https://artsyproduct.atlassian.net/browse/CSGN-210.

When we built out the new sell tab, we forgot that tapping the "consign with Artsy" button on an artist page would direct the user to the new landing page. It wasn't so much a problem that we were sending them there...but they were sent to the landing page in the context of a modal without a mechanism for closing or going back. They were dead-ended, and the only way to escape was to restart the app.

This is what it looked like:

artist-consign-flow-small-stuck

This PR updates the destination of that button to be the new sell tab (if the feature flag is enabled). 

This works really well when they tap an artist consign button while browsing within the home or search tabs of the app:

artist-consign-flow-small-good

It is a little less great when they tap an artist consign button while browsing within the new sell tab: 

artist-consign-flow-small-confusing

The main issue there being that when they're brought to the tab landing page, it has retained their previous scroll position, which feels confusing. 

We're in talks amongst the consignments team to figure out what to do about that last issue. For now, even though it may be confusing, it is significantly better than dead-ending them.

### DIFF
--- a/src/lib/Components/Artist/__tests__/ArtistConsignButton-tests.tsx
+++ b/src/lib/Components/Artist/__tests__/ArtistConsignButton-tests.tsx
@@ -144,6 +144,7 @@ describe("ArtistConsignButton", () => {
       })
     })
 
+    // TODO: make this the default case once the feature flag is removed
     it("tracks the sales tab destination if feature flag is enabled", () => {
       NativeModules.Emission.options.AROptionsMoveCityGuideEnableSales = true
 
@@ -212,6 +213,7 @@ describe("ArtistConsignButton", () => {
       })
     })
 
+    // TODO: make this the default case once the feature flag is removed
     it("tracks the sales tab destination if feature flag is enabled", () => {
       NativeModules.Emission.options.AROptionsMoveCityGuideEnableSales = true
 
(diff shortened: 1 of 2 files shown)


src/lib/Components/Artist/__tests__/ArtistConsignButton-tests.tsx
```diff
@@ -137,6 +143,25 @@ describe("ArtistConsignButton", () => {
         destination_path: "/consign/submission",
       })
     })
+
+    it("tracks the sales tab destination if feature flag is enabled", () => {
```
damassi, line 147: 👍

src/lib/Components/Artist/__tests__/ArtistConsignButton-tests.tsx
```diff
@@ -186,5 +211,24 @@ describe("ArtistConsignButton", () => {
...
```
damassi, line 215: Once this is deployed and feature flag removed will these tests have to be updated? If so can we add a TODO somewhere?

pepopowitz:
```suggestion
    // TODO: make this the default case once the feature flag is removed
    it("tracks the sales tab destination if feature flag is enabled", () => {
```

src/lib/Components/Artist/__tests__/ArtistConsignButton-tests.tsx
pepopowitz, line 147:
```suggestion
    // TODO: make this the default case once the feature flag is removed
    it("tracks the sales tab destination if feature flag is enabled", () => {
```

jonallured:
LGTM!! 🎉 #mergeongreen

PR #3380, opened 2020-05-29
Fix no-text-outside-Text issue

ds300:
react-native throws a full-on Error when you pass a string as children to a component which is not  wrapped by a `Text` component.

We got some mysterious sentry reports about this happening on the Home tab in the latest release and after digging through the changes to all components on the home tab I'm 90% sure this is the culprit.

Trivial change so I'll self-merge, but it's good to be aware that this `blah && <Blah />` pattern is very unsafe in react-native when the thing on the left is a string or a number. It's necessary to use `blah ?? <Blah />` if you're not concerned about falsey values (`""` and `0`) or `blah ? <Blah /> : null` if you are.

### DIFF
--- a/src/lib/Components/ArtworkTileRail/ArtworkTileRailCard.tsx
+++ b/src/lib/Components/ArtworkTileRail/ArtworkTileRailCard.tsx
@@ -37,17 +37,17 @@ export const ArtworkTileRailCard: React.FC<ArtworkTileRailCardProps> = ({
     />
   )
 
-  const artistNamesDisplay = artistNames && (
+  const artistNamesDisplay = artistNames ? (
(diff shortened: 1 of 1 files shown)


pepopowitz:
Thanks for the tip @ds300! Is there a linting rule we could enable to guard against this?

ds300:
yeah, I'll add a ticket to the MX backlog to enable it, good suggestion! 👍

pepopowitz:
I think the rule we'd be looking for is this one: https://palantir.github.io/tslint/rules/strict-boolean-expressions/

ds300:
@pepopowitz that sounds like fun! I put something in your calendar, feel free to move around if it's not convenient.

PR #3389, opened 2020-06-01
[WIP] Toggle the destination of an artist consign tap based on the currently selected tab

pepopowitz:
This is a WIP because it's kind of mess, things are probably mis-named, I didn't put everything in the right places, and I am heavily cargo-culting my objective-C code. I wanted to get eyes on it to review the approach before I clean it up. I'll also add supporting videos.

## The initial problem

When you are in the Sell tab, and you tap an artist consign button, it takes you to the Sell tab landing page....but retains the scroll position of when you were last there. This is because the route associated with the Sell tab is _also_ the route associated with sending you to the Sell tab landing page. The experience would be better if when you tapped artist consign from the Sell tab, it pushed another visit to the landing page onto the navigation stack. We wouldn't have to worry about your scroll position during the previous visit, and you could interact with that page visit in a way that's familiar to you (e.g. tap `back` to go to the previous page).

## The initial solution

This PR adds a new route for visiting the sell tab landing page aside from the sell tab itself. This solves the navigation problem when you're in the Sell tab.

## The wrench

Having done that, it now becomes important to know where you currently are in the app. We don't want these changes to affect your experience while you're on the Home or Search tab. If you're on any tab but Sell, we still want you to go to `/sales`....if you're on the Sales tab, we want you to go to `/collections/my-collection/landing` (there's probably a better path for that).

## The wrench's solution

This PR also adds a bridge between React and the selected nav item. It currently takes advantage of the analytics tab name and passes that through; that certainly doesn't have to be the final solution. Once the React component has the currently selected tab name, it can choose the destination when you tap on an artist consign button.

## Analytics issue

I noticed that the analytics tab name wasn't being set for the new sell tab, so I updated that too.

### DIFF
--- a/Artsy/App/ARAppDelegate+Emission.m
+++ b/Artsy/App/ARAppDelegate+Emission.m
@@ -245,6 +245,10 @@ FollowRequestFailure(RCTResponseSenderBlock block, BOOL following, NSError *erro
         });
     };
 
+    emission.switchBoardModule.getSelectedTabName = ^(RCTPromiseResolveBlock resolve, RCTPromiseRejectBlock reject) {
+        resolve([[ARTopMenuViewController sharedController] selectedTabName]);
+    };
+
     emission.switchBoardModule.presentNavigationViewController = ^(UIViewController *_fromViewController,
                                                                    NSString *_Nonnull route) {
         UIViewController *viewController = [[ARSwitchBoard sharedInstance] loadPath:route];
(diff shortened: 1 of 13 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>Missing Test Files:

- `src/lib/Scenes/Consignments/__tests__/SellTabLanding-tests.tsx`

    </tr>
  </tbody>
</table>




<p align="right">
</p>

pepopowitz:
Closing in favor of #3400 (and a future followup PR)

PR #3392, opened 2020-06-02
[WIP] createdAccount and successfullyLoggedIn events match v2 schema

eessex:
Because account events are not in React native, we can't enforce type-safety on them via cohesion. 
We did however notice that we are not passing properties consistent to the current schema. Properties passed for `createdAccount` and `successfullyLoggedIn` have been updated to 
- Rename `context_type` as `service` 
- Include `type` and `user_id`
- Update event name (trivial, to indicate we've looked at it since v2)

There was one event fired with `context_type`: `safari keychain`. This is not accounted for in our schema and may be a duplicate fire (as part of email login). @abhitip do you have further context on this event?


eessex:
Due to ambiguity around key names and "safari keychain" property, more investigation into marketing integrations is needed before merging this. Closing for now. cc @abhitip

PR #3393, opened 2020-06-02
Update analytics subject for consignment CTA buttons

pepopowitz:
Addresses https://artsyproduct.atlassian.net/browse/CSGN-227

We were sending inaccurate subjects for the consignment header/footer CTA buttons. This PR updates the subjects to be accurate.

#trivial


jonallured:
❤️❤️❤️

PR #3394, opened 2020-06-02
[WIP] Upgrades to styled-system v5

dzucconi:
Palette PR: https://github.com/artsy/palette/pull/705

Opening this for now to see CI pass and ask what the appropriate Q/A process for something like this would be. In general I'm not super concerned since the breaking changes were mostly about pulling values out of the theme directly which isn't done in this codebase. The only real change here I'm making is removing `"keyofStringsOnly": true` from the tsconfig, which is required for type-check to pass (I'm not positive why this was enabled in the first place).

#trivial

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
   },
   "dependencies": {
     "@artsy/cohesion": "1.3.0",
-    "@artsy/palette": "9.13.0",
+    "@artsy/palette": "10.0.0",
     "@mapbox/react-native-mapbox-gl": "6.1.3",
     "@react-native-community/async-storage": "1.6.3",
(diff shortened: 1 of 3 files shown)


ds300:
> what the appropriate Q/A process for something like this would be.

I think just a quick smoke test of all major react-native screens should do the trick. Might need to pay particular attention to interactive components that come from palette like Button.

> keyofStringsOnly

no knowledge of why this was true, it probably broke something in emission when TS changed the default behaviour back in the day.

dzucconi:
Everything looks good to me. Updated the PR to point to the new version. Should be good to go.

PR #3400, opened 2020-06-03
Add a bridge to identify the currently selected tab name

pepopowitz:
In #3389, I demonstrated a proof of concept on how I'd go about identifying the currently selected tab name from a React component. This PR cleans up that implementation, and is limited to setting up the bridge.

A followup PR will take advantage of this new bridge function to modify user flow within the sell tab landing page.

### DIFF
--- a/Artsy/View_Controllers/App_Navigation/ARTopMenuNavigationDataSource.m
+++ b/Artsy/View_Controllers/App_Navigation/ARTopMenuNavigationDataSource.m
@@ -219,6 +219,27 @@
     return [self.tabOrder indexOfObject:@(tabType)];
 }
 
+- (NSString *)tabNameForIndex:(NSInteger)index
+{
+    ARTopTabControllerTabType tab = [self tabTypeForIndex:index];
+    switch (tab) {
+        case ARHomeTab:
+            return @"ARHomeTab";
+        case ARSearchTab:
+            return @"ARSearchTab";
+        case ARMessagingTab:
+            return @"ARMessagingTab";
+        case ARLocalDiscoveryTab:
+            return @"ARLocalDiscoveryTab";
+        case ARFavoritesTab:
+            return @"ARFavoritesTab";
+        case ARSalesTab:
+            return @"ARSalesTab";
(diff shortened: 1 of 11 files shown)


src/lib/NativeModules/TopMenu.tsx
```diff
@@ -0,0 +1,7 @@
+import { NativeModules } from "react-native"
+
+const { ARTopMenuModule } = NativeModules
+
+export function getSelectedTabName() {
```
pepopowitz, line 5: I'm not totally sure about this extra layer of abstraction. I was following the pattern of the `ARSwitchBoardModule`, but as this is really just a pass-through, it feels unnecessary. I could also see it becoming a custom hook as I work through the follow-up PR that's dedicated to my specific navigation issue.

Artsy/View_Controllers/App_Navigation/ARTopMenuNavigationDataSource.m
```diff
@@ -217,6 +219,27 @@ - (NSUInteger)indexForTabType:(ARTopTabControllerTabType)tabType
     return [self.tabOrder indexOfObject:@(tabType)];
 }
 
+- (NSString *)tabNameForIndex:(NSInteger)index
+{
+    ARTopTabControllerTabType tab = [self tabTypeForIndex:index];
```
pepopowitz, line 224: @pvinis I think you mentioned that you had a trick for emitting the name of an enum based on its value. I didn't want to spend a lot of time here, so I hardcoded it; happy to update if someone's got a better approach.

src/lib/NativeModules/TopMenu.tsx
pepopowitz: And yeah, [that's how it turned out](https://github.com/artsy/eigen/pull/3402/files#diff-1a95080af2a04563f97f830b674e87b5) - it holds my custom hook.

ds300:
talked to @pepopowitz on slack and we decided to refactor this as follow-up to unblock other things :+1:

PR #3403, opened 2020-06-03
[MX-189] - Fix for auctions pointy modal corner and non-fullscreen display

brainbicycle:
https://artsyproduct.atlassian.net/browse/MX-189

- Fixes pointy corner of modals in auctions flows
- Makes live auctions views presentation full screen on iPad + iOS 13

While we were working on the pointy corner Damon noticed that the LiveAuctions VC was not being presented correctly on iPad, this is due to the new card presentation style for modals in iOS 13. Fix is to add an exception for this view controller in our presentation code. There may be more layout issues associated with our native modals on iOS 13 that might be worth looking into. 

**Co-authored with @dzucconi** 

**Before**
pointyModal

notFullScreenLiveAuction

**After**
fixedPointyModal

fullScreenLiveAuction

### DIFF
--- a/Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.h
+++ b/Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.h
@@ -72,8 +72,6 @@ NS_ASSUME_NONNULL_BEGIN
 /// Not all view controllers should be presented as a push, use this to determine whether the topVC will present modally or push.
 + (BOOL)shouldPresentViewControllerAsModal:(UIViewController *)viewController;
 
-- (BOOL)isShowingStatusBar;
-
 /// Used by analytics to get the tab name for a particular index
 - (NSString *)descriptionForNavIndex:(NSInteger)index;
 
--- a/Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
+++ b/Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
@@ -461,13 +461,7 @@ static ARTopMenuViewController *_sharedManager = nil;
 }
 
 + (BOOL)shouldPresentModalFullScreen:(UIViewController *)viewController {
-    NSArray *fullScreenClasses = @[LiveAuctionViewController.class];
-    for (Class klass in fullScreenClasses) {
-        if ([viewController isKindOfClass:klass]) {
-            return YES;
-        }
-    }
-    return NO;
+    return [viewController isKindOfClass:LiveAuctionViewController.class];
 }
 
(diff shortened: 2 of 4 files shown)


Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.h
```diff
@@ -72,8 +72,6 @@ NS_ASSUME_NONNULL_BEGIN
 /// Not all view controllers should be presented as a push, use this to determine whether the topVC will present modally or push.
 + (BOOL)shouldPresentViewControllerAsModal:(UIViewController *)viewController;
 
-- (BOOL)isShowingStatusBar;
```
brainbicycle, line 75: This method was removed recently, removed the declaration in header which was causing a warning

Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
```diff
@@ -478,7 +488,9 @@ - (void)pushViewController:(UIViewController *)viewController animated:(BOOL)ani
             } else if ([viewController isKindOfClass:UINavigationController.class] && [[(UINavigationController *)viewController topViewController] isKindOfClass:ARShowConsignmentsFlowViewController.class]) {
                 // Consignments gets full screen
                 viewController.modalPresentationStyle = UIModalPresentationFullScreen;
-            }
+            } else if ([self.class shouldPresentModalFullScreen:viewController]) {
+                viewController.modalPresentationStyle = UIModalPresentationFullScreen;
+            } else {}
```
ds300, line 493: nonblocking

```suggestion
            }
```

Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
```diff
@@ -460,6 +460,16 @@ + (BOOL)shouldPresentViewControllerAsModal:(UIViewController *)viewController
...
 
++ (BOOL)shouldPresentModalFullScreen:(UIViewController *)viewController {
+    NSArray *fullScreenClasses = @[LiveAuctionViewController.class];
+    for (Class klass in fullScreenClasses) {
+        if ([viewController isKindOfClass:klass]) {
+            return YES;
+        }
+    }
+    return NO;
+}
```
ds300, line 471: nonblocking: do we expect this list to get bigger? the method body could be just `return [viewController isKindOfClass:LiveAuctionViewController.class]` for now at least.

brainbicycle: Yeah this is probably an overly complicated implementation, I will update 👍

PR #3408, opened 2020-06-04
[MX-338] Rate limit Relay middleware

dzucconi:
I upgraded `react-relay-network-modern` to get the typings. The only relevant breaking change was `RelayNetworkLayer.fetchFn now returns Observable instead of Promise` in 3.0.0, which, *I believe*, shouldn't actually be a problem here (?). The return type for the `cacheMiddleware` isn't correct but seems to work fine, so I've ignored it, but I can try to correct things if desired. The 4.0.0 breaking change is related to the `batchMiddleware` which isn't used.

Nothing particularly fancy here: fixed window limit and just throws an error/logs to Sentry. If I'm missing something for how we want to actually surface this error let me know.

I checked out the repo prior to the fix to run this against what was causing the issue in the first place:
http://static.damonzucconi.com/_capture/81b6FS49Wj4G.gif — so this is what that would have looked like in development.


src/lib/relay/middlewares/rateLimitMiddleware.ts
```diff
@@ -0,0 +1,42 @@
...
+  let prevTimeElapsed = Date.now()
+
+  return next => {
+    return req => {
+      const timeElapsed = Date.now() - prevTimeElapsed
+      const isWithinInterval = timeElapsed < interval
+      const isOutsideInterval = timeElapsed >= interval
+
+      if (isWithinInterval && count >= limit) {
+        captureMessage("Rate limit exceeded")
```
ds300, line 26: could we add the operation name to this? might help in case we ever see this in production. (see timingMiddleware.ts for an example of how to get it)

src/lib/relay/middlewares/rateLimitMiddleware.ts
```diff
@@ -0,0 +1,42 @@
...
+        throw new Error("Rate limit exceeded")
```
dzucconi, line 27: @ds300 I know the ticket mentioned a distinction between dev/prod environments — how does the app respond to an error thrown in prod? And how does one distinguish between the two environments in code?

dzucconi: Follow up: is the explicit `captureMessage` redundant? Are any errors thrown automatically captured by Sentry anyway?

ds300: I'm pretty sure sentry is smart enough to collapse these into one event, but I'll double-check when I review this PR again on monday.

ds300: Yep all good, it only triggered one event https://sentry.io/organizations/artsynet/issues/1715075469/?project=166818&query=is%3Aunresolved

ds300:
Great work on this, thanks again 🙇

PR #3412, opened 2020-06-05
Typography + Layout fixes Part 1

ds300:
@willanderson @samchieng and I went through the app trying to quickly address typography issues and get rid of Garamond. We're keeping track in a notion doc over here: https://www.notion.so/artsy/iOS-Typography-Audit-8313d1d8ae1140f6addfc5058f37b89f

This PR facelifts the Artist page to be all-sans, and fixes some layout issues. I QA'd the layout changes with @samchieng already.

# iPad

image
image
image
image

# iPhone

image
image
image
image
image

#skip_new_tests


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>Missing Test Files:

- `src/lib/Components/__tests__/Stack-tests.tsx`

    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/Components/Stack.tsx
```diff
@@ -0,0 +1,12 @@
...
+
+export const Stack: React.FC<{ spacing?: SpacingUnit; horizontal?: boolean } & React.ComponentPropsWithoutRef<
+  typeof Flex
+>> = ({ children, spacing = 2, horizontal, ...others }) => {
+  return (
+    <Flex flexDirection={horizontal ? "row" : "column"} {...others}>
+      <Join separator={<Spacer mb={horizontal ? 0 : spacing} mr={horizontal ? spacing : 0} />}>{children}</Join>
+    </Flex>
+  )
+}
```
dzucconi, line 12: 👍 we should extract this to palette as well

dzucconi: I don't love the whole spacer pattern because of how it adds extraneous elements — but in our case we don't uniformly support `space` on everything, and the alternative without the extraneous elements is cloneElement and manipulating space props — so this makes sense to me.

src/lib/Components/Artist/ArtistShows/ArtistShows.tsx
```diff
@@ -14,30 +14,33 @@ interface Props {
 
 class Shows extends React.Component<Props> {
   render() {
+    const currentShows = this.props.artist.currentShows?.edges?.map(edge => edge?.node!) ?? []
+    const upcomingShows = this.props.artist.upcomingShows?.edges?.map(edge => edge?.node!) ?? []
+    const currentAndUpcomingShows = [...currentShows, ...upcomingShows]
+
+    const pastLargeShows = this.props.artist.pastLargeShows?.edges?.map(edge => edge?.node!) ?? []
+    const pastSmallShows = this.props.artist.pastSmallShows?.edges?.map(edge => edge?.node!) ?? []
```
dzucconi, line 22: The nullable connection typings are so annoying.

src/lib/AppRegistry.tsx
```diff
@@ -252,9 +253,11 @@ const InnerPageWrapper: React.FC<PageWrapperProps> = ({ children, fullBleed }) =
 class PageWrapper extends React.Component<PageWrapperProps> {
   render() {
     return (
-      <ProvideScreenDimensions>
-        <InnerPageWrapper {...this.props} />
-      </ProvideScreenDimensions>
+      <Theme>
```
dzucconi, line 256: How was this not wrapped with `<Theme>` before?

dzucconi: I guess my question is why so many screens are individually wrapping their contents with `<Theme>`?

ds300: we only added this wrapper recently, page setup was totally ad-hoc before. I wanted to add the `<Theme>` wrapper at the time but punted on it because @ashfurrow seemed excited about being the one to get to do it.

In this PR I needed to add Theme to a page so I figured I might as well put it here. I left the cleanup for another PR since we already have a Jira ticket for it and it's not totally straightforward thanks to tests.

src/lib/Components/Stack.tsx
ds300: 😄  `<Spacer />` was one of the things I was weary of when I joined Artsy because it's not semantic and I was kinda proud of being good at clean margin organisation.

Now I love it. So convenient. Plus easy to read, debug, and use in conditional contexts.

PR #3414, opened 2020-06-05
Replace sales flag with echo flag, fixes for lab flags, see details

brainbicycle:
- Replaces sales lab flag with echo flag
- Remove exists check preventing lab flags from overriding echo flags
- Rename sales flag to workaround override issue
- Update tests to use echo flag

See slack thread for more context around this PR:
https://artsy.slack.com/archives/C02BAQ5K7/p1591290911223300

We were looking to replace the sales lab feature flag with an echo flag of the same name as recommended by the [docs here](https://github.com/artsy/eigen/blob/master/docs/developing_a_feature.mdl) and noticed the lab flags were not overriding the echo flags as expected. When we load the feature flags for emission [we load the lab features last](https://github.com/artsy/eigen/blob/master/Artsy/App/ARAppDelegate+Emission.m#L322) to allow them to override any set echo features in order to not present in progress work in old versions of the app. This should work if the lab flags defaulted to _false_ however they default to _empty_ so the echo flag are currently taking precedence. 

This pr implements workaround recommended by David to rename the flag and update the native code to use an echo flag. It also [removes an existence check on lab flags](https://github.com/artsy/eigen/pull/3414/files#diff-fa3274520df88be962c7afc891c3b5edL72) that is preventing them from defaulting to false.

**Follow-up**:
- Update the docs with recommended way of enabling production features, example of native changes


Artsy/App/ARSwitchboard+Eigen.m
```diff
@@ -68,7 +68,7 @@ - (ARArtworkComponentViewController *)loadArtworkWithID:(NSString *)artworkID in
 
 - (UIViewController *)loadAuctionWithID:(NSString *)saleID
 {
-    if (self.echo.features[@"DisableNativeAuctions"]) {
+    if (self.echo.features[@"DisableNativeAuctions"].state) {
```
ds300, line 71: nonblocking: could this now be

```suggestion
    if ([self isFeatureEnabled:@"DisableNativeAuctions"]) {
```

?

brainbicycle: ahh yes it definitely could be

PR #3424, opened 2020-06-09
[CSGN-232] Remove horizontal scrollbar from sell tab artist list & update "submission failed" content

pepopowitz:
Addresses https://artsyproduct.atlassian.net/browse/CSGN-232

The horizontal scrollbar was appearing over the bottom artists on the sell tab artist list. This PR removes the horizontal scrollbar from that rail entirely. 

It now looks like this: 

scroll-without

Also included is a small change to some content on the "submission failed" view.

### DIFF
--- a/src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/Components/ArtistList.tsx
+++ b/src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/Components/ArtistList.tsx
@@ -5,7 +5,7 @@ import SwitchBoard from "lib/NativeModules/SwitchBoard"
 import { PlaceholderBox, PlaceholderText } from "lib/utils/placeholders"
 import { chunk, shuffle } from "lodash"
 import React, { useRef } from "react"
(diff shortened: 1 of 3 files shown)


src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/Components/ArtistList.tsx
```diff
@@ -36,26 +36,25 @@ export const ArtistList: React.FC<ArtistListProps> = ({ targetSupply, isLoading
 
         <Spacer mb={2} />
 
-        <ScrollView horizontal>
```
jonallured, line 39: Cool, so if I'm reading this correctly all you had to do was remove the wrapping ScrollView and then the bars went away?

pepopowitz: I also [added a prop to the `FlatList`](https://github.com/artsy/eigen/pull/3424/files#diff-2f28b153a622c0444cdbde38c8d6fd46R44).

PR #3426, opened 2020-06-09
[CSGN-233] Scroll sales tab to top when already on sales tab and icon is tapped

pepopowitz:
Addresses https://artsyproduct.atlassian.net/browse/CSGN-233

This PR adds the sales tab to the list of tabs that should scroll to top when the user is already on the tab and taps its icon. 

scroll-to-top

### DIFF
--- a/Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
+++ b/Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
@@ -628,6 +628,7 @@ static ARTopMenuViewController *_sharedManager = nil;
         // Otherwise find the first scrollview and pop to top
         else if (tabType == ARHomeTab ||
                  tabType == ARMessagingTab ||
+                 tabType == ARSalesTab ||
                  tabType == ARFavoritesTab) {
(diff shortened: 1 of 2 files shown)


Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
```diff
@@ -628,6 +628,7 @@ - (BOOL)tabContentView:(ARTabContentView *)tabContentView shouldChangeToIndex:(N
         // Otherwise find the first scrollview and pop to top
         else if (tabType == ARHomeTab ||
                  tabType == ARMessagingTab ||
+                 tabType == ARSalesTab ||
```
pepopowitz, line 631: Might it make more sense for us to use an exclusion list instead of inclusion at this point?

ashfurrow: Yeah it might, but that might have implications we need to test against and _hopefully_ MX will be moving this whole thing to React Native soon anyway 🤞

jonallured:
LGTM 🏆

PR #3427, opened 2020-06-09
[MX-319] Support principalField extension

dzucconi:
Re: [MX-319](https://artsyproduct.atlassian.net/browse/MX-319)

So: I didn't put `principalField` on everything (yet). To some extent it might not actually matter for us here in a non-web context.

The general idea is that you turn off Relay's default behavior [to throw](https://github.com/relay-tools/react-relay-network-modern/blob/488944f89f40639e9c87319363709c68f366e312/src/fetchWithMiddleware.js#L52-L54) whenever the errors array is present, and just handle errors however you want in the application code. Which makes sense, theoretically if our types are correct/strict then the components should handle any nulls correctly and just render whatever is available. In practice all of the non-strict `@ts-ignore`s are where problems are going to crop up. This makes sense anyway since the application would have thrown regardless, so there's the chance that the component can safely render it and if it errors, then it was going to error anyway.

In the app everything is always going to be linked internally, so the notion of a 404 doesn't really exist — I suppose this can happen if somehow some embedded data got out of sync, but I don't know if that's a real risk (?). So I don't know how critical it is to actually turn on principalField everywhere, though it's probably not a bad idea.

An example:

In b4b6b8273b71a2cb81e80a6a0393ac19cb925e5a I just updated the component to render if the `fair#show#partner` errors and is `null`. What that change amounts to is just removing a `@ts-ignore` and fixing the type errors. So the component renders without the partner name on the fair page, but if you tap it, it doesn't have the partner so the whole thing errors.

I'm thinking a follow-up here would be just removing `@ts-ignore`s where obvious/possible and enabling `principalField` on the top-level fields for the scene queries.

Thoughts?


src/lib/relay/createEnvironment.ts
```diff
@@ -14,27 +15,33 @@ const Constants = NativeModules.ARCocoaConstantsModule
...
-    errorMiddleware({
-      disableServerMiddlewareTip: true,
-    }),
-    metaphysicsExtensionsLoggerMiddleware(),
-    timingMiddleware(),
-  ])
+  const network = new RelayNetworkLayer(
+    [
+      // @ts-ignore
+      cacheMiddleware(),
```
dzucconi, line 21: It's probably worth it to fix the typings in this cache middleware, I noticed at least one error where the id for the cache key is possibly null.

src/lib/relay/createEnvironment.ts
```diff
@@ -14,27 +15,33 @@ const Constants = NativeModules.ARCocoaConstantsModule
...
+      loggerMiddleware(),
+      errorMiddleware({
+        disableServerMiddlewareTip: true,
+      }),
+      metaphysicsExtensionsLoggerMiddleware(),
+      timingMiddleware(),
+    ],
+    // `noThrow` is currently marked as "experimental" and may be deprecated in the future.
+    // See: https://github.com/relay-tools/react-relay-network-modern#advanced-options-2nd-argument-after-middlewares
+    { noThrow: true }
```
dzucconi, line 43: So we were throwing in two places: our own cacheMiddlware (which probably shouldn't have been in the first place) and the default Relay fetch, which this option disables. Not sure why this is 'experimental' as this seems like a pretty reasonable thing to disable.

dzucconi: Force does this as well: https://github.com/artsy/force/blob/d17d03ce40fc5fc9f418b2ed7120e4a77f0d4706/src/v2/Artsy/Relay/createRelaySSREnvironment.ts#L125

ashfurrow:
Okay, I think we're good here. I've resolved the conflicts and we can roll this approach out to other components going forward. #mergeongreen

PR #3434, opened 2020-06-10
[MyCollections] Stub out app structure

damassi:
This stubs out the file structure for for MyCollections and adds a bare-bones [easy-peasy](https://easy-peasy.now.sh/) state management setup. 

Note that there's a `Scenes/MyCollection` folder - this will get migrated in time. For now grouping all consignments related stuff together. 

#trivial 
#skip_new_tests

### DIFF
--- a/fastlane/metadata/en-US/release_notes.txt
+++ b/fastlane/metadata/en-US/release_notes.txt
@@ -1,8 +1,6 @@
 This update includes a number of improvements and bug fixes, including:
 
-• Collections can now be filtered by ways to buy.
-• Fixes a problem with displaying certain fair booths.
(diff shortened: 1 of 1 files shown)


ds300:
:ceiling-cat: going forward we'd love to avoid having index.ts(x) files since they generally make it harder to find and navigate code. I'd appreciate if you could rename these from, e.g. `MyCollectionAddWork/index.tsx` to `MyCollectionAddWork/MyCollectionAddWork.tsx` I know it seems like redundancy in the path but it really pays off when maintaining a big code base.

damassi:
will do @ds300 👍

PR #3402, opened 2020-06-03
[CSGN-225] Choose the destination of artist consign taps based on your active tab

pepopowitz:
This work depends on #3400. The only commits in this diff that are related to this PR are the last two.

Addresses https://artsyproduct.atlassian.net/browse/CSGN-225

This PR uses your currently selected tab to determine where to send you when you tap the artist consign button. 

For users who are on the home or search tab (or anywhere that isn't sell), we'll continue to send them to the sell tab landing page:

artist-consign-1

If you're already in the sell tab, we push another landing page view onto your nav stack. This addresses the experience of tapping on something and not getting enough feedback about where you're going:

artist-consign-2

#trivial (but not really; it's just covered by another entry that's already in the changelog.)

PR #3437, opened 2020-06-10
[CSGN-230] Re-architect selected tab bridge

pepopowitz:
[The original implementation](https://github.com/artsy/eigen/pull/3400) of the bridge that allows React components to identify the currently selected tab [had a "dead zone" where for a moment in time the selected tab would not be accurate](https://github.com/artsy/eigen/pull/3402#discussion_r435310437). 

This PR implements the suggestion in that comment, and re-architects the bridge to use a "push" mechanism from objective-c to react, instead of react "pulling" the info from objective-c.

There should be no visible changes to the user.


emission/Pod/Classes/Core/ARSelectedTab.m
```diff
@@ -0,0 +1,87 @@
+#import "ARSelectedTab.h"
+
+@interface ARSelectedTab ()
+@property (nonatomic, assign, readwrite) BOOL isBeingObserved;
+@property (nonatomic, strong, readwrite) NSDictionary *selectedTab;
+@end
+
+static NSDictionary *buildSelectedTab(NSString *tabName)
+{
+    return @{
```
pepopowitz, line 10: I opted to use an object throughout the stack, instead of just a string, with the thought that it would be easier to extend if we had other info about tabs in the future. It might be YAGNI...

emission/Pod/Classes/Core/ARSelectedTab.m
```diff
@@ -0,0 +1,87 @@
...
+        [sself sendEventWithName:@"selectedTabChanged" body:sself.selectedTab];
+      }
+    });
+}
+
+-(instancetype)init
+{
+  self = [super init];
+  if (self) {
+    self.selectedTab = buildSelectedTab(@"ARHomeTab");
```
pepopowitz, line 57: Specifying the default `ARHomeTab` here doesn't feel great, but I wasn't sure how to hit an eigen view controller (`ARTopMenuViewController`) from an emission module to get the current value.

src/lib/NativeModules/SelectedTab.tsx
```diff
@@ -0,0 +1,57 @@
...
+  name: string
+}
+
+const changes = new NativeEventEmitter(NativeModules.ARSelectedTab)
+
+let initialTab: SelectedTab = {
+  name: NativeModules.ARSelectedTab.name,
+}
+
+changes.addListener("selectedTabChanged", nextTab => {
```
pepopowitz, line 25: You might know this already from the `ARDynamicScreenDimensions` event emitter, on which this code is all based, but this listener picks up any changes when the current view is not on the visible tab. I'm having a hard time putting into words why the other handlers don't fully take care of the non-visible tabs, but if pressed I can pair and demonstrate.

Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
```diff
@@ -609,6 +609,12 @@ - (void)returnToPreviousTab
 - (void)tabContentView:(ARTabContentView *)tabContentView didChangeSelectedIndex:(NSInteger)index
 {
     self.selectedTabIndex = index;
+
+    NSString * selectedTabName = [self.navigationDataSource tabNameForIndex:self.selectedTabIndex];
+
+    [[NSNotificationCenter defaultCenter] postNotificationName:@"ARSelectedTabChangedNotification"
+                      object:self
+                      userInfo:@{@"tabName": selectedTabName }];
```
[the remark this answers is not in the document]
pepopowitz: Yes!

emission/Pod/Classes/Core/ARSelectedTab.m
ashfurrow: This makes sense – I feel like a broken record at this point, but we're looking to move the tab navigation out of Objective-C and into React Native really soon. Consequently, we should avoid any major investments in that existing infrastructure. So this change makes sense to me as-is 👍

src/setupJest.ts
```diff
@@ -59,6 +59,12 @@ jest.mock("./lib/NativeModules/Events.tsx", () => ({
   userHadMeaningfulInteraction: jest.fn(),
 }))
 
+jest.mock("./lib/NativeModules/SelectedTab/SelectedTab.tsx", () => ({
```
pepopowitz, line 62: This fixes the currently broken test, which was failing because of the [initialization logic in `SelectedTab.tsx`](https://github.com/artsy/eigen/pull/3437/files?file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx&hide-deleted-files=true#diff-88991225e78b4a570260c1186b78cef3R9). 

The mocking happens here in `setupJest` because of the immediate execution of that initialization logic. If we didn't do this in `setupJest`, the immediate execution would fire (and break) before the mocking happened in an individual test.

src/lib/NativeModules/SelectedTab/TabName.ts
```diff
@@ -0,0 +1,9 @@
+export enum TabName {
```
pepopowitz, line 1: I extracted this to its own file to avoid having to mock it.

src/lib/Components/Artist/__tests__/ArtistConsignButton-tests.tsx
```diff
@@ -233,4 +241,57 @@ describe("ArtistConsignButton", () => {
...
+        internalID: "fooBarBaz",
+        slug: "alex-katz",
+        name: "Alex Katz",
+        href: "/artist/alex-katz",
+        id: "QXJ0aXN0OjRkOGQxMjBjODc2YzY5N2FlMTAwMDA0Ng==",
+      },
+    }
+
+    beforeEach(() => {
+      NativeModules.Emission.options.AROptionsEnableSales = true
```
pepopowitz, line 261: I only added tests for when the flag is enabled, since no one should experience the old flow (unless something goes really wrong and we turn the echo flag off).

src/lib/Components/Artist/ArtistConsignButton.tsx
```diff
@@ -6,8 +6,9 @@ import { useTracking } from "react-tracking"
 import styled from "styled-components/native"
 
 import { ArtistConsignButton_artist } from "__generated__/ArtistConsignButton_artist.graphql"
+import { useSelectedTab } from "lib/NativeModules/SelectedTab/SelectedTab"
+import { TabName } from "lib/NativeModules/SelectedTab/TabName"
```
jonallured, line 10: #minor but I think it would be cool if these two were exported in tandem so they could be imported like so:

```ts
import { TabName, useSelectedTab } from "lib/NativeModules/SelectedTab"
```

What do you think?

src/lib/NativeModules/SelectedTab/SelectedTab.tsx
```diff
@@ -0,0 +1,47 @@
...
+    name: nextTab.name,
+  }
+})
+
+export const SelectedTabContext = createContext(initialTab)
+
+export const ProvideSelectedTab: React.FC = ({ children }) => {
+  const [selectedTab, setSelectedTab] = useState(initialTab)
+
+  const onChange = useCallback(
```
jonallured, line 26: I hadn't run into this hook before so just dropping a link to the docs in case anybody else is unfamiliar:

https://reactjs.org/docs/hooks-reference.html#usecallback

src/lib/Components/Artist/ArtistConsignButton.tsx
```diff
@@ -18,7 +19,7 @@ export interface ArtistConsignButtonProps {
 export const ArtistConsignButton: React.FC<ArtistConsignButtonProps> = props => {
   const tracking = useTracking()
   const buttonRef = useRef(null)
-  const selectedTabName = useSelectedTabName()
+  const selectedTabName = useSelectedTab()
```
jonallured, line 22: #minor but it might be cool to see this more like:

```ts
const selectedTab = useSelectedTab()
// ...then later
selectedTab.name === TabName.ARSalesTab
```

That way it's more clear that the tab is the object and the name is the property we're comparing, ya know?

src/lib/Components/Artist/ArtistConsignButton.tsx
pepopowitz: I'm not sure what you're specifically suggesting. I [separated them intentionally, to avoid having to mock `TabName`](https://github.com/artsy/eigen/pull/3437#discussion_r439455142) (and thus having unnecessary duplication in the test setup) - I don't want them combined into the same file. If instead you mean we should use a barrel and export `TabName` and `useSelectedTab` from `lib/NativeModules/SelectedTab/index.tsx`, I got the impression yesterday in slack that @artsy/mobile-experience is not interested in that practice.

src/lib/Components/Artist/ArtistConsignButton.tsx
pepopowitz: Yeah, this is me missing a name update while floundering through the work. I will update this name.

pepopowitz: Updated!

src/lib/NativeModules/SelectedTab/TabName.ts
ds300: i would be tempted to delete this file and alter the typings in SelectedTab.tsx to use a string literal union.

```ts
type TabName = "ARHomeTab" | "ARSearchTab" | ...
interface SelectedTab {
  name: TabName
}

// typescript will complain if you try to do useSelectedTab().name === "ARNotAValidTabName"
// plus intellisense works without having to import the TabName enum
```


also, given the use case for this feature, for now I'd be tempted to only expose a hook which returns a boolean representing whether you're on a particular tab(s?), e.g.

```ts
export function useIsCurrentTabOneOf(tabNames: TabName[]) {
  return tabNames.includes(useContext(SelectedTabContext).name)
}
```

and then in react components you can be like

```ts
const isSalesTab = useIsCurrentTabOneOf(["ARSalesTab"])

// or, e.g.

const showConsignmentsCTA = !useIsCurrentTabOneOf(["ARSalesTab"])
```

I can't think of a situation where you'd need access to the actual tab name string in normal application code.

src/lib/NativeModules/SelectedTab/TabName.ts
```diff
@@ -0,0 +1,9 @@
...
+  ARHomeTab = "ARHomeTab",
+  ARSearchTab = "ARSearchTab",
+  ARMessagingTab = "ARMessagingTab",
+  ARLocalDiscoveryTab = "ARLocalDiscoveryTab",
+  ARFavoritesTab = "ARFavoritesTab",
+  ARSalesTab = "ARSalesTab",
+  Unknown = "Unknown",
```
ds300, line 8: after this refactor do we still need the `Unknown` option?

src/lib/Components/Artist/ArtistConsignButton.tsx
jonallured: Hrm, yeah I guess you're right about that - feel free to ignore this one!

src/lib/NativeModules/SelectedTab/TabName.ts
pepopowitz: No, thank you! I removed it.

PR #3438, opened 2020-06-11
[My Collection] Stub out skeleton nav flow 

damassi:
This gets us to a place where we can start building out static views by sketching in the bones of our navigation flow. Things **will** change. 

The flow: 

- Created primitive view controllers and routes for each view: `add-artwork`, `artwork-details`, `artwork-list`, `home`. These routes are basically placeholder stubs and aren't dynamic at all (we're not yet ready to start that phase) 
- Started establishing nav patterns for moving between various sections. (See [flow diagram here](https://www.figma.com/file/yDSHlbJVzBxvSUROD2bfRI/Sell-tab?node-id=1308%3A0). 
- Right now i'm just pushing view controllers so that we can easily move between sections / dismiss while we build out designs. In the future this will change. 
- In the future we'll be needing to check various conditions before proceeding / cancelling / saving, etc, etc, so a centralized navigation center is imperative, versus managing control flow at the component level. 

#### Next Steps
- With this PR in place we can now do a rough pass over the latest comps and get the rough visual feel in via static layouts 
- Once the rough static layouts are in we can start connecting the buttons to navigation actions as established in the patterns in this PR
- Then we can start encoding the business logic and really get to work 
- At some point we might find that some of the view controllers created in this PR are no longer necessary; at that point we can delete, etc. 

stub

#trivial 
#skip_new_tests

### DIFF
--- a/Artsy/App/ARSwitchBoard.m
+++ b/Artsy/App/ARSwitchBoard.m
@@ -28,32 +28,36 @@
 #import "AREigenViewingRoomComponentViewController.h"
 #import "ARFavoritesComponentViewController.h"
 
-#import <Emission/ARNewSubmissionFormComponentViewController.h>
-#import <Emission/ARShowConsignmentsFlowViewController.h>
-#import <Emission/ARSellTabLandingViewController.h>
-#import <Emission/ARFairComponentViewController.h>
-#import <Emission/ARFairBoothComponentViewController.h>
-#import <Emission/ARFairArtworksComponentViewController.h>
+#import <Emission/ARArtworkAttributionClassFAQViewController.h>
+#import <Emission/ARAuctionsComponentViewController.h>
+#import <Emission/ARCityBMWListComponentViewController.h>
+#import <Emission/ARCityFairListComponentViewController.h>
+#import <Emission/ARCitySavedListComponentViewController.h>
+#import <Emission/ARCitySectionListComponentViewController.h>
+#import <Emission/ARCollectionFullFeaturedArtistListComponentViewController.h>
 #import <Emission/ARFairArtistsComponentViewController.h>
+#import <Emission/ARFairComponentViewController.h>
+#import <Emission/ARFairComponentViewController.h>
 #import <Emission/ARFairExhibitorsComponentViewController.h>
+#import <Emission/ARFairMoreInfoComponentViewController.h>
+
+// Breadcrumb
+#import <Emission/ARMyCollectionAddArtworkComponentViewController.h>
+#import <Emission/ARMyCollectionArtworkDetailComponentViewController.h>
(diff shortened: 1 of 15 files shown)


Artsy/App/ARSwitchBoard.m
```diff
@@ -28,31 +28,38 @@
...
-#import <Emission/ARViewingRoomArtworksComponentViewController.h>
+#import <Emission/ARFairArtworksComponentViewController.h>
+#import <Emission/ARFairBMWArtActivationComponentViewController.h>
+#import <Emission/ARFairBoothComponentViewController.h>
 #import <Emission/ARFairComponentViewController.h>
-#import <Emission/ARShowArtworksComponentViewController.h>
-#import <Emission/ARShowArtistsComponentViewController.h>
-#import <Emission/ARShowMoreInfoComponentViewController.h>
+#import <Emission/ARFairComponentViewController.h>
+#import <Emission/ARFairExhibitorsComponentViewController.h>
```
damassi, line 44: Alpha

Artsy/App/ARSwitchBoard.m
```diff
@@ -294,20 +301,41 @@ - (void)updateRoutes
...
+        return [[ARMyCollectionArtworkDetailComponentViewController alloc] init];
+    }];
+
+    [self.routes addRoute:@"/my-collection/artwork-list" handler:JLRouteParams {
+        return [[ARMyCollectionArtworkListComponentViewController alloc] init];
+    }];
+
+    [self.routes addRoute:@"/my-collection/home" handler:JLRouteParams {
+        return [[ARMyCollectionHomeComponentViewController alloc] init];
+    }];
```
jonallured, line 325: I have thoughts on these routes. I did some sketching on this and went in a slightly more RESTful direction, like this:

```
/collections/my-collection/artworks/new => Artwork Form flow with brand new work
/collections/my-collection/artworks/:id/edit => Artwork Form flow with existing work
/collections/my-collection/artworks => list view
/collections/my-collection/artworks/:id => detail view
```

What do you think about something more along those lines?

damassi: Oh yeah, I have a note below this line about discussing this, but would like to save this discussion for a bit later when we start wiring up business logic. This PR is specifically in scope for phase 1 of the build out (per the tech plan), that's why I emphasized that things would change in the description.

damassi: Preliminary thought tho is that I'm down with everything _except_ prefixing `/my-collection` with `/collection` -- that seems a bit premature.

jonallured: Cool, sounds legit!

PR #3443, opened 2020-06-11
[MX-349] added placeholder to my profile screen

MounirDhahri:
Edit: Added placeholders to my profile screen

placeholder image


ds300:
We're trying to move away from that chunky loading spinner, replacing it with the 'skeleton' placeholder style that you see on Home, the artist page, the artwork page, etc. Here's the example from Home: https://github.com/artsy/eigen/blob/057cea7f9adfe5688d2bbc823a4632fdca2bab28/src/lib/Scenes/Home/Home.tsx#L241

It'd be great if you could create something which matched the vague structure of this new profile tab using those placeholder components. If you'd like a design to work from i could throw one together for you in the morning.

MounirDhahri:
Thanks for checking it up. I will adjust it now.

PR #3448, opened 2020-06-11
Updates readme image

ashfurrow:
I noticed earlier this week that our readme's image is really outdated. Most of the screens don't exist or are outdated, and we're missing an opportunity to show off all the cool new things we've built over the past four years, including collections and the new sell tab. Eigen is a popular open source repo and I want our readme to put its best foot forward 😄 

overview

Once I get the 👍 on this new image, I'll send a PR to [the sketch file](https://github.com/artsy/artsy.github.io/blob/source/_design/artsy_open_source.sketch) so that it's easier to update next time.


ds300:
Nice

PR #3454, opened 2020-06-12
Add Notable Works rail to Artist pages

ashleyjelks:
Kapture 2020-06-15 at 16 27 26
Screen Shot 2020-06-15 at 4 28 56 PM

https://artsyproduct.atlassian.net/browse/FX-1964

### DIFF
--- a/src/lib/Components/Artist/ArtistArtworks/ArtistNotableWorksRail.tsx
+++ b/src/lib/Components/Artist/ArtistArtworks/ArtistNotableWorksRail.tsx
@@ -1,7 +1,8 @@
-import { Box, Sans, Spacer } from "@artsy/palette"
+import { Box, Spacer } from "@artsy/palette"
 import { ArtistNotableWorksRail_artist } from "__generated__/ArtistNotableWorksRail_artist.graphql"
 import { AboveTheFoldFlatList } from "lib/Components/AboveTheFoldFlatList"
 import { ArtworkTileRailCard } from "lib/Components/ArtworkTileRail"
+import { SectionTitle } from "lib/Components/SectionTitle"
 import SwitchBoard from "lib/NativeModules/SwitchBoard"
 import React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
@@ -9,12 +10,11 @@ import styled from "styled-components/native"
 
 interface ArtistNotableWorksRailProps {
   artist: ArtistNotableWorksRail_artist
-  hasCollection: boolean
 }
 
-type NotableArtwork = ArtistNotableWorksRail_artist["filterArtworksConnection"]["edges"][0]
+type NotableArtwork = NonNullable<NonNullable<ArtistNotableWorksRail_artist["filterArtworksConnection"]>["edges"]>[0]
 
-const ArtistNotableWorksRail: React.FC<ArtistNotableWorksRailProps> = ({ artist, hasCollection }) => {
+const ArtistNotableWorksRail: React.FC<ArtistNotableWorksRailProps> = ({ artist }) => {
   const artworks = artist?.filterArtworksConnection?.edges ?? []
 
   if (!artist || artworks.length <= 2) {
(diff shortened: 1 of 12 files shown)


src/lib/Components/Artist/ArtistArtworks/ArtistNotableWorksRail.tsx
```diff
@@ -0,0 +1,127 @@
...
+const ArtistNotableWorksRail: React.FC<ArtistNotableWorksRailProps> = ({ artist, hasCollection }) => {
+  const artworks = artist?.filterArtworksConnection?.edges ?? []
+
+  if (!artist || artworks.length <= 2) {
+    return null
+  }
+
+  const navRef = React.useRef<any>()
+
+  const handleNavigation = (slug: NotableArtwork["filterArtworksConnection"]["edges"][0]["slug"]) => {
```
ashleyjelks, line 26: This could also just be a string type here

src/lib/Components/Artist/ArtistArtworks/ArtistNotableWorksRail.tsx
```diff
@@ -0,0 +1,127 @@
...
+import React from "react"
+import { createFragmentContainer, graphql } from "react-relay"
+import styled from "styled-components/native"
+
+interface ArtistNotableWorksRailProps {
+  artist: ArtistNotableWorksRail_artist
+  hasCollection: boolean
+}
+
+type NotableArtwork = ArtistNotableWorksRail_artist["filterArtworksConnection"]["edges"][0]
```
ashleyjelks, line 15: I have no idea why TS is complaining about this type error when the schema is  
```
artist : {
        filterArtworksConnection: {
          edges: [
               {node: ....}
          ]
     }
}
```

src/lib/Components/ArtworkTileRail/ArtworkTileRailCard.tsx
```diff
@@ -13,32 +14,43 @@ interface ArtworkTileRailCardProps {
...
+  useNormalFontWeight?: boolean | null
+  useLighterFont?: boolean | null
 }
 
 export const ArtworkTileRailCard: React.FC<ArtworkTileRailCardProps> = ({
   onPress,
   imageURL,
   artistNames,
   saleMessage,
+  useLargeImageSize,
```
ashleyjelks, line 27: I wanted to reuse the `ArtworkTileRailCard ` component but the design spec differs slightly in the font weight and colors + the image height. I did as @nicoleyeo  to confirm that the artwork metadata font styling should be deviate from other uses of the artworks on tile rails. If the spec doesn't change (still waiting for her response) then I can add a test to this component confirming that the correct sizes/font styles are rendering when these optional props are/not passed.

src/lib/Components/Artist/ArtistArtworks/ArtistNotableWorksRail.tsx
```diff
@@ -0,0 +1,127 @@
...
+                onPress={() => {
+                  handleNavigation(item?.node?.slug)
+                }}
+              />
+            )
+          }}
+          keyExtractor={(item, index) => String(item.image?.internalID || index)}
+        />
+      </ArtistNotableWorksRailWrapper>
+      {!hasCollection && (
```
ashleyjelks, line 83: We need to show a header above the artwork grid when there isn't a an iconic collections rail below the Notable Artworks rail

ds300: do you think it would be possible to move this logic up into the parent component?

ashleyjelks: Ended up removing this title entirely!

src/lib/Components/Artist/ArtistArtworks/ArtistNotableWorksRail.tsx
ashleyjelks: That was it!

PR #3459, opened 2020-06-15
Bumps @artsy/palette => 11.0.0

dzucconi:
Re: https://github.com/artsy/palette/pull/707

Minor breaking change in this release where the `flexGrow/Shrink` props are a bit more strict.

#trival


dzucconi:
OK, weird this is just flat out wrong — investigating

```
Invariant Violation: Invalid prop `borderRadius` of type `string` supplied to `StyleSheet generated`, expected `number`.
    StyleSheet generated: {
      "borderRadius": "50%"
    }
```

dzucconi:
OH, it's the expectation that's wrong, not the type. I updated the avatar component to use a 50% borderRadius instead of a manual one.

dzucconi:
Ah, OK, apparently RN can't have a 50% borderRadius and it has to be a number. I'll release a patch and make sure that's commented :\

ashfurrow:
Ah weird! Thanks!

dzucconi:
K — should be good now!

PR #3464, opened 2020-06-16
[MyCollection] Add Add/Edit artwork patterns

damassi:
This sets down some very initial form patterns for MyCollections, as well as solves an endemically tricky React Native problem around multiple forms / contexts spread across screens and recomposed into a single form for submit. 

Flow: 
- Rather than using the `<Formik>` component we're going a bit lower and using the `useFormik` hook
- This returns the complete formik object that we can pass around from place to place
- We then execute an `initializeFormik` action and pass it our newly created `formik` form, which in turn stores it inside of our easy peasy store, which we can pass to different contexts, eliminating the screen boundary
- This gets around the general inability to _present_ a view controller or _push_ a screen and maintain a connection to Formik's context (or any context, for the matter), solving a bunch of problems in one go (check out old Consignments app multi-screen state patterns for an example of what I mean 😬) 
- Added two actions, `addArtwork` and `editArtwork`, which when called execute `commitMutation` with a payload. (This doesn't do anything ATM.) 
- Added example of a listener, where `addArtwork` action will in turn trigger proper navigation from list to detail per design's animation idea. This will get fleshed out. 
- So far, all of our components are effectively "dumb", with a nice separation of concerns. This is what we want!

#### Gravity 
In a parallel path we need to start working on the Gravity side of things, and get the add / edit code written and wired up to MP.

#### Eigen 
- The next PR will connect up more specific view navigation control states based on overall artwork form state -- locking, unlocking, cancelling, etc.
- After that I think we're in UI building territory (maybe the add artist widget?) while the backend work is being fleshed out. (Can also get the static Artwork Detail view going.)

add

#trivial 
#skip_new_tests


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>Missing Test Files:

- `src/lib/Scenes/Consignments/v2/Screens/MyCollectionAddArtwork/Mutations/__tests__/MyCollectionAddArtworkMutation-tests.ts`
- `src/lib/Scenes/Consignments/v2/Screens/MyCollectionAddArtwork/Mutations/__tests__/MyCollectionEditArtworkMutation-tests.ts`
- `src/lib/Scenes/Consignments/v2/Screens/MyCollectionAddArtwork/__tests__/formValidation-tests.tsx`
- `src/lib/Scenes/Consignments/v2/State/__tests__/artworkModel-tests.tsx`

    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/Scenes/Consignments/v2/State/navigationModel.tsx
```diff
@@ -1,33 +1,39 @@
...
+import { Action, action, ThunkOn, thunkOn } from "easy-peasy"
 import SwitchBoard from "lib/NativeModules/SwitchBoard"
 import { RefObject } from "react"
 import { NavigatorIOS } from "react-native"
 import { MyCollectionAddArtworkAddPhotos } from "../Screens/MyCollectionAddArtwork/Screens/MyCollectionAddArtworkAddPhotos"
 import { MyCollectionAddArtworkTitleAndYear } from "../Screens/MyCollectionAddArtwork/Screens/MyCollectionAddArtworkTitleAndYear"
+import { StoreModel } from "./store"
 
-export interface NavigationState {
+export interface NavigationModel {
```
damassi, line 9: This change just brings the language in line with easy-peasy idioms. A "model" contains state as well as actions, versus "state" (how I had it) which implies simply values.

src/lib/Scenes/Consignments/v2/State/artworkModel.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  addArtworkError: action(() => {
+    console.log("Add artwork error")
+  }),
+
+  /**
+   * Edit Artwork
+   */
+
+  editArtwork: thunk(async (actions, inputPayload) => {
```
pepopowitz, line 64:
```suggestion
  editArtwork: thunk(async (actions, input) => {
```

For consistency between this and the `addArtwork` action. I don't have a preference between `input` or `inputPayload`, but now's our chance to establish consistency.

src/lib/Scenes/Consignments/v2/State/artworkModel.tsx
```diff
@@ -0,0 +1,85 @@
...
+    try {
+      await commitMutation(defaultEnvironment, {
+        query: MyCollectionEditArtworkMutation, // FIXME: Add real mutation once we've completed Gravity API
+        variables: { input: inputPayload },
```
pepopowitz, line 68:
```suggestion
        variables: { input },
```

For consistency between this and the `addArtwork` action. I don't have a preference between `input` or `inputPayload`, but now's our chance to establish consistency.

src/lib/Scenes/Consignments/v2/State/artworkModel.tsx
```diff
@@ -0,0 +1,85 @@
...
+  addArtwork: thunk(async (actions, input) => {
+    try {
+      commitMutation(defaultEnvironment, {
+        query: MyCollectionAddArtworkMutation, // FIXME: Add real mutation once we've completed Gravity API
+        variables: { input },
+        onCompleted: actions.addArtworkComplete,
+        onError: actions.addArtworkError,
+      } as any) // FIXME: any
+    } catch (error) {
+      console.error("Error creating artwork", error)
```
pepopowitz, line 47:
```suggestion
      console.error("Error adding artwork", error)
```

very minor and non-blocking but the goal of this change would be parallel language between the error & the action.

src/lib/Scenes/Consignments/v2/State/artworkModel.tsx
```diff
@@ -0,0 +1,85 @@
...
+  editArtwork: thunk(async (actions, inputPayload) => {
+    try {
+      await commitMutation(defaultEnvironment, {
+        query: MyCollectionEditArtworkMutation, // FIXME: Add real mutation once we've completed Gravity API
+        variables: { input: inputPayload },
+        onCompleted: actions.editArtworkComplete,
+        onError: actions.editArtworkError,
+      } as any) // FIXME: any
+    } catch (error) {
+      console.error("Error creating artwork", error)
```
pepopowitz, line 73:
```suggestion
      console.error("Error editing artwork", error)
```

src/lib/Scenes/Consignments/v2/Screens/MyCollectionAddArtwork/MyCollectionAddArtwork.tsx
```diff
@@ -27,23 +49,31 @@ export const MyCollectionAddArtwork = () => {
         <Join separator={<Spacer my={1} />}>
           <Input
             title="Artist"
-            style={{ height: 50 }}
             placeholder="Search artists"
             icon={<SearchIcon width={18} height={18} />}
+            onChangeText={formik.handleChange("artist") as FormikHandlers["handleChange"]}
```
pepopowitz, line 54: I'm curious about the typing in this line and the next. Looking at the names of the functions, I would have expected them to be already of the type we're casting to (or something similar). Is this a thing we'll need to do in all form fields? Are there additional allowed types we would want to add to `Input#onChangeText` in palette (eventually)?

src/lib/Scenes/Consignments/v2/Screens/MyCollectionAddArtwork/Screens/MyCollectionAddArtworkAddPhotos.tsx
```diff
@@ -16,7 +16,7 @@ export const MyCollectionAddArtworkAddPhotos = () => {
           <Sans size="3">Photo list</Sans>
         </Box>
 
-        <Button block onPress={() => navigator?.pop()}>
+        <Button block onPress={goBack}>
```
pepopowitz, line 19: 👍 to this simple but very useful abstraction.

src/lib/Scenes/Consignments/v2/State/navigationModel.tsx
```diff
@@ -39,6 +45,31 @@ export const navigationState: NavigationState = {
...
 
+  goBack: action(state => {
+    state.navigator?.pop()
+  }),
+
+  /**
+   * Listeners
+   */
+
+  onArtworkAdded: thunkOn(
```
pepopowitz, line 56: Linking the docs for `thunkOn` (and other listeners) for people like me who don't know what they are - https://easy-peasy.now.sh/docs/api/listeners.html

TL;DR  from my understanding is "execute this work as a side-effect when the `storeActions.artwork.addArtwork` action executes"

src/lib/Scenes/Consignments/v2/Screens/MyCollectionAddArtwork/MyCollectionAddArtwork.tsx
damassi: Yeah, this is a tiny bit unfortunate. The typings for `useFormik` and the typings for `<Formik>` -- even though they basically do the same thing -- are subtly different; in the former there are a few more union cases (`undefined`) that the lib accounts for that don't overlap with what React Native expects, in which case I then cast. This is mostly an issue with Formik's type definition file.

src/lib/Scenes/Consignments/v2/State/navigationModel.tsx
damassi: ty!

src/lib/Scenes/Consignments/v2/Screens/MyCollectionAddArtwork/MyCollectionAddArtwork.tsx
```diff
@@ -3,10 +3,32 @@ import SearchIcon from "lib/Icons/SearchIcon"
...
+
+  const initialFormValues = {
+    artist: "Andy Warhol",
+    title: "Artwork title",
+    year: "1982",
+  }
+
+  const formik = useFormik<ArtworkFormValues>({
+    initialValues: initialFormValues,
+    initialErrors: validateAddArtworkForm(initialFormValues),
```
damassi, line 24: The idea here is that we'll eventually be executing this after a fetch, on edit artwork

src/lib/Scenes/Consignments/v2/Screens/MyCollectionAddArtwork/MyCollectionAddArtwork.tsx
jonallured: Ok so if I'm understanding correctly, the awkwardness with the types is not something we address in Palette with an update to the `onChangeText` signature, but instead see what can be done in Formik?

src/lib/Scenes/Consignments/v2/Screens/MyCollectionAddArtwork/MyCollectionAddArtwork.tsx
```diff
@@ -3,10 +3,32 @@ import SearchIcon from "lib/Icons/SearchIcon"
...
+    onSubmit: artworkActions.addArtwork,
+    validate: validateAddArtworkForm,
```
jonallured, line 26: Looks like we can move to a schema for validation by replacing this with `validationSchema`:

https://github.com/jaredpalmer/formik/blob/849ed497e24afbc1c9d41dbd0ad84db8231a48c2/packages/formik/src/types.tsx#L212

src/lib/Scenes/Consignments/v2/State/artworkModel.tsx
```diff
@@ -0,0 +1,85 @@
...
+  addArtworkComplete: Action<ArtworkModel>
+  addArtworkError: Action<ArtworkModel>
+
+  editArtwork: Thunk<ArtworkModel, ArtworkFormValues>
+  editArtworkComplete: Action<ArtworkModel>
+  editArtworkError: Action<ArtworkModel>
+}
+
+export const artworkModel: ArtworkModel = {
+  formik: null,
```
jonallured, line 28: It's a bit awkward that this starts out null - anything we can do about initializing this higher in the stack?

src/lib/Scenes/Consignments/v2/State/navigationModel.tsx
jonallured: Cool, thanks for updating this - that helps!

src/lib/Scenes/Consignments/v2/Screens/MyCollectionAddArtwork/MyCollectionAddArtwork.tsx
jonallured: We could do something similar with a Yup schema too. 👍

src/lib/Scenes/Consignments/v2/State/artworkModel.tsx
damassi: I think it might be possible via the [withFormk HOC](https://jaredpalmer.com/formik/docs/api/withFormik), but that's something i'd want to return to later. Doing it that way could lead to problems with the hook-forced on-change update cycle so would need to experiment a bit....

src/lib/Scenes/Consignments/v2/Screens/MyCollectionAddArtwork/MyCollectionAddArtwork.tsx
damassi: Yah, the problem is Formik's type inconsistency. I want to open an issue over there and ask them why there's a subtle difference, and if there's anything that can be done to make things consistent. The type-definitions def seem to be written by a few different people so it could very well just be a bug.

src/lib/Scenes/Consignments/v2/Screens/MyCollectionAddArtwork/MyCollectionAddArtwork.tsx
damassi: Ah yes, will get yup in the mix before merge 👍

src/lib/Scenes/Consignments/v2/Boot.tsx
```diff
@@ -18,21 +21,35 @@ export const setupMyCollectionScreen = (Component: React.ComponentType<any>) =>
...
   }> = props => {
     const navViewRef = useRef<View>(null)
-    const { setupNavigation } = useStoreActions(actions => actions.navigation)
+    const navigationActions = useStoreActions(actions => actions.navigation)
+    const artworkActions = useStoreActions(actions => actions.artwork)
+    const initialFormValues = useStoreState(state => state.artwork.formValues)
+
+    // FIXME: Don't initialize form for every collection screen; move this to another component
+    const initialForm = useFormik<ArtworkFormValues>({
+      enableReinitialize: true,
```
damassi, line 30: Very important, this lil guy

PR #3475, opened 2020-06-17
[MX-360] Change email address

MounirDhahri:
### Issue
https://artsyproduct.atlassian.net/browse/MX-360

ezgif-5-4514c413fcc2


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>Missing Test Files:

- `src/lib/Components/Modals/__tests__/LoadingModal-tests.tsx`

    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/Components/Modals/LoadingModal.tsx
```diff
@@ -0,0 +1,17 @@
...
+import React from "react"
+import { ActivityIndicator, Modal, ModalProps } from "react-native"
+
+interface LoadingModalProps {
+  isVisible: boolean
+}
+
+const LoadingModal: React.FC<LoadingModalProps & ModalProps> = ({ isVisible, ...rest }) => (
+  <Modal animationType="fade" visible={isVisible} {...(rest as any)} transparent>
+    <Flex flex={1} alignItems="center" justifyContent="center" style={{ backgroundColor: "rgba(0, 0, 0, 0.15)" }}>
```
MounirDhahri, line 11: I couldn't find the color in the list of available colors. maybe we should add it if we'll be using it more often

src/lib/Scenes/MyAccountEditEmail/MyAccountEditEmail.tsx
```diff
@@ -0,0 +1,168 @@
...
+  }, [])
+
+  const handleCancel = () => {
+    // Navigate the user back if nothing changed
+    if (email === me.email) {
+      return SwitchBoard.dismissNavigationViewController(navRef.current!)
+    }
+
+    // Prevent the user from navigating back if he started typing
+    return Alert.alert("Cancel", "Are you sure you want to cancel without saving", [
```
MounirDhahri, line 40: I am not sure about the content here

src/lib/Scenes/MyAccountEditEmail/MyAccountEditEmail.tsx
```diff
@@ -0,0 +1,168 @@
...
+})
+
+const MyAccountEditEmail: React.FC<{ me: MyAccountEditEmail_me; relay: RelayProp }> = ({ me, relay }) => {
+  const [email, setEmail] = useState<string>(me.email || "")
+  const [isSaving, setIsSaving] = useState<boolean>(false)
+  const navRef = useRef(null)
+
+  useEffect(() => {
+    setTimeout(() => {
+      NativeModules.ARSwitchBoardModule.updateShouldHideBackButton(true)
```
MounirDhahri, line 29: Please check this out

src/lib/Components/Modals/LoadingModal.tsx
ds300: i think it's safe to just use `black10` here

src/lib/Scenes/MyAccountEditEmail/MyAccountEditEmail.tsx
ds300: yeah... while this works, I think I know of a better way to achieve this but it involves updating some objective-C stuff. It's kinda involved so I'd be happy to pair on it before or after this gets merged, rather than write up the idea here. Unless you'd prefer to work on it alone?

src/lib/Scenes/MyAccountEditEmail/MyAccountEditEmail.tsx
ds300: I think we don't need this cancel dialog. We can check with sam when she gets back, but for now it seems not too valuable.

src/lib/Scenes/MyAccountEditEmail/MyAccountEditEmail.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+  const onCompletedSave = () => {
+    setIsSaving(false)
+    SwitchBoard.dismissNavigationViewController(navRef.current!)
+  }
+
+  const onErrorSave = () => {
+    InteractionManager.runAfterInteractions(() => {
+      setIsSaving(false)
+    })
```
ds300, line 63: interesting, what happens if you don't use InteractionManager here?

src/lib/Scenes/MyAccountEditEmail/MyAccountEditEmail.tsx
```diff
@@ -0,0 +1,168 @@
...
+          return Alert.alert("Please add a valid email address")
+        }
+        setIsSaving(true)
+        commitMutation<MyAccountEditEmailMutation>(relay.environment, {
+          onCompleted: onCompletedSave,
+          mutation: graphql`
+            mutation MyAccountEditEmailMutation($input: UpdateMyProfileInput!) {
+              updateMyUserProfile(input: $input) {
+                me {
+                  email
```
ds300, line 82: :+1:

src/lib/Scenes/MyAccountEditEmail/MyAccountEditEmail.tsx
```diff
@@ -0,0 +1,168 @@
...
+      })
+  }
+
+  const savingDisabled = Boolean(!email || email === me.email || isSaving)
+
+  return (
+    <Flex pt="2" ref={navRef}>
+      <LoadingModal isVisible={isSaving} />
+      <Join separator={<Separator my={2} />}>
+        <Flex flexDirection="row" justifyContent="space-between" px={2}>
```
ds300, line 103: using "space-between" results in an off-center title here because "Cancel" is wider than "Save". I think it would be better to figure out a way to keep the title in the center.

src/lib/Scenes/MyAccountEditEmail/MyAccountEditEmail.tsx
```diff
@@ -0,0 +1,168 @@
...
+          <TouchableOpacity onPress={handleCancel}>
+            <Sans size="4" weight="medium">
+              Cancel
+            </Sans>
+          </TouchableOpacity>
+
+          <Sans size="4" weight="medium">
+            Email
+          </Sans>
+          <TouchableOpacity disabled={savingDisabled} onPress={handleSave}>
```
ds300, line 113: this and the cancel button seem like their touch targets will be a bit small. maybe you could use the `hitSlop` prop to make them a bit wider?

src/lib/Scenes/MyAccountEditEmail/MyAccountEditEmail.tsx
```diff
@@ -0,0 +1,168 @@
...
+        </Flex>
+        <Flex px={2}>
+          <Input showClearButton value={email} onChangeText={setEmail} autoFocus={false} onSubmitEditing={handleSave} />
+        </Flex>
+      </Join>
+    </Flex>
+  )
+}
+
+const MyAccountEditEmailPlaceholder: React.FC<{}> = ({}) => {
```
ds300, line 127: 🙏

src/lib/Components/Modals/LoadingModal.tsx
MounirDhahri: If I will use black10 as a backgroundColor and set the opacity for 0.15, the activity indicator opacity will also inherit the opacity and will be almost invisible. 
Screenshot 2020-06-17 at 13 47 27

src/lib/Scenes/MyAccountEditEmail/MyAccountEditEmail.tsx
MounirDhahri: sure, I will remove it

src/lib/Scenes/MyAccountEditEmail/MyAccountEditEmail.tsx
MounirDhahri: yes of course

src/lib/Components/Modals/LoadingModal.tsx
ds300: ahhhh right. this is fine then, no need to add the color to palette.

src/lib/Scenes/MyAccountEditEmail/MyAccountEditEmail.tsx
MounirDhahri: nice catch. I will adjust it now

src/lib/Scenes/MyAccountEditEmail/MyAccountEditEmail.tsx
MounirDhahri: If the mutation fails before the saving modal finishes showing up, hiding it free will make the app freeze

src/lib/Scenes/MyAccountEditEmail/MyAccountEditEmail.tsx
MounirDhahri: yes sure, we can pair

PR #3479, opened 2020-06-17
[MyCollection] Add dirty form checking action sheet

damassi:
This adds some more form checking behavior, in this case an action sheet that will present if the add  artwork form is dirty: 

dismiss

#trivial
#skip_new_tests

### DIFF
--- a/src/lib/Scenes/Consignments/v2/State/artworkModel.tsx
+++ b/src/lib/Scenes/Consignments/v2/State/artworkModel.tsx
@@ -1,8 +1,11 @@
 import { Action, action, thunk, Thunk } from "easy-peasy"
 import { defaultEnvironment } from "lib/relay/createEnvironment"
+import { isEqual } from "lodash"
+import { ActionSheetIOS } from "react-native"
 import { commitMutation } from "react-relay"
 import { MyCollectionAddArtworkMutation } from "../Screens/MyCollectionAddArtwork/Mutations/MyCollectionAddArtworkMutation"
 import { MyCollectionEditArtworkMutation } from "../Screens/MyCollectionAddArtwork/Mutations/MyCollectionEditArtworkMutation"
+import { StoreModel } from "./store"
 
 export interface ArtworkFormValues {
   artist: string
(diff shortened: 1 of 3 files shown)


src/lib/Scenes/Consignments/v2/State/artworkModel.tsx
```diff
@@ -56,6 +62,30 @@ export const artworkModel: ArtworkModel = {
     }
   }),
 
+  addArtworkCancel: thunk((actions, _payload, { getState, getStoreActions }) => {
+    const navigationActions = getStoreActions().navigation
+    const formIsDirty = !isEqual(getState().formValues, initialFormValues)
```
damassi, line 67: Once we wire up the "Edit Artwork" functionality this will be revised and we'll tap into formik's `isDirty` -- meaning its changed from the pre-populated state. For now though this just checks to see if any input at all has been added.

src/lib/Scenes/Consignments/v2/State/artworkModel.tsx
```diff
@@ -56,6 +62,30 @@ export const artworkModel: ArtworkModel = {
...
+      ActionSheetIOS.showActionSheetWithOptions(
+        {
+          title: "You sure?",
+          options: ["Discard", "Keep editing"],
+          destructiveButtonIndex: 0,
+          cancelButtonIndex: 1,
+        },
+        buttonIndex => {
+          if (buttonIndex === 0) {
+            actions.setFormValues(initialFormValues)
```
jonallured, line 79: I'm surprised by this - why do we revert the form to initial values? This is the case where one has picked "Discard" so what I was expecting to see was:

* dismiss the action sheet
* trigger the dismiss on the overall form

Is that what `navigationActions.dismissModal` does? Maybe it'll do both or something? Anyway, just curious to know more about what's going on here! ❤️

damassi: When we discard we want to reset the form values in the store, then we dismiss the modal. Not sure i'm totally understanding the question; can you elaborate a bit?

jonallured: Yeah, let me try again!

> we want to reset the form values in the store

Why does this matter? The store won't do anything at this point, correct?

damassi: The store is the "brain" and sits above all behavior and crosses all screen boundaries (its secret sauce); it doesn't go anywhere for the duration that the app is in use. The source of truth for formik is the form values in the store, as they're used to repopulate the form when a view is pushed (or popped, etc); this is a tricky react native context limitation that I was able to solve.

pepopowitz: If I may add - my understanding is that without resetting the initial values, if you came back to submit for consignment a second time, you'd start with all the values from the previous submission. (Right?)

pepopowitz: disregard my comment if it adds confusion, I just saw that y'all sorted this thread out in slack.

damassi: That's correct @pepopowitz, we gotta clean up after ourselves.

PR #3483, opened 2020-06-18
[MX-384] Adds proof of concept for RN modals opting into iOS 13 modal styles

ashfurrow:
This is only a proof of concept. I wasn't sure where the modal you wanted to show with the new style were, so I just created a small component to show.

Screen Shot 2020-06-18 at 11 57 50

Here's a gif:

<details><summary>gif</summary>

</details>

I've tried to leave comments to be helpful. The key bits are the new `newModalStyle` function, the two `wantsNewModalStyle` uses (they are duplicated, I know it's not great but we're working on it 😅)  and `shouldPresentViewControllerAsModal:`. These would be the changes you'd need to pull in.

### DIFF
--- a/Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
+++ b/Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
@@ -30,6 +30,8 @@
 #import <FLKAutoLayout/UIViewController+FLKAutoLayout.h>
 #import <ObjectiveSugar/ObjectiveSugar.h>
 
+// Import is only to make the newModalStyle selector visible.
+#import <Emission/ARNewModalComponentViewController.h>
 #import <Emission/ARHomeComponentViewController.h>
 #import <Emission/ARInboxComponentViewController.h>
 #import <Emission/ARFavoritesComponentViewController.h>
@@ -459,7 +461,7 @@ static ARTopMenuViewController *_sharedManager = nil;
 
 + (BOOL)shouldPresentViewControllerAsModal:(UIViewController *)viewController
 {
-    NSArray *modalClasses = @[ UINavigationController.class, UISplitViewController.class, LiveAuctionViewController.class ];
+    NSArray *modalClasses = @[ UINavigationController.class, UISplitViewController.class, LiveAuctionViewController.class, ARNewModalComponentViewController.class ];
     for (Class klass in modalClasses) {
         if ([viewController isKindOfClass:klass]) {
             return YES;
@@ -482,7 +484,16 @@ static ARTopMenuViewController *_sharedManager = nil;
         // Once that PR is merged and we've upgraded, we can remove the following line
         // of code, which opts us out of the new modal presentation stylel.
         if ([UIDevice isPhone]) {
-            viewController.modalPresentationStyle = UIModalPresentationFullScreen;
+            BOOL wantsNewModalStyle = NO;
+            wantsNewModalStyle |= [viewController respondsToSelector:@selector(newModalStyle)] && [(id)viewController newModalStyle];
+            wantsNewModalStyle |= [viewController isKindOfClass:UINavigationController.class] && [[(UINavigationController *)viewController topViewController] respondsToSelector:@selector(newModalStyle)] && [(id)[(UINavigationController *)viewController topViewController] newModalStyle];
+            if (wantsNewModalStyle) {
+                if (@available(iOS 13.0, *)) {
+                    viewController.modalPresentationStyle = UIModalPresentationAutomatic;
+                }
+            } else {
+                viewController.modalPresentationStyle = UIModalPresentationFullScreen;
+            }
         } else {
             if ([viewController isKindOfClass:UINavigationController.class] && [[(UINavigationController *)viewController topViewController] isKindOfClass:ARBidFlowViewController.class]) {
                 // Bid Flow gets form sheet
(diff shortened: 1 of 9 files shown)


src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/ConsignmentsHome.tsx
```diff
@@ -33,10 +33,18 @@ export const ConsignmentsHome: React.FC<Props> = ({ targetSupply, isLoading }) =
     }
   }
 
+  const handleNewModal = () => {
+    if (navRef.current) {
+      const route = "/new-modal"
+      SwitchBoard.presentNavigationViewController(navRef.current, route)
```
damassi, line 39: @ashfurrow  - Thank you for helping with this! 

Curious about the naming, why "Navigation"? Why no mention of "Modal"? Maybe including `pageSheet` would be a bit more descriptive? https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/pagesheet

Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
```diff
@@ -482,7 +484,16 @@ - (void)pushViewController:(UIViewController *)viewController animated:(BOOL)ani
         // Once that PR is merged and we've upgraded, we can remove the following line
         // of code, which opts us out of the new modal presentation stylel.
         if ([UIDevice isPhone]) {
-            viewController.modalPresentationStyle = UIModalPresentationFullScreen;
+            BOOL wantsNewModalStyle = NO;
+            wantsNewModalStyle |= [viewController respondsToSelector:@selector(newModalStyle)] && [(id)viewController newModalStyle];
+            wantsNewModalStyle |= [viewController isKindOfClass:UINavigationController.class] && [[(UINavigationController *)viewController topViewController] respondsToSelector:@selector(newModalStyle)] && [(id)[(UINavigationController *)viewController topViewController] newModalStyle];
+            if (wantsNewModalStyle) {
+                if (@available(iOS 13.0, *)) {
+                    viewController.modalPresentationStyle = UIModalPresentationAutomatic;
```
damassi, line 492: So this will determine if the "floating" aspect is possible on the device, and then use something else if the viewport doesn't support?

damassi: (Looking here: https://developer.apple.com/documentation/uikit/uimodalpresentationstyle)

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/ConsignmentsHome.tsx
ashfurrow: Yeah, it's a poorly-named function but it's used everywhere so we haven't refactored it yet. It's "navigation" in the sense that it pushes onto the visible navigation stack.

Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m
ashfurrow: Yes that's correct.

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/ConsignmentsHome.tsx
damassi: So would all of these items then be presented inside of a modal? Wouldn't we want to maintain a clear distinction between full-page take-overs and overlay (modal) content? Why not update `presentModalViewController`? 
Screen Shot 2020-06-18 at 11 56 21 AM

ashfurrow:
Okay so next steps are up to you. There are two options:

- Close this PR and apply the approach it outlines to your existing code. I tried to describe the important bits in the description. This would be the best option if you already have a component you want to present this way.
- Modify this PR to rename the `ARNewModalComponentViewController" class and "NewModal" component to some more meaningful identifiers. Remove the new route in "ARSwitchBoard", move the code out of the AppRegistry, and so on. This would be the best option if you haven't yet created the modal you want displayed this way.

This PR was just a proof of concept to show how this could work – definitely shouldn't merge it as-is, but it's not far from working 👍 Let me know where I can clarify.

damassi:
Thanks for the help @ashfurrow 👍

src/lib/Scenes/Consignments/v2/Screens/ConsignmentsHome/ConsignmentsHome.tsx
ashfurrow: Those would present modally depending on the route. It's kind of a subtle distinction here, but the code modified in this PR shows how the Objective-C side introspects the class to determine if it should be presented modally. The RN-visible `presentModalViewController` will wrap whatever route you specify in a `SerifNavigationController`, which gives a standard close button that's not included in your designs. 

Like I said, kind of subtle. Longterm, MX is working on cleaning this up.

damassi: Ah! This now all makes sense 😄

ds300:
@damassi worth noting that while you were OOO we implemented a react-focused way of leveraging this modal style. It's called `FancyModal`. You can check the `Select` implementation if you want to get an idea of how to use it. It's pretty similar to the normal react-native `Modal` in that you just add a `<FancyModal visible={foo}>{modalContent}</FancyModal>` somewhere, anywhere, and then when visible is true it will show.

damassi:
@ds300 - what great news to come back to 😄 😄 🎉

ashfurrow:
This PR has been replaced by the `FancyModal`, going to close.

PR #3487, opened 2020-06-19
[MX-360] Change email address

MounirDhahri:
### Description
Allow users to edit their email address

### Issue
https://artsyproduct.atlassian.net/browse/MX-360

### Screenshots
ezgif-5-dbc3a9ceda4d


src/lib/Scenes/MyAccount/MyAccountEditEmail.tsx
```diff
@@ -0,0 +1,103 @@
...
+          }
+        `,
+        variables: {
+          input: {
+            email,
+          },
+        },
+        onError: () => {
+          InteractionManager.runAfterInteractions(() => {
+            Alert.alert("Something went wrong while saving email")
```
MounirDhahri, line 37: Maybe we will need to adjust the content here

src/lib/Scenes/MyAccount/MyAccountEditEmail.tsx
```diff
@@ -0,0 +1,103 @@
...
+            }
+          }
+        `,
+        variables: {
+          input: {
+            email,
+          },
+        },
+        onError: () => {
+          InteractionManager.runAfterInteractions(() => {
```
MounirDhahri, line 36: This is why I am using the InteractionManager here: https://github.com/facebook/react-native/issues/10471
Waiting for the loading animation to finish and then showing the alert worked fine for me. We could also maybe use `requestAnimationFrame`

src/lib/Scenes/MyAccount/MyAccountEditEmail.tsx
ds300: yeah plus I'm really hesitant to ship this feature without email verification in place, so we can revisit this copy later.

src/lib/Scenes/MyAccount/MyAccountEditEmail.tsx
```diff
@@ -0,0 +1,103 @@
...
+              }
+            }
+          }
+        `,
+        variables: {
+          input: {
+            email,
+          },
+        },
+        onError: () => {
```
ds300, line 35: I think it would be useful to catch and log this error in development

src/lib/Scenes/MyAccount/MyAccountEditEmail.tsx
MounirDhahri: Oki perfect.

src/lib/Scenes/MyAccount/MyAccountEditEmail.tsx
MounirDhahri: sure, I thought it'll be already caught by `principalFieldErrorMiddleware`. I will adjust it

PR #3488, opened 2020-06-22
[MX-362] Change phone number 

MounirDhahri:
### Description
Allow users to edit their phone number

### Issue
https://artsyproduct.atlassian.net/browse/MX-362

### Screenshots
Screen_Recording_2020-06-23_at_11 37 04


data/schema.graphql
```diff
@@ -6153,6 +6174,9 @@ type Message implements Node {
   # A type-specific ID likely used as a database ID.
   internalID: ID!
   invoice: Invoice
+    @deprecated(
+      reason: "Payment Request was deprecated. The field was kept for legacy client support. [Will be removed in v2]"
+    )
```
ds300, line 6179: where did these come from? We shouldn't have to update the schema file in our PRs, it should happen in these automatically-generated PRs like #3168

MounirDhahri: yes sure,

PR #3492, opened 2020-06-24
[MX-361] Change Password Screen

MounirDhahri:
### Description
Allow users to edit their password

### Issue
https://artsyproduct.atlassian.net/browse/MX-361

### Screenshots

Screen_Recording_2020-06-25_at_13 28 18


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>Missing Test Files:

- `src/lib/Scenes/MyAccount/__tests__/MyAccountEditPassword-tests.tsx`

    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/Scenes/MyAccount/MyAccountEditPassword.tsx
```diff
@@ -0,0 +1,109 @@
...
+            description="Must include at least one uppercase letter, one lowercase letter, and one number."
+            onChangeText={setNewPassword}
+            placeholder="New password"
+            secureTextEntry
+            showClearButton
+            title="New password"
+            value={newPassword}
+          />
+          <Input
+            containerStyle={{ marginTop: 30 }}
```
ds300, line 97: it's fine that you added this containerStyle prop, but a more standard pattern (in artsy codebases) for achieving this margin would be to just put `<Spacer mb="3"/>` between the two inputs :)

src/lib/Scenes/MyAccount/MyAccountEditPhone.tsx
```diff
@@ -1,21 +1,26 @@
...
   const [phone, setPhone] = useState<string>(me.phone ?? "")
+  const inputRef = useRef(null)
 
   const onSave = async () => {
     await new Promise((resolve, reject) =>
       commitMutation<MyAccountEditPhoneMutation>(relay.environment, {
-        onCompleted: resolve,
+        onCompleted: () => {
+          resolve()
+          SwitchBoard.dismissNavigationViewController(inputRef.current!)
```
ds300, line 22: MyAccountFieldEditScreen should do this navigation automatically, is it not working?

Artsy/App/ARSwitchBoard.m
```diff
@@ -292,24 +292,27 @@ - (void)updateRoutes
...
-        
+
         [self.routes addRoute:@"/my-account/edit-name" handler:JLRouteParams {
             return [[ARComponentViewController alloc] initWithEmission:nil moduleName:@"MyAccountEditName" initialProperties:parameters hidesBackButton:YES];
         }];
-        
+
+        [self.routes addRoute:@"/my-account/edit-password" handler:JLRouteParams {
+            return [[ARComponentViewController alloc] initWithEmission:nil moduleName:@"MyAccountEditPassword" initialProperties:parameters hidesBackButton:YES];
+        }];
```
brainbicycle, line 306: Non-blocking but should be consistent with spacing in here

src/lib/Scenes/MyAccount/MyAccountEditPassword.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const [currentPassword, setCurrentPassword] = useState<string>("")
+  const [newPassword, setNewPassword] = useState<string>("")
+  const [passwordConfirmation, setPasswordConfirmation] = useState<string>("")
+
+  const onSave = async () => {
+    if (newPassword !== passwordConfirmation) {
+      return Alert.alert("Password confirmation does not match")
+    }
+    try {
+      const res = await fetch(gravityURL + "/api/v1/me/password", {
```
brainbicycle, line 33: I know we talked about it but might be worth putting a ticket to migrate to use a metaphysics mutation in the backlog, think our ideal state for Eigen is to not interact with gravity directly this way.

src/lib/Scenes/MyAccount/MyAccountEditPassword.tsx
```diff
@@ -0,0 +1,109 @@
...
+        },
+        body: JSON.stringify({
+          current_password: currentPassword,
+          new_password: newPassword,
+        }),
+      })
+      const response = await res.json()
+      // The user successfully updated his password
+      if (!response.error) {
+        Alert.alert(
```
brainbicycle, line 48: Not specific to this PR but do we have any thoughts or components that should be used as preferred pattern for quick user facing messages that don't require an action? @samchieng 
Looks like something we could handle with a toast https://github.com/scalessec/Toast (not this one specifically) or maybe a success version of the banner here? https://www.figma.com/file/OSXdLHRQMbe6k6a58uW5Xp/Profile-Tab?node-id=150%3A2890
Think native alert is not most user friendly way of presenting this stuff.

samchieng: @brainbicycle do you have an example of a use case? trying to understand when we'd need a toast message like this

brainbicycle: I am thinking any action where we would want to tell to the user that their action succeeded but there is nothing to be done on their part. 
Some I can think of:
- User places a max bid "Your bid has been placed"
- User submits a consignment "Consignment submitted successfully" (I think we show a full screen confirm for this case so maybe not)
- Changing their settings ("Password saved", "Email updated", "Notification settings updated" etc..)

samchieng: Ah got it. For our existing states, I think bidding + consignments both have a full screen confirmation page. 

For updating settings, let's launch v1 without a success message that follows standard iOS patterns (auto-save, no UI confirmation). I think it's a better add for v2 with trickier flows like sending email verification from this screen.

We should add a toast component to Palette though! I think the lot condition report is a good use case. I'll add a ticket in our backlog for a design spike since we'll need to vet through a lot of design systems.
https://artsyproduct.atlassian.net/browse/MX-394

brainbicycle: Thanks Sam! Also realizing since we are logging out user on change password toast doesn't make sense for this use case 😅 but think it does make sense for others.

src/lib/Scenes/MyAccount/MyAccountEditPassword.tsx
MounirDhahri: done :)

PR #3498, opened 2020-06-25
Recent search easy peasy migration

brainbicycle:

### Description
Migrate Recent Searches from using the Context Api to Easy-Peasy

Co-authored by @ds300 @MounirDhahri


src/lib/Scenes/Search/RecentSearches.tsx
```diff
@@ -1,82 +1,65 @@
...
-  return {
-    get recentSearches() {
-      return searches ? searches.slice(0, numSearches) : []
-    },
-    async notifyRecentSearch(search: RecentSearch) {
-      await updateSearches(oldSearches => {
-        const newSearches = oldSearches.filter(s => s.props.href !== search.props.href)
-        newSearches.unshift(search)
+export const RecentSearchContext = createContextStore<Store>(
+  persist(
```
damassi, line 37: @brainbicycle - not sure how this works in RN's version of local storage, but there's some redux middleware that you can plug in to easy peasy -- here's an example from one of my web projects https://github.com/damassi/fretboard-trainer/blob/master/src/store.tsx#L54

In any case, it would be easy to write a little bit of middleware versus handling this manually.

brainbicycle: Thanks! I'll check it out today

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>Missing Test Files:

- `src/lib/utils/__tests__/storage-tests.ts`

    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/Scenes/Search/RecentSearches.tsx
```diff
@@ -1,82 +1,60 @@
...
-        try {
-          setSearches(JSON.parse(val))
-          return
-          // tslint:disable-next-line:no-empty
-        } catch (e) {}
-      }
-      setSearches([])
-    })
-  }, [])
+export interface StoreModel {
```
damassi, line 19: Would it be worthwhile to break this state stuff out into its own file and keep react components separate?

src/lib/Scenes/Search/Search.tsx
```diff
@@ -9,16 +9,41 @@ import { useTracking } from "react-tracking"
...
 import { AutosuggestResults } from "./AutosuggestResults"
 import { CityGuideCTA } from "./CityGuideCTA"
-import { ProvideRecentSearches, RecentSearches, useRecentSearches } from "./RecentSearches"
+import { getRecentSearches, ProvideRecentSearches, RecentSearchContext, RecentSearches } from "./RecentSearches"
 import { SearchContext, useSearchProviderValues } from "./SearchContext"
 
 const SearchPage: React.FC = () => {
   const [query, setQuery] = useState("")
-  const { recentSearches } = useRecentSearches()
+  const recentSearches = getRecentSearches(RecentSearchContext.useStoreState(state => state.recentSearches))
```
damassi, line 17: Why not make this whole thing its own function / hook? Like
```tsx
cons recentSearches = useRecentSearches()
```

src/lib/Scenes/Search/SearchResult.tsx
```diff
@@ -46,7 +46,8 @@ export const SearchResult: React.FC<{
           setTimeout(() => {
             SwitchBoard.presentNavigationViewController(navRef.current, result.href!)
             if (updateRecentSearchesOnTap) {
-              notifyRecentSearch({ type: "AUTOSUGGEST_RESULT_TAPPED", props: result })
+              addRecentSearch({ type: "AUTOSUGGEST_RESULT_TAPPED", props: result })
```
damassi, line 49: One thought here is that the use of `type` as the name of the key is a bit confusing, because it crosses concepts / libraries / pattterns. Meaning: in vanilla redux you have an action with a type. When using `useReducer` you dispatch an action with a `type`. In `easy-peasy`, however, one just calls functions and passes an argument. Can almost guarantee that future devs reading this will do a double take and be slightly confused.

src/lib/Scenes/Search/__tests__/RecentSearches-tests.tsx
```diff
@@ -3,28 +3,62 @@ import { flushPromiseQueue } from "lib/tests/flushPromiseQueue"
...
+import {
+  getRecentSearches,
+  MAX_SAVED_RECENT_SEARCHES,
+  MAX_SHOWN_RECENT_SEARCHES,
+  RecentSearch,
+  RecentSearchContext,
+} from "../RecentSearches"
+
+const banksy: RecentSearch = {
+  type: "AUTOSUGGEST_RESULT_TAPPED",
```
damassi, line 15: maybe `event`?

ds300:
This is great stuff, thanks to everyone involved!

I have a couple of reservations about merging this as-is

1. We should try to find a way to migrate people's existing recent-searches (from before this refactor is released) otherwise next time their app updates they'd lose all their search history.

2. We will soon need to augment the nascent global app store (in `AppStore.tsx`) with persistence, versioning, and migration infrastructure, and this search state could be moved there already and used to spearhead that work sooner rather than later.

I'd vote for doing 2 since

- this is just a refactor and is not adding any functionality
- we're not working on search right now so this work won't unlock other items in the backlog or block anyone
- we need to do 2 at some point anyway.

ds300:
💡  I'm going to try to get this done today and present in knowledge share tomorrow.

MounirDhahri:
Thanks, I didn't know how to proceed with it

src/lib/store/AppStore.tsx
```diff
@@ -4,10 +4,15 @@ import { defaultsDeep } from "lodash"
 import React, { useRef } from "react"
 import { Action, Middleware } from "redux"
 import { AppStoreModel, appStoreModel, AppStoreState } from "./AppStoreModel"
+import { persistenceMiddleware, unpersist } from "./persistence"
 
 function createAppStore() {
   const middleware: Middleware[] = []
 
+  if (!__TEST__) {
+    middleware.push(persistenceMiddleware)
```
damassi, line 13: 👍

src/lib/store/migration.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export type Migrations = Record<number, (oldState: any) => any>
+/**
+ * If you modify the app state in a backwards-incompatible way, i.e. one of
+ *   - changing the structure of any object type which gets stored in arrays
+ *   - removing a property from a model
+ *   - changing a property's value type
+ *   - renaming a property
+ * you must bump `CURRENT_APP_VERSION` and add a migration for your new version number here
+ * which makes the previous app state version compatible with your new version.
```
damassi, line 12: @ds300 - curious about the thinking that prompted the need for this migration structure, and as a developer is there now some overhead that we need to remember or does the app automatically track and increment this?

ds300: If you need to make backwards-incompatible changes to the app state over time, then you need migration infrastructure like this. Otherwise app updates can end up obliterating any client-local settings the user stores. I've worked on a couple of RN apps with redux before and we had basically this same setup so I figured I'd duplicate it here. I'm gonna add docs and a checkbox item in the PR template to make sure folks understand the implications.

docs/adding_state_migrations.md
```diff
@@ -0,0 +1,68 @@
...
+
+    initialAppState = merge(blankAppState, loadSavedStateFromDeviceStorage())
+
+It is possible to make a code change that would cause this step to produce a corrupt `initialAppState` the next time the user updates their app. To mitigate this we add in a `migrate` step so that old versions of the app state can be brought up-to-date before merging.
+
+    initialAppState = merge(blankAppState, migrate(loadSavedStateFromDeviceStorage()))
+
+When we make a change to the app store which could affect what is stored on disk, we need to create a new migration.
+
+The following kinds of change do _not_ require creating a new migration:
```
damassi, line 15: It would be equally (or more) informative to include examples of what might constitute a migration requirement. As a passer by, I only have a vague idea of what the app store does in the context of a mobile app / persistence.

src/lib/store/migration.ts
```diff
@@ -0,0 +1,43 @@
+import { produce } from "immer-peasy"
```
MounirDhahri, line 1: "Winner of the "**breakthrough of the year**" open source award in 2019". I always wanted to use this 🙈

src/lib/store/persistence.ts
```diff
@@ -0,0 +1,96 @@
...
+    return {}
+  }
+}
+
+export const persistenceMiddleware: Middleware = store => {
+  const throttledPersist = throttle(persist, 1000, { leading: false, trailing: true })
+  return next => action => {
+    next(action)
+    // use requestAnimationFrame to make doubly sure we avoid blocking UI updates
+    requestAnimationFrame(() => {
```
MounirDhahri, line 92: 🔥

src/lib/store/migration.ts
damassi: Why not import directly from `immer`?

docs/adding_state_migrations.md
ds300: good call. I added more examples. Is anything still unclear?

src/lib/store/migration.ts
ds300: it's not in our dependency tree. `immer-peasy` is a fork.

docs/adding_state_migrations.md
damassi: Docs are perfect now 👍 thnx for the updates.

ds300:
I'm going to close this and open a new PR on the same branch to give a clean slate for conversation, now that things have changed so much.

PR #3517, opened 2020-07-01
[MX-336] Adds tracking for home promo space #trivial

ashfurrow:
Not sure if I went overboard on the typings here, open to feedback.

[In the Cohesion PR](https://github.com/artsy/cohesion/pull/69#discussion_r446206905), @eessex mentioned that on the web, we split the URL to calculate these attributes. I took a stab at this, but there are a lot of edge cases and unless we match the web implementation exactly, it will only introduce noise in the data. These attributes can be calculated by the data team based on `destination_path`, so I didn't add the `destination_screen_owner_type`, `destination_screen_owner_id`, or `destination_screen_owner_slug` attributes. /cc @mikehrom 

See: https://github.com/artsy/cohesion/pull/69


src/lib/Scenes/Home/Components/HomeHero.tsx
```diff
@@ -28,12 +31,21 @@ const HomeHero: React.FC<{ homePage: HomeHero_homePage }> = ({ homePage }) => {
 
   const { width, height } = useHeroDimensions()
 
+  const handlePromoSpaceTap = () => {
+    const path = unit.href!
+    const event: TappedPromoSpace = {
```
eessex, line 36: We should be using a helper rather than importing the interface directly- will need to create one in cohesion.

ashfurrow: I see, thanks. I'll take that on next 👍

PR #3521, opened 2020-07-02
[CSGN-269] Stub out my collection artwork list

pepopowitz:
Addresses https://artsyproduct.atlassian.net/browse/CSGN-269

This PR builds out an initial artwork list for My Collection. It uses hard-coded data. 

When tapping on an artwork the user is taken to an artwork detail view, but the data there does not yet match what they tapped on.

The "skeleton" flow is also not removed yet, so there is a "back" nav arrow that appears over the "Your collection" title. Once the skeleton flow is removed, that will be gone.

your-collection

#trivial (but more like WIP and not yet deserving its own changelog entry)


src/lib/Scenes/Consignments/v2/Screens/MyCollectionArtworkList/MyCollectionArtworkList.tsx
```diff
@@ -1,13 +1,121 @@
...
+    medium: "Dry Erase markers",
+    image: {
+      url: "https://d32dm0rphc51dk.cloudfront.net/DkpNiCKRYoqa7BXEtsZSpQ/square.jpg",
+    },
+  },
+  {
+    id: "3",
+    slug: "my-artwork/3",
+    artistNames: "James Rosenquist",
+    medium: "Cat hair",
```
jonallured, line 110: 😹

src/lib/Scenes/Consignments/v2/Screens/MyCollectionArtworkList/MyCollectionArtworkList.tsx
```diff
@@ -1,13 +1,121 @@
...
+      )}
+    />
+  )
+}
+
+const TouchElement = styled.TouchableHighlight.attrs({ underlayColor: color("white100"), activeOpacity: 0.8 })``
+
+interface MyCollectionArtworkListItemProps {
+  item: MyCollectionArtwork
+  onPress: ((event: GestureResponderEvent) => void) | undefined
```
jonallured, line 36: I was surprised to see undefined here, can you help me understand when this would not be a navigate call?

src/lib/Scenes/Consignments/v2/Screens/MyCollectionArtworkList/MyCollectionArtworkList.tsx
```diff
@@ -1,13 +1,121 @@
...
+}
+const MyCollectionArtworkListItem: React.FC<MyCollectionArtworkListItemProps> = ({ item, onPress }) => {
+  const imageURL = item.image?.url
+  const imageDisplay = !!imageURL ? (
+    <OpaqueImageView
+      imageURL={imageURL.replace(":version", "square")}
+      width={90}
+      height={90}
+      // style={{ borderRadius: 2, overflow: "hidden" }}
```
jonallured, line 45: Does this tie back to that question you asked in Slack about which way to run these images?

src/lib/Scenes/Consignments/v2/Screens/MyCollectionArtworkList/MyCollectionArtworkList.tsx
pepopowitz: I'm not really sure - I copy-pasta'd this type from somewhere else and didn't look at it closely. I agree that `undefined` should be removed, and I'll pull it out in my next PR.

src/lib/Scenes/Consignments/v2/Screens/MyCollectionArtworkList/MyCollectionArtworkList.tsx
pepopowitz: Yes! I left these lines in as comments to remind me that I might need to eventually uncomment them.

PR #3529, opened 2020-07-02
Revert #3527 since the root cause has been fixed

rogerluan:
Reverts #3527 
Root cause was fixed here: https://github.com/postmodern/digest-crc/pull/20#event-3508970360 and released in https://github.com/postmodern/digest-crc/releases/tag/v0.6.1

#trivial


ds300:
Nice one, thanks! 🙏  I'll assign @brainbicycle

brainbicycle:
Thanks for the follow-up 🙌

PR #3530, opened 2020-07-03
[MX-401][MX-403] Home Promo QA fixes

ds300:
https://artsyproduct.atlassian.net/browse/MX-401 - add 60% breakpoint for title on iPad

image

https://artsyproduct.atlassian.net/browse/MX-403 - fix spacing above and below hero unit

image

✨ bonus ✨ 

- fix refetch query so it gets the right version of the hero background image

### DIFF
--- a/src/lib/Scenes/Home/Home.tsx
+++ b/src/lib/Scenes/Home/Home.tsx
@@ -104,7 +104,7 @@ const Home = (props: Props) => {
     setIsRefreshing(true)
 
     props.relay.refetch(
-      {},
+      { heroImageVersion: isPad() ? "WIDE" : "NARROW" },
       {},
       error => {
         if (error) {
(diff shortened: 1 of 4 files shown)


src/lib/Scenes/Home/Home.tsx
```diff
@@ -308,6 +310,7 @@ export const HomeQueryRenderer: React.SFC = () => {
       `}
       variables={{ heroImageVersion: isPad() ? "WIDE" : "NARROW" }}
       render={renderWithPlaceholder({ Container: HomeFragmentContainer, renderPlaceholder: () => <HomePlaceholder /> })}
+      cacheConfig={{ force: true }}
```
brainbicycle, line 313: Just out of curiosity what was happening? It was serving same image from cache rather than new images?

ds300: before this, if the hero data changes and you refresh the page by pulling down, then you'll see the updated data thanks to the refetch query. But then if you unmount and remount the home screen (e.g. by hitting cmd+r in dev mode) to show this query ☝️  then the previous version of the data was shown. I think this must be a problem with either our client-side caching, or the ids we generate for hero units in metaphysics are not stable? I would expect the refetch query to update the same cache entries as this query uses. Didn't have time to look into this more and it doesn't seem to be a pervasive problem so I just prevented the home screen from using the cache for now.

brainbicycle: Thanks for explanation!

PR #3534, opened 2020-07-06
[MX-387][MX-388][MX-389] - Remove noisy sentry errors

brainbicycle:
https://artsyproduct.atlassian.net/browse/MX-387
https://artsyproduct.atlassian.net/browse/MX-388
https://artsyproduct.atlassian.net/browse/MX-389

This pr cleans up some of our more noisy sentry errors: 
- Remove untracked home rail sentry message and add missing gallery rail tracking
- Remove volley send failure error message
- Remove deep zoom error message

Deep zoom may require some follow-up, although it seems from code the deepZoom image is expected to be [optional](https://github.com/artsy/metaphysics/blob/master/src/schema/v2/image/deep_zoom.ts#L79) not totally clear on root cause of so many artworks not supporting it, added a [ticket](https://artsyproduct.atlassian.net/browse/MX-408) to backlog to investigate further, 18k events should be enough to go on hopefully 😅. 

[Sentry Events](https://sentry.io/organizations/artsynet/issues/1495784215/?project=166784&query=is%3Aunresolved)


src/lib/utils/volleyClient.ts
```diff
@@ -90,16 +89,6 @@ class VolleyClient {
...
-        .then(e => {
-          if (!__DEV__ && e.status >= 400) {
-            captureMessage(`Failed to post metrics to volley (status ${e.status})`)
-          }
-        })
-        .catch(() => {
-          if (!__DEV__) {
-            captureMessage("Failed to post metrics to volley")
-          }
-        })
```
ds300, line 102: we should probably keep the catch here with an empty block. I think unhandled promise rejections get treated as errors.

brainbicycle: Good call 👍 , added back with console.error, typescript doesn't like the empty block

PR #3537, opened 2020-07-06
Adds Entity Header to Artist Series

ashleyjelks:
https://artsyproduct.atlassian.net/browse/FX-2086

Kapture 2020-07-08 at 20 04 50

Kapture 2020-07-08 at 20 14 35

### DIFF
--- a/src/lib/Scenes/ArtistSeries/ArtistSeries.tsx
+++ b/src/lib/Scenes/ArtistSeries/ArtistSeries.tsx
@@ -1,35 +1,97 @@
-import { Box, Flex, Sans } from "@artsy/palette"
+import { Box, EntityHeader, Flex, Sans } from "@artsy/palette"
 import { ArtistSeries_artistSeries } from "__generated__/ArtistSeries_artistSeries.graphql"
+import { ArtistSeriesFollowMutation } from "__generated__/ArtistSeriesFollowMutation.graphql"
 import { ArtistSeriesQuery } from "__generated__/ArtistSeriesQuery.graphql"
 import OpaqueImageView from "lib/Components/OpaqueImageView/OpaqueImageView"
 import { ReadMore } from "lib/Components/ReadMore"
+import SwitchBoard from "lib/NativeModules/SwitchBoard"
 import { defaultEnvironment } from "lib/relay/createEnvironment"
 import renderWithLoadProgress from "lib/utils/renderWithLoadProgress"
 import { useScreenDimensions } from "lib/utils/useScreenDimensions"
-import React from "react"
-import { View } from "react-native"
-import { createFragmentContainer, graphql, QueryRenderer } from "react-relay"
+import React, { useRef } from "react"
+import { TouchableOpacity, TouchableWithoutFeedback, View } from "react-native"
+import { commitMutation, createFragmentContainer, graphql, QueryRenderer, RelayProp } from "react-relay"
 
 interface ArtistSeriesProps {
   artistSeries: ArtistSeries_artistSeries
+  relay: RelayProp
 }
 
-export const ArtistSeries: React.FC<ArtistSeriesProps> = ({ artistSeries }) => {
+type ArtistToFollowOrUnfollow = NonNullable<NonNullable<ArtistSeriesProps["artistSeries"]["artists"]>[0]>
+
+export const ArtistSeries: React.FC<ArtistSeriesProps> = ({ artistSeries, relay }) => {
   const { width } = useScreenDimensions()
   const isIPad = width > 700
   const maxChars = isIPad ? 200 : 120
+  const ref = useRef<View | null>(null)
+
+  const followOrUnfollowArtist = (followArtist: ArtistToFollowOrUnfollow) => {
+    return new Promise<void>((resolve, reject) => {
+      commitMutation<ArtistSeriesFollowMutation>(relay.environment, {
+        mutation: graphql`
+          mutation ArtistSeriesFollowMutation($input: FollowArtistInput!) {
+            followArtist(input: $input) {
+              artist {
+                isFollowed
+              }
+            }
+          }
+        `,
+        variables: {
+          input: { artistID: followArtist.internalID, unfollow: followArtist.isFollowed },
+        },
+        onError: reject,
+        onCompleted: (_response, errors) => {
+          if (errors && errors.length > 0) {
+            reject(new Error(JSON.stringify(errors)))
+          } else {
+            resolve()
+          }
+        },
+      })
+    })
+  }
 
   return (
-    <View>
+    <View ref={ref}>
       <Box p={2}>
         <Flex flexDirection="row" justifyContent="center">
           {/* TODO: add image url */}
           <OpaqueImageView width={180} height={180} />
         </Flex>
-        <Sans size="8" mt={3} mb={1}>
+        <Sans size="8" mt={3} data-test-id="title">
           {artistSeries.title}
         </Sans>
-        <ReadMore sans content={artistSeries?.description ?? ""} maxChars={maxChars} />
+        {(artistSeries?.artists ?? []).length > 0 &&
+          artistSeries.artists?.map(
+            artist =>
+              !!artist?.name && (
+                <TouchableOpacity
+                  key={artist?.id as string}
+                  onPress={() => {
+                    SwitchBoard.presentNavigationViewController(ref.current!, `/artist/${artist.slug}`)
+                  }}
+                  style={{ marginVertical: 10 }}
+                >
+                  <EntityHeader
+                    smallVariant
+                    name={artist?.name}
+                    imageUrl={artist?.image?.url as string}
+                    FollowButton={
+                      <TouchableWithoutFeedback
+                        onPress={() => followOrUnfollowArtist(artist)}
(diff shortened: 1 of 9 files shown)


ds300:
:ceiling-cat: should we be using the mobile version of the entity header with the right-justified button? like this:

image

ashleyjelks:
> :ceiling-cat: should we be using the mobile version of the entity header with the right-justified button? like this:
> 
> image

The artist series uses a smaller variant of the Entity Header component:

Screen Shot 2020-07-08 at 7 54 41 PM

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>The CHANGELOG is not valid YML:
YAMLSemanticError: Map keys must be unique; "dev" is repeated
    at Object.parseMap [as resolve] (/home/circleci/project/node_modules/yaml/dist/schema/parseMap.js:71:27)
    at Schema.resolveNode (/home/circleci/project/node_modules/yaml/dist/schema/index.js:177:27)
    at Schema.resolveNodeWithFallback (/home/circleci/project/node_modules/yaml/dist/schema/index.js:200:22)
    at Document.resolveNode (/home/circleci/project/node_modules/yaml/dist/Document.js:494:22)
    at resolveBlockMapItems (/home/circleci/project/node_modules/yaml/dist/schema/parseMap.js:198:48)
    at Object.parseMap [as resolve] (/home/circleci/project/node_modules/yaml/dist/schema/parseMap.js:33:79)
    at Schema.resolveNode (/home/circleci/project/node_modules/yaml/dist/schema/index.js:177:27)
    at Schema.resolveNodeWithFallback (/home/circleci/project/node_modules/yaml/dist/schema/index.js:200:22)
    at Document.resolveNode (/home/circleci/project/node_modules/yaml/dist/Document.js:494:22)
    at /home/circleci/project/node_modules/yaml/dist/Document.js:222:26</td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/Scenes/ArtistSeries/ArtistSeries.tsx
```diff
@@ -1,35 +1,103 @@
...
+import { TouchableOpacity, TouchableWithoutFeedback, View } from "react-native"
+import { commitMutation, createFragmentContainer, graphql, QueryRenderer, RelayProp } from "react-relay"
 
 interface ArtistSeriesProps {
   artistSeries: ArtistSeries_artistSeries
+  relay: RelayProp
 }
 
-export const ArtistSeries: React.FC<ArtistSeriesProps> = ({ artistSeries }) => {
+interface ArtistToFollowOrUnfollow {
```
ashleyjelks, line 20: TS complained when I attempted to infer the type from `ArtistSeriesProps["artistSeries"][0]`  even though this pattern is used in lots of places. Not sure why, will follow up with @ds300 in the morning.

https://artsy.slack.com/archives/CP9P4KR35/p1594229446441400

src/lib/Scenes/ArtistSeries/ArtistSeries.tsx
```diff
@@ -1,35 +1,103 @@
...
+  } | null
+}
+
+export const ArtistSeries: React.FC<ArtistSeriesProps> = ({ artistSeries, relay }) => {
   const { width } = useScreenDimensions()
   const isIPad = width > 700
   const maxChars = isIPad ? 200 : 120
+  const ref = useRef<View | null>(null)
+
+  const followOrUnfollowArtist = (followArtist: ArtistToFollowOrUnfollow) => {
```
sweir27, line 37: Since I'm guessing we have this elsewhere, is there a way to extract or re-use an existing function?

ashleyjelks: Yeah I looked at other instances of this in Eigen and the only other place we are using the Entity Header hooked up to a follow artist mutation is here https://github.com/artsy/eigen/blob/07cf2006ba856019f7f55379fe6444fbbf25c0bb/src/lib/Components/Home/ArtistRails/ArtistRail.tsx#L130.  I considered extracting this function for repurposing, but the one other component using this also needs to handle storing the followed/unfollowed artists in state for use in Artist recommendations and is using a deprecated tracking method (https://artsy.slack.com/archives/CN8S32FJR/p1594153073474000?thread_ts=1594152634.473500&cid=CN8S32FJR). So I all things considered I was hesitant to do this. 

I _could_ extract this into a helper file for future uses of the follow artist mutation.

sweir27: No need to extract just yet, but if we add more instances of this we'll likely want to! Thanks for checking. 🙏

ashleyjelks:
CI is failing here due to an unrelated dependency https://app.circleci.com/pipelines/github/artsy/eigen/3107/workflows/4f7786ec-2ed5-49dc-9488-54c4eed259fa/jobs/13390/steps

I tried installing the dependency and bumping to a Palette version that was supposed to have a fix but neither of those changed the CI failures. https://artsy.slack.com/archives/C9XJKPY9W/p1592838422177100

ds300:
> The artist series uses a smaller variant of the Entity Header component:

sure I understand that's what's in the figma spec, but it's not a very mobile-friendly touch target. We explicitly decided that the iOS version of palette's entity header should use a right-aligned button. I just want to make sure that this design is intentionally going against the palette spec, which is used everywhere else in the app, rather than using the web version by accident.

src/lib/Scenes/ArtistSeries/ArtistSeries.tsx
```diff
@@ -1,35 +1,103 @@
...
         <Flex flexDirection="row" justifyContent="center">
           {/* TODO: add image url */}
           <OpaqueImageView width={180} height={180} />
         </Flex>
-        <Sans size="8" mt={3} mb={1}>
+        <Sans size="8" mt={3} data-test-id="title">
           {artistSeries.title}
         </Sans>
-        <ReadMore sans content={artistSeries?.description ?? ""} maxChars={maxChars} />
+        {(artistSeries?.artists ?? []).length > 0 &&
```
williardx, line 74: Could this just be `artistSeries.artists.length > 0 && ...`? Metaphysics should give us an empty array for `artists` when we don't have any artists associated with the series, and `artistSeries` at this point should be defined.

src/lib/Scenes/ArtistSeries/ArtistSeries.tsx
```diff
@@ -1,35 +1,103 @@
...
+          artistSeries.artists?.map(
```
williardx, line 75: Similar here, wondering if you need optional chaining?

ashleyjelks:
> > The artist series uses a smaller variant of the Entity Header component:
> 
> sure I understand that's what's in the figma spec, but it's not a very mobile-friendly touch target. We explicitly decided that the iOS version of palette's entity header should use a right-aligned button. I just want to make sure that this design is intentionally going against the palette spec, which is used everywhere else in the app, rather than using the web version by accident.

Makes sense!  This design is the iOS version but I think its worth following up with @nicoleyeo about this. In the event that she wants to keep this version of the EntityHeader in Eigen I wonder if we should just increase the hitSlop area so that the invisible area around the `Follow` text is tappable.

williardx:
Only minor questions, otherwise looks good to me! It's cool to see how you implemented this vs. what I did on web (https://github.com/artsy/force/pull/5907), I learned some new things!

src/lib/Scenes/ArtistSeries/ArtistSeries.tsx
ashleyjelks: Yes that would work but I used the nullable coalescing operator syntax here bc type strictness is enforced in Eigen and not handling the nullable case throws this TS error:  `Object is possibly 'null'.`

src/lib/Scenes/ArtistSeries/__tests__/ArtistSeries-tests.tsx
```diff
@@ -0,0 +1,108 @@
...
+          )
+        } else if (error) {
+          console.log(error)
+        }
+      }}
+    />
+  )
+
+  const getWrapper = () => {
+    const tree = ReactTestRenderer.create(<TestRenderer />)
```
anandaroop, line 49: Is there a heuristic for when to reach for an `Enzyme` wrapper vs a `ReactTestRenderer`? Is the latter preferred nowadays in Eigen?

ds300: yep we're not writing new tests with enzyme and we're planning to refactor tests currently using enzyme to use react-test-renderer instead.

ashleyjelks:
> CI is failing here due to an unrelated dependency https://app.circleci.com/pipelines/github/artsy/eigen/3107/workflows/4f7786ec-2ed5-49dc-9488-54c4eed259fa/jobs/13390/steps
> 
> I tried installing the dependency and bumping to a Palette version that was supposed to have a fix but neither of those changed the CI failures. https://artsy.slack.com/archives/C9XJKPY9W/p1592838422177100

Going to merge this now and follow up with @dzucconi and @damassi about these dependency issues.

PR #3539, opened 2020-07-07
remove some lines from settings

pvinis:
Just a quick change in VSCode settings:
- flow is not used, so let's remove it.
- editor whitespace showing should not be here, it can be on the user-level settings.
- we don't use svn or hg in this repo, and I'm pretty sure VSCode on mac does not show .DS_Store files, or if it does it will be ignored because of gitignore.


ds300:
hmm #mergeongreen didn't work 🤔  peril must be having an off day

PR #3507, opened 2020-06-26
[MX-289] - Artist Artwork filters

brainbicycle:
https://artsyproduct.atlassian.net/browse/MX-289

This PR is a lot bigger than I had initially thought it would be, apologies, I should have broken this up. Still a couple things to wrap up but would love some eyes and feedback! @ashleyjelks I know you did a lot of the work on the filter modal so would love any feedback if you have any, thanks for building that component! I want to make sure I am not breaking any existing behavior.

**At a high level this PR:**
- adds filter modal to artist artworks page
- loosens filter types a bit to accommodate filters returned from metaphysics (aggregate filters)
- converts some existing filters to use returned filters from API
- adds color, time period, gallery, institution, size filters
- converts collections screen to use updated filters

**Still to do:**
- Clean up some todos
- ~~Revert test code in home~~
- ~~Clean up some typings~~
- ~~Fix color filter layout on iPad~~
- ~~Fix filter ordering~~
- ~~More tests~~
- ~~Use updated filter modal~~

**Follow-ups:**
- Only show filters button when artist artworks are in view as collections does
- loading state for both artist artworks and collections, jarring as is

**Screenshots:**
filters-filter_button copy
filters-all-options copy
filters-colors copy
filters-with-selections copy

PR #3543, opened 2020-07-08
Fix types in filter helper, remove strictness comments #trivial

brainbicycle:
Follow-up from https://github.com/artsy/eigen/pull/3507 to clean up some typings. 

#trivial


src/lib/Scenes/Collection/Helpers/FilterArtworksHelpers.ts
```diff
@@ -84,35 +73,30 @@ export const filterArtworksParams = (appliedFilters: FilterArray) => {
...
+      const omitted = omit(selectedFilterParams, [filterParamName as string])
+      const changedFilter = omitted[filterParamName]
+      if (changedFilter) {
+        changedFilters[filterParamName] = omitted[filterParamName]
       }
     } else {
-      // @ts-ignore STRICTNESS_MIGRATION
-      changedFilters[filterType] = selectedFilterParams[filterType]
+      changedFilters[filterParamName] = selectedFilterParams[filterParamName]
     }
```
ds300, line 100: I think this whole if..else block can be replaced with 

```ts
if (currentFilterParams[filterParamName] !== selectedFilterParams[filterParamName]) {
  changedFilters[filterParamName] = selectedFilterParams[filterParamName]
}
```

I know it was like this before but we might as well clean it up now too :)

brainbicycle: 👍 much simpler

PR #3549, opened 2020-07-10
[MX-206] Fixes layout problem on LAI max bid selection screen

ashfurrow:
I was surprised to learn that this view is laid out in storyboards, so I will highlight the relevant code changes inline.

The problem was that the picker view for the max bid sits within a container view that is smaller than the picker; the container clips the bounds of the picker so it's okay. But the picker was aligned along the _bottom_ of the container, rather than at its center, which caused the misalignment.

Before:
Unknown-2

After:
Screen Shot 2020-07-10 at 10 34 19 AM

The import thing is that the little arrows on the left and right are vertically centered. I wouldn't call this a great UI or anything, but at least it's a little more polished. /cc @samchieng


Artsy/View_Controllers/Live_Auctions/Resources/LiveAuctions.storyboard
```diff
@@ -145,7 +143,7 @@
                                     <constraint firstAttribute="trailing" secondItem="bXr-xU-Qxb" secondAttribute="trailing" id="NLT-50-aox"/>
                                     <constraint firstAttribute="height" constant="138" id="bC8-3A-KmO"/>
                                     <constraint firstItem="kmx-cd-5vN" firstAttribute="centerY" secondItem="bXr-xU-Qxb" secondAttribute="centerY" id="lSH-R6-7bF"/>
-                                    <constraint firstAttribute="bottom" secondItem="bXr-xU-Qxb" secondAttribute="bottom" id="ra6-Va-OGP"/>
+                                    <constraint firstAttribute="centerY" secondItem="bXr-xU-Qxb" secondAttribute="centerY" id="ra6-Va-OGP"/>
```
ashfurrow, line 146: This is the only relevant change in this file (thank you, Xcode).

erikdstock:
Sorry I missed this review! I went ahead and fixed the merge conflict. Should be good to merge.

PR #3551, opened 2020-07-10
Fixes blank alert after successful password change #trivial

ashfurrow:
I was QA'ing [MX-361] when I noticed that, even in the success case, an empty alert was shown. This PR only shows the alert on error cases. It's a bit weird, since it's hitting Gravity's REST API directly, so the error-handling code is repeated twice.

[MX-361]: https://artsyproduct.atlassian.net/browse/MX-361


src/lib/Scenes/MyAccount/MyAccountEditPassword.tsx
```diff
@@ -34,7 +34,9 @@ export const MyAccountEditPassword: React.FC<{}> = ({}) => {
       })
       const response = await res.json()
       // The user successfully updated his password
-      if (!response.error) {
+      if (response.error) {
+        Alert.alert(typeof response.error === "string" ? response.error : "Something went wrong.")
+      } else {
         Alert.alert(
           "Password Changed",
           "Your Password has been changed successfully. Use your new password to log in",
```
brainbicycle, line 42: I know this is not part of this PR but should password be lowercase?

```suggestion
          "Your password has been changed successfully. Use your new password to log in",
```

ashfurrow: Mmmmm yes, let's do it 👍

PR #3555, opened 2020-07-10
[MX-431] - Fix some warnings

brainbicycle:
https://artsyproduct.atlassian.net/browse/MX-431

This fixes a few of the warnings in our native code in Eigen. 

**Before**

warningsbefore

**After**

warningsafter

**Fixed warnings:**
- **Implicit unwraps in view code**: we have a lot of methods that just setup views and assume root view is present, fix in most cases is just to add a guard and return early at top of method if assumption doesn't hold
- **handling unknown default cases in enums**: where possible marks objective-C enums as closed so we can just be exhaustive in handling cases otherwise add a default case
- **deprecated apis**: `adjustScrollViewInsets` replaced with `scrollView.contentInsetAdjustmentBehavior = .never` wherever possible, in auctions where it doesn't seem to be having an effect removed. Update notification registration to use new api
- **category method name collisions: UILabel+Typography**:  we have the same category in the Extraction pod repeated in the Eigen code, remove the category in the Eigen code

**Still to fix:**
The only warnings remaining are linker warnings about category method collisions, we should fix these if possible because the behavior is undefined when two categories override the same class method: 
 https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/CustomizingExistingClasses/CustomizingExistingClasses.html

- **category method name collision: RCTScrollView**: we deliberately monkey-patch RCTScrollView to allow getting onScroll callbacks from nested scrollViews: https://github.com/artsy/eigen/blob/master/emission/Pod/Classes/Core/RCTScrollView+Artsy.m.  It seems this is still necessary: https://github.com/facebook/react-native/issues/8024, not sure best way to fix this
- **category method name collision: SDWebImage**:  `setImageWithPlaceholder` method has a conflict with `AFNetworking` UIImageCategory. Versions of SDWebImage later than 4.0 prefix the method and remove the old method name so can be fixed by upgrading but requires a bit of refactoring on our side
- **category method name collision: EDColor**: `clamp` method is defined in two separate UIColor categories, we only use pod for converting hex strings to UIColors or converting hexValues to UIColors so can probably just get rid of it


Artsy/View_Controllers/Live_Auctions/ViewModels/LiveAuctionEventViewModel.swift
```diff
@@ -165,6 +165,7 @@ class LiveAuctionEventViewModel: NSObject, LiveAuctionEventViewModelType {
         case .unknown: return NSAttributedString()
         case .bidComposite: return NSAttributedString()
         case .undo: return NSAttributedString()
+        default: return NSAttributedString();
```
ashfurrow, line 168: _Very_ tiny nitpick:

```suggestion
        default: return NSAttributedString()
```

PR #3556, opened 2020-07-13
fix: notable works qa #trivial

adamvert:
https://artsyproduct.atlassian.net/browse/FX-2045 

Three post-QA tweaks in the Artist page:

1. Reduce space under Notable Works and Collections headings (N.B. the ticket specifies 10px - this is slightly more than 10px, but better than it was; reducing to precisely 10px would require an unconventional approach)
image

2. Hide separator if no Notable Works or Collections
image

3. Higher res images in Notable Works rail (see image in (1))


src/lib/Components/Artist/ArtistArtworks/ArtistCollectionsRail.tsx
```diff
@@ -19,7 +19,7 @@ export const ArtistCollectionsRail: React.FC<ArtistCollectionsRailProps> = props
   if (collections && collections.length > 1) {
     return (
       <Box>
-        <Box mb={1}>
+        <Box mb={0}>
           <SectionTitle title="Iconic Collections" />
         </Box>
```
ds300, line 24: again, you can remove the outer box

src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
```diff
@@ -108,6 +108,8 @@ const ArtistArtworksContainer: React.FC<ArtworksGridProps> = ({ artist, relay, .
   const filterParams = filterArtworksParams(state.appliedFilters)
   const artworks = artist.artworks
   const artworksTotal = artworks?.edges?.length
+  const shouldShowCollections = artist.iconicCollections && artist.iconicCollections.length > 1
+  const [shouldShowNotables, setShouldShowNotables] = useState(true)
```
dzucconi, line 112: Rather than reaching into a child component with state like this it seems like this component should just specify its own data requirements with a fragment.

ds300, line 112: This pattern where you pass the `setShouldShowNotables` down to the child is problematic on mobile since it can result in jumpy content. I think the content would indeed jump in this case, because when there aren't enough iconic collections or notable works we'd show the separator for a single frame before removing it.

Our normal pattern for solving this problem is to duplicate the query fragment in the parent component and perform the check ahead-of-time. Let me know if you need any more info or want to pair on it!

adamvert: Right, I figured it would make sense to get the data in advance, but I wasn't sure how to go about doing it. It would be great to pair on it if you have time - maybe tomorrow morning?

ds300: sure feel free to ping me tomorrow!

adamvert:
Can you take a final look at this pls @dzucconi ?

src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
```diff
@@ -172,8 +171,15 @@ const ArtistArtworksContainer: React.FC<ArtworksGridProps> = ({ artist, relay, .
...
     <>
       <Spacer mb={2} />
-      <ArtistNotableWorksRailFragmentContainer artist={artist} {...props} />
-      <ArtistCollectionsRailFragmentContainer collections={artist.iconicCollections} artist={artist} {...props} />
+      {shouldShowNotables && <ArtistNotableWorksRailFragmentContainer artist={artist} {...props} />}
+      {shouldShowCollections && (
+        <ArtistCollectionsRailFragmentContainer collections={artist.iconicCollections} artist={artist} {...props} />
+      )}
+      {(shouldShowNotables || shouldShowCollections) && (
+        <Box mx={"-20px"} mb={3} mt={1}>
```
dzucconi, line 179: Negative values should be supported OK so `m={-2}` would keep this using the space tokens.

src/lib/Components/Artist/ArtistArtworks/ArtistCollectionsRail.tsx
dzucconi: ➕

src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
```diff
@@ -108,6 +108,8 @@ const ArtistArtworksContainer: React.FC<ArtworksGridProps> = ({ artist, relay, .
   const filterParams = filterArtworksParams(state.appliedFilters)
   const artworks = artist.artworks
   const artworksTotal = artworks?.edges?.length
+  const shouldShowCollections = artist.iconicCollections && artist.iconicCollections.length > 1
```
dzucconi, line 111: `artist.iconicCollections?.length > 1` works OK as well

dzucconi:
Looks good, would deal with the minor clean up comments and needs a rebase to resolve the conflicts 💚

src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
adamvert: I tried this, but I get a type error:

```
(property) iconicCollections: readonly ({
    readonly " $fragmentRefs": FragmentRefs<"ArtistCollectionsRail_collections">;
} | null)[] | null
Object is possibly 'undefined'.ts(2532)

Comment (dzucconi) [reply src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx:111]: Ah probably like `(artist.iconicCollections?.length ?? 0 > 1)` or something, but the guard clause is clearer in this case anyway 👍

File: src/lib/Components/Artist/ArtistArtworks/ArtistNotableWorksRail.tsx [hunk 5]
```diff
@@ -50,7 +50,7 @@ const ArtistNotableWorksRail: React.FC<ArtistNotableWorksRailProps> = ({ artist
 
   return (
     <Box>
-      <Box mt={2} mb={1}>
+      <Box mt={2} mb={0}>
```
brainbicycle, line 53: Can mb just be left out here?
```suggestion
      <Box mt={2}>
```
Nitpicky/nonblocking

sweir27:
Gonna merge this in while it's 🍏 !

PR #3559, opened 2020-07-13
[MX-434] - Filter button visibility

brainbicycle:
https://artsyproduct.atlassian.net/browse/MX-434

This hides the filter button on the artist artworks page until the artist artwork are visible, it is pretty close to the implementation in the collections component: https://github.com/artsy/eigen/blob/master/src/lib/Scenes/Collection/Collection.tsx#L59

- Adds a flatlist to contain the artist artworks and rails in section
- Adds callbacks for viewableItems changed and displays filter button when filtered artworks section is visible

### DIFF
--- a/src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
+++ b/src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
@@ -62,17 +62,10 @@ const ArtworksGrid: React.FC<ArtworksGridProps> = ({ artist, relay, ...props }) 
   }
 
   const onViewRef = React.useRef(({ viewableItems }: ViewableItems) => {
-    console.log(viewableItems)
-    ;(viewableItems! ?? []).map((viewableItem: ViewToken) => {
-      const artworksRenderItem = viewableItem?.item ?? ""
-      const artworksRenderItemViewable = viewableItem?.isViewable || false
-
-      if (artworksRenderItem === "filteredArtworks" && artworksRenderItemViewable) {
-        return setArtworksGridVisible(true)
-      }
-
-      return setArtworksGridVisible(false)
+    const artworksItem = (viewableItems! ?? []).find((viewableItem: ViewToken) => {
+      return viewableItem?.item === "filteredArtworks"
     })
+    setArtworksGridVisible(artworksItem?.isViewable ?? false)
   })
 
   const viewConfigRef = React.useRef({ viewAreaCoveragePercentThreshold: 25 })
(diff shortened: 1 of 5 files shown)


src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
```diff
@@ -60,14 +61,36 @@ const ArtworksGrid: React.FC<ArtworksGridProps> = ({ artist, relay, ...props })
     handleFilterArtworksModal()
   }
 
+  const onViewRef = React.useRef(({ viewableItems }: ViewableItems) => {
+    console.log(viewableItems)
+    ;(viewableItems! ?? []).map((viewableItem: ViewToken) => {
```
ashfurrow, line 66: I'm reading this, and it feels like the implementation only works because the filteredArtworks section always comes last? If `viewableItems` are `"collections"` and `"filteredArtworks"`, then we would call `setArtworksGridVisible(false)` for `collections` and then `setArtworksGridVisible(true)` for `filteredArtworks`. Maybe we could move from a map to a filter?

```tsx
const filteredArtworksVisible = (viewableItems! ?? []).filter((viewableItem: ViewToken) => viewableItem?.isViewable && viewableItem?.item === "filteredArtworks").count > 0
setArtworksGridVisible(filteredArtworksVisible)
```

Or something like that?

src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
```diff
@@ -60,14 +61,36 @@ const ArtworksGrid: React.FC<ArtworksGridProps> = ({ artist, relay, ...props })
     handleFilterArtworksModal()
   }
 
+  const onViewRef = React.useRef(({ viewableItems }: ViewableItems) => {
+    console.log(viewableItems)
```
ashfurrow, line 65:
```suggestion
```

brainbicycle: 😬 good catch

src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
brainbicycle: 🙌 That is way simpler, I will update in Collection component as well

src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
```diff
@@ -60,14 +61,29 @@ const ArtworksGrid: React.FC<ArtworksGridProps> = ({ artist, relay, ...props })
     handleFilterArtworksModal()
   }
 
+  const onViewRef = React.useRef(({ viewableItems }: ViewableItems) => {
+    const artworksItem = (viewableItems! ?? []).find((viewableItem: ViewToken) => {
+      return viewableItem?.item === "filteredArtworks"
+    })
+    setArtworksGridVisible(artworksItem?.isViewable ?? false)
```
ashfurrow, line 68: Oh this is even better than my recommendation, nice one 🚀

PR #3535, opened 2020-07-06
[MX-109] Moving bottom tabs to TS

ds300:
Things that have changed in this PR:

- Moved bottom tabs UI out of obj-c and into TS
  - replacing the icon pngs with SVGs
  - moving analytics to TS
  - moving source of truth for the app icon badge and the messages tab badge to TS
- Set up global easy-peasy app store to track tab selection state and unread messages count.
  - This is intended to serve as an all-purpose glue layer while we migrate app infra out of obj-c and into TS+easy-peasy
  - Also fleshed out some test utils for the new app store.
- Set up events/state glue layer between obj-c and new global easy-peasy store
   - Commandeering the `ARNotificationsManager` module for this.
- Deleted a bunch of unused code obj-c code related to the things I was working on
- Added 'fade' transition between tabs
- Removed feature flags related to the sell tab and the profile tab. After this is deployed there is no going back 🙅‍♀️ so let's do a thorough prod QA of the profile tab and navigating around the app before releasing.

  - If we want to be super careful I can re-write this to all be behind an echo flag. But with the time cost + the extra complexity that would be created + the relatively low risk of us needing to flip the echo flag off... I'd vote for just forging ahead and QA-ing a bit harder than usual this time.

image

### Fade

new-fade

### New message!

pop-notification

#skip_new_tests

https://artsyproduct.atlassian.net/browse/MX-109

PR #3566, opened 2020-07-14
[MX-398] Fixes greyscale images getting displayed as red in AR VIR

ashfurrow:

This PR resolves: **MX-398**

### Description

This PR fixes a problem caused by a quirk of ARKit (Apple's Augmented Reality SDK). The problem was, as described in the comments I've left, that ARKit is interpreting the single-channel greyscale image as an RGB image. This caused it to only see the R channel. 

The problem is caused because our image processor notices that parter-uploaded artwork images don't contain non-greyscale pixels, and "helpfully" changes the images to greyscale. [Here is an example artwork](https://www.artsy.net/artwork/ronny-sen-khmer-din-14). Here is a side-by-side comparison of the original upload metadata with the metadata of the image being used by the app:

Screen Shot 2020-07-14 at 4 18 39 PM

### Test Plan

We'll QA in the beta. I've tested locally on my device. I've also included a unit test.

### Screenshots

IMG_1657


Artsy.xcodeproj/project.pbxproj
```diff
@@ -3420,7 +3425,6 @@
 				60C1C31A1C5FB00A00EC98EE /* Auctions */,
 				60E719FF1B4C60A0008948FA /* Launch */,
 				60C1C3191C5FAF6B00EC98EE /* Modal */,
-				608C6DBC1FAA10000044E235 /* Navigation */,
```
ashfurrow, line 3423: This, and a few other Xcode project changes, are due to some deleted files from #3535 (the changes were made when I ran `pod install`).

PR #3567, opened 2020-07-14
Fixes problem displaying alert after promise await

ashfurrow:
The type of this PR is: bugfix

This PR resolves: **MX-361**

### Description

I was QA'ing [MX-361] and noticed that the confirmation alert dialogue was not getting displayed. I reproduced the issue in the simulator. I noticed that [we had a hack](https://github.com/artsy/eigen/commit/2f534259cf4ed77cb762c2bb012bd25316c346d2#diff-3cc054076411de0f74147aca113dd7d2L12-L22) before, and I tried it again, and it did fix the problem. I managed to reduce the hack down and make it a little more elegant.

The thing is, I don't know why it's necessary. From my testing, it looks like `Alert.alert(...)` would only reliably work _before_ this line:

```tsx
const response = await res.json()
```

I think it might be a threading issue. But truth be told, I don't totally understand `requestAnimationFrame()` itself, either. So I'm assigning to @ds300 to get some feedback here: do you think this is related to threading? is this something we can fix another way? maybe open an issue?

[MX-361]: https://artsyproduct.atlassian.net/browse/MX-361


ds300:
Ah yes. This is to do with the fact that we present the loading spinner in a `Modal` while it's saving, and then we try to show the alert while the modal is dismissing, which there is a long-standing issue about: https://github.com/facebook/react-native/issues/10471 

I deleted the hack while cleaning up the MyProfile code because the alert had worked for me in the simulator and I forgot we were using a modal for the loading spinner, apologies for the churn.

I'm happy to keep this as-is. To follow-up, a more holistic solution would be to submit a PR to react-native fixing the underlying issue.

PR #3572, opened 2020-07-15
Viewing room artwork section

pvinis:

This PR resolves: **GALL-2878**

### Description

This PR adds a new screen to view an artwork that is in a Viewing Room.

- New route: `/viewing-room/:vrSlug/:artworkSlug`.

- We reuse `ArtworkTileRailContainer` so I added an `extra` prop to allow us to override the route on press.

- I needed to convert between cm and inches, and I saw this was done in one other place, so rather than putting hardcoded numbers and calculations, I just move that in the functions `cm2in` and `in2cm`.

- Created `LoadingScreen` and used it for the new components using relay-hooks, as well as the existing ones using `renderWithLoadProgress`.


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>Missing Test Files:

- `src/lib/Scenes/ViewingRoom/__tests__/ViewingRoomArtwork-tests.tsx`
- `src/lib/utils/__tests__/LoadingScreen-tests.tsx`

    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/Scenes/ViewingRoom/ViewingRoomArtwork.tsx
```diff
@@ -0,0 +1,235 @@
...
+      selectedArtwork.id
+    )
+  }
+
+  const tag = tagForStatus(vrInfo.status, vrInfo.distanceToOpen, vrInfo.distanceToClose)
+
+  return (
+    <ScrollView ref={navRef}>
+      <Flex>
+        <ImageCarousel images={[selectedArtwork.images![0]] as any} cardHeight={screenHeight} />
```
pvinis, line 105: a shameful `as any`, I couldn't figure out how to fix it correctly.

src/lib/Components/ArtworkTileRail/ArtworkTileRail.tsx
```diff
@@ -12,7 +12,8 @@ import { ArtworkTileRailCard } from "./ArtworkTileRailCard"
 export const ArtworkTileRailContainer: React.FC<{
   artworksConnection: ArtworkTileRail_artworksConnection
   contextModule: Schema.ContextModules
-}> = ({ artworksConnection, contextModule }) => {
+  extra?: { route: (slug: string) => string }
```
mdole, line 15: I'd suggest keeping this specific and dropping the `extra` layer so that it's just `route: (slug: string) => string` - I feel like we might just end up packing `extra` with assorted stuff 🙂

src/lib/Scenes/Artwork/Components/ArtworkActions.tsx
```diff
@@ -100,9 +101,9 @@ export class ArtworkActions extends React.Component<ArtworkActionsProps> {
       artwork: { image, id, slug, heightCm, widthCm },
     } = this.props
     // @ts-ignore STRICTNESS_MIGRATION
-    const heightIn = heightCm / 2.54
+    const heightIn = cm2in(heightCm)
```
mdole, line 104: yay for creating reusable functions! do you know if this would make it possible to drop the `ts-ignore`, or is that still a pain?

src/lib/Scenes/Artwork/Components/ArtworkHeader.tsx
```diff
@@ -1,32 +1,37 @@
...
+export const ArtworkHeader: React.FC<ArtworkHeaderProps> = props => {
+  const { artwork } = props
+  const screenDimensions = useScreenDimensions()
+
+  return (
+    <Box>
+      <Spacer mb={2} />
+      <ImageCarouselFragmentContainer
+        images={artwork.images as any /* STRICTNESS_MIGRATION */}
+        // The logic for cardHeight comes from the zeplin spec https://zpl.io/25JLX0Q
```
mdole, line 23: haha I couldn't figure out where this was coming from till I realized it was a different file!

Let's get rid of this comment - the spec it links to is archived and no longer accessible.

src/lib/utils/LoadingScreen.tsx
```diff
@@ -0,0 +1,6 @@
+import Spinner from "lib/Components/Spinner"
+import React from "react"
+
+export const LoadingTestID = "relay-loading"
+
+export const LoadingScreen = () => <Spinner testID={LoadingTestID} style={{ flex: 1 }} />
```
mdole, line 6: nice thought to turn this into its own component

src/lib/Components/ArtworkTileRail/ArtworkTileRail.tsx
```diff
@@ -32,7 +33,11 @@ export const ArtworkTileRailContainer: React.FC<{
           <ArtworkTileRailCard
             onPress={() => {
               tracking.trackEvent(tappedArtworkGroupThumbnail(contextModule, item.internalID, item.slug))
-              SwitchBoard.presentNavigationViewController(navRef.current!, item.href!)
+              {
+                !!extra?.route
+                  ? SwitchBoard.presentNavigationViewController(navRef.current!, extra.route(item.slug))
+                  : SwitchBoard.presentNavigationViewController(navRef.current!, item.href!)
+              }
```
ds300, line 40: this `extra/route` thing is a strange API. How about adding an `onTilePress` prop which, if provided, gets called here instead of SwitchBoard?

src/lib/utils/LoadingScreen.tsx
ds300: we are actively replacing this old loading spinner with 'skeleton' placeholder views, and it would be good to avoid shipping this. Do you have follow-up work scheduled to create a proper loading state?

src/lib/utils/conversions.ts
```diff
@@ -0,0 +1,3 @@
+export const in2cm = (inches: number) => inches * 2.54
+
+export const cm2in = (cm: number) => cm / 2.54
```
ds300, line 3: 🙌

src/lib/Scenes/ViewingRoom/ViewingRoomArtwork.tsx
```diff
@@ -0,0 +1,244 @@
...
+  const artworks = extractNodes(artworksList.artworksConnection)
+  const vrInfo = useFragment(viewingRoomInfoFragmentSpec, props.viewingRoomInfo)
+
+  const navRef = useRef(null)
+  const { height: screenHeight } = useScreenDimensions()
+
+  const viewInAR = () => {
+    const [widthIn, heightIn] = [selectedArtwork.widthCm!, selectedArtwork.heightCm!].map(cm2in)
+
+    // @ts-ignore STRICTNESS_MIGRATION
```
ds300, line 91: you shouldn't be adding new STRICTNESS_MIGRATION comments

emission/Pod/Classes/ViewControllers/ARViewingRoomArtworkComponentViewController.h
```diff
@@ -0,0 +1,20 @@
...
+                            artworkID:(NSString *)artworkID
+                             emission:(nullable AREmission *)emission NS_DESIGNATED_INITIALIZER;
+
+- (instancetype)initWithEmission:(nullable AREmission *)emission
+                      moduleName:(NSString *)moduleName
+               initialProperties:(nullable NSDictionary *)initialProperties NS_UNAVAILABLE;
+
+@end
+
+NS_ASSUME_NONNULL_END
```
ds300, line 20: for future reference we don't need to create new classes when adding new screens any more. you can just create a `ARComponentViewController` directly and then pass the route parameters in as props to the registered component. See, e.g. all the MyProfile/MyAccount routes.

pvinis:
I resolved all the comments. Thank you both for the good ideas in here! 💜

src/lib/Scenes/ViewingRoom/ViewingRoomArtwork.tsx
MounirDhahri: I think it could be `ImageCarousel_images` from [here](https://github.com/artsy/eigen/blob/a3cc6ac3410617d37f78f4c9cc19b5917431185f/src/lib/Scenes/Artwork/Components/ImageCarousel/ImageCarousel.tsx#L3)

pvinis: That's what it is, and I did try to make it right, but I failed. Maybe we could take a look at some point later, if you're feeling like fighting with ts 😅!

MounirDhahri: aha not at all, it's not important

MounirDhahri:
nice one 🎉

pvinis:
Merging this as it's been accepted and addressed. Also, I want it in master before the QA meeting today, so we can possibly use it as an example ✋.

PR #3577, opened 2020-07-16
Remove sales lab option

ds300:

### Description

Following up #3535 , removing the now-defunct sales lab option.


src/lib/Scenes/Consignments/__tests__/index-tests.tsx
```diff
@@ -8,6 +8,8 @@ import { Consignments } from "../"
 
 import { Theme } from "@artsy/palette"
 
+jest.unmock("react-relay")
```
pepopowitz, line 11: Question: Were there any noticeable side-effects of not doing this, or was this added as a precaution?

src/lib/Scenes/Consignments/index.tsx
```diff
@@ -1,8 +1,5 @@
 import { ConsignmentSubmissionCategoryAggregation } from "__generated__/createConsignmentSubmissionMutation.graphql"
 import React from "react"
-import { NativeModules, ViewProperties } from "react-native"
-import NavigatorIOS, { Route } from "react-native-navigator-ios"
-import Welcome from "./Screens/Welcome"
```
pepopowitz, line 5: I suspect this removal also means we could delete the `Welcome` component (and other related components). Heads up that [we have an upcoming ticket in our backlog](https://artsyproduct.atlassian.net/browse/CSGN-256) to take care of this. (And thank you for doing this work to remove the flag!)

src/lib/Scenes/Consignments/__tests__/index-tests.tsx
ds300: yeah the test failed because now this component tries to mount a `QueryRenderer` which is undefined at test time without this line

Issue #3562, opened 2020-07-14
[RFC] New Dependency: `relay-hooks`


pvinis:
### New Dependency

Name: `relay-hooks`

URL: https://github.com/relay-tools/relay-hooks

### Motivation

We use relay with `react-relay` already, and it works great. We use `createFragmentContainer` and `QueryRenderer` to make our components.

We already use hooks if a bunch of places and we are moving towards them more and more with new code.

`relay-hooks` is the bridge of these two "worlds". It provides hooks like `useFragment` and `useQuery` that replace the functions I mentioned earlier. This makes our components cleaner. We create a component and use hooks to connect it with relay, rather than having relay create and manage all that for us.

Look at this:
```
export const ViewingRoomQueryRenderer: React.FC<{ viewingRoomID: string }> = ({ viewingRoomID }) => {
  return (
    <QueryRenderer<ViewingRoomQuery>
      environment={defaultEnvironment}
      query={graphql`
        query ViewingRoomQuery($viewingRoomID: ID!) {
          viewingRoom(id: $viewingRoomID) {
            ...ViewingRoom_viewingRoom
          }
        }
      `}
      cacheConfig={{ force: true }}
      variables={{
        viewingRoomID,
      }}
      render={renderWithLoadProgress(ViewingRoomFragmentContainer)}
    />
  )
}
```
where our component is basically a relay component that we customize and adjust using the props.

Now look at this:
```
const query = graphql`
  query ViewingRoomQuery($viewingRoomID: ID!) {
    viewingRoom(id: $viewingRoomID) {
      ...ViewingRoom_viewingRoom
    }
  }
`

export const ViewingRoomQueryRenderer: React.FC<{ viewingRoomID: string }> = ({ viewingRoomID }) => {
  const { props, error } = useQuery(query, { viewingRoomID }, { networkCacheConfig: { force: true } })


  if (props && props.viewingRoom) {
    return <ViewingRoomFragmentContainer viewingRoom={props.viewingRoom} />
  }
  if (error) {
    throw error
  }

  return <LoadingScreen />
}
```
where our component is actually our code, and we control the flow from what we get from the hook.

[...]

PR #3587, opened 2020-07-17
[FX-2010] Add artworks to Artist Series

ashleyjelks:

This PR resolves: [FX-2010](https://artsyproduct.atlassian.net/browse/FX-2010) and [FX-2050](https://artsyproduct.atlassian.net/browse/FX-2050)

### Description

Implements an infinite scroll artwork grid on the Artist Series view.  In order to implement this feature I had to refactor the Artist Series "home page" to use a `Flatlist`.  This refactor involved breaking out the existing components into separate files (including test files). I also added a header image url to the Artist Series Header, which takes care of  [FX-2050](https://artsyproduct.atlassian.net/browse/FX-2050).

### Test Plan
 
@ashleyjelks  and @nicoleyeo will QA the new changes in a beta.

### Screenshots

Kapture 2020-07-17 at 16 40 21

### DIFF
--- a/src/lib/Scenes/ArtistSeries/ArtistSeries.tsx
+++ b/src/lib/Scenes/ArtistSeries/ArtistSeries.tsx
@@ -1,13 +1,12 @@
-import { Box, Flex, Spacer, Theme } from "@artsy/palette"
+import { Box, Spacer, Theme } from "@artsy/palette"
 import { ArtistSeries_artistSeries } from "__generated__/ArtistSeries_artistSeries.graphql"
 import { ArtistSeriesQuery } from "__generated__/ArtistSeriesQuery.graphql"
 import { AboveTheFoldFlatList } from "lib/Components/AboveTheFoldFlatList"
 import { defaultEnvironment } from "lib/relay/createEnvironment"
-import { ArtistSeriesHeader } from "lib/Scenes/ArtistSeries/ArtistSeriesHeader"
+import { ArtistSeriesHeaderFragmentContainer } from "lib/Scenes/ArtistSeries/ArtistSeriesHeader"
 import { ArtistSeriesMetaFragmentContainer } from "lib/Scenes/ArtistSeries/ArtistSeriesMeta"
 import renderWithLoadProgress from "lib/utils/renderWithLoadProgress"
 import React from "react"
-import { View } from "react-native"
 import { createFragmentContainer, graphql, QueryRenderer } from "react-relay"
 interface ArtistSeriesProps {
   artistSeries: ArtistSeries_artistSeries
@@ -22,7 +21,7 @@ export const ArtistSeries: React.FC<ArtistSeriesProps> = ({ artistSeries }) => {
           showsVerticalScrollIndicator={false}
           initialNumToRender={3}
           data={sections}
-          ListHeaderComponent={<ArtistSeriesHeader />}
+          ListHeaderComponent={<ArtistSeriesHeaderFragmentContainer artistSeries={artistSeries} />}
           ItemSeparatorComponent={() => <Spacer mb={2} />}
           keyExtractor={(_item, index) => String(index)}
           renderItem={({ item }): null | any => {
@@ -41,6 +40,7 @@ export const ArtistSeriesFragmentContainer = createFragmentContainer(ArtistSerie
   artistSeries: graphql`
     fragment ArtistSeries_artistSeries on ArtistSeries {
       ...ArtistSeriesMeta_artistSeries
+      ...ArtistSeriesHeader_artistSeries
     }
   `,
 })
--- a/src/lib/Scenes/ArtistSeries/ArtistSeries.tsx
+++ b/src/lib/Scenes/ArtistSeries/ArtistSeries.tsx
@@ -3,17 +3,19 @@ import { ArtistSeries_artistSeries } from "__generated__/ArtistSeries_artistSeri
 import { ArtistSeriesQuery } from "__generated__/ArtistSeriesQuery.graphql"
 import { AboveTheFoldFlatList } from "lib/Components/AboveTheFoldFlatList"
 import { defaultEnvironment } from "lib/relay/createEnvironment"
+import { ArtistSeriesArtworksFragmentContainer } from "lib/Scenes/ArtistSeries/ArtistSeriesArtworks"
 import { ArtistSeriesHeaderFragmentContainer } from "lib/Scenes/ArtistSeries/ArtistSeriesHeader"
 import { ArtistSeriesMetaFragmentContainer } from "lib/Scenes/ArtistSeries/ArtistSeriesMeta"
 import renderWithLoadProgress from "lib/utils/renderWithLoadProgress"
 import React from "react"
+
 import { createFragmentContainer, graphql, QueryRenderer } from "react-relay"
 interface ArtistSeriesProps {
   artistSeries: ArtistSeries_artistSeries
 }
 
 export const ArtistSeries: React.FC<ArtistSeriesProps> = ({ artistSeries }) => {
-  const sections = ["artistSeriesMeta", "artistSeriesArtwork"]
+  const sections = ["artistSeriesMeta", "artistSeriesArtworks"]
   return (
     <Theme>
       <Box px={2}>
@@ -24,11 +26,14 @@ export const ArtistSeries: React.FC<ArtistSeriesProps> = ({ artistSeries }) => {
           ListHeaderComponent={<ArtistSeriesHeaderFragmentContainer artistSeries={artistSeries} />}
           ItemSeparatorComponent={() => <Spacer mb={2} />}
           keyExtractor={(_item, index) => String(index)}
-          renderItem={({ item }): null | any => {
+          renderItem={({ item }) => {
             switch (item) {
               case "artistSeriesMeta":
                 return <ArtistSeriesMetaFragmentContainer artistSeries={artistSeries} />
+              case "artistSeriesArtworks":
+                return <ArtistSeriesArtworksFragmentContainer artistSeries={artistSeries} />
             }
+            return null
           }}
         />
       </Box>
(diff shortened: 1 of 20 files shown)


src/lib/Scenes/ArtistSeries/ArtistSeries.tsx
```diff
@@ -1,117 +1,51 @@
...
-                      </TouchableWithoutFeedback>
-                    }
-                  />
-                </TouchableOpacity>
-              )
-          )}
-        <ReadMore data-test-id="description" sans content={artistSeries?.description ?? ""} maxChars={maxChars} />
+    <Theme>
+      <Box px={2}>
+        <AboveTheFoldFlatList
```
ashleyjelks, line 22: Got stuck on adding the correct type to this Flatlist. Will reach out to @ds300 or @pvinis next week.

src/lib/Scenes/ArtistSeries/ArtistSeriesArtworks.tsx
```diff
@@ -0,0 +1,83 @@
...
+import { createPaginationContainer, graphql, RelayPaginationProp } from "react-relay"
+
+interface ArtistSeriesArtworksProps {
+  artistSeries: ArtistSeriesArtworks_artistSeries
+  relay: RelayPaginationProp
+}
+export const ArtistSeriesArtworks: React.FC<ArtistSeriesArtworksProps> = ({ artistSeries, relay }) => {
+  const artworks = artistSeries?.artistSeriesArtworks!
+
+  if (artworks?.counts?.total === 0) {
```
ashleyjelks, line 14: Might be worth following up with @nicoleyeo to see if there's a Zero State spec here for the (likely rare) instance when an Artist Series has no artworks, but still want to exit early to avoid rendering the `InfiniteScrollArtworksGridContainer` with no data.

src/lib/Scenes/ArtistSeries/ArtistSeriesArtworks.tsx
```diff
@@ -0,0 +1,83 @@
...
+        connection={artworks}
+        loadMore={relay.loadMore}
+        hasMore={relay.hasMore}
+        isLoading={relay.isLoading}
+      />
+    </Box>
+  )
+}
+
+export const ArtistSeriesArtworksFragmentContainer = createPaginationContainer(
```
ashleyjelks, line 31: I saw the Relay hooks RFC. https://github.com/artsy/eigen/issues/3562 A good followup here could be to use those instead.

src/lib/Scenes/ArtistSeries/ArtistSeriesHeader.tsx
```diff
@@ -0,0 +1,36 @@
...
+    <Flex flexDirection="row" justifyContent="center">
+      <OpaqueImageView width={180} height={180} imageURL={url} />
+    </Flex>
+  )
+}
+
+export const ArtistSeriesHeaderFragmentContainer = createFragmentContainer(ArtistSeriesHeader, {
+  artistSeries: graphql`
+    fragment ArtistSeriesHeader_artistSeries on ArtistSeries {
+      imageURL
```
ashleyjelks, line 25: I need to confirm with @mzikherman that this is the correct MP field for the admin uploaded Artist Series header.

src/lib/Scenes/ArtistSeries/ArtistSeriesMeta.tsx
```diff
@@ -0,0 +1,112 @@
...
+    })
+  }
+
+  return (
+    <Box ref={ref}>
+      <Sans size="8" mt={3} data-test-id="title">
+        {artistSeries.title}
+      </Sans>
+      {(artistSeries?.artists ?? []).length > 0 &&
+        artistSeries.artists?.map(
```
williardx, line 58: Minor, but given that we clarified that this design is only for one artist it might be clearer to just take the first artist we get instead of mapping over all artists. That'd also make it consistent with the behavior in Force.

src/lib/Scenes/ArtistSeries/ArtistSeriesArtworks.tsx
```diff
@@ -0,0 +1,83 @@
...
+import { InfiniteScrollArtworksGridContainer } from "lib/Components/ArtworkGrids/InfiniteScrollArtworksGrid"
+import React from "react"
+import { createPaginationContainer, graphql, RelayPaginationProp } from "react-relay"
+
+interface ArtistSeriesArtworksProps {
+  artistSeries: ArtistSeriesArtworks_artistSeries
+  relay: RelayPaginationProp
+}
+export const ArtistSeriesArtworks: React.FC<ArtistSeriesArtworksProps> = ({ artistSeries, relay }) => {
+  const artworks = artistSeries?.artistSeriesArtworks!
```
williardx, line 12: I didn't know about TypeScript's post-fix `!`, cool!

src/lib/Scenes/ArtistSeries/ArtistSeries.tsx
```diff
@@ -1,117 +1,52 @@
...
-          } else {
-            resolve()
-          }
-        },
-      })
-    })
-  }
-
+export const ArtistSeries: React.FC<ArtistSeriesProps> = ({ artistSeries }) => {
+  const sections = ["artistSeriesMeta", "artistSeriesArtwork"]
```
anandaroop, line 18: #minor should this be `artistSeriesArtworks` (plural)?

src/lib/Scenes/ArtistSeries/ArtistSeriesHeader.tsx
ashleyjelks: Just check Gravity's GravQL schema and this _is_ the field intended for the header image!

src/lib/Scenes/ArtistSeries/ArtistSeriesArtworks.tsx
sweir27: This seems like a fine/graceful solution for now!

src/lib/Scenes/ArtistSeries/ArtistSeriesMeta.tsx
ashleyjelks: I thought we were supposed to handle the (likely rare) instance where there are multiple artists in a series.

src/lib/Scenes/ArtistSeries/ArtistSeriesHeader.tsx
sweir27: I _think_ we actually want to use the `image` field here (which I'm guessing is what is hooked up to the fallback option)--

image

src/lib/Scenes/ArtistSeries/ArtistSeriesHeader.tsx
```diff
@@ -0,0 +1,36 @@
...
+import React from "react"
+import { createFragmentContainer, graphql } from "react-relay"
+
+interface ArtistSeriesHeaderProps {
+  artistSeries: ArtistSeriesHeader_artistSeries
+}
+export const ArtistSeriesHeader: React.SFC<ArtistSeriesHeaderProps> = ({ artistSeries }) => {
+  const url = artistSeries.imageURL
+    ? artistSeries.imageURL
+    : artistSeries.filterArtworksConnection?.edges?.[0]?.node?.imageUrl
```
sweir27, line 13: I don't think we need this logic here (assuming that https://github.com/artsy/metaphysics/pull/2531 is handling it for us).

src/lib/Scenes/ArtistSeries/ArtistSeriesMeta.tsx
```diff
@@ -0,0 +1,108 @@
...
+    </View>
+  )
+}
+
+export const ArtistSeriesMetaFragmentContainer = createFragmentContainer(ArtistSeriesMeta, {
+  artistSeries: graphql`
+    fragment ArtistSeriesMeta_artistSeries on ArtistSeries {
+      title
+      description
+      artists(size: 1) {
```
sweir27, line 96: Since you're restricting it to 1 artist here, I agree with @williardx above-- you can avoid mapping and just grab the `first` artist!

src/lib/Scenes/ArtistSeries/ArtistSeriesHeader.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+export const ArtistSeriesHeaderFragmentContainer = createFragmentContainer(ArtistSeriesHeader, {
+  artistSeries: graphql`
+    fragment ArtistSeriesHeader_artistSeries on ArtistSeries {
+      imageURL
+      filterArtworksConnection(first: 1, sort: "-merchandisability") {
+        edges {
+          node {
+            title
+            imageUrl
```
anandaroop, line 30: This could probably change, due to recent work on Artist Series images.

We now have an `image` field that will resolve to _either_ the automatically selected artwork image _or_ the admin-uploaded image. So clients don't need to worry about selecting the right one. 

Example query:

Screen Shot 2020-07-20 at 1 46 34 PM

Here `artistSeries.image.url` is the correct "raw" image and `artistSeries.image.resized.url` is the correct on-the-fly resized one.

(Note to us: This would probably make a good addition to the inline docs in MP for these image related fields, which are looking kind of bare on the right hand side there!)

src/lib/Scenes/ArtistSeries/ArtistSeriesHeader.tsx
anandaroop: See my other comment about images — we should be able to use the `image` field instead

src/lib/Scenes/ArtistSeries/ArtistSeriesHeader.tsx
ashleyjelks: Agreed, it isn't immediately clear which url field is the correct one to use!

PR #3588, opened 2020-07-20
Small things #trivial

pvinis:

### Description

Three small things in here:

- Changing the way types work in `useQuery`.
There was an issue that caused us to add a patch for relay-hooks. After reporting it, a new version with the fix will come out soon, and with this change we can just remove the patch when we update to the new relay-hooks version, and all should be good.
More context: https://github.com/relay-tools/relay-hooks/pull/101

- I was getting a buuuuunch of warnings about ios 8 in the pods, moving to 9 removed all these.

- There were a few files in Xcode that were referenced but don't exists, so I think they are just remnants of a deletion. I removed these too.


Podfile
```diff
@@ -199,12 +199,12 @@ post_install do |installer|
...
 
-  # Forces the minimum to be 8.0 as that's our last deployment targer, and new xcode build tools
+  # Forces the minimum to be 9.0 as that's our last deployment targer, and new xcode build tools
   # give an error in Xcode 10
   installer.pods_project.targets.each do |target|
     target.build_configurations.each do |config|
-      if config.build_settings['IPHONEOS_DEPLOYMENT_TARGET'].to_f < 8.0
-        config.build_settings['IPHONEOS_DEPLOYMENT_TARGET'] = '8.0'
+      if config.build_settings['IPHONEOS_DEPLOYMENT_TARGET'].to_f < 9.0
+        config.build_settings['IPHONEOS_DEPLOYMENT_TARGET'] = '9.0'
```
ashfurrow, line 207: 👍

src/lib/Scenes/ViewingRoom/ViewingRoomArtwork.tsx
```diff
@@ -237,13 +238,17 @@ export const ViewingRoomArtworkQueryRenderer: React.FC<{ viewing_room_id: string
   viewing_room_id: viewingRoomID,
   artwork_id: artworkID,
 }) => {
-  const { props, error } = useQuery(query, { viewingRoomID, artworkID }, { networkCacheConfig: { force: true } })
+  const { props, error } = useQuery<ViewingRoomArtworkQuery>(
```
ashfurrow, line 241: Oh very cool.

PR #3595, opened 2020-07-20
Viewing room featured rail

pvinis:

This PR resolves **[GALL-3040]**

### Description

Adding the top part of the VR list screen, the featured rail.

One thing to note is that for the medium cards I gave `372` pixels height. I guess the way I added the cards in palette are not the perfect size. I can make a PR later to fix them. We are kinda unsure about how to handle the sizing of the medium cards, so for now it's all good.

I also changed the text in the dev menu, because we had things like `Show Viewing Rooms` to **navigate** there, and `Show Viewing Rooms` to **enable** the feature. I thought it's simpler to have an arrow for navigation instead.

### Screenshots

Screenshot 2020-07-20 at 22 04 39
Screenshot 2020-07-20 at 22 04 36

[GALL-3040]: https://artsyproduct.atlassian.net/browse/GALL-3040


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>Missing Test Files:

- `src/lib/Scenes/ViewingRoom/Components/__tests__/ViewingRoomsListFeatured-tests.tsx`

    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/Scenes/ViewingRoom/Components/ViewingRoomsListFeatured.tsx
```diff
@@ -0,0 +1,83 @@
...
+    <Flex flexDirection="row">
+      {_.times(2).map(() => (
+        <PlaceholderBox width={280} height={370} marginRight={15} />
+      ))}
+    </Flex>
+  </ProvidePlaceholderContext>
+)
+
+export const FeaturedRail = () => {
+  const { props, error } = useQuery<ViewingRoomsListFeaturedQuery>(query)
```
ds300, line 44: seems like it would be better to use a fragment here? Otherwise we'd end up with staggered loading on whatever page this is shown on. Given that this rail is above-the-fold on the viewing rooms list I think it makes sense to fetch it's data with the rest of the page to show it as early as possible.

src/lib/Scenes/ViewingRoom/Components/ViewingRoomsListFeatured.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const navRef = useRef(null)
+
+  if (props?.viewingRooms) {
+    const featured = extractNodes(props.viewingRooms)
+    return (
+      <FlatList
```
ds300, line 50: Let's change this to use an `AboveTheFoldFlatList` to avoid rendering too many images and hurting initial page load times.

src/lib/Scenes/ViewingRoom/Components/ViewingRoomsListFeatured.tsx
```diff
@@ -0,0 +1,83 @@
...
+        showsHorizontalScrollIndicator={false}
+        keyExtractor={item => item.internalID}
+        data={featured}
+        ItemSeparatorComponent={() => <Spacer ml={15} />}
+        renderItem={({ item }) => {
+          const tag = tagForStatus(item.status, item.distanceToOpen, item.distanceToClose)
+          return (
+            <TouchableHighlight
+              onPress={() =>
+                void SwitchBoard.presentNavigationViewController(navRef.current!, `/viewing-room/${item.slug!}`)
```
ds300, line 64:
```suggestion
               SwitchBoard.presentNavigationViewController(navRef.current!, `/viewing-room/${item.slug!}`)
```

src/lib/Scenes/ViewingRoom/ViewingRoomsList.tsx
```diff
@@ -84,10 +88,16 @@ export const ViewingRoomsListContainer: React.FC<ViewingRoomsListProps> = props
...
+                <Spacer mt={15} />
+              </Flex>
               <FeaturedRail />
-              <Sans size="4t">Latest</Sans>
-              <Spacer mt={1} />
+              <Flex mx="2">
+                <Spacer mt="4" />
+                <Sans size="4t">Latest</Sans>
+                <Spacer mt={15} />
+              </Flex>
```
ds300, line 100: we have a standard component for this called `SectionTitle`

pvinis:
Fixing things now. Thanks!

For the rail using a fragment, I'll give it another try. I started trying that, but gave up when the code because too messy for what we need. The fact that the query on that screen is just to spread the fragment with the root field `viewingRooms` makes typing a bit funky.

pvinis:
All issues addressed! Should be good to go.

pvinis:
ill google how we can make changelog conflict less.. :(

ashfurrow:
It's tricky, we already have this:

https://github.com/artsy/eigen/blob/172ba188823d62236a0b613914d4ccdc666be057/.gitattributes#L1

But it's not perfect.

Will review this PR shortly!

Artsy/View_Controllers/Admin/ARAdminSettingsViewController.m
```diff
@@ -273,7 +273,7 @@ - (ARCellData *)generateOnScreenMartsy
 
 - (ARCellData *)generateEchoContents
 {
-    return [self tappableCellDataWithTitle:@"View Echo Configuration" selection:^{
+    return [self tappableCellDataWithTitle:@"→ Echo Configuration" selection:^{
```
ashfurrow, line 276: Yeah, these changes make sense 👍

src/lib/Scenes/ViewingRoom/Components/ViewingRoomsListFeatured.tsx
```diff
@@ -0,0 +1,76 @@
...
+        listRef={listRef}
+        showsHorizontalScrollIndicator={false}
+        initialNumToRender={2}
+        keyExtractor={item => item.internalID}
+        data={featured}
+        ItemSeparatorComponent={() => <Spacer ml={15} />}
+        renderItem={({ item }) => {
+          const tag = tagForStatus(item.status, item.distanceToOpen, item.distanceToClose)
+          return (
+            <TouchableHighlight
```
ashfurrow, line 60: Eventually, hopefully soon, we should factor out the `TouchableHighligh` with these props into its own reusable component 👍

src/lib/Scenes/ViewingRoom/ViewingRoomsList.tsx
```diff
@@ -40,20 +42,32 @@ const useNumColumns = () => {
...
+    setRefreshing(true)
+    refetchConnection(connectionConfig, PAGE_SIZE)
+    setRefreshing(false)
+    scrollRefIPhone.current?.scrollToTop()
+    scrollRefIPad.current?.scrollToTop()
+  }
+
   const numColumns = useNumColumns()
+  const scrollRefIPhone = useRef<RailScrollRef>(null)
+  const scrollRefIPad = useRef<RailScrollRef>(null)
```
ashfurrow, line 70: Hmm. I see what's happening here, we're tying the number of columns to the iPhone/iPad device, and then using one component tree or another. But if we only ever use one or the other, could we use just one ref? Since only one or the other will be included, and never both?

src/lib/Scenes/ViewingRoom/Components/ViewingRoomsListFeatured.tsx
ds300: 🤔 isn't this behaving the same as `TouchableOpacity`? Assuming the component is on a white background, that is.

pvinis: To be honest, I just took that from another file and pasted it here. I saw it was in a few places, so we could move it into its own thing, but I didn't want to do something about it here.

I'm ok with whatever you guys think. I can leave it as it, I can replace it with `TouchableOpacity`, or I can make a styled-component out of it.

src/lib/Scenes/ViewingRoom/ViewingRoomsList.tsx
pvinis: Yup, that makes sense. In my mind I thought, "If I use one in both places, something will complain for reusing something that should be used once, being a ref. How could it know it will only ever be used once.".

Then I thought that if somehow we ever remove the ternary and we have both flatlists with the same ref, we might not get a complaint, but only one will work. But having both flatlists will obviously never happen.

All that is to say, I guess I was tired and I did a bit of a weird overthinking of this ref 😅. Good catch.

src/lib/Scenes/ViewingRoom/Components/ViewingRoomsListFeatured.tsx
ashfurrow: Good catch, I didn't even realize we had both! I've opened [MX-456] to update these throughout the app to be consistent, but I don't think we need to block this PR for now.

PR #3604, opened 2020-07-21
Fixes problem selecting home tab initially

ashfurrow:
The type of this PR is: **bugfix**

### Description

Here's the bug:

- Open the app. It shows the home tab.
- Without tapping into another tab, tap the home tab icon.

What happens is we see a blank screen, and the other tabs are unresponsive. 

The cause of the bug is that the `currentTab` property isn't initially set to anything, leading the comparison to check if the user has selected the  current tab to fail (comparison to `nil`). If the user selects another tab, then the bug doesn't appear (the `currentTab` property has been set).

I also fixed a stylistic Objective-C issue with a `==` versus `isEqual:`. `==` is a pointer comparison, but happens to succeed because Objective-C string literals are hoisted into static memory, so every Objective-C `@"home"` string literal _is_ the identical `NSString` instance. 

We should probably release a hot fix with this tomorrow; it probably won't affect many users, but the app becomes unresponsive until it's restarted.

### Test Plan

Let's test on device, too.


ds300:
Thanks for finding and fixing this! Yikes!

PR #3608, opened 2020-07-22
Move card in eigen and fix

pvinis:

### Description

This PR tries to fix two bugs, and at the same time brings the problematic components to eigen.

- Bug 1: the medium card was not fixed size, so the featured list had to have some extra height (372 instead of 370). Now the card is fixed size, as spec-ed, and no need for extra height.

- Bug 2: on ipad, when the bottom row of the list was less than full (not all columns had items), the size of those bottom row items would flex to get all the available space, like here:
Screenshot 2020-07-22 at 19 28 22
Screenshot 2020-07-22 at 19 28 29
Fixed:
Screenshot 2020-07-22 at 19 24 57
Screenshot 2020-07-22 at 19 24 50
The way this works for FlatLists with multiple columns, is that each item is wrapped in a `Flex` with `flex={1 / numColumns}`, so that no more than the right space is taken. The other part of the fix here is that inside that `Flex` comes the `Spacer` for left, a `Flex` for the actual content, and the `Spacer` for right.

- Lastly, I moved the cards in the future place for palette and imported them like `import { Card } from "palette"`, as we will be using it in the future. For now, I changed the imports inside the cards to point to the current palette. We will change it to import from local when we move everything else too.


src/lib/Scenes/ViewingRoom/ViewingRoomsList.tsx
```diff
@@ -121,14 +121,15 @@ export const ViewingRoomsListContainer: React.FC<ViewingRoomsListProps> = props
           refreshControl={<RefreshControl refreshing={refreshing} onRefresh={handleRefresh} />}
           keyExtractor={item => `${item.internalID}-${numColumns}`}
           renderItem={({ item, index }) => (
-            <>
-              {index % numColumns > 0 && <Spacer ml={2} />}
+            <Flex flex={1 / numColumns} flexDirection="row">
```
mdole, line 124: 🧠 💥  very cool!

pvinis:
Ugh. I need to make a ci step to try to fix all these things, that the test fails but are fixable.

src/lib/Scenes/ViewingRoom/ViewingRoomsList.tsx
```diff
@@ -121,14 +121,15 @@ export const ViewingRoomsListContainer: React.FC<ViewingRoomsListProps> = props
...
+              {index % numColumns > 0 && <Spacer ml="1" />}
               <Flex flex={1}>
                 <ViewingRoomsListItem item={item} />
               </Flex>
-            </>
+              {index % numColumns < numColumns - 1 && <Spacer mr="1" />}
```
ashfurrow, line 129: Not related to this PR, but I'm curious why this is needed in addition to the `ItemSeparatorComponent` prop below?

src/palette/package.json
```diff
@@ -0,0 +1,3 @@
+{
+  "name": "palette"
```
ashfurrow, line 2: This is a clever workaround while we deal with porting Palette over into Eigen properly, really neat!

src/lib/Scenes/ViewingRoom/ViewingRoomsList.tsx
pvinis: So the way FlatList works is a bit funky. The `ItemSeparatorComponent` is the component that goes below the first item and above the second item. If the FlatList has 3 columns, the `ItemSeparatorComponent` appears below the first **row** and above the second **row**, so the items per row are considered as one thing. That's the reason we need both that, and the spacers.

Does that explain it? If not, we can do a little more in depth explanation in the next KS?

ashfurrow: Ahhhh I see. Yeah that makes sense 👍

PR #3609, opened 2020-07-23
[My Collection] Cleanup AddArtwork screen 

damassi:

### Description

This PR cleans up the Add Artwork screen and brings it in line with the comps, as well as:
- Fixes `NavigatorIOS` structure to account for new FancyModal component 
- Adds a new `FancyModalHeader` component for sharing common title, backButton, etc layout, and updates the collection filter as well as various modal headers in My Collection
-  Adds a subtitle to the `<Select>` component
- Adjusts `title` font weight to `medium` on `Select` and `Input`, but need to follow up with design about this change, which matches latest form comps. If we change this here a few other spots in the app will also update (but not many, since `Select` and latest `Input` are new). Thoughts @ds300?
- Updates modal backdrop presses to tap into "is form dirty" checks, and present confirmation action sheet 
- Swaps `<Buttons>` with `<TouchableOpacity>` 

### Screenshots

addartwork

#skip_new_tests


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>Missing Test Files:

- `src/lib/Components/FancyModal/__tests__/FancyModalHeader-tests.tsx`

    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/AppRegistry.tsx
```diff
@@ -35,7 +35,7 @@ import { MyCollectionHome } from "./Scenes/Consignments/v2/Screens/MyCollectionH
 import { MyCollectionMarketingHome } from "./Scenes/Consignments/v2/Screens/MyCollectionHome/MyCollectionMarketingHome"
 import { SellTabApp } from "./Scenes/Consignments/v2/SellTabApp"
 
-import { _FancyModalPageWrapper } from "./Components/FancyModal"
+import { _FancyModalPageWrapper } from "./Components/FancyModal/FancyModal"
```
damassi, line 38: @ds300 - I think I asked this before but didn't see the response -- what does this `_` signify?

src/lib/Scenes/Consignments/v2/State/artworkModel.tsx
```diff
@@ -23,8 +23,8 @@ export interface ArtworkFormValues {
 const initialFormValues: ArtworkFormValues = {
   artist: "",
   artistSearchResult: null,
-  medium: "",
-  size: "",
+  medium: "PAINTING",
+  size: "medium",
```
damassi, line 27: This sets the defaults for our selects

src/lib/Components/FancyModal/FancyModalHeader.tsx
```diff
@@ -0,0 +1,46 @@
...
+export const FancyModalHeader: React.FC<{
+  backButtonText?: string
+  onBackPress?: () => void
+}> = ({ children, backButtonText, onBackPress }) => {
+  return (
+    <Container>
+      <Flex mt={0.5} flexDirection="column" alignContent="center" alignItems="center">
+        {!!onBackPress && (
+          <NavigateBackContainer
+            hitSlop={{ top: space(1), bottom: space(1), left: space(1), right: space(1) }}
```
jonallured, line 15: `hitSlop`?? I LOLed.

src/lib/Components/FancyModal/FancyModalHeader.tsx
```diff
@@ -0,0 +1,46 @@
...
+            onPress={() => onBackPress()}
+          >
+            {backButtonText ? <Sans size="3">{backButtonText}</Sans> : <ArrowLeftIcon fill="black100" top="2px" />}
+          </NavigateBackContainer>
+        )}
+      </Flex>
+      <Sans mt={2} weight="medium" size="4" color="black100">
+        {children}
```
pepopowitz, line 23: **question [non-blocking]**: Since we're expecting `children` to always be a text element, should we specify that in the interface on lines 7-8?

src/lib/AppRegistry.tsx
ds300: 'implementation detail, do not use'

src/lib/Scenes/Consignments/v2/Screens/MyCollectionAddArtwork/Form/artworkSchema.tsx
```diff
@@ -3,7 +3,7 @@ import { ArtworkFormValues } from "lib/Scenes/Consignments/v2/State/artworkModel
 import * as Yup from "yup"
 
 export const artworkSchema = Yup.object().shape({
-  artist: Yup.string().test("artist", "Artist must be cindy sherman", value => value === "Cindy Sherman"),
+  artist: Yup.string().test("artist", "Artist must not be empty", value => value !== ""),
```
jonallured, line 6: Haha, much better 😝

src/lib/Scenes/Consignments/v2/State/artworkModel.tsx
jonallured: And I assume these defaults come from the comps?

src/lib/Components/Input/Input.tsx
```diff
@@ -31,7 +31,7 @@ export const Input = React.forwardRef<TextInput, InputProps>(
     return (
       <Flex flexGrow={1} style={containerStyle}>
         {!!title && (
-          <Sans mb={0.5} size="3">
+          <Sans mb={0.5} size="3" weight="medium">
```
ds300, line 34: can you extract this out into its own component? We have to create our own input title for the stripe credit card input on on the `MyProfilePaymentNewCreditCard` form so we'll need to update it there too. Better to have a single source of truth

src/lib/AppRegistry.tsx
damassi: makes sense 👍

src/lib/Scenes/Consignments/v2/State/artworkModel.tsx
damassi: Ahh, actually, looking closer at the comps they were updated tot just "Select" -- will fix.

src/lib/Components/FancyModal/FancyModalHeader.tsx
damassi: Children can technically be whatever -- if we wanted icons, other components, etc.

src/lib/Components/Input/Input.tsx
damassi: Sure thing 👍

src/lib/Scenes/Consignments/v2/Screens/MyCollectionAddArtwork/MyCollectionAddArtwork.tsx
```diff
@@ -42,17 +37,37 @@ export const MyCollectionAddArtwork = () => {
...
-        </BorderBox>
-        {/* FIXME: Is there a <StackableBorderBox> for iOS? */}
-        <BorderBox px={1} position="relative" top={-1}>
-          <Button variant="noOutline" onPress={navActions.navigateToAddTitleAndYear}>
-            Title & year (optional)
-          </Button>
-        </BorderBox>
-      </Box>
+      <BorderBox px={0}>
+        <ScreenMargin>
```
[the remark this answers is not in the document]
damassi: Ah, you're right, it's very light but zooming in they're visible. Will add a FIXME and update in a subsequent PR.

src/lib/Components/Input/Input.tsx
damassi: Fixed in https://github.com/artsy/eigen/pull/3609/commits/d892e8face211fcc6bf36e289cfc58565225e9e8

PR #3611, opened 2020-07-23
Prepare for echo option

pvinis:

### Description

Adding support for the echo flag for VR, plus a couple of maintenance changes.
More specifically:

- Made the routes for VRs work already, so if you click a link like `artsy:///viewing-rooms`, you will get the right screen.
- Added `AREnableViewingRooms` to support that echo flag.
- Made the home VR rail appear if the echo flag or the lab option is enabled. Either works, so we can test and we can remote enable it. Once it's remotely enabled, we can go ahead and remove the lab option.
- Adding json in editorconfig, and formatting the echo.json.example file.
- Renaming `ARSwitchboard+Eigen.h` to `ARSwitchBoard+Eigen.h`, so it's consistant with the `.m` file.

### Test Plan

We kinda all tested it together in KS, but you can test the link above in safari, it should open the app and take you to the right screen, even with the lab option turned off in the dev menu.


.editorconfig
```diff
@@ -1,3 +1,7 @@
 [*.{js,jsx,ts,tsx}]
 indent_style = space
 indent_size = 2
+
+[*.json]
+indent_style = space
+indent_size = 2
```
ds300, line 7: 🙇

PR #3612, opened 2020-07-23
[FX-2011] Enable Pagination By Click Button On Infinite Scroll Artworks Grid

ashleyjelks:

This PR resolves **[FX-2011](https://artsyproduct.atlassian.net/browse/FX-2011)**

### Description

This PR modifies the `InfiniteScrollArtworkGrid` component so that it accepts an optional boolean prop `autoFetch`.  By default this prop is true; when true the artwork grid fetches the next page  of artworks via a scroll event (existing implementation, and when false it renders a button that fetches the next page when tapped.

### Test Plan

@ashleyjelks and @nicoleyeo will QA in a beta

### Screenshots

Kapture 2020-07-23 at 14 08 25


src/lib/Components/ArtworkGrids/InfiniteScrollArtworksGrid.tsx
```diff
@@ -210,12 +214,20 @@ class InfiniteScrollArtworksGrid extends React.Component<Props & PrivateProps, S
...
+    const hideShowMoreButton = autoFetch === true
+
     return (
       <Theme>
         <ScrollView
-          onScroll={isCloseToBottom(this.fetchNextPage)}
+          onScroll={
+            hideShowMoreButton
+              ? isCloseToBottom(this.fetchNextPage)
+              : () => {
```
sweir27, line 227: #superminor can this be `() => null` to avoid the extra brackets/line break?

ashleyjelks:
> This looks good! Just two minor comments 😄 .
> 
> Another question-- in the ticket/spec the desire was to show 20 artworks at first and then every time "Show more" is clicked, we fetch 30 artworks at a time. I think what you have here is totally mergeable, but curious if that was left out intentionally (or is it the default behavior? totally could be!).
> 
> If it's not included here, we should add a ticket to capture the follow-up work! 🙏

No, good catch, I just added a this in a followup commit!

src/lib/Components/ArtworkGrids/InfiniteScrollArtworksGrid.tsx
```diff
@@ -227,7 +232,21 @@ class InfiniteScrollArtworksGrid extends React.Component<Props & PrivateProps, S
...
-          {!!this.props.isLoading() && (
+
+          {!autoFetch && !!hasMore() && (
+            <Button
+              mt={5}
+              mb={3}
+              variant="secondaryGray"
+              size="large"
+              block
+              onPress={() => this.fetchNextPage(ARTIST_SERIES_PAGE_SIZE)}
```
sweir27, line 243: Instead of tying the page size to the Artist Series context, I think it would be more flexible to have this be a prop that can be passed in (and defaults to the existing `PAGE_SIZE`). That will allow the "Show more" button to more easily work on other pages as well!

.circleci/config.yml
```diff
@@ -5,7 +5,7 @@ version: 2.1
 orbs:
   node: circleci/node@1.1.6
   queue: eddiewebb/queue@1.0.110
-  horizon: artsy/release@0.0.1
+  # horizon: artsy/release@0.0.1
```
sweir27, line 8: Also can these changes be reverted? I _think_ horizon is back up.

ashleyjelks: I think these changes were added after a merge with `master` but the latest rebase seems too have removed the change.

ashleyjelks:
> Code wise this looks great, but I think the tap interaction could use some UX refinement.
> 
> At the moment it feels a little sluggish after tapping because the button goes back to its normal state.
> 
> To fix that we could do one of the following
> 
> * On tap, Disappear the button (see `Disappearable`) and show an `ActivityIndicator` until the next page loads.
> * Prefetch the next page so that it can be shown instantly
> 
> What do you think?

Yes, I have some clarifying questions that I'll follow up with you about via Slack but I think these changes can happen in a follow up PR!

PR #3617, opened 2020-07-24
Require only one app launch for an echo change to apply

pvinis:

### Description

This PR is trying to make eigen **not** require a second launch to apply changes from echo.

So far it seems that after we change something in echo, eigen needs to launch twice. First time it launches with the previously saved options while also fetching and saving the fresh ones, but not using them. Second time it launches, all the fresh options are locally available, so all good.

But having to launch twice sucks. I think the way I did it here can save us from the second launch.

#### Before ARNotificationManager

~The way this works is with a new hook. Start from `Home.tsx`. We used to get the options from `Emission.options`, now from `useEmissionOptions()`. Basically the same thing, only now it can refresh and rerender if needed.~

~Now, for the native code, we had the options as constants, so the moment the app starts, we are stuck with whatever we have locally (so it's fast), and we fetch and store async after that. What I do here is that I turned Emission into an event emitter. This way the native code can call the js code. When the echo fetch is done, we can call the js part of the app and notify it that we have some fresh options.~

~The native code changes are basically three things:~
- ~One function that js can call to get the freshest options.~
- ~One function that calls js to notify of new options available.~
- ~One function call after the echo fetch, to run the function above.~

~The js code changes are the ones in the hook, and it is this (I have some doc in there, but here is the idea again):~

~The initial `latestOptions` are whatever `Emission.options` gives us, same as before.~

~Then we add a listener for the event that fresh options are available and keep them in `latestOptions`. We remove that listener after 2 minutes, as echo will only fire zero or one time, only at app start.~

~We also get fresh options manually at RN initiation, so if we missed the event, we have the chance to get the freshest options.~

~In the hook, we initially get whatever `latestOptions` is (which is up to date for all hooks), and we also add a listener that will update our hook's `opts` as well as `latestOptions` for future hooks.~

~I'll try to add some tests too might I might #skip_new_tests. I was testing manually.~

##### ~Two things I don't like but I think are ok~
- ~The native function are in EigenConfiguration and I think they don't really belong there, but that's where we have the Eigen module, so I put them there.~
- ~Options are exported as constants still, and when echo changes them we only update them in the EigenConfiguration and sent to RN. I'm not sure and I didn't test if they are also refreshed when read in native code.~

#### After ARNotificationManager

I refactored all the above after a suggestion of @ds300, using the ARNotificationManager. Functionally it's pretty much the same, but here's the explanation of the code changes.

We call `emissionOptionsChanged` from native in two places, one is at init, using the local `echo.json`, and the other is after the echo fetch finishes.

The hook is using the `AppStore` for all the native<->js communication.

### Test Plan

- Make sure you have `Enable Viewing Rooms` in `LAUNCHER` in dev menu turned **off**.
- Go here https://echo-web-production.herokuapp.com/accounts/1/features.
- Toggle the `AREnableViewingRooms`.
- Close and open the app.
- Bask in the awesomeness that is single app launch for an echo toggle.
- Toggle it again and try it again, because you might not be sure it worked! :p

### Screenshots

Screen Recording 2020-07-24 at 12 41 57


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>Missing Test Files:

- `src/lib/utils/__tests__/useEmissionOptions-tests.ts`

    </tr>
  </tbody>
</table>




<p align="right">
</p>

ds300:
Nice work, but can you refactor this to use `NativeModel` + `ARNotificationsManager`? It was created exactly for this purpose, and to consolidate state into easy-peasy while we migrate app infrastructure away from obj-c. I want to avoid proliferating native bridges and context providers during this process. Happy to pair on this if you need a hand, but it should be pretty obvious.

pvinis:
I didn't know about this one. I will try it now, and if I have problems I'll let you know. Thanks!

pvinis:
All done and refactored. Thanks for pairing with me and for making a nice framework 🙏.

Artsy/App/ARSwitchBoard.h
```diff
@@ -14,7 +14,7 @@
 @class ARArtworkComponentViewController;
 @class ARAuctionArtworkResultsViewController;
 @class ARUserSettingsViewController;
-@class Aerodramus;
+@class ArtsyEcho;
```
brainbicycle, line 17: 👍 nice

emission/Pod/Classes/Core/AREmission.m
```diff
@@ -186,6 +188,8 @@ - (instancetype)initWithConfiguration:(AREmissionConfiguration *)config packager
         [ARCocoaConstantsModule new],
     ];
 
+      [self.notificationsManagerModule emissionOptionsChanged:config.options];
```
brainbicycle, line 191: minor/non-blocking but extra space at beginning of line here
```suggestion
     [self.notificationsManagerModule emissionOptionsChanged:config.options];
```

src/lib/Scenes/Home/Home.tsx
```diff
@@ -265,6 +267,9 @@ export const HomeFragmentContainer = createRefetchContainer(
 const HomePlaceholder: React.FC<{}> = () => {
   // We use Math.random() here instead of PlaceholderRaggedText because its random
   // length is too deterministic, and we don't have any snapshot tests to worry about.
+
+  const EmissionOptions = useEmissionOptions()
```
brainbicycle, line 271: ✨

emission/Pod/Classes/Core/AREmission.m
pvinis: We saw it with david, apparently it's messy because of 2/4 spaces in eigen and old emission. Sure, let's put it old-style on this file, so it's consistent. We can clean up later.

ashfurrow: Yeah, really wish Prettier supported Objective-C 😭

pvinis:
I think we could. I was very focused to make this one work before trying others, but I'm confident it's just a 1:1 replacement everywhere else.

pvinis:
The code for this PR was a rollercoaster to make 😀. Maybe I can showoff some things in a KS or s&t if people want.

ashfurrow:
Yeah definitely! I think this will likely be how we move the user ID and access token from constants into an app context, which we need to do to support logged-out users.

Issue #3473, opened 2020-06-17
[RFC] New dependency: `react-native-image-picker` 


damassi:
#### New Dependency
Name: react-native-image-picker
URL: https://github.com/react-native-community/react-native-image-picker

### Focus

A React Native module that allows you to use native UI to select a photo/video from the device library or directly from the camera.

Check out [the example code](https://github.com/react-native-community/react-native-image-picker/blob/master/example/App.js) and a test out PR [here](https://github.com/artsy/eigen/pull/3474). 

#### Check List
- [x] Have read over the source code, and we can maintain it
- [x] Has had a release in the last year, or looks done and stable

#### Alternatives
`react-native-image-picker` seems to be the standard. There's also [react-native-cameraroll](https://github.com/react-native-community/react-native-cameraroll), which we already have in the app, but it is extremely low level and doesn't give us all of the nice native bits.

MounirDhahri:
I think we should also consider trying [react-native-image-crop-picker ](https://github.com/ivpusic/react-native-image-crop-picker) since it supports multiple image selection and it's fairly stable and maintained as well. About react-native-cameraroll, I agree with what you said.

ds300:
The existing picker we have in the current consignments flow provides a significantly better UX than either of these options IMO. Is there some functionality it's missing for the new flow?

ashfurrow:
Yeah, I'd like to hear more about the specifics UX that's motivating the RFC. Maybe some mocks of what this will look like to users, in contrast to what we already have?

damassi:
We'd like to go with native UI, versus custom UI.

damassi:
@MounirDhahri - I did see that come up! The name seemed a bit misleading, but now that I'm thinking about it this could be beneficial for users when they take photo. Will dig in.

ashfurrow:
Okay, let's circle back to this. 

I've reviewed [the new designs](https://www.figma.com/file/yDSHlbJVzBxvSUROD2bfRI/Sell-tab) around image picking for consignment submissions and I have some thoughts. The downside of the designs is that users need to choose between choosing an existing image or taking a new one (whereas our current implementation lets users do either/both without choosing upfront). 

![Screen Shot 2020-07-20 at 4 11 27 PM](https://user-images.githubusercontent.com/498212/87981716-b1d18600-caa3-11ea-95d1-c17b59fbe33e.png)

[...]

PR #3618, opened 2020-07-24
Adds react-native-image-crop-picker

ashfurrow:
The type of this PR is: **dependency**

See: https://github.com/artsy/eigen/issues/3473

### Description

The Objective-C code had a bug specific to our app setup, so I had to patch it. I've [submitted a PR](https://github.com/ivpusic/react-native-image-crop-picker/pull/1354) to address it for others, too. Prettier mangled the Podfile, too 😭 

### Screenshots

<details>
  <summary>GIF proof of concept</summary>

</details>

### DIFF
--- a/patches/react-native-image-crop-picker+0.32.2.patch
+++ b/patches/react-native-image-crop-picker+0.32.2.patch
@@ -1,0 +1,17 @@
+diff --git a/node_modules/react-native-image-crop-picker/ios/.DS_Store b/node_modules/react-native-image-crop-picker/ios/.DS_Store
+new file mode 100644
+index 0000000..5172429
+Binary files /dev/null and b/node_modules/react-native-image-crop-picker/ios/.DS_Store differ
+diff --git a/node_modules/react-native-image-crop-picker/ios/src/ImageCropPicker.m b/node_modules/react-native-image-crop-picker/ios/src/ImageCropPicker.m
+index 25ead1a..3b431c8 100644
+--- a/node_modules/react-native-image-crop-picker/ios/src/ImageCropPicker.m
++++ b/node_modules/react-native-image-crop-picker/ios/src/ImageCropPicker.m
+@@ -124,7 +124,7 @@ - (void) setConfiguration:(NSDictionary *)options
+ }
+ 
+ - (UIViewController*) getRootVC {
+-    UIViewController *root = [[[[UIApplication sharedApplication] delegate] window] rootViewController];
++    UIViewController *root = [[[[UIApplication sharedApplication] windows] lastObject] rootViewController];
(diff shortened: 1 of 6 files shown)


patches/react-native-image-crop-picker+0.32.2.patch
```diff
@@ -0,0 +1,17 @@
...
+Binary files /dev/null and b/node_modules/react-native-image-crop-picker/ios/.DS_Store differ
+diff --git a/node_modules/react-native-image-crop-picker/ios/src/ImageCropPicker.m b/node_modules/react-native-image-crop-picker/ios/src/ImageCropPicker.m
+index 25ead1a..3b431c8 100644
+--- a/node_modules/react-native-image-crop-picker/ios/src/ImageCropPicker.m
++++ b/node_modules/react-native-image-crop-picker/ios/src/ImageCropPicker.m
+@@ -124,7 +124,7 @@ - (void) setConfiguration:(NSDictionary *)options
+ }
+ 
+ - (UIViewController*) getRootVC {
+-    UIViewController *root = [[[[UIApplication sharedApplication] delegate] window] rootViewController];
```
damassi, line 13: @ashfurrow - curious how you were able to debug this?

ashfurrow: Ah! I should have explained. I stepped through [the Objective-C function that is exported to React Native](https://github.com/ivpusic/react-native-image-crop-picker/blob/a0c7a8f4b591c864c8025d0eb2360b8270311301/ios/src/ImageCropPicker.m#L135-L137) in the Xcode debugger, line by line. A bit tricky since there are `dispatch` calls to different threads, but eventually I got to this line, where we have a configured picker modal that's ready to present:

https://github.com/ivpusic/react-native-image-crop-picker/blob/a0c7a8f4b591c864c8025d0eb2360b8270311301/ios/src/ImageCropPicker.m#L174

Out of curiosity, I evaluated `[self getRootVC]` in the debugger to see which view controller the library was trying to present _from_, and found it returned `nil`. The view controller code piqued my interest initially because RN's integration with them is often kind of weird, but the problem turned out to be a lot more straightforward. I don't know why the library author initially went through the app's delegate to get the window, instead of just the app directly, but it wasn't working in our case. As a bonus, the PR I made should make the library better supported on multi-window iPad apps 🌟 

Let me know what I can clarify here 👍

damassi: 👍 👍

MounirDhahri: thanks for explaining the issue @ashfurrow

PR #3620, opened 2020-07-24
dont crash if no image. just show an empty one. #trivial

pvinis:

### Description

Tiny change, but in VR list, when there is no hero image, the app would crash. now its just showing an empty square, as it should. Normally we wouldn't allow no hero image on cms and we would not return it from MP, but for now since we are still building the verifications, ill just add this here.

### Test Plan

- Go to VR list page.
- Your app didn't crash.


pvinis:
I'll go ahead and mark this as #mergeongreen. It's tiny and pretty simple.

PR #3623, opened 2020-07-24
Starts tracking auction page load time

ashfurrow:

This PR resolves **[MX-475]**

### Description

This adds a Segment analytics event when the network calls necessary to render the current Auction view are completed. This is a good approximation of time-to-interactive.
### Test Plan

Get @mikehrom to look at the event and give a 👍 on the naming: `sale_screen_load_complete` with a property of `load_time_ms`.

Then I'll assign it to an engineer to review the code

### Screenshots

Original: 

Screen Shot 2020-07-24 at 16 26 44

Update: 

Screen Shot 2020-07-27 at 17 08 48

[MX-475]: https://artsyproduct.atlassian.net/browse/MX-475

### DIFF
--- a/Artsy/View_Controllers/Auction/AuctionViewController.swift
+++ b/Artsy/View_Controllers/Auction/AuctionViewController.swift
@@ -13,6 +13,8 @@ class AuctionViewController: UIViewController {
     var buyNowView: AuctionBuyNowView?
 
     var allowAnimations = true
+    
+    var initialTime = Date()
 
     fileprivate var showLiveInterfaceWhenAuctionOpensTimer: Timer?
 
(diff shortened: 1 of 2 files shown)


mikehrom:
This naming seems good to me! I wonder... will we need to tie this event directly back to specific screens or sale_ids or any other descriptive properties? Or are we just measuring to see what `load_time_ms` looks like over time?

ashfurrow:
Mmmm yeah let's add some additional context around the sale ID and the number of lots (which is currently the performance-limiting factor here).

Artsy/View_Controllers/Auction/AuctionViewController.swift
```diff
@@ -66,17 +68,26 @@ class AuctionViewController: UIViewController {
...
             .fetch()
             .next { [weak self] saleViewModel in
+                guard let self = self else { return }
+                let now = Date()
+                let loadDuration = now.timeIntervalSince(self.initialTime)
+                ARAnalytics.event("sale_screen_load_complete", withProperties: [
+                    "load_time_ms": String(format: "%.0f", loadDuration*1000),
+                    "sale_slug": saleViewModel.saleID,
+                    "number_of_lots": saleViewModel.numberOfLots,
+                    "jump_to_lai_interface": saleViewModel.shouldShowLiveInterface
```
ashfurrow, line 78: Okay, based on @mikehrom's suggestion, we're going to also track:

- The sale slug.
- The number of lots in the sale.
- Whether or not the app immediately switched to the LAI interface (ie: the LAI sale was live currently).

PR #3626, opened 2020-07-27
[do not merge] proof-of-concept sticky content changing

ds300:
changing-sticky-content

### DIFF
--- a/src/lib/Containers/Artist.tsx
+++ b/src/lib/Containers/Artist.tsx
@@ -1,4 +1,4 @@
-import { Flex, Message, Separator, Spacer, Theme } from "@artsy/palette"
+import { Flex, Message, Sans, Separator, Spacer, Theme } from "@artsy/palette"
 import {
   ArtistAboveTheFoldQuery,
   ArtistAboveTheFoldQueryVariables,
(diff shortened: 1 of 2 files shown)


src/lib/Containers/Artist.tsx
```diff
@@ -81,14 +84,56 @@ export const Artist: React.FC<{
...
   )
 }
 
+const ArtistStickyHeader: React.FC<{}> = ({}) => {
+  const { headerOffsetY, staticHeaderHeight } = useStickyTabPageContext()
+  const _isHeaderStuck = useMemo(() => {
+    return Animated.eq(headerOffsetY, Animated.multiply(-1, staticHeaderHeight ?? +9999999))
+  }, [headerOffsetY, staticHeaderHeight])
+
+  const isHeaderStuck = Boolean(useNativeValue(_isHeaderStuck, 0))
```
ds300, line 105: the sticky tab page is built on these so-called 'Animated' values, which are abstract wrappers for native values. Here we create a new native value to represent whether or not the header is in it's 'stuck' position, and then deref the native value using this special hook `useNativeValue`

src/lib/Containers/Artist.tsx
```diff
@@ -81,14 +84,56 @@ export const Artist: React.FC<{
...
+
+  const overlayOpacity = useRef(new Animated.Value(0)).current
+
+  useEffect(() => {
+    Animated.timing(overlayOpacity, { toValue: isHeaderStuck ? 1 : 0, duration: 140, easing: Easing.ease }).start()
+  }, [isHeaderStuck])
```
ds300, line 111: then we feed that `isHeaderStuck` boolean into some animation code to control the opacity of the  overlay content

ds300:
gonna close this to clean up

MounirDhahri:
Nice initiative @ds300 !

damassi:
(Yah @ds300 - it will be a sec before this gets implemented.)

PR #3636, opened 2020-07-28
Some fixes after VR QA session

pvinis:

Context: https://www.notion.so/artsy/2020-07-27-iOS-Feature-QA-Viewing-Rooms-50bcd643fa084afcaff5f51eb5137325

### Description

This PR solves a few things:
- replace `heroImageURL` with `image.imageURLs.normalized`. We have deprecated the former.
- stop a warning about unique keys in the VR rail placeholder.
- fix padding on medium card.
- fix margin on the VR list in the VR landing page.
- make text black in tags.
- if there are 0 artworks, don't show the "View works" pill.
- use OpageImageView for images in the cards, for faster loading with gemini resizing.
- in artwork page, show more images of the same artwork, not others from the vr.

### Screenshots

In this you can see most of the above things fixed.
Screenshot 2020-07-28 at 15 48 52


src/lib/Scenes/ViewingRoom/ViewingRoomArtwork.tsx
```diff
@@ -180,7 +184,7 @@ export const ViewingRoomArtworkContainer: React.FC<ViewingRoomArtworkProps> = pr
                   underlayColor={color("white100")}
                   activeOpacity={0.8}
                 >
-                  <OpaqueImageView imageURL={item.image?.url} aspectRatio={item.image!.aspectRatio} />
+                  <OpaqueImageView useRawURL imageURL={item.image?.url} aspectRatio={item.image!.aspectRatio} />
```
mdole, line 187: @pvinis do we actually want the raw URL here? I'm pretty sure we want to be resizing this with Cloudfront, probably using the built-in functionality included with `OpaqueImageView`. I forget exactly how that works, but maybe we pass in the height and width we want to crop to?

src/lib/Scenes/ViewingRoom/Components/ViewingRoomsListItem.tsx
```diff
@@ -80,7 +84,7 @@ export const ViewingRoomsListItem: React.FC<ViewingRoomsListItemProps> = props =
   } else if (extractedArtworks.length > 1) {
     artworks = extractedArtworks.map(a => a.image!.square!)
   }
-  const images = [heroImageURL ?? "", ...artworks]
+  const images = [image?.imageURLs?.heroImage ?? "", ...artworks]
```
mdole, line 87: Hmm I'm realizing that probably everywhere we reference the hero image we should do so with Gemini + Cloudfront resizing - even after the `normalized` processing, a lot of these images would still be way bigger (and also weird sizes) than we actually want, especially on mobile. 

We have a `resizer` util in [Force](https://github.com/artsy/force/blob/master/src/v2/Utils/resizer.ts), and we [implemented one in Volt](https://github.com/artsy/volt/pull/4463) recently as well. The closest thing I see in Eigen is [this](https://github.com/artsy/eigen/blob/master/src/lib/Components/OpaqueImageView/createGeminiUrl.tsx); maybe worth moving that to a more general location and making use of it?

src/lib/Scenes/ViewingRoom/ViewingRoomArtwork.tsx
pvinis: Aaah goootcha! I'll fix this.

src/palette/elements/Cards/MediumCard.tsx
```diff
@@ -55,7 +55,7 @@ export const MediumCard: React.FC<MediumCardProps> = ({ image, title, subtitle,
             {subtitle}
           </Sans>
         )}
-        <Spacer mt={1} />
+        <Spacer mt={15} />
```
dblandin, line 58: Would `mt={3}` represent 15px within our design system?

src/palette/elements/Cards/MediumCard.tsx
```diff
@@ -43,7 +43,7 @@ export const MediumCard: React.FC<MediumCardProps> = ({ image, title, subtitle,
         style={{
           position: "absolute",
           bottom: 0,
-          left: space(1),
+          left: 15,
```
dblandin, line 46: Would `space(3)` represent 15px within our design system?

src/lib/Scenes/ViewingRoom/ViewingRoomArtwork.tsx
```diff
@@ -203,7 +207,12 @@ export const ViewingRoomArtworkContainer: React.FC<ViewingRoomArtworkProps> = pr
         underlayColor={color("white100")}
         activeOpacity={0.8}
       >
-        <LargeCard title={vrInfo.title} subtitle={vrInfo.partner!.name!} image={vrInfo.heroImageURL!} tag={tag} />
+        <LargeCard
+          title={vrInfo.title}
+          subtitle={vrInfo.partner!.name!}
+          image={vrInfo.image?.imageURLs?.heroImage ?? ""}
```
dblandin, line 213: What do you think of keeping this is `normalized` instead of aliasing it to `heroImage` to better reflect the image version name?

src/lib/Scenes/ViewingRoom/ViewingRoomArtwork.tsx
dblandin, line 187: Will we need to support `aspectRatio` for ViewingRoom / ViewingRoomSubsection images? We expose width/height for those images at the moment.

src/palette/elements/Cards/MediumCard.tsx
pvinis: no.. 1 is 10px, 2 is 20px. 15px dont exist in our design system, and the designers said we just use 15px hardcoded.

dblandin: Ahh, that's right. I sometimes use `{.5}` for 5px. Maybe `{1.5}` would work?

src/palette/elements/Cards/MediumCard.tsx
mdole: :ceiling-cat: pretty sure our design system is in increments of 10px, so `1` is 10, `2` is 20, etc.

And confusingly, I believe there IS a `0.5` value (and I want to say a `0.3` value) , but otherwise I believe it's integers all the way

src/lib/Scenes/ViewingRoom/ViewingRoomArtwork.tsx
pvinis: Yep. I'm fixing this. I misunderstood this a bit.

src/palette/elements/Cards/MediumCard.tsx
dblandin: Yep, my mistake :sweat_smile: 

Totally fine if we can't represent this in design system increments. Totally non-blocking :)

src/lib/Scenes/ViewingRoom/Components/ViewingRoomsListItem.tsx
ashfurrow: That sounds like a good idea to me too.

ashfurrow: Wait, `OpaqueImageView` will call `createGeminiUrl` for you automatically: 

https://github.com/artsy/eigen/blob/f33577ebb09800224731365734be411b66ad8126/src/lib/Components/OpaqueImageView/OpaqueImageView.tsx#L106-L111

src/lib/Scenes/ViewingRoom/ViewingRoomArtwork.tsx
ashfurrow: I think it might make sense to still call it a hero image, if that's what it's being used as. Either way works for me.

dblandin: Yep, I think that makes sense. I might alias `image` in this case instead of `normalized` though. Such as `vrInfo.heroImage.imageURLs.normalized`.

src/lib/Scenes/ViewingRoom/Components/ViewingRoomsListItem.tsx
mdole: good point! yeah seems good to just use `OpaqueImageView`

src/palette/elements/Cards/MediumCard.tsx
mdole: wait I'm so confused, I wrote a response to this and I'm not sure where it went! I believe we only have `0.3, 0.5, 1, 2, 3...`

pvinis: Yep. I try to use `mx="1"` when it's 1 from our design system, and `mx={15}` if it's 15 pixels. Right now in eigen unfortunately there is no really difference between strings and numbers, but hopefully at some point we can get to that point, of using a string for our system and a number for a hardcoded number of pixels.

dblandin: Gotcha. TIL! Thanks :)

src/lib/Scenes/ViewingRoom/Components/ViewingRoomsListItem.tsx
pvinis: Yup, doing that 😃!

pvinis:
#mergeongreen

PR #3639, opened 2020-07-28
[MX-453] Add FancyModal nesting support

ds300:

This PR resolves https://artsyproduct.atlassian.net/browse/MX-453

### Description

This PR refactors the `FancyModal` component to add support for nested fancy modals. It arranges the nested modals into a 'stack' of cards, and manages the cards' positions in a centralised context. 

✨ Bonus ✨ 

Some things I fixed along the way.

- Add `@principalField` to bottom tabs query.

   Occasionally the bottom tabs query, which fetches the unread messages count for the current user, will fail. According to our recent error handling behaviour change, this is treated as a success by relay. But the response is `{"me": null}` which relay then uses to update the global cache entry for the root `me` field. Consequently any other active views that depend on `me` being present will throw null pointer errors. (e.g. email confirmation sash). Adding the `@principalField` directive causes the error to be treated as an actual error. I'm beginning to doubt whether our change to the error handling was a good idea for app reliability  since this kind of thing might happen in other places without us realising (cc @ashfurrow).

- Fix `KeyboardAvoidingView` padding offset on profile pages, which was about 40px off. I think this was affected somehow by the bottom tabs infra changes, but I didn't have time to dig into why. I checked other usages of `KeyboardAvoidingView` and they all seem fine.

- Fix filter modal header, which had recently been given one too many borders. (cc @damassi)

- Updated the credit card input to use the new `medium` title component `InputTitle` (also cc @damassi)

### Screenshots

nested-modal-goodness

modal-roller-coaster

#skip_new_tests


src/lib/Scenes/BottomTabs/BottomTabsModel.ts
```diff
@@ -26,7 +26,7 @@ export const BottomTabsModel: BottomTabsModel = {
       defaultEnvironment,
       graphql`
         query BottomTabsModelFetchCurrentUnreadConversationCountQuery {
-          me {
+          me @principalField {
```
ashfurrow, line 29: 🙌

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>Missing Test Files:

- `src/lib/Components/FancyModal/__tests__/FancyModalCard-tests.tsx`
- `src/lib/Components/FancyModal/__tests__/FancyModalContext-tests.tsx`

    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/lib/Components/FancyModal/FancyModalCard.tsx
```diff
@@ -0,0 +1,167 @@
...
+ * It has a tappable backdrop and a content overlay and it is responsible for positioning the overlay and controlling
+ * the opacity of the backdrop in response to directions from FancyModalContext, which manages a 'stack' of these cards
+ */
+export const FancyModalCard = React.forwardRef<
+  FancyModalCard,
+  React.PropsWithChildren<{ level: number; height: number; onBackgroundPressed(): void }>
+>((props, ref) => {
+  const screen = useScreenDimensions()
+  const isRootCard = props.level === 0
+  const backdropOpacity = useRef(new Animated.Value(0)).current
```
ashfurrow, line 37: Is `useRef(...).current` roughly the same as `useMemo(() => ..., [])`?

src/lib/Components/FancyModal/FancyModalCard.tsx
```diff
@@ -0,0 +1,167 @@
...
+  const scale = useRef(new Animated.Value(1)).current
+  const translateY = useRef(new Animated.Value(isRootCard ? 0 : props.height)).current
+  const borderRadius = useRef(new Animated.Value(isRootCard ? 0 : 10)).current
+
+  useImperativeHandle(
```
damassi, line 42: Nice, have never seen this used in the wild.

src/lib/Components/FancyModal/FancyModalCard.tsx
```diff
@@ -0,0 +1,167 @@
...
+            createAnimation(backdropOpacity, 0),
+            createAnimation(borderRadius, 0),
+            createAnimation(scale, 1),
+            createAnimation(translateY, 0),
+          ]
+        }
+
+        const distanceFromTopOfStack = Math.max(stack.length - props.level - 1, 0)
+
+        // the degree by which the cards should shrink at each layer
```
damassi, line 67: This is such a terribly minor no need to fix nit! but wanted to call it out. In review i'm always noticing inconsistent sentence casing, like the previous file uses proper sentence casing (ideal) but this file is all lowercase. Would be gentler on the eye if everything was consistent.

src/lib/utils/ExecutionQueue.ts
```diff
@@ -0,0 +1,30 @@
+export class ExecutionQueue {
```
damassi, line 1: Curious what led to the creation of this class?

damassi:
Thanks for jumping on this and solving the problem so well @ds300 👌

src/lib/Components/FancyModal/FancyModalCard.tsx
ds300: Yep, functionally it's the same.

src/lib/utils/ExecutionQueue.ts
ds300: I want to guarantee there wouldn't be any interleaved execution of the `show`/`hide` methods, to prevent race conditions and janky interactions.

MounirDhahri:
Nice one 🎉

MounirDhahri:
Nice One 👏
Btw, do you know why the animation is still running slow although you are using the `nativeDriver`

PR #3644, opened 2020-07-29
Qa fixes 2

pvinis:

#trivial

### Description

- Adding a few new analytics events and fixing some existing ones. Didn't use cohesion here as the needed changes were not there already, and I need to fix this quickly for the launch. We will make them nice next week.
For context, here is the doc: https://docs.google.com/spreadsheets/u/1/d/1o1J8xs7R3wLY7MXAK28J2XKmYe4NSiKT0aNdwMkryZA/edit#gid=0.

- Removed some testing buttons we added a while ago.

- Added a `shouldTrack` prop in the ArtworkRail so we can disable it for the VR page and do our own tracking. It's a bit messy since some events use the old names, some the new names, and some cohesion. Once we move these to cohesion we could remove this prop too, or at least fix up the other props that are used. For the FeaturedRail for example I added a `trackInfo` prop to send the needed extra options.

### DIFF
--- a/src/lib/Components/ArtworkTileRail/ArtworkTileRail.tsx
+++ b/src/lib/Components/ArtworkTileRail/ArtworkTileRail.tsx
@@ -9,11 +9,26 @@ import { createFragmentContainer, graphql } from "react-relay"
 import { useTracking } from "react-tracking"
 import { ArtworkTileRailCard } from "./ArtworkTileRailCard"
 
-export const ArtworkTileRailContainer: React.FC<{
+type ArtworkTileRailContainerProps = {
   artworksConnection: ArtworkTileRail_artworksConnection
-  contextModule: Schema.ContextModules
-  onTilePress?: (slug: string) => void
-}> = ({ artworksConnection, contextModule, onTilePress }) => {
+  onTilePress?: (slug: string, id: string) => void
+} & (
+  | {
+      contextModule?: undefined
+      shouldTrack: false
+    }
+  | {
+      contextModule: Schema.ContextModules
+      shouldTrack?: boolean
+    }
+)
+
+export const ArtworkTileRailContainer: React.FC<ArtworkTileRailContainerProps> = ({
+  artworksConnection,
+  contextModule,
+  onTilePress,
+  shouldTrack = true,
+}) => {
   const artworks = extractNodes(artworksConnection)
   const tracking = useTracking()
   const navRef = useRef<any>()
@@ -32,7 +47,9 @@ export const ArtworkTileRailContainer: React.FC<{
         renderItem={({ item }) => (
           <ArtworkTileRailCard
             onPress={() => {
-              tracking.trackEvent(tappedArtworkGroupThumbnail(contextModule, item.internalID, item.slug))
+              if (shouldTrack) {
+                tracking.trackEvent(tappedArtworkGroupThumbnail(contextModule!, item.internalID, item.slug))
+              }
               {
                 !!onTilePress
                   ? onTilePress(item.slug)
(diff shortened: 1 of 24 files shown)


src/lib/Scenes/Partner/Components/PartnerShows.tsx
```diff
@@ -68,25 +68,6 @@ export const PartnerShows: React.FC<{
...
-              SwitchBoard.presentNavigationViewController(navRef.current!, `/viewing-room/${viewingRoom.slug}`)
-            }
-          >
-            {viewingRoom.title}
-          </Button>
-        ),
-      })
-    })
-  }
-
```
mdole, line 89: lol so glad tin caught this

pvinis:
Screenshot 2020-07-29 at 23 03 21

This ends now. I also made sure the pre-commit does that.

src/lib/Scenes/ViewingRoom/Components/ViewingRoomViewWorksButton.tsx
```diff
@@ -54,13 +54,14 @@ const ViewWorksButton = styled(Flex)`
...
 export const tracks = {
-  tappedViewWorksButton: (ownerID: string, slug: string) => {
+  tappedViewWorksButton: (id: string, slug: string) => {
     return {
-      action_name: Schema.ActionNames.TappedViewWorksButton,
-      action_type: Schema.ActionTypes.Tap,
+      action: Schema.ActionNames.TappedViewWorksButton,
+      context_screen_owner_type: Schema.OwnerEntityTypes.ViewingRoom,
+      context_screen_owner_id: id,
+      context_screen_owner_slug: slug,
```
mdole, line 62: I think technically we don't need to add context here because we [get the context from the parent](https://github.com/artsy/eigen/blob/e2221b56cf8628e2e635608ab68fd0838c78388a/src/lib/Scenes/ViewingRoom/ViewingRoom.tsx#L127). It's probably fine to have it in both places though - realistically it's much easier to read this way and we probably won't reuse this button in another context 🤷‍♂️

tslint.json
```diff
@@ -10,6 +10,7 @@
     "no-console": [true, ["error", ["warn", "error"]]],
     "object-literal-sort-keys": false,
     "ordered-imports": true,
+    "no-unused-variable": true,
```
ashfurrow, line 13: 🚀

ds300: 🤔 @pvinis is there some situation where this works but `tsc` with `noUnusedLocals` + `noUnusedParameters` doesn't work?

pvinis: Seems like the change here https://github.com/artsy/eigen/pull/3644/commits/461a8dc6e0b0932499eac4fdef7cf6fee189bfb4 didn't happen. Maybe I was missing or mis-running some scripts? I'm not sure, but if you think it should be happening, then let's check it and I can fix this another way. 🤷‍♂️

src/lib/Components/ArtworkTileRail/ArtworkTileRail.tsx
```diff
@@ -9,11 +9,26 @@ import { createFragmentContainer, graphql } from "react-relay"
...
+} & (
+  | {
+      contextModule?: undefined
+      shouldTrack: false
+    }
+  | {
+      contextModule: Schema.ContextModules
+      shouldTrack?: boolean
+    }
+)
```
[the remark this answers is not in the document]
pvinis: I also didn't like this a lot, and I'm down to try to make this better. I just wanted the analytics finished! Only one of them is required, and the other can be missing, but I failed to make this type right first try, so I wrote this in a rush. I like your idea, I can play with this today.

src/lib/Scenes/Partner/Components/PartnerShows.tsx
pvinis: Yup 😅. I didn't even know this was there haha.

src/lib/Scenes/ViewingRoom/Components/ViewingRoomViewWorksButton.tsx
pvinis: Yea. I noticed the context was duplicated in one or two places. I have scheduled some time with Eve to move this the right way to cohesion, and then I can clean all this up 👍.

PR #3648, opened 2020-07-30
Qa fixes 3

pvinis:

### Description

A few things in here again:
- Copy changes, using the new font and removing the Serif one from VR related things.
- If there are no featured VRs, we hide that section in the VR landing page.
- If there are no featured VRs, we show the regular VRs in home.
- Fixed tag color.
- Show description in artwork page, it was hiding in `additionalInformation`.

I'll add #trivial and #skip_new_tests for the new component. It's a funky component that shows the featured rail or if its empty then does a new query to get the regular VRs.
https://f.p0l.co/file/dropshare-public-pavlos/Screen-Recording-2020-07-30-at-20.13.00.mov
I'm guessing we will always/usually have featured VRs so it would be fine to do another request? I felt it was too much to fetch these in the first place juuuust in case we don't have featured.
Also the promo placeholder is making things a bit ugly on wait, but eh. 🤷‍♂️

### DIFF
--- a/src/__generated__/ViewingRoomArtworkQuery.graphql.ts
+++ b/src/__generated__/ViewingRoomArtworkQuery.graphql.ts
@@ -1,6 +1,6 @@
 /* tslint:disable */
 /* eslint-disable */
-/* @relayHash 4f87048717c1a28f4208313641fe1193 */
+/* @relayHash 055fecd300f1300eda1f1332701f704e */
 
 import { ConcreteRequest } from "relay-runtime";
 import { FragmentRefs } from "relay-runtime";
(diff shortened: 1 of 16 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>Missing Test Files:

- `src/lib/Scenes/ViewingRoom/Components/__tests__/ViewingRoomsHomeRail-tests.tsx`

    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #3650, opened 2020-07-31
QA: The Last Stand

pvinis:

### Description

Fixing the rest of the cards at https://www.notion.so/artsy/d008b8f10c4842d793686cf3d7a37492?v=e0e9110490c64318a15983136cb6e2c6.

- Added space at the bottom of the VR list page.
- Added new-style placeholders when loading more item in that list.
See both here: https://f.p0l.co/file/dropshare-public-pavlos/Screen-Recording-2020-07-31-at-14.30.25.mov
- Show 10 artworks in the VR in the artwork rail instead of 5.
- Fixed navbar height on artworks list page.
- Added `Touchable` so we have consistent touchables, since we had the same thing copy-pasted a bunch.

#trivial
#skip_new_tests


src/palette/elements/Touchable.tsx
```diff
@@ -0,0 +1,9 @@
+import { color } from "@artsy/palette"
+import React from "react"
+import { TouchableHighlight, TouchableHighlightProps, View } from "react-native"
+
+export const Touchable: React.FC<TouchableHighlightProps> = ({ children, ...props }) => (
```
ds300, line 5: 🙏 🎉

src/palette/elements/Touchable.tsx
```diff
@@ -0,0 +1,9 @@
...
+  <TouchableHighlight underlayColor={color("white100")} activeOpacity={0.8} {...props}>
+    <View>{children}</View>
```
ds300, line 7: I don't think we need this `View` wrapper?

ds300: ok but let's make it stretch to fit it's container in that case, by adding `flex: 1`

pvinis: With `flex: 1` I get some funny things happening 🤣.
Screenshot 2020-07-31 at 16 30 50

I changed it to `<Flex flex={flex}>{children}</Flex>`, so default works everywhere I see, and if we need `flex: 1` we can easily add it.

I'm sure we will refine this soon, since it's in eigen. Also there is a new `Pressable` component in rn 63 which we might switch to for this etc.

PR #3659, opened 2020-08-03
Partner view on ipad

pvinis:

### Description

Enable the ipad app to go to partners natively, not in a webview.

### Test Plan

I guess try to go to one, or maybe try to see if there is a case where this might be/look broken?

### Screenshots

These are a bit slow on my network and device, but you see the diff.

#### before
https://f.p0l.co/file/dropshare-public-pavlos/RPReplay_Final1596450747.MP4

#### after
https://f.p0l.co/file/dropshare-public-pavlos/RPReplay_Final1596450608.MP4

### DIFF
--- a/Artsy/View_Controllers/Fair/ARProfileViewController.m
+++ b/Artsy/View_Controllers/Fair/ARProfileViewController.m
@@ -62,37 +62,37 @@
 
 - (void)loadProfile
 {
-        // We need to figure out if it's a fair URL or not
-        //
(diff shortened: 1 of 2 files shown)


ds300:
that loading UX is very sketchy. is that recording from a physical device?

pvinis:
yea.. while debugging..

ds300:
Oh, maybe the debugging explains the extra delays. I tested it out on my simulator and it seems fine. Happy to merge once the tests are fixed.

PR #3662, opened 2020-08-03
Final qa vrs

pvinis:

### Description

Quick change to make only live VRs appear in the lists.
No closed, no scheduled.

#trivial

### DIFF
--- a/src/lib/Scenes/ViewingRoom/ViewingRoomsList.tsx
+++ b/src/lib/Scenes/ViewingRoom/ViewingRoomsList.tsx
@@ -23,6 +23,7 @@ const fragmentSpec = graphql`
       edges {
         node {
           internalID
+          status
           ...ViewingRoomsListItem_item
         }
       }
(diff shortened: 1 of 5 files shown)


src/lib/Scenes/ViewingRoom/ViewingRoomsList.tsx
```diff
@@ -48,7 +49,7 @@ interface ViewingRoomsListProps {
 
 export const ViewingRoomsListContainer: React.FC<ViewingRoomsListProps> = props => {
   const [queryData, { isLoading, hasMore, loadMore, refetchConnection }] = usePagination(fragmentSpec, props.query)
-  const viewingRooms = extractNodes(queryData.viewingRooms)
+  const viewingRooms = extractNodes(queryData.viewingRooms).filter(vr => vr.status === "live")
```
mdole, line 52: just to confirm - we definitely don't want to show upcoming or closed rooms in the VR list? at least not for launch?

jbeyen: Correct. The current issue w/ doing that is it mixes up the VR states, making them a little tricky to browse (bc of sorting by start date - opening soon ones will be on top of live ones, some live ones are mixed in with closed ones, etc. It would be cool to include recently closed and opening soon VRs when we can be more nuanced with sorting (or design new rails to hold these? cc @tinn )

pvinis: Yup. Seems for the launch version, this is good. I'll merge 👍.

pvinis:
#mergeongreen

PR #3663, opened 2020-08-03
Small ios things for vrs

pvinis:

### Description

- Small copy changes for non-live VRs
- Add support for new `Text` in `ReadMore`, and use it in VR
- Fix wrong aspect ratio of subsection image


pvinis:
#mergeongreen

PR #3671, opened 2020-08-04
Maxwidth vr

pvinis:

This PR resolves **[GALL-3061]**

### Description

Adding maxwidth for VRs too.

I added `width="100%"` before `maxWidth` so it will extend even if the text is not enough to fill one whole line. Without it, a short line of text would just go centered. With `flex: 1` it would extend downwards too. So the best choice is to extend width before shortening it.

### Screenshots
Screenshot 2020-08-04 at 13 01 15
Screenshot 2020-08-04 at 13 01 29

[GALL-3061]: https://artsyproduct.atlassian.net/browse/GALL-3061


src/lib/Scenes/ViewingRoom/Components/ViewingRoomSubsections.tsx
```diff
@@ -1,6 +1,7 @@
 import { Box, Flex, Text } from "@artsy/palette"
 import { ViewingRoomSubsections_viewingRoom } from "__generated__/ViewingRoomSubsections_viewingRoom.graphql"
 import OpaqueImageView from "lib/Components/OpaqueImageView/OpaqueImageView"
+import { maxWidth } from "palette"
```
ds300, line 4: I feel like this should be added as a builtin style to the eigen `Text` component, and then we could add a prop called `fullWidth` for the (very rare) cases where we need the text to stretch to the full width of the screen. What do you think?

pvinis: Yep. That's exactly what I thought, but I didn't wanna move things from palette or do changes on it, since we are moving the whole thing in eigen. So I added it as a style there temporarily 🤔.

What do you think we should do? I could make a PR in palette with the change, or we can wait for the absorption to happen and then rebase?

ds300: let's wait for the absorption. In the meantime can you add some `TODO` comments to say this is temporary? and I'll add a note about it to https://artsyproduct.atlassian.net/browse/MX-443

pvinis: That would have been a good candidate for `HACKS.md` 👀 😃. I put it in right above the style definition as a TODO.

jbeyen:
I like the width of the text in your first screenshot.

pvinis:
The second is there just to show that the text starts from the same spot when the text is shorter than a full line.
Yea, I think it looks nice.

PR #3694, opened 2020-08-06
[FX-2017] View Artist series on Artist Page

ashleyjelks:

This PR resolves **[FX-2017](https://artsyproduct.atlassian.net/browse/FX-2017)**

### Description
Adds the artist series component to the artist page. Links to the full artist series view when greater than 4 series associated with an artist. This PR also introduces a feature flag to enable development and QA of the Artist Series component on the Artist page without releasing it to the public.

### Test Plan
@ashleyjelks and @nicoleyeo will QA in a beta

### Screenshots

Kapture 2020-08-07 at 16 32 55

Kapture 2020-08-06 at 18 45 49
Screen Shot 2020-08-07 at 4 32 51 PM


src/lib/Scenes/ArtistSeries/ArtistSeriesMoreSeries.tsx
```diff
@@ -12,12 +12,20 @@ export type ArtistSeriesConnectionEdge = NonNullable<
...
+  artist,
+  artistSeriesHeader,
+  currentArtistSeriesExcluded,
+  ...rest
+}) => {
   const navRef = useRef<Component>(null)
   const series = artist?.artistSeriesConnection?.edges ?? []
-  const totalCount = Number(artist?.artistSeriesConnection?.totalCount ?? 0) + 1
+  const excludedArtistSeriesCount = currentArtistSeriesExcluded ? 1 : 0
+  const totalCount = Number(artist?.artistSeriesConnection?.totalCount ?? 0) + excludedArtistSeriesCount
```
williardx, line 28: I feel like we should change Metaphysics to always give the correct total number of artist series instead -- that way the UI doesn't need to worry about whether or not the current artist series is excluded. Thoughts?

src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
```diff
@@ -192,6 +193,7 @@ const ArtistArtworksContainer: React.FC<ArtworksGridProps & ViewableItemRefs> =
   }
 
   const sections = [
+    "topArtistSeries",
```
sweir27, line 196: Since this is exposed on an existing page, I think we'll want to gate it behind a feature flag within eigen that can be enabled with the admin menu. (This may affect @williardx 's upcoming PR as well!) Let me know if I can explain more-- basically I think we'll want to be safe here so that the QA/launch of these feature doesn't block the upcoming app release.

williardx: Also wondering if rendering of `topArtistSeries` should be controlled by a similar `shouldShowTopArtistSeries` method? I know we use that pattern here and on the Artwork page but I'm not super clear on when it's best to do so.

src/lib/Scenes/ArtistSeries/ArtistSeriesMoreSeries.tsx
ashleyjelks: I think the issue here is that we already have MP set up to exclude the current artist series so you don't see the current series in the list of other artist series. BUT now this component is used on the Artist page, which does not need to exclude any artist series.

ashleyjelks: I'm not sure how to go about having MP handle excluding or not excluding an artist series id based on the page requesting data. Which is why the UI is handling the total count depending on the page the component is being used on.

ashleyjelks: @anandaroop @mzikherman  do you have thoughts on this, considering how stitching is configured here: https://github.com/artsy/metaphysics/blob/bee9fd5cf3cabc20020977e2f13a104da553a9b3/src/lib/stitching/gravity/stitching.ts#L488-L493 ??

williardx: When there's no series in context (e.g. the artist page) MP shouldn't exclude anything and should give you the full list of results. But looks like in Gravity we set `totalCount` to be the total count of items _after_ we do any filtering: https://github.com/artsy/gravity/pull/13346#discussion_r465306022. So if we want to have a total count regardless of any `excludeIds` we'd have to change it there if we want to avoid having this logic in the UI.

anandaroop: @williardx is this basically the issue you were anticipating in https://github.com/artsy/metaphysics/pull/2575

> This bakes in an assumption in our product that, when querying for an artistSeriesConnection on an artist, we don't want to see the current artist series in that list if there is one. As of now there's no product use case where we would want that to occur.

And we are now finding that "there's no product use case where…" is no longer true?

anandaroop: Still wrapping my head around this, but I _think_ I find it preferable to have the client code be responsible for the necessary exclusions rather than MP. But you may have weighed this tradeoff previously when arriving at that^

williardx: Given how `totalCount` is implemented (i.e. takes into account any filters) we'd always have to adjust the count by 1 when `ArtistSeriesMoreSeries` is on the artist series page vs. the artist page. The exclusion logic lives in Metaphysics but if it were on the UI side it wouldn't solve this issue. We'd either need to update `totalCount` or create a new field that gives the total number of published artist series regardless of filters.

Re: the exclusion logic in MP vs. UI, @mzikherman and I decided to have it in Metaphysics after running into some difficulties in implementing it in the UI, and we thought it was a good bonus that we'd be able to enforce consistent behavior across platforms. The downside is that we then can't easily create a feature that would show a list of artist series that includes the current artist series in context (e.g. artist series page). But since we're on the artist page and don't have an artist series in context it's okay.

ashleyjelks: I think one other thing to consider is that if we didn't exclude the current artist series (when nested under and Artist Series Query) in MP, then we'd have to handle removing that series from the list in the UI...which seems like something we shouldn't have to manually filter. FWIW It's also less work to add +1 or +0 to the totalCount client side than it is to filter the current artist series from the list in an artist series view, but not in an artist page view. It seems cleaner somehow. Anyways, I don't think it should be a blocker to getting this merged/we can always followup with a more optimal solution. In the meantime let's keep the team in the loop so everyone has context when/if we re-visit this: cc @sweir27 @mzikherman @williardx @anandaroop

src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
ashleyjelks: Yep @williardx, good catch. It makes sense to do it here bc the Artist page is already user facing, but as we want to hold off on exposing any of the Artist Series content until its launched, it needs to be behind a feat flag here, but its safe (for the time being/provided we don't start exposing Artist Series in search before the feature is launched) to not add this flag to the Artist Series page since there's no way for user to access it.

src/lib/Scenes/ArtistSeries/ArtistSeriesMoreSeries.tsx
williardx: Agreed, not a blocker for merging 👍

PR #3699, opened 2020-08-07
Users can see related artworks from an artist series on the artwork page

williardx:

This PR resolves **[FX-2019](https://artsyproduct.atlassian.net/browse/FX-2019)**

### Description

- Creates `ArtworksInSeriesRail` component that, for an artwork in an artist series, displays other artworks from the series on the artwork page
- Updates `ArtworkTileRailCard`:
  - Displays artwork title, date and partner (if given) 
  - Supports true aspect ratio images
- `ArtistNotableWorksRail` to use true aspect ratio images instead of square aspect ratio images
- Updates other artwork rails as needed to use new `ArtworkTileRailCard` props

### Test Plan

* @williardx to test in beta

### Screenshots

**`ArtworksInSeriesRail`**

fx-2019

**`ArtistNotableWorksRail`**

Screen Shot 2020-08-06 at 11 49 46 AM

### DIFF
--- a/src/lib/Components/ArtworkTileRail/ArtworkTileRailCard.tsx
+++ b/src/lib/Components/ArtworkTileRail/ArtworkTileRailCard.tsx
@@ -12,7 +12,7 @@ const IMAGE_SIZES = {
 
 const ArtworkCard = styled.TouchableHighlight.attrs({ underlayColor: color("white100"), activeOpacity: 0.8 })``
 
-interface ArtworkTileRailCardProps {
+export interface ArtworkTileRailCardProps {
   onPress: ((event: GestureResponderEvent) => void) | null | undefined
   imageURL: string | null | undefined
   imageSize: "small" | "medium" | "large"
@@ -23,7 +23,6 @@ interface ArtworkTileRailCardProps {
   title?: string | null | undefined
   imageAspectRatio?: number | null | undefined
   useSquareAspectRatio?: boolean | null
-  useLighterFont?: boolean | null
 }
 
 export const ArtworkTileRailCard: React.FC<ArtworkTileRailCardProps> = ({
@@ -38,6 +37,10 @@ export const ArtworkTileRailCard: React.FC<ArtworkTileRailCardProps> = ({
   imageSize,
   useSquareAspectRatio = false,
 }) => {
+  if (!!imageURL && !imageAspectRatio && !useSquareAspectRatio) {
+    throw new Error("imageAspectRatio is required for non-square images")
+  }
+
   const size = IMAGE_SIZES[imageSize]
   const imageHeight = size
   const imageWidth = useSquareAspectRatio ? size : (imageAspectRatio ?? 1) * size
--- a/src/lib/Components/ArtworkTileRail/ArtworkTileRailCard.tsx
+++ b/src/lib/Components/ArtworkTileRail/ArtworkTileRailCard.tsx
@@ -15,16 +15,14 @@ const ArtworkCard = styled.TouchableHighlight.attrs({ underlayColor: color("whit
 interface ArtworkTileRailCardProps {
   onPress: ((event: GestureResponderEvent) => void) | null | undefined
   imageURL: string | null | undefined
-  artistNames?: string | null | undefined
+  imageSize: "small" | "medium" | "large"
   saleMessage: string | null | undefined
+  artistNames?: string | null | undefined
   date?: string | null | undefined
   partner?: { name: string | null } | null | undefined
   title?: string | null | undefined
-  // todo: default to "small"
-  imageSize?: string
   imageAspectRatio?: number | null | undefined
   useSquareAspectRatio?: boolean | null
-  useNormalFontWeight?: boolean | null
   useLighterFont?: boolean | null
 }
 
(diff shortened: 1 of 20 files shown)


src/lib/Scenes/Artwork/Artwork.tsx
```diff
@@ -133,6 +134,13 @@ export class Artwork extends React.Component<Props, State> {
     }
   }
 
+  shouldRenderArtworksInArtistSeries = () => {
```
williardx, line 137: Flagging that we'll likely want a feature flag here before we can release.

src/lib/Components/ArtworkTileRail/ArtworkTileRailCard.tsx
```diff
@@ -4,53 +4,61 @@ import { GestureResponderEvent } from "react-native"
...
   saleMessage,
-  useLargeImageSize,
-  useNormalFontWeight,
-  useLighterFont,
+  date,
+  partner,
+  title,
+  imageAspectRatio,
+  imageSize,
+  useSquareAspectRatio = false,
```
sweir27, line 38: Can we infer the value of this based on whether `imageAspectRatio` exists or not? It seems redundant to require both 🤔 .

src/lib/Components/ArtworkTileRail/ArtworkTileRailCard.tsx
```diff
@@ -4,53 +4,61 @@ import { GestureResponderEvent } from "react-native"
...
 }) => {
+  if (!!imageURL && !imageAspectRatio && !useSquareAspectRatio) {
```
sweir27, line 40: Why do we need the `!!imageURL` check here?

src/lib/Components/ArtworkTileRail/ArtworkTileRailCard.tsx
```diff
@@ -4,53 +4,61 @@ import { GestureResponderEvent } from "react-native"
...
-      bg={color("black30")}
-      width={useLargeImageSize ? LARGE_TILE_IMAGE_SIZE : SMALL_TILE_IMAGE_SIZE}
-      height={useLargeImageSize ? LARGE_TILE_IMAGE_SIZE : SMALL_TILE_IMAGE_SIZE}
-      style={{ borderRadius: 2 }}
-    />
+    <Box bg={color("black30")} width={imageWidth} height={imageHeight} style={{ borderRadius: 2 }} />
   )
 
   const artistNamesDisplay = artistNames ? (
-    <Sans
```
sweir27, line 48: Just confirming that we not longer want different font variants for this component?

src/lib/Components/ArtworkTileRail/__tests__/ArtworkTileRailCard-tests.tsx
```diff
@@ -64,4 +101,76 @@ describe("ArtworkTileRailCard", () => {
...
+    expect(() =>
+      mount(
+        <Theme>
+          <ArtworkTileRailCard {...props} />
+        </Theme>
+      )
+    ).toThrowError(error)
+  })
+
+  it("renders a square image when useSquareAspectRatio is true", () => {
```
sweir27, line 160: Nice tests!

src/lib/Scenes/Artwork/Artwork.tsx
```diff
@@ -342,13 +357,27 @@ export const ArtworkContainer = createRefetchContainer(
             }
           }
         }
+        artistSeriesConnection(first: 1) {
```
sweir27, line 360: Non-blocking, but once we're able to access the `totalCount` of artworks within a series, we should be able to use that instead of having to dig into the edges to conditionally show this component ✨ .

src/lib/Scenes/Artwork/Components/ArtworksInSeriesRail.tsx
```diff
@@ -0,0 +1,110 @@
...
+  const navRef = useRef<Component>(null)
+
+  const artistSeriesSlug = artwork?.artistSeriesConnection?.edges?.[0]?.node?.slug
+  const artworksConnection = artwork?.artistSeriesConnection?.edges?.[0]?.node?.artworksConnection
+
+  if (!artworksConnection) {
+    return null
+  }
+
+  const artworks = extractNodes(artworksConnection)
```
sweir27, line 25: ooooh didn't know about `extractNodes`! 😍

src/lib/Components/ArtworkTileRail/ArtworkTileRailCard.tsx
williardx: `imageURL` could be `null` or `undefined`, and there were preexisting tests that ensure that the component still works even in the absence of an image. So following that it made sense to me that if we don't have an `imageURL` then we don't really care about aspect ratios -- we should just let the component render anyway.

src/lib/Components/ArtworkTileRail/ArtworkTileRailCard.tsx
williardx: I don't think so -- the only place we were using them was in `ArtistNotableWorksRail` where we were using it to have a single component for both artist name and for title. But now that we have a first-class `title` we shouldn't need this anymore.

src/lib/Components/ArtworkTileRail/ArtworkTileRailCard.tsx
williardx: I see what you're saying but I do like the fact that it's explicit. If you don't include an aspect ratio then you may have just forgotten and don't necessarily want to use a square image.

PR #3692, opened 2020-08-06
[FX-2109] Artist Series QA

anandaroop:
The type of this PR is: **Bugfix** (← I'm guessing, since I don't see a better fit for a Q/A ticket like this)

This PR resolves **https://artsyproduct.atlassian.net/browse/FX-2109**

### Description

This is a grab bag of visual Q/A fixes for the Artist Series view

### Test Plan

n/a, style changes only

### Screenshots

#trivial

PR #3710, opened 2020-08-10
Fix Artist Series header image border radius

anandaroop:

This PR resolves **[FX-2109]**

### Description

This is take 2 (after #3692 ) at setting the `borderRadius` for the Artist Series header image. Turns out that you need to set an overflow on the `OpaqueImageView` for the image to actually be constrained. 

Here is an exaggerated version, for demo purposes —

Without no `overflow` on the left -vs-  with `overflow: hidden` on the right.

without with

### Screenshots

Here's the actual, final, very subtle 2px radius!

final

#trivial

[FX-2109]: https://artsyproduct.atlassian.net/browse/FX-2109


sweir27:
Huh! Who knew! Nice job tracking down this fix.

PR #3714, opened 2020-08-11
React native first

ds300:

This PR resolves **[MX-491]**

### Description

This PR is moving things around so that when the app launches the first thing it does is show a react-native view.

This involved:

- Turning the emission configuration class into dynamic state (so we can update userID and authenticationToken after logging in) and removing `NativeModules.Emission` which is now all in `AppStore`.
- Presenting the onboarding view when there are no auth details, or onboarding was not completed in a previous session.
- Removing the unused demo mode
- Removing the unused error modal in VIR
- Changing VIR presentation from stack-based to modal-based (I left extra comments explaining this)
- Replacing our custom `ARDynamicScreenDimensions` class with [`react-native-safe-area-context`](https://github.com/th3rdwave/react-native-safe-area-context) early because I noticed a minor UI bug which I won't bother detailing here. This lib is very obviously the right one to use (de facto standard, maintained by a RN core maintainer, used in expo, have read the source, etc etc..) so I think we can skip the RFC.
- Adding a hook for more easily using the now-stateful lab options in functional components `useEmissionOption` (I'm not crazy about that name, but it's consistent with what was there before. Let's think about it again when we're cleaning up our lab/echo infra)
- Checking the current user's session is valid whenever a metaphysics request fails for whatever reason. Before, this was only done on startup, so if your token expired during a session the app would just be broken until the user manually reboots.
- Refactoring `__appStoreTestUtils__` to be more friendly to tests that do use the app store but don't render react components (e.g. `NativeModel-tests`)
- Migrating consignments to `AppStore`. This was required since the consignments screens previously used a global easy peasy store, and we can't use more than one of those in any given react tree. Now that lab options are in `AppStore` the consignments screens were throwing when trying to access lab options. cc @damassi  
- Simplified how we mock `NativeModules` in `setupJest.ts`
- Patching react-native to allow nested `RCTRootView` instances. This was needed because each root view has its own touch handler, which caused duplicate touch event dispatches over the bridge, so we just needed to make sure that any nested `RCTRootView` instances don't create a new touch handler).

### Follow-up

- I need someone with a physical iPhone 10 + to check the back button position on VIR.
- Very very thorough QA-ing before next release!

[MX-491]: https://artsyproduct.atlassian.net/browse/MX-491

  - The only real UI change is that VIR is now presented with a fade transition rather than a slide transition.


src/lib/Scenes/Consignments/v2/Screens/AddArtwork/MyCollectionAddArtwork.tsx
```diff
@@ -97,6 +98,6 @@ export const MyCollectionAddArtwork: React.FC = () => {
           </Box>
         </ScreenMargin>
       )}
-    </Box>
+    </ScrollView>
```
ds300, line 101: I noticed this view wasn't scrollable

src/lib/store/AppStore.tsx
```diff
@@ -15,28 +16,29 @@ function createAppStore() {
...
 
-  // at test time let's allow individual tests to deep-merge an initial state before mounting
-  const mergedModel: AppStoreModel = __TEST__
-    ? defaultsDeep((__appStoreTestUtils__ && __appStoreTestUtils__.initialStateProvider()) ?? {}, appStoreModel)
-    : appStoreModel
+  if (__TEST__) {
+    ;(appStoreModel as any).__injectState = action((state, injectedState) => {
+      assignDeep(state, injectedState)
+    })
+  }
```
ds300, line 39: we used to only be able to inject state before the app store was initialised, which was problematic for any tests that don't mount react components. After this change you can inject state at any time during a test.

src/setupJest.ts
```diff
@@ -113,94 +113,98 @@ mockedModule("./lib/Components/Artist/ArtistAbout.tsx", "ArtistAbout")
...
+        if (a[k]?.mockReset) {
+          a[k].mockReset()
+        } else {
+          a[k] = b?.[k] ?? a[k]
+        }
+      }
+    })
+  }
+  reset(NativeModules, getNativeModules())
+})
```
ds300, line 206: before we had a hodgepodge of different ways to reset `NativeModules` during tests. This unifies it into one little framework.

Artsy/View_Controllers/Login_and_Onboarding/ARNativeViewControllerManager.m
```diff
@@ -0,0 +1,74 @@
...
+        UIViewController *parentVC = self.myParentViewController;
+        if (!parentVC) {
+            return;
+        }
+        [parentVC addChildViewController:self.wrappedViewController];
+        self.wrappedViewController.view.frame = self.bounds;
+        [self addSubview:self.wrappedViewController.view];
+        [self.wrappedViewController didMoveToParentViewController:parentVC];
+    }
+    self.wrappedViewController.view.frame = self.bounds;
```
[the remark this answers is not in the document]
ashfurrow: Works for me 👍

src/lib/Scenes/Consignments/v2/State/ConsignmentsArtworkModel.tsx
```diff
@@ -33,47 +33,53 @@ const initialFormValues: ArtworkFormValues = {
...
-export const artworkModel: ArtworkModel = {
-  formValues: initialFormValues,
+export const ConsignmentsArtworkModel: ConsignmentsArtworkModel = {
+  sessionState: {
+    formValues: initialFormValues,
+  },
 
   setFormValues: action((state, input) => {
-    state.formValues = input
+    state.sessionState.formValues = input
```
damassi, line 64: @ds300 - why move this under `sessionState`? Because under the hood these values are preserved across boots?

src/lib/store/AppStore.tsx
```diff
@@ -80,19 +90,24 @@ export const AppStore = {
...
   return <StoreProvider store={appStoreInstance}>{children}</StoreProvider>
 }
 
 export function useSelectedTab() {
   return hooks.useStoreState(state => state.native.sessionState.selectedTab)
 }
 
-export const useEmissionOptions = () => {
-  return hooks.useStoreState(state => state.native.sessionState.emissionOptions)
+let appStoreInstance = createAppStore()
```
damassi, line 100: @ds300 - Now that `consignments` is in the global app store do i need to increment anything after changes? Or does that only apply to lower level app session stuff?

damassi:
Amazing 💯

src/lib/Scenes/Consignments/v2/State/ConsignmentsArtworkModel.tsx
ds300: any state values in the app store are persisted to disk and restored the next time the user opens the app _unless_ those values are behind a key called `sessionState`. Since the previous consignments easy-peasy store didn't use persistence I put its state behind a `sessionState` key to avoid changing the behaviour.

src/lib/store/AppStore.tsx
ds300: Only if you make changes that affect what is persisted. See [adding_state_migrations.md](https://github.com/artsy/eigen/blob/master/docs/adding_state_migrations.md)) for more info. If it's still not clear let me know.

src/lib/Scenes/Consignments/v2/State/ConsignmentsArtworkModel.tsx
damassi: Makes sense 👍

PR #3718, opened 2020-08-12
[MX-437] Update type gene category

MounirDhahri:

This PR resolves **[MX-426]**

### Description

Design adjustments on gene/category
- Header to Sans 8, left-aligned
- Tabs to Sans 3t (to visually match Artist, Gallery pages) but does not need to be sticky (quick fix only)
- number of works to Sans 3t, black60

### Screenshots
Screen_Recording_2020-08-14_at_13 41 05

[MX-426]: https://artsyproduct.atlassian.net/browse/MX-426


ds300:
Looks like you've accidentally removed the sticky header.

ds300:
btw I think we can easily migrate this page to `StickyTabPage` to make the tab bar UI consistent with all the other pages in the app (at the moment the font and padding are both off). We could also keep the refine button sticky. And then afterwards we could delete the `ARSwitchView` class forever 🎉

MounirDhahri:
Oki perfect, I will do it now 😊

src/lib/Components/Gene/Header.tsx
```diff
@@ -28,11 +26,11 @@ class Header extends React.Component<Props, State> {
     const { gene } = this.props
     return (
       <Theme>
```
brainbicycle, line 28: wonder if this <Theme> is necessary? Should get it from page wrappers I believe: https://github.com/artsy/eigen/blob/master/src/lib/AppRegistry.tsx#L285

src/lib/Containers/Gene.tsx
```diff
@@ -238,11 +255,11 @@ export class Gene extends React.Component<Props, State> {
     return (
       <Theme>
```
brainbicycle, line 256: **nitpick (non-blocking):** Same as above wonder if we can get rid of this

MounirDhahri: yes sure

src/lib/Components/Gene/Header.tsx
MounirDhahri: yes definitely

ds300:
Looking great! Thanks for switching to `StickyTabPage` 🙌 

One more visual thing I noticed is that there should be a touch more padding between the button and the tabs. Probably just 10px.

PR #3722, opened 2020-08-12
AppStore Persistence + Migration

ds300:

### Description

This PR is the reincarnation of #3498. It refactors the state management + storage behind the 'recent searches' feature to use the `AppStore`. In doing so it builds out a framework for client-local state persistence and migration, which will set us up for success when adding or modifying features in the app which need to use client-local storage. 

#### Why migration is important

"We've gotten this far without it" you might think. True! The reason it's more relevant now is twofold:

- The app is growing and being used for more demanding features than before. Usage of client-local storage will only grow in proportion to that.
- We're moving to a system of **storing state in one place**, whereas up until now every individual module that used storage (either `AsyncStorage` in react land or `NSUserDefaults` in obj-c/swift) managed its own little piece of the storage pie.

At launch time we will rehydrate the local app state from a serialized json string written during a previous user session, if available. We need a way to make sure we can deserialize the json string in a way that prevents data loss or corruption after publishing app updates.

A trivial example of an app update that could cause data loss is simply renaming a property in the app store:

```diff
 interface RecentSearch {
   entityID: string
   link: string
 }
 interface SearchModel {
-  recentSearches: RecentSearch[]
+  searchHistory: RecentSearch[]
 }
```

After this update, any state that was stored under the name `recentSearches` will be lost forever because the app is now looking for a property called `searchHistory`.

And a simple property rename could also be responsible for data corruption:

```diff
 interface RecentSearch {
   entityID: string
-  link: string
+  href: string
 }
 interface SearchModel {
   recentSearches: RecentSearch[]
 }
```

After this update, existing recent search entries will still be there. But when you tap on one, the `href` property will be `undefined` and you won't be routed to where you want to go.

Migrations help us prevent these kinds of issues by injecting extra deserialization logic in the app launch process. This logic can take previous versions of the app state and iteratively bring them up-to-date, one version at a time. You can check out [`migration.ts`](https://github.com/artsy/eigen/pull/3722/files#diff-2d296fdfde5cac237ff73436fbf50fdf) for the implementation of this, along with [`adding_state_migrations.md`](https://github.com/artsy/eigen/pull/3722/files#diff-6a516f126c63f641c078cc905fae4556) for the docs that product developers will be expected to read and understand.

I think it's best to add this infrastructure now rather than the first time we actually need it, so that we are alert to these issues and have the infrastructure and documentation in place to deal with them. Otherwise I worry that little problems like the above would slip through unnoticed.

#### `sessionState`

Not all state should be persisted across user sessions. e.g. the currently selected tab (we want new sessions to begin on the Home screen). In these situations, developers can put their state behind a property called `sessionState` and it will not be included in the serialized version of the state.

#### The State Structure Test

I added a test which will enforce some of the things outlined above. It checks that applying all the migrations to an empty object will result in an object identical to the one that you'd get from serializing the initial state of the current version of the app store. That's quite an awkward thing to express in english so you can also [read the code](https://github.com/artsy/eigen/blob/9f7fdd21f51e495c65b21049e468054c80614bc8/src/lib/store/__tests__/migration-tests.ts#L132).

This test will fail if anyone does the following

 - accidentally add a property that should be session-scoped
 - add/delete/rename a property and forget to add a migration for it

This test will unfortunately not fail if anyone does the following 

 - modify the way that a property is _used_ in a backwards-incompatible way (e.g. changing a plain number dimension from inches to centimetres)
 - changing the shape of objects that are not present in the `initial`, dehydrated, app store state. e.g. recent searches.

To mitigate those two cases we've added plenty of encouragement around adding and testing migrations, but it would be great if anyone has any ideas for how we can catch them automatically 💭

#### ✨ Bonus! ✨ 

- We're updating the PR template. See #3688 for more info.
- We've removed the `#skip_new_tests` danger thing.
- Fixed the ref prop on the `AboveTheFoldFlatList` component
- Simplified `__appStoreTestUtils__.injectInitialState`
- Fix a minor bug in search on iPad where it wouldn't load enough autosuggest results to fill the screen.

Co-authored by @brainbicycle  @MounirDhahri 


ashfurrow:
Just seeing this now – I will review first thing tomorrow 👍

MounirDhahri:
👏

docs/adding_state_migrations.md
```diff
@@ -0,0 +1,182 @@
...
+- Modifying, adding, or removing a [computed](https://easy-peasy.now.sh/docs/api/computed.html) property.
+- Modifying, adding, or removing [actions](https://easy-peasy.now.sh/docs/api/action.html) or [thunks](https://easy-peasy.now.sh/docs/api/thunk.html).
+
+The following kinds of changes to easy-peasy models in the global app store will **always** require a migration:
+
+- Adding a property
+  ```diff
+   interface MyModel = {
+     existingProperty: string
+  +  newPropeerty: number
```
brainbicycle, line 28: nitpick (non-blocking): typo here
```suggestion
  +  newProperty: number
```

src/lib/store/__tests__/migration-tests.ts
```diff
@@ -0,0 +1,137 @@
...
+        },
+      },
+    })
+
+    expect((result as any).value).toBe("modified")
+    expect(result.version).toBe(1)
+  })
+})
+
+describe("artsy app store migrations", () => {
```
brainbicycle, line 132: Did the changing `recentSearches` to `searchHistory` example in the docs and got a failure here 🎉 
Wonder if it would be good to have a comment pointing to the docs on migrations in this test so when developers come across these failures they can have a starting point for what they have to do?
```suggestion
/**
 * Getting a failure in this test? You may need a migration.
 * See adding_state_migrations.md for more information.
 * Reach out to #product-mobile-experience with questions.
 **/
describe("artsy app store migrations", () => {

```

docs/adding_state_migrations.md
```diff
@@ -0,0 +1,182 @@
...
+     })
+   })
+   ```
+
+5. Test your migration manually on a device.
+   1. Remove the app from your phone.
+   2. Install the latest beta from testflight.
+   3. Log in and perform actions to accumulate some of the state that will be affected by your changes.
+   4. Install your development version of the app from xcode.
+   5. Check that the affected code works as intended.
```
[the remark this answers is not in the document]
damassi: I've been really fighting this too. The only one that doesn't make me feel that way is "non-blocking"; all the others are so bureaucratic IMO.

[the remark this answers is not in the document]
brainbicycle: Yeah I am personally leaning towards dropping **praise** at least, think the appeal for me in first place is just clarifying ambiguity and sometimes softening tone but neither or those are really applicable to **praise** comments.

src/lib/store/__tests__/migration-tests.ts
```diff
@@ -0,0 +1,137 @@
...
+        },
+      },
+    })
+
+    expect((result as any).zero).toBe(undefined)
+    expect((result as any).one).toBe(true)
+    expect((result as any).two).toBe(true)
+    expect((result as any).three).toBe(true)
+    expect((result as any).four).toBe(true)
+    expect((result as any).five).toBe(undefined)
```
ashfurrow, line 75: This is a nice touch!

src/lib/store/__tests__/persistence-tests.ts
```diff
@@ -0,0 +1,204 @@
...
+        bar: false,
+        foo: {
+          bar: false,
+          biz: true,
+        },
+        biz: true,
+      },
+    }
+    assignDeep(obj, { foo: { bar: true, baz: true, foo: { bar: true, baz: true } } })
+    expect(obj).toEqual({ foo: { bar: true, baz: true, biz: true, foo: { bar: true, baz: true, biz: true } } })
```
ashfurrow, line 79: **thought** this isn't the easiest line of code to read, making the purpose of this unit test difficult to infer. But I also played around and couldn't come up with a better idea of how to represent this kind of logic – maybe this is just part of the incidental complexity of this kind of deep-merge behaviour?

src/lib/store/__tests__/migration-tests.ts
ds300: great suggestion! 🙌

src/lib/store/__tests__/persistence-tests.ts
ds300: I tried to improve this by using a different naming scheme and using multiline formatting consistency.

pvinis:
I know I'm late to the party here, but I want to say these:

- Ok, first, obviously, amazing work! Good stuff, good docs, good tests, good ideas!

- Second is a suggestion for a small change. Instead of just doing `export const CURRENT_APP_VERSION = 1` and bumping the number, and then adding a migration like `[1]:` which is great, it will get messy very quickly. What I have done in the past and I feel helped a lot, is to have named migrations like:
```
const versions = {
  Init: 0,
  RenamingSearchField: 1,
  RemovingHrefFromBla: 2,
  CalculatingTheAnswer: 42,
  DeletingEverything: 44,
}

export const CURRENT_APP_VERSION = versions.DeletingEverything

const doTheMigration = () => {
...
  [RenamingSearchField]: ...,
  [DeletingEverything]: ...,
...
}
```

- I can't help but feel a bit concerned that we might be using redux and redux-persist without actually using redux and redux-persist. By that, I mean that we are starting to use the one-global-store-for-everything-just-add-stuff-keep-it-coming-more-please, and then we save and hydrate the next time, and also have the temp store. It all feels like we are kinda recreating the functionality that some libs give us. I might be missing something, and in general, this concern is not exactly easy to express. I just have this gut feeling that we might be setting ourselves up for more work to untangle things later.

ds300:
@pvinis I like the idea of naming migrations, happy for you to make a PR to do that and update the docs 🙌 

> It all feels like we are kinda recreating the functionality that some libs give us.

In this PR yeah for sure. `redux-persist` does much of the stuff I wrote from scratch here. Only problem is that it doesn't integrate with easy-peasy because it wants to wrap your reducer. `easy-peasy` gives a `persist` helper, but it doesn't support migrations. If we have a good time with this we could totally PR the migration bit upstream.

> I just have this gut feeling that we might be setting ourselves up for more work to untangle things later.

Yeah I also have the feeling that `easy-peasy` is a decision that is hard to undo, but isn't it the same with any state management solution? Admittedly plain redux might give us more wiggle room to adopt different technologies over time as our needs change, but the setup and maintenance burden of crafting out own idiosyncratic redux setup is also worth taking into consideration. I like how `easy-peasy` just has everything (more or less) working out of the box, with strong conventions for how to do things.

pvinis:
I'll try to make a PR soon!

Yea.. I prefer if we don't use redux etc, so it is more what you say about using a state management solution. I just wanted to try to express this, without blocking anything. While we use it more, I'm sure we will find it great or we will fix whatever we find not great 👍.

damassi:
+1 on the named migrations, that's a really great suggestion.

Issue #3724, opened 2020-08-12
[RFC] Add react-native-config (forked) dependency


ashfurrow:
### New Dependency

Name: react-native-config (forked, for now)

URL: https://github.com/artsy/react-native-config
 
### Motivation

Currently, Eigen uses [`cocoapods-keys`](https://github.com/orta/cocoapods-keys) to store environment configuration for the app ([the list of used keys is here](https://github.com/artsy/eigen/blob/30b74c04d0eaed113085a016bb283b504c0c70dc/Podfile#L34-L44)). CocoaPods only works for iOS apps, and we are planning in [MX-489] to migrate to a cross-platform solution, to support upcoming Android work. 

[`react-native-config`](https://github.com/luggit/react-native-config) is a good solution. However, it lacks one important feature from `cocoapods-keys`: key obfuscation. **Please note** that it is impossible to ship a binary to users that guarantees security of the keys, but key obfuscation makes it much more difficult to extract them. Unobfuscated keys are present as plaintext in the app binary, which is not acceptable. We have forked `react-native-config` and ported the obfuscation code from `cocoapods-keys` [in this PR](https://github.com/artsy/react-native-config/pull/1). Eventually, we'll need to recreate that logic for the Android app too.

A big upside to using `react-native-config` over `cocoapods-keys` is that we will migrate to storing local keys in a `.env` file, just like most other apps at Artsy. This will reduce the overhead for new contributors to the app. We will need to change our beta deploys to write the environment variables to a `.env` file prior to deploying.

### Check List

- [x] Have you read over the source code?
- [x] Has had a release in the last year, or looks done and stable?
- [x] Could you fit this codebase in your head after reading the source?
- [x] Is this the stand-out obvious answer to a particular domain problem?
- [x] Do you expect your team to be the only people who know about this dependency?
  - nope
- [x] Is this obviously being used in production by the maintainers? Is it battle-tested?
- [x] Does our bundle already include a (transitive) dependency that solves the problem and could we use that
      instead?
- [x] Do you feel well versed in the domain of this dependency and/or could you maintain it if that needs to become
      an option?

### Alternatives

We chatted about alternatives in MX Knowledge Share yesterday, and nothing came up.

[MX-489]: https://artsyproduct.atlassian.net/browse/MX-489

admbtlr:
`react-native-config` is nice, so def 👍 on that. 

[...]

PR #3725, opened 2020-08-12
[Hold] [MX-489] React native config

ashfurrow:

This PR resolves **[MX-489]**

See: https://github.com/artsy/eigen/issues/3724

### Description

This PR migrates us away from `cocoapods-keys` to (a fork) of `react-native-config`

- [x] Remove `cocoapods-keys`
- [x] Update `make update_echo` command to not use `cocoapods-keys`
- [x] Add our fork (in progress, locally)
- [x] Update our docs
- [x] Migrate all native uses of `cocoapods-keys` to `react-native-config`
- [x] (Optional, I'm not sure about this yet) Update `AREmission` to remove keys, and have JavaScript use `react-native-config` directly
  - I say "optional" because we have a lot of logic to decide between staging and production keys in `ARAppDelegate+Emission.m`. I suppose we should probably move that logic to TypeScript anyway, but I might punt on that for a follow-up work item. Open to feedback on this.
- [x] Update CI beta deploys to use `react-native-config`
- [x] Extensive local testing

### Test Plan

This will require extensive testing in beta.

## Screenshots

It's working 🎉

Screen Shot 2020-08-12 at 17 31 33

[MX-489]: https://artsyproduct.atlassian.net/browse/MX-489


ashfurrow:
I think I need to mock this out for Jest tests, since Jest can't call the under-the-hood Objective-C in `react-native-config`.

ashfurrow:
I've added new env vars to CI to reflect the new naming convention (`MyKey` versus `MY_KEY`) so hopefully tests will pass this time 🤞

ashfurrow:
```
ls Artsy/App/Echo.json || make update_echo
v12.16.2 is already installed.
Now using node v12.16.2 (npm v6.14.4)
default -> v12.16.2 *
ls: Artsy/App/Echo.json: No such file or directory
No such file or directory @ rb_sysopen - /Users/distiller/project/.env
curl https://echo-api-production.herokuapp.com/accounts/1 --header "Http-Authorization: " --header "Accept: application/vnd.echo-v2+json"
{"error":"Forbidden"}
CircleCI received exit code 0
```

So, two problems here. First, the key needs to be present in the `curl` command. Second, it needs to not be in the logs 😬

ashfurrow:
Yes! CI is green and the logs are clean 😊

Makefile
```diff
@@ -39,29 +39,16 @@ next: update_bundle_version
...
-	bundle exec pod keys set "ArtsyAPIClientSecret" "3a33d2085cbd1176153f99781bbce7c6" Artsy
-	bundle exec pod keys set "ArtsyAPIClientKey" "e750db60ac506978fc70"
-	bundle exec pod keys set "ArtsyFacebookAppID" "-"
-	bundle exec pod keys set "SegmentProductionWriteKey" "-"
-	bundle exec pod keys set "SegmentDevWriteKey" "-"
-	bundle exec pod keys set "AdjustProductionAppToken" "-"
-	bundle exec pod keys set "ArtsyEchoProductionToken" "-"
-	bundle exec pod keys set "SentryProductionDSN" "-"
-	bundle exec pod keys set "SentryStagingDSN" "-"
+	cp .env.example .env
```
brainbicycle, line 44: 🎉

patches/react-native-config+1.3.3.patch
```diff
@@ -0,0 +1,14 @@
+diff --git a/node_modules/react-native-config/react-native-config.podspec b/node_modules/react-native-config/react-native-config.podspec
+index 2864c53..43cf197 100644
+--- a/node_modules/react-native-config/react-native-config.podspec
++++ b/node_modules/react-native-config/react-native-config.podspec
+@@ -21,7 +21,8 @@ Pod::Spec.new do |s|
+     name: 'Config codegen',
+     script: %(
+ set -ex
+-HOST_PATH="$SRCROOT/../.."
++# Eigen doesn't use standard React Native folder structure (yet) so we have to do this hack.
```
brainbicycle, line 10: 🏴‍☠️

patches/react-native-config+1.3.3.patch
```diff
@@ -1,14 +1,22 @@
...
 +# Eigen doesn't use standard React Native folder structure (yet) so we have to do this hack.
 +HOST_PATH="$SRCROOT/.."
++# Use CI or shared env file.
++if [ -z ${CI+x} ]; then
++  echo "Not running on CI, using shared env file";
++  export ENVFILE=".env.shared";
++else
++  echo "Running on CI, using CI env file";
++  export ENVFILE=".env.ci";
++fi
```
ashfurrow, line 19: This was the easiest way to modify the Xcode run script to check if it was on CI. It feels like a good use of patch-package, rather than adding this to our fork, because it is specific to Eigen.

ashfurrow:
Okay, next steps on this:

- [ ] Verify that CI is getting the correct env file (`.env.ci`).
- [ ] Wait for 6.6.1 to be released to users and 6.6.2 development to start.
- [ ] Pull in upstream changes and merge.
- [ ] Announce in Slack.

ashfurrow:
Hmm, native unit tests failed because they didn't have access to the environment variables. Looks like the build of the app was entirely cached (in `derived_data`) but should have been cleared because of the change to the `react-native-config` patch. A bit of an edge case, I'll remedy later.

ashfurrow:
Looks like the script is getting the right env file name, but our build caching is replacing the generated `.m` file with the default one.

Makefile
```diff
@@ -136,7 +136,7 @@ set_git_properties:
 
 update_echo:
 	# The @ prevents the command from being printed to console logs.
-	@curl https://echo-api-production.herokuapp.com/accounts/1 --header "Http-Authorization: $(shell dotenv env | grep ARTSY_ECHO_PRODUCTION_TOKEN | awk -F "=" {'print $$2'})" --header "Accept: application/vnd.echo-v2+json" > Artsy/App/Echo.json
+	@curl https://echo-api-production.herokuapp.com/accounts/1 --header "Http-Authorization: $(shell dotenv env -f ".env.shared,.env.ci" | grep ARTSY_ECHO_PRODUCTION_TOKEN | awk -F "=" {'print $$2'})" --header "Accept: application/vnd.echo-v2+json" > Artsy/App/Echo.json
```
ashfurrow, line 139: The tests weren't failing due to missing keys, they were failing due to missing Echo data. This should fix it.

ashfurrow: It didn't, `dotenv` needs explicit file names: https://github.com/artsy/eigen/pull/3725/commits/5c9858409e740226cba27a351fcde62205da9493

.circleci/config.yml
```diff
@@ -48,7 +48,7 @@ commands:
     steps:
       - run:
           name: "Set up .env file"
-          command: scripts/generate-env.js
+          command: touch .env.shared && scripts/generate-env.js
```
ashfurrow, line 51: Both files specified by `dotenv -f` need to exist or the whole command fails.

.circleci/config.yml
```diff
@@ -144,14 +149,15 @@ commands:
...
+            - v6-app_build-{{ checksum ".manifests/app_build" }}
       - run:
           name: Build App
           command: ls derived_data || make ci
       - save_cache:
-          key: v4-app_build-{{ checksum ".manifests/app_build" }}
+          key: v6-app_build-{{ checksum ".manifests/app_build" }}
           paths:
             - derived_data
+            - node_modules/react-native-config
```
ashfurrow, line 160: This might not be necessary, the problem was with `make update_echo`, try removing tomorrow.

.env.example
```diff
@@ -0,0 +1,17 @@
...
+ARTSY_API_CLIENT_KEY=e750db60ac506978fc70
+ARTSY_FACEBOOK_APP_ID=-
+SEGMENT_PRODUCTION_WRITE_KEY=-
+SEGMENT_STAGING_WRITE_KEY=-
+ARTSY_ECHO_PRODUCTION_TOKEN=-
+SEGMENT_PRODUCTION_DSN=-
+SEGMENT_STAGING_DSN=-
+GOOGLE_MAPS_API_KEY=-
+MAPBOX_API_CLIENT_KEY=-
+SAILTHRU_KEY=-
```
ashfurrow, line 17: We should add this to the native code manifest list so that the build cache gets cleared when new keys are added.

.circleci/config.yml
```diff
@@ -38,12 +38,17 @@ commands:
           name: "Generate initial checksum manifests"
           command: |
             mkdir .manifests
-            scripts/generate-manifest.js .manifests/native_code '^Podfile' '^Makefile' '^Gemfile' '^emission/' '^Artsy'
+            scripts/generate-manifest.js .manifests/native_code '^Podfile' '^Makefile' '^Gemfile' '^emission/' '^Artsy' '^patches/react-native-*'
```
ds300, line 41: these strings are regexes so the `*` in `'^patches/react-native-*'` means "any number of dashes". You probably meant `'^patches/react-native-.*'`, although the current version will still match all file name that begin with `react-native`

We could simplify as follows though

```suggestion
            scripts/generate-manifest.js .manifests/native_code '^Podfile' '^Makefile' '^Gemfile' '^emission/' '^Artsy' '^patches/react-native'
```

ashfurrow: Ahhhh yes you're right, good point 👍

.circleci/config.yml
ashfurrow: Because the Pods are cached before the generated `.m` file is created, I think we will need this still. Even if the derived data folder is cached, I don't want to get into a weird state where the generated source file is expected to be there but isn't.

PR #3729, opened 2020-08-13
[FX-2123] Update artworks counts for Artist Series 

anandaroop:

This PR resolves **[FX-2123]**

### Description

Updates artwork count that are displayed in the context of Artist Series, such that they have a fallback value when the for-sale artworks count is zero. 

(This fallback is derived in Metaphysics; nothing special being done here other than querying a new `artworksCountMessage` field in the MP schema.)

### Test Plan

Mostly a no-op, but existing test fixtures are updated to reflect the new upstream field being utilized in MP queries

### Screenshots

|Artist Series - more series|Artist - top series|Artist - all series|Artwork - more series|
|---|---|---|---|
|||||

[FX-2123]: https://artsyproduct.atlassian.net/browse/FX-2123

### DIFF
--- a/src/lib/Scenes/ArtistSeries/__tests__/ArtistSeriesListItem-tests.tsx
+++ b/src/lib/Scenes/ArtistSeries/__tests__/ArtistSeriesListItem-tests.tsx
@@ -35,7 +35,7 @@ describe("ArtistSeriesListItem", () => {
     expect(instance.findByType(OpaqueImageView).props.imageURL).toBe(
       "https://d32dm0rphc51dk.cloudfront.net/dL3hz4h6f_tMHQjVHsdO4w/medium.jpg"
     )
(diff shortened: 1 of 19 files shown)


src/lib/Scenes/ArtistSeries/__tests__/ArtistSeriesListItem-tests.tsx
```diff
@@ -35,7 +35,7 @@ describe("ArtistSeriesListItem", () => {
     expect(instance.findByType(OpaqueImageView).props.imageURL).toBe(
       "https://d32dm0rphc51dk.cloudfront.net/dL3hz4h6f_tMHQjVHsdO4w/medium.jpg"
     )
-    expect(instance.findByProps({ "data-test-id": "count" }).props.children).toBe("25 Available")
```
anandaroop, line 38: Per the spec, the casing was wrong before, so that's fixed in this PR as well "25 Available" &rarr; "25 available"

PR #3735, opened 2020-08-14
adding names to migration versions

pvinis:

### Description

Adding names to migration versions.


docs/adding_state_migrations.md
```diff
@@ -77,20 +77,32 @@ If you're still unsure whether your change requires a new migration, please reac
 ## How to add a new state migration
 
 1. Navigate to `migration.ts`.
-2. Increment `CURRENT_APP_VERSION`.
+2. Add a new version in `Versions`. The key should be a meaningful name, and the value should be at least one higher than the one above.
```
ds300, line 80: **suggestion:** it should be exactly one higher. Maybe we should use an enum without explicit values instead so the increment is automatic 🤔

src/lib/store/migration.ts
```diff
@@ -6,11 +6,16 @@ import { produce } from "immer-peasy"
  * Before you modify this file please read docs/adding_state_migrations.md
  */
 
-export const CURRENT_APP_VERSION = 1
+const Versions = {
+  Init: 0,
+  AddSearchesAndNativeAndBottomTabs: 1,
```
ds300, line 11: **suggestion:** i think we can just have `Init: 1` since `0` will never be used.

docs/adding_state_migrations.md
pvinis: I thought of enum, but I like an explicit number more, just so things are more "set in stone".

I will change it to "exactly one", but I guess there is no problem with being more than one, as long as it's ever-increasing? 🤔

src/lib/store/migration.ts
pvinis: Indeed.. I guess I was thinking what if we didn't have that first migration already.. If you don't mind it being there as a placeholder, then let's keep it. Otherwise I can remove it.

docs/adding_state_migrations.md
ds300: The migration code expects numbers to increase by 1 every time. As a developer, so do I. Is there some use case for allowing arbitrary gaps in the sequence?

src/lib/store/migration.ts
ds300: yeah let's remove

docs/adding_state_migrations.md
pvinis: I used to have it in other apps that this number was not specifically for migrations. I was getting this number from the build number of the app. It was also ever-increasing, but it also meant that I might have one migration at build number 200 and another at 319. As long as the rule was right, and it was always increasing, everything worked well. For us it's the same, we just need it to always go up, other than that, it doesn't really matter if we use only even number or add one or use like prime numbers or whatever else :p!

That's the reason I also don't like having an enum here with implicit values. It's too easy to just `meta-↑` a line and from
```
enum Versions = {
  First: 1,
  Second,
  Third,
}
```
to go to
```
enum Versions = {
  First: 1,
  Third,
  Second,
}
```
and possibly break the universe. Keeping the numbers explicitly seems like an easy way to avoid that. But again, in my mind, the numbers are there just to know the order of things. It doesn't make any difference if there's an `increment` func or no, besides mapping over them, but we don't map over, we just compare, or sort and compare.

pvinis: I did change the docs though to keep them simple 🤷.

src/lib/store/migration.ts
```diff
@@ -6,11 +6,15 @@ import { produce } from "immer-peasy"
  * Before you modify this file please read docs/adding_state_migrations.md
  */
 
-export const CURRENT_APP_VERSION = 1
+const Versions = {
+  AddSearchesAndNativeAndBottomTabs: 1,
+}
```
ds300, line 11: I think it would make sense to add a sanity check test for this to make sure the following two things are true

1. The version numbers increase one-by-one
2. The highest version number is the same as `CURRENT_APP_VERSION`

pvinis: mm I like this, I'll work on that.

pvinis:
Added some tests and tried them out. I think this should be ready to go. 👍

ds300 (approved):
Amazing, thank you 🙏 🙇

PR #3739, opened 2020-08-17
[MX-459] - Add skeleton loader for partner page, move unknown route logic to react side

brainbicycle:
This PR resolves [MX-459]

### Description

This was slightly complicated by the fact that we are navigating to partners using`/:slug` route and losing context of the entity type when routing through Eigen. The result was an additional loading state on native side when making API calls to determine the entity type. In most cases we know the entity type on javascript side so can pass over to Eigen and can pass it along to avoid this. There is an open question as to whether slug given in search results is adequate to load a partner page.

- Passes entity info in Search Results and Rails for fairs and partners
- Adds a wrapper viewController that just embeds the ARPartnerComponentViewController in the same way ARProfileViewController currently does
- Makes Artist Placeholder a general shared component and uses in Partner view, layout is the same

#### TODO
- [x] Might make sense to update app to allow institutional partners to use partner page as part of this https://artsyproduct.atlassian.net/browse/MX-339
- [x] Open question about inconsistent slugs: https://artsy.slack.com/archives/CN8S32FJR/p1597688248069000

[MX-459]: https://artsyproduct.atlassian.net/browse/MX-459


ds300:
This would be a sensible approach, except a lot of this code you're adding will need to be deleted soon and I wonder whether it makes sense to skip to the longer-lasting solution already:

1. Add a new root field in metaphysics like
   ```graphql
   type VanityURLEntity = Fair | Partner
   type Query {
     vanityURLEntity(slug: String!): VanityURLEntity
   }
   ```
2. Delete `ARMutableLinkViewController` and replace the switchboard entry for `@"/:slug"` with 
   ```objc
    [self.routes addRoute:@"/:slug" priority:0 handler:JLRouteParams {
        return [[ARComponentViewController alloc] initWithEmission:nil
                                                        moduleName:@"VanityURLEntity"
                                                 initialProperties:parameters];
    }];
   ```
3. Create a new component like

   ```ts
    const VanityURLEntity: React.FC<{ fairOrPartner: VanityURLEntity_fairOrPartner }> = ({ fairOrPartner }) => {
      if (fairOrPartner.__typename === "Fair") {
        return <Fair fair={fairOrPartner} />
      } else if (fairOrPartner.__typename === "Partner") {
        // might need to add `safeAreaInsets.top` top padding here since `Fair` is full-bleed and `Partner` isn't.
        return <Partner partner={fairOrPartner} />
      }
      throw new Error(`404`)
    }

    const VanityURLEntityContainer = createFragmentContainer(VanityURLEntity, {
      fairOrPartner: graphql`
        fragment VanityURLEntity_fairOrPartner on VanityURLEntity {
          ... on Fair {
            __typename
            ...Fair_fair
          }
          ... on Partner {
            __typename
            ...Partner_partner
          }
        }
      `,
    })

    export const VanityURLEntityRenderer: React.SFC<{ entity: "fair" | "partner"; slug: string }> = ({
      entity,
      slug,
    }) => {
      return (
        <QueryRenderer<VanityURLEntityQuery>
          environment={defaultEnvironment}
          query={graphql`
            query VanityURLEntityQuery($slug: String!) {
              vanityURLEntity(slug: $slug) {
                ...VanityURLEntity_fairOrPartner
              }
            }
          `}
          variables={{ slug }}
            Container: VanityURLEntityContainer,
            renderPlaceholder: () => (entity === "fair" ? <FairPlaceholder /> : <PartnerPlaceholder />),
          })}
        />
      )
    }

   ``` 

4. Add it to AppRegistry.tsx

   ```ts
   register("VanityURLEntity", VanityURLEntityRenderer, { fullBleed: true })
   ```

None of this should be too tricksy, and we'll need to do something along these lines for android anyway. What do you think?

Also note that we'll still be able to make use of the infra you built for passing the `?entity=fair` param in, to decide which kind of skeleton loader to show.

brainbicycle:
@ds300 Thanks for the in depth comment 👍 , I think it makes sense to move this logic to react native side and that should allow us to delete a fair bit of Objective-C. I will need to do something along the lines of the root metaphysics field because of the slug issue Joey mentioned here anyway: https://artsy.slack.com/archives/CN8S32FJR/p1597688248069000
I'll give it a shot today.

Artsy/View_Controllers/Web_Browsing/ARInternalWebView.m
```diff
@@ -0,0 +1,40 @@
...
+    }
+}
+
+- (void)embedWebView
+{
+    NSURL *resolvedURL = [ARSwitchBoard.sharedInstance resolveRelativeUrl:self.route];
+    ARInternalMobileWebViewController *webVC = [[ARInternalMobileWebViewController alloc] initWithURL:resolvedURL];
+    UIViewController *parentVC = [self reactViewController];
+    [parentVC ar_addModernChildViewController:webVC];
+    [webVC.view alignToView:parentVC.view];
```
ds300, line 37: Woah, nice one. We could simplify `ARNativeViewControllerManager` by using this approach there too.

brainbicycle:
This PR resolves [MX-459]

### Description

Followed @ds300's approach outlined above.

This was complicated by the fact that we are navigating to partners using`/:slug` route and need to look up the entity associated with the slug in order to render the page in react native. This is currently happening in Objective-C via [ARMutableLinkViewController]() and [ARProfileViewController](). The result was an additional loading state on native side when making API calls to determine the entity type. This PR moves some of that logic for profiles specifically handling rendering **fairs** and **partners** to react-native and adds a skeleton loader for both fairs and partners. Also adds a wrapper around our native web view to handle case where unknown slug is an unknown profile type or not a profile. This can be replaced with react-native-webview eventually.

- Adds a new VanityURLEntity component that takes a profile id and renders either a partner or fair
	- call this new metaphysics endpoing to make that determination: (add link)
- Adds an entity parameter to presentation of partners and fairs to determine placeholder to render
- Makes Artist Placeholder a general shared component and uses in Partner view, layout is the same
- Adds placeholder loading to fair view
- Adds a react native wrapper around our native web view controller to handle case where profile is neither a partner nor a fair


#### TODO
- [x] Might make sense to update app to allow institutional partners to use partner page as part of this https://artsyproduct.atlassian.net/browse/MX-339
- [x] Open question about inconsistent slugs: https://artsy.slack.com/archives/CN8S32FJR/p1597688248069000
       _This was my misunderstanding about returned ids, in some cases a profile id is returned and in others a partner
       or fair id_

### Follow-ups

- We are still handling gravity redirects in Objective-C side in [ARMutableLinkViewController]() this can be handled in this same component but decided against for now since the PR has already expanded a bit in scope.
- Native web view wrapper can be replaced with react-native-webview once we migrate

[Migrate to RN-based web views](https://artsyproduct.atlassian.net/browse/MX-409)
[Spike: Web views in the app should look like native views](https://artsyproduct.atlassian.net/browse/MX-436)

[MX-459]: https://artsyproduct.atlassian.net/browse/MX-459

src/lib/Scenes/VanityURL/VanityURLEntity.tsx
```diff
@@ -0,0 +1,100 @@
...
+  fairOrPartner: VanityURLEntity_fairOrPartner
+}
+
+const VanityURLEntity: React.FC<EntityProps> = ({ fairOrPartner, originalSlug, shouldInsertInset }) => {
+  if (fairOrPartner.__typename === "Fair") {
+    return <FairContainer fair={fairOrPartner} />
+  } else if (fairOrPartner.__typename === "Partner") {
+    const { safeAreaInsets } = useScreenDimensions()
+    const insets = shouldInsertInset ? safeAreaInsets : undefined
+    return <PartnerContainer safeAreaInsets={insets} partner={fairOrPartner} />
```
ds300, line 26: I think it would be cleaner to just wrap the `PartnerContainer` in a view here to add the safe area inset.

Also, shouldn't the top safe area always be added for the partner view since it's not a full-bleed view?

ds300 (commented):
looking great 🙌  I had some questions about a few of the details, but overall this is very good stuff.

src/lib/NativeModules/SwitchBoard.tsx
```diff
@@ -13,6 +13,35 @@ function presentNavigationViewController(component: React.Component<any, any>, r
...
+}
+
+function presentPartnerViewController(component: React.Component<any, any>, slug: string) {
+  presentEntityViewController(component, slug, TopLevelEntity.Partner)
+}
+
+function presentFairViewController(component: React.Component<any, any>, slug: string, slugType?: SlugType) {
+  presentEntityViewController(component, slug, TopLevelEntity.Fair, slugType ?? SlugType.ProfileID)
+}
+
```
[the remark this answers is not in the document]
brainbicycle: Let me check, anyplace we know it is a fair we should probably be calling this method 👍

src/lib/NativeModules/SwitchBoard.tsx
```diff
@@ -13,6 +13,35 @@ function presentNavigationViewController(component: React.Component<any, any>, r
...
 
+export enum TopLevelEntity {
+  Partner = "partner",
+  Fair = "fair",
+}
+
+export enum SlugType {
+  ProfileID = "profileID",
+  FairID = "fairID",
+}
```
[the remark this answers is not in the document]
brainbicycle: Sure thing 👍

src/lib/Scenes/VanityURL/VanityURLEntity.tsx
brainbicycle: Agreed wrapping is better than adding a prop. Let me get back to you on second part, I think it is necessary because of how ARMutableLinkViewController handles presenting the containing view.

brainbicycle: 👍  I updated on the Objective-C side so we can just add every time, our childViewController adding method was binding to top safeAreaLayout guide rather than view layout guide.

src/lib/NativeModules/SwitchBoard.tsx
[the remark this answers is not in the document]
brainbicycle: Updated everywhere I could find

src/lib/NativeModules/SwitchBoard.tsx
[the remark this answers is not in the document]
brainbicycle: Added a comment on the method with param descriptions and explanations let me know if it isn't clear 🙏

Artsy/View_Controllers/Web_Browsing/ARInternalWebViewManager.m
```diff
@@ -0,0 +1,14 @@
+#import "ARInternalWebViewManager.h"
+#import "ARInternalWebView.h"
+
+@implementation ARInternalWebViewManager
+
+RCT_EXPORT_MODULE()
```
ashfurrow, line 6: Mmmm very cool!

ds300 (approved):
Nice one, thanks for addressing those questions 🙌  Looks like some native snapshots need updating and then it should be good to go.

PR #606, opened 2015-07-10
@alloy => Migrate to WKWebView 

orta:
This still isn't 100% but it's getting there, and greens.

PR #3748, opened 2020-08-19
[MX-286] Remove UIWebView usage in tests

MounirDhahri:

This PR resolves [MX-286]

### Description

Migrate to `WKWebView ` because Apple will stop accepting submissions of app updates that use UIWebView APIs starting from December 2020. See https://developer.apple.com/documentation/uikit/uiwebview for more information.

there were only 2 references for UIWebView in the project

This PR is a follow-up from [this PR](https://github.com/artsy/eigen/pull/606)

#trivial

[MX-286]: https://artsyproduct.atlassian.net/browse/MX-286


MounirDhahri:
I just realized some of the pods are still using the UIWebview, I will adjust them now

ashfurrow:
Yeah, was just about to point that out 👍 Thanks Mounir! And good catch – I'm glad that #606 already took care of the lion's share of the work here 🎉

ashfurrow:
Looks like we might need to update to a newer version of the Facebook SDK – might be tricky. @brainbicycle has most recently worked with the login code (which is the only place we use Facebook's SDK) so reach out if you need a hand 👍

MounirDhahri:
Cool, I will reach out to @brainbicycle if I had any issues

PR #3751, opened 2020-08-19
[MX-510] Removes url request VCR developer tool

ashfurrow:
The type of this PR is: **cleanup**

This PR resolves [MX-510]

### Description

This removes the `VCURLConnection` dependency. This would previously allow developers to record network requests and replay them in subsequent debug sessions. While useful, this doesn't really play nicely with Relay and has become unused. I'm not sure if it currently works or not. Its use was often for things like developing locally without hitting the actual API, but nowadays we use local versions of Metaphysics anyway. So I'm open to keeping it, but I don't think it's providing much value to our developer experience.

[MX-510]: https://artsyproduct.atlassian.net/browse/MX-510


ds300 (approved):
🎉 goodbye another unused thing 👋

PR #3763, opened 2020-08-21
[PURCHASE-2053] Restyled conversations list to look closer to web 

jo-rs:

This PR resolves [PURCHASE-2053]

### Description

Currently conversations list looks like this on mobile:
Screenshot 2020-08-18 at 15 47 26
but on Figma:
Screenshot 2020-08-25 at 01 02 29

This PR updates styles to make mobile conversations list look like this:
Screenshot 2020-08-25 at 00 44 08

There are follow up tickets to continue styling the rest of Inbox.
#trivial

[PURCHASE-2053]: https://artsyproduct.atlassian.net/browse/PURCHASE-2053


zephraph:
Here's a link to the design: https://www.figma.com/file/vtXzdPefblvRiAapA79fTh/Conversations-Inbox?node-id=120%3A31.

Give that this is a different platform there's definitely always room for design adjustments if it looks better, just run it by @willanderson.

I'd note that if the margins stay the same it looks like there's a little bit _too_ much space under Inbox. There's only supposed to be ~20px between the bottom of Inbox and the separator (based on the current designs)

src/lib/Components/Inbox/Conversations/ConversationSnippet.tsx
```diff
@@ -13,11 +13,23 @@ import styled from "styled-components/native"
...
+  top: -7;
+  z-index: 99;
+  justify-content: center;
+  align-items: center;
+`
+
+const Indicator = styled.View`
+  height: 10;
+  width: 10;
+  border-radius: 5;
```
jo-rs, line 32: In theory this could be accomplished in a much simpler fashion by just using one circle with a border. However, it seems like React Native has a bug on iOS that causes some of the background to bleed when doing so. So using this as a workaround.
Screenshot 2020-08-25 at 00 32 04
Screenshot 2020-08-25 at 00 32 48

jo-rs:
Styles have been updated once more to be closer to Figma designs instead, and include the unread counter on the header that is present on Figma, as well. (Screenshots on PR body have been updated)

src/lib/Components/Inbox/Conversations/Conversations.tsx
```diff
@@ -76,15 +76,20 @@ export class Conversations extends Component<Props, State> {
...
+    const unreadCount = this.props.me.conversations?.totalUnreadCount
+    const unredCounter = unreadCount ? `(${unreadCount})` : null
+
     return (
       <View>
-        <Serif m={2} size="8">
-          Messages
-        </Serif>
+        <Flex py={1} style={{ borderBottomWidth: 1, borderBottomColor: color("black10") }}>
+          <Sans mx={2} mt={1} size="6" style={{ borderBottomWidth: 1, borderBottomColor: color("black10") }}>
```
MounirDhahri, line 85: **suggestion:** I think we are using 8 as a fontsize for titles. You can copy it from My profile tab here
https://github.com/artsy/eigen/blob/5d3ba24de8aebed1b369428a4dfe0da73125c874/src/lib/Scenes/MyProfile/MyProfile.tsx#L33

src/lib/Components/Inbox/Conversations/ConversationSnippet.tsx
```diff
@@ -92,32 +92,52 @@ export class ConversationSnippet extends React.Component<Props> {
...
+            </Flex>
             <Flex ml={1} style={{ flex: 1 }} justifyContent="flex-start">
-              <Flex flexDirection="row" style={{ flex: 0, alignItems: "center" }}>
-                <Sans size="3t" weight="medium" ellipsizeMode="tail" numberOfLines={1} style={{ flex: 1 }}>
+              <Flex flexDirection="row" mb="2px" style={{ flex: 0, alignItems: "center" }}>
+                <Sans
+                  size="3t"
+                  weight="medium"
+                  ellipsizeMode="tail"
+                  numberOfLines={1}
```
MounirDhahri, line 116: Nice👏

src/lib/Components/Inbox/Conversations/ConversationSnippet.tsx
```diff
@@ -92,32 +92,52 @@ export class ConversationSnippet extends React.Component<Props> {
...
+                  color={conversation.unread ? "black" : "black60"}
+                  style={{ flex: 3 }}
+                >
                   {partnerName}
                 </Sans>
-                {!!conversation.unread && <UnreadIndicator />}
+                <Sans size="3t" color="black30" style={{ flex: 1 }}>
+                  {conversation.messagesConnection?.totalCount}
+                </Sans>
+                <Sans textAlign="right" size="3t" color="black30" style={{ flex: 0 }}>
```
MounirDhahri, line 126: **suggestion:** Maybe we need to add `numberOfLines={1}` here as well

jo-rs: Added 👍

src/lib/Components/Inbox/Conversations/Conversations.tsx
jo-rs: Applied 💪

MounirDhahri (approved):

MounirDhahri:
Looks good! thanks 👏👏👏

PR #3771, opened 2020-08-25
chore: migrate routing logic to TS

ds300:

This PR resolves [MX-485]

## BEFORE YOU START

This PR is written as a guided tour. Each comment links to the next, and hopefully by the end you'll have a pretty firm grasp on what's changed in this PR and why. Sometimes GitHub can't find the next link unless you open it in a new tab 🤷‍♂️. Sometimes, even after opening a link in a new tab it will take GitHub up to 10 seconds to make the browser scroll you to the comment 🤷‍♂️. Sorry about that. Have patience 🧘 

### Follow-up work

- remove usages of TS SwitchBoard and nav refs
- non-modal web view presentation, w/r/t back button
- move modal presentation/transition style logic to TS
- give modally-presented VCs navigation stacks + optional back buttons to allow us to
  - replace usages of NavigatorIOS
  - give modally-presented web views a back button when they want to show an artwork or an artist
- tell ARComponentVCs whether they are being presented modally, let them manage their own close button (this will allow modal web views rendered by VanityURL to be dismissed nicely, not a big deal right now since we don't present any vanity urls modally)
- add logic to check that all declared RouteMatchers are exercised in `routes-tests.tsx`
- delete `landingURLRepresentation` ?

[MX-485]: https://artsyproduct.atlassian.net/browse/MX-485


Artsy/View_Controllers/ARMutableLinkViewController.m
```diff
@@ -1,71 +0,0 @@
...
-            return;
-        }
-
-        // We need to call begin/endAppearanceTransition manually, because at this point, our view controller
-        // has already appeared (so newly added children won't receive viewWill/DidAppear callbacks automatically).
-        [internalViewController beginAppearanceTransition:true animated:false];
-        [self ar_addAlignedModernChildViewController:internalViewController];
-        [internalViewController endAppearanceTransition];
-    }];
-}
```
ds300, line 69: [21] This class `ARMutableLinkViewController` was previously used by `ARSwitchBoard` when it saw a vanity URL (an artsy URI path with only one segment e.g. `artsy.net/alpha-137-gallery`) without a query param hint to tell us what kind of entity the vanity URL was associated with (either a Partner or a Fair)

https://github.com/artsy/eigen/blob/38046929f68590032ae9df1c13e4476d74ff4586/Artsy/App/ARSwitchBoard.m#L505-L515

https://github.com/artsy/eigen/blob/38046929f68590032ae9df1c13e4476d74ff4586/Artsy/App/ARSwitchBoard+Eigen.m#L180-L183

e.g. previously `/gagosian?type=partner` would directly display the TS app registry module `VanityURLEntity` while just `/gagosian` would display one of these, an `ARMutableLinkViewController`

The job of this class was to resolve any vanity URL redirect, e.g. `/artsy-vanguard-2019` redirects to `/series/artsy-vanguard-2019`, and use the switchboard to display that redirect route if it is indeed a redirect. If there was no redirect it would go on to show the `VanityURLEntity` module and try to resolve the vanity slug as a partner or fair, falling back to showing a web view if neither of those worked.

But since we deleted the native switchboard we had to move this logic to the TS-side too.

src/lib/navigation/routes.tsx
```diff
@@ -0,0 +1,144 @@
...
+    new RouteMatcher("/fair/:fairID/bmw-sponsored-content", "FairBMWArtActivation"),
+    new RouteMatcher("/city/:citySlug/:section", "CitySectionList"),
+    new RouteMatcher("/city-fair/:citySlug", "CityFairList"),
+    new RouteMatcher("/city-save/:citySlug", "CitySavedList"),
+    new RouteMatcher("/auctions", "Auctions"),
+    new RouteMatcher("/works-for-you", "WorksForYou"),
+    new RouteMatcher("/categories", "WebView", () => ({ url: "/categories" })),
+
+    new RouteMatcher("/city-bmw-list/:citySlug", "CityBMWList"),
+    new RouteMatcher("/:slug", "VanityURLEntity"),
```
ds300, line 130: [22] Now all single-segment URI paths on an artsy.net domain are routed directly to the `VanityURLEntity` module.

src/lib/Scenes/VanityURL/VanityURLEntity.tsx
```diff
@@ -99,11 +91,7 @@ export const VanityURLEntityRenderer: React.FC<RendererProps> = ({ entity, slugT
             if (props.vanityURLEntity) {
               return <VanityURLEntityFragmentContainer fairOrPartner={props.vanityURLEntity} originalSlug={slug} />
             } else {
-              if (useReactNativeWebView) {
-                return <RelativeURLWebView route={slug} />
-              } else {
-                return <InternalWebView route={slug} />
-              }
+              return <VanityURLPossibleRedirect slug={slug} />
```
ds300, line 94: [23] Now we always try to resolve the vanity url with metaphysics first. If that fails, or [if it returns an unexpected entity type](https://github.com/artsy/eigen/blob/151de479c835b805e3a8d5dfeef2923301da7979/src/lib/Scenes/VanityURL/VanityURLEntity.tsx#L20-L33) then we show this new component called `VanityURLPossibleRedirect`

src/lib/Scenes/VanityURL/VanityURLPossibleRedirect.tsx
```diff
@@ -0,0 +1,129 @@
...
+        case "show_native_view":
+          // TODO: implement `replace` mode of navigation
+          goBack()
+          navigate(result.url)
+          break
+        default:
+          assertNever(result)
+      }
+    })
+  }, [])
```
ds300, line 37: [24] `VanityURLPossibleRedirect` shows a loading spinner while we send a HEAD request to force (details coming soon) to decide whether there is a redirect and what to do about it. 

💡 Notice the `assertNever` that I added in `ambient.d.ts` https://github.com/artsy/eigen/blob/151de479c835b805e3a8d5dfeef2923301da7979/src/ambient.d.ts#L36

It lets us make sure that all cases are handled in a switch statement that operates on a discriminated union type. See this [awesome guide to discriminated unions](https://basarat.gitbook.io/typescript/type-system/discriminated-unions). Don't be put-off, It's a very jargony name for a simple and powerful concept that you already mostly understand 💪

src/lib/Scenes/VanityURL/VanityURLPossibleRedirect.tsx
```diff
@@ -0,0 +1,129 @@
...
+          variant="secondaryOutline"
+          onPress={() => {
+            Linking.openURL(join(webURL, slug))
+          }}
+        >
+          Open in browser
+        </Button>
+      </Stack>
+    )
+  }
```
ds300, line 65: [25] I added a 404 page in the (vanishingly rare) case that a vanity URL is not a valid page on force.

image

src/lib/Scenes/VanityURL/VanityURLPossibleRedirect.tsx
```diff
@@ -0,0 +1,131 @@
...
+    } else {
+      // The app has a native screen for the redirect URL, let's show it.
+      // Test this with `artsy:///auction` which force should redirect to /auctions
+      return { type: "show_native_view", url: response.url }
+    }
+  } catch (e) {
+    // Happens only if there's a network problem
+    return { type: "error" }
+  }
+}
```
ds300, line 131: [26] This code handles making the HEAD request and deciding what to do about it.

The `join(webURL, slug)` pattern implements the logic that was previously in the `ARSwitchBoard#resolveRelativeURL` method. https://github.com/artsy/eigen/blob/38046929f68590032ae9df1c13e4476d74ff4586/Artsy/App/ARSwitchBoard.m#L693-L696

Returning the full `response.url` here also lets us avoid the need for the `RelativeURLWebView` component that was added recently https://github.com/artsy/eigen/blob/38046929f68590032ae9df1c13e4476d74ff4586/src/lib/Components/WebView/RelativeURLWebView.tsx#L10-L27

So this PR removes that, along with all associated code.

Reading this file now, I think I could remove a fair amount of indirection here, but I don't have time so I'll add it to the list of follow-up work.

ds300 (commented):
First batch of comments, many more to come tomorrow, and then I'll link them all together into a story.

Artsy/App/ARAppNotificationsDelegate.m
```diff
@@ -277,76 +275,36 @@ - (void)applicationDidReceiveRemoteNotification:(NSDictionary *)userInfo inAppli
...
                                             }];
-            }
 
         } else if (applicationState == UIApplicationStateInactive) {
             // The user tapped a notification while the app was in background.
-            [self tappedNotification:notificationInfo viewController:viewController];
+            [self tappedNotification:notificationInfo url:url];
         }
     }
 }
```
ds300, line 296: [36] The way we handle in-app notifications has been simplified a lot too. I paired on this with @ashfurrow 

Before, it had a couple of special cases where the notification would not be shown, these have been removed since they don't matter much for UX and we're in the gradual process of rethinking in-app notifications for android.

Artsy/App/ARAppNotificationsDelegate.m
```diff
@@ -277,76 +275,36 @@ - (void)applicationDidReceiveRemoteNotification:(NSDictionary *)userInfo inAppli
...
-
-    if (isConversation) {
-        [topMenuController presentRootViewControllerInTab:[ARTabType inbox] animated:NO];
-    }
-
-    if (viewController) {
-        [[ARTopMenuViewController sharedController] pushViewController:viewController];
-    }
+    [[AREmission sharedInstance] navigate:url];
 }
```
ds300, line 308: [37] when a user taps an in-app notification we now simply call `navigate`

emission/Pod/Classes/ViewControllers/ARArtworkComponentViewController.m
```diff
@@ -1,24 +0,0 @@
...
-{
-    if ((self = [super initWithEmission:emission
-                        moduleName:@"Artwork"
-                      initialProperties:@{ @"artworkID": artworkID }])) {
-        _artworkID = artworkID;
-    }
-    return self;
-}
-
-@end
```
ds300, line 24: [38] a happy side-effect of removing `ARSwitchBoard` was that we were able to delete the vast majority of our subclasses of `ARComponentViewController`

emission/Pod/Classes/ViewControllers/ARArtistComponentViewController.m
```diff
@@ -1,32 +0,0 @@
-#import "ARArtistComponentViewController.h"
-#import "ARGraphQLQueryPreloader.h"
-
-@implementation ARArtistComponentViewController
-
-+ (NSDictionary *)propsWithArtistID:(NSString *)artistID;
-{
-    BOOL isPad = UI_USER_INTERFACE_IDIOM() == UIUserInterfaceIdiomPad;
-    return @{ @"artistID": artistID, @"isPad": @(isPad) };
-}
```
ds300, line 10: [39] There were a few minor things like this where a subclass of `ARComponentViewController` was doing something a little differently. These were resolved on a case-by-case basis. e.g. here this `isPad` prop that we were passing to `Artist.tsx` was replaced on the TS side with the `isPad()` helper:

https://github.com/artsy/eigen/blob/3c33ede1896742a52ad8a6d7cbf7fa8c235885bb/src/lib/Containers/Artist.tsx#L125

Artsy/View_Controllers/Embedded/Generics/AREmbeddedModelsPreviewDelegate.m
```diff
@@ -62,24 +63,9 @@ - (nullable UIViewController *)previewingContext:(id<UIViewControllerPreviewing>
...
-        [[ARTopMenuViewController sharedController] pushViewController:viewController];
-    } else {
-        if ([object respondsToSelector:@selector(publicArtsyPath)]) {
-            [switchBoard loadPath:[object publicArtsyPath]];
-        }
+    
+    if ([object respondsToSelector:@selector(publicArtsyPath)]) {
+        [[AREmission sharedInstance] navigate:[object publicArtsyPath]];
     }
 }
```
ds300, line 70: [40] For the `Spotlight` links we had this complicated logic that loaded different kinds of view controllers based on what kind of object was associated with the spotlight search result. The thing was that all of the objects have this `publicArtsyPath` method that we can just pass to `navigate`, so I deleted all the conditional junk.

Artsy/Models/API_Models/Partner_Metadata/Artist.m
```diff
@@ -104,13 +103,6 @@ - (void)getFollowState:(void (^)(ARHeartStatus status))success failure:(void (^)
     } failure:failure];
 }
 
-- (NSString *)publicURL
-{
-    NSString *path = [NSString stringWithFormat:ARArtistInformationURLFormat, self.artistID];
-    NSURL *url = [ARSwitchBoard.sharedInstance resolveRelativeUrl:path];
-    return [url absoluteString];
-}
-
```
ds300, line 113: [41] I noticed that a bunch of our domain classes had this `publicURL` method, in addition to the `publicArtsyPath`. They weren't being used so i deleted them.

Artsy/App/ARSwitchBoard.m
```diff
@@ -1,712 +0,0 @@
...
-        return [sself loadGeneWithID:parameters[@"id"] refineParams:parameters];
-    }];
-
-    [self registerEchoRouteForKey:@"ARShowRoute" handler:JLRouteParams {
-        if ([parameters[@"entity"] isEqualToString:@"fair-booth"]) {
-            return [[ARFairBoothComponentViewController alloc] initWithFairBoothID:parameters[@"id"]];
-        }
-        __strong typeof (wself) sself = wself;
-        return [sself loadShowWithID:parameters[@"id"]];
-    }];
```
ds300, line 241: [42] One final thing to note regarding routing logic that moved somewhere else, we have this URI query param `?entity=fair-booth` that can be used to redirect a show url to a fair booth view controller. This moved here https://github.com/artsy/eigen/blob/3c33ede1896742a52ad8a6d7cbf7fa8c235885bb/src/lib/Scenes/Show/Show.tsx#L35-L38

Artsy/View_Controllers/Auction_Results/ARAuctionWebViewController.m
```diff
@@ -102,8 +85,8 @@ - (void)bidHasBeenConfirmed;
                       object:self
                     userInfo:@{ARAuctionIDKey : self.auctionID, ARAuctionArtworkIDKey : self.artworkID}];
 
-    [self ensureArtworkViewControllerIsLowerInStack];
     [self.navigationController popViewControllerAnimated:ARPerformWorkAsynchronously];
+    [[AREmission sharedInstance] navigate:[NSString stringWithFormat:@"/artwork/%@", self.artworkID]];
```
ds300, line 89: [44] The `AuctionWebViewController`, which probably has not seen the light of day for about a few years, was previously doing this thing where it would manipulate the navigation stack to show the artwork a user has bid on after their bid has been confirmed. 

This change degrades the UX to simplify the code. Instead of manipulating the navigation stack so that the view controller is already there when the bid flow is popped, we just pop the bid flow and then show the artwork immediately afterwards.

I think it's extremely unlikely that anyone will ever see this, but if they do, it shouldn't be too bad.

ds300 (commented):
stashing the next batch of comments before lunch.

emission/Pod/Classes/EigenCommunications/ARNotificationsManager.m
```diff
@@ -114,11 +130,29 @@ - (void)stopObserving {
...
+RCT_EXPORT_METHOD(didFinishBootstrapping)
+{
+    self.didBootStrap = true;
+    while (self.bootstrapQueue.count > 0) {
+        void (^completion)() = [self.bootstrapQueue firstObject];
+        [self.bootstrapQueue removeObjectAtIndex:0];
+        completion();
+    }
+}
+
```
ds300, line 155: [45] There's a few places where we need to call `navigate` before the app has had a chance to get into a usable state (e.g. handling deep links, shortcuts menu, etc). To solve that, I added a feature to this class that allows it to put operations in a queue, `bootstrapQueue`, and execute them one-by-one after the TS `AppStore` has finished loading.

We already had something similar in `ARTopMenuViewController` (added in #3792) which I adapted to use the same pattern. https://github.com/artsy/eigen/blob/6ec84a1ae810c6170370a7e4fd8a7d9154c56ce9/Artsy/View_Controllers/App_Navigation/ARTopMenuViewController.m#L199-L222

Artsy/View_Controllers/App_Navigation/ARTopMenuNavigationDataSource.m
```diff
@@ -98,11 +83,6 @@ - (ARNavigationController *)navigationControllerForTabType:(NSString *)tabType
     return [[self.config objectForKey:tabType] navigationController];
 }
 
-- (NSString *)switchBoardRouteForTabType:(NSString *)tabType
-{
-    return [[self.config objectForKey:tabType] route];
-}
-
```
ds300, line 105: [46] we no longer register tab routes in obj-c now so we can clean this file up a bit.

src/lib/navigation/routes.tsx
```diff
@@ -0,0 +1,144 @@
...
+    new RouteMatcher("/fair/:fairID/artists", "FairArtists"),
+    new RouteMatcher("/fair/:fairID/exhibitors", "FairExhibitors"),
+    new RouteMatcher("/fair/:fairID/info", "FairMoreInfo"),
+    new RouteMatcher("/fair/:fairID/bmw-sponsored-content", "FairBMWArtActivation"),
+    new RouteMatcher("/city/:citySlug/:section", "CitySectionList"),
+    new RouteMatcher("/city-fair/:citySlug", "CityFairList"),
+    new RouteMatcher("/city-save/:citySlug", "CitySavedList"),
+    new RouteMatcher("/auctions", "Auctions"),
+    new RouteMatcher("/works-for-you", "WorksForYou"),
+    new RouteMatcher("/categories", "WebView", () => ({ url: "/categories" })),
```
ds300, line 127: [47] We've had a ticket for moving `/categories` to a web view in our backlog for so long 🎉

src/lib/Components/Inbox/Conversations/Message.tsx
```diff
@@ -96,38 +97,27 @@ const track: Track<Props> = _track
...
+      const attachment = compact(attachments).find(({ internalID }) => internalID === attachmentID)!
+      NativeModules.ARScreenPresenterModule.presentMediaPreviewController(
+        reactNodeHandle,
         attachment.downloadURL,
-        // @ts-ignore STRICTNESS_MIGRATION
         attachment.contentType,
-        // @ts-ignore STRICTNESS_MIGRATION
         attachment.internalID
       )
     }
```
ds300, line 110: [48] there was a slight API change for `presentMediaViewController`, the one final place where we actually do need to pass a react ref to obj-c. But I've hidden the ref resolution in the `AttachmentPreview` class.

https://github.com/artsy/eigen/blob/6ec84a1ae810c6170370a7e4fd8a7d9154c56ce9/src/lib/Components/Inbox/Conversations/Preview/Attachment/AttachmentPreview.tsx#L23-L36

.circleci/config.yml
```diff
@@ -43,7 +43,8 @@ commands:
             scripts/generate-manifest.js .manifests/js_transform '^\.manifests/node_modules' '^babel\.config\.js' '^relay\.config\.js' '^jest\.config\.js'
             scripts/generate-manifest.js .manifests/js_bundle '^\.manifests/js_transform' '^data/' '^index\.ios\.js' '^src/(?!.*(__tests__|__mocks__|__fixtures__))'
             scripts/generate-manifest.js .manifests/js_test_results '^\.manifests/js_transform' '^data/' '^src/'
-            scripts/generate-manifest.js .manifests/cocoapods '^Podfile' '^Gemfile'
+            ls -lR emission > .manifests/emission_pod_file_list
+            scripts/generate-manifest.js .manifests/cocoapods '^Podfile' '^Gemfile' '^\.manifests/emission_pod_file_list'
```
ds300, line 47: [49] I noticed a caching issue. We need to reinstall the pods if the list of files in the emission pods dir changes..

Artsy_Tests/App_Tests/ARAppActivityContinuationDelegateTests.m
```diff
@@ -39,44 +39,43 @@
...
         [userManagerMock stopMocking];
-        
-        [topMenuMock verify];
-        [topMenuMock stopMocking];
-        
+
+        // TODO: figure out why this mock is not working
+//        [emissionMock verify];
+        [emissionMock stopMocking];
+
```
ds300, line 70: [50] Not sure if this is worth worrying about, but this mock for the emission `navigate` method is not working. If I debug the test it does call `navigate` but it steps into the actual implementation. I don't know much about this mocking framework and it didn't seem too critical to figure it out since this isn't a core feature of the app and I'm pretty sure it's still working.

Artsy_Tests/App_Tests/ARAppNotificationsDelegateTests.m
```diff
@@ -152,43 +151,6 @@ - (UIViewController *)getGlobalTopViewController;
...
-
-            id mock = [OCMockObject mockForClass:[ARNotificationView class]];
-            [[mock expect] showNoticeInView:OCMOCK_ANY title:OCMOCK_ANY response:OCMOCK_ANY];
-
-            [delegate applicationDidReceiveRemoteNotification:@{ @"sale_id" : @"123ffg3edfsd", @"artwork_id": @"asd1432asda", @"action": @"winning" } inApplicationState:appState];
-
-            [mock verify];
-            [mock stopMocking];
-            [mockedNotificationsDelegate stopMocking];
-        });
```
ds300, line 191: [51] these tests are no longer relevant. See https://github.com/artsy/eigen/pull/3771/files#r494263577

Artsy_Tests/View_Controller_Tests/Web_Browsing/ARExternalWebBrowserViewControllerTests.m
```diff
@@ -54,50 +54,23 @@
     expect([vc shouldLoadNavigationResponse: (id)mock]).to.equal(WKNavigationResponsePolicyAllow);
 });
 
+// TODO: figure out how to test that it calls openURLInExternalService now that it's a local method
+//it(@"handles showing an alert punting a user to safari if we can't show something in a webview", ^{
```
ds300, line 58: [52] as the comment says, I couldn't figure out how to use this mocking framework to check  that it calls `openURRLInExternalService` and it didn't seem like worrying since we're migrating to RN web views.

Artsy_Tests/View_Controller_Tests/Web_Browsing/ARInternalMobileWebViewControllerTests.m
```diff
@@ -26,27 +25,6 @@ - (WKNavigationActionPolicy)shouldLoadNavigationAction:(WKNavigationAction *)nav
...
-    NSURLRequest *request = [NSURLRequest requestWithURL:url];
-
-    id action = StubNavActionForRequest(request, WKNavigationTypeLinkActivated);
-    [controller shouldLoadNavigationAction:action];
-
-    [switchboardMock verify];
-    [switchboardMock stopMocking];
-    [fair stopMocking];
-});
-
```
ds300, line 49: [53] This fair context business hasn't been relevant for a long time.

ds300 (commented):
final block of story comments. some individual comments to follow.

src/lib/Scenes/VanityURL/VanityURLPossibleRedirect.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  if (useReactNativeWebView) {
+    return <WebView source={{ uri: url }} />
+  } else {
+    return <InternalWebView route={url} />
+  }
+}
+
+export const VanityURLPossibleRedirect: React.FC<{ slug: string }> = ({ slug }) => {
+  const [jsx, setJSX] = useState(<Loading />)
```
ds300, line 26: `VanityURLPossibleRedirect` shows a loading spinner while we send a HEAD request to force (details coming soon) to decide whether there is a redirect and what to do about it.

🔛 [23] `VanityURLPossibleRedirect` implementation\
_- - - - - - -_\
[Introducing `VanityURLPossibleRedirect `](https://github.com/artsy/eigen/pull/3771/files#r493793440) Prev ⬅️ ••• ➡️ Next **[404 fallback](https://github.com/artsy/eigen/pull/3771/files#r493799029)**

src/lib/Scenes/VanityURL/VanityURLPossibleRedirect.tsx
```diff
@@ -0,0 +1,99 @@
...
+        variant="secondaryOutline"
+        onPress={() => {
+          Linking.openURL(url)
+        }}
+      >
+        Open in browser
+      </Button>
+    </Stack>
+  )
+}
```
ds300, line 99: I added a 404 page in the (vanishingly rare) case that a vanity URL is not a valid page on force.

image

🔛 [24] 404 fallback\
_- - - - - - -_\
[`VanityURLPossibleRedirect` implementation](https://github.com/artsy/eigen/pull/3771/files#r493797958) Prev ⬅️ ••• ➡️ Next **[Performing the redirect `HEAD` request](https://github.com/artsy/eigen/pull/3771/files#r493808431)**

src/lib/Scenes/VanityURL/VanityURLPossibleRedirect.tsx
```diff
@@ -0,0 +1,99 @@
...
+          // Test this with `artsy:///auction` which force should redirect to /auctions
+
+          // TODO: add `replace` mode of navigation
+          goBack()
+          navigate(response.url)
+        }
+      })
+      .catch(() => {
+        setJSX(<NotFound url={resolvedURL} />)
+      })
```
ds300, line 67: This code handles making the HEAD request and deciding what to do about it.

The `join(webURL, slug)` pattern implements the logic that was previously in the `ARSwitchBoard#resolveRelativeURL` method. https://github.com/artsy/eigen/blob/38046929f68590032ae9df1c13e4476d74ff4586/Artsy/App/ARSwitchBoard.m#L693-L696

Returning the full `response.url` here also lets us avoid the need for the `RelativeURLWebView` component that was added recently https://github.com/artsy/eigen/blob/38046929f68590032ae9df1c13e4476d74ff4586/src/lib/Components/WebView/RelativeURLWebView.tsx#L10-L27

So this PR removes that, along with all associated code.

Reading this file now, I think I could remove a fair amount of indirection here, but I don't have time so I'll add it to the list of follow-up work.

🔛 [25] Performing the redirect `HEAD` request\
_- - - - - - -_\
[404 fallback](https://github.com/artsy/eigen/pull/3771/files#r493799029) Prev ⬅️ ••• ➡️ Next **[Goodbye `MakeMenuAware`](https://github.com/artsy/eigen/pull/3771/files#r494112752)**

src/lib/store/NativeModel.ts
```diff
@@ -87,5 +95,10 @@ listenToNativeEvents((event: NativeEvent) => {
       AppStore.actions.reset()
       AppStore.actions.native.setLocalState(event.payload)
       return
+    case "REQUEST_NAVIGATION":
+      navigate(event.payload.route)
+      return
+    default:
+      assertNever(event)
```
ds300, line 102: 💡 This `assertNever` function doesn't really exist, it was added as a type declaration in `ambient.d.ts` https://github.com/artsy/eigen/blob/151de479c835b805e3a8d5dfeef2923301da7979/src/ambient.d.ts#L36

It lets us make sure that all cases are handled in a switch statement that operates on a discriminated union type. See this [awesome guide to discriminated unions](https://basarat.gitbook.io/typescript/type-system/discriminated-unions). Don't be put-off, It's a very jargony name for a simple and powerful concept that you already understand 💪 

[100] `assertNever`

.circleci/config.yml
```diff
@@ -116,30 +117,30 @@ commands:
     steps:
       - restore_cache:
           keys:
-            - v1-gems-{{ checksum "Gemfile.lock" }}
+            - v2-gems-{{ checksum "Gemfile.lock" }}
```
ds300, line 120: bumped the cache keys in this file, hoping that it would fix an issue with the swift snapshot tests

src/lib/navigation/navigate.ts
```diff
@@ -0,0 +1,42 @@
...
+      }
+      NativeModules.ARScreenPresenterModule.presentReactScreen(
+        result.module,
+        result.params,
+        presentModally,
+        module.hidesBackButton ?? false
+      )
+    }
+  }
+}
```
ds300, line 34: `navigate`'s job is to call `matchRoute` and decide how to present the screen that it returns

`navigate` ends up either opening an external link in Safari, or calling a method in the new objective-c class `ARScreenPresenterModule`, which we'll take a look at next.

🔛 [8.1] `navigate` implementation\
_- - - - - - -_\
[Finding a route match](https://github.com/artsy/eigen/pull/3771/files#r493714588) Prev ⬅️ ••• ➡️ Next **[`ARScreenPresenterModule#presentReactScreen`](https://github.com/artsy/eigen/pull/3771/files#r493729550)**

Artsy/App/ARAppDelegate+Emission.m
```diff
@@ -235,66 +203,6 @@ - (void)setupSharedEmissionWithPackagerURL:(NSURL *)packagerURL;
...
-    emission.switchBoardModule.presentModalViewController = ^(UIViewController *_fromViewController,
-                                                              NSString *_Nonnull route) {
-        UIViewController *viewController = [[ARSwitchBoard sharedInstance] loadPath:route];
-        UIViewController *targetViewController = [ARTopMenuViewController sharedController];
-
-        // We need to accomodate presenting a modal _on top_ of an existing modal view controller. Consignments
-        // and BidFlow are presented modally, and we want to let them present modal view controllers on top of themselves.
-        if (targetViewController.presentedViewController) {
-            targetViewController = targetViewController.presentedViewController;
-        }
```
ds300, line 274: Now handled by `ARScreenPresenterModule`: https://github.com/artsy/eigen/blob/cef7e47f67722b7efc30b6656c9dd9d97a777132/Artsy/App/ARScreenPresenterModule.m#L102

Artsy/App/ARAppDelegate+Emission.m
```diff
@@ -235,66 +203,6 @@ - (void)setupSharedEmissionWithPackagerURL:(NSURL *)packagerURL;
...
-
-        // When presenting modally, view controller generally have to be wrapped in a navigation controller
-        // so the user can hit the close button. Consignments is the exception, and it has its own close button.
-        if (!([viewController isKindOfClass:[UINavigationController class]] || [viewController isKindOfClass:[LiveAuctionViewController class]])) {
-            viewController = [[ARSerifNavigationViewController alloc] initWithRootViewController:viewController];
-        }
```
ds300, line 280: We now do this for web views presented modally, and all others have their own close buttons. I've checked all the places `presentModalViewController` was being used ✅  as well as all the view controllers that previously had special-cases in `ARTopMenuViewController` ✅

Artsy/App/ARAppDelegate+Emission.m
```diff
@@ -235,66 +203,6 @@ - (void)setupSharedEmissionWithPackagerURL:(NSURL *)packagerURL;
...
-        // pushViewController:animated: method. Once that is removed, we can remove this.
-        if ([UIDevice isPhone]) {
-            viewController.modalPresentationStyle = UIModalPresentationFullScreen;
-        } else {
-            // BNMO goes through this code path instead of the one in ARTopMenuViewController.
-            if ([viewController isKindOfClass:ARSerifNavigationViewController.class] &&
-                [[[(ARInternalMobileWebViewController *)[(ARSerifNavigationViewController *)viewController topViewController] initialURL] absoluteString] containsString:@"/orders/"]) {
-                viewController.modalPresentationStyle = UIModalPresentationFormSheet;
-            }
-        }
```
ds300, line 291: BNMO is now presented with the default `PageSheet` (aka fancy modal) on both phones and tablets. If we want to change that we could do it in a follow-up PR. But it shouldn't affect UX too much.

Artsy/App/ARAppDelegate+Emission.m
```diff
@@ -344,124 +252,4 @@ - (NSDictionary *)getOptionsForEmission:(NSDictionary *)echoFeatures labOptions:
...
-                        [alert addAction:defaultAction];
-                        [[ARTopMenuViewController sharedController] presentViewController:alert animated:YES completion:nil];
-                    }
-                }];
-            }
-        }];
-    } else {
-        // nop: we don't expect Emission to call this on non-AR devices.
-    }
-}
```
ds300, line 407: Moved to `ARScreenPresenterModule`

Artsy/App/ARAppDelegate.m
```diff
@@ -340,23 +345,15 @@ - (BOOL)application:(UIApplication *)app openURL:(NSURL *)url options:(NSDiction
         NSString *urlString = [data valueForKey:@"url"];
 
         if (urlString) {
-            _landingURLRepresentation = urlString;
```
ds300, line 343: This is only ever written to, never read. We can delete it in a follow-up PR.

Artsy/App/ARSwitchBoard+Eigen.m
```diff
@@ -1,197 +0,0 @@
...
--(NSURL *)liveAuctionsURL
-{
-    BOOL useStaging = [AROptions boolForOption:ARUseStagingDefault];
-    NSString *echoDomainKey = useStaging ? @"ARLiveAuctionsStagingURLDomain" : @"ARLiveAuctionsURLDomain";
-    NSString *domain = self.echo.routes[echoDomainKey].path;
-    if (domain == nil) {
-        return [NSURL URLWithString:(useStaging ? @"https://live-staging.artsy.net" : @"https://live.artsy.net")];
-    }
-    return [NSURL URLWithString:[@"https://" stringByAppendingString:domain]];
-}
```
ds300, line 41: this is now hard-coded when we initialise emission https://github.com/artsy/eigen/blob/cef7e47f67722b7efc30b6656c9dd9d97a777132/Artsy/App/ARAppDelegate+Emission.m#L137-L146

and accessible native-side through AREmission 
https://github.com/artsy/eigen/blob/cef7e47f67722b7efc30b6656c9dd9d97a777132/emission/Pod/Classes/Core/AREmission.m#L100-L103

Artsy/App/ARScreenPresenterModule.m
```diff
@@ -0,0 +1,325 @@
...
+    [self presentViewController:vc modalPresentationStyle:modalPresentationStyle];
+}
+
+- (void)presentViewController:(UIViewController *)vc modalPresentationStyle:(UIModalPresentationStyle)modalPresentationStyle
+{
+    UIViewController *currentVC = [self currentlyPresentedVC];
+    if (![currentVC isKindOfClass:UINavigationController.class]) {
+        modalPresentationStyle = UIModalPresentationFullScreen;
+    }
+    if (modalPresentationStyle != -1) {
```
ashfurrow, line 108: *Minor comment* (not blocking): Objective-C actually has a sentinel value built in, `NSNotFound`, that is equal to `NSIntegerMax`. I think `-1` is closer to what a JS developer would be used to in Eigen so I think we should use that, but it might make sense to have this sentinel in a constant (eg: `ARModalPresentationStyleNone` or something).

Artsy/App/ARScreenPresenterModule.m
```diff
@@ -0,0 +1,325 @@
...
+
+    ARComponentViewController *vc = [[ARComponentViewController alloc] initWithEmission:nil
+                                                                    moduleName:moduleName
+                                                             initialProperties:props];
+    vc.hidesBackButton = hidesBackButton;
+
+    [self presentViewController:vc modalPresentationStyle:modalPresentationStyle];
+}
+
+- (void)presentViewController:(UIViewController *)vc modalPresentationStyle:(UIModalPresentationStyle)modalPresentationStyle
```
ashfurrow, line 102: *Thought* (non-blocking): later we could have this function accept a `RCTResponseSenderBlock` callback so JS callers can `await` on it, which might unblock some of the Auctions LAI transition UX work.

Artsy/App/ARAppDelegate+Emission.m
```diff
@@ -344,124 +252,4 @@ - (NSDictionary *)getOptionsForEmission:(NSDictionary *)echoFeatures labOptions:
...
-MakeMenuAware(ARShowComponentViewController)
-MakeMenuAware(ARShowArtworksComponentViewController)
-MakeMenuAware(ARShowArtistsComponentViewController)
-MakeMenuAware(ARShowMoreInfoComponentViewController)
-
-// Make Fairs menu-aware
-MakeMenuAware(ARFairMoreInfoComponentViewController)
-MakeMenuAware(ARFairArtistsComponentViewController)
-MakeMenuAware(ARFairArtworksComponentViewController)
-MakeMenuAware(ARFairExhibitorsComponentViewController)
```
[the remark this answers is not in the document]
ashfurrow: Oh thank goodness – of all the code I've written at Artsy, this macro is what I'm most ashamed of 😭 Glad to see it go.

Artsy/App/ARAppDelegate+Echo.m
```diff
@@ -0,0 +1,29 @@
...
+        [aero update:^(BOOL updated, NSError *error) {
+            if (!ARAppStatus.isRunningTests) {
+                [[ARAppDelegate sharedInstance] updateEmissionOptions];
+            }
+        }];
+    }];
+}
+
+
+@end
```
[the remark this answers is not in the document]
ashfurrow: *comment* (non-blocking): Moving this into its own `ARAppDelegate` is very idiomatic Objective-C, but there's a more idiomatic way to make sure Echo is only set up once. I'll follow-up with a PR 👍

Artsy/View_Controllers/Web_Browsing/ARExternalWebBrowserViewController.m
```diff
@@ -179,17 +176,23 @@ - (void)webView:(WKWebView *)webView decidePolicyForNavigationAction:(WKNavigati
...
+//        NSURL *URL = navigationAction.request.URL;
+//        ARSwitchBoard *switchboard = ARSwitchBoard.sharedInstance;
+//        if ([switchboard canRouteURL:URL]) {
+//            UIViewController *controller = [switchboard loadURL:URL];
+//            if (controller) {
+//                [switchboard presentViewController:controller];
+//            }
+//            return WKNavigationActionPolicyCancel;
+//        }
+//    }
```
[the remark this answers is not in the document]
ashfurrow: That makes sense – the biggest use case I think is editorial web views, which we should definitely follow up with those common cases mentioned in the comment.

Artsy_Tests/App_Tests/ARAppActivityContinuationDelegateTests.m
ashfurrow: Oh yeah, this is likely because `emissionMock` is a partial mock, so any non-stubbed or non-expected method calls get passed through to the underlying instance.

ashfurrow (commented):
Okay! I've finished my first pass with the comments as my guide and things look good! No blocking feedback.

I'm going to do another pass shortly, going top-to-bottom with all this context in mind. Great work!

Artsy/App/ARScreenPresenterModule.m
ds300: yeah good idea. i think for the LAI transition work we'd also need a way to pop a navigation stack which is hidden beneath a modal.

src/lib/navigation/__tests__/routes-tests.tsx
```diff
@@ -0,0 +1,1096 @@
...
+})
+
+describe("other domains", () => {
+  it("open in a browser", () => {
+    expect(matchRoute("https://google.com")).toEqual({
+      type: "external_url",
+      url: "https://google.com",
+    })
+  })
+})
```
ashfurrow, line 1096: Wow! This test file is really extensive! Awesome work – do we expect to maintain this as we add new routes?

ashfurrow (approved):

src/lib/navigation/__tests__/routes-tests.tsx
ds300: I'm not 100% sure it's worth doing but that's my plan yeah. I added a follow-up item to make sure when people add a new route they add a corresponding test, but I don't see it as being very important for app stability.

src/lib/navigation/routes.tsx
```diff
@@ -0,0 +1,144 @@
...
+    // For artists in a gallery context, like https://artsy.net/spruth-magers/artist/astrid-klein . Until we have a native
+    // version of the gallery profile/context, we will use the normal native artist view instead of showing a web view.
+    new RouteMatcher("/:profile_id_ignored/artist/:artistID", "Artist"),
+    new RouteMatcher("/auction-registration/:id", "AuctionRegistration"),
+    getCurrentEmissionState().options.AROptionsNewSalePage
+      ? new RouteMatcher("/auction/:saleID", "Auction2")
+      : new RouteMatcher("/auction/:id", "Auction"),
+    new RouteMatcher("/auction/:id/bid/:artwork_id", "AuctionBidArtwork"),
+    new RouteMatcher("/gene/:geneID", "Gene"),
+    new RouteMatcher("/show/:showID", "Show"),
```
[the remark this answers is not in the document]
damassi: Would it possible to DRY this up a bit by proving a series of objects to this array and mapping over it to invoke `new RouteMatcher` in one location? That also seems more flexible. 

Also it would be a little more developer friendly if we could used key / value names for the arguments vs bare args.

src/lib/navigation/RouteMatcher.ts
```diff
@@ -0,0 +1,66 @@
...
+          break
+        case "wildcard":
+          params["*"] = pathParts.slice(i).join("/")
+          break loop
+      }
+    }
+
+    return this.paramsMapper?.(params) ?? params
+  }
+}
```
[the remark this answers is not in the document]
damassi: I would recommend investigating [path-to-regex](https://github.com/pillarjs/path-to-regexp) its what many routing library implementations use under the hood and its pretty powerful.

src/lib/AppRegistry.tsx
```diff
@@ -363,29 +395,37 @@ const modules = defineModules({
...
+    nativeModuleName: "LocalDiscovery",
+  },
+  WebView: { nativeModuleName: "WebView" },
   Map: { Component: MapContainer, fullBleed: true },
   MyAccount: { Component: MyAccountQueryRenderer },
-  MyAccountEditEmail: { Component: MyAccountEditEmailQueryRenderer },
-  MyAccountEditName: { Component: MyAccountEditNameQueryRenderer },
-  MyAccountEditPassword: { Component: MyAccountEditPassword },
-  MyAccountEditPhone: { Component: MyAccountEditPhoneQueryRenderer },
+  MyAccountEditEmail: { Component: MyAccountEditEmailQueryRenderer, hidesBackButton: true },
```
damassi, line 411: nice!

src/lib/AppRegistry.tsx
```diff
@@ -363,29 +395,37 @@ const modules = defineModules({
...
+  MyAccountEditName: { Component: MyAccountEditNameQueryRenderer, hidesBackButton: true },
+  MyAccountEditPassword: { Component: MyAccountEditPassword, hidesBackButton: true },
+  MyAccountEditPhone: { Component: MyAccountEditPhoneQueryRenderer, hidesBackButton: true },
   MyBids: { Component: MyBidsQueryRenderer },
   AddEditArtwork: { Component: setupMyCollectionScreen(AddEditArtwork) },
   MyCollectionArtworkDetail: { Component: setupMyCollectionScreen(MyCollectionArtworkDetail) },
   MyCollectionArtworkList: { Component: setupMyCollectionScreen(MyCollectionArtworkList) },
-  MyProfile: { Component: MyProfileQueryRenderer },
+  MyProfile: { Component: MyProfileQueryRenderer, isRootViewForTabName: "profile" },
```
damassi, line 419: Also really nice 👍

src/lib/navigation/routes.tsx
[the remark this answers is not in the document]
ds300: > Would it possible to DRY this up a bit by proving a series of objects to this array and mapping over it to invoke new RouteMatcher in one location? That also seems more flexible.

I tried something along those lines at one point but it made it so that TS would only report one type error at a time, whereas like this if you make a change in AppRegistry that affects more than one route you can see all the errors at a glance.

> Also it would be a little more developer friendly and react-router-esq if we used key / value names for the constructor vs bare args.

sure, happy for someone to PR that change.

src/lib/navigation/RouteMatcher.ts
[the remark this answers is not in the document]
ds300: I looked at that lib, but the feature set goes way beyond our needs and the syntax is tricky; I didn't want folks adding routes to have to learn that stuff.

I feel like this 50-line implementation is easier to absorb than `path-to-regex`'s README.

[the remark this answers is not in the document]
damassi: This is the lib that sits under most routing libraries in the JS ecosystem; there's nothing to learn that one doesn't already know and has used a bunch across many of our codebases. If we need something that's more feature complete we can always return to this.

src/lib/Scenes/VanityURL/VanityURLPossibleRedirect.tsx
pvinis: @samchieng lets make this cool like the force 404? :D

Artsy/App/ARAppDelegate+Emission.m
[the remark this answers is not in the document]
pvinis: If it works, it works :D

src/lib/store/NativeModel.ts
pvinis: YES!

pvinis:
Super good PR! Awesome job with the changes and the explanations!!

PR #3784, opened 2020-08-26
[MyCollection] Finish up core features

damassi:

### Description

This PR somewhat brutally defeats the scoped PR rule but was moving fast. Finishes up the raw core stuff for my collections, namely 
- Adding
- Editing
- Deleting 
- And seeing stuff update on the screen after these actions are complete

Also noticed in @ds300's [PR](https://github.com/artsy/eigen/pull/3714) that some stuff was moved around, and the My Collections state stuff was stored under a `consignments` state key. I assume this was done because this work was under a `Consignments` folder, but that was some tech debt that needed to be cleaned up and things have now been moved to `/MyCollections` and the state key updated accordingly. 

⚠️ Note that most of the important changes happen in the `src/lib/Scenes/MyCollection/State/MyCollectionArtworkModel.tsx` file, but due to its size its collapsed in the PR review UI.

myc

#trivial


src/lib/store/migration.ts
```diff
@@ -26,6 +27,10 @@ export const artsyAppMigrations: Migrations = {
       navigation: {},
     }
   },
+  [Versions.RenameConsignmentsToMyCollection]: state => {
```
pepopowitz, line 30: 👍👍👍👍  this is totally the kind of thing I would have forgotten to do.

src/lib/store/AppStoreModel.ts
```diff
@@ -50,7 +50,7 @@ export const appStoreModel: AppStoreModel = {
   // APP MODULE STATE
   bottomTabs: BottomTabsModel,
   search: SearchModel,
-  consignments: ConsignmentsModel,
+  myCollection: MyCollectionModel,
```
damassi, line 53: @ds300 regarding your comment in https://github.com/artsy/eigen/pull/3714

> Migrating consignments to AppStore. This was required since the consignments screens previously used a global easy peasy store, and we can't use more than one of those in any given react tree. Now that lab options are in AppStore the consignments screens were throwing when trying to access lab options. cc @damassi

I'm wondering why this was an issue? Looking in the docs one can [assign a name to the store](https://easy-peasy.now.sh/docs/api/store-config.html) and thereby separate it from other stores within a tree. 

Adding this to the global store isn't _so_ bad, but its overhead that doesn't seem necessary atm, versus a tightly scoped easy peasy store embedded directly within the MyCollections app and unaware of the outside world.

src/lib/store/migration.ts
pepopowitz: oh, but I guess that's why it's called out in the PR template 😄

jonallured: I was def confused by this but [the doc](https://github.com/artsy/eigen/blob/master/docs/adding_state_migrations.md) was super helpful, thanks @artsy/mobile-experience team!! ❤️

damassi:
@jonallured - reassigned

src/lib/store/AppStoreModel.ts
ds300: > Looking in the docs one can assign a name to the store and thereby separate it from other stores within a tree.

I don't think that's true. The name seems to be for debugging purposes only.

If this is causing overhead for you I apologise. I wasn't anticipating any real difference in DX aside from using the centralised hooks to access state and actions. If this is slowing you down and you can find a way to avoid the need to keep the MyCollectionModel in the global store then by all means feel free to submit a PR. Although bear in mind that you haven't written any tests yet, and we've got some nice juicy infra for doing that that you'd be leaving behind.

damassi: Yeah its not that big of a deal! and no real changes. The overhead is more conceptual than anything and in practice not noticable from the way things were.

src/lib/Components/ArtworkFilterOptions/ColorOptions.tsx
```diff
@@ -137,7 +137,7 @@ export const ColorOptionsScreen: React.SFC<ColorOptionsScreenProps> = ({ navigat
   return (
     <View onLayout={handleLayout}>
       <Flex flexGrow={1}>
-        <FancyModalHeader onBackPress={handleBackNavigation}>{FilterDisplayName.color}</FancyModalHeader>
+        <FancyModalHeader onLeftButtonPress={handleBackNavigation}>{FilterDisplayName.color}</FancyModalHeader>
```
damassi, line 140: Made this a bit more generic as we now need to support left button / right button actions, we share this header with some views

src/lib/Scenes/MyCollection/State/MyCollectionArtworkModel.tsx
```diff
@@ -0,0 +1,402 @@
...
+            artistIds: [input!.artistSearchResult!.internalID as string],
+            depth: input.depth,
+            height: input.height,
+            medium: input.medium,
+            title: input.title,
+            width: input.width,
+            date: input.date,
+          },
+        },
+        updater: store => {
```
damassi, line 147: Figuring this stuff out was a bit of a journey.   Really wanted to get things working using the [declarative config](https://relay.dev/docs/en/mutations#updater-configs) but in the end couldn't figure it out. (And according to some folks on the relay slack the declarative form is gonna go away.)

src/lib/Scenes/MyCollection/State/MyCollectionArtworkModel.tsx
```diff
@@ -0,0 +1,402 @@
...
+          const payload = store
+            .getRootField("myCollectionCreateArtwork")
+            .getLinkedRecord("artworkOrError")
+            .getLinkedRecord("artworkEdge")
+
+          // Use me.id's globalID which is the parent to `myCollectionConnection`
+          const parentID = store.get("TWU6NTg4MjhiMWU5YzE4ZGIzMGYzMDAyZmJh")
```
damassi, line 154: Since there's only one `me` field, this ID is static. In a future cleanup pass we should pass this in dynamically though, just to be safe.

src/lib/Scenes/MyCollection/State/MyCollectionArtworkModel.tsx
```diff
@@ -0,0 +1,402 @@
...
+  startEditingArtwork: thunk((actions, artwork) => {
+    const dimensions = artwork.dimensions.in ?? ""
+    const [height = "", width = "", depth = ""] = dimensions
+      .replace("in", "") // FIXME: currently this only supports inches
+      .split("×")
+      .map((dimension: string) => dimension.trim())
+
+    actions.setArtworkId({
+      artworkId: artwork.internalID,
+      artworkGlobalId: artwork.id,
```
damassi, line 200: We need to pass the global id so that, similar to how we update the list view after adding an artwork, we know how which relay store record to edit after editing an artwork.

src/lib/Scenes/MyCollection/State/MyCollectionArtworkModel.tsx
```diff
@@ -0,0 +1,402 @@
...
+        },
+        updater: store => {
+          const artwork = store.get(sessionState.artworkGlobalId)
+          artwork!.setValue(input.artistSearchResult?.displayLabel, "artistNames")
+          artwork!.setValue(input.date, "date")
+          artwork!.setValue(input.depth, "depth")
+          artwork!.setValue(input.height, "height")
+          artwork!.setValue(input.medium, "medium")
+          artwork!.setValue(input.title, "title")
+          artwork!.setValue(input.width, "width")
```
damassi, line 260: Imagine this query

```graphql
artwork(id: "foo") {
  artistNames
  date
  ...
}
```
This updates code block updates the above form; and `artworkGlobalId` refers to the global id of the `artwork` identifier at the top-most part of the query.

src/lib/Scenes/MyCollection/State/MyCollectionArtworkModel.tsx
```diff
@@ -0,0 +1,402 @@
...
+              updater: store => {
+                const parentID = store.get("TWU6NTg4MjhiMWU5YzE4ZGIzMGYzMDAyZmJh") // Use me.id's globalID
+
+                if (parentID) {
+                  const connection = ConnectionHandler.getConnection(
+                    parentID,
+                    "MyCollectionArtworkList_myCollectionConnection"
+                  )
+                  if (connection) {
+                    ConnectionHandler.deleteNode(connection, input.artworkGlobalId)
```
damassi, line 324: Just as we can use the globalId to find the node to update, we can use `relay-runtime`'s `ConnectionHandler` util to find that node and delete it from a connection.

src/lib/Scenes/MyCollection/State/MyCollectionArtworkModel.tsx
```diff
@@ -0,0 +1,402 @@
...
+          // Use me.id's globalID which is the parent to `myCollectionConnection`
+          const parentID = store.get("TWU6NTg4MjhiMWU5YzE4ZGIzMGYzMDAyZmJh")
+
+          if (parentID) {
+            const connection = ConnectionHandler.getConnection(
+              parentID,
+              "MyCollectionArtworkList_myCollectionConnection"
+            )
+            if (connection) {
+              ConnectionHandler.insertEdgeBefore(connection, payload)
```
damassi, line 162: Note that the `payload` we're inserting -- the new edge -- comes directly from what we return from the mutation:

```graphql
mutation MyCollectionArtworkModelCreateArtworkMutation($input: MyCollectionCreateArtworkInput!) {
  myCollectionCreateArtwork(input: $input) {
    artworkOrError {
      ... on MyCollectionArtworkMutationSuccess {
        artworkEdge {
          ...
        }
     }
   }
}
```

And visualize further, the above block of code used to insert a new edge modifies this query structure: 

```graphql
me {
  myCollectionConnection {
    edges {
      node {
        ...
      }
    }
  }
}
```

Lastly, in order to get in to this connection, we had to identify it with a relay directive here:

https://github.com/artsy/eigen/blob/aa8663052d94e58c55da947bcf29ed9375429029/src/lib/Scenes/MyCollection/Screens/ArtworkList/MyCollectionArtworkList.tsx#L18

src/lib/store/__tests__/migration-tests.ts
```diff
@@ -150,3 +150,14 @@ describe("artsy app store migrations", () => {
     expect(Object.values(Versions).sort()).toStrictEqual(_.range(1, Object.values(Versions).length + 1))
   })
 })
+
+describe("App version Versions.RenameConsingmentsToMyCollections", () => {
+  it("renames `consignments` to `myCollection`", () => {
+    const result = migrate({ state: { version: 0 } })
+    expect(result).toMatchObject({
+      version: Versions.RenameConsignmentsToMyCollection,
+    })
```
ds300, line 159: this expect is going to fail as soon as someone bumps the version again. We should normally be providing explicit target versions to the `migrate` function in tests.

For example, the ideal version of this test would look like:

```ts

const previousState = migrate({ state: { version: 0 }, toVersion: Versions.RenameConsingmentsToMyCollections - 1 })

expect("consignments" in previousState).toBe(true)
expect("myCollection" in previousState).toBe(false)

const migratedState = migrate({state: previousState, toVersion: Versions.RenameConsingmentsToMyCollections})

expect("consignments" in migratedState).toBe(false)
expect("myCollection" in migratedState).toBe(true)
```

ds300 (commented):
Looking good. i had a few suggestions and one blocking comment but otherwise really nice work on the easy-peasy front 🎉

src/lib/Scenes/MyCollection/State/MyCollectionArtworkModel.tsx
ds300: Looks like in v10 they added new directives for deleting/appending/prepending edges after mutations, i wonder if that was to replace the `configs` api. 

https://github.com/facebook/relay/releases/tag/v10.0.0

Not really worth adopting now unless you're eager to get back into the weeds on this 🙈

pepopowitz (commented):
This is impressive. A few comments:

1. I noticed a couple commented-out lines. I'm not sure if they were unintentionally commented-out, or if they could be removed, but they're marked below. 

2. One thing that would help me more fully understand the state management of this app would be tests at some level. I don't personally like the idea of testing the store directly, but tests around the form that is using the store would help me better understand how the store is _intended_ to work. I'm not suggesting these be added as part of this PR - it's definitely work for the future. I'm curious if you've thought about what those tests would look like, though.

3. If you're interested in suggestions on how this PR could have been broken into smaller ones, here are some specific thoughts I had. (Inspired by the conversation in @admbtlr 's [PR RFC](https://github.com/artsy/README/issues/327)) 
- Start with the commits that move things from "consignments" to "MyCollection". I think this would have been the biggest payoff of all my suggestions. 
- There is at least one refactoring commit after that that could be its own PR. This method of breaking refactoring into its own PR is suggested by both [the article @zephraph linked to](https://www.thedroidsonroids.com/blog/splitting-pull-request#method-2) and [the talk @jonallured linked to](https://railsconf.com/2020/video/kent-beck-tidy-first) in the aforementioned RFC, and it plays well with the [git strategy of basing PRs on previous milestones described in the article](https://www.thedroidsonroids.com/blog/splitting-pull-request#git-strategy-3). It can be hard to predict refactoring before you start the other work, but you can always wip/stash your current work, roll back to the previous milestone, do the refactoring, and then rebase your wip/stash on the refactoring PR. 
- Following that, the remaining commits seem like they could be broken down into several PRs based on functionality, possibly by type of user action (add/edit). But also, the previous suggestions would have removed a lot of noise from the meat of this PR, and these remaining commits might just have been fine as one PR.

src/lib/Scenes/MyCollection/State/MyCollectionArtworkModel.tsx
```diff
@@ -0,0 +1,402 @@
...
+            title: input.title,
+            width: input.width,
+            date: input.date,
+          },
+        },
+        updater: store => {
+          const payload = store
+            .getRootField("myCollectionCreateArtwork")
+            .getLinkedRecord("artworkOrError")
+            .getLinkedRecord("artworkEdge")
```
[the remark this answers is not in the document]
damassi: When either the mutation _or_ the updater returns an error, the mutation error handler will execute.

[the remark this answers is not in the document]
ds300: I mean if `artworkOrError` is not an `Artwork` but some kind of error.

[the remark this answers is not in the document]
damassi: Ah, I see, good call 🤔

src/lib/Scenes/MyCollection/State/MyCollectionArtworkModel.tsx
damassi: > New Declarative connection mutations: we added several directives to make common store updates in mutations easier. An updater is no longer required for following cases: @deleteRecord (applied to a field of type ID) will delete the record with the given id from the store. @appendEdge and @prependEdge (applied to a connection edge field) will append or prepend the edge to the listed connections (07ccab7, 687d89b, 2719324)

Very cool! I think this can be a follow-up. I'll add a comment inline.

src/lib/Scenes/MyCollection/State/MyCollectionArtworkModel.tsx
damassi: We're now storing it in the store. (This bit can be removed after the relay 10 api update tho.)

src/lib/Scenes/MyCollection/State/MyCollectionArtworkModel.tsx
damassi: Not too sure -- these things are required to update the view. I think there will be room to optimize this with the relay v10 update. Will add a note.

src/lib/Scenes/MyCollection/State/MyCollectionArtworkModel.tsx
```diff
@@ -0,0 +1,402 @@
...
+
+          // Use me.id's globalID which is the parent to `myCollectionConnection`
+          const parentID = store.get("TWU6NTg4MjhiMWU5YzE4ZGIzMGYzMDAyZmJh")
+
+          if (parentID) {
```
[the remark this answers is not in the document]
damassi: Updated to `meNode` 👍

src/lib/Scenes/MyCollection/State/MyCollectionArtworkModel.tsx
```diff
@@ -0,0 +1,402 @@
...
+          },
+        },
+        updater: store => {
+          const artwork = store.get(sessionState.artworkGlobalId)
+          artwork!.setValue(input.artistSearchResult?.displayLabel, "artistNames")
+          artwork!.setValue(input.date, "date")
+          artwork!.setValue(input.depth, "depth")
+          artwork!.setValue(input.height, "height")
+          artwork!.setValue(input.medium, "medium")
+          artwork!.setValue(input.title, "title")
```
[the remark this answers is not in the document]
damassi: Added a TODO to revisit in follow-up PR around relay 10 update

damassi:
@pepopowitz - thank you for the detailed response. When I started adding commits to this pr I started feeling 😬 because of the recent convo's around git hygiene, as well as the fact that this PR is something of an exemplary example of what _not_ to do. I made a conscious decision to wing this given the presentation today, but I will certainly give those links a read and be more disciplined about this in the future.

To the rest, thank you for the feedback. I've addressed most comments, and in a few places added TODO's for followup in subsequent PRs.

PR #3788, opened 2020-08-27
Remove lab option for ArtistSeries in favor of Echo flag

sweir27:

This PR resolves [FX-2154](https://artsyproduct.atlassian.net/browse/FX-2154)

### Description

We already had a lab option for enabling the ArtistSeries features on-device. From following the instructions, I _think_ this is all I need to do (aside from actually creating the flag in Echo and naming it the same). But it's very possible I'm missing something.

## Follow-ups
- Add the flag in Echo via the UI


sweir27:
I meant to open this as a draft! But there's still one small bug to fix that we need to get out this week. So if this is the way to do it, would love a 👍 and I can merge when we're ready!

PR #3793, opened 2020-08-28
[MX-505] - Add back deleted slider assets #trivial

brainbicycle:

This PR resolves [MX-505]

### Description

Adds back some deleted images used for customizing the price range slider in Gene and Auctions refine view. 

### Screenshots

slider-with-assets

[MX-505]: https://artsyproduct.atlassian.net/browse/MX-505


brainbicycle:
> Thanks Brian, great catch! Whoops about those assets 😅 I think the fact that we had failing screenshot tests, and the just [overwrote them](https://github.com/artsy/eigen/commit/c61e32155e3b30e59ac3d765a8a9c11186ac081d?short_path=690e34f#diff-690e34feeaa183ee806378b5d5de238f), well... it speaks to how little value snapshot tests really provide.

On the other hand they did catch a legitimate layout issue in my PR earlier today!
https://github.com/artsy/eigen/pull/3739/commits/3303e0efd8cfe3846e9b5f7f08e03844b4232b8f
I think the Snapshot tests are perhaps a bit too fine-grained and they could be a bit more friendly in showing why they failed. As far as I know you are just told the snapshots don't match, which it makes it easy to just assume it is some minor layout change and overwrite them. I think if there was some friendly diff presented on failure they would be more valuable because you could easily tell if it was an actual issue or not. That is probably easier said than done though.

PR #3798, opened 2020-08-31
Update Artwork Details section to display colored dot that indicates the availability of the artwork

lilyfromseattle:
[PURCHASE-2086]

paired with @xtina-starr 

Update Artwork Details section to display colored dot that indicates the availability of the artwork.

### Description

Change “Contact for Price” to “For Sale”.

Change Sold by <gallery name> to Listed by <gallery name>

May be some logic changes. For works that are for sale we will always show “For Sale”.

# Before
Screen Shot 2020-08-31 at 10 15 00 AM

# After
Screen Shot 2020-08-31 at 10 07 50 AM

[PURCHASE-2086]: https://artsyproduct.atlassian.net/browse/PURCHASE-2086


src/lib/Scenes/Artwork/Components/CommercialInformation.tsx
```diff
@@ -110,28 +133,34 @@ export class CommercialInformation extends React.Component<CommercialInformation
...
+    if (artwork.availability?.toLowerCase() === "on loan" || artwork.availability?.toLowerCase() === "on hold") {
+      indicatorColor = color("yellow100")
+    } else if (
+      artwork.availability?.toLowerCase() === "sold" ||
+      artwork.availability?.toLowerCase() === "not for sale"
+    ) {
+      indicatorColor = color("red100")
+    } else if (artworkIsInClosedAuction) {
+      newSaleMessage = "Bidding closed"
+    } else if (artwork.saleMessage?.toLowerCase() === "contact for price" && artwork.isForSale) {
```
oxaudo, line 154: I know it might have been discussed but wondering what is the reasoning to do this change only on Mobile and not across the board in MP? I think the whole idea of this `saleMessage` was that it is going to be used across the board and consistent on devices?

This change feels different that just changing capitalization or color. It seems like it's actually changing the value of the 'Sale message'. Maybe we can move this change up to MP?

zephraph: This is a great point @oxaudo. I'm leaning towards advising to pull that into a follow up PR. Do you think that would be okay or do you think the change should be made here?

oxaudo: My question here was more high-level. I think we probably should keep the sale message displayed in UI consistent between Web and Mobile app. Probably less confusing for users and less confusing for us to debug in case there are some issues. If we agree on that I just think that updating MP logic for determining `saleMessage` in MP is less prone to bugs and confusion.

Totally fine with whatever you decide - was just wondering if you've discussed that change in this light.

PR #3800, opened 2020-08-31
add active lot standing count to profile page

erikdstock:
_One question about testing: We don't have a high-level test for the MyProfile screen where I could check for active bid count. Is this okay to skip for now in order to QA in tomorrow's session?_

### Description
  This PR adds the active lot count to the user profile page. It needs tests and is blocked by (includes commits from) #3787. Once that is merged I will rebase and add tests here.

It also fixes a few visual bugs:
- missing/broken sale end time on sale card
- Reserve not met message should not should for live lots with UnknownReserve status

One point of note - I created a `lotStandingIsClosed` helper since I was reusing this logic, and imported it into the MyProfile screen from the `MyBids` screen helpers directory. I don't know if that is okay.

- [ ] I have added tests for my changes, or my changes don't require testing, or I have included a link to a separate Jira ticket covering the tests.  _TODO: I'm not sure the best way to do this or if we should for this feature - it is small and would require some new test setup that I haven't done yet. If we deploy this we could QA it tomorrow._

image

### DIFF
--- a/src/lib/Scenes/MyBids/__tests__/ActiveLot-tests.tsx
+++ b/src/lib/Scenes/MyBids/__tests__/ActiveLot-tests.tsx
@@ -1,0 +1,89 @@
+import React from "react"
+
+import { ActiveLot_lotStanding } from "__generated__/ActiveLot_lotStanding.graphql"
+import { extractText } from "lib/tests/extractText"
+import { renderWithWrappers } from "lib/tests/renderWithWrappers"
+import { merge } from "lodash"
+import { ActiveLot } from "../Components/ActiveLot"
+
+const defaultLotStanding = {
+  isHighestBidder: true,
+  lotState: {
+    internalID: "123",
+    bidCount: 1,
+    soldStatus: "Passed",
+    reserveStatus: "ReserveMet",
+    sellingPrice: {
+      displayAmount: "CHF 1,800",
+    },
+    askingPrice: {
+      displayAmount: "CHF 2,000",
+    },
+  },
+  saleArtwork: {
+    lotLabel: "3",
+    artwork: {
+      artistNames: "Maskull Lasserre",
+      href: "/artwork/maskull-lasserre-painting",
+      image: {
+        url: "https://d2v80f5yrouhh2.cloudfront.net/zrtyPc3hnFNl-1yv80qS2w/medium.jpg",
+      },
+    },
+    sale: {
+      displayTimelyAt: "Closed on 7/15/20",
+    },
+  },
+}
+
+const lotStandingFixture = (overrides = {}) => {
+  return (merge({}, defaultLotStanding, overrides) as unknown) as ActiveLot_lotStanding
+}
+
+describe(ActiveLot, () => {
+  describe("User winning status", () => {
+    it("says 'Highest bid' if the user is winning the lot", () => {
+      const tree = renderWithWrappers(
+        <ActiveLot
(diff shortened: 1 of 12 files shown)


src/lib/Scenes/MyBids/__tests__/ActiveLot-tests.tsx
```diff
@@ -0,0 +1,71 @@
...
+      },
+    },
+    sale: {
+      displayTimelyAt: "Closed on 7/15/20",
+    },
+  },
+}
+
+const lotStandingFixture = (overrides = {}) => {
+  return (merge({}, defaultLotStanding, overrides) as unknown) as ActiveLot_lotStanding
```
erikdstock, line 39: I just did this in the fastest possible way because the nested relay types were not being cooperative with making a proper fixture.

src/lib/Scenes/MyBids/helpers/lotStanding.ts
```diff
@@ -0,0 +1,2 @@
+export const lotStandingIsClosed: (lotStanding: { lotState?: { soldStatus?: string } }) => boolean = lotStanding =>
```
erikdstock, line 1: I'm now importing and using this from a difference Scene.

src/lib/Scenes/MyProfile/MyProfile.tsx
```diff
@@ -27,6 +28,7 @@ const MyProfile: React.FC<{ me: MyProfile_me; relay: RelayRefetchProp }> = ({ me
       listRef.current?.scrollToOffset({ offset: 0, animated: false })
     })
   }, [])
+  const activeBidCount = extractNodes(me.auctionsLotStandingConnection).filter(ls => !lotStandingIsClosed(ls)).length
```
erikdstock, line 31: I wonder if we should eventually add a `active` filter to the causality query so we could just request the count rather than doing this kind of logic

src/lib/Scenes/MyProfile/MyProfile.tsx
```diff
@@ -39,6 +41,16 @@ const MyProfile: React.FC<{ me: MyProfile_me; relay: RelayRefetchProp }> = ({ me
         <MyProfileMenuItem
           title="My Bids"
           onPress={() => SwitchBoard.presentNavigationViewController(navRef.current!, "my-bids")}
+          chevron={
+            <Flex flexDirection="row" alignItems="center">
+              {activeBidCount > 0 && (
+                <Sans size="4" color="black60" mr={1}>
+                  {activeBidCount} Active
```
erikdstock, line 48: The logic here is trivial and I could not find a test with complete data for this view, so i did not add a test for it.

ds300, line 48: should probably be lower-case `active`

src/lib/Scenes/MyProfile/MyProfile.tsx
```diff
@@ -39,6 +41,16 @@ const MyProfile: React.FC<{ me: MyProfile_me; relay: RelayRefetchProp }> = ({ me
         <MyProfileMenuItem
           title="My Bids"
           onPress={() => SwitchBoard.presentNavigationViewController(navRef.current!, "my-bids")}
+          chevron={
+            <Flex flexDirection="row" alignItems="center">
+              {activeBidCount > 0 && (
+                <Sans size="4" color="black60" mr={1}>
```
ds300, line 47: MyProfileMenuItem has a `value` prop that can be used for this

ds300 (approved):
Great stuff 👍

MounirDhahri (approved):

PR #3804, opened 2020-09-01
Typeable app module definitions

ds300:

This is some groundwork for #3771 

### Description

I want to update how the top-level app modules are registered so that we can use the module names as a type in the upcoming route definitions.

Doing this now to avoid conflicts down the line.

### DIFF
--- a/src/lib/AppRegistry.tsx
+++ b/src/lib/AppRegistry.tsx
@@ -320,83 +320,91 @@ function register(screenName: string, Component: React.ComponentType<any>, optio
   AppRegistry.registerComponent(screenName, () => WrappedComponent)
 }
 
-// TODO: Change everything to BidderFlow? AuctionAction?
-register("Artist", ArtistQueryRenderer)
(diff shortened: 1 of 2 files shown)


src/lib/AppRegistry.tsx
```diff
@@ -320,83 +320,91 @@ function register(screenName: string, Component: React.ComponentType<any>, optio
...
-register("BottomTabs", BottomTabs, { fullBleed: true })
-register("Feature", FeatureQueryRenderer, { fullBleed: true })
-register("VanityURLEntityRenderer", VanityURLEntityRenderer, { fullBleed: true })
+interface ModuleDescriptor {
+  fullBleed?: boolean
+  Component: React.ComponentType<any>
+}
+
+// little helper function to make sure we get both intellisense and good type information on the result
+function defineModules<T extends string>(obj: Record<T, ModuleDescriptor>) {
```
brainbicycle, line 329: ✨

PR #3808, opened 2020-09-01
Create Lab Option to develop new inquiry UI features

lilyfromseattle:
[PURCHASE-2087]

Creates Lab Option to develop new inquiry UI features while supporting the current implementation. See documentation in eigen for creating a new lab option.

First time adding a lab option in Eigen, so I'm not sure if I added everything I needed to add, also unsure if I may have added lines that were unnecessary.

[PURCHASE-2087]: https://artsyproduct.atlassian.net/browse/PURCHASE-2087

### DIFF
--- a/Artsy/App/ARAppDelegate+Emission.m
+++ b/Artsy/App/ARAppDelegate+Emission.m
@@ -281,13 +281,13 @@ SOFTWARE.
                                            animated:ARPerformWorkAsynchronously
                                          completion:nil];
     };
-    
+
     emission.APIModule.authValidationChecker = ^() {
         if ([User currentUser]) {
             [ARAuthValidator validateAuthCredentialsAreCorrect];
         };
     };
-    
+
 
 #pragma mark - Native Module: Events/Analytics
     emission.eventsModule.eventOccurred = ^(NSDictionary *_Nonnull info) {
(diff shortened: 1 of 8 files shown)


Artsy/App/ARAppDelegate+Emission.m
```diff
@@ -327,6 +327,7 @@ - (NSDictionary *)getOptionsForEmission:(NSDictionary *)echoFeatures labOptions:
 
     options[@"AROptionsPriceTransparency"] = @([options[@"AROptionsPriceTransparency"] boolValue] || [labOptions[AROptionsPriceTransparency] boolValue]);
     options[@"AROptionsArtistSeries"] = @([options[@"AROptionsArtistSeries"] boolValue] || [labOptions[AROptionsArtistSeries] boolValue]);
+    options[@"AROptionsNewFirstInquiry"] = @([options[@"AROptionsNewFirstInquiry"] boolValue] || [labOptions[AROptionsNewFirstInquiry] boolValue]);
```
brainbicycle, line 330: I don't think this line is necessary but let me double check.

brainbicycle: Yeah I think the first one was added because of a naming difference with echo flag and lab option key and then others were just added as lab options were added copying the code. No real harm I am going to follow up to remove these and update the docs so looks fine to merge to me! 👍

Artsy_Tests/App_Tests/ARAppDelegateEmissionSpec.m
```diff
@@ -28,6 +29,7 @@
       @"AROptionsDisableNativeLiveAuctions": @(YES),
       @"AROptionsPriceTransparency": @(NO),
       @"AROptionsArtistSeries": @(NO)
```
brainbicycle, line 31:
```suggestion
      @"AROptionsArtistSeries": @(NO),
```

PR #3810, opened 2020-09-01
Avoids installing CocoaPods on CI if cache is available #trivial

ashfurrow:
The type of this PR is: **enhancement**

### Description

I noticed when merging a few PRs in a row that `pod install` was being ran every build, despite having manifest-style caching like our other dependencies. This isn't a huge deal but it takes a minute or two for CocoaPods just to download the podspecs to do dependency resolution. We already have [`cocoapods-check`](https://github.com/square/cocoapods-check) in our `Gemfile`, which compares `Podfile.lock` to `Pods/Manifest.lock` and returns the resulting exit code. This should shave some time off our builds. 

I tried to think of a reason this wouldn't work, but I think it should be safe. Open to feedback here.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -134,7 +134,7 @@ commands:
             - v5-pods-{{ checksum ".manifests/cocoapods" }}
       - run:
           name: Install Pods
-          command: bundle exec pod install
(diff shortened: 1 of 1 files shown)


ds300:
Nice one, thanks 🙏

PR #3812, opened 2020-09-01
Feature flag doc updates cleanup remove unnecessary OR #trivial

brainbicycle:

### Description

- Update some places feature flag docs are out of date
- Fixes some warnings in tests
- Remove unnecessary OR of newly added feature flag

The logical or part came up in this PR: https://github.com/artsy/eigen/pull/3808#discussion_r481372160 
My best guess is the initial price transparency feature flag had a key that didn't match the string for the echo flag so this line was necessary: https://github.com/artsy/eigen/pull/2934/files#diff-fa3274520df88be962c7afc891c3b5edR27. Also we call out in docs that we have had problems with OR'ing these flags so probably do not want to do this. 
https://github.com/artsy/eigen/blob/master/docs/developing_a_feature.md#why-do-we-need-to-remove-the-lab-flag-when-adding-echo-flag Unfortunately simply removing the OR is problematic for released features that are counting on this behavior so this was removed from PR. We are going to discuss how to proceed in front-end-ios practice before removing the others.

### DIFF
--- a/Artsy/App/ARAppDelegate+Emission.m
+++ b/Artsy/App/ARAppDelegate+Emission.m
@@ -325,6 +325,10 @@ SOFTWARE.
     // Lab options come last (as they are admin/dev controlled, giving them a chance to override)
     [options addEntriesFromDictionary:labOptions];
 
+    options[@"AROptionsPriceTransparency"] = @([options[@"AROptionsPriceTransparency"] boolValue] || [labOptions[AROptionsPriceTransparency] boolValue]);
+    options[@"AROptionsArtistSeries"] = @([options[@"AROptionsArtistSeries"] boolValue] || [labOptions[AROptionsArtistSeries] boolValue]);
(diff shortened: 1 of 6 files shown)


Artsy/App/ARAppDelegate+Emission.m
```diff
@@ -327,7 +327,6 @@ - (NSDictionary *)getOptionsForEmission:(NSDictionary *)echoFeatures labOptions:
 
     options[@"AROptionsPriceTransparency"] = @([options[@"AROptionsPriceTransparency"] boolValue] || [labOptions[AROptionsPriceTransparency] boolValue]);
     options[@"AROptionsArtistSeries"] = @([options[@"AROptionsArtistSeries"] boolValue] || [labOptions[AROptionsArtistSeries] boolValue]);
-    options[@"AROptionsNewFirstInquiry"] = @([options[@"AROptionsNewFirstInquiry"] boolValue] || [labOptions[AROptionsNewFirstInquiry] boolValue]);
```
brainbicycle, line 330: We are going to talk about removing the others in front-end-ios practice meeting, for now it should be safe to remove the new feature flag that hasn't been released yet.

ashfurrow (approved):

PR #3838, opened 2020-09-08
Add new buttons with clear call to action in commercial information

lilyfromseattle:
[PURCHASE-2088]

In Commercial Information section add new buttons with clear call to action based on price visibility and wire up new inquiry buttons to launch new inquiry modal shell.

NOTE: MessageIcon will be added after making changes to the button component to support icons

I can provide screenshots but was having simulator issues when i made this PR

[PURCHASE-2088]: https://artsyproduct.atlassian.net/browse/PURCHASE-2088


src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryButtons.tsx
```diff
@@ -0,0 +1,49 @@
...
+          Request Price
+        </Button>
+      )}
+      {!artwork.isPriceHidden && (
+        <Button onPress={() => setModalVisibility(true)} size="large" mb={1} block width={100} variant={props.variant}>
+          Inquire to Purchase
+        </Button>
+      )}
+      <Button onPress={() => setModalVisibility(true)} size="large" block width={100} variant="secondaryOutline">
+        Contact Gallery
```
xtina-starr, line 35: Was the conclusion that button component just doesn't support text with an icon?

src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
```diff
@@ -0,0 +1,22 @@
...
+  navigator?: NavigatorIOS
+  modalIsVisible: boolean
+}
+
+export const InquiryModal: React.FC<InquiryModalProps> = props => {
+  const { toggleVisibility, modalIsVisible } = props
+
+  return (
+    <FancyModal visible={modalIsVisible} onBackgroundPressed={() => toggleVisibility()}>
+      <CloseIcon onPress={() => toggleVisibility()} />
```
xtina-starr, line 19: This minor and can definitely be addressed in a follow up PR but worth noting: this will be text that says cancel instead of the close icon.

xtina-starr (commented):
🎉 this is such a great start!!! 

I think it's worth adding a spec or two. Testing the the InquiryButtons component would be dope, but I definitely think we should add something for the `CommericalButtons` component. This could test that new `else if` block you added (we might have to mock out the `getCurrentEmissionState` function). 

Let me know if this is something you want to pair on.

src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryButtons.tsx
zephraph: @lilyfromseattle can add more details but yeah, it looks like the button component itself needs to be updated. @willanderson had some ideas around that anyway, so the icon can be in a follow up.

src/lib/Scenes/Artwork/Components/CommercialButtons/CommercialButtons.tsx
```diff
@@ -79,12 +82,15 @@ export class CommercialButtons extends React.Component<CommercialButtonProps> {
...
       return <MakeOfferButton artwork={artwork} editionSetID={this.props.editionSetID} />
-    } else if (isInquireable) {
+    } else if (isInquireable && !newFirstInquiry) {
       return (
         <Button onPress={() => this.handleInquiry()} size="large" block width={100}>
           Contact gallery
         </Button>
       )
+    } else if (isInquireable && newFirstInquiry) {
+      // @ts-ignore STRICTNESS_MIGRATION
```
zephraph, line 92: **(non-blocking)** Out of curiosity, what type errors were generated before this was added? We might want to try to fix that type error instead of adding a new ts-ignore.

src/lib/Scenes/Artwork/Components/CommercialButtons/__tests__/CommercialButtons-tests.tsx
```diff
@@ -68,13 +70,73 @@ describe("CommercialButtons", () => {
     expect(commercialButtons.text()).toContain("Contact gallery")
   })
 
+  it("renders Inquire to Purchase button if isInquireable, price is visible, and lab option is true", async () => {
+    __appStoreTestUtils__?.injectEmissionOptions({ AROptionsNewFirstInquiry: true })
```
zephraph, line 74: 🙌

xtina-starr: Right!!!!

xtina-starr: Shout out to @ashleyjelks! She showed this to me when we were talking about testing yesterday.

PR #3842, opened 2020-09-09
IDFA toggle

pvinis:

This PR resolves [MX-502]

### Description

I just set `add_id_info_uses_idfa` to `false`.

In the ticket there was a link to the line with `export_compliance_available_on_french_store`, I'm not sure if that was by mistake or if we need to toggle that too.

Also, I'll try to make a build locally and run fastlane to upload the build to see if it worked of if there was any complaint. I guess we can also check it after merging with the nightly?

[MX-502]: https://artsyproduct.atlassian.net/browse/MX-502

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -10,6 +10,7 @@ upcoming:
     - Adds lab option for new inquiry flow - lily
     - Migrate away from UIWebview - mounir
     - Removes more unused Objective-C code - ash
+    - Change IDFA for AppStoreConnect to false - pavlos
   user_facing:
     - Fix app crash when tapping fair map - brian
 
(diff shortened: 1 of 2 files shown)


brainbicycle (approved):
This looks good to me 👍, pretty sure the link to the other line was a mistake. I will say that although we are no longer using it the code to access the idfa is still there so I am not sure if apple's automatic checks will be okay with it. Perhaps we can try and find out.

ashfurrow (approved):
Looks good! The link was a mistake, yes. It linked to a line number of the `Fastfile` on _master_, which isn't reliable because new lines can be added/removed from the file after the link is created.

```
https://github.com/artsy/eigen/blob/master/fastlane/Fastfile#L163
```

A better link would have looked like this:

```
https://github.com/artsy/eigen/blob/52719c2f38acabff1c32e10e32eea277e10a79e9/fastlane/Fastfile#L152
```

Which you can get to by pressing `y` in GitHub after selecting a line number.

PR #3853, opened 2020-09-11
Remove local VR flag

pvinis:

### Description

Removing the local Viewing Rooms flag. The Echo one is still there. It doesn't make much sense to keep a local one now, since VRs are enabled for users. If we ever disable them, it would be easy enough to add a local flag again or turn them on locally to test/fix something.

#trivial

⚠️ I think we don't need a migration for this, but just to double-check, @ds300 🤔.


ds300:
yes you don't need a migration 👍

brainbicycle (approved):
👍 👍

PR #3865, opened 2020-09-14
[Consignments] Updates layout on My Collection Insights view

ashleyjelks:

This PR resolves [CSGN-373](https://artsyproduct.atlassian.net/browse/CSGN-373)
Before:
Screen Shot 2020-09-14 at 5 00 58 PM

After:
Kapture 2020-09-15 at 11 24 32

#trivial

### Description
Updates the My Collection view to reflect latest design specifications.

### DIFF
--- a/src/__generated__/MyCollectionArtworkListItem_artwork.graphql.ts
+++ b/src/__generated__/MyCollectionArtworkListItem_artwork.graphql.ts
@@ -9,6 +9,7 @@ export type MyCollectionArtworkListItem_artwork = {
     readonly slug: string;
     readonly artistNames: string | null;
     readonly medium: string | null;
+    readonly title: string | null;
     readonly image: {
         readonly url: string | null;
     } | null;
@@ -65,6 +66,13 @@ const node: ReaderFragment = {
       "storageKey": null
     },
     {
+      "kind": "ScalarField",
+      "alias": null,
+      "name": "title",
+      "args": null,
+      "storageKey": null
+    },
+    {
       "kind": "LinkedField",
       "alias": null,
       "name": "image",
@@ -84,5 +92,5 @@ const node: ReaderFragment = {
     }
   ]
 };
-(node as any).hash = 'c31d382360718bb4be47c0bc5242b70f';
+(node as any).hash = 'db9df505dce68b7c6af92cd0d78e1ea9';
 export default node;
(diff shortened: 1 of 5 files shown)


damassi:
@ashleyjelks - not sure what changed based on the screen-cap? Looking at the JIRA ticket the layout doesn't quite correspond 🤔

damassi:
Also, there's a `description` section in the ticket template, if you wouldn't mind filling that out with specifics that helps with review.

src/lib/utils/formatArtworkMedium.ts
```diff
@@ -0,0 +1,22 @@
...
+    OTHER: "Other",
+    PRINT: "Print",
+    MIXED_MEDIA: "Mixed media",
+    DRAWING_COLLAGE_OR_OTHER_WORK_ON_PAPER: "Drawing, Collage, or other Work on paper",
+    PERFORMANCE_ART: "Performance Art",
+    VIDEO_FILM_ANIMATION: "Video/Film/Animation",
+    FASHION_DESIGN_AND_WEARABLE_ART: "Fashion Design and Wearable Art",
+    DESIGN_DECORATIVE_ART: "Design/Decorative Art",
+    TEXTILE_ARTS: "Textile Arts",
+  }
```
damassi, line 20: Is there anywhere else where this list might be used that we can share from?

src/lib/utils/formatArtworkMedium.ts
```diff
@@ -0,0 +1,22 @@
...
+  return mediums[medium] || capitalize(medium)
+}
```
damassi, line 22: Can we add a test for this file?

src/lib/utils/formatArtworkMedium.ts
ashleyjelks: Not really, I'd like to do a 1:1 lookup rather than looping through an array, especially since there's no limit to how long the list of artworks in a collection can be. 

It's in an array here https://github.com/artsy/eigen/blob/a13dd17b26c8e7aae1af669e005aa6bc517b82c1/src/lib/Scenes/Consignments/Screens/Metadata.tsx#L31 and here: https://github.com/artsy/eigen/blob/a13dd17b26c8e7aae1af669e005aa6bc517b82c1/src/lib/Scenes/MyCollection/Screens/AddArtwork/Components/MediumPicker.tsx#L32

damassi: For an array of this size its totally cool to use a loop; we'd get into trouble if this category list was of a variable number of items though.

src/lib/utils/formatArtworkMedium.ts
damassi: Basically, the plan is to go through the MyCollections folder and add tests 1 by 1 next sprint, but if stuff is outside of that folder it could be overlooked.

PR #3871, opened 2020-09-16
[MyCollection] Wire up artwork details view to real data

damassi:

### Description

This wires up the details view to real data, including stitched market data from diffusion, articles and auction results. Not _all_ the data has been connected however as there are some questions about what fields go where, but things are coming back from diffusion and things are wrapped in fragment containers, etc. Also covers a ton of over misc items like info modals, price index bar, and so on. 

At a good stopping point as this PR is getting kinda large. 

Next up:
- Connect up `Price Estimate` and `Market Insights` sections to data (how the field names map is confusing, need to talk to J)
- Wire up click handlers for articles / auction results to go to webpage

detail-2

#trivial


src/lib/Scenes/MyCollection/Screens/ArtworkDetail/Components/ArtworkInsights/InfoModal.tsx
```diff
@@ -0,0 +1,99 @@
...
+        }
+    }
+  }
+
+  const { title, content } = getModalContent() as { title: string; content: JSX.Element }
+  const hideModal = () => {
+    setShowModal(false)
+    setTimeout(() => {
+      navActions.showInfoModal(null)
+    }, 400)
```
damassi, line 89: This pattern comes from needing to not zero out modal view when setting the modal type. If we were to set it to null immediately the view would update and the content would flash away. So what we do is control show / hide locally, then on close, set to null so that we get nice slide down animation.

src/lib/Scenes/MyCollection/Screens/ArtworkDetail/MyCollectionArtworkDetail.tsx
```diff
@@ -1,96 +1,51 @@
...
+import { WhySell } from "./Components/WhySell"
 
 export interface MyCollectionArtworkDetailProps {
   artwork: NonNullable<MyCollectionArtworkDetailQueryResponse["artwork"]>
+  marketPriceInsights: NonNullable<MyCollectionArtworkDetailMarketInsightsQueryResponse["marketPriceInsights"]>
 }
 
-const MyCollectionArtworkDetail: React.FC<MyCollectionArtworkDetailProps> = ({ artwork }) => {
-  const dimensions = useScreenDimensions()
+const MyCollectionArtworkDetail: React.FC<MyCollectionArtworkDetailProps> = ({ artwork, marketPriceInsights }) => {
```
jonallured, line 28: **praise:** consistent levels of abstraction

This component is starting to look _really_ nice!! I opened just this file and enjoyed how easy it was to read it as an "outline" for the screen. I like how each section is at the same level of abstraction. 👍 

*format via [Conventional Comments](https://conventionalcomments.org/)*

jonallured (approved):
No real meaningful feedback from me - looks wonderful!!

PR #3873, opened 2020-09-16
[MX-517] Auction Lots grid

MounirDhahri:

This PR resolves [MX-517]

### Description

- Add auction lots grid
Screenshot 2020-09-21 at 14 47 17

[MX-517]: https://artsyproduct.atlassian.net/browse/MX-517


src/lib/Scenes/Sales/index.tsx
```diff
@@ -61,7 +61,7 @@ class Sales extends React.Component<Props, State> {
           <Stack py={2} spacing={3}>
             <SaleList title="Current Live Auctions" sales={liveAuctions} />
             <SaleList title="Current Timed Auctions" sales={timedAuctions} />
-            <LotsByFollowedArtists title={"Lots by Artists You Follow"} me={this.props.me} />
+            <LotsByFollowedArtists title={"Lots by Artists You Follow"} me={this.props.me} showUrgencyTags />
```
MounirDhahri, line 64: The title will be changed once we implement filters

src/lib/Components/ArtworkGrids/ArtworkGridItem.tsx
```diff
@@ -35,6 +39,8 @@ export const Artwork: React.FC<Props> = ({
   contextScreenOwnerId,
   contextScreenOwnerSlug,
   contextScreenOwnerType,
+  showUrgencyTags = false,
```
ds300, line 42: Should this be `true` by default?

ds300: I guess my question is: after this PR these urgency tags will only be shown on the `Sales` page. Is that correct?

MounirDhahri: right, good catch, the default value should be `true`. I adjusted it to be `hideUrgencyTags` now with `false` as a default value

src/lib/Scenes/Sale/Sale.tsx
```diff
@@ -1,27 +1,62 @@
...
 interface Props {
   sale: Sale_sale
+  me: Sale_me
+}
+
+const saleSectionsData: SaleSection[] = [
+  { key: "header" },
+  { key: "registerToBid" },
+  { key: "saleArtworksRail" },
+  { key: "saleLotsList" },
```
ds300, line 26: nonblocking: A pattern I prefer for this is to add the jsx in the array like 
```
  { key: "header", content: <SaleHeader sale={props.sale} scrollAnim={scrollAnim} /> },
  { key: "registerToBid", content: (
      <Flex mx="2" mt={2}>
        <RegisterToBidButton sale={props.sale} />
      </Flex>
    ) },
    // etc
```

and then the `renderItem` can just be `({item}) => item.content`

it's a bit less indirection since you don't need the switch case.

I'll merge this for now since it's not a big deal.

ds300 (approved):
🎉 Really nice work, thanks 🙏

src/lib/Scenes/Sale/Sale.tsx
MounirDhahri: I will refactor it now in my other PR

PR #3877, opened 2020-09-16
[My Collection] Add pagination container to My Collection Artworks List 

ashleyjelks:

This PR resolves [CSGN-375](https://artsyproduct.atlassian.net/browse/CSGN-375)

### Description

Adds a pagination container to the My Collection artworks list component.

Kapture 2020-09-15 at 11 24 32

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -6,33 +6,34 @@ upcoming:
   user_facing:
     - Fix navigation bug on Artwork consign button - david
     - Add auction lots rail to the auctions screen - mounir
+    - Enable pagination on My Collection artworks screen - ashley
 
 releases:
   - version: 6.6.3
     date: Sept 16, 2020
     dev:
-     - Connects new My Bids view to causality's lot standings - erik
-     - Continued enhancements for user clarity to My Bids view - erik
-     - Make TS type for app module names - david
-     - Disable idfa collection in Segment - brian
-     - Removes AFNetworking/UIKit subspec - ash
-     - Disable facebook access to idfa - brian
-     - Adds lab option for new inquiry flow - lily
-     - Migrate away from UIWebview - mounir
-     - Add react-native-webview behind flag - brian
-     - Removes more unused Objective-C code - ash
-     - Change IDFA for AppStoreConnect to false - pavlos
-     - Add download option for non-preview-able conversation attachments - sepand
-     - Support switching between environments on login screen - ash, brian, david, mounir, pavlos
+      - Connects new My Bids view to causality's lot standings - erik
+      - Continued enhancements for user clarity to My Bids view - erik
+      - Make TS type for app module names - david
+      - Disable idfa collection in Segment - brian
+      - Removes AFNetworking/UIKit subspec - ash
+      - Disable facebook access to idfa - brian
+      - Adds lab option for new inquiry flow - lily
+      - Migrate away from UIWebview - mounir
+      - Add react-native-webview behind flag - brian
+      - Removes more unused Objective-C code - ash
+      - Change IDFA for AppStoreConnect to false - pavlos
+      - Add download option for non-preview-able conversation attachments - sepand
+      - Support switching between environments on login screen - ash, brian, david, mounir, pavlos
     user_facing:
-     - Fix shows save button - mounir
-     - Add colored indicator dot and adjusted messaging for artwork availability - lily
-     - Fix hidden slider handle bug in refine - brian
-     - Fix app crash when opened from a 404 page - mounir
-     - Fix app crash when tapping fair map - brian
-     - Adds new inquiry buttons and modal behind lab option - lily
-     - Fix new works for you rail sorting- mounir
-     - Fix missing date issue on auctions rail at home - mounir
+      - Fix shows save button - mounir
+      - Add colored indicator dot and adjusted messaging for artwork availability - lily
+      - Fix hidden slider handle bug in refine - brian
+      - Fix app crash when opened from a 404 page - mounir
+      - Fix app crash when tapping fair map - brian
+      - Adds new inquiry buttons and modal behind lab option - lily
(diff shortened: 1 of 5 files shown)


src/lib/Scenes/MyCollection/Screens/ArtworkList/MyCollectionArtworkList.tsx
```diff
@@ -1,87 +1,141 @@
...
+          <Sans size="3" m={2}>
+            Add artwork
+          </Sans>
+        </TouchableOpacity>
+        <Sans ml={2} mb={2} size="8">
+          Artwork Insights
+        </Sans>
+      </>
+    )
+  }
```
damassi, line 59: Can we move this Header component out of this function?

src/lib/Scenes/MyCollection/Screens/ArtworkList/MyCollectionArtworkList.tsx
```diff
@@ -1,87 +1,141 @@
...
+            }
+          }
+        }
+      }
+    `,
+  },
+  {
+    getConnectionFromProps(props) {
+      return props?.me?.artworks
+    },
```
damassi, line 108: Per [this blog post](https://artsy.github.io/blog/2020/08/31/relay-pagination-containers-demystified/) I _think_ this isn't necessary...

src/lib/Scenes/MyCollection/Screens/ArtworkList/MyCollectionArtworkList.tsx
```diff
@@ -1,87 +1,141 @@
...
-import { FlatList, TouchableOpacity } from "react-native"
-
-import {
-  MyCollectionArtworkListQuery,
-  MyCollectionArtworkListQueryResponse,
-} from "__generated__/MyCollectionArtworkListQuery.graphql"
+import { MyCollectionArtworkList_me } from "__generated__/MyCollectionArtworkList_me.graphql"
+import { MyCollectionArtworkListQuery } from "__generated__/MyCollectionArtworkListQuery.graphql"
 import { defaultEnvironment } from "lib/relay/createEnvironment"
+import { MyCollectionArtworkListItemFragmentContainer } from "lib/Scenes/MyCollection/Screens/ArtworkList/MyCollectionArtworkListItem"
```
damassi, line 4:
```
MyCollectionArtworkListItemFragmentContainer as ArtworkListItem
```

damassi (approved):
A couple minor comments, but after that LGTM!

src/lib/Scenes/MyCollection/Screens/ArtworkList/MyCollectionArtworkList.tsx
ashleyjelks: Haven't seen this post, really helpful @ashfurrow !

src/lib/Scenes/MyCollection/Screens/ArtworkList/MyCollectionArtworkList.tsx
ashleyjelks: I've gotten PR feedback from MX to *not* alias pagination/fragment containers to their component names. Pinging @ds300 to see if he can remind of the reason for moving away from that?

damassi: I don't think its a hard and fast rule, but that's def what we've done all throughout the MyCollection code and I would argue against refactoring that.

ashleyjelks: Cool. Going to merge this now while its in sync with what's already upstream and follow up with @ds300 tomorrow!

CHANGELOG.yml
```diff
@@ -6,33 +6,34 @@ upcoming:
   user_facing:
     - Fix navigation bug on Artwork consign button - david
     - Add auction lots rail to the auctions screen - mounir
+    - Enable pagination on My Collection artworks screen - ashley
```
damassi, line 9: @ashleyjelks - we're not quite to the point where we need to add entries like this as nothing has been released. What we're working on right now is effectively a prototype.

src/lib/Scenes/MyCollection/Screens/ArtworkList/MyCollectionArtworkList.tsx
ds300: Yeah we decided a while back to avoid aliasing components, to make it easier to find things and navigate the codebase, which are IMO much more important concerns than avoiding long component names.

If you want you can just use the suffix `*Container` instead of `*FragmentContainer` to reduce line noise.

damassi: With the namespacing requirements of Relay some of these components can get absurdly long. What is the problem with navigation? Cmd click in VSCode goes to exactly the same place. @ds300 what navigation problems are you referring to?

ds300: find-text and find-file within a project both take you to the wrong place. Today MX had a mob session in force and I noticed it causing confusion there too. I personally remember feeling this was a persistent source of minor annoyance in reaction and eigen, and I remember others feeling similarly when we had the conversation about cleaning all the aliases up in eigen. Plus it's helpful to be reading some JSX and knowing which components take relay data and which don't. 

Consistent naming is a really important way to make codebases more maintainable and approachable IMO. Long component names look ugly and are sometimes slower to read, but at least they do not cause confusion.

damassi: 👍 cool, will update MC stuff in that case.

PR #3883, opened 2020-09-17
feat: inquiry artwork details

admbtlr:
The type of this PR is: feature

This PR resolves [PURCHASE-2092]

### Description

Adds a `CollapsibleArtworkDetails` component that is displayed at the top of the `InquiryModal`, as defined in the ticket.

image
image

I wanted to make the animation nicer, but I think it would be good to wait until `reanimated` is upgraded and `react-native-redash` added, since that will make it possible to use some nice hooks (looks like the MX team are already onboard).

[PURCHASE-2092]: https://artsyproduct.atlassian.net/browse/PURCHASE-2092


src/lib/Scenes/Artwork/Components/CommercialButtons/CollapsibleArtworkDetails.tsx
```diff
@@ -0,0 +1,60 @@
...
+  <Flex flexDirection="row" mb={1}>
+    <Text style={{ flex: 1 }}>{name}</Text>
+    <Text color="black60" style={{ flex: 1, flexGrow: 3 }}>
+      {value}
+    </Text>
+  </Flex>
+)
+
+export const CollapsibleArtworkDetails: React.FC<{ artwork: InquiryButtons_artwork }> = ({ artwork }) => {
+  // const { artwork } = props
```
xtina-starr, line 18: Stray comment here.

src/lib/Scenes/Artwork/Components/CommercialButtons/CollapsibleArtworkDetails.tsx
```diff
@@ -0,0 +1,60 @@
...
+          {!!artwork.medium && makeRow("Medium", artwork.medium)}
+          {!!artwork.dimensions && makeRow("Size", artwork.dimensions.in + "\n" + artwork.dimensions.cm)}
+          {!!artwork.editionOf && makeRow("Availability", artwork.editionOf)}
+          {!!artwork.signatureInfo?.details && makeRow("Signature", artwork.signatureInfo.details)}
+        </Flex>
+      )}
+      <Separator />
+    </>
+  ) : null
+}
```
xtina-starr, line 60: The more I look at this, I'm seeing that this component could be wrapped in a fragment container. In that case the query for these artwork metadata props would live in here and we add something like `...CollapsibleArtworkDetails_artwork` to the `InquiryButtons_artwork` fragment.

I think this is also fine as is. @zephraph do you have any thoughts?

src/lib/Scenes/Artwork/Components/CommercialButtons/CollapsibleArtworkDetails.tsx
```diff
@@ -0,0 +1,59 @@
...
+          {!!artwork.image && (
+            <OpaqueImageView
+              height={40}
+              aspectRatio={(artwork.image.width || 1) / (artwork.image.height || 1)}
+              imageURL={artwork.image.url}
+              width={40}
+              style={{ alignSelf: "center" }}
+            />
+          )}
+          <Flex ml={2} style={{ flex: 1 }}>
```
zephraph, line 39:
```suggestion
          <Flex ml={2} flex="1">
```

`flex` should be provided by styled-system's [flexbox prop api](https://styled-system.com/api#flexbox)

src/lib/Scenes/Artwork/Components/CommercialButtons/CollapsibleArtworkDetails.tsx
```diff
@@ -0,0 +1,59 @@
...
+            <Text mb={0.25}>{artwork.artist?.name}</Text>
```
zephraph, line 40: I'm not sure what the `0.25` margin value here is supposed to represent.

In our design system, we have a predetermined set of [space values](https://github.com/artsy/palette/blob/master/packages/palette/src/Theme.tsx#L156). `0.3` exists (which translates over to 3px), but there's no 0.25. I think that might get interpreted by RN as a quarter of 1dp (dp being the [density-independent pixels](https://reactnative.dev/docs/height-and-width.html#fixed-dimensions) value that RN uses for sizing)

src/lib/Scenes/Artwork/Components/CommercialButtons/CollapsibleArtworkDetails.tsx
zephraph: Yep, 💯. 

@admbtlr we pretty much _always_ try to colocate UI components w/ their data dependencies by placing a fragment container in the same file. That's generally a relay best practice. 

So the update would be as @xtina-starr suggests above:
1. Create a fragment container in this file that wraps the CollapsibleArtworkDetails component
2. Construct the fragment to have whatever data dependencies this component has
2. Remove the query updates from the `InquiryButtons_artwork` and replace it with `...CollapsibleArtworkDetails_artwork`

Happy to pair on this!

src/lib/Scenes/Artwork/Components/CommercialButtons/CollapsibleArtworkDetails.tsx
admbtlr: I guess I'm used to vertical rhythm base values being divisible in rhythmically meaningful ways (half, quarter, etc.) - I would normally always use a number like 12 or 24 as a base, to make this possible. Using 10 seems weird to me, but I guess I'll have to get used to it!

I was trying to match the spacing in the mockup, but it looks like you're right, and this has no effect at all. I'll remove it.

src/lib/Scenes/Artwork/Components/CommercialButtons/CollapsibleArtworkDetails.tsx
admbtlr: Thanks @zephraph - yes, would love to pair on this, since I'm still fumbling my way through relay.

src/lib/Scenes/Artwork/Components/CommercialButtons/CollapsibleArtworkDetails.tsx
```diff
@@ -0,0 +1,92 @@
...
+export const CollapsibleArtworkDetailsFragmentContainer = createFragmentContainer(CollapsibleArtworkDetails, {
+  artwork: graphql`
+    fragment CollapsibleArtworkDetails_artwork on Artwork {
+      image {
+        url
+        width
+        height
+      }
+      internalID
+      isPriceHidden
```
xtina-starr, line 75: minor: I don't think the `isPriceHidden` prop is needed.

admbtlr: Ah, well spotted!

zephraph (approved):

PR #3516, opened 2020-07-01
Adds Artist Series Entry Point

ashleyjelks:
Eigen needs an entry point for the Artist Series work that is a part of: https://artsyproduct.atlassian.net/browse/FX-2008

This PR adds an Artist Series View Controller and an entry point on the Admin screen.

## New admin-only entry point

#skip_new_tests

PR #3884, opened 2020-09-17
Add skeleton Fair2 view

williardx:

This PR resolves [FX-2207](https://artsyproduct.atlassian.net/browse/FX-2207)

### Description

Adds a new admin-only skeleton Fair2 view for our revamp of the Fair page. The view can be accessed in the Admin screen. The implementation here mirrors https://github.com/artsy/eigen/pull/3516, where we added the artist series skeleton view.

- [x] **Unnecessary since this is just a skeleton view.** ~I have added tests for my changes, or my changes don't require testing, or I have included a link to a separate Jira ticket covering the tests.~
- [x] **Follow up work will be future Fair2 tickets that leverage the skeleton view.** ~I have documented any follow-up work that this PR will require, or it does not require any.~
- [x] **Unnecessary** ~I have added an app state migration, or my changes do not require one. ([What are migrations?](https://github.com/artsy/eigen/blob/master/docs/adding_state_migrations.md))~

Screen Shot 2020-09-17 at 3 39 05 PM


src/lib/AppRegistry.tsx
```diff
@@ -111,37 +112,37 @@ interface ArtworkProps {
   isVisible: boolean
 }
 
-const Artwork: React.SFC<ArtworkProps> = props => <ArtworkQueryRenderer {...props} />
```
anandaroop, line 114: Is this on account of a recent linting update?

anandaroop (approved):
Thanks, looks good to me! 👍 

But I would feel better about getting MX eyes on this before merging.

(Also I'm not sure what's up with the CI fail. The file mentioned in the error is present in this PR 🤷‍♂️ )

src/lib/AppRegistry.tsx
ashfurrow: We just merged #3890 which replaces `SFC` with just `FC`.

Artsy.xcodeproj/project.pbxproj
```diff
@@ -129,6 +129,8 @@
 		3CFB078B18EB417F00792024 /* ARSecureTextFieldWithPlaceholder.m in Sources */ = {isa = PBXBuildFile; fileRef = 3CFB078A18EB417F00792024 /* ARSecureTextFieldWithPlaceholder.m */; };
 		3CFBE33618C5848900C781D0 /* ARFileUtilsTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 3CFBE33518C5848900C781D0 /* ARFileUtilsTests.m */; };
 		4191F43F1C8745D90054F823 /* ARAnalyticsVisualizer.m in Sources */ = {isa = PBXBuildFile; fileRef = 4191F43E1C8745D90054F823 /* ARAnalyticsVisualizer.m */; };
+		44152E8C2513ABEF0041B34B /* AREigenFair2ComponentViewController.m in Sources */ = {isa = PBXBuildFile; fileRef = 44152E8B2513ABEE0041B34B /* AREigenFair2ComponentViewController.m */; };
+		4441DD4B2512AC8E005DFBBC /* Fair2 in Resources */ = {isa = PBXBuildFile; fileRef = 4441DD4A2512AC8E005DFBBC /* Fair2 */; };
```
ashfurrow, line 133: Feedback (non-blocking): we can use the existing `Fair` group in Xcode, that's okay 👍

Artsy/View_Controllers/Admin/Fair2/AREigenFair2ComponentViewController.h
```diff
@@ -0,0 +1,10 @@
+#import "ARFair2ComponentViewController.h"
+#import "ARTopMenuViewController.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface AREigenFair2ComponentViewController : ARFair2ComponentViewController
```
ashfurrow, line 6: Generally, we only need Eigen-specific subclasses like this if the `.m` file does something that only it can do (like conform to the `ARMenuAwareViewController` protocol you have below). Just wanted to make sure this is done with purpose? 

(This weirdness won't be necessary much longer, after #3771 is finished.)

emission/Pod/Classes/ViewControllers/ARFair2ComponentViewController.h
```diff
@@ -0,0 +1,22 @@
+#import <Emission/ARComponentViewController.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface ARFair2ComponentViewController : ARComponentViewController
```
ashfurrow, line 5: Since this view controller only does boilerplate work, another option is to forsake it entirely and just instantiate its superclass with the module name and props you need, like here:

https://github.com/artsy/eigen/pull/3785/files#diff-14845bae47c92c89fde88c9dcc3a80bfR77

src/lib/AppRegistry.tsx
williardx: I just rebased this PR on master so this isn't part of the changes anymore.

ashfurrow (approved):
This is mergeable – none of my feedback is blocking. Our navigation strategy is in flux right now and I wouldn't want to slow down your work for the sake of adhering to best practices that we don't yet have defined or documented 👍 

That said, this approach does a lot of work that we don't need to do, even without #3771 being merged soon. I'm curious where some of these patterns came from – are there any docs we can update in the meantime? /cc @ds300 

I linked to it in a comment but https://github.com/artsy/eigen/pull/3785 does something very similar, adding a new component to replace an Swift view controller (where this PR replaces an existing React component). Let me know what I can clarify.

williardx:
@ashfurrow This approach is based on https://github.com/artsy/eigen/pull/3516, where the FX team added a new admin-only skeleton view for artist series. I effectively copied what that PR did, so if something seems strange there's not necessarily a good reason for it.

brainbicycle (approved):

PR #3890, opened 2020-09-18
Replace SFC with FC

pvinis:

#trivial

### Description

It's one of those evenings, were your brain can't unsee 
Screenshot 2020-09-18 at 15 08 21

Now it's fixed.

### DIFF
--- a/src/lib/AppRegistry.tsx
+++ b/src/lib/AppRegistry.tsx
@@ -111,7 +111,7 @@ interface ArtworkProps {
   isVisible: boolean
 }
 
-const Artwork: React.SFC<ArtworkProps> = props => <ArtworkQueryRenderer {...props} />
+const Artwork: React.FC<ArtworkProps> = (props) => <ArtworkQueryRenderer {...props} />
 
 interface PartnerProps {
   partnerID: string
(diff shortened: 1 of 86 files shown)


ashfurrow (approved):
Yes, yes!

src/lib/AppRegistry.tsx
```diff
@@ -407,9 +407,9 @@ for (const moduleName of Object.keys(modules)) {
 }
 
 const Main: React.FC<{}> = track()(({}) => {
-  const isHydrated = AppStore.useAppState(state => state.sessionState.isHydrated)
-  const isLoggedIn = AppStore.useAppState(state => !!state.native.sessionState.userID)
-  const onboardingState = AppStore.useAppState(state => state.native.sessionState.onboardingState)
+  const isHydrated = AppStore.useAppState((state) => state.sessionState.isHydrated)
```
damassi, line 410: @ds300 - What did we settle on in terms of prettier churn from the other day? Noticing this kind of stuff in PRs lately

pvinis: I'll jump in here :D prettier 2 has this as a default, and the [reason](https://prettier.io/blog/2020/03/21/2.0.0.html#change-default-value-for-arrowparens-to-always-7430httpsgithubcomprettierprettierpull7430-by-kachkaevhttpsgithubcomkachkaev) (which i agree with) is that it makes it easier to add args/types etc.

PR #3896, opened 2020-09-19
[MyCollection] Refactor navigator flow 

damassi:

### Description

This PR continues work started in https://github.com/artsy/eigen/pull/3893 and refactors our navigation flow to exclusively use NavigatorIOS for pushing / popping views. This keeps the UX consistent and uniform and simpler. Additionally, fixes some bugs around editing an artwork and wires up the basic cost fields.

Follow-up
- Delete now unused Obj-C view controller files 
- Wire up cost field to translate between currencies 
- Add dirty form check on edit artwork cancel button check -- right now its specific to create, not edit
- General code cleanup / types 

cc @artsy/csgn-devs 

#### All app transitions

mc


src/lib/Scenes/Consignments/Screens/Overview.tsx
```diff
@@ -243,56 +245,64 @@ export default class Overview extends React.Component<Props, State> {
...
-              {!this.props.isArrivingFromMyCollection && (
-                <Button variant="noOutline" onPress={() => SwitchBoard.dismissModalViewController(this)}>
-                  Close
-                </Button>
-              )}
-            </Flex>
-          </View>
-        </ScrollView>
-      </Theme>
+      <>
```
damassi, line 248: This consignments stuff is just temporary, we're close to being able to merge the two apps.

src/lib/Scenes/MyCollection/Boot.tsx
```diff
@@ -36,7 +37,9 @@ export const setupMyCollectionScreen = (Component: React.ComponentType<any>) =>
     return (
       <View ref={navViewRef} style={{ flex: 1 }}>
         <FormikProvider value={initialForm}>
-          <Component {...props} />
+          <Navigator name="main">
```
damassi, line 40: This fixes a pretty nasty bug which relates to having multiple layers of `NavigatorIOS` instances -- the main one, which allows us to slide between screens, and the fancy modal instance, for the "more details screen". In order to know which layer we're on I had to give it a name, which is then looked up.

src/lib/Scenes/MyCollection/Screens/AddArtwork/Screens/AddPhotos.tsx
```diff
@@ -15,7 +15,7 @@ export const AddArtworkAddPhotos = () => {
 
   return (
     <ScrollView>
-      <FancyModalHeader onLeftButtonPress={() => navActions.goBack()}>
+      <FancyModalHeader onLeftButtonPress={() => navActions.goBackInModal()}>
```
damassi, line 18: If we're in a modal view, we'll use this action; it targets the correct NavigatorIOS instance.

src/lib/Scenes/MyCollection/Screens/ArtworkDetail/Components/MyCollectionArtworkMeta.tsx
```diff
@@ -33,7 +46,7 @@ export const MyCollectionArtworkMeta: React.FC<MyCollectionArtworkMetaProps> = (
         <Field label="Materials" value={capitalize(category as string)} />
         <Field label="Dimensions" value={dimensions} />
         <Field label="Edition" value="TODO" />
-        <Field label="Price paid" value="TODO" />
+        <Field label="Price paid" value={`${costMinor} ${costCurrencyCode}`} />
```
damassi, line 49: I'm not doing any translation between minor cents to dollars, etc, just yet

src/lib/Scenes/MyCollection/Screens/ArtworkDetail/MyCollectionArtworkDetail.tsx
```diff
@@ -17,109 +13,114 @@ import { Button, Flex, Join, Spacer } from "palette"
...
+    id
+    image {
+      url
+    }
+    internalID
+    medium
+    metric
+    title
+    width
+  }
```
damassi, line 91: This bit is the most meh part of the my collections app IMO, and we should think  of a way to make this a little more automatic. 

Whats happening is: 
- we load an artwork detail page view and populate it with data
- fields in this fragment are all **editable** 
- when we click the edit button on the artwork, we pass these returned values to  formik as initial values to be edited

This all seems pretty ok / standard. _However_...

We now need to update two other places with similar data! And when additional fields are added we need to be sure to update those other places as well:
- we need to update the mutation input on create / edit
- we need to update the mutation `updater` function to then update the relay store so that when we submit and the edit modal is dismissed, the view updates with the latest changes. (This last point should be fiixed when we upgrade Relay and swap in the new Relay 10 mutation API.)

src/lib/Scenes/MyCollection/Screens/ArtworkList/MyCollectionArtworkListItem.tsx
```diff
@@ -30,21 +32,29 @@ export const MyCollectionArtworkListItem: React.FC<MyCollectionArtworkListItemPr
...
 
   return (
-    <TouchElement onPress={onPress}>
+    <TouchElement
+      onPress={() =>
+        navActions.navigateToArtworkDetail({
+          artistInternalID: artist!.internalID,
+          artworkSlug: slug,
+          medium,
+        })
```
damassi, line 55: The way this used to be (which broke some stuff) was we would stack two relay `QueryRenderers` in sequence. The first would fetch data that we would then pass to the second in order to fetch diffusion data. I realized we could combine everything into one query by expanding the intial input data.

src/lib/Scenes/MyCollection/Screens/ConsignmentsHome/ConsignmentsHome.tsx
```diff
@@ -39,15 +40,20 @@ export const ConsignmentsHome: React.FC<Props> = ({ targetSupply, isLoading, isA
...
-        <HowItWorks />
-        <ArtistList targetSupply={targetSupply} isLoading={isLoading} />
-        <Footer onConsignPress={handleConsignPress} />
-      </Join>
-    </ScrollView>
+    <>
+      <ScrollView ref={navRef}>
+        {!!isArrivingFromMyCollection && (
+          <FancyModalHeader onLeftButtonPress={() => navActions.goBack()} hideBottomDivider />
+        )}
```
damassi, line 47: When consign / myCollection is combined this will go away.

src/lib/Scenes/MyCollection/State/MyCollectionArtworkModel.tsx
```diff
@@ -314,6 +305,8 @@ export const MyCollectionArtworkModel: MyCollectionArtworkModel = {
           const artwork = store.get(sessionState.artworkGlobalId)
           artwork!.setValue(input.artistSearchResult?.displayLabel, "artistNames")
           artwork!.setValue(input.category, "category")
+          artwork!.setValue(input.costMinor, "costMinor")
+          artwork!.setValue(input.costCurrencyCode, "costCurrencyCode")
```
damassi, line 309: You can see the "two other spots" areas of the code (as mentioned above) here and in the previous file.

src/lib/Scenes/MyCollection/Screens/ConsignmentsHome/ConsignmentsHome.tsx
jonallured: That will be a happy day!

src/lib/Scenes/MyCollection/Boot.tsx
jonallured: Seems like a fine approach to me!

src/lib/Scenes/MyCollection/Components/Navigator.tsx
```diff
@@ -1,35 +1,39 @@
...
+import NavigatorIOS from "react-native-navigator-ios"
 
 /**
  * Navigator is a helper component for automatically connecting a component tree
  * to State/MyCollectionNavigationModel. When wrapped, we can dispatch actions
  * and push / pop views from a central location (the model) to NavigatorIOS.
  */
-export const Navigator: React.FC<Route> = ({ children, title = "" }) => {
+
+export type NavigatorTarget = "main" | "modal"
```
jonallured, line 12: Hopefully this list won't grow! 🙈

src/lib/Scenes/MyCollection/Screens/AddArtwork/Screens/AddPhotos.tsx
jonallured: Did you consider an API more like:

```
goBack("modal")
```

Probably not worth it, but should the list of nav names grow, could be something to think about.

src/lib/Scenes/MyCollection/Screens/ArtworkDetail/MyCollectionArtworkDetail.tsx
jonallured: Hrm, good call out - I don't have a good thought here but I know this pain. A higher-order abstraction might help but I can't see what it would look like from where I'm sitting. 🤔

src/lib/Scenes/MyCollection/utils/formatArtworkDimensions.ts
```diff
@@ -7,7 +7,12 @@ interface Props {
   metric: "in" | "cm" | string | null
 }
 
-export function formatArtworkDimensions({ height, width, depth, metric }: Props): string {
+export function formatArtworkDimensions({ height, width, depth, metric }: Props): string | null {
+  const hasValues = [height, width, depth, metric].every(Boolean)
```
jonallured, line 11: I think this means that all four values are required in order to format the dimensions, right? But I think it's likely that for a painting one would only provide width and height, for example, no? Is this used prior to the creation of the artwork in Gravity? I ask because I believe artwork has a dimensions property that should do this for us!

jonallured (approved):
Nothing blocking from my end and this looks so good!! 🎉 Going to merge and that should set us up for some QA which will be nice. Thank you!!

src/lib/Scenes/MyCollection/Screens/AddArtwork/Screens/AddPhotos.tsx
damassi: I did -- but used something a little more explicit: 
```
goBack({ navigator: 'main' }) 
```
But then remembered its trivial to spin up new actions which are self-describing and handle implementation for the user and went this route.

src/lib/Scenes/MyCollection/utils/formatArtworkDimensions.ts
damassi: Ah good point -- will fix 👍

PR #3897, opened 2020-09-21
refactor(tests): replacing some enzyme with react-test-renderer

pvinis:

This PR resolves [MX-]

### Description

I'd like this merged so we have the dep ready, and we can start using it randomly when we want.

Changes:
- `Button`s have a `HiddenText` in them that is used to make the buttons sized big enough to fit the largest text we will put inside that button. Now it is only used outside tests, because otherwise tests see the buttons having texts like `FollowFollowed` when it's just `Follow`.

~~PR submitted to RNTL: https://github.com/callstack/react-native-testing-library/pull/569~~

### Follow-up work

- [ ] wrapper
I made `renderWithWrappersTL` for this PR because the existing `renderWithWrappers` has some extra things like
```
 const wrappedComponent = componentWithWrappers(component)
  // tslint:disable-next-line:use-wrapped-components
  const renderedComponent = ReactTestRenderer.create(wrappedComponent)
  // monkey patch update method to wrap components
  const originalUpdate = renderedComponent.update
  renderedComponent.update = (nextElement: ReactElement) => {
    originalUpdate(componentWithWrappers(nextElement))
  }
  return renderedComponent
```
that I didn't look too much into. I would like to have a specific case where the new one doesn't work. The eventual follow-up work here would be to only have one of these at some point.

- [ ] Add a TODO on every file using enzyme, to prompt devs to migrate that file from enzyme to RNTL

### DIFF
--- a/src/lib/tests/renderWithWrappers.tsx
+++ b/src/lib/tests/renderWithWrappers.tsx
@@ -1,3 +1,4 @@
+import { render } from "@testing-library/react-native"
 import { ToastProvider } from "lib/Components/Toast/toastHook"
 import { GlobalStoreProvider } from "lib/store/GlobalStore"
 import { ProvideScreenDimensions } from "lib/utils/useScreenDimensions"
@@ -6,11 +7,32 @@ import React from "react"
 import ReactTestRenderer from "react-test-renderer"
 import { ReactElement } from "simple-markdown"
 
+export const Wrappers: React.FC = ({ children }) => {
+  return (
+    <GlobalStoreProvider>
+      <Theme>
+        <ToastProvider>
+          <ProvideScreenDimensions>{children}</ProvideScreenDimensions>
+        </ToastProvider>
+      </Theme>
+    </GlobalStoreProvider>
+  )
+}
+
+/**
+ * Returns given component wrapped with our page wrappers
+ * @param component
+ */
+export const componentWithWrappers = (component: ReactElement) => {
+  return <Wrappers>{component}</Wrappers>
+}
+
 /**
  * Renders a React Component with our page wrappers
  * @param component
+ * @deprecated Try to use `renderWithWrappers`. If there is any problem and this function works, please report it to CX.
  */
-export const renderWithWrappers = (component: ReactElement) => {
+export const renderWithWrappers_legacy = (component: ReactElement) => {
   const wrappedComponent = componentWithWrappers(component)
   try {
     // tslint:disable-next-line:use-wrapped-components
@@ -38,17 +60,9 @@ export const renderWithWrappers = (component: ReactElement) => {
 }
 
 /**
- * Returns given component wrapped with our page wrappers
+ * Renders a React Component with our page wrappers
  * @param component
  */
-export const componentWithWrappers = (component: ReactElement) => {
-  return (
-    <GlobalStoreProvider>
-      <Theme>
-        <ToastProvider>
-          <ProvideScreenDimensions>{component}</ProvideScreenDimensions>
-        </ToastProvider>
-      </Theme>
-    </GlobalStoreProvider>
-  )
+export const renderWithWrappers = (component: ReactElement) => {
+  return render(component, { wrapper: Wrappers })
 }
--- a/package.json
+++ b/package.json
@@ -148,6 +148,8 @@
     "@octokit/rest": "16.34.0",
     "@omakase/cli": "0.0.10",
     "@react-native-community/cli-platform-ios": "3.0.0",
+    "@testing-library/jest-native": "3.4.3",
+    "@testing-library/react-native": "7.1.0",
     "@types/chalk": "2.2.0",
     "@types/dedent": "0.7.0",
     "@types/jest": "25.1.5",
--- a/src/lib/Components/Toast/__tests__/Toast-tests.tsx
+++ b/src/lib/Components/Toast/__tests__/Toast-tests.tsx
@@ -21,15 +21,15 @@ describe("Toasts", () => {
   it("renders a toast when", async () => {
     const tree = renderWithWrappers(<TestRenderer />)
 
-    expect(tree.root.findAllByType(Toast)).toHaveLength(0)
+    expect(tree.UNSAFE_queryAllByType(Toast)).toHaveLength(0)
 
-    const buttonInstance = tree.root.findByType(Button)
+    const buttonInstance = tree.UNSAFE_getByType(Button)
     act(() => buttonInstance.props.onPress())
 
-    expect(tree.root.findAllByType(Toast)).toHaveLength(1)
+    expect(tree.UNSAFE_queryAllByType(Toast)).toHaveLength(1)
 
     jest.advanceTimersByTime(3000)
 
-    expect(tree.root.findAllByType(Toast)).toHaveLength(0)
+    expect(tree.UNSAFE_queryAllByType(Toast)).toHaveLength(0)
   })
 })
--- a/src/lib/Scenes/MyBids/__tests__/ActiveLotStanding-tests.tsx
+++ b/src/lib/Scenes/MyBids/__tests__/ActiveLotStanding-tests.tsx
@@ -2,7 +2,7 @@ import React from "react"
 
 import { ActiveLotStanding_lotStanding } from "__generated__/ActiveLotStanding_lotStanding.graphql"
 import { extractText } from "lib/tests/extractText"
-import { renderWithWrappers } from "lib/tests/renderWithWrappers"
+import { renderWithWrappers_legacy } from "lib/tests/renderWithWrappers"
 import { merge } from "lodash"
 import { ActiveLotStanding } from "../Components/ActiveLotStanding"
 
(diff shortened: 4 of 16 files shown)


src/lib/Scenes/Artwork/Components/__tests__/ContextCard-tests.tsx
```diff
@@ -15,82 +14,68 @@ jest.unmock("react-relay")
...
-        <Theme>
-          <ContextCard relay={{ environment: {} } as RelayProp} artwork={fairContextArtwork as any} />
-        </Theme>
-      )
-      expect(component.find(EntityHeader).length).toEqual(1)
-
-      expect(component.find(EntityHeader).text()).toContain(`Market Art + Design 2019`)
+      const tree = renderWithWrappers(
+        <ContextCard relay={{ environment: {} } as RelayProp} artwork={fairContextArtwork as any} />
+      ).root
```
pvinis, line 19: I thought it might be a good idea to add `.root` here rather than inside the expect like `expect(tree.root.findAll..)` 🤔.

No idea if this might be making other things trickier. I'm open to get some pros or cons for this.

src/lib/Scenes/Artwork/Components/__tests__/CommercialPartnerInformation-tests.tsx
```diff
@@ -1,23 +1,21 @@
...
-    expect(component.find(Sans).at(3).render().text()).toMatchInlineSnapshot(`"VAT included in price"`)
+    const tree = renderWithWrappers(
+      <CommercialPartnerInformation
+        // @ts-ignore STRICTNESS_MIGRATION
+        artwork={CommercialPartnerInformationArtwork}
+      />
+    ).root
+
+    expect(tree.findAllByType(Text)[0].props.children.join("")).toMatch("From Bob's Gallery")
+    expect(tree.findAllByType(Text)[1].props.children.join("")).toMatch("Ships from Brooklyn")
```
pvinis, line 16: I added `.join("")` since `Text` has an array of string to form the final text. If there is a better way for this, let me know.

src/lib/Scenes/City/Components/BMWEventSection/__tests__/index-tests.tsx
```diff
@@ -21,24 +20,22 @@ const show = {
...
+          },
+          featuredShows: [show as any],
+        }}
+        relay={{ environment: {} } as RelayProp}
+        citySlug={"new-york-us"}
+      />
+    ).root
 
-    expect(comp.text()).toContain("BMW Art Guide")
+    expect(tree.findAllByType(Text)[0].props.children).toMatch("BMW Art Guide")
```
pvinis, line 39: In a few places we use `toContain`, `toMatch` and `toBe`. I tried to look for the differences, and I think for strings that are exact we should use `toBe`, for strings we want to just check a substring we should use `toMatch`, and only for arrays we should use `toContain`.

There is an interesting thing here where `toContain` works for `Text`s that have multiple children combined into one string. I tried using `.join("")` (I have a comment in this PR about this. Interestingly we could avoid using `.join("")` and use `toContain`, and that kinda gives us the same as using `.join("")` with `toBe`.

Thoughts? 🤔

src/lib/Scenes/Collection/Screens/__tests__/CollectionHeader-tests.tsx
```diff
@@ -40,54 +35,39 @@ describe("collection header", () => {
...
-    const wrapper = mount(
-      <Theme>
-        <CollectionHeader {...props} />
-      </Theme>
-    )
+    const tree = renderWithWrappers(<CollectionHeader {...props} />).root
 
-    expect(wrapper.find(ReadMore).exists()).toBe(true)
-    expect(wrapper.find(ReadMore).find(Sans).text()).toContain(
+    expect(tree.findAllByType(ReadMore)).toHaveLength(1) // doest work?
```
pvinis, line 68: I need to figure out why this doesn't seem to find the component.

src/lib/Scenes/Collection/Screens/__tests__/CollectionHeader-tests.tsx
```diff
@@ -40,54 +35,39 @@ describe("collection header", () => {
...
+    expect(tree.findAllByType(Text)[1].props.children.join("")).toBe(
       "A beach towel by Yayoi Kusama, a classic print by Alexander Calder, or a piggy bank by Yoshitomo Nara"
-    )
+    ) // maybe .join("") is not the best way?
```
pvinis, line 71: One problem with the `.join("")` I ran into here is that the array is deeply-nested, so the string coming back was something like `"A beach towel by [Object object], a classic print by [Object object], or a piggy bank by [Object object]"`.

🤔 Again, not sure how the best way would be here, so I'd like opinions. Maybe a helper func to deeply join? In this specific case here `toContain` would still not help us.

src/lib/Scenes/Artwork/Components/__tests__/ContextCard-tests.tsx
ashfurrow: Makes sense to me 👍 Not going through `.root` is the exception, and we should optimize for the common case.

src/lib/Scenes/Artwork/Components/__tests__/CommercialPartnerInformation-tests.tsx
ashfurrow: Could we use the `extractText` helper?

pvinis: I don't know about that one. I'll check it tomorrow!

pvinis:
I'll use this PR to play around with react-native-testing-library and see how the helpers are used and can help us, like `fireEvent` and `getByText`, compared to raw `act` and things like `findAllByType(Button)[1].props.children`.

admbtlr:
Maybe change the PR title to `chore: replace some enzyme with react-test-renderer`? 😬

MounirDhahri (approved):

MounirDhahri:
This looks good to me, I left few suggestions/questions that we not clear to me. 
Also, it looks like the build is failing some tests!

src/lib/tests/renderWithWrappers.tsx
```diff
@@ -36,13 +54,9 @@ export const renderWithWrappers = (component: ReactElement) => {
...
+ * Renders a React Component with our page wrappers
  * @param component
  */
-export const componentWithWrappers = (component: ReactElement) => {
-  return (
-    <GlobalStoreProvider>
-      <Theme>{component}</Theme>
-    </GlobalStoreProvider>
-  )
+export const renderWithWrappersTL = (component: ReactElement) => {
```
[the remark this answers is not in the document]
brainbicycle: Agreed think we can rename to make it clear it is deprecated/legacy

src/palette/elements/Button/Button.tsx
```diff
@@ -44,6 +44,7 @@ export interface ButtonProps extends ButtonBaseProps {
    * to add haptic feedback on the button.
    */
   haptic?: HapticFeedbackTypes | true
+  testID?: string
```
pvinis, line 47:
```suggestion
```

src/palette/elements/Button/Button.tsx
```diff
@@ -272,9 +273,12 @@ export const Button: React.FC<ButtonProps> = (props) => {
                     {children}
                   </Sans>
                 </VisibleTextContainer>
-                <HiddenText role="presentation" weight="medium" size={s.size}>
-                  {longestText ? longestText : children}
-                </HiddenText>
+                {!__TEST__ && (
```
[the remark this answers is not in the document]
pvinis: So there is this funky thing we do in Button, and it's tripping the tests.

When a Button has two texts, say `Follow` and `Following`, we put both texts in the button, in order to make the button get the right size, so that both texts fit. We use `HiddenText` for that.

The problem with this is that during tests, the text that is extracted from a Button would be something like `FollowFollowing`, because even though it's hidden, it's still "visible" in the tests.

I added this `__TEST__` check in order to **not** render the hidden text in the tests, and only render the text we give by default, so we can have our tests run correctly.

src/lib/tests/renderWithWrappers.tsx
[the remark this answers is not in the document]
pvinis: Hm. I will look into that.

src/palette/elements/Button/Button.tsx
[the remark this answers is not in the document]
MounirDhahri: thanks for explaining!

src/lib/tests/renderWithWrappers.tsx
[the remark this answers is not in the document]
pvinis: ~~I tried renaming the old but that doesn't just work because of the different return objects. The old one has things like `tree.root.findByType`, the new has `tree.UNSAFE_findByType` with root missing etc. The API is different so I would need to replace in every file.

Because I'd like this merged so we can start using it and replace enzyme, I see two options:
- We keep the old as it is, we add the new `renderWithWrappersTL`, and we go from there.
- I rename the old one and replace the name in all test files. Then we slowly start only using the new.

I see option 2 as slightly better for "future planning", as the more time goes by, the more ok things will be. But I also see that it will touch a bunch of files (which is fine I guess?).
~~

MounirDhahri:
I will close this for inactivity, please feel free to open if it's still relevant

PR #3900, opened 2020-09-22
protect from having a nil string

pvinis:

This PR resolves [MX-]

### Description

`domain` can get `nil` as a value and that crashes app in the line below with `stringByAppendingString`.

We add a check and return a silly url to make the app not crash. This happens when the locally saved echo.json is broken, something like `{"error": "forbidden"}`. This shouldn't happen ever and we will address this in the next PR. In this PR we are trying to fix the crashing of the app in the case that the broken echo.json got saved, and instead of crashing forever, we just can't access auctions?

Are we thinking correctly here or do we need to do something else?

#trivial 


Artsy/App/ARSwitchBoard+Eigen.m
```diff
@@ -34,6 +34,9 @@ -(NSURL *)liveAuctionsURL
     BOOL useStaging = [AROptions boolForOption:ARUseStagingDefault];
     NSString *echoDomainKey = useStaging ? @"ARLiveAuctionsStagingURLDomain" : @"ARLiveAuctionsURLDomain";
     NSString *domain = self.echo.routes[echoDomainKey].path;
+    if (domain == nil) {
+        return [NSURL URLWithString:@"https://notready.yet"];
```
ds300, line 38: might as well use the actual values here

```suggestion
        return [NSURL URLWithString:(useStaging ? @"live-staging.artsy.net" : @"live.artsy.net")];
```

ds300, line 38: might as well use the actual values here 

```suggestion
        return [NSURL URLWithString:(useStaging ? @"https://live-staging.artsy.net" : @"https://live.artsy.net")];
```

pvinis: Makes sense. No harm in duplicating these here.

ashfurrow (approved):
Looks good 👍

PR #3901, opened 2020-09-22
FX-2265: User can filter artworks within an artist series

sweir27:

This PR resolves [FX-2265]

### Description

This PR adds the ability to filter the artworks on the artist series view. On the way, I had to make a couple of adjustments to existing components.

Full list of changes is:
- Converts the top-level `ArtistSeries` view to be a `FlatList`. This is both because my understanding is that FlatLists are slightly more performant (especially given the artist series view has components far below-the-fold), and because FlatLists provide helpers for understanding the placement of components on the screen-- this made it possible for us to conditionally render the Filter button.
  - Caveat: It's possible you can achieve something similar with a `ScrollView`, but I didn't see anything obvious in my research.
- Using the existing filter components, adds the ability to filter the artworks on the artist series view.
- Slightly adjusts the spacing around the `FilterArtworksZeroState` component, which was somewhat specific to the existing uses but didn't work for artist series.

**Zero state for Artist Series:**
Screen Shot 2020-09-22 at 9 12 11 AM

**Zero state for Collections:**
Screen Shot 2020-09-22 at 9 38 57 AM

**Full gif of behavior:**
filter-artist-series

[FX-2265]: https://artsyproduct.atlassian.net/browse/FX-2265

### DIFF
--- a/src/lib/Components/ArtworkGrids/FilteredArtworkGridZeroState.tsx
+++ b/src/lib/Components/ArtworkGrids/FilteredArtworkGridZeroState.tsx
@@ -18,9 +18,9 @@ export const FilteredArtworkGridZeroState: React.FC<ZeroStateProps> = (props) =>
   }
 
   return (
-    <ZeroStateContainer>
+    <Flex flexDirection="column" px={4}>
       <ZeroStateMessage size="3">Unfortunately, there are no works that meet your criteria.</ZeroStateMessage>
-      <ButtonBox>
+      <Flex m="0 auto" pt={2}>
         <Button
           size="medium"
           variant="secondaryGray"
@@ -31,21 +31,12 @@ export const FilteredArtworkGridZeroState: React.FC<ZeroStateProps> = (props) =>
         >
           Clear filters
         </Button>
-      </ButtonBox>
-    </ZeroStateContainer>
+      </Flex>
+    </Flex>
   )
 }
 
 const ZeroStateMessage = styled(Sans)`
   color: ${color("black100")};
   text-align: center;
-`
-const ZeroStateContainer = styled(Flex)`
-  padding: 25px 35px 50px 35px;
-  flex-direction: column;
-`
-
-const ButtonBox = styled(Flex)`
-  margin: 0 auto;
-  padding: 15px 0 25px 0;
 `
--- a/src/lib/Components/ArtworkGrids/InfiniteScrollArtworksGrid.tsx
+++ b/src/lib/Components/ArtworkGrids/InfiniteScrollArtworksGrid.tsx
@@ -268,7 +268,7 @@ class InfiniteScrollArtworksGrid extends React.Component<Props & PrivateProps, S
           </Box>
 
           {!autoFetch && !!hasMore() && (
-            <Button mt={5} mb={3} variant="secondaryGray" size="large" block onPress={this.fetchNextPage}>
+            <Button variant="secondaryGray" size="large" block onPress={this.fetchNextPage}>
               Show more
             </Button>
           )}
--- a/src/lib/Components/ArtworkGrids/InfiniteScrollArtworksGrid.tsx
+++ b/src/lib/Components/ArtworkGrids/InfiniteScrollArtworksGrid.tsx
@@ -267,7 +267,7 @@ class InfiniteScrollArtworksGrid extends React.Component<Props & PrivateProps, S
             </View>
           </Box>
 
-          {!autoFetch && !!hasMore() && (
+          {!autoFetch && !!hasMore() && !isLoading() && (
             <Button variant="secondaryGray" size="large" block onPress={this.fetchNextPage}>
               Show more
             </Button>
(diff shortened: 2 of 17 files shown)


src/lib/Components/ArtworkGrids/FilteredArtworkGridZeroState.tsx
```diff
@@ -18,9 +18,9 @@ export const FilteredArtworkGridZeroState: React.FC<ZeroStateProps> = (props) =>
   }
 
   return (
-    <ZeroStateContainer>
+    <Flex flexDirection="column" px={4}>
```
sweir27, line 21: I find it easier to read when the styles are inline-- but more significantly, this also removes the exterior padding from the component so that call-sites can choose their own treatment.

src/lib/Components/ArtworkGrids/InfiniteScrollArtworksGrid.tsx
```diff
@@ -259,8 +259,8 @@ class InfiniteScrollArtworksGrid extends React.Component<Props & PrivateProps, S
             </View>
           </Box>
 
-          {!autoFetch && !!hasMore() && (
-            <Button mt={5} mb={3} variant="secondaryGray" size="large" block onPress={this.fetchNextPage}>
+          {!autoFetch && !!hasMore() && !isLoading() && (
```
sweir27, line 262: We don't want to show the `Show more` button if the data is loading (this gets around some weird behavior where the spinner + Show more button were displaying at the same time)

src/lib/Components/FilterModal.tsx
```diff
@@ -223,6 +224,19 @@ export const FilterOptions: React.FC<FilterOptionsProps> = (props) => {
           "color",
         ]
         break
+      case "ArtistSeries":
```
sweir27, line 227: I used the same sort order that we have for collections, given the use-case is the same.

src/lib/Scenes/ArtistSeries/ArtistSeries.tsx
```diff
@@ -31,35 +82,78 @@ export const ArtistSeries: React.FC<ArtistSeriesProps> = ({ artistSeries }) => {
...
+                        )
+                      case "artistSeriesArtworks":
+                        return (
+                          <Box px={2}>
+                            <ArtistSeriesArtworksFragmentContainer artistSeries={artistSeries} />
+                          </Box>
+                        )
+                      case "artistSeriesMoreSeries":
+                        return (
+                          !((artist?.artistSeriesConnection?.totalCount ?? 0) === 0) && (
```
sweir27, line 119: I added a little more data to this parent component (due to how relay caches / resolves its queries, it shouldn't result in more data being _fetched_) so that we can conditionally display the separator + component here.

brainbicycle (approved):
🎉

src/lib/Scenes/ArtistSeries/ArtistSeries.tsx
```diff
@@ -6,22 +6,73 @@ import { ArtistSeriesArtworksFragmentContainer } from "lib/Scenes/ArtistSeries/A
...
+      context_screen_owner_slug: artistSeries.slug,
+      action_type: Schema.ActionTypes.Tap,
+    })
+    handleFilterArtworksModal()
+  }
+
+  const closeFilterArtworksModal = () => {
+    tracking.trackEvent({
+      action_name: "closeFilterWindow",
+      context_screen_owner_type: Schema.OwnerEntityTypes.Artist,
```
williardx, line 67: Should this be `Schema.OwnerEntityTypes.ArtistSeries` instead?

src/lib/Scenes/ArtistSeries/ArtistSeries.tsx
```diff
@@ -6,22 +6,73 @@ import { ArtistSeriesArtworksFragmentContainer } from "lib/Scenes/ArtistSeries/A
...
+      context_screen: Schema.PageNames.ArtistPage,
```
williardx, line 68: Same here -- `Schema.PageNames.ArtistSeriesPage`?

src/lib/Scenes/ArtistSeries/ArtistSeries.tsx
sweir27: AH! Good catch!

src/lib/Scenes/ArtistSeries/ArtistSeries.tsx
williardx: [non-blocking] Curious to hear more about this!

src/lib/Scenes/ArtistSeries/ArtistSeriesArtworks.tsx
```diff
@@ -1,38 +1,88 @@
...
-      />
-    </Box>
-  )
+  useEffect(() => {
+    if (state.applyFilters) {
+      relay.refetchConnection(
+        PAGE_SIZE,
+        (error) => {
+          if (error) {
+            throw new Error("ArtistArtworks/ArtistArtworks filter error: " + error.message)
```
williardx, line 45: `ArtistSeries/ArtistSeriesArtworks`?

williardx:
Looks great, feel free to merge whenever you're ready!

PR #3908, opened 2020-09-23
[FX-2246] Utilizes state to track loading

dzucconi:
#trivial

This PR resolves [FX-2246]

### Description

The [`isLoading` function](https://relay.dev/docs/en/pagination-container#isloading) provided by Relay pagination containers does not trigger a render. It "is convenient for avoiding duplicate load calls" rather than transitioning the state of the UI. This PR just switches it to a simple state flag and removes what, I assume, is some code that was not actually rendering. [It *looks* like this broke in this commit.](https://github.com/artsy/eigen/commit/1090fc6251514ba157220c86fdcc0f1827eed5b2)

I did not include a spec for this because of the simplicity, but let me know if you believe it should have one and I'll add it.

-----

[FX-2246]: https://artsyproduct.atlassian.net/browse/FX-2246

### DIFF
--- a/src/lib/Components/ArtworkGrids/InfiniteScrollArtworksGrid.tsx
+++ b/src/lib/Components/ArtworkGrids/InfiniteScrollArtworksGrid.tsx
@@ -10,7 +10,6 @@ import React from "react"
 import { Dimensions, LayoutChangeEvent, ScrollView, StyleSheet, View, ViewStyle } from "react-native"
 import { createFragmentContainer, RelayPaginationProp } from "react-relay"
 
-import Spinner from "../Spinner"
 import Artwork from "./ArtworkGridItem"
 
 import { isCloseToBottom } from "lib/utils/isCloseToBottom"
@@ -84,11 +83,11 @@ interface PrivateProps {
   connection: InfiniteScrollArtworksGrid_connection
   loadMore: RelayPaginationProp["loadMore"]
   hasMore: RelayPaginationProp["hasMore"]
-  isLoading: RelayPaginationProp["isLoading"]
 }
 
 interface State {
   sectionDimension: number
+  isLoading: boolean
 }
 
 class InfiniteScrollArtworksGrid extends React.Component<Props & PrivateProps, State> {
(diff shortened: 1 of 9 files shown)


src/lib/Components/ArtworkGrids/InfiniteScrollArtworksGrid.tsx
```diff
@@ -85,6 +84,7 @@ interface PrivateProps {
 
 interface State {
   sectionDimension: number
+  isLoading: boolean
```
sweir27, line 87: Ooh good catch on this-- (and sry for breaking before! makes sense the relay prop wouldn't work for render updates 😅 )

Would you mind including a comment here to explain why we need this in addition to the relay prop?

ashfurrow: Yeah, I agree a comment would be good here. I'm a little confused why the `isLoading` in `PrivateProps` isn't/wasn't working, and if we need to keep both?

dzucconi: `isLoading` from Relay isn't a boolean value but rather a function that returns a boolean because they don't want to trigger re-renders. But yeah we can remove it, unless it's doing something clever under the hood that wouldn't be covered by just flipping the state flag back and forth.

ashfurrow: Right, right. Yeah it seems like we're handling state internally now anyway so if this is unused, I'd say that's remove it 👍

sweir27: I think we still need it to determine if we are able to fetch the next page? I guess I'd prefer to rely on relay for those decisions (and leave internal state handling to control the UI).

dzucconi: I think this is more just a case of Relay's docs being misleading. See a stale issue about this here: https://github.com/facebook/relay/issues/1973#issuecomment-325441743

This does wind up touching quite a few call sites in removing. In the future we'd probably want to refactor things using pagination container to accept `relay: RelayRefetchProp` rather than pulling off each individual sub-prop.

dzucconi: Also yeah, the problem is more complex than it seems — it's interesting: Apollo's approach is just to trigger re-renders on every state change. It's at least nice to see Relay concerned about causing all those wasted renders.

> we might create a client side field on the connection so we would only trigger a re-render when the status changes and the client is interested in the change of the status, but this seems a difficult change as it would require adding a client field extension to every *Connection type that has a @connection

sweir27: Oy. Good catch @dzucconi. I completely misunderstood those docs originally!

ashfurrow (approved):
Sorry for the delayed review – looks good to me! #3771 introduced a merge conflict with loading state. Once resolved we can #mergeongreen.

PR #3910, opened 2020-09-23
[MyCollection] Add tests

damassi:

### Description

This PR implements tests for all files in MyCollection (with the exception of the store which is coming in an immediate follow-up).

cc @artsy/csgn-devs 

#trivial

### DIFF
--- a/src/lib/Scenes/MyCollection/Screens/AddArtwork/Components/__tests__/ArrowButton-tests.tsx
+++ b/src/lib/Scenes/MyCollection/Screens/AddArtwork/Components/__tests__/ArrowButton-tests.tsx
@@ -1,0 +1,26 @@
+import { renderWithWrappers } from "lib/tests/renderWithWrappers"
+import { ArrowRightIcon } from "palette"
+import React from "react"
+import { TouchableOpacity, View } from "react-native"
+import { ArrowButton } from "../ArrowButton"
+
+describe("ArrowButton", () => {
+  it("renders correct components", () => {
+    const spy = jest.fn()
+    const Noop: React.FC = () => <View />
+    const wrapper = renderWithWrappers(
+      <ArrowButton onPress={spy}>
+        <Noop />
+      </ArrowButton>
+    )
+    expect(wrapper.root.findByType(Noop)).toBeDefined()
+    expect(wrapper.root.findByType(ArrowRightIcon)).toBeDefined()
+  })
+
+  it("calls onPress callback on click", () => {
+    const spy = jest.fn()
+    const wrapper = renderWithWrappers(<ArrowButton onPress={spy} />)
+    wrapper.root.findByType(TouchableOpacity).props.onPress()
+    expect(spy).toHaveBeenCalled()
+  })
+})
--- a/src/lib/Scenes/MyCollection/Screens/AddArtwork/Components/__tests__/ArtistSearchResult-tests.tsx
+++ b/src/lib/Scenes/MyCollection/Screens/AddArtwork/Components/__tests__/ArtistSearchResult-tests.tsx
@@ -1,0 +1,30 @@
+import OpaqueImageView from "lib/Components/OpaqueImageView/OpaqueImageView"
+import { AutosuggestResult } from "lib/Scenes/Search/AutosuggestResults"
+import { AppStore } from "lib/store/AppStore"
+import { renderWithWrappers } from "lib/tests/renderWithWrappers"
+import { Button } from "palette"
+import React from "react"
+import { ArtistSearchResult } from "../ArtistSearchResult"
+
+describe("ArtistSearchResult", () => {
+  const displayLabel = "some display label"
+  const result = {
+    imageUrl: "",
+    displayLabel,
+  }
+
+  it("renders correct components", () => {
+    const wrapper = renderWithWrappers(<ArtistSearchResult result={result as AutosuggestResult} />)
+    expect(wrapper.root.findByType(OpaqueImageView)).toBeDefined()
+    expect(wrapper.root.findByType(Button)).toBeDefined()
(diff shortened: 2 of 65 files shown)


src/lib/tests/renderWithWrappers.tsx
```diff
@@ -31,7 +31,7 @@ export const renderWithWrappers = (component: ReactElement) => {
           error
       )
     } else {
-      throw new Error(error.message)
+      throw new Error(error.stack)
```
damassi, line 34: Better error traces

src/lib/Scenes/MyCollection/Screens/AddArtwork/Components/__tests__/ArrowButton-tests.tsx
```diff
@@ -0,0 +1,26 @@
...
+    const wrapper = renderWithWrappers(
+      <ArrowButton onPress={spy}>
+        <Noop />
+      </ArrowButton>
+    )
+    expect(wrapper.root.findByType(Noop)).toBeDefined()
+    expect(wrapper.root.findByType(ArrowRightIcon)).toBeDefined()
+  })
+
+  it("calls onPress callback on click", () => {
```
ashleyjelks, line 20: Is this testing anything relevant other than prop drilling?

src/lib/Scenes/MyCollection/Screens/AddArtwork/Components/__tests__/ArtistSearchResult-tests.tsx
```diff
@@ -0,0 +1,30 @@
...
+    expect(wrapper.root.findByType(Button)).toBeDefined()
+    expect(wrapper.root.findByProps({ "data-test-id": "displayLabel" }).props.children).toEqual("some display label")
+  })
+
+  it("sets the search results to null on remove button click", () => {
+    const spy = jest.fn()
+    AppStore.actions.myCollection.artwork.setArtistSearchResult = spy as any
+    const wrapper = renderWithWrappers(<ArtistSearchResult result={result as AutosuggestResult} />)
+    wrapper.root.findByType(Button).props.onPress()
+    expect(spy).toHaveBeenCalled()
```
ashleyjelks, line 28: Call this with .`toHaveBeenCalledWith(null)` ?

src/lib/Scenes/MyCollection/Screens/AddArtwork/Components/__tests__/MediumPicker-tests.tsx
```diff
@@ -0,0 +1,26 @@
...
+  beforeEach(() => {
+    useFormikContextMock.mockImplementation(() => ({
+      handleChange: jest.fn(),
+      values: {
+        medium: "Painting",
+      },
+    }))
+  })
+
+  it("displays the correct medium", () => {
```
ashleyjelks, line 21: When we add more form validation, this test might be more relevant, as now its just testing data input maps to UI output.

src/lib/Scenes/MyCollection/utils/__tests__/formatArtworkMedium-tests.tsx
```diff
@@ -0,0 +1,8 @@
+import { formatMedium } from "../formatArtworkMedium"
+
+describe("formatArtworkMedium", () => {
+  it("formats medium correctly", () => {
```
ashleyjelks, line 4: Can you test how it behaves when invalid props are received vs valid ones?

src/lib/Scenes/MyCollection/utils/__tests__/formatCentsToDollars-tests.tsx
```diff
@@ -0,0 +1,7 @@
+import { formatCentsToDollars } from "../formatCentsToDollars"
+
+describe("formatCentsToDollars", () => {
+  it("works", () => {
+    expect(formatCentsToDollars(200)).toBe("$2")
```
ashleyjelks, line 5: Can you test invalid input as well e.g. negative ints, very large ints, floats, etc?

src/lib/Scenes/MyCollection/Screens/AddArtwork/Components/__tests__/ArtistSearchResult-tests.tsx
damassi: If it were called with an actual `null` argument then that would be the proper test, but its just a regular old bare function invocation.

src/lib/Scenes/MyCollection/Screens/AddArtwork/Components/__tests__/ArrowButton-tests.tsx
damassi: The idea is to test that all behavior is correctly wired up -- if I pass a prop `onPress` in, I should assert that when the main button in the component is pressed that its invoked.

ashleyjelks:
@jonallured and I went through most of the PR and added some general comments/feedback. It seems like the most important elements to add test coverage for are on the state and actions being manage via Formik (and I believe Yup as well). I noticed those are still marked TODO, so I think we can circle back on them.

damassi:
>  state and actions

Yup! That's coming in a follow up PR with the store tests.

src/lib/Scenes/MyCollection/utils/__tests__/formatArtworkMedium-tests.tsx
damassi: This will take any string value thats passed through and spit it back out capitalize, or match against the old `CAPS_STRINGS` that were there before -- if a number or array or whatever is passed through the typechecker would fail in a couple different ways, so we're covered there. This just tests behavior -- that we get a capitalized string back, and that it correctly looks up values from the defaults.

PR #3914, opened 2020-09-24
[FX-2210] Implements fair editorial display

dzucconi:

This PR resolves [FX-2210]

### Description

User can see up to 5 articles associated with the fair

[FX-2210]: https://artsyproduct.atlassian.net/browse/FX-2210

### DIFF
--- a/src/lib/Scenes/Fair2/Components/Fair2Editorial.tsx
+++ b/src/lib/Scenes/Fair2/Components/Fair2Editorial.tsx
@@ -1,0 +1,78 @@
+import { Fair2Editorial_fair } from "__generated__/Fair2Editorial_fair.graphql"
+import OpaqueImageView from "lib/Components/OpaqueImageView/OpaqueImageView"
+import SwitchBoard from "lib/NativeModules/SwitchBoard"
+import { Box, BoxProps, color, Text, Touchable } from "palette"
+import React, { useRef } from "react"
+import { createFragmentContainer, graphql } from "react-relay"
+
+interface Fair2EditorialProps extends BoxProps {
+  fair: Fair2Editorial_fair
+}
+
+export const Fair2Editorial: React.FC<Fair2EditorialProps> = ({ fair, ...rest }) => {
+  const ref = useRef<any>()
+
+  if (!fair.articles?.edges || fair.articles.edges.length === 0) {
+    return null
+  }
+
+  return (
+    <Box ref={ref} {...rest}>
+      <Text mx={2} mb={2} variant="subtitle">
+        Related articles
+      </Text>
+
+      {(fair.articles.edges || []).map((edge) => {
+        const article = edge!.node!
+
+        return (
+          <Touchable
+            key={article.id}
+            underlayColor={color("black5")}
+            onPress={() => {
+              if (!article.href) {
+                return
+              }
+
+              SwitchBoard.presentNavigationViewController(ref.current, article.href)
+            }}
+          >
+            <Box flexDirection="row" py={1} px={2}>
+              <Box flex={1} pr={2}>
+                <Text variant="subtitle">{article.title}</Text>
+
+                <Text variant="text" color="black60">
+                  {article.publishedAt}
+                </Text>
+              </Box>
+
+              {!!article.thumbnailImage?.src && (
+                <OpaqueImageView width={90} height={50} imageURL={article.thumbnailImage.src} />
+              )}
+            </Box>
+          </Touchable>
+        )
+      })}
+    </Box>
+  )
+}
+
+export const Fair2EditorialFragmentContainer = createFragmentContainer(Fair2Editorial, {
+  fair: graphql`
+    fragment Fair2Editorial_fair on Fair {
+      articles: articlesConnection(first: 5, sort: PUBLISHED_AT_DESC) {
+        edges {
+          node {
+            id
+            title
+            href
+            publishedAt(format: "MMM Do, YY")
+            thumbnailImage {
+              src: imageURL
+            }
+          }
+        }
+      }
+    }
+  `,
+})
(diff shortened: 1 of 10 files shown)


src/lib/Scenes/Fair2/Components/Fair2Editorial.tsx
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  return (
+    <Box ref={ref} {...rest}>
+      <Text mx={2} mb={2} variant="subtitle">
+        Related articles
+      </Text>
+
+      {(fair.articles.edges || []).map((edge) => {
+        if (!edge || !edge.node) {
```
sweir27, line 26: `!edge?.node`?

src/lib/Scenes/Fair2/Components/Fair2Editorial.tsx
```diff
@@ -0,0 +1,84 @@
...
+              <Box flex={1} pr={2}>
+                <Text variant="subtitle">{article.title}</Text>
+
+                <Text variant="text" color="black60">
+                  {article.publishedAt}
+                </Text>
+              </Box>
+
+              {!!article.thumbnailImage?.img && (
+                <OpaqueImageView width={90} height={50} imageURL={article.thumbnailImage.img.src} />
```
sweir27, line 54: I _think_ this `OpaqueImageView` will also try to create a resized URL (and I think you're passing one in here?). Is it possible that's causing the image issues?

src/lib/Scenes/Fair2/Fair2.tsx
```diff
@@ -27,6 +28,10 @@ export const Fair2: React.FC<Fair2Props> = ({ fair }) => {
           switch (item) {
             case "fair2Header":
               return <Fair2HeaderFragmentContainer fair={fair} />
+            case "fair2Editorial":
+              return (
+                <Fair2EditorialFragmentContainer fair={fair} my={3} pt={3} borderTopWidth={1} borderColor="black10" />
```
sweir27, line 33: Should we use a `<Separator` here instead of border?

They achieve the same thing, just curious if you had a reason to use a border instead.

sweir27 (commented):
Some minor questions about TS syntax and a question about image URLing.

src/lib/Scenes/Fair2/Components/Fair2Editorial.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { createFragmentContainer, graphql } from "react-relay"
+
+interface Fair2EditorialProps extends BoxProps {
+  fair: Fair2Editorial_fair
+}
+
+export const Fair2Editorial: React.FC<Fair2EditorialProps> = ({ fair, ...rest }) => {
+  const ref = useRef<any>()
+
+  if (!fair.articles || !fair.articles.edges || fair.articles.edges.length === 0) {
```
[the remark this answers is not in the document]
dzucconi: I did sort of mean this explicitly in this case: each of those fields are nullable (okay can use the ? for one of them) then it might be an empty array.

src/lib/Scenes/Fair2/Components/Fair2Editorial.tsx
dzucconi: I truly wish our types were correct.

src/lib/Scenes/Fair2/Components/Fair2Editorial.tsx
dzucconi: Ah, almost certainly. Will address.

src/lib/Scenes/Fair2/Fair2.tsx
dzucconi: I'd probably use `Separator` in web-land if it was an actual `<hr />`. Here it doesn't matter much to me either way.

src/lib/Scenes/Fair2/Components/Fair2Editorial.tsx
dzucconi: `edge!.node!` and remove the conditional is probably better — this will never be null so this conditional is just here to satisfy the incorrect typing.

src/lib/Scenes/Fair2/Fair2.tsx
dzucconi: Ah, I see with `FlatList` — addressed.

src/lib/Scenes/Fair2/Components/Fair2Editorial.tsx
dzucconi: 👍

src/lib/Scenes/Fair2/Fair2.tsx
```diff
@@ -17,16 +18,19 @@ interface Fair2Props {
 }
 
 export const Fair2: React.FC<Fair2Props> = ({ fair }) => {
-  const sections = ["fair2Header"]
+  const sections = ["fair2Header", "fair2Editorial"]
```
dzucconi, line 21: This pattern feels very awkward to me

sweir27: Given the `Separator` appears between each section, it's more standard to conditionally add these sections to the list (so like, verify that there are articles at all before pushing it onto this array)

dzucconi: That's true — though I would also hope that the `FlatList` implementation compacts the return values?

dzucconi: But I more mean this piece of indirection of assigning things arbitrary keys then using a switch statement. Why not just an array of components?

sweir27: It makes sense in the case where sections could be repeated (not sure if there's another use-case).

I'm also not sure about compacting the return values, but should be verifiable 😬 .

dzucconi: Ignoring the null bit for a moment:

```tsx
<FlatList
        data={[<Fair2HeaderFragmentContainer fair={fair} />, <Fair2EditorialFragmentContainer fair={fair} />]}
        ItemSeparatorComponent={() => <Separator my={3} />}
        renderItem={({ item }) => item} // Should be optional/default
      />
```

williardx (approved):
LGTM!

sweir27:
#mergeongreen

PR #3918, opened 2020-09-24
[MyCollection] Fix invalid payload

damassi:

### Description

This fixes an issue where we were sending falsy values to MP, leading to failures in create / edit, and also automates a lot of the manual update [documented here](https://github.com/artsy/eigen/pull/3896#discussion_r493188885).

#trivial

### DIFF
--- a/src/lib/Scenes/MyCollection/utils/cleanArtworkPayload.ts
+++ b/src/lib/Scenes/MyCollection/utils/cleanArtworkPayload.ts
@@ -1,6 +1,8 @@
+import { isNil } from "lodash"
+
 export function cleanArtworkPayload(payload = {}): any {
   const cleanPayload = Object.entries(payload).reduce((acc, [key, value]) => {
-    if (!!value) {
+    if (value !== "" && !isNil(value)) {
       return { ...acc, [key]: value }
     }
     return acc
--- a/src/lib/Scenes/MyCollection/utils/cleanArtworkPayload.ts
+++ b/src/lib/Scenes/MyCollection/utils/cleanArtworkPayload.ts
@@ -1,0 +1,9 @@
+export function cleanArtworkPayload(payload = {}): any {
+  const cleanPayload = Object.entries(payload).reduce((acc, [key, value]) => {
+    if (!!value) {
+      return { ...acc, [key]: value }
+    }
+    return acc
+  }, {})
+  return cleanPayload
+}
--- a/src/__generated__/MyCollectionArtworkModelCreateArtworkMutation.graphql.ts
+++ b/src/__generated__/MyCollectionArtworkModelCreateArtworkMutation.graphql.ts
@@ -1,6 +1,6 @@
 /* tslint:disable */
 /* eslint-disable */
-/* @relayHash 1079ed6888958e98aeacb10b229d7f2d */
+/* @relayHash 87110181faf533b605a71e1b0f96dbe0 */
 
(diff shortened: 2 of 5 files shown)


src/lib/Scenes/MyCollection/utils/cleanArtworkPayload.ts
```diff
@@ -0,0 +1,9 @@
+export function cleanArtworkPayload(payload = {}): any {
```
jonallured, line 1: **issue [blocking]:** clobbering false values (and missing spec file)

I'd love to see a spec that demonstrates what this function does, maybe something like this:

```ts
describe("cleanArtworkPayload", () => {
  describe("with a payload including a key with a null value", () => {
    it("removes that key from the cleaned payload", () => {
      const initialPayload = { attribute: null }
      const cleanedPayload = cleanArtworkPayload(initialPayload)
      expect(cleanedPayload).not.toHaveProperty('attribute')
    })
  })

  describe("with a payload including a key with a false value", () => {
    it("passes that value through to the cleaned payload", () => {
      const initialPayload = { attribute: false }
      const cleanedPayload = cleanArtworkPayload(initialPayload)
      expect(cleanedPayload).toHaveProperty('attribute', false)
    })
  })

  describe("with a payload including a key with a true value", () => {
    it("passes that value through to the cleaned payload", () => {
      const initialPayload = { attribute: true }
      const cleanedPayload = cleanArtworkPayload(initialPayload)
      expect(cleanedPayload).toHaveProperty('attribute', true)
    })
  })

  // etc...
})
```

And I think that's worth doing because I believe this implementation will clobber a false value, no??

*format via [Conventional Comments](https://conventionalcomments.org/)*

jonallured (changes requested):
@damassi I think there's a bug here with the cleaner function. I think we're at the point with this project where we should be more strict about testing so I have written up what I believe would be a good spec to cover the behavior we're wanting here but LMK if I missing something!

src/lib/Scenes/MyCollection/utils/cleanArtworkPayload.ts
damassi: @jonallured i'm adding tests all at once in here: https://github.com/artsy/eigen/pull/3910. Can we merge and i'll cover this comment over there?

src/lib/Scenes/MyCollection/utils/cleanArtworkPayload.ts
```diff
@@ -0,0 +1,11 @@
+import { isNil } from "lodash"
+
+export function cleanArtworkPayload(payload = {}): any {
+  const cleanPayload = Object.entries(payload).reduce((acc, [key, value]) => {
+    if (value !== "" && !isNil(value)) {
```
damassi, line 5: `isNil` covers undefined and null

src/lib/Scenes/MyCollection/utils/cleanArtworkPayload.ts
jonallured: I'm not a fan of this approach - I don't see a good reason to skip the unit test here. 🤷

damassi: Updated!

jonallured (approved):
Thanks for the spec file!! ❤️ ❤️ ❤️ 

#mergeongreen

damassi:
Haha sorry for being ridiculous 😄

PR #3920, opened 2020-09-25
Make mocked relay error more obvious

damassi:

### Description

When writing a relay test today I found myself hopelessly lost as to why the test renderer wasn't rendering the `<QueryRenderer>`. It turns out I forgot to add `jest.unmock('react-relay')` at the top of my test.

This updates our `renderWithWrappers` function to check for this condition, and to alert the user of the possibility. 

Screen Shot 2020-09-24 at 5 47 46 PM

### DIFF
--- a/src/lib/tests/renderWithWrappers.tsx
+++ b/src/lib/tests/renderWithWrappers.tsx
@@ -11,16 +11,29 @@ import { ReactElement } from "simple-markdown"
  */
 export const renderWithWrappers = (component: ReactElement) => {
   const wrappedComponent = componentWithWrappers(component)
-  // tslint:disable-next-line:use-wrapped-components
-  const renderedComponent = ReactTestRenderer.create(wrappedComponent)
+  try {
+    // tslint:disable-next-line:use-wrapped-components
+    const renderedComponent = ReactTestRenderer.create(wrappedComponent)
 
-  // monkey patch update method to wrap components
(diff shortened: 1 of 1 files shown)


src/lib/utils/__tests__/placeholders-tests.tsx
```diff
@@ -6,7 +6,7 @@ describe(PlaceholderBox, () => {
   it(`requires a placeholder context`, async () => {
     expect(() => {
       renderWithWrappers(<PlaceholderBox width={400} />)
-    }).toThrowErrorMatchingInlineSnapshot(`"You're using a Placeholder outside of a PlaceholderContext"`)
+    }).toThrowErrorMatchingInlineSnapshot(`"Error: You're using a Placeholder outside of a PlaceholderContext"`)
```
ds300, line 9: Any idea why this has changed?

ds300 (approved):
Awesome, thank you 🙏 

There's an odd unrelated snapshot change here, let's make sure it's not going to cause churn before merging.

src/lib/utils/__tests__/placeholders-tests.tsx
damassi: Will investigate.

damassi: difference between `Error(error)` and `Error(error.message)`

PR #3934, opened 2020-09-28
FX-2213, FX-2214, FX-2215: Adds the Artworks and Exhibitors content to fair page

sweir27:

This PR resolves [FX-2213], [FX-2214], [FX-2215]

### Description

This PR adds the bottom-of-the-page content to the new fair page, which includes the Exhibitors and Artworks shelves/grid.

Some notes:
- To get the sticky header effect, I used the built-in `stickyHeaderIndices` prop on `FlatList`. It seems to work fine, the only weird thing is that when you click into the "Artworks" tab, the screen jumps. I'm pretty sure this is due to how we lazily-load that artworks grid (as the same does not occur when switching to the "Exhibitors" tab).
- The rest is pretty standard and builds off the logic we have on web.
- In order to validate that the basic layout works as-intended, I also included the animated Filter Button in this PR. Will follow up with the actual filtering behavior.

Screen Shot 2020-09-28 at 11 16 20 AM
Screen Shot 2020-09-28 at 11 16 26 AM
fairtabs

[FX-2213]: https://artsyproduct.atlassian.net/browse/FX-2213
[FX-2214]: https://artsyproduct.atlassian.net/browse/FX-2214
[FX-2215]: https://artsyproduct.atlassian.net/browse/FX-2215

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -13,6 +13,7 @@ upcoming:
     - Add collections rail for Fair2 (visible only to admins) - damon
     - Add More Info view for Fiar2 (visible only to admins) - will
     - Add More Info view for Fair2 (visible only to admins) - will
+    - Add Artworks/Exhibitors rails to Fair2 - sweir
   user_facing:
     - Fix navigation bug on Artwork consign button - david
     - Add auction lots rail to the auctions screen - mounir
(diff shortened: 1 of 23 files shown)


pvinis:
Good job with the sticky tabs, looks sweeeeet.

About the scrollOffset jump on tab switch, yea, probably it's the lazy loading. I wonder if there is a quick easy way to have something like the placeholders we have with the grey animated boxes. We might need to incorporate this in the InfiniteGrid component. If it's out-of-scope for this PR, we can make a ticket for this, and then just having a few boxes should make it not jump.

williardx:
LGTM! Nice job figuring out the sticky header 💪

sweir27:
Going to merge this in so we can keep building on it, but yeah @pvinis we should have a ticket for the skeleton loading `InfiniteGrid`!

pvinis:
Made one here. https://artsyproduct.atlassian.net/browse/MX-556

PR #3935, opened 2020-09-28
[MX-549] - align internal webview to parent vc, add dedicated routes for faqs

brainbicycle:

This PR resolves [MX-549]

### Description

These routes were actually loading correctly but there was a layout issue with the internal web view component, not really clear how it broke.

- Aligns internal web view native component to parent vc
- Adds dedicated routes for identity verification FAQ and auction FAQ to avoid profile api call in VanityURLEntity

#### Screenshots

auction-faqs

id-verify-faqs

#### Follow-ups

- The back button is covering the Auction FAQ header in the webview, seems like this is an existing bug in older builds

[MX-549]: https://artsyproduct.atlassian.net/browse/MX-549


ashfurrow (approved):
Sounds good to me 👍 I suspect this might have been broken by the routing changes and how VCs are displayed to the user.

PR #3936, opened 2020-09-28
[MX-553] fix: Bottom buttons in modals are obscured by keyboard

MounirDhahri:

This PR resolves [MX-553]

### Description
The Bottom buttons in modals are obscured by the keyboard.

**Before:**
Group 2

___
**After**
Screenshot 2020-09-28 at 19 06 27
Screenshot 2020-09-28 at 19 14 46

___

[MX-553]: https://artsyproduct.atlassian.net/browse/MX-553

### DIFF
--- a/src/lib/Components/Buttons/BottomAlignedButtonWrapper.tsx
+++ b/src/lib/Components/Buttons/BottomAlignedButtonWrapper.tsx
@@ -1,6 +1,6 @@
+import { ICON_HEIGHT } from "lib/Scenes/BottomTabs/BottomTabsIcon"
 import React from "react"
 import { Dimensions, KeyboardAvoidingView, View } from "react-native"
-import { CARD_STACK_OVERLAY_HEIGHT } from "../FancyModal/FancyModalCard"
 
 export interface BottomAlignedProps extends React.Props<JSX.Element> {
   onPress: () => void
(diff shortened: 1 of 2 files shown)


src/lib/Components/Buttons/BottomAlignedButtonWrapper.tsx
```diff
@@ -14,7 +15,7 @@ const defaultVerticalOffset = isPhoneX ? 30 : 15
 const BottomAlignedButtonWrapper: React.FC<BottomAlignedProps> = (props) => (
   <KeyboardAvoidingView
     behavior="padding"
-    keyboardVerticalOffset={props.verticalOffset || defaultVerticalOffset}
+    keyboardVerticalOffset={props.verticalOffset || defaultVerticalOffset + CARD_STACK_OVERLAY_HEIGHT}
```
MounirDhahri, line 18: I had to add the `CARD_STACK_OVERLAY_HEIGHT` since we are opening the screen inside a fancy modal

brainbicycle: Can we instead pass in at callsite as part of verticalOffset prop? It seems this will break if we use this button outside a fancy modal otherwise?

src/lib/Scenes/Consignments/Components/BottomAlignedButton.tsx
```diff
@@ -10,10 +10,13 @@ export interface BottomAlignedProps extends React.Props<JSX.Element> {
   verticalOffset?: number
 }
 
+const isPhoneX = Dimensions.get("window").height === 812 && Dimensions.get("window").width === 375
+const defaultVerticalOffset = isPhoneX ? 34 : 20
```
brainbicycle, line 14: If at all possible think we should avoid having device specific checks like this because they can be fairly brittle, is safeAreaInsets.top not returning correct values?

MounirDhahri: you're right, I think I found out why they broke

MounirDhahri:
I was investigating the issue and it turned out that the buttons broke after the bottom tabs migration to typescript. I think adding the bottom tab height (`ICON_HEIGHT`) is enough to fix it

brainbicycle (approved):
👍 Nice find!

PR #3949, opened 2020-09-30
FX-2216: Allows filtering on new fair page by defaults + Artists I Follow

sweir27:

This PR resolves [FX-2216]

### Description

This PR adds the ability filter a fair by "Artists I Follow". For the most part, I was able to follow existing patterns but refactored slightly to make it easier to include the new aggregation/filter.

Adding the default filters at this point is trivial boilerplate (the filter button was included in a separate PR).

followfilter

[FX-2216]: https://artsyproduct.atlassian.net/browse/FX-2216

### DIFF
--- a/src/lib/Components/ArtworkFilterOptions/ArtistsIFollowOptions.tsx
+++ b/src/lib/Components/ArtworkFilterOptions/ArtistsIFollowOptions.tsx
@@ -1,0 +1,37 @@
+import { FilterDisplayName, FilterParamName } from "lib/Scenes/Collection/Helpers/FilterArtworksHelpers"
+import { ArtworkFilterContext, FilterData, useSelectedOptionsDisplay } from "lib/utils/ArtworkFiltersStore"
+import React, { useContext } from "react"
+import { NavigatorIOS } from "react-native"
+import { MultiSelectOptionScreen } from "./MultiSelectOption"
+
+interface ArtistsIFollowOptionsScreenProps {
+  navigator: NavigatorIOS
+}
+
+export const ArtistsIFollowOptionsScreen: React.FC<ArtistsIFollowOptionsScreenProps> = ({ navigator }) => {
+  const { dispatch } = useContext(ArtworkFilterContext)
+  const selectedOptions = useSelectedOptionsDisplay()
+
+  const artistsIFollowFilterNames = [FilterParamName.artistsIFollow]
+  const artistsIFollowOptions = selectedOptions.filter((value) => artistsIFollowFilterNames.includes(value.paramName))
+
+  const selectOption = (option: FilterData, updatedValue: boolean) => {
+    dispatch({
+      type: "selectFilters",
+      payload: {
+        displayText: option.displayText,
+        paramValue: updatedValue,
+        paramName: option.paramName,
+      },
+    })
+  }
+
+  return (
+    <MultiSelectOptionScreen
+      onSelect={selectOption}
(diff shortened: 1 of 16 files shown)


src/lib/Components/ArtworkFilterOptions/ArtistsIFollowOptions.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { ArtworkFilterContext, FilterData, useSelectedOptionsDisplay } from "lib/utils/ArtworkFiltersStore"
+import React, { useContext } from "react"
+import { NavigatorIOS } from "react-native"
+import { MultiSelectOptionScreen } from "./MultiSelectOption"
+
+interface ArtistsIFollowOptionsScreenProps {
+  navigator: NavigatorIOS
+}
+
+export const ArtistsIFollowOptionsScreen: React.FC<ArtistsIFollowOptionsScreenProps> = ({ navigator }) => {
```
sweir27, line 11: Each of the individual filter options (sort, medium, etc.) have their own screen like this. Right now it just has a single "Artists I follow" toggle, but in the future we'll add a full list of artists.

src/lib/Scenes/Collection/Helpers/FilterArtworksHelpers.ts
```diff
@@ -96,3 +100,63 @@ export const changedFiltersParams = (currentFilterParams: FilterParams, selected
 
   return changedFilters
 }
+
+export const selectedOption = (selectedOptions: FilterArray, filterType: FilterScreen) => {
```
sweir27, line 104: I extracted this because the logic was getting more complex and isolating makes it easier to reason about/test.

This function returns the "selected display" text on the home screen of the filter modal.

src/lib/Scenes/Collection/Helpers/FilterArtworksHelpers.ts
```diff
@@ -96,3 +100,63 @@ export const changedFiltersParams = (currentFilterParams: FilterParams, selected
...
+}
+
+export type aggregationsType =
+  | ReadonlyArray<{
+      slice: string
+      counts: Array<{ count: number; value: string; name?: string }>
+    }>
+  | []
+
+export const aggregationsWithFollowedArtists = (
```
sweir27, line 148: I also wanted to test this in isolation, and we can use it when we decide to add "Artists I follow" as a filter to other screens.

I'm not sure the original reasoning, but the format of the aggregation result (in MP) differs between `FOLLOWED_ARTISTS` and the rest of the aggregations, so we need some special logic to coerce it into the format that the filter modal expects.

src/lib/Scenes/Fair2/Fair2.tsx
```diff
@@ -109,15 +131,39 @@ export const Fair2: React.FC<Fair2Props> = ({ fair }) => {
                       return <Tabs setActiveTab={setActiveTab} activeTab={activeTab} tabs={tabs} />
                     }
                     case "fairTabChildContent": {
-                      return <TabChildContent activeTab={activeTab} tabs={tabs} />
```
sweir27, line 112: We could improve this interface in the future but since I'm not sure what the future holds for tab treatments in the app, I made the logic simpler for now.

src/lib/Scenes/Collection/Helpers/FilterArtworksHelpers.ts
dzucconi: 🙏

PR #3954, opened 2020-09-30
RFC-ish: reworking LoadingModal

pvinis:

This PR resolves [CX-681]

### Description

This came up because of https://artsy.slack.com/archives/CN8S32FJR/p1601474431112100.
The dismissal of the edit screen (for example phone edit) was not working.

I figured out that the reason for it is that here https://github.com/artsy/eigen/blob/master/src/lib/Scenes/MyAccount/Components/MyAccountFieldEditScreen.tsx#L34-L41 we set `saving` to true, then we do the mutation, then `dismiss` runs. When that happens, the `LoadingModal` is the topmost view, so the dismissal goes to that, and nothing happens 🤔.

I had two ideas, one was to make sure `onDismiss` (a few lines above) turns `saving` to false, waits for that to happen so the LoadingModal goes away, and then run the Switchboard dismiss. I didn't go for that because it feels kinda fragile?

The other idea was to make the LoadingModal **not** a modal. I turned it into a Provider and a hook, so anywhere in the app we can just do `const setLoadingBlock = useLoadingBlock()` and then we can set it to true and the LoadingBlockView will appear on top of everything. It works great except for this 👇.
https://f.p0l.co/file/dropshare-public-pavlos/Screen-Recording-2020-09-30-at-18.36.01.mov
The tabs right now are on a separate native view, so the blocking view is not going over that. I tried tapping another tab while the blocking view was showing, and things work but the dismissal doesnt happen if we are looking at another tab. Kind of an edge case but still I'm not in love with that behavior.

Ideas? How should we handle this bug?

[CX-681]: https://artsyproduct.atlassian.net/browse/CX-681


damassi:
👍 We need something like this for MyCollections stuff as well!

src/lib/Components/Modals/LoadingModal.tsx
```diff
@@ -1,17 +1,42 @@
...
-      <ActivityIndicator size="large" />
-    </Flex>
-  </Modal>
+const LoadingBlockContext = React.createContext<{ loadingBlock: boolean; setLoadingBlock: (val: boolean) => void }>(
+  null!
 )
 
-export default LoadingModal
+export const LoadingBlockProvider: React.FC = ({ children }) => {
+  const [loadingBlock, setLoadingBlock] = useState(false)
```
damassi, line 29: We should be using our app_store for this, no? And then we can use hooks to tap into things and not need to recreate a lot of stateful infra.

src/lib/Scenes/MyAccount/Components/MyAccountFieldEditScreen.tsx
```diff
@@ -19,25 +19,26 @@ export const MyAccountFieldEditScreen = React.forwardRef<
     onSave(dismiss: () => void): Promise<any>
   }>
 >(({ children, canSave, onSave, title, contentContainerStyle }, ref) => {
-  const [isSaving, setIsSaving] = useState<boolean>(false)
+  const setLoadingBlock = useLoadingBlock()
```
damassi, line 22:
```tsx
AppStore.actions.showLoadingIndicator()
```
and 
```tsx
AppStore.actions.showLoadingIndicator()

**Comment by pvinis** [comment]
totally! good suggestion. It didn't come to my mind.

**Comment by damassi** [comment]
One thing I ran into is an iOS limitation around displaying two modals at once -- given this example: the FancyModal appears with a CTA that triggers a mutation; on click, we can't display the `LoadingModal` because the `FancyModal` is already visible. But if we dismiss the `FancyModal` first and delay the `LoadingModal`, we risk running into a race condition between the mutation request succeeding before the timeout has executed (and so on). 

A good way around this would be to extend the layered modal code that @ds300 used for nested `FancyModals` but for any RN modal that we may want to use. (Not saying we use the animation, but the layering seems to work nicely.)

File: src/lib/AppRegistry.tsx [hunk 3]
```diff
@@ -304,11 +305,13 @@ class PageWrapper extends React.Component<PageWrapperProps> {
       <ProvideScreenDimensions>
         <RelayEnvironmentProvider environment={defaultEnvironment}>
           <AppStoreProvider>
-            <Theme>
-              <_FancyModalPageWrapper>
-                <InnerPageWrapper {...this.props} />
-              </_FancyModalPageWrapper>
-            </Theme>
+            <LoadingBlockProvider>
```
damassi, line 308: Since this isn't a react context the naming seems slightly misleading, but that said it _is_ providing something -- the modal.

src/lib/Components/Modals/LoadingBlock.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+export const LoadingBlockModel: LoadingBlockModel = {
+  sessionState: {
+    isShowing: false,
+  },
+
+  setIsShowing: action((state, input) => {
+    state.sessionState.isShowing = input
+  }),
+}
```
damassi, line 59: This will grow nicely with time 👍

src/lib/AppRegistry.tsx
pvinis: yea I have some questions and concerns for this pr. I'll check things with David next week 🤔

pvinis:
I'll close this. If we decide this needs to happen, we can rework it in a ticket.

PR #3957, opened 2020-10-01
chore: replace Echo with an S3 bucket

ashfurrow:

This PR resolves [MX-528]

### Description

Paired on this in KS.

Remaining work here:

- [x] Update the URL from a raw S3 bucket to a CNAME from Cloudflare once the Echo work is done.
- [x] ⚠️ Remove `ARTSY_ECHO_PRODUCTION_TOKEN` from our shared env in s3, or keep it and use it for s3. ⚠️
  In one commit in this PR I removed it and replaced it with `_OSS_` like the suggestion from https://artsyproduct.atlassian.net/browse/CX-531. If we decide to keep it and use it with s3, just revert that commit and let's adjust s3.

[MX-528]: https://artsyproduct.atlassian.net/browse/MX-528


pvinis:
This is almost ready. First we should decide what to do about the echo key, if we should add it back on s3, or if we keep it public and not store sensitive things there. Otherwise the code is ready.

brainbicycle (approved):
🎉

ds300 (approved):
🎉

Podfile
```diff
@@ -169,8 +169,7 @@ post_install do |installer|
 
   # Note: we don't want EchoNew.json checked in, so Artsy staff download it at pod install time. We
   # use a stubbed copy for OSS developers.
-  echo_key = `dotenv -f ".env.shared,.env.ci" env | grep ARTSY_ECHO_PRODUCTION_TOKEN | awk -F "=" {'print $2'}`
-  if echo_key.length > 1 # OSS contributors have "-" as their key
+  if !ENV['_OSS_']
```
ashfurrow, line 172: We should remove this check and have everyone (Artsy staff and OSS people) run `make update_echo` instead. So that would mean:

- Removing this `if` check so everyone runs `make update_echo`.
- Removing `Echo.json.example` from the repo.
- Removing `cp Artsy/App/Echo.json.example Artsy/App/EchoNew.json` from the `oss` script in the Makefile.

ashfurrow (commented):
Looks good! Just one other thing and then we can merge. Happy to take on that work, just let me know.

Podfile
pvinis: Yup, makes sense. Feel free to add that on if ou have time, otherwise I'll take a look tomorrow.

pvinis:
I addressed the suggestions. Removed the `if`, removed `Echo.json.example`, and removed the `cp` command, so now both `make oss` and `make artsy` will not touch echo.json, and the pod install will handle the fetching and running of `make update_echo`.

One extra change I made is that the `make update_echo` now formats the file. The difference in file size is so small that there is basically no difference for the app. The benefit of it is that in our editors (vscode, xcode) we can read that json file easily, and search/add things while testing. Also the moment this is replaced in the app, the minified version will be the new locally stored file, but at that point we don't really look at that file as devs, so all good. In the next native release of the app, the prettified file will be embedded again, but once again, the size increase is basically nothing. If we still want, we can add a build phase to minify, but honestly I don't think we need it for that size of file.

Podfile
pvinis: done

PR #3959, opened 2020-10-01
CSGN-356: Upload photos to S3 and pass off urls to MP

jonallured:
Note: This PR relies on artsy/metaphysics#2742 and the update to the schema that it adds.

The purpose of this PR is to begin supporting the uploading of My Collection artwork images. I worked down the stack to support this behavior so it is now possible. For the UX in the app I did the dumbest thing possible: block sending the mutation until all the photos have been uploaded and don't start uploading until the "Create" button is pressed.

This is almost certainly NOT what we'll end up shipping to end users, but it's a start! And it moves us from zero support for images to one that is naive but possible. I hope to demonstrate how this works today and then get feedback and improve things.

https://artsyproduct.atlassian.net/browse/CSGN-356


pepopowitz (approved):
💪 💪 💪

damassi:
LGTM 👍

PR #3961, opened 2020-10-02
Fixes Changelog parse error #trivial

ashfurrow:
The type of this PR is: **bugfix**

### Description

This was failing to parse in our `Fastfile` with the following error:

```
Psych::SyntaxError: (<unknown>): found character that cannot start any token while scanning for the next token at line 19 column 7
```

This caused the beta deploy to fail.

/cc @sepans 


sepans:
Thanks for the fix 🙏. Sorry about that

ashfurrow:
No worries at all! We actually have tooling to check this on PRs:

https://github.com/artsy/eigen/blob/320bf7c212723eba7140747bb389bb06661efce5/dangerfile.ts#L100-L101

 But this error is specific to the Ruby library we use in the Fastfile 🤷

PR #3964, opened 2020-10-02
Use navigate where possible for fairs routing #trivial

sweir27:

### Description

This is part of a series of PRs that will update fairs routing to support the new fair view.

Fairs routing is inherently confusing because:
- Externally (i.e. search results, from web), fairs are linked to via a URL that looks like `/:profileSlug`. This means that for any fair view, we need to be able to support this top-level URL.
- Internally within eigen, we currently use a combination of that top-level URL and routes that are scoped like `/fair/:fairID`.

What this PR does:
- Gets rid of `SwitchBoard.presentFairViewController` in favor of our new `navigate` helper. (I'm guessing/assuming this is desired from a general ecosystem perspective!)
- Adds a new route to our routes file: `/fair/:fairID`, which will route to the main Fair screen. (Previously, the only way to get to this screen was via the vanity URL).
- Updates as many places as I could find to, internally, route us to `/fair/:fairID` instead of the vanity URL.

Why?
- In future PRs, I need to add the following logic:
  - As a user, if I have a `Feature flag enabled` _or_ the fair I'm looking at is not part of a `specified list of fair slugs`, take me to the new fair view. Else, take me to the old one.
- I've already dug into that ☝️ implementation, and it is especially difficult given how many places and ways we render and route to the fair view. The goal of this PR is to simplify the logic in preparation for adding on the additional complexity.

~Note: the changes are pretty innocuous, but I'm having difficulty testing every place as city guide does not work with Xcode 12 😭!~

(Downgraded to Xcode 11 and was able to test!)

### DIFF
--- a/src/lib/NativeModules/SwitchBoard.tsx
+++ b/src/lib/NativeModules/SwitchBoard.tsx
@@ -39,10 +39,6 @@ function presentPartnerViewController(component: React.Component<any, any>, slug
   presentEntityViewController(component, slug, EntityType.Partner, SlugType.ProfileID)
 }
 
-function presentFairViewController(component: React.Component<any, any>, slug: string, slugType?: SlugType) {
-  presentEntityViewController(component, slug, EntityType.Fair, slugType ?? SlugType.FairID)
-}
-
 function presentModalViewController(_component: React.Component<any, any>, route: string) {
   navigate(route, { modal: true })
 }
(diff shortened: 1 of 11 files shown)


sweir27:
I verified locally and it seems to be 👍 on all of the city guide views I could find!

ashfurrow (approved):
Awesome! Yes, `navigate()` is the way of the future but we haven't mass-migrated the code yet. Thank you for the thorough test coverage 🙇

PR #3967, opened 2020-10-05
fix: Gallery follow button doesnt follow

pvinis:

This PR resolves [MX-535]

### Description

Turns out the error was not from a gallery not being there, it was from a wrong if check.

We were checking on `!partnerFollowed` with `partnerFollowed: boolean | null | undefined`, which means that the above check would get be truthy for `null` (correctly), `undefined` (correctly), and `false` (incorrectly). Basically if the gallery was not followed, then we would not procceed, which would block us from following.

I changed the check from `!partnerFollowed` to `partnerFollowed == null` so it only becomes truthy in the `null` and `undefined` cases. I feel this is a nice way to check but if anyone can think of a better way, I'm open to suggestions.

As an extra thing, I thought I'll add https://palantir.github.io/tslint/rules/strict-boolean-expressions/ to our lint config. When I added it, it showed a warning under the old check, and it's gone with the new check. Feedback for this is welcome as well. The `ignore-rhs` is for jsx cases like
```jsx
{someCheck && <View />} 
```

### DIFF
--- a/src/lib/Scenes/Fair/Components/FairBoothHeader.tsx
+++ b/src/lib/Scenes/Fair/Components/FairBoothHeader.tsx
@@ -34,7 +34,7 @@ export class FairBoothHeader extends React.Component<Props, State> {
   @track((props, _, args) => {
     const partnerSlug = args[0]
     const partnerID = args[1]
-    const partnerFollowed = props.show.partner?.profile?.is_followed
+    const partnerFollowed = props.show.partner?.profile?.is_followed!
     const actionName = partnerFollowed ? Schema.ActionNames.GalleryFollow : Schema.ActionNames.GalleryUnfollow
     return {
       action_name: actionName,
@@ -61,7 +61,7 @@ export class FairBoothHeader extends React.Component<Props, State> {
         id: partnerRelayID,
       },
     } = show
-    const { is_followed: partnerFollowed, internalID: profileID } = show.partner?.profile || {}
+    const { is_followed: partnerFollowed, internalID: profileID } = show.partner?.profile ?? {}
 
     if (profileID === undefined || partnerFollowed == null) {
       return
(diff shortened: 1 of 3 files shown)


src/lib/Scenes/Fair/Components/FairBoothHeader.tsx
```diff
@@ -61,9 +61,9 @@ export class FairBoothHeader extends React.Component<Props, State> {
         id: partnerRelayID,
       },
     } = show
-    const { is_followed: partnerFollowed, internalID: profileID } = show.partner?.profile || {}
+    const { is_followed: partnerFollowed, internalID: profileID } = show.partner?.profile ?? {}
 
-    if (!profileID || !partnerFollowed) {
+    if (profileID === undefined || partnerFollowed == null) {
```
brainbicycle, line 66: Should the second also be strict equals or are they deliberately different? `partnerFollowed === null`

pvinis: `=== undefined` checks for `undefined`

`=== null` checks for `null`

`== null` checks for both `undefined` and `null`

so I put `===` on the first because it's `string | undefined` and `==` on the second because it's `boolean | null | undefined` 🤔.

If it's not too clean I could replace it with
```
if (profileID === undefined || partnerFollowed === null || partnerFollowed === undefined) {
```
?

brainbicycle: Thanks for clarification 👍 if it is intentional I think it is fine as is.

tslint.json
```diff
@@ -17,6 +17,7 @@
     "ordered-imports": true,
     "switch-default": false,
     "relay-operation-generics": [true, { "artifactDirectory": "__generated__", "makeRelative": false }],
-    "variable-name": [true, "allow-pascal-case", "allow-leading-underscore"]
+    "variable-name": [true, "allow-pascal-case", "allow-leading-underscore"],
+    "strict-boolean-expressions": [true, "ignore-rhs"]
```
brainbicycle, line 21: 🎉 I think it is a good addition personally especially if it prevents bugs like this in future

ashfurrow (approved):
Ah, great catch! Good thinking with the TSLint config too 👍

PR #3968, opened 2020-10-05
fix(formatting): Fix register text in auction

pvinis:

This PR resolves [CX-537] (woop! first CX issue!)

### Description

The text was displayed incorrectly. Now it should be good.
Screenshot 2020-10-05 at 17 19 57
Screenshot 2020-10-05 at 17 45 36

I basically added two more `replace`s. One is for removing the newline, and the other is for making the months capitalized, since `capitalize` doesn't do it (I guess it cares only about "sentences", and we don't have a full-stop, so it skips the month).

I tried adding tests but it got a bit tricky because I would need to have them at the top level, in the QueryRenderer, and that feels a bit meh? I might add some there..

[CX-537]: https://artsyproduct.atlassian.net/browse/CX-537

### DIFF
--- a/src/lib/Scenes/Sales/Components/SaleListItem.tsx
+++ b/src/lib/Scenes/Sales/Components/SaleListItem.tsx
@@ -26,7 +26,9 @@ export class SaleListItem extends React.Component<Props> {
   render() {
     const sale = this.props.sale
     const image = sale.coverImage
-    const timestamp = capitalize(sale.displayTimelyAt?.replace(/M$/, "mo").replace("\n", " "))
+    const timestamp = capitalize(
+      sale.displayTimelyAt?.replace(/M$/, "mo").replace("\n", " ")
+    ).replace(/(jan|feb|mar|apr|may|jun|jul|aug|sept|oct|nov|dec)/, (s) => capitalize(s))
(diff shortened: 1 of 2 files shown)


src/lib/Scenes/Sales/Components/SaleListItem.tsx
```diff
@@ -26,7 +26,9 @@ export class SaleListItem extends React.Component<Props> {
   render() {
     const sale = this.props.sale
     const image = sale.coverImage
-    const timestamp = capitalize(sale.displayTimelyAt?.replace(/M$/, "mo").toLowerCase())
+    const timestamp = capitalize(
+      sale.displayTimelyAt?.replace(/M$/, "mo").replace("\n", " ")
```
ashfurrow, line 30: It was coming back with a newline? That's so weird! Okay.

ashfurrow (approved):

src/lib/Scenes/Sales/Components/SaleListItem.tsx
pvinis: It was.. and the weird thing is that I didn't see it changed in MP for the last year or more 🤷.

PR #3972, opened 2020-10-06
Feat: Adds lab option and allows for routing to new fair page

sweir27:

This PR resolves [FX-2327]

### Description

This PR:
1. Adds a new lab option (`AROptionsNewFairPage`)
2. Adds logic to route to the new fair page if this option is enabled in our `routes` file.
3. Adds logic to route to the new fair view when we click into a fair from search results (or otherwise access the fair at a `/:profileSlug` route).

**Old view routing**
oldview

**New view routing**
newview

[FX-2327]: https://artsyproduct.atlassian.net/browse/FX-2327


src/lib/Scenes/VanityURL/VanityURLEntity.tsx
```diff
@@ -34,10 +41,12 @@ const VanityURLEntity: React.FC<EntityProps> = ({ fairOrPartner, originalSlug })
 
 const VanityURLEntityFragmentContainer = createFragmentContainer(VanityURLEntity, {
   fairOrPartner: graphql`
-    fragment VanityURLEntity_fairOrPartner on VanityURLEntityType {
+    fragment VanityURLEntity_fairOrPartner on VanityURLEntityType
+    @argumentDefinitions(useNewFairView: { type: "Boolean", defaultValue: false }) {
       __typename
       ... on Fair {
-        ...Fair_fair
+        ...Fair2_fair @include(if: $useNewFairView)
```
dblandin, line 48: Cool GraphQL field decorators!

sweir27, line 48: I added these `@include` and `@skip` directives so we can avoid querying for more than we need.

Unfortunately... even with this, relay still yells at us for field name conflicts between these two top-level fragment containers. I believe this is because `@skip` and `@include` only affect the graphQL query itself, but [relay doesn't consider them at compile time](https://github.com/facebook/relay/issues/1791#issuecomment-303401853).

I aliased the conflicting fields above, will leave a comment up there too!

src/lib/navigation/routes.tsx
```diff
@@ -48,6 +48,24 @@ function getDomainMap(): Record<string, RouteMatcher[] | null> {
     new RouteMatcher("/*", "LiveAuction", (params) => ({ slug: params["*"] })),
   ])
 
+  const fairRoutes = getCurrentEmissionState().options.AROptionsNewFairPage
```
dblandin, line 51: I think this is true but just to confirm, does changing one of these AR options cause a full app reload? If not, would these routing rules refresh somehow?

src/lib/Scenes/Fair2/Components/Fair2Header.tsx
```diff
@@ -91,10 +104,10 @@ export const Fair2HeaderFragmentContainer = createFragmentContainer(Fair2Header,
         summary
       }
       ticketsLink
-      hours(format: MARKDOWN)
-      links(format: MARKDOWN)
-      tickets(format: MARKDOWN)
-      contact(format: MARKDOWN)
+      fairHours: hours(format: MARKDOWN) # aliased to avoid conflicts in the VanityURLQueryRenderer
```
sweir27, line 107: The fields here had to be aliased to avoid conflicts.

dblandin (approved):
One question about routing but LGTM otherwise!

src/lib/Scenes/Fair2/Fair2MoreInfo.tsx
```diff
@@ -154,11 +154,11 @@ export const Fair2MoreInfoFragmentContainer = createFragmentContainer(Fair2MoreI
         summary
       }
       ticketsLink
-      hours(format: MARKDOWN)
-      links(format: MARKDOWN)
-      tickets(format: MARKDOWN)
+      fairHours: hours(format: MARKDOWN)
```
sweir27, line 157: I aliased them here as well so they'd resolve to the same field.

src/lib/navigation/routes.tsx
sweir27: I believe in order to see the changes, you have to restart the app after selecting a new option (it doesn't happen automatically).

anandaroop (approved):
Nice!

src/lib/navigation/__tests__/routes-tests.tsx
```diff
@@ -826,8 +826,13 @@ describe("artsy.net routes", () => {
...
 
-  it("routes to Fair2", () => {
-    expect(matchRoute("/fair2/red")).toMatchInlineSnapshot(`
+  describe("Fair routing", () => {
+    describe("with the new option enabled", () => {
+      beforeEach(() => {
+        __appStoreTestUtils__?.injectEmissionOptions({ AROptionsNewFairPage: true })
+      })
+      it("routes to Fair2", () => {
+        expect(matchRoute("/fair/red")).toMatchInlineSnapshot(`
```
damassi, line 835: Thought we were trying to avoid snapshots?

ashfurrow: We're trying to avoid snapshot tests that snapshot a large view hierarchy, store them in a `.snap` file, and cause code churn. This kind of inline snapshot feels more valuable to me, especially since asserting on the individual object keys that come back isn't super-helpful (compared to this solution).

Also, speaking with David, this file was helpful mostly to aide the migration. I think longterm we probably won't need a test for each route.

sweir27: (I like these tests! They helped me make sure I had my bases covered 😅)

damassi: One of the issues with this is that one can still do `-u` to quickly update, leading to less safety. As a proposal, it seems easy enough to do something like 

```tsx
expect(matchRoute("/fair/red")).toContain({
  foo: "bar"
})
```
Which would force a dev's attention if something unexpectedly changed. Or even something like this

```tsx
const user = prepareUserInfo('test-user'); 

// user object is something like ... 
// {
//   id: 123,
//   name: 'test-user',
//   profile: {...},
//   passwordHash: '*****',
//   whatever: {...}
// }
// ... but for the test we are interested only in name and id

expect(user).toEqual(expect.objectContaining({
  id: 123,
  name: 'test-user'
}));
```
(From [this article](https://medium.com/@boriscoder/the-hidden-power-of-jest-matchers-f3d86d8101b0))

PR #3978, opened 2020-10-06
FX-2328: (feat) Adds legacyFairSlugs and legacyProfileSlugs for tailored fair routing #trivial

sweir27:

This PR resolves [FX-2328]

### Description

This PR allows us to continue to show the _old_ fair view for certain fairs (configured via `LegacyFairSlugs` and `LegacyProfileSlugs`). This is necessary so that, when we launch the new view of fairs, any currently-running (or otherwise specified) fairs can continue to use the old view for a period of time.

Some notes:
- We need to route to the new or old view based on whether the `fairID` param is in that configured list. I couldn't find a "nice" way to do that in our new router, but I tried to make it as clean as possible. Will comment inline.
- Aside from the router, we have to include this logic in our `VanityURLEntity` renderer so we can correctly route from search results (and externally, based on `:profileSlug` routing).
- It's frustrating but necessary to include _both_ the profile slug and the fair slug in our config. This is because, from some entry points, we only have the profile slug (vs. others, where we have the fair slug). Both need to be present for a given fair for this to work.

In the example below, `swab-barcelona-2020` and `swab-barcelona` are configured as `legacyFairSlugs` and `legacyProfileSlugs`, respectively.

fairrouting

[FX-2328]: https://artsyproduct.atlassian.net/browse/FX-2328

### DIFF
--- a/Artsy/App/ArtsyEcho.m
+++ b/Artsy/App/ArtsyEcho.m
@@ -5,6 +5,7 @@
 
 #import "ARAppStatus.h"
 
+
 @implementation ArtsyEcho
 
 - (instancetype)init
@@ -41,13 +42,28 @@
     return [mutableOptions copy];
 }
 
-- (BOOL)isFeatureEnabled:(NSString *)featureFlag {
+- (BOOL)isFeatureEnabled:(NSString *)featureFlag
+{
     Feature *currentFeature = self.features[featureFlag];
     if (currentFeature) {
         return currentFeature.state;
     } else {
         return NO;
     }
+}
+
+- (NSArray *)legacyFairSlugs
+{
+    Message *legacyFairMessage = self.messages[@"LegacyFairSlugs"];
+    NSArray *fairSlugs = [legacyFairMessage.content componentsSeparatedByString:@","];
+    return fairSlugs;
+}
+
+- (NSArray *)legacyFairProfileSlugs
+{
+    Message *legacyFairProfileSlugsMessage = self.messages[@"LegacyFairProfileSlugs"];
+    NSArray *fairProfileSlugs = [legacyFairProfileSlugsMessage.content componentsSeparatedByString:@","];
+    return fairProfileSlugs;
 }
 
 @end
--- a/Artsy/App/ArtsyEcho.m
+++ b/Artsy/App/ArtsyEcho.m
@@ -56,14 +56,14 @@
 {
     Message *legacyFairMessage = self.messages[@"LegacyFairSlugs"];
     NSArray *fairSlugs = [legacyFairMessage.content componentsSeparatedByString:@","];
-    return fairSlugs;
+    return fairSlugs ? fairSlugs : @[];
 }
 
 - (NSArray *)legacyFairProfileSlugs
 {
     Message *legacyFairProfileSlugsMessage = self.messages[@"LegacyFairProfileSlugs"];
     NSArray *fairProfileSlugs = [legacyFairProfileSlugsMessage.content componentsSeparatedByString:@","];
-    return fairProfileSlugs;
+    return fairProfileSlugs ? fairProfileSlugs : @[];
 }
 
 @end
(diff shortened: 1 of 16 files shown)


Artsy/App/ArtsyEcho.m
```diff
@@ -50,4 +52,18 @@ - (BOOL)isFeatureEnabled:(NSString *)featureFlag {
     }
 }
 
+- (NSArray *)legacyFairSlugs
```
sweir27, line 55: Many thanks to @brainbicycle for pairing on this!!

src/lib/navigation/util.ts
```diff
@@ -0,0 +1,32 @@
+import { AppModule } from "lib/AppRegistry"
+import { getCurrentEmissionState } from "lib/store/AppStore"
+
+export interface MatchResult {
+  type: "match"
+  module: AppModule
+  params: object
+}
+
+export function handleFairRouting(result: MatchResult): MatchResult {
```
sweir27, line 10: Because I couldn't figure out a way (based on the current API) to conditionally, based on the _actual_ param, include different matchers within the routes file... this does it based on the content of the "matched" value.

There are likely other ways of doing this!! Happy to also revisit when @ds300 is back, but let me know if you all have preferences on the approach.

Artsy/App/ArtsyEcho.m
sweir27: @brainbicycle how/where can I configure this to fall back to an empty array if the option is not set at all in echo? Right now if I remove the fields from my local echo config, I get an `NSInvalidArgumentException` (but am not sure where to put the fallback).

Artsy/App/ArtsyEcho.m
```diff
@@ -50,4 +52,18 @@ - (BOOL)isFeatureEnabled:(NSString *)featureFlag {
...
+{
+    Message *legacyFairMessage = self.messages[@"LegacyFairSlugs"];
+    NSArray *fairSlugs = [legacyFairMessage.content componentsSeparatedByString:@","];
+    return fairSlugs;
```
brainbicycle, line 59:
```suggestion
   return fairSlugs ? fairSlugs : @[];
```
You should be able to use a ternary here to do that 👍

Artsy/App/ArtsyEcho.m
brainbicycle: Added a suggestion below 👇

src/lib/Scenes/VanityURL/__tests__/VanityURLEntity-tests.tsx
```diff
@@ -188,6 +188,31 @@ describe("VanityURLEntity", () => {
...
+        errors: [],
+        data: {
+          vanityURLEntity: {
+            ...fairFixture,
+          },
+        },
+      })
+    })
+    const fairComponent = tree.root.findByType(FairContainer)
+    expect(fairComponent).toBeDefined()
```
brainbicycle, line 214: Nice test!

src/lib/Scenes/VanityURL/VanityURLEntity.tsx
```diff
@@ -63,6 +63,8 @@ interface RendererProps {
 
 export const VanityURLEntityRenderer: React.FC<RendererProps> = ({ entity, slugType, slug }) => {
   const showNewFairViewFeatureEnabled = getCurrentEmissionState().options.AROptionsNewFairPage
+  const fairProfileSlugs = getCurrentEmissionState().legacyFairProfileSlugs
+  const useNewFairView = showNewFairViewFeatureEnabled && !fairProfileSlugs?.includes(slug)
 
   if (slugType === "fairID") {
     return <FairQueryRenderer fairID={slug} />
```
anandaroop, line 70: Just confirming that it makes sense for this code path to not care about the `useNew...` conditional logic?

anandaroop (approved):
Tricky stuff! Looks good as far as I can tell.

brainbicycle (approved):

src/lib/navigation/util.ts
damassi: @sweir27 - it would be a little clearer if this file was named `handleFairRouting`; `util` is ambiguous, and not discoverable by looking in the file tree.

sweir27: Good call @damassi !

PR #3980, opened 2020-10-07
feat: shipping modal - first inquiry

jo-rs:

This PR resolves [PURCHASE-2097]

### Description

This PR adds the shipping location modal, with a GMaps autocomplete field, for first inquiry.

It also:
- adds ability to disable the right button on the FancyModalHeader component.
- creates an e-e `InquiryModal` test and a test for the `LocationAutocomplete` component
  - The tests include some `press` and `type` helpers which might be good or controversial. They mostly work well!
- creates a `lib/utils/googleMaps` module which extracts the querying logic from the Consignments `Location` component, but does not change that component (which also does some other stuff using the same api, so a partial update there felt both inadequate and out of scope).

location_modal

Long texts are handled well in all locations (suggestions dropdown, input box and here):
image

[PURCHASE-2097]: https://artsyproduct.atlassian.net/browse/PURCHASE-2097


src/lib/Scenes/Artwork/Components/CommercialButtons/__tests__/LocationAutocomplete-tests.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  it("Highlights matched text in the autocomplete", async () => {
+    ;(queryLocation as jest.Mock).mockResolvedValue([{ id: "x", name: "Hello, World" }])
+    await typeInInput(wrapper.root, "hell World")
+
+    expect(extractText(wrapper.root)).toContain("Hello, World")
+    expect(wrapper.root.findAllByProps({ fontWeight: "bold" }).map((t) => extractText(t))).toEqual(
+      expect.arrayContaining(["Hell", "World"])
+    )
+  })
```
jo-rs, line 87: 💯

src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
```diff
@@ -25,48 +26,73 @@ interface InquiryModalProps {
...
+    const isShipping = inquiry === InquiryOptions.Shipping
+
     return (
-      <TouchableOpacity onPress={() => toggleExpanded()} key={inquiry}>
+      <React.Fragment key={inquiry}>
         <InquiryField>
           <Flex flexDirection="row" justifyContent="space-between">
             <Flex flexDirection="row">
               <Checkbox
+                data-test-id={`checkbox-${inquiry}`}
```
zephraph, line 51: Nice!

src/lib/Scenes/Artwork/Components/CommercialButtons/__tests__/helpers.ts
```diff
@@ -0,0 +1,38 @@
...
+  {
+    text = "",
+    componentType = TouchableOpacity,
+  }: {
+    /** A string or RegExp (for exactness) to match - defaults to "" (everything) */
+    text?: string | RegExp
+    /** The type of component to search for - default TouchableOpacity */
+    componentType?: React.ComponentType
+  }
+) => {
```
zephraph, line 19: cc @pvinis, this is the kind of reason why I think react native testing library could be a good addition. While you _can_ implement these more generic handlers, we really shouldn't have to. I think for this PR I'm 👍 with this handler, but I'd like to make an RFC to introduce that lib. It just feels like it'd be better than w/ the limited enzyme usage we have.

src/lib/Scenes/Artwork/Components/CommercialButtons/ShippingModal.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const { location, toggleVisibility, modalIsVisible, setLocation } = props
+
+  const [locationInput, setLocationInput] = useState("")
+
+  return (
+    <FancyModal visible={modalIsVisible} onBackgroundPressed={() => toggleVisibility()}>
+      <FancyModalHeader
+        leftButtonText="Cancel"
+        onLeftButtonPress={() => {
+          throw new Error("honk")
```
zephraph, line 25: honk

src/lib/Scenes/Artwork/Components/CommercialButtons/LocationAutocomplete.tsx
```diff
@@ -0,0 +1,158 @@
...
+    } else {
+      ;(async () => {
+        const googlePredictions = await queryLocation(query)
+        setPredictions(googlePredictions)
+      })()
+    }
+  }, [query])
+
+  const reset = () => {
+    if (Boolean(selectedLocation)) {
```
zephraph, line 45: **non-blocking**: Don't see this type of boolean casting often. Probably should choose between `!!` or `Boolean` and apply that a bit more consistently.

src/lib/Scenes/Artwork/Components/CommercialButtons/__tests__/InquiryModal-tests.tsx
```diff
@@ -0,0 +1,150 @@
+jest.mock("../../../../../utils/googleMaps", () => ({ queryLocation: jest.fn() }))
```
zephraph, line 1: Thanks for these tests! They're great.

zephraph (approved):

zephraph:
This is great! I don't see anything major to address and we can follow up on anything minor. Thanks @jo-rs, @erikdstock.

src/lib/Scenes/Artwork/Components/CommercialButtons/__tests__/helpers.ts
pvinis: This feels like it's exactly what I'm doing [here](https://github.com/pvinis/react-native-testing-library/blob/add-getByTextWithType/src/helpers/getByAPI.js#L107) and [use like this](https://github.com/pvinis/react-native-testing-library/blob/add-getByTextWithType/src/__tests__/render.test.js#L199-L222).

In [this PR](https://github.com/artsy/eigen/pull/3897) I'm using that too. I would say once this PR here for the modal is merged, we can try it out and remove this helper and just use the one I have in combination with `fireEvent.press`.

xtina-starr (approved):
This looks great. Merging 🚢

PR #3983, opened 2020-10-07
Adds support for new text styles within Markdown #trivial

sweir27:
The type of this PR is: ** Feature**

### Description

I noticed while working on Fairs that we don't yet have great support for our new text styles within the existing `Markdown` component.

This PR takes a stab (using my best guess and existing tokens/variants) at providing support for this.

Within the ReadMore component:
image

Within the Markdown component:
image

### DIFF
--- a/src/lib/Components/Bidding/Screens/RegistrationResult.tsx
+++ b/src/lib/Components/Bidding/Screens/RegistrationResult.tsx
@@ -67,15 +67,18 @@ const registrationNetworkErrorMessage = {
   description: "Please\ncheck your internet connection\nand try again.",
 }
 
-const markdownRules = defaultRules(true, {
-  paragraph: {
-    match: blockRegex(/^((?:[^\n]|\n(?! *\n))+)(?:\n *)/),
-    react: (node, output, state) => {
-      return (
-        <Sans size="4" key={state.key} textAlign="center">
-          {output(node.content, state)}
-        </Sans>
-      )
+const markdownRules = defaultRules({
+  modal: true,
+  ruleOverrides: {
+    paragraph: {
+      match: blockRegex(/^((?:[^\n]|\n(?! *\n))+)(?:\n *)/),
+      react: (node, output, state) => {
+        return (
+          <Sans size="4" key={state.key} textAlign="center">
+            {output(node.content, state)}
+          </Sans>
+        )
+      },
     },
   },
 })
--- a/src/lib/Components/ReadMore.tsx
+++ b/src/lib/Components/ReadMore.tsx
@@ -32,7 +32,8 @@ export const ReadMore = React.memo(
   ({ content, maxChars, presentLinksModally, color, trackingFlow, contextModule, textStyle = "serif" }: Props) => {
     const [isExpanded, setIsExpanded] = useState(false)
     const tracking = useTracking()
-    const basicRules = defaultRules(presentLinksModally)
+    const useNewTextStyles = textStyle === "new"
+    const basicRules = defaultRules({ modal: presentLinksModally, useNewTextStyles })
     const TextComponent: React.ComponentType<SansProps | SerifProps | PaletteTextProps> = (textStyle === "new"
       ? PaletteText
       : textStyle === "sans"
@@ -55,6 +56,7 @@ export const ReadMore = React.memo(
         },
       },
     }
+
     const root = renderMarkdown(content, rules)
     // Removes the last empty space in the markdown array
     if (Array.isArray(root)) {
--- a/src/lib/utils/renderMarkdown.tsx
+++ b/src/lib/utils/renderMarkdown.tsx
@@ -161,7 +161,7 @@ export function defaultRules({
           }
 
           const listItemText = useNewTextStyles ? (
-            <Text variant="text" key={state.key}>
+            <Text variant="text" key={String(state.key) + 1}>
               {output(node.content, state)}
             </Text>
           ) : (
--- a/src/lib/utils/renderMarkdown.tsx
+++ b/src/lib/utils/renderMarkdown.tsx
@@ -1,9 +1,9 @@
 import { LinkText } from "lib/Components/Text/LinkText"
 import SwitchBoard from "lib/NativeModules/SwitchBoard"
 import _ from "lodash"
-import { color, Sans, Separator, Serif, space } from "palette"
+import { color, Sans, Separator, Serif, space, Text } from "palette"
 import React from "react"
-import { Linking, Text, View } from "react-native"
+import { Linking, Text as RNText, View } from "react-native"
 import SimpleMarkdown, { ParserRule, ParserRules, ReactNodeOutput } from "simple-markdown"
 
 interface OurReactRule extends Partial<ParserRule> {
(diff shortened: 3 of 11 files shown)


src/lib/utils/renderMarkdown.tsx
```diff
@@ -211,11 +262,11 @@ export function defaultRules(modal: boolean = false, roleOverrides: MarkdownRule
     hr: {
       react: () => <Separator mb={2}></Separator>,
     },
-    ...roleOverrides,
+    ...ruleOverrides,
   })
 }
 
-export function renderMarkdown(markdown: string, rules: any = defaultRules(false)): React.ReactElement {
+export function renderMarkdown(markdown: string, rules: any = defaultRules({})): React.ReactElement {
```
sweir27, line 269: This defaults to `false` above.

src/lib/utils/renderMarkdown.tsx
```diff
@@ -33,7 +33,15 @@ function createReactRules(rules: MarkdownRules): ParserRules {
 // https://github.com/CharlesMangwa/react-native-simple-markdown/blob/next/src/rules.js for new functionalities.
 //
 // Default rules: https://github.com/Khan/simple-markdown/blob/f1a75785703832bbff146d0b98e76cd7ac74b8e8/simple-markdown.js#L806
-export function defaultRules(modal: boolean = false, roleOverrides: MarkdownRules = {}): ParserRules {
+export function defaultRules({
+  modal = false,
+  ruleOverrides = {},
```
sweir27, line 38: Renamed `roleOverrides` -> `ruleOverrides` (since I'm pretty sure that was a typo?)

src/lib/utils/renderMarkdown.tsx
```diff
@@ -33,7 +33,15 @@ function createReactRules(rules: MarkdownRules): ParserRules {
 // https://github.com/CharlesMangwa/react-native-simple-markdown/blob/next/src/rules.js for new functionalities.
 //
 // Default rules: https://github.com/Khan/simple-markdown/blob/f1a75785703832bbff146d0b98e76cd7ac74b8e8/simple-markdown.js#L806
-export function defaultRules(modal: boolean = false, roleOverrides: MarkdownRules = {}): ParserRules {
+export function defaultRules({
```
sweir27, line 36: Decided to name the arguments here so we weren't passing around a bunch of opaque booleans (now that I'm adding another 🙃)!

src/lib/Components/Bidding/Screens/RegistrationResult.tsx
```diff
@@ -67,15 +67,18 @@ const registrationNetworkErrorMessage = {
...
-const markdownRules = defaultRules(true, {
-  paragraph: {
-    match: blockRegex(/^((?:[^\n]|\n(?! *\n))+)(?:\n *)/),
-    react: (node, output, state) => {
-      return (
-        <Sans size="4" key={state.key} textAlign="center">
-          {output(node.content, state)}
-        </Sans>
-      )
+const markdownRules = defaultRules({
```
sweir27, line 70: Most of the changes in this PR are updating the argument format for this function-- the underlying functionality should be the same.

src/lib/Components/ReadMore.tsx
```diff
@@ -32,7 +32,8 @@ export const ReadMore = React.memo(
   ({ content, maxChars, presentLinksModally, color, trackingFlow, contextModule, textStyle = "serif" }: Props) => {
     const [isExpanded, setIsExpanded] = useState(false)
     const tracking = useTracking()
-    const basicRules = defaultRules(presentLinksModally)
+    const useNewTextStyles = textStyle === "new"
+    const basicRules = defaultRules({ modal: presentLinksModally, useNewTextStyles })
```
sweir27, line 36: The existing `textStyle` update appears to only affect the `paragraph` rule. `useNewTextStyles` is more extensive.

src/lib/utils/renderMarkdown.tsx
```diff
@@ -98,7 +114,11 @@ export function defaultRules(modal: boolean = false, roleOverrides: MarkdownRule
 
     em: {
       react: (node, output, state) => {
-        return (
+        return useNewTextStyles ? (
+          <Text variant="caption" key={state.key}>
```
sweir27, line 118: Most of the choices here should be non-controversial 🤞 , but this one is a little weird. I don't think we have an `italic` variant? I chose `caption` but not a strong opinion by any means.

sweir27: cc @dzucconi

dzucconi: It's not part of any "variant" but `fontStyle="italic"` works for regular and bold weights of Unica on web. So long as eigen supports has the italic fonts it should work here as well.

sweir27: Will check...

dzucconi: that is: `<Text variant="text" fontStyle="italic">`

sweir27: image

Seems okay!

anandaroop: Oh nice! That reminds me of [this observation](https://artsy.slack.com/archives/C9SATFLUU/p1601305940105000), which I've been meaning to follow up on. Seems like this PR will either fix that or make it an easy change 🎉

src/lib/utils/renderMarkdown.tsx
```diff
@@ -123,25 +143,37 @@ export function defaultRules(modal: boolean = false, roleOverrides: MarkdownRule
...
+            ) : (
               <Serif size="3t" key={state.key}>
                 -{" "}
               </Serif>
             )
           }
 
-          const listItemText = (
+          const listItemText = useNewTextStyles ? (
+            <Text variant="text" key={state.key}>
```
williardx, line 164: The `Serif` below has a different `key` value than this `Text` -- did you mean to do that?

src/lib/utils/renderMarkdown.tsx
sweir27: The screenshot right above is actually from that series! So it should be fixed 😄 .

src/lib/utils/renderMarkdown.tsx
williardx: Ah, I thought for some reason we were rendering those values but we're not, so it probably doesn't matter. Though I'm curious as to why we were using `String(state.key) + 1` for `Serif` 🤔

sweir27: Ohh good catch! I'll keep it the same.

williardx:
LGTM!

src/lib/utils/renderMarkdown.tsx
```diff
@@ -177,9 +217,20 @@ export function defaultRules(modal: boolean = false, roleOverrides: MarkdownRule
...
+        const newTextMap = {
+          1: "largeTitle",
+          2: "title",
+          3: "subtitle",
+          4: "text",
+        }
         // @ts-ignore STRICTNESS_MIGRATION
-        const size = map[node.level] || "4"
-        return (
+        const size = useNewTextStyles ? newTextMap[node.level] || "subtitle" : map[node.level] || "4"
```
ashfurrow, line 228: This took me a second to parse. It looks like we're falling back to either `"subtitle"`, or `"4"`. Just double-checking that this is intended?

ashfurrow (approved):
Looks good! Just one question, not blocking 👍

src/lib/utils/renderMarkdown.tsx
sweir27: Yep! I took a guess-- `subtitle` is slightly larger than `text` (which roughly maps to the defaulting to `4` behavior).

sweir27:
Gonna merge so we can test tomorrow but happy to follow up!

PR #3993, opened 2020-10-08
fix: Fixes a few issues from fair QA #trivial

sweir27:

### Description

This PR fixes two issues:
1. Clicking into "Photo London Fair" search result led to an "Unable to load" screen. This is actually a bug in the current app version, due to some mismatch between the `slug` returned from the search result and the `href`. I believe that `href` [should be the correct field](https://github.com/artsy/metaphysics/blob/3f16e6c5a9c0f2e40bb91ab00cddad5ff307b1eb/src/schema/v2/SearchableItem/SearchableItemPresenter.ts#L50-L51).
2. Collections return `null` for their `internalID` field. I [created a ticket](https://artsyproduct.atlassian.net/browse/FX-2343) to address the underlying issue (either in MP or KAWS), but this will fix the current problem within eigen.

### DIFF
--- a/src/lib/Scenes/Fair2/Components/Fair2Collections.tsx
+++ b/src/lib/Scenes/Fair2/Components/Fair2Collections.tsx
@@ -57,7 +57,7 @@ export const Fair2Collections: React.FC<Fair2CollectionsProps> = ({ fair, ...res
             <TouchableWithScale
               key={collection.slug}
               onPress={() => {
-                trackTappedCollection(collection.internalID, collection.slug)
+                trackTappedCollection(collection.slug, collection.slug)
                 navigate(`/collection/${collection.slug}`)
               }}
             >
@@ -76,7 +76,6 @@ export const Fair2CollectionsFragmentContainer = createFragmentContainer(Fair2Co
       internalID
       slug
       marketingCollections(size: 4) {
-        internalID
         slug
         title
         category
--- a/src/lib/Scenes/Search/SearchResult.tsx
+++ b/src/lib/Scenes/Search/SearchResult.tsx
@@ -107,7 +107,7 @@ function navigateToResult(result: AutosuggestResult, navRef: React.MutableRefObj
   if (result.displayType === "Gallery" || result.displayType === "Institution") {
     SwitchBoard.presentPartnerViewController(navRef.current, result.slug!)
   } else if (result.displayType === "Fair") {
-    const fairProfileUrl = `/${result.slug!}?entity=${EntityType.Fair}&slugType=${SlugType.ProfileID}`
+    const fairProfileUrl = `${result.href}?entity=${EntityType.Fair}&slugType=${SlugType.ProfileID}`
     navigate(fairProfileUrl)
   } else {
     SwitchBoard.presentNavigationViewController(navRef.current, result.href!)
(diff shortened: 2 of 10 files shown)


src/lib/Scenes/Fair2/Components/Fair2Collections.tsx
```diff
@@ -57,7 +57,7 @@ export const Fair2Collections: React.FC<Fair2CollectionsProps> = ({ fair, ...res
             <TouchableWithScale
               key={collection.slug}
               onPress={() => {
-                trackTappedCollection(collection.internalID, collection.slug)
+                trackTappedCollection(collection.slug, collection.slug)
```
sweir27, line 60: I think this is fine for now.

src/lib/Scenes/Search/SearchResult.tsx
```diff
@@ -107,7 +107,7 @@ function navigateToResult(result: AutosuggestResult, navRef: React.MutableRefObj
   if (result.displayType === "Gallery" || result.displayType === "Institution") {
     SwitchBoard.presentPartnerViewController(navRef.current, result.slug!)
   } else if (result.displayType === "Fair") {
-    const fairProfileUrl = `/${result.slug!}?entity=${EntityType.Fair}&slugType=${SlugType.ProfileID}`
+    const fairProfileUrl = `${result.href}?entity=${EntityType.Fair}&slugType=${SlugType.ProfileID}`
```
sweir27, line 110: `href` includes a slash already.

dblandin (approved):
LGTM! 👍

dblandin:
A couple meta non-blocking notes before merging:

- You mentioned this addresses two bugs. Would it be worthwhile to split the fixes into two different commits?
- Also wondering if we're also encouraging semantic/conventional commit formatting in this repository as per https://github.com/artsy/README/issues/327

Could see these potentially represented as something like:

```
- fix(FX-2343): Use slug instead of internalID for collection analytics tracking
- fix(?): Use fair href instead of slug to fix search result navigation 
```

The scope above is optional but I've found it useful to use it for the Jira ticket number ☝️ 

Assigning back to you. Feel free to merge when ready 👍

sweir27:
Ah, sorry! I made this PR before we had a Jira ticket. Agree in general about linking them, although in general I feel like project management software comes and goes so the PR feels like a more useful places to put that.

You're right that this would be cleaner with two commits, but aren't we supposed to squash now anyways?

dblandin:
> Ah, sorry! I made this PR before we had a Jira ticket. Agree in general about linking them, although in general I feel like project management software comes and goes so the PR feels like a more useful place to put that.

Nothing to be sorry about! And for sure, it's nice to have the context in the PR as well.

> You're right that this would be cleaner with two commits, but aren't we supposed to squash now anyways?

Pretty sure the gist is that whichever commits land on the master branch should follow semantic formatting but it's up to the PR author whether they want to squash into a single commit or craft multiple commits, each using that formatting, and rebase merge.

sweir27:
> Pretty sure the gist is that whichever commits land on the master branch should follow semantic formatting but it's up to the PR author whether they want to squash into a single commit or craft multiple commits, each using that formatting, and rebase merge.

Ah, am still learning the new way.

PR #3994, opened 2020-10-08
PURCHASE Add Artwork context provider for passing/persisting modal state

ashleyjelks:

This PR resolves [PURCHASE-2096]

### Description

This PR adds a context provider around the Artwork app to enable passing and persisting state between views in the inquiry modal.

This is an initial pass, so we can safely assume the shape of the global data state will evolve.

Kapture 2020-10-09 at 14 47 47

Kapture 2020-10-09 at 14 02 57

[PURCHASE-2096]: https://artsyproduct.atlassian.net/browse/PURCHASE-2096

### DIFF
--- a/src/lib/Scenes/Artwork/Artwork.tsx
+++ b/src/lib/Scenes/Artwork/Artwork.tsx
@@ -10,6 +10,7 @@ import { defaultEnvironment } from "lib/relay/createEnvironment"
 import { ArtistSeriesMoreSeriesFragmentContainer as ArtistSeriesMoreSeries } from "lib/Scenes/ArtistSeries/ArtistSeriesMoreSeries"
 import { getCurrentEmissionState } from "lib/store/AppStore"
 import { AboveTheFoldQueryRenderer } from "lib/utils/AboveTheFoldQueryRenderer"
+import { ArtworkInquiryContext, ArtworkInquiryStateProvider } from "lib/utils/ArtworkInquiry/ArtworkInquiryStore"
 import {
   PlaceholderBox,
   PlaceholderRaggedText,
@@ -288,17 +289,25 @@ export class Artwork extends React.Component<Props, State> {
 
   render() {
     return (
-      <FlatList<ArtworkPageSection>
-        data={this.sections()}
-        ItemSeparatorComponent={() => (
-          <Box mx={2} my={3}>
-            <Separator />
-          </Box>
-        )}
-        refreshControl={<RefreshControl refreshing={this.state.refreshing} onRefresh={this.onRefresh} />}
-        contentContainerStyle={{ paddingBottom: 40 }}
-        renderItem={({ item }) => (item.excludePadding ? item.element : <Box px={2}>{item.element}</Box>)}
-      />
+      <ArtworkInquiryStateProvider>
+        <ArtworkInquiryContext.Consumer>
+          {(_context) => {
+            return (
+              <FlatList<ArtworkPageSection>
+                data={this.sections()}
+                ItemSeparatorComponent={() => (
+                  <Box mx={2} my={3}>
+                    <Separator />
+                  </Box>
+                )}
+                refreshControl={<RefreshControl refreshing={this.state.refreshing} onRefresh={this.onRefresh} />}
+                contentContainerStyle={{ paddingBottom: 40 }}
+                renderItem={({ item }) => (item.excludePadding ? item.element : <Box px={2}>{item.element}</Box>)}
+              />
+            )
+          }}
+        </ArtworkInquiryContext.Consumer>
+      </ArtworkInquiryStateProvider>
     )
   }
 }
--- a/src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
+++ b/src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
@@ -6,7 +6,8 @@ import ChevronIcon from "lib/Icons/ChevronIcon"
 import { ArtworkInquiryContext } from "lib/utils/ArtworkInquiry/ArtworkInquiryStore"
 import { InquiryOptions } from "lib/utils/ArtworkInquiry/ArtworkInquiryTypes"
 import { Box, color, Flex, Separator, space, Text } from "palette"
-import React, { useContext } from "react"
+import React, { useContext, useState } from "react"
+import { LayoutAnimation, TouchableOpacity } from "react-native"
 import NavigatorIOS from "react-native-navigator-ios"
 import { createFragmentContainer, graphql } from "react-relay"
 import styled from "styled-components/native"
@@ -25,30 +26,47 @@ export const InquiryModal: React.FC<InquiryModalProps> = ({ artwork, ...props })
   const { toggleVisibility, modalIsVisible } = props
   const questions = artwork?.inquiryQuestions!
   const { state } = useContext(ArtworkInquiryContext)
+  const [isExpanded, setExpanded] = useState(false)
+  const toggleExpanded = () => {
+    LayoutAnimation.configureNext({
+      ...LayoutAnimation.Presets.linear,
+      duration: 200,
+    })
+    setExpanded(!isExpanded)
+  }
 
   const renderInquiryQuestion = (inquiry: string): JSX.Element => {
     return (
-      <InfoBox key={inquiry}>
-        <Flex flexDirection="row">
-          <Checkbox
-            checked={state.inquiryType === InquiryOptions.RequestPrice && inquiry === InquiryOptions.PriceAvailability}
-          />
-          <Text variant="text">{inquiry}</Text>
-        </Flex>
-        {inquiry === InquiryOptions.Shipping && (
-          <>
-            <Separator my={2} />
-            <Flex flexDirection="row" justifyContent="space-between">
-              <Text variant="text" color="black60">
-                Add your location
-              </Text>
-              <Box mt={0.5}>
-                <ChevronIcon color="black60" />
-              </Box>
+      <TouchableOpacity onPress={() => toggleExpanded()} key={inquiry}>
+        <InquiryField>
+          <Flex flexDirection="row" justifyContent="space-between">
+            <Flex flexDirection="row">
+              <Checkbox
+                checked={
+                  state.inquiryType === InquiryOptions.RequestPrice && inquiry === InquiryOptions.PriceAvailability
+                }
+              />
+              <Text variant="text">{inquiry}</Text>
             </Flex>
-          </>
-        )}
-      </InfoBox>
+            {inquiry === InquiryOptions.Shipping && (
+              <Flex flexDirection="row" mt={0.5}>
+                <ChevronIcon color="black60" expanded={isExpanded} initialDirection="down" />
+              </Flex>
+            )}
+          </Flex>
+
+          {inquiry === InquiryOptions.Shipping && !!isExpanded && (
+            <>
+              <Separator my={2} />
+              <Flex flexDirection="row" justifyContent="space-between">
+                <Text variant="text" color="black60">
+                  Add your location
+                </Text>
+              </Flex>
+            </>
+          )}
+        </InquiryField>
+      </TouchableOpacity>
     )
   }
 
@@ -72,7 +90,7 @@ export const InquiryModal: React.FC<InquiryModalProps> = ({ artwork, ...props })
   )
 }
 
-const InfoBox = styled(Flex)`
+const InquiryField = styled(Flex)`
   border-radius: 5;
   border: solid 1px ${color("black10")};
   flex-direction: column;
--- a/src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
+++ b/src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
@@ -1,10 +1,15 @@
 import { InquiryModal_artwork } from "__generated__/InquiryModal_artwork.graphql"
+import { Checkbox } from "lib/Components/Bidding/Components/Checkbox"
 import { FancyModal } from "lib/Components/FancyModal/FancyModal"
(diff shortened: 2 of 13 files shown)


src/lib/utils/ArtworkInquiry/__tests__/ArtworkInquiryStore-tests.tsx
```diff
@@ -0,0 +1,68 @@
...
+    inquiryAction = {
+      type: "selectInquiryType",
+      payload: InquiryOptions.RequestPrice,
+    }
+
+    const r = reducer(inquiryState, inquiryAction)
+
+    expect(r).toEqual({
+      shippingLocation: null,
+      inquiryType: "Request Price",
```
erikdstock, line 24: Can the inquiry only be about a single question?

src/lib/utils/ArtworkInquiry/ArtworkInquiryTypes.ts
```diff
@@ -0,0 +1,33 @@
...
+  payload: InquiryTypes
+}
+
+interface SelectLocation {
+  type: "selectShippingLocation"
+  payload: string
+}
+
+export enum InquiryOptions {
+  RequestPrice = "Request Price",
```
erikdstock, line 28: Cool use of enum, resulting in the display text for the button being the underlying value for the state as well.

src/lib/utils/ArtworkInquiry/ArtworkInquiryTypes.ts
```diff
@@ -0,0 +1,33 @@
...
+  state: ArtworkInquiryContextState
+  dispatch: Dispatch<ArtworkInquiryActions>
+}
+
+export interface ArtworkInquiryContextState {
+  readonly inquiryType: InquiryTypes | null
+  readonly shippingLocation: string | null
+}
+
+export type InquiryTypes = "Request Price" | "Contact Gallery" | "Inquire to Purchase"
```
erikdstock, line 15: Can this be inferred from the `InquiryOptions` enum in any way? Maybe i'm confusing them - is supposed to be the header? If they are different things, maybe ignore me. One way I have approached this before has been to use a string enum/union for keys in normal objects, but then give them a `Record<Key, string>`. 
```ts
type InquiryType = "Request Price" | "Contact Gallery" | "Inquire to Purchase" | "Shipping"

const InquiryOptions: Record<string, InquiryType>  = {
  ContactGallery: "Contact Gallery",
   InquireToPurchase: "Inquire to Purchase",
  // ...
}
more fleshed-out example [here](https://www.typescriptlang.org/play?#code/PTAEHUFMBsGMHsC2lQBd5oBYoCoE8AHSAZVgCcBLA1UABWgEM8BzM+AVwDsATAGiwoBnUENANQAd0gAjQRVSQAUCEmYKsTKGYUAbpGF4OY0BoadYKdJMoL+gzAzIoz3UNEiPOofEVKVqAHSKymAAmkYI7NCuqGqcANag8ABmIjQUXrFOKBJMggBcISGgoAC0oACCoASMFmgY7p7ehCTkVOle4jUMdRLYTqCc8LEZzCZmoNJODPHFZZXVtZYYkAAeRJTInDQS8po+rf40gnjbDKv8LqD2jpbYoACqAEoAMsK7sUmxkGSCc+VVQQuaTwVb1UBrDYULY7PagbgUZLJH6QbYmJAECjuMigZEMVDsJzCFLNXxtajBVAtUAASU4AEd2BQyHgDqAALygABEtEoFgqOgYWIY0ix8jwXNAAB9uQBheBnWCoADiDGg2Il0u5dMZzMgOHgtEJpkEkElMq5xDUBExnGYXOCCE4ghoOqZLIA8tQKAqcGtUPlQE9IAgyNwADxu5mslp2VCUO0APhKnIA3ooSvLFSq1RrA1ys6gejRVeqfhLeBmSlGnAajWQTZB8zXlnRjQ5TVzKwBfRROl20hnuvAACQ83B+geDoYjNZjRDjCeYyY5oHTJTofMgAqFjFF0HF+d56hQADJKoLhfvxV3gt2gA)
```

This might be completely missing your point so if it is not helpful feel free to ignore!

src/lib/utils/ArtworkInquiry/ArtworkInquiryStore.tsx
```diff
@@ -0,0 +1,42 @@
...
+} from "lib/utils/ArtworkInquiry/ArtworkInquiryTypes"
+import React, { createContext, Reducer, useReducer } from "react"
+
+const artworkInquiryState: ArtworkInquiryContextState = {
+  shippingLocation: null,
+  inquiryType: null,
+}
+
+// NOTE: We will need to handle clearing the location fields and other CRUD like actions
+// But since we are working this in different streams we'll have to come back to this
```
erikdstock, line 14: FWIW - https://github.com/artsy/eigen/pull/3980 currently manages input state within the shipping modal and only sets it using the reducer after the user clicks apply. Not sure if that is what you are getting at here but this looks fine to me.

src/lib/utils/ArtworkInquiry/ArtworkInquiryStore.tsx
```diff
@@ -0,0 +1,42 @@
...
+        shippingLocation: action.payload,
+        inquiryType: inquiryState.inquiryType,
+      }
+  }
+}
+
+export const ArtworkInquiryContext = createContext<ArtworkInquiryContextProps>(null as any)
+
+export const ArtworkInquiryStateProvider = ({ children }: any) => {
+  const [state, dispatch] = useReducer<Reducer<ArtworkInquiryContextState, ArtworkInquiryActions>>(
```
erikdstock, line 37: I'm experienced with redux as well as `useState` and `useContext` but have never actually used the `useReducer` hook, so I don't feel qualified to review this area. On first glance it feels like a lot of overhead to construct the reducer and all of the interdependent types (compared to relying on `useState`), but I was able to understand it pretty quickly so definitely not a blocker.

erikdstock (commented):
All seems sensible to me - a few questions but no blockers from me.

src/lib/Scenes/Artwork/Artwork.tsx
```diff
@@ -288,17 +289,25 @@ export class Artwork extends React.Component<Props, State> {
...
-        ItemSeparatorComponent={() => (
-          <Box mx={2} my={3}>
-            <Separator />
-          </Box>
-        )}
-        refreshControl={<RefreshControl refreshing={this.state.refreshing} onRefresh={this.onRefresh} />}
-        contentContainerStyle={{ paddingBottom: 40 }}
-        renderItem={({ item }) => (item.excludePadding ? item.element : <Box px={2}>{item.element}</Box>)}
-      />
+      <ArtworkInquiryStateProvider>
```
xtina-starr, line 292: This might be a somewhat naive question, but could the `ArtworkInquiryStateProvider` wrap the `InquiryModal` versus the entire Artwork Component? I'm thinking the entire artwork component might not need to have knowledge of the InquiryModal's state.

src/lib/utils/ArtworkInquiry/ArtworkInquiryStore.tsx
```diff
@@ -0,0 +1,42 @@
...
+  ArtworkInquiryActions,
+  ArtworkInquiryContextProps,
+  ArtworkInquiryContextState,
+} from "lib/utils/ArtworkInquiry/ArtworkInquiryTypes"
+import React, { createContext, Reducer, useReducer } from "react"
+
+const artworkInquiryState: ArtworkInquiryContextState = {
+  shippingLocation: null,
+  inquiryType: null,
+}
```
xtina-starr, line 11: Does `ArtworkInquiryContextState` support the fact that multiple inquiry questions can be checked (i.e. price & availability and shipping quote)?

src/lib/Scenes/Artwork/Artwork.tsx
ashleyjelks: Yes, I think we could wrap the context around any component in the Artwork hierarchy, but looking at other uses of context providers of it in the app it seems follow the pattern of wrapping the index/parent.

src/lib/utils/ArtworkInquiry/ArtworkInquiryStore.tsx
ashleyjelks: This does not, no, but I added a note that the shape of the state's data will likely change to support other streams of work. InquiryType could look something like `{inquiryType: {shipping: null}, {price: null}, {condition: null}}`

src/lib/utils/ArtworkInquiry/ArtworkInquiryTypes.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export type ArtworkInquiryActions = SelectInquiryType | SelectLocation
+
+export interface ArtworkInquiryContextProps {
+  state: ArtworkInquiryContextState
+  dispatch: Dispatch<ArtworkInquiryActions>
+}
+
+export interface ArtworkInquiryContextState {
+  readonly inquiryType: InquiryTypes | null
```
xtina-starr, line 11: Because multiple inquiry questions can be selected, I'm thinking the state for those questions might be more accurately reflected as an array which contains the questions that where selected or an object where the key is the question id and value is a boolean.

This visual of how the inquiry questions could possibly be rendered in CMS might be helpful to create a mental model:
Screen Shot 2020-10-09 at 3 59 18 PM

src/lib/utils/ArtworkInquiry/ArtworkInquiryStore.tsx
ashleyjelks: Interesting, I see your point, I guess i think its less to reason about to have changes made in one place propagated to all the components that need to know about them/managing a single global state object is less complex than using `useState` with callbacks to detect state changes in between different components. I can see a case for both!

src/lib/utils/ArtworkInquiry/ArtworkInquiryTypes.ts
ashleyjelks: Good call, let me see if I can clean it up!

src/lib/utils/ArtworkInquiry/__tests__/ArtworkInquiryStore-tests.tsx
ashleyjelks: Nope, @xtina-starr asked a similar question, this is just to handle the reqs of this ticket, but I would expect this data shape to change as we handle the other inquiry fields.

src/lib/utils/ArtworkInquiry/ArtworkInquiryStore.tsx
erikdstock: yep, non-blocking, all good.

src/lib/Scenes/Artwork/Artwork.tsx
xtina-starr: My hesitation around having the `ArtworkInquiryStateProvider` wrap the Artwork component is that it introduces complexity to the artwork component root that it doesn't really need. Also most of, if not all, the work we were doing for first inquiry is hidden behind a lab option and this breaks that pattern which is a bigger concern for me. If we do stick with this pattern I would feel at ease with a test covering it.

src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
```diff
@@ -18,26 +24,87 @@ interface InquiryModalProps {
 
 export const InquiryModal: React.FC<InquiryModalProps> = ({ artwork, ...props }) => {
   const { toggleVisibility, modalIsVisible } = props
+  const questions = artwork?.inquiryQuestions!
```
xtina-starr, line 27: Alternative to the context provider approach we could also grabbed the `internalID` (i.e. `"price_and_availability"`, `"shipping"`, & `"condition_and_provenance"`) for the questions in metaphysics and initialize state somewhere around here with the `useState` hook.

src/lib/Scenes/Artwork/Artwork.tsx
zephraph: I'm in agreement w/ @xtina-starr here. While theoretically providers can be placed anywhere in the tree, having it as close to its intended usage goes a long way to communicate intent. Given that this state will only every be consumed from inside the modal, I'd expect it to be contained there too. 

That said, I'd be interested if @ashfurrow or @ds300 had thoughts otherwise.

src/lib/Scenes/Artwork/Artwork.tsx
```diff
@@ -288,17 +289,25 @@ export class Artwork extends React.Component<Props, State> {
...
+        <ArtworkInquiryContext.Consumer>
+          {(_context) => {
+            return (
```
zephraph, line 295: Does this block need to be wrapped in a consumer? Unless I'm missing something, I believe this should be removed. 

From the [docs](https://reactjs.org/docs/context.html#contextconsumer) this is useful for when you need to use data from the provider programmatically in the tree. It seems like this would just lead to unnecessary re-renders as it's unused.

src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
```diff
@@ -18,26 +24,87 @@ interface InquiryModalProps {
...
+  const { state } = useContext(ArtworkInquiryContext)
+  const [isExpanded, setExpanded] = useState(false)
+  const toggleExpanded = () => {
+    LayoutAnimation.configureNext({
+      ...LayoutAnimation.Presets.linear,
```
ds300, line 32: you should probably use `LayoutAnimation.Presets.easeInEaseOut ` here, since linear animations usually feel clunky

src/lib/Scenes/Artwork/Artwork.tsx
ds300: I'd agree that this should be scoped to the inquiry modal unless other elements on the artwork page need to have access to the inquiry state for some reason.

PR #4001, opened 2020-10-09
fix: conversation pagination layout fixes #trivial

jo-rs:

This PR resolves [PURCHASE-2156][PURCHASE-2157]

### Description

This PR fixes issues in layout when paginating a conversation. Bugs discovered after Purchase QA.
1-Fixes issue of messages sometimes disappearing on pagination.
2-Fixes flat-list top-margin/wrong layout.
3-Fixes issue of keyboard blocking last message if its a long message.
4-Fixes crash caused by image attachments resizing on pagination.

There's one QA bug which wasn't reproducible which for one person caused the whole screen to flicker. There will be QA conducted to make sure it's not happening any more.

[PURCHASE-2156]: https://artsyproduct.atlassian.net/browse/PURCHASE-2156
[PURCHASE-2157]: https://artsyproduct.atlassian.net/browse/PURCHASE-2157

### DIFF
--- a/src/lib/Components/Inbox/Conversations/Message.tsx
+++ b/src/lib/Components/Inbox/Conversations/Message.tsx
@@ -20,8 +20,9 @@ import { Schema, Track, track as _track } from "../../../utils/track"
 
 import { compact } from "lodash"
 
-const AttachmentContainer = styled(Flex)`
+const AttachmentContainer = styled(View)`
   border-radius: 15px;
+  padding: 10px;
 `
 
 const linkStyle = {
@@ -39,7 +40,7 @@ const track: Track<Props> = _track
 
 @track()
 export class Message extends React.Component<Props> {
-  renderAttachmentPreviews(attachments: Props["message"]["attachments"], backgroundColor: "black100" | "black10") {
+  renderAttachmentPreviews(attachments: Props["message"]["attachments"], backgroundColor: string) {
     // reactNodeHandle is passed to the native side to decide which UIView to show the
     // download progress bar on.
     const previewAttachment = (reactNodeHandle: number, attachmentID: string) => {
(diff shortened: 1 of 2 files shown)


erikdstock (commented):
This all looks sensible to me, but it's hard to fully understand the style changes or how they directly fix the bugs. If @starsirius approves then I am 👍 !

jo-rs:
> This all looks sensible to me, but it's hard to fully understand the style changes or how they directly fix the bugs. If @starsirius approves then I am 👍 !

@erikdstock There was a bit of [this issue](https://github.com/facebook/react-native/issues/15990) happening combined with a few `Flex`es that would mess with the height calculation, and cause the content to collapse. So by adding a workaround for the first issue, and replacing a few `Flex`es with a couple of `View`s the issues seem to (hopefully) be resolved. 😬

src/lib/Components/Inbox/Conversations/Message.tsx
```diff
@@ -57,11 +58,13 @@ export class Message extends React.Component<Props> {
...
-          p={1}
-          flex={1}
-          mb={index === attachments!.length - 1 ? 0 : 0.5}
+          style={{
+            backgroundColor,
+            flex: 1,
+            marginBottom: index === attachments!.length - 1 ? 0 : 5,
+            width: "100%",
+          }}
+          key={attachment.internalID}
```
sepans, line 67: ~non-blocking: curious to know if setting these as `style` was different from doing it with props?~ Got it because it is now a `View` and not `Flex`. The same should be the case for changing `backgroundColor` type to `string` 👍

sepans (approved):
💯

src/lib/Components/Inbox/Conversations/Message.tsx
jo-rs: Yes, sir! 😊

sepans:
Cool merging. Thanks @jo-rs

PR #4003, opened 2020-10-09
feat: switch to live auction view

brainbicycle:

This PR resolves [CX-516]

### Description

Switch to live auction view when a sale goes live from auction view.

- Adds a poll every .5 seconds to check if sale has gone live
- Exposes native method for popping top view controller 

Co-authored with @bhoggard 

### Screenshots

switchToLive

[CX-516]: https://artsyproduct.atlassian.net/browse/CX-516


src/lib/Scenes/Sale/Sale.tsx
```diff
@@ -25,16 +28,49 @@ interface SaleSection {
...
   const [showGrid, setShowGrid] = useState(true)
 
   const saleArtworks = extractNodes(props.sale.saleArtworksConnection)
   const scrollAnim = useRef(new Animated.Value(0)).current
 
+  let intervalId: NodeJS.Timeout
+
+  useEffect(() => {
+    if (props.sale.liveStartAt) {
+      intervalId = setInterval(checkIfSaleIsLive, 500)
```
ashfurrow, line 41: I would have thought this would be `setTimeout`, so it might be worth adding a comment here clarifying.

src/lib/Scenes/Sale/Sale.tsx
```diff
@@ -25,16 +28,49 @@ interface SaleSection {
...
+        switchToLive()
+      }
+    }
+  }
+
   const switchView = (value: boolean) => {
     setShowGrid(value)
   }
 
+  const liveAuctionUrl = (slug: string): string => {
```
ashfurrow, line 62: If we only reference this function once, maybe we can inline this in `switchToLive()`?

src/lib/Scenes/Sale/__tests__/Sale-tests.tsx
```diff
@@ -0,0 +1,81 @@
...
+          liveStartAt: inHalfSecond.toISOString(),
+          internalID: "the-sale-internal",
+        }),
+      })
+    )
+
+    setTimeout(() => {
+      expect(navigate).toHaveBeenCalledWith("https://live-staging.artsy.net/live-sale-slug")
+      done()
+    }, 1000)
```
ashfurrow, line 79: I always worry about timeouts like this in unit tests, which should be as fast as possible. But it looks like this is a standard approach in our codebase already. Could we decrease the intervals a bit to keep the tests fast? It's a balance between what is a reliable test, and what is a fast test. If we could decrease these down by ten percent, and it works, let's do that and monitor CI and developers for intermittent failures.

ashfurrow (commented):
The approach looks good! I just had some questions about comments and unit testing approaches.

src/lib/Scenes/Sale/Sale.tsx
brainbicycle: `setInterval` will repeat indefinitely whereas `setTimeout` will only get executed once. I can add a comment that we are essentially polling to see if sale is live.

src/lib/Scenes/Sale/Sale.tsx
brainbicycle: 👍

src/lib/Scenes/Sale/__tests__/Sale-tests.tsx
brainbicycle: I agree it's not ideal, I think I might be able to do this synchronously or at least faster with https://jestjs.io/docs/en/timer-mocks 
Barry and I tried for a bit and had issues but I will revisit and see if I can get it working

ashfurrow: It's not a huge blocker – give it a shot but don't let it block the PR from being merged 👍

brainbicycle: Done! Looks like our problems were with navigation functions not being mocked, mockTimers are actually pretty straightforward

bhoggard (approved):

MounirDhahri (approved):

PR #4012, opened 2020-10-13
Fix: [My Collection] reset form on delete; auction list click

damassi:

This PR resolves 
- https://artsyproduct.atlassian.net/browse/CX-701
- https://artsyproduct.atlassian.net/browse/CX-700

### Description

This fixes an issue where on artwork delete, the form is still populated. This is because we "load" the form on edit button click, but then don't reset the form on delete (which happens from the edit screen). When a user goes to add a new work, the form appears populated with the work that was just deleted. This fixes that. 

Also makes it so that the entire auctions section is clickable, taking the user to `/artist/id/auction-results`, and removes the "optional" text around photo addition (since we require 1 photo before a work can be submitted). 


bhoggard (approved):

brainbicycle (approved):
LGTM 👍

PR #4016, opened 2020-10-14
feat: add inactive fair state to Fair2

anandaroop:

This PR resolves [FX-2347]

### Description

To be consistent with the current Fair view this PR updates the Fair2 view to be pared down for inactive fairs:

Also updates handling for missing images (which maybe shouldn't happen, but I was able to navigate to this one from City Guide, e.g.)

[FX-2347]: https://artsyproduct.atlassian.net/browse/FX-2347

### DIFF
--- a/src/lib/Scenes/Fair2/Components/Fair2Header.tsx
+++ b/src/lib/Scenes/Fair2/Components/Fair2Header.tsx
@@ -5,7 +5,7 @@ import { navigate } from "lib/navigation/navigate"
 import { shouldShowFairBMWArtActivationLink } from "lib/Scenes/Fair/Screens/FairBMWArtActivation"
 import { shouldShowLocationMap } from "lib/Scenes/Fair2/Fair2MoreInfo"
 import { truncatedTextLimit } from "lib/utils/hardware"
-import { Box, ChevronIcon, Flex, Text } from "palette"
+import { Box, ChevronIcon, Flex, Spacer, Text } from "palette"
 import React from "react"
 import { Dimensions, TouchableOpacity } from "react-native"
 import { createFragmentContainer, graphql } from "react-relay"
@@ -50,7 +50,7 @@ export const Fair2Header: React.FC<Fair2HeaderProps> = ({ fair }) => {
 
   return (
     <Box>
-      {!!image && (
+      {!!image ? (
         <Flex alignItems="center" justifyContent="center" style={{ position: "relative" }}>
           <OpaqueImageView width={screenWidth} height={screenWidth / image.aspectRatio} imageURL={image.imageUrl} />
           {!!profileImageUrl && (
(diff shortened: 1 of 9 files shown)


src/lib/Scenes/Fair2/Components/Fair2Header.tsx
```diff
@@ -50,7 +50,7 @@ export const Fair2Header: React.FC<Fair2HeaderProps> = ({ fair }) => {
 
   return (
     <Box>
-      {!!image && (
+      {image ? (
```
sweir27, line 53: Should we keep the `!!`?

src/lib/Scenes/Fair2/Components/Fair2Header.tsx
```diff
@@ -89,6 +91,8 @@ export const Fair2Header: React.FC<Fair2HeaderProps> = ({ fair }) => {
   )
 }
 
+const SafeTopMargin = () => <Spacer mt={6} pt={2} />
```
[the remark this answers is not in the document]
sweir27: I still feel like one of these values should be:
```
const { safeAreaInsets } = useScreenDimensions()
// safeAreaInsets.top
```

But would be worth getting an MX friend to answer that.

src/lib/Scenes/Fair2/Fair2.tsx
```diff
@@ -56,13 +57,15 @@ export const Fair2: React.FC<Fair2Props> = ({ fair }) => {
...
-  ]
+  const sections = isActive
+    ? [
+        "fairHeader",
+        ...(hasArticles ? ["fairEditorial"] : []),
+        ...(hasCollections ? ["fairCollections"] : []),
+        ...(hasFollowedArtistArtworks ? ["fairFollowedArtistsRail"] : []),
+        ...(hasArtworks && hasExhibitors ? ["fairTabs", "fairTabChildContent"] : []),
+      ]
+    : ["fairHeader", "notActive"]
```
sweir27, line 68: Based on the description in the ticket, I believe we also want to include the `fairEditorial` section if it has content + the fair is not yet open.

src/lib/Scenes/Fair2/Components/Fair2Header.tsx
anandaroop: `image` is already truthy if present? But I don't feel strongly, can add back (I don't think this was intentional 😃 )

src/lib/Scenes/Fair2/Fair2.tsx
anandaroop: Ah good catch about editorial, lemme update...

anandaroop: Updated to include Editorial, e.g.

sweir27:
We can follow up with the top-padding/margin question if we want! I don't think it blocks this PR.

PR #4020, opened 2020-10-14
fix: fallback to local photo when processing

brainbicycle:

This PR resolves [CX-703]

### Description

Use local path to display photo in users collection list while photo is processing. 

- Adds an additional lastPhotoUploaded field to the session state
- When rendering artwork list view use this as a backup if imageURL is not present

Note: I  initially tried to use relay's optimisticResponse for this but ultimately decided against it because it requires recreating an entire fake artwork model object and we only really need the local path to the last uploaded photo. If there is a better way please let me know.

Co-authored with @bhoggard
Co-authored with @damassi 

#### Before

photoStateBefore

#### After 
photoStateAfter

[CX-703]: https://artsyproduct.atlassian.net/browse/CX-703

### DIFF
--- a/src/lib/Scenes/MyCollection/Screens/ArtworkList/MyCollectionArtworkListItem.tsx
+++ b/src/lib/Scenes/MyCollection/Screens/ArtworkList/MyCollectionArtworkListItem.tsx
@@ -6,6 +6,7 @@ import { useScreenDimensions } from "lib/utils/useScreenDimensions"
 import { capitalize } from "lodash"
 import { Box, color, Flex, Sans } from "palette"
 import React from "react"
+import { Image as RNImage } from "react-native"
 import { createFragmentContainer, graphql } from "react-relay"
 import styled from "styled-components/native"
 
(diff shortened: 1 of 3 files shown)


src/lib/Scenes/MyCollection/State/MyCollectionArtworkModel.tsx
```diff
@@ -187,8 +188,10 @@ export const MyCollectionArtworkModel: MyCollectionArtworkModel = {
     )
   }),
 
-  uploadPhotos: thunk(async (actions, photos) => {
+  uploadPhotos: thunk(async (actions, photos, { getState }) => {
     try {
+      const state = getState()
+      state.sessionState.lastUploadedPhoto = photos[0]
```
damassi, line 194: This is much simpler! Nice 👍 

You'll want to add 
```
state.sessionState.lastUploadedPhoto = null
```
to the `resetForm` action as well, which will handle cleanup.

src/lib/Scenes/MyCollection/Screens/ArtworkList/MyCollectionArtworkListItem.tsx
```diff
@@ -24,12 +25,17 @@ const MyCollectionArtworkListItem: React.FC<MyCollectionArtworkListItemProps> =
 
   const { artist, artistNames, medium, slug, title } = artwork
 
-  const Image = () =>
-    !!imageURL ? (
-      <OpaqueImageView imageURL={imageURL.replace(":version", "square")} width={90} height={90} />
-    ) : (
-      <Box bg={color("black30")} width={90} height={90} />
-    )
+  const lastUploadedPhoto = AppStore.useAppState((state) => state.myCollection.artwork.sessionState.lastUploadedPhoto)
```
damassi, line 28: Can we update the associated test file with a test around this behavior as well?

damassi (approved):
Two minor requests, otherwise LGTM 👍

ds300 (commented):
I noticed it's possible to upload an artwork without a photo. In that situation will it still try to show the `lastUploadedPhoto` which could be pointing to a different image?

src/lib/Scenes/MyCollection/State/MyCollectionArtworkModel.tsx
brainbicycle: I don't think I can do it there because that reset will happen in the `addArtworkComplete` action which will occur before the artwork list is displayed. Maybe I can clean it up after it is displayed the first time and assume after that we will get a url for image with completed processing, since fallback is the gray box think that should be okay.

ds300 (approved):
LGTM, feel free to merge when ready 🙇

PR #4026, opened 2020-10-16
fix: vanity url fallback handling

ds300:
### Description

When #3867 was merged while I was away, it broke the `VanityURLEntity` fallback handling which depended on the old `principalField` behaviour. Here I added a special `renderFallback` option to our `renderWithPlaceholder` utility so that when metaphysics errors out we can show something other than the `LoadFailureView`, in this case we show our `VanityURLPossibleRedirect` fallback component.


brainbicycle (approved):
👍 Looks good to me, good catch!

PR #4031, opened 2020-10-16
fix: Use aggregations for all fair artworks in filter options

dblandin:
This PR resolves [FX-2358]

### Description

When a collector views "Artworks by Artists You Follow", the filter options become pinned to the aggregations calculated for that scoped set. Even if the toggle for "by Artists you Follow" is disabled, the aggregations are not recalculated and still pinned to that initial set of artworks.

This commit adds a query for the aggregations of the full artwork set in the fair. This does result in a second request to our API.

When pairing, we touched on conditionally including this field with a GraphQL directive so that it's not always triggered or optimizing some other way.

[FX-2358]: https://artsyproduct.atlassian.net/browse/FX-2358

### DIFF
--- a/src/__generated__/Fair2AllFollowedArtistsQuery.graphql.ts
+++ b/src/__generated__/Fair2AllFollowedArtistsQuery.graphql.ts
@@ -1,7 +1,7 @@
 /* tslint:disable */
 /* eslint-disable */
 // @ts-nocheck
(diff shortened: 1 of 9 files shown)


dblandin:
Closing in favor of https://github.com/artsy/eigen/pull/4031.

PR #4033, opened 2020-10-19
Ensure the modal has the correct title based on the CTA clicked

lilyfromseattle:
# [PURCHASE-2147]

Updates modal header text to match the call to action.

__Before:__
Screen Shot 2020-10-19 at 12 45 42 PM

__After:__
Screen Shot 2020-10-19 at 12 24 32 PM
Screen Shot 2020-10-19 at 12 24 43 PM
Screen Shot 2020-10-19 at 12 25 38 PM

[PURCHASE-2147]: https://artsyproduct.atlassian.net/browse/PURCHASE-2147

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -13,6 +13,7 @@ upcoming:
     - Handle artist page subroutes natively by default - david
     - Fix Vanity URL fallbacks - david
     - Adds Show2 basic more information route, behind lab option - damon
+    - Adds header text to new inquiry modal - lily
 
 releases:
   - version: 6.6.6
--- a/src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryButtons.tsx
+++ b/src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryButtons.tsx
@@ -20,21 +20,28 @@ export interface InquiryButtonsState {
 
 export const InquiryButtons: React.FC<InquiryButtonsProps> = ({ artwork, ...props }) => {
   const [modalVisibility, setModalVisibility] = useState(false)
+  const [modalHeaderText, setModalHeaderText] = useState("Contact gallery")
+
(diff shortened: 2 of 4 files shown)


ashleyjelks:
You can leverage the global state object value of `inquiryType` and pass that value in as the modal header directly rather than passing state around. It's maybe a 1-2 line code change in `src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx` 

```
<FancyModalHeader leftButtonText="Cancel" onLeftButtonPress={() => toggleVisibility()}>
        {state.inquiryType}
      </FancyModalHeader>
```

lilyfromseattle:
> You can leverage the global state object value of `inquiryType` and pass that value in as the modal header directly rather than passing state around. It's maybe a 1-2 line code change in `src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx`
> 
> ```
> <FancyModalHeader leftButtonText="Cancel" onLeftButtonPress={() => toggleVisibility()}>
>   {state.inquiryType}
> </FancyModalHeader>
> ```

The modal header copy and casing in the designs is slightly different from the `inquiryType`s ("Inquire on price" vs "Request Price" etc) but I could change the logic to use `inquiryType` and decide on the title within the modal component if you think that would be better

zephraph:
Actually, @lilyfromseattle that's a good point. 

In talking to Will before he left, the intent was to have the text the same in the buttons as is in the header. 

Looking back at the designs, the text in [InquiryOptions](https://github.com/artsy/eigen/blob/9ce4fe02ef774fa63e10ea4fd4b119da9a07fac7/src/lib/utils/ArtworkInquiry/ArtworkInquiryTypes.ts#L28-L30) needs to be updated.

So, I'd recommend taking @ashleyjelks' advice here and using the modal state to set the header, but also to update the text in InquiryOptions to match the specs (i.e. fix casing, `Request Price` -> `Inquire on price`)

PR #4035, opened 2020-10-20
enhancement: follow up on first inquiry functionality

zephraph:

### Description

This PR follows up on some of the first inquiry work to make some minor updates.

There's only a few things here:

1. Moves the state provider down to be closer to it's usage in the tree
2. 

### DIFF
--- a/src/lib/Scenes/Artwork/Artwork.tsx
+++ b/src/lib/Scenes/Artwork/Artwork.tsx
@@ -10,7 +10,6 @@ import { defaultEnvironment } from "lib/relay/createEnvironment"
 import { ArtistSeriesMoreSeriesFragmentContainer as ArtistSeriesMoreSeries } from "lib/Scenes/ArtistSeries/ArtistSeriesMoreSeries"
 import { getCurrentEmissionState } from "lib/store/AppStore"
 import { AboveTheFoldQueryRenderer } from "lib/utils/AboveTheFoldQueryRenderer"
-import { ArtworkInquiryContext, ArtworkInquiryStateProvider } from "lib/utils/ArtworkInquiry/ArtworkInquiryStore"
 import {
   PlaceholderBox,
   PlaceholderRaggedText,
@@ -289,25 +288,17 @@ export class Artwork extends React.Component<Props, State> {
 
   render() {
     return (
-      <ArtworkInquiryStateProvider>
-        <ArtworkInquiryContext.Consumer>
-          {(_context) => {
-            return (
-              <FlatList<ArtworkPageSection>
-                data={this.sections()}
-                ItemSeparatorComponent={() => (
-                  <Box mx={2} my={3}>
-                    <Separator />
-                  </Box>
-                )}
-                refreshControl={<RefreshControl refreshing={this.state.refreshing} onRefresh={this.onRefresh} />}
-                contentContainerStyle={{ paddingBottom: 40 }}
-                renderItem={({ item }) => (item.excludePadding ? item.element : <Box px={2}>{item.element}</Box>)}
-              />
-            )
-          }}
-        </ArtworkInquiryContext.Consumer>
-      </ArtworkInquiryStateProvider>
+      <FlatList<ArtworkPageSection>
+        data={this.sections()}
+        ItemSeparatorComponent={() => (
+          <Box mx={2} my={3}>
+            <Separator />
+          </Box>
+        )}
+        refreshControl={<RefreshControl refreshing={this.state.refreshing} onRefresh={this.onRefresh} />}
+        contentContainerStyle={{ paddingBottom: 40 }}
+        renderItem={({ item }) => (item.excludePadding ? item.element : <Box px={2}>{item.element}</Box>)}
+      />
     )
   }
 }
--- a/src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryButtons.tsx
+++ b/src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryButtons.tsx
@@ -17,7 +17,7 @@ export interface InquiryButtonsState {
   modalIsVisible: boolean
 }
 
-export const InquiryButtons: React.FC<InquiryButtonsProps> = ({ artwork, ...props }) => {
+const InquiryButtons: React.FC<InquiryButtonsProps> = ({ artwork, ...props }) => {
   const [modalVisibility, setModalVisibility] = useState(false)
   const { dispatch } = useContext(ArtworkInquiryContext)
   const dispatchAction = (buttonText: string) => {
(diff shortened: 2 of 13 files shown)


src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryButtons.tsx
```diff
@@ -31,7 +30,7 @@ export const InquiryButtons: React.FC<InquiryButtonsProps> = ({ artwork, ...prop
   }
 
   return (
-    <>
+    <ArtworkInquiryStateProvider>
```
zephraph, line 33: Given that `useContext` is referenced above, we may actually need to hoist this one more level higher.

src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
```diff
@@ -38,23 +38,23 @@ export const InquiryModal: React.FC<InquiryModalProps> = ({ artwork, ...props })
   const [shippingModalVisibility, setShippingModalVisibility] = useState(false)
   const selectShippingLocation = (l: string) => dispatch({ type: "selectShippingLocation", payload: l })
 
-  const renderInquiryQuestion = (inquiry: string): JSX.Element => {
+  const renderInquiryQuestion = (id: string, inquiryQuestion: string): JSX.Element => {
     // Shipping requires special logic to accomodate dropdown and shipping modal
-    const isShipping = inquiry === InquiryOptions.Shipping
+    const isShipping = id === InquiryQuestionIDs.Shipping
```
zephraph, line 43: This is a bit of a subtle distinction, but the question text is controlled by gravity and considered variable over time. To preserve backwards compatibility, we're relying on ids which we're considering more stable.

src/lib/Scenes/Artwork/Components/CommercialButtons/__tests__/helpers.ts
```diff
@@ -22,12 +22,12 @@ export const press = (
   })
   const touchable = touchables[0]
   if (touchable && touchable.props.onPress) {
-    return act(async () => {
```
zephraph, line 25: cc @erikdstock everything that happens inside of act is supposed to be synchronous. Typescript was failing w/ an error here on the precommit hook. 

Hopefully this re-arrangement works relatively the same way.

src/lib/Scenes/Artwork/Components/CommercialButtons/__tests__/helpers.ts
```diff
@@ -22,12 +22,12 @@ export const press = (
...
+    act(() => {
       touchable.props.onPress()
-      await flushPromiseQueue()
     })
+    return flushPromiseQueue()
   } else {
-    return Promise.resolve
+    return Promise.resolve()
```
zephraph, line 30: Also, not sure if this was intentional, but it didn't seem so. If it was, I can change it back.

zephraph:
I've got to follow up and do some more testing. Having some precompiled header issues.

src/lib/Scenes/Artwork/Components/CommercialButtons/__tests__/helpers.ts
erikdstock: I'm okay with whatever is most simple and keeping promises out of `act` seems more simple, but I upgraded `react-test-renderer` and its [types](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-test-renderer/index.d.ts#L70) so promises should be allowed now (otherwise we wouldn't have been able to push/lint/merge #3980). This is probably better though.

xtina-starr (approved):
This looks good to me  🎉

src/lib/Scenes/Artwork/Artwork.tsx
```diff
@@ -289,25 +288,17 @@ export class Artwork extends React.Component<Props, State> {
...
-                )}
-                refreshControl={<RefreshControl refreshing={this.state.refreshing} onRefresh={this.onRefresh} />}
-                contentContainerStyle={{ paddingBottom: 40 }}
-                renderItem={({ item }) => (item.excludePadding ? item.element : <Box px={2}>{item.element}</Box>)}
-              />
-            )
-          }}
-        </ArtworkInquiryContext.Consumer>
-      </ArtworkInquiryStateProvider>
+      <FlatList<ArtworkPageSection>
```
erikdstock, line 291: Is this just removing the provider/consumer?

src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
```diff
@@ -38,23 +38,23 @@ export const InquiryModal: React.FC<InquiryModalProps> = ({ artwork, ...props })
...
+      <React.Fragment key={inquiryQuestion}>
         <InquiryField>
           <Flex flexDirection="row" justifyContent="space-between">
             <Flex flexDirection="row">
               <Checkbox
-                data-test-id={`checkbox-${inquiry}`}
+                data-test-id={`checkbox-${inquiryQuestion}`}
                 checked={
-                  state.inquiryType === InquiryOptions.RequestPrice && inquiry === InquiryOptions.PriceAvailability
+                  state.inquiryType === InquiryOptions.RequestPrice && id === InquiryQuestionIDs.PriceAndAvailability
```
erikdstock, line 53: i feel like I already noticed this and maybe it wasn't an issue so I didn't end up changing it, but shouldn't this property be general rather than specific to `RequestPrice`? I know the checkboxes are working correctly visually so if this question sounds right maybe there is something else going on with that component.

src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
```diff
@@ -105,10 +105,13 @@ export const InquiryModal: React.FC<InquiryModalProps> = ({ artwork, ...props })
...
-          // @ts-ignore
           // NOTE: For now the inquiryQuestions field values are always present and therefore never null, so it is safe to destructure them
-          questions!.map(({ question }: string) => {
-            return renderInquiryQuestion(question)
+          questions!.map((inquiryQuestion) => {
+            if (!inquiryQuestion) {
+              return false
+            }
+            const { internalID: id, question } = inquiryQuestion
+            return renderInquiryQuestion(id, question)
```
erikdstock, line 114: :+1:

erikdstock (commented):
all makes sense here

src/lib/Scenes/Artwork/Artwork.tsx
zephraph: Well, we _did_ remove the consumer here because it was unnecessary. More specifically though, the provder is moved closer to where it's used (i.e. within the inquiry buttons component).

src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
zephraph: That's a very valid point. I'll follow up on this.

zephraph:
I'm going to move forward w/ this. There's one small update that I'll follow up w/ in addressing the prop provided to the checkbox, but I'll do so in a separate PR.

PR #4038, opened 2020-10-20
fix(city guide): fix cta images not showing up

pvinis:

This PR resolves [CX-693]

### Description

After @admbtlr's suggestion, upgrading RN to the latest fixed the image issue 🙏.

I'd like to test the tipsi-stripe thing, maybe we can do this in QA? I saw that in the Podfile.lock it used to point to a specific commit, but now it's not. Of course, as always, after an upgrade, we should do our normal QA anyway.

Before
cityGuideiOS14

After
Screenshot 2020-11-09 at 11 49 26

#known 

[CX-693]: https://artsyproduct.atlassian.net/browse/CX-693


Podfile
```diff
@@ -68,56 +75,6 @@ target 'Artsy' do
 
   pod 'Emission', path: './emission'
 
-  pod 'React', path: 'node_modules/react-native'
-  if ENV['CIRCLE_BUILD_NUM']
-    pod 'React-Core', path: 'node_modules/react-native'
-  else
-    pod 'React-Core', path: 'node_modules/react-native', subspecs: ['DevSupport']
```
ashfurrow, line 75: We should make sure that the beta builds don't ship with this enabled 😅

pvinis:
This should be ready now 👍.

ashfurrow (approved):
Looks good! Just one note about testing the next beta. I think we can merge when ready 👍

Podfile
pvinis: mm I'll check this, one sec.

pvinis: I'm seeing this: https://github.com/facebook/react-native/blob/d03c0f965ed26c7c8a5557ab183396b64d25a9a1/scripts/react_native_pods.rb#L14-L15

ashfurrow: Okay, so it seems we can move this to when we invoke `use_react_native!`?

```ruby
use_react_native! production:  ENV['CIRCLE_BUILD_NUM']
```

pvinis: Yep! Took me a while, cause node_modules broke again 😅. Let's see if that makes ci happy.

pvinis:
This should be good to go.

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>Developer Specific file shouldn't be changed, you can skip by adding #known to the PR body and re-running CI</td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

pvinis:
Oh no.. this had the squashongreen and I rebased it on top of code 12 to make it green 😬. #4117 is the one. Since it's green and approved too, I'll do a tiny change and we are good.

PR #4041, opened 2020-10-21
fix(auctions): Closed auctions fixes

pvinis:

This PR resolves [CX-725], [CX-728]

### Description

This includes two fixes:
- when a closed auction is accessed, it doesn't try to go live anymore.
- when a closed auction is accessed, the register button is not displayed.

[CX-725]: https://artsyproduct.atlassian.net/browse/CX-725

[CX-728]: https://artsyproduct.atlassian.net/browse/CX-728

### DIFF
--- a/src/lib/Scenes/Sale/Sale.tsx
+++ b/src/lib/Scenes/Sale/Sale.tsx
@@ -114,14 +114,16 @@ export const Sale: React.FC<Props> = ({ queryRes }) => {
       key: "header",
       content: <SaleHeader sale={sale} scrollAnim={scrollAnim} />,
     },
-    (sale.endAt === null || moment().isBefore(sale.endAt)) && {
-      key: "registerToBid",
-      content: (
-        <Flex mx="2" mt={2}>
-          <RegisterToBidButton sale={sale} />
-        </Flex>
-      ),
-    },
+    (sale.endAt === null || moment().isBefore(sale.endAt)) &&
+      sale.registrationEndsAt !== null &&
+      moment().isBefore(sale.registrationEndsAt) && {
+        key: "registerToBid",
+        content: (
+          <Flex mx="2" mt={2}>
+            <RegisterToBidButton sale={sale} />
+          </Flex>
+        ),
+      },
     {
       key: "saleArtworksRail",
       content: <SaleArtworksRailContainer me={me} />,
@@ -187,6 +189,7 @@ export const SaleQueryRenderer: React.FC<{ saleID: string }> = ({ saleID }) => {
             endAt
             ...SaleHeader_sale
             ...RegisterToBidButton_sale
+            registrationEndsAt
           }
           me {
             ...SaleArtworksRail_me
--- a/src/lib/Scenes/Sale/Sale.tsx
+++ b/src/lib/Scenes/Sale/Sale.tsx
@@ -59,9 +59,24 @@ export const Sale: React.FC<Props> = ({ queryRes }) => {
   // poll every .5 seconds to check if sale has gone live
   useInterval(() => {
     const now = moment()
-    setIsLive(
-      sale.liveStartAt !== null && now.isAfter(sale.liveStartAt) && sale.endAt !== null && now.isBefore(sale.endAt)
-    )
+    if (sale.liveStartAt === null) {
+      setIsLive(false)
+      return
+    }
+    if (now.isAfter(sale.liveStartAt)) {
+      if (sale.endAt === null) {
+        setIsLive(true)
+        return
+      }
+      if (now.isBefore(sale.endAt)) {
+        setIsLive(true)
+        return
+      }
+      setIsLive(false)
+      return
+    }
+    setIsLive(false)
+    return
   }, 500)
 
   useEffect(() => {
--- a/src/lib/Scenes/Sale/Sale.tsx
+++ b/src/lib/Scenes/Sale/Sale.tsx
@@ -8,6 +8,7 @@ import { defaultEnvironment } from "lib/relay/createEnvironment"
 import { getCurrentEmissionState } from "lib/store/AppStore"
 import { ArtworkFilterContext, ArtworkFilterGlobalStateProvider } from "lib/utils/ArtworkFilter/ArtworkFiltersStore"
 import { Schema } from "lib/utils/track"
+import _ from "lodash"
 import moment from "moment"
 import { Flex } from "palette"
 import React, { useEffect, useRef, useState } from "react"
@@ -112,12 +113,12 @@ export const Sale: React.FC<Props> = ({ queryRes }) => {
     setTimeout(popParentViewController, 500)
   }
 
-  const saleSectionsData: SaleSection[] = [
+  const saleSectionsData: SaleSection[] = _.compact([
     {
       key: "header",
       content: <SaleHeader sale={sale} scrollAnim={scrollAnim} />,
     },
-    {
+    (sale.endAt === null || moment().isBefore(sale.endAt)) && {
       key: "registerToBid",
       content: (
         <Flex mx="2" mt={2}>
@@ -133,7 +134,7 @@ export const Sale: React.FC<Props> = ({ queryRes }) => {
       key: "saleLotsList",
       content: <SaleLotsListContainer saleArtworksConnection={queryRes} saleID={sale.slug} saleSlug={sale.slug} />,
     },
-  ]
+  ])
 
   return (
     <ArtworkFilterGlobalStateProvider>
--- a/src/lib/Scenes/Sale/Sale.tsx
+++ b/src/lib/Scenes/Sale/Sale.tsx
@@ -58,11 +58,11 @@ export const Sale: React.FC<Props> = ({ queryRes }) => {
 
   // poll every .5 seconds to check if sale has gone live
   useInterval(() => {
-    const now = moment()
     if (sale.liveStartAt === null) {
       setIsLive(false)
       return
     }
+    const now = moment()
     if (now.isAfter(sale.liveStartAt)) {
       if (sale.endAt === null) {
         setIsLive(true)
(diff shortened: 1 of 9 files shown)


src/lib/Scenes/Sale/Sale.tsx
```diff
@@ -48,29 +51,30 @@ export const Sale: React.FC<Props> = ({ queryRes }) => {
...
+  const prevIsLive = usePrevious(isLive, false)
 
   const scrollAnim = useRef(new Animated.Value(0)).current
 
-  let intervalId: NodeJS.Timeout
+  // poll every .5 seconds to check if sale has gone live
+  useInterval(() => {
+    const now = moment()
+    setIsLive(
+      sale.liveStartAt !== null && now.isAfter(sale.liveStartAt) && sale.endAt !== null && now.isBefore(sale.endAt)
```
MounirDhahri, line 63: I think we should also make this check if the sale `endAt` is null and we're after the `liveStartAt`

src/lib/utils/usePrevious.ts
```diff
@@ -0,0 +1,11 @@
+import { useEffect, useRef } from "react"
+
+export const usePrevious = <T>(value: T, initialValue: T) => {
+  const ref = useRef<T>(initialValue)
+
+  useEffect(() => {
+    ref.current = value
+  }, [value])
```
MounirDhahri, line 8: Since we have value in the dependencies array I guess it's no longer the previous value right? it will be updated once the isLive value is updated and the screen is mounted

MounirDhahri: **thought:** I guess you are trying to get the new `isLive` value before the screen is mounted, if that's the case I think you should move the reference and the the `useEffect`to the sale screen

MounirDhahri:
Thanks for fixing these fast 🙏, I left few minor **non-blocking** comments. But I think we should also add a test for `when a closed auction is accessed, it doesn't try to go live anymore.` to avoid this regression in the future

MounirDhahri:
Thanks for fixing these fast 🙏 good job. I left few minor non-blocking comments. But I think we should also add a test for _when a closed auction is accessed, it doesn't try to go live anymore._ to avoid this regression in the future

src/lib/Scenes/Sale/Sale.tsx
pvinis: 🤔 makes sense.

pvinis:
Yup, I'll add that too.

src/lib/Scenes/Sale/__tests__/Sale-tests.tsx
```diff
@@ -78,22 +71,79 @@ describe("Sale", () => {
...
+          slug: "closed-sale-slug",
+          startAt: moment().subtract(3, "days").toISOString(),
+          liveStartAt: moment().subtract(2, "days").toISOString(),
+          endAt: moment().subtract(1, "day").toISOString(),
+          name: "closed sale!",
+        }),
+      })
+    )
+
+    expect(tree.findAllByType(RegisterToBidButton)).toHaveLength(0)
```
brainbicycle, line 147: 🎉

src/lib/Scenes/Sale/Sale.tsx
```diff
@@ -48,29 +51,30 @@ export const Sale: React.FC<Props> = ({ queryRes }) => {
...
+    )
+  }, 500)
```
brainbicycle, line 65: nice new useInterval is cleaner ✨

brainbicycle (approved):

src/lib/utils/usePrevious.ts
pvinis: Mm I'm not sure what you mean. I used https://usehooks.com/usePrevious/ and modified it to have an initial value. The return is happening before the useEffect, so the usePrevious hook always gives us the previous value. What I use it for is to know when `isLive` is `true` and `prevIsLive` is `false`. That's the only case where we should navigate to the live screen. In any other case we do nothing.

pvinis:
This should be ready. I'll add @admbtlr as a reviewer because I changed stuff in the `saleTime`. I tried to make types strict, and ended up with a couple of edge cases that were missing. I talked with Sam for how to handle these, changed the tests to reflect that, and then changed the code to make them all pass.

src/lib/Scenes/Sale/Sale.tsx
```diff
@@ -8,6 +8,9 @@ import { defaultEnvironment } from "lib/relay/createEnvironment"
 import { getCurrentEmissionState } from "lib/store/AppStore"
 import { ArtworkFilterContext, ArtworkFilterGlobalStateProvider } from "lib/utils/ArtworkFilter/ArtworkFiltersStore"
 import { Schema } from "lib/utils/track"
+import { useInterval } from "lib/utils/useInterval"
+import { usePrevious } from "lib/utils/usePrevious"
+import _ from "lodash"
```
admbtlr, line 13: Shouldn't this rather be 
```
import compact from 'lodash/compact'
```
in order to avoid importing the whole caboodle?

src/lib/Scenes/Sale/Sale.tsx
```diff
@@ -48,29 +51,45 @@ export const Sale: React.FC<Props> = ({ queryRes }) => {
...
-    if (sale.liveStartAt) {
-      // poll every .5 seconds to check if sale has gone live
-      intervalId = setInterval(checkIfSaleIsLive, 500)
-      return () => {
-        clearInterval(intervalId)
-      }
+  // poll every .5 seconds to check if sale has gone live
+  useInterval(() => {
+    const now = moment()
+    if (sale.liveStartAt === null) {
```
admbtlr, line 62: If it's null, then we don't need to do the `useInterval` in the first place, do we? Unless we're expecting that value to change from null at some point?

src/lib/Scenes/Sale/helpers/__tests__/auctionTime-tests.ts
```diff
@@ -1,105 +1,227 @@
-// import moment from "moment-timezone"
-
-const TIME_ZONE_NY = "America/New_York"
-const TIME_ZONE_LA = "America/Los_Angeles"
-const TIME_ZONE_DE = "Europe/Berlin"
+const timezones = {
```
admbtlr, line 1: Nice work on these tests!

src/lib/Scenes/Sale/helpers/saleTime.ts
```diff
@@ -1,68 +1,122 @@
...
+  if (startDateMoment !== null && now.isBefore(startDateMoment)) {
+    return begins(startDateMoment, userTimeZone, saleType)
+  }
+
+  // definitely already closed
+  if (endDateMoment !== null && now.isAfter(endDateMoment)) {
+    return closed(endDateMoment)
+  }
+
+  // if we have both start and end and we in the middle
```
admbtlr, line 45:
```suggestion
  // if we have both start and end and we're in between them
```

src/lib/utils/usePrevious.ts
```diff
@@ -0,0 +1,11 @@
+import { useEffect, useRef } from "react"
```
admbtlr, line 1: Do we need this util? This is a known pattern, and hiding it in a 6 line library function is kinda obfuscating

admbtlr (changes requested):

src/lib/Scenes/Sale/Sale.tsx
pvinis: Nothing really changes in RN. We are not web (thankfully, in this case 😅). In most files we import like that, and since there is no change in perf or whatever, why not keep it consistent and simple. 👍

src/lib/Scenes/Sale/helpers/saleTime.ts
pvinis: 🇬🇧

src/lib/utils/usePrevious.ts
pvinis: 🤔 I'm not sure I understand 😬. What do you mean?

The `usePrevious` is kind of a nice and helpful hook. Apparently we are not using it anywhere in eigen already, so I thought I'll add this here.

src/lib/Scenes/Sale/Sale.tsx
pvinis: Indeed. That was one of my thoughts against using useInterval. But then again when I saw the useEffect with an if and an interval and cleanup return func inside that if, I felt confused and a bit uncertain of how that works, and worse, if/how that could break.
So I decided to flip that and run the interval outside the if, and have the if check inside.

This means the interval now will run even if it's not a live auction and the code is simpler. I felt I'm ok with that tradeoff.

pvinis, line 62:
```suggestion
    if (sale.liveStartAt === null) {
      const now = moment()
```
I'll quickly do this. This feels slightly nicer.

src/lib/Scenes/Sale/Sale.tsx
```diff
@@ -48,29 +51,45 @@ export const Sale: React.FC<Props> = ({ queryRes }) => {
...
+      setIsLive(false)
+      return
     }
```
pvinis, line 65:
```suggestion
    if (sale.liveStartAt === null) {
      setIsLive(false)
      return
    }
    const now = moment()
```
I'll do this. This feels slightly nicer.

src/lib/utils/usePrevious.ts
MounirDhahri: Sorry, I must have misunderstood the effect then 😬

src/lib/utils/usePrevious.ts
admbtlr: Actually, you're right, creating a named hook to do this is a nice approach 👍 . (I was just thinking that it was unnecessary to create a separate hook in order to abstract away 6 lines of code)

pvinis: Aaah gotcha. Thanks for explaining. Yes, I think it's nice to have, and convenient to have it as a reusable hook.

MounirDhahri:
Thanks for adding the tests and the quick fixes 🌷

MounirDhahri (approved):

admbtlr (approved):

PR #4048, opened 2020-10-21
fix: select correct window in react-image-crop-picker when input window is present

brainbicycle:

This PR resolves [CX-711]

### Description

We do some swizzling in the AppDelegate that causes  `[[UIApplication sharedApplication] delegate] window]` to return nil. react-native-image-crop-picker uses this method to find the viewController to present the picker onto so we had a patch that grabbed the last window in the hierarchy to workaround this issue. In the My Collection flow this was returning the wrong window, specifically a UIInputWindow that manages presentation of the keyboard. This PR updates our existing patch for react-native-image-crop-picker to look for our custom window subclass instead which should return the correct window, and falls back to the last window if it can't find the subclass.

#### Before 

before-image-crop

#### After 

after-image-crop

[CX-711]: https://artsyproduct.atlassian.net/browse/CX-711

### DIFF
--- a/patches/react-native-image-crop-picker+0.32.2.patch
+++ b/patches/react-native-image-crop-picker+0.32.2.patch
@@ -1,17 +1,29 @@
 diff --git a/node_modules/react-native-image-crop-picker/ios/.DS_Store b/node_modules/react-native-image-crop-picker/ios/.DS_Store
 new file mode 100644
-index 0000000..5172429
-Binary files /dev/null and b/node_modules/react-native-image-crop-picker/ios/.DS_Store differ
+index 0000000..e69de29
 diff --git a/node_modules/react-native-image-crop-picker/ios/src/ImageCropPicker.m b/node_modules/react-native-image-crop-picker/ios/src/ImageCropPicker.m
-index 25ead1a..3b431c8 100644
+index 25ead1a..ea872d1 100644
 --- a/node_modules/react-native-image-crop-picker/ios/src/ImageCropPicker.m
 +++ b/node_modules/react-native-image-crop-picker/ios/src/ImageCropPicker.m
-@@ -124,7 +124,7 @@ - (void) setConfiguration:(NSDictionary *)options
+@@ -123,8 +123,19 @@ - (void) setConfiguration:(NSDictionary *)options
+     }
  }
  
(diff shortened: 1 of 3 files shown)


damassi:
So happy to see this fixed! Ty 👍

patches/react-native-image-crop-picker+0.32.2.patch
```diff
@@ -1,17 +1,29 @@
...
++        if ([window isKindOfClass:NSClassFromString(@"ARWindow")]) {
++            presentingWindow = window;
++        }
++    }
++    if (!presentingWindow) {
++        NSAssert(false, @"Unable to find ARWindow did it get renamed?");
++        presentingWindow = [[[UIApplication sharedApplication] windows] lastObject];
++    }
++
++    UIViewController *root = [presentingWindow rootViewController];
```
damassi, line 26: @ashfurrow - did you make some kind of similar change recently? This vaguely brings to mind another bug we encountered, but I can't remember what.

ds300 (approved):
nice one! TIL about NSClassFromString!

patches/react-native-image-crop-picker+0.32.2.patch
ashfurrow: Yeah, it was to fix this same bug actually! But that change didn't cover the case that Brian outlined, this patch is much more specific 👍

patches/react-native-image-crop-picker+0.32.2.patch
```diff
@@ -1,17 +1,29 @@
...
-+    UIViewController *root = [[[[UIApplication sharedApplication] windows] lastObject] rootViewController];
++- (UIViewController *) getRootVC {
++    UIWindow *presentingWindow;
++    for (UIWindow *window in [[UIApplication sharedApplication] windows]) {
++        if ([window isKindOfClass:NSClassFromString(@"ARWindow")]) {
++            presentingWindow = window;
++        }
++    }
++    if (!presentingWindow) {
++        NSAssert(false, @"Unable to find ARWindow did it get renamed?");
```
ashfurrow, line 22: Great work, this 👍

ashfurrow (approved):
🙌

PR #4054, opened 2020-10-22
chore: update FB SDK

jonallured:
I think this is all that's required to update us to the latest FB SDK but lmk if I am missing something! Seems trivial enough that I'm going to mark this PR as squash on green. 👍 

https://artsyproduct.atlassian.net/browse/GRO-29

/cc @artsy/grow-devs


ashfurrow (approved):
Makes sense! That is how I'd expect the update to look. /cc @brainbicycle

PR #4060, opened 2020-10-26
fix: scroll to top after applying filters

MounirDhahri:

This PR resolves [CX-733]

### Description
Instead of staying in the same scroll position on the page, we should instantly move users to the top of the list after applying any filtering/sorting

Screen_Recording_2020-10-26_at_13 10 12

[CX-733]: https://artsyproduct.atlassian.net/browse/CX-733

### DIFF
--- a/src/lib/Scenes/Sale/Sale.tsx
+++ b/src/lib/Scenes/Sale/Sale.tsx
@@ -134,7 +134,7 @@ export const Sale: React.FC<Props> = ({ queryRes }) => {
   }
 
   const scrollToTop = () => {
-    const saleLotsListIndex = saleSectionsData.map((section) => section.key).indexOf(SALE_LOTS_LIST) || 0
+    const saleLotsListIndex = saleSectionsData.findIndex((section) => section.key === SALE_LOTS_LIST)
     flatListRef.current?.scrollToIndex({ index: saleLotsListIndex, viewOffset: 50 })
   }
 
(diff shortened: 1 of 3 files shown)


MounirDhahri:
Sorry I pushed to the wrong branch 😅

src/lib/Scenes/Sale/Sale.tsx
```diff
@@ -125,32 +133,44 @@ export const Sale: React.FC<Props> = ({ queryRes }) => {
     setFilterArtworkModalVisible(false)
   }
 
+  const scrollToTop = () => {
+    const saleLotsListIndex = saleSectionsData.map((section) => section.key).indexOf(SALE_LOTS_LIST) || 0
```
ds300, line 137: nonblocking: the `|| 0` here isn't needed. `indexOf` always returns a number and `0` is the only falsey number.

Also, instead of `.map(...).indexOf(...)` you can use `.findIndex(...)`

ds300 (approved):
Nice, feel free to merge when ready!

src/lib/Scenes/Sale/Sale.tsx
MounirDhahri: about the || 0, I had it because `scrollToIndex` supports only index bigger than 0.
About the `map().indexOf()`  absolutely makes sense 👏, I will change it right away and merge

ds300:
Gonna merge this since it's blocking our QA today.

PR #4063, opened 2020-10-26
fix(fair2): sticky tabs interactions + UI

ds300:
This PR resolves [CX-556]

### Description

- Fixes the scroll jump that happened when you switch tab.

   This happened because the `InfiniteScrollArtworkGrid` component needs to know how wide it is to layout its masonry grid manually. By default it needs to wait until it has mounted to have an `onLayout` callback invoked. This causes a single frame to be rendered where the content has no height. To fix that we allow manually specifying a width so that the content can be laid out before first paint. We could also have worked around this issue in this one case by adding a `min-height: screenHeight` style, but I went for the general-purpose solution since we are likely to need it again one day.
- Hides the back button while the user scrolls down. Otherwise the back button obscures one of the sticky tabs.
- Make the sticky tab UI match what we have elsewhere in the app.
- Make the artwork grid have the standard 20px margins

#### Before

fair2-scroll-bad

#### After

fair2-scroll-good

[CX-556]: https://artsyproduct.atlassian.net/browse/CX-556

### DIFF
--- a/src/lib/Components/ArtworkGrids/InfiniteScrollArtworksGrid.tsx
+++ b/src/lib/Components/ArtworkGrids/InfiniteScrollArtworksGrid.tsx
@@ -80,6 +80,9 @@ export interface Props {
 
   /** Set as true to automatically manage the back button visibility as the user scrolls */
   hideBackButtonOnScroll?: boolean
+
+  /** To avoid layout jank, supply the width of the grid ahead of time. */
+  width?: number
 }
 
 interface PrivateProps {
@@ -106,7 +109,7 @@ class InfiniteScrollArtworksGrid extends React.Component<Props & PrivateProps, S
   }
 
   state = {
-    sectionDimension: 0,
+    sectionDimension: this.getSectionDimension(this.props.width),
     isLoading: false,
   }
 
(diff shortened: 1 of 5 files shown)


src/lib/Components/ArtworkGrids/InfiniteScrollArtworksGrid.tsx
```diff
@@ -80,6 +80,9 @@ export interface Props {
 
   /** Set as true to automatically manage the back button visibility as the user scrolls */
   hideBackButtonOnScroll?: boolean
+
+  /** To avoid layout jank, supply the width of the grid ahead of time. */
+  width?: number
```
sweir27, line 85: 👍

src/lib/Scenes/Fair2/Fair2.tsx
```diff
@@ -192,6 +193,8 @@ export const Fair2: React.FC<Fair2Props> = ({ fair }) => {
                   ItemSeparatorComponent={() => <Spacer mb={3} />}
                   keyExtractor={(_item, index) => String(index)}
                   stickyHeaderIndices={[tabIndex]}
+                  onScroll={hideBackButtonOnScroll}
```
sweir27, line 196: Nice, thanks for this!

sweir27 (approved):
Thanks for the quick fix!

non-blocking: The only thing I'm not sure on-- the UI of the sticky tabs were built to match [the spec](https://www.figma.com/file/I5sz0m16SiW7F3tsSoLRxK/Fair-Page?node-id=294%3A13886):
image

Should we consult with the design team before updating their display?

PR #4067, opened 2020-10-27
fix: add artist names to filtermodal

MounirDhahri:

This PR resolves [CX-718]

### Description
- If you filter for an artist, the top-level filter modal shows the artist-slug instead of the artist name
___
**Before the changes**
Screen_Recording_2020-10-27_at_13 05 10

**After the changes**
Screen_Recording_2020-10-27_at_13 03 43

[CX-718]: https://artsyproduct.atlassian.net/browse/CX-718

### DIFF
--- a/src/lib/utils/ArtworkFilter/FilterArtworksHelpers.ts
+++ b/src/lib/utils/ArtworkFilter/FilterArtworksHelpers.ts
@@ -6,7 +6,7 @@ import {
   FilterCounts,
   FilterType,
 } from "lib/utils/ArtworkFilter/ArtworkFiltersStore"
-import { forOwn, groupBy, sortBy } from "lodash"
+import { compact, forOwn, groupBy, sortBy } from "lodash"
 
 // General filter types and objects
 export enum FilterParamName {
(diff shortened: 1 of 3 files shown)


src/lib/utils/ArtworkFilter/__tests__/FilterArtworksHelpers-tests.tsx
```diff
@@ -504,61 +508,57 @@ describe("selectedOption", () => {
...
-            paramValue: true,
+            paramValue: ["banksy", "andy-warhol"],
             displayText: "Artists",
           },
         ]
 
-        expect(selectedOption(selectedOptions, "artistIDs")).toEqual("All artists I follow, 2 more")
+        expect(
+          selectedOption({ selectedOptions, filterScreen: "artistIDs", aggregations, filterType: "saleArtwork" })
+        ).toEqual("Andy Warhol, 1 more")
```
pvinis, line 561: 👌

src/lib/utils/ArtworkFilter/FilterArtworksHelpers.ts
```diff
@@ -184,16 +190,24 @@ export const selectedOption = (
 
     let selectedArtistNames: string[]
 
-    if (filterType === "artwork") {
+    if (filterType === "saleArtwork") {
```
pvinis, line 193: Non-blocking, just curious:
Why did you swap these around? The first block (in the if) was for `artwork`, the second (in the else) was for `saleArtwork`. Was it a specific reason why you flipped them or just happened?

MounirDhahri: Yes sure, it is because after refactoring the `selectedOption` function, I realized actually that the default value (and the one used in the 4 out of the 5 filter modals we have in the app) is `artwork` and we are only using the  `saleArtwork` filterType in the `artistIDs` context and it made sense for me to fix it (My bad 😅, I should have had it that way since the beginning).

pvinis: No no, no one's bad! I was just curious if there was any other interesting case or something. Thanks for explaining. 👍

pvinis (approved):

PR #4078, opened 2020-10-29
User can see empty state when they don't have any active bids

lilyfromseattle:
[PURCHASE-2163]

Adds empty bids state with CTA that takes user to all auctions view

__Before:__
Screen Shot 2020-10-29 at 2 28 09 PM

__After:__
Screen Shot 2020-10-29 at 2 24 47 PM
Screen Shot 2020-10-29 at 2 24 39 PM

[PURCHASE-2163]: https://artsyproduct.atlassian.net/browse/PURCHASE-2163


src/lib/Scenes/MyBids/MyBids.tsx
```diff
@@ -46,6 +47,13 @@ class MyBids extends React.Component<MyBidsProps> {
       return moment(timelySale.relevantEnd).unix()
     })
 
+    const noActiveBids = activeStandings.length === 0
```
erikdstock, line 50: minor - I know we already have a sort of jumble of intermediate values as a result of sorting our queried lots above.  I wonder if it makes sense to say (pseudocode) `{activeStandings.length === 0 ? <NoBids title="You don't have any upcoming bids" /> : <ActiveBids lotStandings={activeStandings} /> }`? This would dry up some of the repeated new logic below (It looks like everything except the main text at the top is identical) and allow use to move L38-48 into a separate component that deals only with active bids.

src/lib/Scenes/MyBids/__tests__/MyBids-tests.tsx
```diff
@@ -132,4 +132,54 @@ describe(MyBidsQueryRenderer, () => {
...
+        errors: [],
+        data: {
+          me,
+        },
+      })
+    })
+
+    expect(extractText(tree.root)).toContain("You don't have any upcoming bids.")
+    expect(extractText(tree.root)).toContain(
+      "Watch a live auction and place bids in advance or in real time, or you can bid in our curated timed auction"
```
erikdstock, line 157: :+1: I think there is a MyBids2-tests.tsx now which uses the new relay mocking tools, but tbh I'm not sure how you would mock this. I do wonder if we should prefer writing them in the newer place, though.

src/lib/Scenes/MyBids/MyBids.tsx
```diff
@@ -90,6 +122,30 @@ class MyBids extends React.Component<MyBidsProps> {
...
+                          fontWeight="normal"
+                          color="black60"
+                        >
+                          Watch a live auction and place bids in advance or in real time, or you can bid in our curated
+                          timed auction
+                        </Text>
+                        <Flex width="100%" justifyContent="center" flexDirection="row">
+                          <Button variant="primaryBlack" onPress={() => handleViewAllAuctions()}>
+                            Explore Auctions
+                          </Button>
```
erikdstock, line 145: Most of this section seems to be repeated. Maybe we can extract an empty state component?

erikdstock (commented):
Looks great! A few  minor comments, nothing I would call blocking - basically was thinking these messages might be extracted into a single component, but perhaps that's premature.

src/lib/Scenes/MyBids/MyBids.tsx
lilyfromseattle: I made `NoBids` its own component but idk if it makes sense to make `ActiveBids` its own component since its so short and I think it makes more sense for the active bids and closed bids logic to be in the same place

starsirius (approved):

PR #4085, opened 2020-11-02
fix: my collection nav bugs

brainbicycle:

This PR resolves [CX-708]

### Description

This was actually a few different bugs.
1. We were showing the native back button when entering the My Collection flow, since this flow is managed by react native navigators instead we should use react-native back buttons that pop from the react native navigators. 
2. The goBack method for our native back buttons was selecting the wrong NavigationController to pop from and the result was tapping back would first pop all viewControllers from React Native managed navigators and then pop the showing VC, this has been updated to check that the navigationController is our subclass before popping.
3. Our set up navigation call was a little too eager and was replacing navigators which still had content causing the back button and pushing vcs to stop working. Added a check to see if a navigator exists before creating a new one and a clean up call to reset navigation when we leaving the MyCollection flow. 
4. Since we are no longer showing native back button in this flow when leaving update the calls to use the `navigate` function so other flows can get the native back button. 

[CX-708]: https://artsyproduct.atlassian.net/browse/CX-708

### DIFF
--- a/src/lib/AppRegistry.tsx
+++ b/src/lib/AppRegistry.tsx
@@ -427,7 +427,7 @@ export const modules = defineModules({
   MyBids: { Component: MyBidsQueryRenderer },
   AddEditArtwork: { Component: setupMyCollectionScreen(AddEditArtwork) },
   MyCollectionArtworkDetail: { Component: setupMyCollectionScreen(MyCollectionArtworkDetail) },
-  MyCollectionArtworkList: { Component: setupMyCollectionScreen(MyCollectionArtworkList) },
+  MyCollectionArtworkList: { Component: setupMyCollectionScreen(MyCollectionArtworkList), hidesBackButton: true },
   MyProfile: { Component: MyProfileQueryRenderer, isRootViewForTabName: "profile" },
   MyProfilePayment: { Component: MyProfilePaymentQueryRenderer },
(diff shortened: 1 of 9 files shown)


src/lib/AppRegistry.tsx
```diff
@@ -427,7 +427,7 @@ export const modules = defineModules({
   MyBids: { Component: MyBidsQueryRenderer },
   AddEditArtwork: { Component: setupMyCollectionScreen(AddEditArtwork) },
   MyCollectionArtworkDetail: { Component: setupMyCollectionScreen(MyCollectionArtworkDetail) },
-  MyCollectionArtworkList: { Component: setupMyCollectionScreen(MyCollectionArtworkList) },
+  MyCollectionArtworkList: { Component: setupMyCollectionScreen(MyCollectionArtworkList), hidesBackButton: true },
```
damassi, line 430: Nice, wasn't aware of this prop

ds300 (approved):
This all makes sense to me. Is there a reason it's still marked as draft?

brainbicycle:
> This all makes sense to me. Is there a reason it's still marked as draft?

Yeah there is another bug here where other My Collection screens are replacing the navigation controllers of previous My Collection screens, going to try to wrap up today.

brainbicycle:
Closing in favor of https://github.com/artsy/eigen/pull/4102 which addresses these issues 🎉

PR #4086, opened 2020-11-03
fix(my collection): set correct keyboard types #trivial

admbtlr:

This PR resolves [CX-786]

### Description

This specifies keyboard types for the Add Artwork form:

- Dimensions (height/width/depth): numeric with decimal
- Edition (number/size): numeric
- Price paid: numeric with decimal
- Date: numeric with punctuation

The Date field is not ideal - I wonder whether we should change the label to "Year" and then have a straight numeric keyboard @joanna20 ? Or just leave it as "Date" but use a numeric keyboard?

image
image

[CX-786]: https://artsyproduct.atlassian.net/browse/CX-786

### DIFF
--- a/src/lib/Scenes/MyCollection/Screens/AddArtwork/Components/Dimensions.tsx
+++ b/src/lib/Scenes/MyCollection/Screens/AddArtwork/Components/Dimensions.tsx
@@ -22,6 +22,7 @@ export const Dimensions: React.FC = () => {
       <Flex flexDirection="row">
         <Input
           placeholder="Height"
+          keyboardType="decimal-pad"
           onChangeText={formik.handleChange("height")}
           onBlur={formik.handleBlur("height")}
           defaultValue={formik.values.height}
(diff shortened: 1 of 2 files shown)


brainbicycle (approved):
Looks good! +1 on question about date input, if we are just looking for year would be better to label as year and use numeric

joanna20:
Yes "date" should be "year" and +1 to numeric keyboard. thank you!

ds300 (approved):
:+1:

admbtlr:
I just changed the label and placeholder to `Year`, but left everything everything else as `Date` (field, test-id, etc.) since the `Artwork` model is complicated. Does that seem like a bad thing to do?

admbtlr:
@ds300 ^

ds300:
> Does that seem like a bad thing to do?

Well I think `year` would be a better name but I'm happy for you to merge this as-is if you think it would be difficult to update the name right now.

PR #4090, opened 2020-11-03
fix: fix upper limit label of demand rank #trivial

brainbicycle:

This PR resolves [CX-783]

### Description

Update upper limit label for demand index progress bar.

### Screenshots 

DemandRankUpdatedAgain

[CX-783]: https://artsyproduct.atlassian.net/browse/CX-783

### DIFF
--- a/src/lib/Scenes/MyCollection/Screens/ArtworkDetail/Components/ArtworkInsights/MyCollectionArtworkDemandIndex.tsx
+++ b/src/lib/Scenes/MyCollection/Screens/ArtworkDetail/Components/ArtworkInsights/MyCollectionArtworkDemandIndex.tsx
@@ -91,7 +91,7 @@ const DemandRankScale: React.FC<{ demandRank: number }> = ({ demandRank }) => {
       <Spacer my={0.3} />
       <Flex flexDirection="row" justifyContent="space-between">
         <Text>0.0</Text>
-        <Text>{demandRank}</Text>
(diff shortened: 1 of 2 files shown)


admbtlr:
Just wrote a comment on this ticket: I feel like the labels should actually be `0` and `10` - `0.0` and `10.0` look a bit silly to me 🤔

admbtlr:
I also realise now that (a) it should probably say _Very strong demand (>**9.0**)_, shouldn't it? (b) The text underneath that sentence contains placeholders

brainbicycle:
> I also realise now that (a) it should probably say _Very strong demand (>**9.0**)_, shouldn't it? (b) The text underneath that sentence contains placeholders

Good catch on the 0.9, I think at some point this value changed from a value between 0 and 1 to a value between 1 and 10. I will update. On the text underneath, that is to spec although I agree it looks a bit weird. https://www.figma.com/file/yDSHlbJVzBxvSUROD2bfRI/Sell-tab?node-id=2943%3A19660

samchieng:
@admbtlr good callout on the labels – lets use `0 – 10` and drop the decimal here. For the actual demand index value, I think we should just use 1 decimal place, i.e. `9.7` instead of `9.73`. If anything is above 9.9, I think we should round down to 9.9 though (10/10 feels odd in concept) for this.

admbtlr:
💯

PR #4094, opened 2020-11-03
fix(fx-2407): adjusts show2 install shots query #trivial

dzucconi:

This PR resolves [FX-2407]

[FX-2407]: https://artsyproduct.atlassian.net/browse/FX-2407

### DIFF
--- a/src/lib/Scenes/Show2/Components/Show2InstallShots.tsx
+++ b/src/lib/Scenes/Show2/Components/Show2InstallShots.tsx
@@ -56,7 +56,7 @@ export const Show2InstallShotsFragmentContainer = createFragmentContainer(Show2I
   show: graphql`
     fragment Show2InstallShots_show on Show {
       name
-      images {
+      images(default: false) {
         internalID
         caption
         src: url(version: ["larger", "large"])
--- a/src/__generated__/Show2InstallShots_show.graphql.ts
+++ b/src/__generated__/Show2InstallShots_show.graphql.ts
@@ -40,7 +40,13 @@ const node: ReaderFragment = {
     },
     {
       "alias": null,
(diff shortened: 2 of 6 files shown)


src/lib/Scenes/Show2/Components/Show2InstallShots.tsx
```diff
@@ -56,7 +56,7 @@ export const Show2InstallShotsFragmentContainer = createFragmentContainer(Show2I
   show: graphql`
     fragment Show2InstallShots_show on Show {
       name
-      images {
+      images(default: false) {
```
dblandin, line 59: Wasn't sure what this argument did so I quickly looked it up:

https://github.com/artsy/metaphysics/blob/0b4ebf1dd09e6a515f8294dd0e480c2465e65a4a/src/schema/v2/show.ts#L355-L358

Looks like it determines whether default/cover images are included in the response set which makes sense 👍

dblandin (approved):
LGTM!

src/lib/Scenes/Show2/Components/Show2InstallShots.tsx
dzucconi: Yeah, two things about this interface:

* `default` is a very bad name for this argument
* `coverImage` should be it's own distinct field probably, separate from installation views

It's probably a decent idea to deprecate the `images` field and build `installationViews` and `coverImage` as their own fields.

dblandin: Agreed on both points!

Deprecating the existing field and moving to separate interfaces sounds good. Maybe `coverImage` and `installationImages`? We'd still be modeling images, right?

PR #4100, opened 2020-11-04
fix(fx-2408): improves guards for rendering hours and locations #trivial

dzucconi:

This PR resolves [FX-2408]

### Description

I think this pattern of rendering is very brittle because of the implicit linking between the parent context and the rendered children which may have conditionals that drift. I don't really have any good answers for this. I could imagine some kind of interface where a component contains its own guard for rendering. For now this just syncs them up.

[FX-2408]: https://artsyproduct.atlassian.net/browse/FX-2408

### DIFF
--- a/src/__generated__/Show2MoreInfoQuery.graphql.ts
+++ b/src/__generated__/Show2MoreInfoQuery.graphql.ts
@@ -1,7 +1,7 @@
 /* tslint:disable */
 /* eslint-disable */
 // @ts-nocheck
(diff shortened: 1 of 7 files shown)


dblandin (approved):
Fix looks good to me!

> I think this pattern of rendering is very brittle because of the implicit linking between the parent context and the rendered children which may have conditionals that drift.

Feel free to create an upkeep ticket if you think we should explore some refactoring later.

PR #4109, opened 2020-11-06
fix(show2): updates remaining time once per second #trivial

dzucconi:

This PR resolves [FX-2404]

### Description

This simply causes the timer to update once per second. We don't need to do any fancy syncing with the server for millisecond offsets or anything since this isn't an auction.

The ticket says to use the same component in viewing rooms. However:

* That component isn't entirely shared. [Part of it lives in viewing rooms and is specific to viewing rooms status.](https://github.com/artsy/eigen/blob/b9caccdd3a487a199095413e5bd9984de7dfe6ea/src/lib/Scenes/ViewingRoom/Components/ViewingRoomHeader.tsx#L50-L75)
* We already use this hook on fairs

So rather than use viewing rooms' component: we should settle on a detailed spec for how to format time, update the hook to match that spec, then refactor the viewing rooms header to use the hook. This can/should be a separate ticket.

[FX-2404]: https://artsyproduct.atlassian.net/browse/FX-2404

### DIFF
--- a/src/lib/Scenes/Show2/Components/Show2Header.tsx
+++ b/src/lib/Scenes/Show2/Components/Show2Header.tsx
@@ -1,7 +1,8 @@
 import { Show2Header_show } from "__generated__/Show2Header_show.graphql"
 import { useEventTiming } from "lib/utils/useEventTiming"
+import { DateTime } from "luxon"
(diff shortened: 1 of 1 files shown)


anandaroop (approved):
👌

PR #4112, opened 2020-11-06
fix: reset form when returning to artwork screen

erikdstock:

This PR resolves a bug found in QA. [PURCHASE-2223](https://artsyproduct.atlassian.net/browse/PURCHASE-2223]) is the one I set out to address (shipping checkbox gets its logic flipped if you exit/enter the form with it checked) but I think this actually will fix another bug where the 'send' button gets similarly messed up. I don't know if that was found in QA since I reported the former and assumed they were the same thing - they are not.

On an implementation level it:
- Extracts the 'inquiry question' checkbox into its own component
- cleans up some logic within the component, moving Animation setup into `useLayoutEffect`
- resets the form when exiting the inquiry modal, which prevents another bug causing the shipping checkbox to get flipped after exiting and re-entering
- removes some reducer code that mutates our starting value, making it possible to wind up in an unsubmittable state after exiting and re-entering

### Description
To see the bug, enter the inquiry form, tick the shipping box, press cancel to exit the form, and then re-enter the form. The form will be in a bad and unactionable state. This PR fixes it by resetting form state when leaving.

### DIFF
--- a/src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
+++ b/src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
@@ -48,9 +48,7 @@ const InquiryQuestionOption: React.FC<{
     }
   }
 
-  React.useLayoutEffect(() => {
-    maybeRegisterAnimation()
-  }, [questionSelected])
+  React.useLayoutEffect(maybeRegisterAnimation, [questionSelected])
 
   return (
     <React.Fragment>
--- a/src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
+++ b/src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
@@ -4,7 +4,7 @@ import { FancyModal } from "lib/Components/FancyModal/FancyModal"
 import { FancyModalHeader } from "lib/Components/FancyModal/FancyModalHeader"
 import ChevronIcon from "lib/Icons/ChevronIcon"
 import { ArtworkInquiryContext } from "lib/utils/ArtworkInquiry/ArtworkInquiryStore"
-import { InquiryOptions, InquiryQuestionIDs } from "lib/utils/ArtworkInquiry/ArtworkInquiryTypes"
+import { InquiryQuestionIDs } from "lib/utils/ArtworkInquiry/ArtworkInquiryTypes"
 import { Box, color, Flex, Separator, space, Text } from "palette"
 import React, { useContext, useState } from "react"
 import { LayoutAnimation, TouchableOpacity } from "react-native"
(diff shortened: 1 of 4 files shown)


src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
```diff
@@ -25,100 +25,115 @@ interface InquiryModalProps {
...
+        duration: 200,
+      })
+    }
+  }
 
-                  const shouldBeChecked: boolean = !state.inquiryQuestions.find((question) => {
-                    return question.questionID === id
-                  })
+  React.useLayoutEffect(() => {
+    maybeRegisterAnimation()
```
zephraph, line 52: Given that the `maybeRegisterAnimation` is only used in one place, should we just inject it here?

erikdstock: Ah yes, this effect had more going on inside on an earlier pass. I'll clean it up.

src/lib/utils/ArtworkInquiry/ArtworkInquiryStore.tsx
```diff
@@ -1,14 +1,14 @@
...
   ArtworkInquiryContextProps,
   ArtworkInquiryContextState,
   InquiryOptions,
   InquiryQuestionIDs,
 } from "lib/utils/ArtworkInquiry/ArtworkInquiryTypes"
-import { remove } from "lodash"
 import React, { createContext, Reducer, useReducer } from "react"
 
-const artworkInquiryState: ArtworkInquiryContextState = {
+const initialArtworkInquiryState: ArtworkInquiryContextState = {
```
zephraph, line 11: 👍

src/lib/utils/ArtworkInquiry/ArtworkInquiryStore.tsx
```diff
@@ -37,26 +39,23 @@ export const reducer = (
...
       const { isChecked, ...payloadQuestion } = action.payload
-
-      if (isChecked) {
-        inquiryState.inquiryQuestions.push(payloadQuestion)
-      } else {
-        remove(inquiryState.inquiryQuestions, (question) => question.questionID === payloadQuestion.questionID)
-      }
+      const { inquiryQuestions } = inquiryState
+      const newSelection: InquiryQuestionInput[] = isChecked
+        ? [...inquiryQuestions, payloadQuestion]
```
zephraph, line 44: I was wondering what would happen here if `inquiryQuestions` was `null`. 

TIL: When applying a spread operation on an arrayLike, if it's not arrayLike (i.e. null or something else) then the browser will give a type error. When applying a spread operation on an object (like `{ ...foo }`) if `foo` is null the result just evaluates to an empty object.

src/lib/utils/ArtworkInquiry/ArtworkInquiryStore.tsx
```diff
@@ -37,26 +39,23 @@ export const reducer = (
 
     case "selectInquiryQuestion":
       const { isChecked, ...payloadQuestion } = action.payload
-
-      if (isChecked) {
-        inquiryState.inquiryQuestions.push(payloadQuestion)
-      } else {
-        remove(inquiryState.inquiryQuestions, (question) => question.questionID === payloadQuestion.questionID)
-      }
+      const { inquiryQuestions } = inquiryState
```
zephraph, line 42:
```suggestion
      const { inquiryQuestions = [] } = inquiryState
```

This'll add a default value to inquiryQuestions just in case some wonkiness has happened and the questions are empty.

src/lib/utils/ArtworkInquiry/ArtworkInquiryStore.tsx
```diff
@@ -37,26 +39,23 @@ export const reducer = (
...
 
 export const ArtworkInquiryContext = createContext<ArtworkInquiryContextProps>(null as any)
 
 export const ArtworkInquiryStateProvider = ({ children }: any) => {
-  const [state, dispatch] = useReducer<Reducer<ArtworkInquiryContextState, ArtworkInquiryActions>>(
-    reducer,
-    artworkInquiryState
-  )
+  const [state, dispatch] = useReducer<Reducer<ArtworkInquiryContextState, ArtworkInquiryActions>>(reducer, {
+    ...initialArtworkInquiryState,
```
zephraph, line 58: What's the value of doing `{ ...initialArtworkInquiryState }` here vs just passing `initialArtworkInquiryState`?

xtina-starr (approved):

src/lib/utils/ArtworkInquiry/ArtworkInquiryStore.tsx
xtina-starr: I had the same questions.

src/lib/utils/ArtworkInquiry/ArtworkInquiryStore.tsx
```diff
@@ -37,26 +39,23 @@ export const reducer = (
...
-      }
+      const { inquiryQuestions } = inquiryState
+      const newSelection: InquiryQuestionInput[] = isChecked
+        ? [...inquiryQuestions, payloadQuestion]
+        : inquiryQuestions.filter((q) => q.questionID !== payloadQuestion.questionID)
 
       return {
         ...inquiryState,
-        inquiryQuestions: inquiryState.inquiryQuestions,
+        inquiryQuestions: newSelection,
```
xtina-starr, line 49: 👍

zephraph (approved):
Looks good to me. Some minor questions, comments, and one code update suggestion. Take a look at the suggestion, but I consider nothing strictly blocking. Feel free to merge when you're ready. 👍

src/lib/utils/ArtworkInquiry/ArtworkInquiryStore.tsx
erikdstock: @zephraph does `as const` guard against mutating the inner values?

src/lib/utils/ArtworkInquiry/ArtworkInquiryStore.tsx
erikdstock: hmm thanks, I don't see how this would happen but i will put a fallback value there to be sure since this is not even a browser

src/lib/utils/ArtworkInquiry/ArtworkInquiryStore.tsx
erikdstock: i guess you did actually

src/lib/utils/ArtworkInquiry/ArtworkInquiryStore.tsx
erikdstock: This is probably unnecessary - I did it in the same pass as removing the mutating reducer code thinking it's best to always start with a copy.

erikdstock: And come to think of it this probably isn't even a real copy

erikdstock:
#squashongreen

PR #4119, opened 2020-11-09
fix: round down to one decimal in demand label #trivial

brainbicycle:

### Description

This is a quick follow-up to https://github.com/artsy/eigen/pull/4090, I noticed after the fact that @samchieng had asked to round down to one decimal place to avoid 10/10.

### Screenshots

DemandIndexiPad


admbtlr:
I think that @samchieng 's comment specifically addressed scores of 10. I don't think it's necessary to round down _everything_, but only to put in a special case that `score = score === 10 ? 9.9 : score`. At least that's how I understood it - but maybe I misunderstood?

brainbicycle:
This feels a bit arbitrary to me and I think in general we should try to avoid arbitrary transformations of the data on the client, that said I don't feel too strongly about it so I will make the change 👍

admbtlr:
Yeah, I hear you on that. But I guess it makes sense from a product perspective, and I also guess it's pretty unlikely that demand will round to 10.

admbtlr (approved):

admbtlr:
#squashongreen

PR #4123, opened 2020-11-09
fix: register to bid button tracking

MounirDhahri:

This PR resolves [CX-804]

### Description
Use cohesion for register to bid button tracking

[CX-804]: https://artsyproduct.atlassian.net/browse/CX-804


src/lib/Scenes/Sale/Components/RegisterToBidButton.tsx
```diff
@@ -107,10 +108,21 @@ export const RegisterToBidButtonContainer = createFragmentContainer(RegisterToBi
...
+    contextScreenOwnerId,
+    contextScreenOwnerSlug,
+    contextType,
+  }: {
+    contextScreenOwnerId: string
+    contextScreenOwnerSlug: string
+    contextType: ScreenOwnerType
+  }) => {
+    const trackArgs: TappedRegisterToBidArgs = {
+      contextModule: contextType === "sale" ? ContextModule.auctionHome : ContextModule.auctionsInfo,
```
pepopowitz, line 121: [blocking] 2 suggestions in this line: (1) compare to the enum value instead of the string to avoid hardcoded strings, and (2) per @mikehrom we should use `ContextModule.aboutThisAuction` for the "about this auction" screen.

```suggestion
      contextModule: contextType === OwnerType.sale ? ContextModule.auctionHome : ContextModule.aboutThisAuction,
```

pepopowitz: [non-blocking] I also prefer to pass in an additional argument for the `contextModule` than to make a decision here based on `contextType`, but that's only a preference. If the caller is already making a decision about what to pass for `contextType`, it seems reasonable to me that it can also make a decision about what to pass for `contextModule`.

src/lib/Scenes/Sale/Components/RegisterToBidButton.tsx
```diff
@@ -1,16 +1,16 @@
+import { ContextModule, ScreenOwnerType, tappedRegisterToBid, TappedRegisterToBidArgs } from "@artsy/cohesion"
```
pepopowitz, line 1:
```suggestion
import { ContextModule, OwnerType, ScreenOwnerType, tappedRegisterToBid, TappedRegisterToBidArgs } from "@artsy/cohesion"
```

This suggestion goes along with my suggestion below.

pepopowitz (changes requested):
Thank you!!! I added a couple suggestions but only one of them is blocking, marked accordingly.

src/lib/Scenes/Sale/Components/RegisterToBidButton.tsx
MounirDhahri: you're absolutely true 🙏

MounirDhahri: Submitting PRs after 8pm is never a good idea 🙈

PR #4125, opened 2020-11-09
fix(echo): Echo config sort and only skip update if `_OSS_=true`

pvinis:

This PR resolves [CX-]

### Description

This check was assuming that the `_OSS_` var will always be there when it shouldn't. This caused the app to basically never check for echo updates 😱. Now it checks as usual. 

Also in the admin menu, the "echo configuration" screen now sorts things alphabetically so we can find things more easily.

### DIFF
--- a/Artsy/View_Controllers/Admin/AREchoContentsViewController.m
+++ b/Artsy/View_Controllers/Admin/AREchoContentsViewController.m
@@ -43,8 +43,8 @@ static NSString *CellIdentifier = @"Cell";
 
     self.echo = [[ARAppDelegate sharedInstance] echo];
     self.messages = self.echo.messages.allValues;
-    self.routeKeys = self.echo.routes.allKeys;
-    self.featureKeys = self.echo.features.allKeys;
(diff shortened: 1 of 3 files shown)


dblandin (approved):
LGTM 👍

damassi:
👌

Artsy/View_Controllers/Admin/AREchoContentsViewController.m
```diff
@@ -43,8 +43,8 @@ - (void)viewDidLoad
 
     self.echo = [[ARAppDelegate sharedInstance] echo];
     self.messages = self.echo.messages.allValues;
-    self.routeKeys = self.echo.routes.allKeys;
-    self.featureKeys = self.echo.features.allKeys;
+    self.routeKeys = [self.echo.routes.allKeys sortedArrayUsingSelector:@selector(caseInsensitiveCompare:)];
+    self.featureKeys = [self.echo.features.allKeys sortedArrayUsingSelector:@selector(caseInsensitiveCompare:)];
```
brainbicycle, line 47: Nice ✨

PR #4137, opened 2020-11-12
feature(share): allow instagram share

pvinis:

This PR is part of Hackathon 8. It's about https://www.notion.so/artsy/Instagram-share-button-on-the-Artwork-page-in-the-Artsy-app-and-mobile-web-c08f321c8d414ba996cbdcc2296798d1.

Jira link: [CX-832]

### Description

I am making this PR so we have a diff basically, and maybe we can have a way to try this on the beta app.

In order to make it wor, we share two "representations", one is the URL that takes us to the artwork page/screen, and the other one is a base64 of the image. This way, if we share with an app that needs an image, it uses that, like Instagram. If an app needs a URL, it will grab the URL.

~~I added a button behind a flag. When that button is clicked, Instagram opens and you can share the image on your feed or stories.~~

~~There is also a second flag about removing after sharing. To share with Instagram, the image needs to be in your camera roll. If the flag is on, the app will remove it after the share. If the flag is off, it will keep it.~~

Demos:
~~https://f.p0l.co/file/dropshare-public-pavlos/RPReplay_Final1605182083.MP4~~
~~https://f.p0l.co/file/dropshare-public-pavlos/RPReplay_Final1605279585.MP4~~
https://f.p0l.co/file/dropshare-public-pavlos/RPReplay_Final1608654922.MOV
https://f.p0l.co/file/dropshare-public-pavlos/RPReplay_Final1608654922-2.MOV

[CX-832]: https://artsyproduct.atlassian.net/browse/CX-832


src/lib/Scenes/Artwork/Components/ArtworkHeader.tsx
```diff
@@ -14,16 +19,45 @@ interface ArtworkHeaderProps {
...
+    })
+
+    const { title, href, artists } = artwork
+    const details = shareContent(title!, href!, artists)
+
+    const url = ((artwork.images ?? [])[currentImageIndex]?.url ?? "").replace(":version", "large")
+
+    const resp = await RNFetchBlob.config({
+      fileCache: true,
+    }).fetch("GET", url)
```
ashfurrow, line 39: It feels like we should be able to get the image from the image cache, since it's likely already downloaded. But also, ¯\_(ツ)_/¯ I'd be fine with showing a spinner while we wait for this, since the user might have to wait a bit for the image to download first.

ashfurrow (commented):
Looks good! Just one thing I'd like to address about UX before we merge 👍

src/lib/Scenes/Artwork/Components/ArtworkHeader.tsx
pvinis: hm I think from the cache would be good way, but since everything has its own cache, it's weird. I'll look into using the existing cache, where it is etc.

pvinis:
This should be good to go.

src/lib/Scenes/Artwork/Components/ArtworkHeader.tsx
```diff
@@ -1,9 +1,14 @@
 import { ArtworkHeader_artwork } from "__generated__/ArtworkHeader_artwork.graphql"
+import { Schema } from "lib/utils/track"
 import { useScreenDimensions } from "lib/utils/useScreenDimensions"
 import { Box, Flex, Spacer } from "palette"
-import React from "react"
+import React, { useState } from "react"
+// @ts-ignore
```
brainbicycle, line 6: do we have type issues with `react-native-share`?

brainbicycle (approved):

src/lib/Scenes/Artwork/Components/ArtworkHeader.tsx
pvinis: yea.. it doesn't have types 🤦‍♂️

PR #4143, opened 2020-11-16
[PURCHASE-2099] Enhancement: Enable inquiry request notifications

ashleyjelks:

This PR resolves [PURCHASE-2099]

### Description
Enables a 2 second notification after an inquiry message mutation successfully completes.

Kapture 2020-11-16 at 14 09 15
Kapture 2020-11-18 at 15 56 10

cc @artsy/purchase-devs 

[PURCHASE-2099]: https://artsyproduct.atlassian.net/browse/PURCHASE-2099

### DIFF
--- a/src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
+++ b/src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
@@ -7,7 +7,7 @@ import { ArtworkInquiryContext } from "lib/utils/ArtworkInquiry/ArtworkInquirySt
 import { InquiryQuestionIDs } from "lib/utils/ArtworkInquiry/ArtworkInquiryTypes"
 import { LocationWithDetails } from "lib/utils/googleMaps"
 import { Box, color, Flex, Separator, space, Text } from "palette"
-import React, { useContext, useState } from "react"
+import React, { useContext, useEffect, useState } from "react"
 import { LayoutAnimation, TouchableOpacity } from "react-native"
 import NavigatorIOS from "react-native-navigator-ios"
 import { createFragmentContainer, graphql, RelayProp } from "react-relay"
@@ -24,6 +24,7 @@ interface InquiryModalProps {
   navigator?: NavigatorIOS
   modalIsVisible: boolean
   relay: RelayProp
+  onMutationSuccessful: (state: boolean) => void
 }
 
 const ErrorMessageFlex = styled(Flex)`
@@ -122,18 +123,33 @@ const InquiryQuestionOption: React.FC<{
 }
 
 export const InquiryModal: React.FC<InquiryModalProps> = ({ artwork, ...props }) => {
-  const { toggleVisibility, modalIsVisible, relay } = props
+  const { toggleVisibility, modalIsVisible, relay, onMutationSuccessful } = props
   const questions = artwork?.inquiryQuestions!
   const { state, dispatch } = useContext(ArtworkInquiryContext)
   const [shippingModalVisibility, setShippingModalVisibility] = useState(false)
   const [errorMessageVisibility, setErrorMessageVisibility] = useState(false)
   const selectShippingLocation = (locationDetails: LocationWithDetails) =>
     dispatch({ type: "selectShippingLocation", payload: locationDetails })
-
+  const [mutationSuccessful, setMutationSuccessful] = useState(false)
   const resetAndExit = () => {
     dispatch({ type: "resetForm", payload: null })
     toggleVisibility()
   }
+
+  useEffect(() => {
+    if (mutationSuccessful) {
+      toggleVisibility()
+
+      const delayNotification = setTimeout(() => {
+        onMutationSuccessful(true)
+        setMutationSuccessful(false)
+      }, 500)
+      return () => {
+        clearTimeout(delayNotification)
+      }
+    }
+  }, [mutationSuccessful])
+
   return (
     <FancyModal visible={modalIsVisible} onBackgroundPressed={() => resetAndExit()}>
       <FancyModalHeader
@@ -144,7 +160,7 @@ export const InquiryModal: React.FC<InquiryModalProps> = ({ artwork, ...props })
         rightButtonText="Send"
         rightButtonDisabled={state.inquiryQuestions.length === 0}
         onRightButtonPress={() => {
-          SubmitInquiryRequest(relay.environment, artwork, state, setErrorMessageVisibility)
+          SubmitInquiryRequest(relay.environment, artwork, state, setMutationSuccessful, setErrorMessageVisibility)
         }}
       >
         {state.inquiryType}
--- a/src/lib/Scenes/Artwork/Components/CommercialButtons/InquirySuccessNotification.tsx
+++ b/src/lib/Scenes/Artwork/Components/CommercialButtons/InquirySuccessNotification.tsx
@@ -1,0 +1,88 @@
+import { navigate } from "lib/navigation/navigate"
+import { useScreenDimensions } from "lib/utils/useScreenDimensions"
+import { CloseIcon, color, Flex, Text, Theme } from "palette"
+import React, { useEffect } from "react"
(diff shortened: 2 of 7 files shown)


lilyfromseattle (approved):
👏 👏 👏

src/lib/Scenes/Artwork/Components/CommercialButtons/__tests__/InquiryButtons-tests.tsx
```diff
@@ -0,0 +1,93 @@
...
+  presentNavigationViewController: jest.fn(),
+}))
+jest.mock("lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal", () => {
+  return {
+    InquiryModalFragmentContainer: ({ toggleNotification }: any) => {
+      mockSuccessfulMutation.mockImplementation((mockState) => {
+        toggleNotification(mockState)
+      })
+      return null
+    },
```
erikdstock, line 23: This is a fake modal, basically a function component, that lets us mock a successful mutation without needing to stub all the implementation details of the shipping modal.

src/lib/Scenes/Artwork/Components/CommercialButtons/__tests__/InquiryButtons-tests.tsx
```diff
@@ -0,0 +1,93 @@
...
+    mockSuccessfulMutation(true)
+    expect(wrapper.root.findByType(InquirySuccessNotification).props.modalVisible).toBe(true)
+  })
+
+  it("clears the message sent notifcation after 2000 ms", () => {
+    const wrapper = getWrapper()
+    mockSuccessfulMutation(true)
+    expect(wrapper.root.findByType(InquirySuccessNotification).props.modalVisible).toBe(true)
+    jest.advanceTimersByTime(2001)
+    jest.runAllTicks()
```
erikdstock, line 82: because `flushPromiseQueue` uses a timeout it does not play with with fake timers. We started on something like

```
jest.advanceTimersByTime(2001)
flushPromiseQueue().then(() => {
  expect(...) 
})
```
but forgot to use the `done()` callback in the moment, and when we switched to `runAllTicks` it worked so we stopped trying new things.

src/lib/Scenes/Artwork/Components/CommercialButtons/__tests__/InquiryButtons-tests.tsx
```diff
@@ -0,0 +1,93 @@
...
+beforeEach(() => {
+  jest.useFakeTimers()
+  env = createMockEnvironment()
+})
+
+afterEach(() => {
+  jest.clearAllMocks()
+})
+
+const mockSuccessfulMutation = jest.fn()
```
erikdstock, line 36: this is our special little button that pretends we did the whole modal mutation bit.

erikdstock (commented):
Left comments on approach, some of which might be anti-patterns?

src/lib/Scenes/Artwork/Components/CommercialButtons/InquirySuccessNotification.tsx
```diff
@@ -0,0 +1,94 @@
...
+      transparent
+      ref={navRef}
+    >
+      <Animated.View
+        style={{
+          position: "absolute",
+          left: 0,
+          right: 0,
+          bottom: 0,
+          height: "95%",
```
ds300, line 46: Rather than using a proportional height to avoid the notch here we can use `safeAreaInsets` to make sure this works on all devices.

```suggestion
          top: useScreenDimensions().safeAreaInsets.top
```

src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
```diff
@@ -122,18 +123,34 @@ const InquiryQuestionOption: React.FC<{
 }
 
 export const InquiryModal: React.FC<InquiryModalProps> = ({ artwork, ...props }) => {
-  const { toggleVisibility, modalIsVisible, relay } = props
+  let delayNotification: NodeJS.Timeout | any
```
ds300, line 126: A few things here:

- The type `NodeJS.Timeout | any` is exactly the same as `any` as far as typescript is concerned. So we don't get any useful type checking on this variable. It should probably be `NodeJS.Timeout | undefined`
- Stateful variables in react functional components should be created with a hook of some kind, e.g. `useState/useReducer/useRef`, but in this case that's not necessary since it can live inside the `useEffect` as a `const`. i.e. you can delete this line and then do


```diff
  useEffect(() => {
    if (mutationSuccessful) {
      toggleVisibility()

-     delayNotification = setTimeout(() => {
+     const delayNotification = setTimeout(() => {
        toggleNotification(true)
        setMutationSuccessful(false)
      }, 500)
      return () => {
        clearTimeout(delayNotification)
      }
    }
```

And this way it doesn't need an explicit type annotation either!

ds300 (commented):
This looks really nice, good work! 🎉

src/lib/Scenes/Artwork/Components/CommercialButtons/__tests__/InquiryButtons-tests.tsx
```diff
@@ -0,0 +1,93 @@
...
+          }
+        }
+      `}
+      variables={{ id: "great-artttt" }}
+      render={({ props, error }) => {
+        if (Boolean(props?.artwork!)) {
+          return <InquiryButtonsFragmentContainer artwork={props!.artwork!} />
+        } else if (Boolean(error)) {
+          console.error(error)
+        }
```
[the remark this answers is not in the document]
erikdstock: We added this to address an error the linter has been reporting lately that demands explicit booleans (not even `!!` are allowed)

src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
```diff
@@ -122,18 +123,34 @@ const InquiryQuestionOption: React.FC<{
...
     dispatch({ type: "resetForm", payload: null })
     toggleVisibility()
   }
+
+  useEffect(() => {
+    if (mutationSuccessful) {
+      toggleVisibility()
+
+      delayNotification = setTimeout(() => {
+        toggleNotification(true)
```
erikdstock, line 145: Thinking about this prop (`toggleNotification`) and the way we went about testing I have a proposal to change this name to correspond more to the interface. In short, I think this prop should be called `onSuccess` or something similar; since props are public interface this means that the `InquiryModal` doesn't need to know that it is showing a banner, only that it has a callback to execute when the mutation is successful.

Likewise in our tests for the containing Button component, we stub the modal out completely and trigger a `mockSuccessfulMutation` function. A similar test for the inquiry modal can be `it('calls its onSuccess callback if the mutation is successful', ...`, making the interface clear from both ends. Test output would look something like this, corresponding to what the outer and inner components need to know about each other:
```
InquiryButtons:
  - displays the banner if the inquiry mutation is successful (which we mock out in the tests)
InquiryModal:
  - calls its `onSuccess` prop if the mutation is successful (to validate the assumptions we mocked above)
```

erikdstock (commented):
One minor comment that i think would give us a less leaky public interface on InquiryModal.

src/lib/Scenes/Artwork/Components/CommercialButtons/__tests__/InquiryButtons-tests.tsx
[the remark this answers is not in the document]
ds300: hmm that sounds like a broken lint rule to me, I'll look into it

src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
ashleyjelks: Niiice!

src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
ashleyjelks: Cool, I update the prop name in InquiryModal to `onMutationSuccessful`!

PR #4149, opened 2020-11-19
fix: persist shipping location for inquiry requests

xtina-starr:

This PR resolves [PURCHASE-2226]

### Description
When a user checks the shipping quote checkbox and adds a shipping location, the location was not being persisted through to the Inquiry Request instance in gravity. This is caused by two bugs:

1. The way we were previously assigning the shipping location to details was incorrect.
2. We were passing the shipping location as a string when the backend expects an object/hash.

The new implementation makes an additional api request to get an object that represents the city location and handles updating the shipping quote inquiry questions with details in the submit function when we have all the necessary data. 

shipping location

Screen Shot 2020-11-19 at 11 36 12 AM
 
Paired with @erikdstock 

[PURCHASE-2226]: https://artsyproduct.atlassian.net/browse/PURCHASE-2226

### DIFF
--- a/src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
+++ b/src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
@@ -5,7 +5,7 @@ import { FancyModalHeader } from "lib/Components/FancyModal/FancyModalHeader"
 import ChevronIcon from "lib/Icons/ChevronIcon"
 import { ArtworkInquiryContext } from "lib/utils/ArtworkInquiry/ArtworkInquiryStore"
 import { InquiryQuestionIDs } from "lib/utils/ArtworkInquiry/ArtworkInquiryTypes"
-import { LocationDetails } from "lib/utils/googleMaps"
+import { LocationWithDetails } from "lib/utils/googleMaps"
 import { Box, color, Flex, Separator, space, Text } from "palette"
 import React, { useContext, useState } from "react"
 import { LayoutAnimation, TouchableOpacity } from "react-native"
@@ -127,7 +127,7 @@ export const InquiryModal: React.FC<InquiryModalProps> = ({ artwork, ...props })
   const { state, dispatch } = useContext(ArtworkInquiryContext)
   const [shippingModalVisibility, setShippingModalVisibility] = useState(false)
   const [errorMessageVisibility, setErrorMessageVisibility] = useState(false)
-  const selectShippingLocation = (locationDetails: LocationDetails) =>
+  const selectShippingLocation = (locationDetails: LocationWithDetails) =>
     dispatch({ type: "selectShippingLocation", payload: locationDetails })
 
   const resetAndExit = () => {
--- a/src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
+++ b/src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
@@ -5,6 +5,7 @@ import { FancyModalHeader } from "lib/Components/FancyModal/FancyModalHeader"
 import ChevronIcon from "lib/Icons/ChevronIcon"
 import { ArtworkInquiryContext } from "lib/utils/ArtworkInquiry/ArtworkInquiryStore"
 import { InquiryQuestionIDs } from "lib/utils/ArtworkInquiry/ArtworkInquiryTypes"
+import { LocationDetails } from "lib/utils/googleMaps"
 import { Box, color, Flex, Separator, space, Text } from "palette"
 import React, { useContext, useState } from "react"
 import { LayoutAnimation, TouchableOpacity } from "react-native"
@@ -70,7 +71,7 @@ const InquiryQuestionOption: React.FC<{
                   type: "selectInquiryQuestion",
                   payload: {
                     questionID: id,
-                    details: isShipping ? state.shippingLocation : null,
+                    details: isShipping ? state.shippingLocation?.name : null,
                     isChecked: !questionSelected,
                   },
                 })
@@ -104,7 +105,7 @@ const InquiryQuestionOption: React.FC<{
                 ) : (
                   <>
                     <Text variant="text" color="black100" style={{ width: "70%" }}>
-                      {state.shippingLocation}
+                      {state.shippingLocation.name}
                     </Text>
                     <Text variant="text" color="purple100">
                       Edit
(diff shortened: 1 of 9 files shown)


src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
```diff
@@ -70,7 +71,7 @@ const InquiryQuestionOption: React.FC<{
                   type: "selectInquiryQuestion",
                   payload: {
                     questionID: id,
-                    details: isShipping ? state.shippingLocation : null,
```
xtina-starr, line 73: This will basically always be `null` or possibly an outdated shipping location because this dispatch checks if the shipping question is check to assign the shippingLocation from state, but the user needs to check the box before being able to set the location.

src/lib/Scenes/Artwork/Components/Mutation/SubmitInquiryRequest.ts
```diff
@@ -1,12 +1,32 @@
...
+        postal_code: postalCode,
+        state,
+        state_code: stateCode,
+      }
+      const details = JSON.stringify(locationInput)
+      return { ...q, details }
+    } else {
+      return q
+    }
+  })
```
xtina-starr, line 28: Here is where update the shipping quote details with the latest shipping location from the inquiryState.

src/lib/utils/googleMaps.ts
```diff
@@ -26,6 +36,60 @@ export const queryLocation = async (query: string): Promise<GMapsLocation[]> =>
   return results.predictions.map(predictionToResult)
 }
 
-const predictionToResult = (prediction: LocationResult): GMapsLocation => {
+const predictionToResult = (prediction: QueryAutocompletePrediction): SimpleLocation => {
   return { id: prediction.place_id, name: prediction.description }
 }
+
+export const getLocationDetails = async ({ id, name }: SimpleLocation): Promise<LocationWithDetails> => {
```
xtina-starr, line 43: This function get the full details for the city which was selected by the user.

jo-rs (approved):
LGTM! 💪 💪

src/lib/Scenes/Artwork/Components/CommercialButtons/LocationAutocomplete.tsx
```diff
@@ -1,19 +1,19 @@
...
-  initialLocation: string
+  initialLocation: SimpleLocation | null
 }
 
 export const LocationAutocomplete: React.FC<Props> = ({ onChange, initialLocation }) => {
-  const [predictions, setPredictions] = useState<GMapsLocation[]>([])
-  const [selectedLocation, setSelectedLocation] = useState(initialLocation ? initialLocation : "")
-  const [query, setQuery] = useState(selectedLocation)
+  const [predictions, setPredictions] = useState<SimpleLocation[]>([])
+  const [selectedLocation, setSelectedLocation] = useState<SimpleLocation | null>(initialLocation)
```
zephraph, line 15: One slight improvement that you could potentially make here. 

Given that this `useState` hook is being provided a prop directly, typescript can infer its type to the correct thing. 

Check out what the type results to [here](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAJQKYEMDGMA0cDecCuAzkgMowoxJwC+cAZlBCHAORSoYsCwAULzAE8wVEqDAAbJABkIaCsAgA7OAF5cvOJriKUIJAC44hGFGCKA5r2q9eZylDroqABUZhC6nprPAYwFOIycn5KhqLgkkHySnAAPtr44uJWNjxIAB6QsHBoSsZwUSGKAIL4MLJMEkiUhsjoMAB0AGIAwgA8rhDuAHyqcAAUeD5+AYUKytQAlKq9OBpw87mK+QDaxJIYSAAmY0rYxDAkSBuUO7LRigC6fUSk5JT9w-6B50WTqZrsMPhQym0AFgBGboACWO4ggbQA9EDelYgA). 

If, for whatever reason, a case like this happens where you _have_ to explicitly declare the type, there's two ways you could do this to ensure things don't become out of sync. 

You could do...

 ```
useState<typeof initialLocation>(initialLocation)
```
or
```
useState<Props['initialLocation']>
```

Either of could work. In this case though, I think letting typescript infer the usage is likely a better approach. In general the more types we have to manually specify, the more types we have to maintain. Just opens us up to more unintentional bugs (especially if we're copying a type)

```suggestion
  const [selectedLocation, setSelectedLocation] = useState(initialLocation)
```

src/lib/Scenes/Artwork/Components/Mutation/SubmitInquiryRequest.ts
```diff
@@ -19,8 +39,8 @@ export const SubmitInquiryRequest = (
       input: {
         inquireableID: inquireable.internalID,
         inquireableType: "Artwork",
-        message: payload.message,
-        questions: payload.inquiryQuestions,
+        // message: inquiryState.message,
```
zephraph, line 42: Should this be removed?

src/lib/Scenes/Artwork/Components/CommercialButtons/LocationAutocomplete.tsx
xtina-starr: So the suggestion here is to let typescript infer the type here versus explicitly set it?

src/lib/Scenes/Artwork/Components/Mutation/SubmitInquiryRequest.ts
xtina-starr: Yeah it can be.

src/lib/utils/googleMaps.ts
```diff
@@ -26,6 +36,60 @@ export const queryLocation = async (query: string): Promise<GMapsLocation[]> =>
...
+  const data = await response.json()
+
+  // TODO: Add dedicated error handling to the maps response
+  const { address_components, geometry } = data.result as PlaceResult
+  // @ts-ignore STRICTNESS_MIGRATION
+  const cityPlace = address_components.find((comp) => comp.types[0] === "locality")
+  // @ts-ignore STRICTNESS_MIGRATION
+  const statePlace = address_components.find((comp) => comp.types[0] === "administrative_area_level_1")
+  // @ts-ignore STRICTNESS_MIGRATION
+  const countryPlace = address_components.find((comp) => comp.types[0] === "country")
```
zephraph, line 59: I'm sure the typing ere is a little cumbersome, but is there an alternative to the ts-ignores? I don't know the best practice, but I'm wary of adding more `STRICTNESS_MIGRATION` flags because I believe (and @ds300, correct me if I'm wrong) that this was essentially supposed to be a one time thing.

That said, this is definitely something we can follow up on and non-blocker for this PR.

zephraph (approved):
Overall, this looks great! I've left a few notes of feedback and one suggestion, but nothing I feel super strongly about blocking this PR. There are multiple places that this PR improves types or naming in a way that adds clarity and makes things easier to follow, so props for that. 

Whenever you feel you're satisfied, feel free to merge.

src/lib/Scenes/Artwork/Components/CommercialButtons/LocationAutocomplete.tsx
zephraph: Yep. That way if the type if the type changes in the future it'll automatically flow through vs having to be manually updated.

PR #4150, opened 2020-11-19
fix: minimize startup api requests

brainbicycle:

This PR resolves [CX-823]

### Description

We were making a few api requests on app launch from a killed state. 

1. `/api/v1/devices` we call this endpoint when we get a success push notification registration callback to register a users device. Previously this would only happen on login but since integrating Sailthru this started happening on every launch from a killed state. Solution is to only call when the token has changed since last registration or the user has logged out.

2. `/dev/blank` we were making this call to pre-warm cache for our native WebViews, since we are less reliant on native WebViews nowadays I think it should be safe to just remove this. Anecdotally it doesn't seem to make a noticeable difference in loading time.  

See slack thread for more context: https://artsy.slack.com/archives/C01B2P6LJUU/p1605301828178900

[CX-823]: https://artsyproduct.atlassian.net/browse/CX-823


Artsy/App/ARAppNotificationsDelegate.m
```diff
@@ -236,13 +236,14 @@ - (void)application:(UIApplication *)application didRegisterForRemoteNotificatio
...
     // service always has an up-to-date record of devices and associated users.
-    //
-    // https://developer.apple.com/library/ios/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/Chapters/IPhoneOSClientImp.html#//apple_ref/doc/uid/TP40008194-CH103-SW2
-    [ArtsyAPI setAPNTokenForCurrentDevice:deviceToken success:^(id response) {
-        ARActionLog(@"Pushed device token to Artsy's servers");
-    } failure:^(NSError *error) {
-        ARErrorLog(@"Couldn't push the device token to Artsy, error: %@", error.localizedDescription);
-    }];
+    // https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622958-application?language=objc
+    if (![self tokensAreTheSame:previousToken newToken:deviceToken]) {
```
joeyAghion, line 240: 🙏 Thanks!

(Aside: how often/when does this change?)

joeyAghion (approved):

Artsy/App/ARAppNotificationsDelegate.m
brainbicycle: It seems like [Apple's docs](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622958-application?language=objc) (at least those I have seen) are deliberately vague on this, they just say periodically. From SO it sounds like there a various situations like resetting a device, upgrading OS or reinstalling an app that cause the token to change. That said on our end we are deleting tokens from gravity and recreating on logout/login based on [this conversation]( https://artsy.slack.com/archives/CN8S32FJR/p1580316910106300) from a while back which I am guessing logouts/logins will be generating the large majority of update calls to gravity but other cases should be handled by this check.

PR #4154, opened 2020-11-20
feature: Move my bids

lilyfromseattle:
[PURCHASE-2172]

Currently this view exists as its own screen under the profile page. It is feature-flagged and only visible in the debug menu. This change moves MyBids into the inbox view, and adds tabs to allow users to switch between bids and conversations. Had lots of pairing help from @erikdstock, and he refactored my changes for clarity in a PR to this branch.

Screen Shot 2020-11-30 at 5 02 09 PM

Screen Shot 2020-11-30 at 5 01 58 PM

[PURCHASE-2172]: https://artsyproduct.atlassian.net/browse/PURCHASE-2172


erikdstock:
I have a PR (#4160) that I think could make this one a bit less complicated and safer to work on as we continue the inbox/my bids work. It is directed at this move-my-bids branch, not master.

src/lib/Containers/Inbox/InboxOld.tsx
```diff
@@ -1,5 +1,5 @@
-import { Inbox_me } from "__generated__/Inbox_me.graphql"
-import { InboxQuery } from "__generated__/InboxQuery.graphql"
+import { InboxOld_me } from "__generated__/InboxOld_me.graphql"
+import { InboxOldQuery } from "__generated__/InboxOldQuery.graphql"
```
zephraph, line 2: So this update is from @erikdstock's other PR to make it easier to toggle between the two modes, right?

src/lib/Scenes/Inbox/Components/Conversations/Conversations.tsx
```diff
@@ -77,15 +78,20 @@ export class Conversations extends Component<Props, State> {
...
-        </Flex>
+        {!!shouldDisplayMyBids ? (
+          <Separator />
+        ) : (
+          <Flex py={1} style={{ borderBottomWidth: 1, borderBottomColor: color("black10") }}>
+            <Sans mx={2} mt={1} size="8" style={{ borderBottomWidth: 1, borderBottomColor: color("black10") }}>
+              Inbox {unreadCounter}
+            </Sans>
+          </Flex>
+        )}
```
zephraph, line 94: This feels a little bit like a code smell to me. Why should conversations need to know about bids? Is it possible for us to hoist this up to a higher level?

cc @lilyfromseattle, @erikdstock

src/lib/Scenes/MyBids/MyBids.tsx
```diff
@@ -53,13 +53,7 @@ class MyBids extends React.Component<MyBidsProps> {
...
       <Flex flex={1}>
         <StickyTabPage
-          staticHeaderContent={
-            <Flex mt={2}>
-              <Text variant="mediumText" textAlign="center">
-                My Bids
-              </Text>
-            </Flex>
-          }
+          staticHeaderContent={<></>}
```
zephraph, line 56: Is this prop required if nothing is being passed to it?

src/lib/Containers/Inbox/InboxOld.tsx
lilyfromseattle: Yup!

src/lib/Scenes/Inbox/Components/Conversations/Conversations.tsx
lilyfromseattle: It's just that the "Inbox" title only appears when bids isn't integrated into the inbox (because its made redundant by the Inquiries tab up top). I could move it up to the parent but then I'd have to move the unread logic too and since it's going to get ripped out once the new my bids goes live I thought it would be ok to just have it like this for now, but I'm open to other suggestions.

src/lib/Scenes/MyBids/MyBids.tsx
lilyfromseattle: Yes, I tried making it optional but I was running into some errors and thought it best not to mess with it. But I can work through the errors if you think it's important

zephraph (approved):
Thanks for the comments @lilyfromseattle! I think there's a little clean up we could do, but it's something I'm comfortable pushing off to another PR. Before we launch this (when removing the lab flag) we'll want to come back and delete the old inbox too.

PR #4156, opened 2020-11-20
feat: change title and ordering of artwork price range filter

anandaroop:

This PR resolves [FX-2398]

### Description

This tackles two of the tasks in the associated ticket:

- Change the title of the price range filter from "Price/estimate range"
- Change the ordering of the price buckets from highest to lowest

After looking into it, this PR _punts on_ the following change from the ticket, since this appears to require upstream changes to Gravity ~and Metaphysics(?)~ (I will split out the ticket accordingly):

- Split the USD 10,000–50,000 bucket into 10,000–25,000 and 25,000–50,000

While the change was ticketed for Fairs…

fair: all filters fair: price filter selected

…it is also intended to take effect globally (artist, series, collection, show), which these screen caps partly demonstrate: 

series collection

The one exception is the auction filters, which should continue to show the "Price/estimate" language…

auctions

[FX-2398]: https://artsyproduct.atlassian.net/browse/FX-2398

### DIFF
--- a/src/lib/Components/ArtworkFilterOptions/PriceRangeOptions.tsx
+++ b/src/lib/Components/ArtworkFilterOptions/PriceRangeOptions.tsx
@@ -27,7 +27,7 @@ const priceRangeDisplayText: Map<string, string> = new Map([
 ])
 
 const priceSort = (left: FilterData, right: FilterData): number => {
-  const sortOrder = ["*-*", "*-1000", "1000-5000", "5000-10000", "10000-50000", "50000-*"]
+  const sortOrder = ["*-*", "50000-*", "10000-50000", "5000-10000", "1000-5000", "*-1000"]
   const leftParam = left.paramValue as string
   const rightParam = right.paramValue as string
   if (sortOrder.indexOf(leftParam) < sortOrder.indexOf(rightParam)) {
--- a/src/lib/Components/ArtworkFilterOptions/__tests__/PriceRangeOptions-tests.tsx
+++ b/src/lib/Components/ArtworkFilterOptions/__tests__/PriceRangeOptions-tests.tsx
@@ -104,6 +104,14 @@ describe("Price Range Options Screen", () => {
     expect(extractText(firstOption)).toContain("All")
   })
 
+  it("renders price range options from most to least expensive", () => {
+    const tree = renderWithWrappers(<MockPriceRangeScreen initialState={state} />)
+    const options = tree.root.findAllByType(OptionListItem)
+    const renderedOptions = options.map(extractText)
+
+    expect(renderedOptions).toEqual(["All", "$50,000+", "$10,000-50,000", "$5,000-10,000", "$1000-5,000", "$0-1,000"])
+  })
+
   describe("selectedPriceRangeOption", () => {
     it("returns the default option if there are no selected or applied filters", () => {
       const tree = renderWithWrappers(<MockPriceRangeScreen initialState={state} />)
--- a/src/lib/utils/ArtworkFilter/FilterArtworksHelpers.ts
+++ b/src/lib/utils/ArtworkFilter/FilterArtworksHelpers.ts
@@ -42,7 +42,7 @@ export enum FilterDisplayName {
   gallery = "Gallery",
   institution = "Institution",
   medium = "Medium",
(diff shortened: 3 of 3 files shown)


src/lib/Components/ArtworkFilterOptions/PriceRangeOptions.tsx
```diff
@@ -27,7 +27,7 @@ const priceRangeDisplayText: Map<string, string> = new Map([
 ])
 
 const priceSort = (left: FilterData, right: FilterData): number => {
-  const sortOrder = ["*-*", "*-1000", "1000-5000", "5000-10000", "10000-50000", "50000-*"]
+  const sortOrder = ["*-*", "50000-*", "10000-50000", "5000-10000", "1000-5000", "*-1000"]
```
mzikherman, line 30: Ah, so this is the 'essential' change!

mzikherman:
> this appears to require upstream changes to Gravity and Metaphysics(?)

I don't think it technically needs any. The filter API can take arbitrary ranges of the form `10000-25000` and `25000-50000` already. The aggregation in Gravity still would use the `10000-50000` bucket, but if we're just displaying all the ranges, in order, we don't _need_ to update the aggregation bucketing in Gravity (and we could save requesting that as well).

anandaroop:
I thought I could just changes all these bucketing thresholds in the client but when I did I observed the weirdness below.

I traced this to [code that relies on the actual agg buckets in the search response](https://github.com/artsy/eigen/blob/ec839528b1640c3264f44a155e70fb926d19be34/src/lib/Components/ArtworkFilterOptions/PriceRangeOptions.tsx#L45), thus my conclusion that changes need to happen further upstream.

I could be misunderstanding, would be happy to pair on this!

Screen Shot 2020-11-20 at 4 29 54 PM

mzikherman:
Hmm, interesting! I'm not super familiar with the Eigen artwork filter, but was more thinking it would be like the Force one: https://github.com/artsy/force/blob/a4b251a38b50eeb4e95458068d67be757aa627ef/src/v2/Components/v2/ArtworkFilter/ArtworkFilters/PriceRangeFilter.tsx#L36

Rather than ask for a price bucket aggregation, why not list all the desired buckets, in order? Using the returned aggregation to limit options to just those with results definitely makes sense for large/unbounded potential filters (gallery, artist, etc.), but for filters with just a few options ('Ways to Buy'- four different options, 'Size'- small, medium, large, 'Price'- just these 6 predefined buckets), instead of asking for the aggregation, they can all be listed? I posit that requesting those aggregations and then only listing those returned ones adds complexity for very little benefit.

In essence, I'm recommending hard-coding the desired price ranges which should always be displayed, much like Force: https://github.com/artsy/force/blob/a4b251a38b50eeb4e95458068d67be757aa627ef/src/v2/Components/v2/ArtworkFilter/ArtworkFilters/PriceRangeFilter.tsx#L36 . I'm not sure how easily refactorable that is into the current Eigen filter (ie- is there a single component/place this update could happen?).

anandaroop:
Yeah, seems like it is a [deliberate design decision](https://artsyproduct.atlassian.net/browse/FX-2398?focusedCommentId=32662) on iOS (as compared to web) to only display buckets that are represented in the current search hits. 

I guess the implementation hinges on that UX decision, maybe one we should revisit.

mzikherman:
Yea, I wonder how 'intentional' that UX is, or isn't. I believe web started off w/ a similar UX (the reason for the complex aggregation computations possible, including the nuance of having an accurate 'medium' filter _even if_ you were already filtering by 'medium', for instance).

I guess wanting to minimize empty states, generally (even if they will always still be possible). On web we basically...stopped worrying about those? We don't display the counts next to each filter option, and instead have the more reliable/consistent set of filter options always shown. Not sure which is better for Eigen (or if the decision/design was ever re-visited).

In the case of wanting to preserve this behavior in Eigen, the 'correct' upstream fix looks surprisingly simple! Literally adding `25000` to the list of breaks for the bucketing: https://github.com/artsy/gravity/blob/a05d75810cf09fe95a40500d20afbd615a7e7e37/app/models/util/for_sale.rb#L92 (and re-indexing all artworks of course!).

PR #4160, opened 2020-11-21
chore: split old and new inboxes

erikdstock:
This is a PR to your development branch @lilyfromseattle - totally your call on whether to incorporate it, and I'm not sure if it would be considered best practice or not so tagging some others as well.
I was thinking about the sort of complicated display logic in our Inbox component and in particular how we are fetching both  the old `lotStandings` (from gravity) and the new `auctionsLotStandingConnection`* and wondered if we can separate these out, so I made a PR to do that.

In short, I split `Inbox.tsx` into a directory with `Inbox` & `InboxOld` (the version of `Inbox.tsx` currently on master) plus an `InboxWrapper` (which checks the AREmissionOptions to decide which to display. 

As a side benefit we are working with a new single component now so can iterate on it without worrying about breaking old functionality, and when we do switch over completely we can just delete the OldInbox + wrapper files and move back to a single-file module.

*side note- if this isn't the best way to go about it we will want to change the `hadBids` logic in some other way, since the two queries for lot standings return different results.

### DIFF
--- a/src/lib/Containers/Inbox/Inbox.tsx
+++ b/src/lib/Containers/Inbox/Inbox.tsx
@@ -1,0 +1,178 @@
+import { Inbox_me } from "__generated__/Inbox_me.graphql"
+import { InboxQuery } from "__generated__/InboxQuery.graphql"
+import { defaultEnvironment } from "lib/relay/createEnvironment"
+import { ActiveBids as ActiveBidsRef } from "lib/Scenes/Inbox/Components/ActiveBids"
+import { ConversationsContainer } from "lib/Scenes/Inbox/Components/Conversations/Conversations"
+import ZeroStateInbox from "lib/Scenes/Inbox/Components/Conversations/ZeroStateInbox"
+import { MyBidsContainer } from "lib/Scenes/MyBids/MyBids"
+import { listenToNativeEvents } from "lib/store/NativeModel"
+import renderWithLoadProgress from "lib/utils/renderWithLoadProgress"
+import { Flex, Spacer, Text } from "palette"
+import React from "react"
+import { EmitterSubscription, RefreshControl } from "react-native"
+import { createRefetchContainer, graphql, QueryRenderer, RelayRefetchProp } from "react-relay"
+import styled from "styled-components/native"
+
+interface State {
+  fetchingData: boolean
+  inquiryTabIsSelected: boolean
+}
+
+interface Props {
+  me: Inbox_me
+  relay: RelayRefetchProp
+  isVisible: boolean
+}
+
+const Container = styled.ScrollView`
+  flex: 1;
+`
+export class Inbox extends React.Component<Props, State> {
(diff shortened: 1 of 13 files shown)


src/lib/Containers/Inbox/InboxOld.tsx
```diff
@@ -88,58 +88,19 @@ export class Inbox extends React.Component<Props, State> {
...
-            componentRef={(conversations) => (this.conversations = conversations)}
-          />
-        </Container>
-      ) : (
-        <Flex style={{ flex: 1 }}>
-          <ZeroStateInbox />
-        </Flex>
-      )
-    }
+    return hasBids || hasConversations ? (
```
erikdstock, line 91: This logic, for showing the empty state, survives in OldInbox but is changed in the new Inbox.

[next](https://github.com/artsy/eigen/pull/4160/files#diff-6d7095fdbd37d5afeb7c1405e1adcee32cf710ec37f79ac60d77f80077eb53e5R84-R89)

src/lib/Containers/Inbox/Inbox.tsx
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  render() {
+    const somethingToShow = Boolean(
+      // tslint:disable-next-line: strict-boolean-expressions
+      this.props.me.lotStandingsExistenceCheck.edges?.length || this.props.me.conversationsExistenceCheck?.edges?.length
+    )
+
+    if (!somethingToShow) {
+      return (
```
erikdstock, line 90: The new inbox uses a different relay query to achieve the same result for showing the empty state.

[next](https://github.com/artsy/eigen/pull/4160/files#diff-52f8cfecf66fc7db8f94250ccddda826c3d7bdea35a5ec7c52b12a1687ef1c10R10)

src/lib/Containers/Inbox/index.tsx
```diff
@@ -0,0 +1,13 @@
...
+import { InboxQueryRenderer } from "./Inbox"
+import { InboxOldQueryRenderer } from "./InboxOld"
+
+/**
+ * A wrapper containing the logic for whether to display our new inbox
+ */
+export const InboxWrapper: React.FC = () => {
+  const shouldDisplayMyBids = getCurrentEmissionState().options.AROptionsBidManagement
+
+  return shouldDisplayMyBids ? <InboxQueryRenderer /> : <InboxOldQueryRenderer />
```
erikdstock, line 12: This wrapper component (temporarily) houses the logic deciding which component to render, and by extension which query to make.

erikdstock (commented):
Left a miniature code tour showing how things are broken up.

erikdstock:
Failing tests are expected because this is a PR to a still in-progress branch. If we like the changes we could just merge the commits directly into that branch and close this PR.

ds300:
👍 this approach makes total sense, we've done similar things when redesigning a bunch of different screens.

PR #4161, opened 2020-11-23
feat: first inquiry custom message

jo-rs:

This PR resolves [PURCHASE-2209]

### Description

This PR enables the user to submit a custom message through the first inquiry flow.
input_small

[PURCHASE-2209]: https://artsyproduct.atlassian.net/browse/PURCHASE-2209

### DIFF
--- a/src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
+++ b/src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
@@ -8,7 +8,7 @@ import { InquiryQuestionIDs } from "lib/utils/ArtworkInquiry/ArtworkInquiryTypes
 import { LocationWithDetails } from "lib/utils/googleMaps"
 import { Box, color, Flex, Separator, space, Text } from "palette"
 import React, { useContext, useState } from "react"
-import { LayoutAnimation, TouchableOpacity } from "react-native"
+import { LayoutAnimation, TextInput, TextInputProps, TouchableOpacity } from "react-native"
 import NavigatorIOS from "react-native-navigator-ios"
 import { createFragmentContainer, graphql, RelayProp } from "react-relay"
 import styled from "styled-components/native"
@@ -129,6 +129,7 @@ export const InquiryModal: React.FC<InquiryModalProps> = ({ artwork, ...props })
   const [errorMessageVisibility, setErrorMessageVisibility] = useState(false)
   const selectShippingLocation = (locationDetails: LocationWithDetails) =>
     dispatch({ type: "selectShippingLocation", payload: locationDetails })
+  const setMessage = (message: string) => dispatch({ type: "setMessage", payload: message })
 
   const resetAndExit = () => {
     dispatch({ type: "resetForm", payload: null })
@@ -179,6 +180,14 @@ export const InquiryModal: React.FC<InquiryModalProps> = ({ artwork, ...props })
           })
         }
       </Box>
+      <Box mx={2}>
+        <TextArea
+          placeholder="Add a custom note..."
+          title="Add Message"
+          value={state.message ? state.message : ""}
+          onChangeText={setMessage}
+        />
+      </Box>
       <ShippingModal
         toggleVisibility={() => setShippingModalVisibility(!shippingModalVisibility)}
         modalIsVisible={shippingModalVisibility}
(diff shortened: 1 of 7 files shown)


src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
```diff
@@ -213,6 +222,42 @@ const InquiryField = styled(Flex)`
...
+        onFocus={() => {
+          setBorderColor(color("purple100"))
+        }}
+        onBlur={() => {
+          setBorderColor(color("black10"))
+        }}
+        style={{ borderColor }}
+        numberOfLines={3}
+        multiline={true}
+      />
```
xtina-starr, line 256: Do we know how the `InputText` handles long messages? Like does the text area grow or scroll when it exceeds the space available? Also how does the 'numberOfLines` prop work?

jo-rs: I'll kill `numberOfLines` it only works on Android, I forgot to take that out.
Currently the input stays the same size, it scrolls on longer text.

src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
```diff
@@ -158,7 +159,9 @@ export const InquiryModal: React.FC<InquiryModalProps> = ({ artwork, ...props })
           resetAndExit()
         }}
         rightButtonText="Send"
-        rightButtonDisabled={state.inquiryQuestions.length === 0}
+        rightButtonDisabled={
+          (state.inquiryQuestions.length === 0 && !!questions.length) || (!questions.length && !state.message)
```
zephraph, line 163: I'm trying to work through this in my head, but I think this logic isn't correct. 

I think it's supposed to be as follows:

If no question is checked and the message is empty, disable the button.

zephraph: There are three factors considered here

1. Are there questions
2. Are any questions selected
3. Is there a message

In the situation that there are no questions we expect that there must be a message. If there's no message we expect that there must be questions selected. I think we can leverage those two facts to simplify the overall logic to just two factors.

```suggestion
          (state.inquiryQuestions.length === 0 && !state.message)
```

jo-rs: I think my brain might still not be in full working order; this took me a minute.

Yeah, you're right. Nice, @zephraph! 💪

src/lib/utils/ArtworkInquiry/ArtworkInquiryStore.tsx
```diff
@@ -48,6 +49,11 @@ export const reducer = (
         ...inquiryState,
         inquiryQuestions: newSelection,
       }
+    case "setMessage":
+      return {
+        ...inquiryState,
+        message: action.payload,
```
zephraph, line 55: I think it'd be good for us to remove any extra whitespace from the payload to the user isn't just submitting with a lot of whitespace or that there's weird padding in the email that goes out to partners. 

```suggestion
        message: action.payload.trim(),
```

src/lib/utils/ArtworkInquiry/ArtworkInquiryStore.tsx
```diff
@@ -52,7 +52,7 @@ export const reducer = (
     case "setMessage":
       return {
         ...inquiryState,
-        message: action.payload,
+        message: action.payload.trim(),
```
xtina-starr, line 55: 👌

zephraph (approved):
This looks good to me. Thanks for seeing this through!

xtina-starr:
I agree! I'm going to add the `SquashOnGreen` label.

src/lib/utils/ArtworkInquiry/ArtworkInquiryStore.tsx
jo-rs: Actually, @zephraph, this cuts spaces on every keyup. So, I'll modify to trim only on submit.

jo-rs: @xtina-starr the `trim` actually doesn't work the way it is now. I'll submit another PR with a fix.

jo-rs:
Bugfix on [this](https://github.com/artsy/eigen/pull/4165) PR

PR #4163, opened 2020-11-23
chore: improve STRICTNESS_MIGRATION situation

ds300:
### Description

Folks are frequently copy-pasting `STRICTNESS_MIGRATION` comments and receiving feedback about it in PR reviews. I've done a couple of things here to give earlier and more useful feedback:

- Added extra inline context to let people know that code with `STRICTNESS_MIGRATION` comments is unsafe, gently implying that the code should not be reproduced.
- Use the new (since we turned on strict mode at least) TS compiler feature `@ts-expect-error` instead of `@ts-ignore`. `@ts-expect-error` will cause TS compiler errors when the comment is not actually suppressing one or more errors.

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -10,6 +10,7 @@ upcoming:
     - Fix minimize startup api requests - brian
     - Remove analytics tracking on back-arrow taps - pepopowitz
     - Rename AppStore to GlobalStore - david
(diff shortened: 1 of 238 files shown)


ashfurrow (approved):
Good idea 👍 Good use of emojis too!

MounirDhahri (approved):

MounirDhahri:
Makes sense 🌟

PR #4164, opened 2020-11-23
fix(MyAccount): field edit nav issues

ds300:
This PR resolves 

- https://www.notion.so/artsy/Profile-Account-updating-phone-number-or-name-probably-others-doesn-t-dismiss-edit-field-scree-e46392df7c214ea388e1c4f11ee3d1c6
- https://www.notion.so/artsy/Profile-Payment-Add-new-card-error-message-doesn-t-appear-or-appears-for-only-a-split-second--2818c677b1b643c4a0527199d897593d

### Description

These bugs were triggered when react-native changed the way `Modal#onDismiss` is invoked. It used to be called whenever the native modal was hidden (controlled by the `visible` prop), and now it is only called when the react component unmounts. https://github.com/facebook/react-native/commit/bd2b7d6c0366b5f19de56b71cb706a0af4b0be43

On this edit screen we use the `onDismiss` callback to trigger alerts or navigation actions safely (these can cause bugs if triggered before the modal has a chance to be removed from the native view hierarchy).

So the way to fix this was to make sure that we unmount the loading modal component when we don't want to show it, triggering the callback at the appropriate time.

I also cleaned up some related code before figuring that out:

- Use the 'alert' callback in the credit card form, to opt in to the special `onDismiss` behaviour.
- Make our own native nav code more robust so that it works even if there's a `Modal` being shown.

TODO:

- [x] QA modal navigation in the whole app

### DIFF
--- a/src/lib/Scenes/MyAccount/Components/MyAccountFieldEditScreen.tsx
+++ b/src/lib/Scenes/MyAccount/Components/MyAccountFieldEditScreen.tsx
@@ -1,6 +1,6 @@
 import LoadingModal from "lib/Components/Modals/LoadingModal"
 import { PageWithSimpleHeader } from "lib/Components/PageWithSimpleHeader"
-import SwitchBoard from "lib/NativeModules/SwitchBoard"
+import { goBack } from "lib/navigation/navigate"
 import { useScreenDimensions } from "lib/utils/useScreenDimensions"
 import { Sans } from "palette"
 import React, { useImperativeHandle, useRef, useState } from "react"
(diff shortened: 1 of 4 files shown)


src/lib/Scenes/MyProfile/MyProfilePaymentNewCreditCard.tsx
```diff
@@ -113,7 +112,7 @@ export const MyProfilePaymentNewCreditCard: React.FC<{}> = ({}) => {
       ref={screenRef}
       canSave={state.allPresent}
       title="Add new card"
-      onSave={async (dismiss) => {
+      onSave={async (dismiss, alert) => {
```
MounirDhahri, line 115: 🙌

src/lib/Scenes/MyAccount/Components/MyAccountFieldEditScreen.tsx
```diff
@@ -105,17 +96,17 @@ export const MyAccountFieldEditScreen = React.forwardRef<
...
-            onDismiss={() => {
-              if (behindTheModalDismiss) {
-                onDismiss()
-              }
-              if (behindTheModalAlert !== undefined) {
-                Alert.alert(...behindTheModalAlert)
-              }
-            }}
-          />
+          {!!isSaving && (
```
MounirDhahri, line 99: 👏

MounirDhahri (approved):

PR #4165, opened 2020-11-23
fix: Move first inquiry message trim

jo-rs:

### Description

Moved the trim to the submit to avoid it getting triggered on every keyup. Otherwise, every time the user would put in a space between words the trim would trigger.


src/lib/Scenes/Artwork/Components/Mutation/SubmitInquiryRequest.ts
```diff
@@ -40,7 +40,7 @@ export const SubmitInquiryRequest = (
         inquireableID: inquireable.internalID,
         inquireableType: "Artwork",
         questions: formattedQuestions,
-        message: inquiryState.message,
+        message: inquiryState.message?.trim(),
```
xtina-starr, line 43: This makes sense!

xtina-starr:
💯

PR #4173, opened 2020-11-24
chore: update colors

mdole:

### Description

Updating colors in Palette, so needed to update here too! (Thanks @zephraph for working towards a unified tokens package 😉) https://github.com/artsy/palette/pull/819

- [?] I have included screenshots or videos to illustrate my changes, or I have not changed anything that impacts the UI.


Podfile.lock
```diff
@@ -882,7 +882,7 @@ SPEC CHECKSUMS:
   CocoaLibEvent: 2fab71b8bd46dd33ddb959f7928ec5909f838e3f
   CocoaLumberjack: aa9dcab71bdf9eaf2a63bbd9ddc87863efe45457
   DHCShakeNotifier: 64048427ecaa763f2472d0032f58bf7a10074eee
-  DoubleConversion: cde416483dac037923206447da6e1454df403714
+  DoubleConversion: 5805e889d232975c086db112ece9ed034df7a0b2
```
mdole, line 885: 👋 mobile noob here - I'm not sure what caused this or if it belongs in this PR! any pointers?

pvinis: All good. Maybe we had an older version. 👍

brainbicycle (approved):

Podfile.lock
brainbicycle: 🤔 Hmm, I am not getting this when running `pod install` on master, maybe this is a difference in versions somewhere in your local setup? Cocoapods, react-native or node? I don't think we should put this in the PR but I don't think it is too harmful either.

brainbicycle: Orta has a blog explainer on the Podfile.lock and checksums and what can cause churn but not obvious to me what is causing it in this case: https://artsy.github.io/blog/2016/05/03/podspec-checksums/

pvinis: I was getting this last week, and I didn't include it in a few PRs until I did. It might be because the pod specs are in a different version, and since for these pods we don't have a pinned version, it might be pulling the different one for you two (or us three I guess). 🤷

brainbicycle: I would expect the version in the Podfile.lock to change if that were the case though but maybe someone more knowledgable on the subject understands 🤷 like I said I don't think it is a blocker at worst it will cause a bit of churn in other PRs

PR #4176, opened 2020-11-25
chore: upgrade typescript

ds300:
### Description

The main breaking change was that function signatures like  `blah(arg: any): whatever` can no longer be called without an explicit argument like `blah()`. You have to explicitly declare the parameter as having type `void`.

This was so painless 🙌 


ashfurrow (approved):
Awesome! Nice and easy – thanks David!

PR #4179, opened 2020-11-27
feat(my collection): allow multiple images for an artwork

admbtlr:

This PR resolves [CX-782]

### Description

Makes is possible to add multiple images to a My Collection Artwork, and adds a screen to show all the images, as per the mockups.

image

image

Along the way I discovered that there are two bugs relating to artwork images that need fixing:

1. It's not possible to remove artwork images [CX-851](https://artsyproduct.atlassian.net/browse/CX-851)
2. Uploading multiple images in one go sometimes fails because of rate limiting [CX-850](https://artsyproduct.atlassian.net/browse/CX-850)

[CX-782]: https://artsyproduct.atlassian.net/browse/CX-782


src/lib/Scenes/MyCollection/State/MyCollectionArtworkModel.tsx
```diff
@@ -123,12 +131,15 @@ export const MyCollectionArtworkModel: MyCollectionArtworkModel = {
...
+    state.sessionState.formValues.photos = uniqBy(
+      state.sessionState.formValues.photos.concat(photos),
+      (photo) => photo.url || photo.path
+    )
   }),
 
   removePhoto: action((state, photoToRemove) => {
     state.sessionState.formValues.photos = state.sessionState.formValues.photos.filter(
-      (photo) => photo.path !== photoToRemove.path
+      (photo) => photo.path !== photoToRemove.path || photo.url !== photoToRemove.url
```
damassi, line 142: Curious why we need to now check by `path` and `url`? Is there any way to land on one or the other across the board so that the additional logic branch is no longer necessary?

src/lib/Scenes/MyCollection/State/MyCollectionArtworkModel.tsx
```diff
@@ -180,7 +191,7 @@ export const MyCollectionArtworkModel: MyCollectionArtworkModel = {
       artistSearchResult: {
         internalID: artwork?.artist?.internalID,
         displayLabel: artwork?.artistNames,
-        imageUrl: artwork?.image?.url?.replace(":version", "square"),
+        imageUrl: artwork?.images?.length > 0 && artwork.images[0].url?.replace(":version", "square"),
```
damassi, line 194: Can also do 
```
artwork.images?.[0].url?
```
which will allow one to skip the initial length check

src/lib/Scenes/MyCollection/Screens/ArtworkImages/MyCollectionArtworkImages.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { ActivityIndicator, ScrollView, View } from "react-native"
+import { createFragmentContainer, graphql, QueryRenderer } from "react-relay"
+
+interface MyCollectionArtworkImagesProps {
+  artwork: MyCollectionArtworkImages_artwork
+}
+
+const MyCollectionArtworkImages: React.FC<MyCollectionArtworkImagesProps> = ({ artwork }) => {
+  const { images } = artwork
+  // const { navigation: navActions } = AppStore.actions.myCollection
```
damassi, line 19: This can go now that we no longer have a navigation model

src/lib/Scenes/MyCollection/Screens/ArtworkFormModal/Screens/MyCollectionArtworkFormAddPhotos.tsx
```diff
@@ -8,7 +8,7 @@ import { chunk } from "lodash"
 import { AddIcon, BorderBox, Box, color, Flex, XCircleIcon } from "palette"
 import React from "react"
 import { Image, ScrollView, TouchableOpacity } from "react-native"
-import { Image as ImageProps } from "react-native-image-crop-picker"
+import { Image as ImageProps } from "../../../State/MyCollectionArtworkModel"
```
damassi, line 11: #minor - can we make this path absolute? Generally one level `../` is fine, but when nested further refactoring can get a little tricky

src/lib/Scenes/MyCollection/Screens/Artwork/Components/__tests__/MyCollectionArtworkHeader-tests.tsx
```diff
@@ -35,25 +38,41 @@ describe("MyCollectionArtworkHeader", () => {
     mockEnvironment = createMockEnvironment()
   })
 
-  it("renders without throwing an error", () => {
-    const wrapper = renderWithWrappers(<TestRenderer />)
+  const getWrapper = (mockResolvers = {}) => {
+    const tree = renderWithWrappers(<TestRenderer />)
+    act(() => {
```
damassi, line 43: Don't think `act` is necessary here

src/lib/Scenes/MyCollection/Screens/ArtworkImages/MyCollectionArtworkImages.tsx
```diff
@@ -0,0 +1,84 @@
...
+            />
+            <Spacer my={1} />
+          </View>
+        ))}
+      </ScrollView>
+    </>
+  ) : null
+}
+
+export const MyCollectionArtworkImagesFragmentContainer = createFragmentContainer(MyCollectionArtworkImages, {
```
damassi, line 42: Does this need to be exported? Or are we only using the `QueryRenderer`?

src/lib/Scenes/MyCollection/Screens/ArtworkFormModal/Screens/MyCollectionArtworkFormAddPhotos.tsx
```diff
@@ -25,7 +25,10 @@ export const MyCollectionAddPhotos: React.FC<StackScreenProps<ArtworkFormModalSc
     photos.map((photo, index) => {
       return (
         <Box key={index}>
-          <Image style={{ width: imageSize, height: imageSize, resizeMode: "cover" }} source={{ uri: photo.path }} />
+          <Image
+            style={{ width: imageSize, height: imageSize, resizeMode: "cover" }}
+            source={{ uri: photo.url || photo.path }}
```
damassi, line 30: Mentioned elsewhere, but it would be nice if the overall `photo` API was narrowed down to just one option

src/lib/Scenes/MyCollection/Screens/ArtworkImages/MyCollectionArtworkImages.tsx
```diff
@@ -0,0 +1,84 @@
...
+interface MyCollectionArtworkImagesProps {
+  artwork: MyCollectionArtworkImages_artwork
+}
+
+const MyCollectionArtworkImages: React.FC<MyCollectionArtworkImagesProps> = ({ artwork }) => {
+  const { images } = artwork
+  // const { navigation: navActions } = AppStore.actions.myCollection
+  const dimensions = useScreenDimensions()
+
+  return images !== null ? (
```
damassi, line 22: #superminor - its a bit more readable for future devs if one exits early 

```tsx
if (!images) {
  return null
}

return (
  <>
    <FancyModalHeader>Artwork Images</FancyModalHeader>
    <ScrollView>
       ...
    </ScrollView>
  </>
)
```

src/lib/Scenes/MyCollection/Screens/ArtworkImages/MyCollectionArtworkImages.tsx
```diff
@@ -0,0 +1,84 @@
...
+    <>
+      <FancyModalHeader>Artwork Images</FancyModalHeader>
+      <ScrollView>
+        {images.map((image: any, index: number) => (
+          <View key={index}>
+            <OpaqueImageView
+              // TODO: figure out if "normalized" is the correct version
```
damassi, line 29: This comment has been in since the initial static POC, we should find the answer and address in this PR

src/lib/Scenes/MyCollection/State/MyCollectionArtworkModel.tsx
```diff
@@ -3,10 +3,18 @@ import { AutosuggestResult } from "lib/Scenes/Search/AutosuggestResults"
 import { GlobalStoreModel } from "lib/store/GlobalStoreModel"
 import { uniqBy } from "lodash"
 import { ActionSheetIOS } from "react-native"
-import ImagePicker, { Image } from "react-native-image-crop-picker"
+import ImagePicker from "react-native-image-crop-picker"
 
 import { Metric } from "../Screens/ArtworkFormModal/Components/Dimensions"
 
+export interface Image {
```
damassi, line 10: Wondering - why not `extends Image`? (vs completely replacing `react-native-image-crop-picker`'s `Image` interface )

admbtlr: This is probably the best place to answer your questions above about `path` vs `url`, plus the question about extending `Image` from `react-native-image-picker`: 

There are actually two different kinds of `Image` in play here, local images (that have been selected from the camera roll) or remote images (that have already been uploaded). Local images have a path, since they only exist on the local filesystem; remote images have a URL, since they only exist on Gemini. It's awkward to have two types coexisting like this, but since they are both treated in exactly the same way within the UI, it actually makes things a lot more readable.

I thought about extending the `ImagePicker` version of `Image`, but ultimately we don't actually need any of its fields apart from `path`, and this only until the image is uploaded to Gemini, so it seemed unhelpful. But I'm happy to reconsider if you have a clearer way of organising the types.

src/lib/Scenes/MyCollection/State/MyCollectionArtworkModel.tsx
admbtlr: Oh, nice! (although a little weird to read)

src/lib/Scenes/MyCollection/Screens/ArtworkFormModal/Screens/MyCollectionArtworkFormAddPhotos.tsx
admbtlr: I feel like it makes sense to keep this relative because its referring to a file within the same "package", i.e. `MyCollection`. Like this it's possible to move `MyCollection` around within the filesystem without having to adjust any internal import paths.

src/lib/Scenes/MyCollection/Screens/ArtworkImages/MyCollectionArtworkImages.tsx
admbtlr: Yup, good call.

src/lib/Scenes/MyCollection/Screens/ArtworkFormModal/Screens/MyCollectionArtworkFormAddPhotos.tsx
admbtlr: (see answer below)

src/lib/Scenes/MyCollection/State/MyCollectionArtworkModel.tsx
admbtlr: (see answer below)

src/lib/Scenes/MyCollection/Screens/ArtworkImages/MyCollectionArtworkImages.tsx
```diff
@@ -0,0 +1,84 @@
...
+              imageURL={image.url?.replace(":version", "normalized")}
+              height={(dimensions.width / (image.width || 1)) * (image.height || 1)}
+              width={dimensions.width}
+            />
+            <Spacer my={1} />
+          </View>
+        ))}
```
ds300, line 36: It's a bit safer to avoid the `any` here and use the infix `!` operator.

```suggestion
        {images.map((image, index) => (
          <View key={index}>
            <OpaqueImageView
              // TODO: figure out if "normalized" is the correct version
              imageURL={image!.url?.replace(":version", "normalized")}
              height={(dimensions.width / (image!.width || 1)) * (image!.height || 1)}
              width={dimensions.width}
            />
            <Spacer my={1} />
          </View>
        ))}
```

ds300 (approved):
This is really good stuff Adam! 🎉 I had one non-blocking comment, feel free to merge when ready.

src/lib/Scenes/MyCollection/Screens/Artwork/Components/MyCollectionArtworkHeader.tsx
```diff
@@ -12,11 +15,15 @@ interface MyCollectionArtworkHeaderProps {
...
   const {
-    artwork: { artistNames, date, image, title },
+    artwork: { artistNames, date, images, internalID, title },
   } = props
   const dimensions = useScreenDimensions()
   const formattedTitleAndYear = [title, date].filter(Boolean).join(", ")
 
+  const defaultImage = images?.find((i: any) => i.isDefault) || (images && images[0])
+
+  const isImage = (toCheck: any): toCheck is Image => !!toCheck
```
MounirDhahri, line 25: **praise:** Cool, I haven't used type guards before

src/lib/Scenes/MyCollection/Screens/Artwork/Components/MyCollectionArtworkHeader.tsx
```diff
@@ -12,11 +15,15 @@ interface MyCollectionArtworkHeaderProps {
 
 const MyCollectionArtworkHeader: React.FC<MyCollectionArtworkHeaderProps> = (props) => {
   const {
-    artwork: { artistNames, date, image, title },
+    artwork: { artistNames, date, images, internalID, title },
   } = props
   const dimensions = useScreenDimensions()
   const formattedTitleAndYear = [title, date].filter(Boolean).join(", ")
 
+  const defaultImage = images?.find((i: any) => i.isDefault) || (images && images[0])
```
MounirDhahri, line 23: **thought:** I think we should remove the any here since apparently, images can contain null values and use `i?.isDefault ` instead of `i.isDefault `

src/lib/Scenes/MyCollection/Screens/Artwork/Components/MyCollectionArtworkHeader.tsx
```diff
@@ -28,12 +35,38 @@ const MyCollectionArtworkHeader: React.FC<MyCollectionArtworkHeaderProps> = (pro
...
-      <OpaqueImageView
-        // TODO: figure out if "normalized" is the correct version
-        imageURL={image?.url?.replace(":version", "normalized")}
-        height={300}
-        width={dimensions.width}
-      />
+      <TouchableOpacity onPress={() => navigate(`/my-collection/artwork-images/${internalID}`)}>
+        {!!isImage(defaultImage) && (
+          <OpaqueImageView
+            // TODO: figure out if "normalized" is the correct version
```
MounirDhahri, line 41: Let's discuss this with @samchieng

src/lib/Scenes/MyCollection/Screens/Artwork/Components/__tests__/MyCollectionArtworkHeader-tests.tsx
```diff
@@ -35,25 +38,41 @@ describe("MyCollectionArtworkHeader", () => {
...
+      mockEnvironment.mock.resolveMostRecentOperation((operation) =>
+        MockPayloadGenerator.generate(operation, mockResolvers)
+      )
+    })
+    return tree
+  }
```
MounirDhahri, line 49: **thought:** you can maybe use `mockEnvironmentPayload` here, it is placed in `lib/tests/mockEnvironmentPayload`

MounirDhahri (approved):

src/lib/Scenes/MyCollection/Screens/ArtworkImages/MyCollectionArtworkImages.tsx
admbtlr: Good catch!

src/lib/Scenes/MyCollection/Screens/Artwork/Components/__tests__/MyCollectionArtworkHeader-tests.tsx
admbtlr: Good call, didn't know about that!

src/lib/Scenes/MyCollection/Screens/ArtworkFormModal/Screens/MyCollectionArtworkFormAddPhotos.tsx
damassi: > I feel like it makes sense to keep this relative because its referring to a file within the same "package"

This makes assumptions about code over time, including the overall file structure of the package. With absolute paths no assumptions need to be made and it helps with maintenance at scale. Looking around Eigen over time, it's increasingly rare to see relative imports greater than one level deep. 

@ds300 might be able to chime in on this having done so much refactoring

src/lib/Scenes/MyCollection/State/MyCollectionArtworkModel.tsx
damassi: Makes sense 👍

src/lib/Scenes/MyCollection/Screens/ArtworkFormModal/Screens/MyCollectionArtworkFormAddPhotos.tsx
ds300: VSCode automatically updates absolute imports when moving files around, so that's not an issue for me personally. And yes I've found inconsistent relative imports occasionally cause friction when doing manual find+replace.

I'd be in favour of a TSLint rule enforcing absolute imports everywhere, but I fear that proposing such a thing might cause more discussion than it warrants.

damassi: > VSCode automatically updates absolute imports when moving files around

I've gotten into some confusing trouble with this setting in the past so I have it disabled. So its likely inconsistent between devs. 

But yeah, either way, this is totally non-blocking! I do think we should call out deeply nested relative imports when we can though for said friction reasons.

admbtlr:
#squashongreen

PR #4184, opened 2020-11-30
fix: avoid shrinking the background on non-fullscreen fancy modals

ds300:
This PR resolves [CX-765]

### Description

This ticket makes it so that half-page (non-fullscreen) fancy modals do not cause their backgrounds to shrink.

Fancy Modal is implemented as a stack of cards, of arbitrary depth. Before this change each card would shrink to a linear degree when it was below the top card. Now not all cards shrink relative to the card above, so it's no longer a linear calculation. Now we need to look at the whole stack and kinda merge the cards that don't shrink into the cards above them.

no-shrinky

[CX-765]: https://artsyproduct.atlassian.net/browse/CX-765

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -17,6 +17,7 @@ upcoming:
     - Fix device token checking logic - david
   user_facing:
     - Add loading spinner to infinite artwork grid on artist page - david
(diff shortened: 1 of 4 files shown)


MounirDhahri (approved):

MounirDhahri:
LGTM 🎉

PR #4185, opened 2020-11-30
tooling(relay-tests): add default mock resolvers and an array generator and some helpers

pvinis:

This PR resolves [CX-]

### Description

The motivation for this PR was me wanting to test without having to fill all the required fields, and ideally no fields at all. I guess it's kinda RFC too?

- I added `DefaultMockResolvers` for `ID` and `String` fields, so that they generate mock data like 
```
{
  id: "id-1",
  title: "title-1"
}
```

- One extra thing here is that if we have something with possible duplicate names, it will fix it like so:
```
{
  artist: {
    id: "id-1"
    name: "name-1",
    artworks: {
      edges: [
        { node: {
          id: "artworks.id-1",
          slug: "slug-1"
        }},
      ]
    }
  }
}
```

- Also, in an `edges` array, this will generate this like `id-1`, `title-1` for the first item, `id-2`, `title-2` for the second etc.

- Lastly, I added a `mockArray` that takes a number for the length of the array to generate. Normally only one item is generated, so when we need to test for empty edges, `edges: []`, when we need 3 edges, `edges: mockArray(3)` and it will generate 3 items with the fields filled in as above.

Of course, the ability to override any of the autogenerated fields is possible. I have three test files I touched here, so show how all possibilities would work. I can follow up with another PR to change more files if we like this way.

### DIFF
--- a/patches/@types+relay-test-utils+6.0.1.patch
+++ b/patches/@types+relay-test-utils+6.0.1.patch
@@ -1,0 +1,13 @@
+diff --git a/node_modules/@types/relay-test-utils/lib/RelayMockPayloadGenerator.d.ts b/node_modules/@types/relay-test-utils/lib/RelayMockPayloadGenerator.d.ts
+index 04d356e..89c707c 100644
+--- a/node_modules/@types/relay-test-utils/lib/RelayMockPayloadGenerator.d.ts
++++ b/node_modules/@types/relay-test-utils/lib/RelayMockPayloadGenerator.d.ts
+@@ -10,7 +10,7 @@ import {
+     NormalizationSplitOperation,
+ } from 'relay-runtime';
+ 
+-interface MockResolverContext {
++export interface MockResolverContext {
+     parentType?: string;
+     name?: string;
+     alias?: string;
--- a/src/lib/tests/mockEnvironmentPayload.ts
+++ b/src/lib/tests/mockEnvironmentPayload.ts
@@ -1,3 +1,4 @@
+import { takeRight } from "lodash"
 import { createMockEnvironment, MockPayloadGenerator } from "relay-test-utils"
 import { MockResolverContext, MockResolvers } from "relay-test-utils/lib/RelayMockPayloadGenerator"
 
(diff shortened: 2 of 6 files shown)


src/lib/tests/mockEnvironmentPayload.ts
```diff
@@ -1,7 +1,57 @@
...
+ * Used to generate results in an array.
+ * Usage: ```
+ * Artist: () => ({
+ *   auctionResultsConnection: {
+ *     edges: mockArray(10), // <- this will generate an array with 10 results, all prefilled
+ *   }
+ * })
+ * ```
+ */
+export const mockArray = (length: number) => new Array(length).fill({ node: {} })
```
ds300, line 26: `mockEdges` might be a better name for this, since it's specific to connections rather than arrays.

src/lib/tests/mockEnvironmentPayload.ts
```diff
@@ -1,7 +1,57 @@
+import { takeRight } from "lodash"
 import { createMockEnvironment, MockPayloadGenerator } from "relay-test-utils"
+import { MockResolverContext, MockResolvers } from "relay-test-utils/lib/RelayMockPayloadGenerator"
 
-export type RelayMockEnvironment = ReturnType<typeof createMockEnvironment>
+let counters: { [path: string]: number } = {}
+const resetCounters = () => {
+  counters = {}
```
ds300, line 7: should you also reset the `paths` object here? Otherwise you might get nondeterministic behaviour based on the order that tests run in? I could be misreading the `goodMockResolver` code 🤔

ds300 (approved):
This looks really good, super useful, nice one! 🙏

src/lib/tests/mockEnvironmentPayload.ts
pvinis: Not a bad idea 👍.

pvinis:
#squashongreen

damassi (approved):
Nice, will bring this one over to force 👍 💯

patches/@types+relay-test-utils+6.0.1.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node_modules/@types/relay-test-utils/lib/RelayMockPayloadGenerator.d.ts b/node_modules/@types/relay-test-utils/lib/RelayMockPayloadGenerator.d.ts
+index 04d356e..89c707c 100644
+--- a/node_modules/@types/relay-test-utils/lib/RelayMockPayloadGenerator.d.ts
++++ b/node_modules/@types/relay-test-utils/lib/RelayMockPayloadGenerator.d.ts
+@@ -10,7 +10,7 @@ import {
+     NormalizationSplitOperation,
+ } from 'relay-runtime';
+ 
+-interface MockResolverContext {
++export interface MockResolverContext {
```
damassi, line 10: @pvinis - it seems self-evident, but can we update the HACKS.md file with this patch addition?

pvinis: Yup, sorry I forgot.

PR #4189, opened 2020-11-30
Enhancement: Update to use palette tokens

zephraph:

### Description

Since palette was vendored there's been drift in the base token values. This attempts to integrate the base tokens from a separate package. This _should_ be good to go. I haven't been able to test it like I'd like, so I'd definitely recommend giving it a run on the simulator. The typography integration in particular was _quite_ gnarly to get working and I suspect someone fresher than I am could come through and add some major improvements. 

I know there's a lot of file changes here, but _most_ of them are icons that have had their helper references updated. The helper types weren't quite right so I fixed them (but I also moved those definitions upstream to the tokens package b/c their unlikely to change and more broadly useful). 


ds300:
Thanks for taking care of this @zephraph ❤️ 🎉 sorry it took a while to get merged.

PR #4195, opened 2020-12-02
fix(my collection): save changes button state #trivial

admbtlr:

This PR resolves [CX-772]

### Description

This is a bug that only happens when running directly on a device (i.e. not when debugging on the device). I'm guessing that this is some problem with JavascriptCore, the iOS JS engine, although I tried debugging with Safari, which usually picks up problems specific to JavascriptCore. So I think we should try merging this, and then test it once it's been deployed to a beta.

[CX-772]: https://artsyproduct.atlassian.net/browse/CX-772


ds300 (approved):
I guess this will work as long as there are no nested fields. Fine by me 👍

PR #4200, opened 2020-12-02
chore: Remove first inquiry flag

lilyfromseattle:
[PURCHASE-2240]

This PR removes the feature flag for the new inquiry flow, so that we can use an echo flag instead (added [here](https://github.com/artsy/echo/pull/57))

May need to hold off on merging until the echo PR is merged

[PURCHASE-2240]: https://artsyproduct.atlassian.net/browse/PURCHASE-2240

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -6,6 +6,7 @@ upcoming:
     - Allow for multiple artwork images in My Collection - adam
     - Set price and marketing text dynamically - adam
     - Migrate floor selling price and online asking price in My Bids to the new type - starsirius
+    - remove new inquiry flow feature flag in favor of echo flag - lily
   user_facing:
     - Add artist past shows screen - mounir
     - Add loading spinner to infinite artwork grid on artist page - david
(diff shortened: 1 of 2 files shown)


CHANGELOG.yml
```diff
@@ -6,6 +6,7 @@ upcoming:
     - Allow for multiple artwork images in My Collection - adam
     - Set price and marketing text dynamically - adam
     - Migrate floor selling price and online asking price in My Bids to the new type - starsirius
+    - remove new inquiry flow feature flag in favor of echo flag - lily
```
zephraph, line 9: One of those small nuances where this is technically what you did, but from the perspective of someone consuming the changelog it may be more helpful to describe what it means. 

```suggestion
    - Enable new fist inquiry flow - lily
```

zephraph (approved):
This looks good to me. A small note on the changelog, but I don't feel strongly about it. Let's wait until the echo PR is merged and deployed before merging this though.

xtina-starr:
To make sure I understand this correctly, the `newFirstInquiry` property will now be controlled by echo once this gets merged?

https://github.com/artsy/eigen/blob/56c1edfa8970919594254e95030ceb6a8f7d33ba/src/lib/Scenes/Artwork/Components/CommercialButtons/CommercialButtons.tsx#L40

https://github.com/artsy/eigen/blob/56c1edfa8970919594254e95030ceb6a8f7d33ba/src/lib/Scenes/Artwork/Components/CommercialButtons/CommercialButtons.tsx#L90-L92

PR #4204, opened 2020-12-02
[PURCHASE-2249] First inquiry visual qa

xtina-starr:

Screen Shot 2020-12-02 at 2 32 32 PM

This PR resolves [PURCHASE-2249]

This vertically centers the chevron to the left of the 'Add your location' copy.

[PURCHASE-2249]: https://artsyproduct.atlassian.net/browse/PURCHASE-2249

### DIFF
--- a/src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
+++ b/src/lib/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
@@ -99,7 +99,7 @@ const InquiryQuestionOption: React.FC<{
                     <Text variant="text" color="black60">
                       Add your location
                     </Text>
-                    <Box mt={0.5}>
+                    <Box>
                       <ChevronIcon color="black60" />
                     </Box>
                   </>


Podfile.lock
```diff
@@ -904,7 +904,7 @@ SPEC CHECKSUMS:
   Folly: b73c3869541e86821df3c387eb0af5f65addfab4
   Forgeries: 64ced144ea8341d89a7eec9d1d7986f0f1366250
   FXBlurView: 5121730176fd52bcf7bffd0bd8c1485e8aabc3cb
-  glog: 1f3da668190260b06b429bb211bfbee5cd790c28
+  glog: 40a13f7840415b9a77023fbcae0f1e6f43192af3
```
ashleyjelks, line 907: Would the app not compile without this? I was under the impression that we should avoid Podfile.lock commits unless that's the case. cc @artsy/collector-experience maybe someone on this team clarify?

ashleyjelks (approved):

Podfile.lock
xtina-starr: Yeah that's a good question. Should these not be checked in?

xtina-starr: @zephraph do you know if this is needed?

ashleyjelks: I saw a change to this file after making some change recently, and to test my assumption, I stashed it to find that the app would not compile without it, in which case it made sense to commit the change. I've also tried that and seen the app work as expected without the new changes to this file, so I'm not sure what the best practice here is. But do you know if the app still builds without this change?

PR #4205, opened 2020-12-02
fix: handle iOS 14 photo picker UI in my collection

brainbicycle:

This PR resolves [CX-865]

### Description

iOS 14 introduces a new photo picker UI that operates outside the app process and allows users to share only those photos explicitly selected through the picker:  https://developer.apple.com/videos/play/wwdc2020/10652/ 
What it is cool about this picker is it can be shown without first requesting permissions to access to users photos. 
Alongside this change any apps that _do_ request permissions for access to a users photos will now have an option in the permissions dialog to select photos to share using the new picker. RNCImageCropPicker requests access to users photos so we are seeing this new dialog, and actually seeing both Apple's picker and then the RNCImageCropPicker. See behavior in **before** gif below. This PR first checks if a user is on iOS 14 or greater and if so shows the new photo picker, otherwise fallback to RNCImageCropPicker.

- Adds a new NativeModule for presenting the photo picker and returning photos
- Updates `RNImageCropPicker` and our associated patch file
- Adds a utility typescript method for requesting photos which falls back to image-crop-picker for other OS's and versions

#### Before

beforeiOS14

#### After

**iOS 14**

afteriOS14

**Other OS's and versions**

afteriOS12

### TODO

- [x] add tests
- [x] fix cast to any in promises
- [x] check behavior in other flows + address if necessary
        https://artsyproduct.atlassian.net/browse/CX-866 - will address consignments flow with another PR

[CX-754]: https://artsyproduct.atlassian.net/browse/CX-754

[CX-865]: https://artsyproduct.atlassian.net/browse/CX-865


emission/Pod/Classes/EigenCommunications/ARPHPhotoPickerModule.m
```diff
@@ -0,0 +1,151 @@
...
+ */
+
+@implementation ARPHPhotoPickerModule
+
+static NSString * ErrorDomain = @"net.artsy.ARPHPhotoPicker";
+
+static NSString * UnsupportedOSErrorMessage = @"PHPhotoPicker unavailable before iOS 14.";
+static NSString * NoPhotosErrorMessage = @"No photos returned from picker.";
+static NSString * LoadFailedErrorMessage = @"Failed to load photos from picker.";
+static NSString * SaveFailedErrorMessage = @"Failed to save photos locally.";
```
ashfurrow, line 31: Super-minor, but we can use `static const` here and have the `*` hug the variable name, to match Eigen's Objective-C code style.

```suggestion
static const NSString *ErrorDomain = @"net.artsy.ARPHPhotoPicker";

static const NSString *UnsupportedOSErrorMessage = @"PHPhotoPicker unavailable before iOS 14.";
static const NSString *NoPhotosErrorMessage = @"No photos returned from picker.";
static const NSString *LoadFailedErrorMessage = @"Failed to load photos from picker.";
static const NSString *SaveFailedErrorMessage = @"Failed to save photos locally.";
```

emission/Pod/Classes/EigenCommunications/ARPHPhotoPickerModule.m
```diff
@@ -0,0 +1,151 @@
...
+    dispatch_async(dispatch_get_main_queue(), ^{
+        UIViewController *currentVC = RCTPresentedViewController();
+        [currentVC presentViewController:picker animated:true completion:nil];
+    });
+}
+
+#pragma mark - PHPickerViewControllerDelegate
+
+- (void)picker:(PHPickerViewController *)picker
+didFinishPicking:(NSArray<PHPickerResult *> *)results  API_AVAILABLE(ios(14)) {
```
ashfurrow, line 63: Also minor, and maybe a Prettier thing, but we can keep these on one line to match the rest of the app

```suggestion
- (void)picker:(PHPickerViewController *)picker didFinishPicking:(NSArray<PHPickerResult *> *)results API_AVAILABLE(ios(14)) {
```

emission/Pod/Classes/EigenCommunications/ARPHPhotoPickerModule.m
```diff
@@ -0,0 +1,151 @@
...
+RCT_EXPORT_MODULE();
+
+RCT_EXPORT_METHOD(requestPhotos:(RCTPromiseResolveBlock)resolve
+                  reject:(RCTPromiseRejectBlock)reject)
+{
+    self.resolve = resolve;
+    self.reject = reject;
+    if (@available(iOS 14.0, *)) {
+        [self presentPhotoPicker];
+    } else {
```
ashfurrow, line 42: So if someone is using an older OS, they won't be able to upload photos? Probably okay to ship going forward, let's just make sure the user-facing error is descriptive 👍 I checked Looker and 83% of users are on iOS 14.

emission/Pod/Classes/EigenCommunications/ARPHPhotoPickerModule.m
```diff
@@ -0,0 +1,151 @@
...
+        UIViewController *currentVC = RCTPresentedViewController();
+        [currentVC dismissViewControllerAnimated:true completion:nil];
+    });
+
+    if (results.count == 0) {
+         NSError *noPhotosError = [NSError errorWithDomain:ErrorDomain code:ARPHPhotoPickerErrorNoPhotosReturned userInfo:@{ NSLocalizedDescriptionKey: NoPhotosErrorMessage }];
+        _reject(ErrorDomain, NoPhotosErrorMessage, noPhotosError);
+    } else {
+        NSMutableArray *images = [[NSMutableArray alloc] init];
+        Compression *compression = [[Compression alloc] init];
```
ashfurrow, line 74: Nice solution here 👍

ashfurrow (commented):
Nice work here 👍

emission/Pod/Classes/EigenCommunications/ARPHPhotoPickerModule.m
brainbicycle: The idea was to falllback to to asking permission and using the react-native-image-crop-picker in this case and only use this module if the user is on iOS 14 or above:
https://github.com/artsy/eigen/blob/deb38e3ce016974cf1a0bf810df325882dcef1f2/src/lib/utils/requestPhotos.ts#L5

brainbicycle: The idea was to falllback to to asking permission and using the react-native-image-crop-picker in this case and only use this module if the user is on iOS 14 or above:
https://github.com/artsy/eigen/blob/deb38e3ce016974cf1a0bf810df325882dcef1f2/src/lib/utils/requestPhotos.ts#L5-L14

emission/Pod/Classes/EigenCommunications/ARPHPhotoPickerModule.m
```diff
@@ -0,0 +1,150 @@
...
+
+- (void)picker:(PHPickerViewController *)picker didFinishPicking:(NSArray<PHPickerResult *> *)results API_AVAILABLE(ios(14)) {
+    dispatch_async(dispatch_get_main_queue(), ^{
+        UIViewController *currentVC = RCTPresentedViewController();
+        [currentVC dismissViewControllerAnimated:true completion:nil];
+    });
+
+    if (results.count == 0) {
+         NSError *noPhotosError = [NSError errorWithDomain:ErrorDomain code:ARPHPhotoPickerErrorNoPhotosReturned userInfo:@{ NSLocalizedDescriptionKey: NoPhotosErrorMessage }];
+        _reject(ErrorDomain, NoPhotosErrorMessage, noPhotosError);
```
ds300, line 70: does it make sense to treat this as an error? As a consumer I would expect an empty array if the user simply decided not to pick any images.

emission/Pod/Classes/EigenCommunications/ARPHPhotoPickerModule.m
```diff
@@ -0,0 +1,150 @@
...
+                        [images addObject:imageDict];
+                    }
+                    dispatch_group_leave(imageLoadGroup);
+                }];
+            }
+        }
+
+        double delayInSeconds = 3.0;
+        dispatch_time_t timeout = dispatch_time(DISPATCH_TIME_NOW, delayInSeconds * NSEC_PER_SEC);
+        intptr_t status = dispatch_group_wait(imageLoadGroup, timeout);
```
ds300, line 93: Cool, I didn't know about this! It's like `await Promise.all(...)` for the 1990s 😎 🛹

emission/Pod/Classes/EigenCommunications/ARPHPhotoPickerModule.m
```diff
@@ -0,0 +1,150 @@
...
+    });
+}
+
+#pragma mark - PHPickerViewControllerDelegate
+
+- (void)picker:(PHPickerViewController *)picker didFinishPicking:(NSArray<PHPickerResult *> *)results API_AVAILABLE(ios(14)) {
+    dispatch_async(dispatch_get_main_queue(), ^{
+        UIViewController *currentVC = RCTPresentedViewController();
+        [currentVC dismissViewControllerAnimated:true completion:nil];
+    });
```
ds300, line 66: what's happening here? will `currentVC` not be the same as `picker`?

ds300 (approved):
Looking great! Thanks for figuring this stuff out. Aside from error handling, what needs to be done before this is ready?

emission/Pod/Classes/EigenCommunications/ARPHPhotoPickerModule.m
ashfurrow: Yes! [Grand Central Dispatch](https://developer.apple.com/documentation/dispatch?language=objc) was the thing I was most excited about OS X Snow Leopard 🤩

emission/Pod/Classes/EigenCommunications/ARPHPhotoPickerModule.m
brainbicycle: We actually want the VC that is presenting the picker in this case to call dismiss on and not the picker itself, my understanding is that this helper method will return the top react vc in our app and not the picker since it is another process, it seems to be working but let me double check 😅

emission/Pod/Classes/EigenCommunications/ARPHPhotoPickerModule.m
brainbicycle: Yeah you are right it makes more sense to not be an error

src/lib/utils/requestPhotos.ts
```diff
@@ -0,0 +1,14 @@
+import { NativeModules, Platform } from "react-native"
+import ImagePicker, { Image } from "react-native-image-crop-picker"
+import { osMajorVersion } from "./hardware"
+
+export async function requestPhotos(): Promise<Image[]> {
+  if (Platform.OS === "ios" && osMajorVersion() >= 14) {
```
ashfurrow, line 6: Ohhhh this is cool 👍

typings/emission.d.ts
```diff
@@ -17,6 +18,9 @@ declare module "react-native" {
       postNotificationName(type: string, data: object): void
       didFinishBootstrapping(): void
     }
+    ARPHPhotoPickerModule: {
+      requestPhotos(): Promise<any> // why?
```
ashfurrow, line 22: Would it be possible to add a more specific type here? If we can't right now, leaving a `// TODO:` comment would be okay too 👍

emission/Pod/Classes/EigenCommunications/ARPHPhotoPickerModule.m
ashfurrow: Okay cool, that makes sense! I'd recommend we leave a comment inside this `else` block pointing the reader to `requestPhotos.ts` and explaining that, in practice, this shouldn't be executed.

ashfurrow (approved):
Great work! I don't have any blocking feedback, just a few things that would be nice to cleanup.

This is a really great example of where we choose to use Objective-C and write a native module, for the sake of a good user experience. Would be a great blog post – no pressure! But I'd be happy to help you write one up.

typings/emission.d.ts
brainbicycle: Yeah let me look into this, I am not clear why this isn't working, the types match but seems to be some conflict with react-native's Image type

PR #4207, opened 2020-12-03
fix: selecting search suggestion requires double tap (to clear keyboard)

erikdstock:
#minor

This PR resolves [PURCHASE-2250]

### Description
This PR fixes a bug where the user, when trying to tap a search suggestion, must tap twice since the first dismisses the keyboard. It also includes an optimization for autofocusing the input.

It requires adding a prop at the very top of the Artwork Scene component hierarchy, so we should be aware if there are any unusual keyboard behaviors within an artwork view that we want to avoid. I'm not aware of any.

[PURCHASE-2250]: https://artsyproduct.atlassian.net/browse/PURCHASE-2250

### DIFF
--- a/src/lib/Scenes/Artwork/Artwork.tsx
+++ b/src/lib/Scenes/Artwork/Artwork.tsx
@@ -289,6 +289,7 @@ export class Artwork extends React.Component<Props, State> {
   render() {
     return (
       <FlatList<ArtworkPageSection>
+        keyboardShouldPersistTaps="always"
         data={this.sections()}
         ItemSeparatorComponent={() => (
           <Box mx={2} my={3}>
--- a/src/lib/Scenes/Artwork/Artwork.tsx
+++ b/src/lib/Scenes/Artwork/Artwork.tsx
@@ -289,7 +289,7 @@ export class Artwork extends React.Component<Props, State> {
   render() {
     return (
       <FlatList<ArtworkPageSection>
-        keyboardShouldPersistTaps="always"
+        keyboardShouldPersistTaps="handled"
         data={this.sections()}
         ItemSeparatorComponent={() => (
           <Box mx={2} my={3}>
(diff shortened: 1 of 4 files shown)


src/lib/Scenes/Artwork/Artwork.tsx
```diff
@@ -289,6 +289,7 @@ export class Artwork extends React.Component<Props, State> {
   render() {
     return (
       <FlatList<ArtworkPageSection>
+        keyboardShouldPersistTaps="handled"
```
MounirDhahri, line 292: @erikdstock we can use handled as a value here to avoid future regressions.
_the keyboard will not dismiss automatically when the tap was handled by a children, (or captured by an ancestor)._

src/lib/Scenes/Artwork/Components/CommercialButtons/ShippingModal.tsx
```diff
@@ -36,9 +37,11 @@ export const ShippingModal: React.FC<ShippingModalProps> = (props) => {
       >
         Add Location
       </FancyModalHeader>
-      <Flex m={2} flex={1}>
-        <LocationAutocomplete onChange={setLocationInput} initialLocation={location} />
-      </Flex>
+      <ScrollView keyboardShouldPersistTaps="always">
```
erikdstock, line 40: what is the difference between `handled` above in Artwork and `always` here? Do we need both, and does this one change the behavior? The explanations for `always|never|handled` don't really make sense to me, seems like something is missing.

In any case this is my only question and I propose we merge if this is otherwise good to you.

MounirDhahri: In the `Artwork` screen I set the `ScrollView` `keyboardShouldPersistTaps` to `handled` to be able to delegate how we handle the keyboard taps to components/screens defined within that ScrollView (usually we don't need it since using nested ScrollViews is a react-native anti-pattern). And since the ShippingModal is defined within the Artwork screen I set `keyboardShouldPersistTaps` to `always` there. (I think it would have been better if we defined the InquiryModal and the ShippingModal as new screens that get presented modally instead of a modal)

MounirDhahri (approved):

src/lib/Scenes/Artwork/Components/CommercialButtons/ShippingModal.tsx
MounirDhahri: I will merge this then and if you want we can talk about how we can redefine navigation there

PR #4209, opened 2020-12-03
enhancement: Remove the 'X' to exit the confirmation modal

lilyfromseattle:
# [PURCHASE-2273]

This 'X' is intended to dismiss the modal but because this delay is short tapping the 'X' in the 'Toast' is mainly just linking me to the inbox as its a very quick tap. We should remove the 'X' to eliminate confusion. 

_It now looks like this:_ 
Screen Shot 2020-12-03 at 11 55 45 AM

[PURCHASE-2273]: https://artsyproduct.atlassian.net/browse/PURCHASE-2273


zephraph (approved):

zephraph:
Excellent, thanks @lilyfromseattle.

PR #4213, opened 2020-12-03
feat: wrap first inquiry checkbox with touchable

jo-rs:

This PR resolves [PURCHASE-2272]

### Description

This PR makes the whole area, not just the checkbox, clickable for first inquiry fields. It also adds functionality to the checkbox component to change when the `checked` prop changes.
checkbox_click

[PURCHASE-2272]: https://artsyproduct.atlassian.net/browse/PURCHASE-2272

### DIFF
--- a/src/lib/Components/Bidding/Components/Checkbox.tsx
+++ b/src/lib/Components/Bidding/Components/Checkbox.tsx
@@ -1,4 +1,4 @@
-import React, { Component } from "react"
+import React, { Component, GetDerivedStateFromProps } from "react"
 import { StyleSheet, TouchableWithoutFeedback, TouchableWithoutFeedbackProps } from "react-native"
 import styled from "styled-components/native"
 
@@ -17,12 +17,13 @@ interface CheckboxProps extends TouchableWithoutFeedbackProps, FlexProps {
 }
 
 export class Checkbox extends Component<CheckboxProps, CheckboxState> {
-  static getDerivedStateFromProps(props: CheckboxProps, state: CheckboxState) {
+  static getDerivedStateFromProps: GetDerivedStateFromProps<CheckboxProps, CheckboxState> = (props, state) => {
     if (props.checked !== state.checked) {
       return {
         checked: props.checked,
       }
     }
+    return {}
   }
 
   private readonly checkboxSize = 20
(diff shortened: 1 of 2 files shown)


src/lib/Components/Bidding/Components/Checkbox.tsx
```diff
@@ -10,13 +10,22 @@ interface CheckboxState {
...
 
 export class Checkbox extends Component<CheckboxProps, CheckboxState> {
+  static getDerivedStateFromProps: GetDerivedStateFromProps<CheckboxProps, CheckboxState> = (props, state) => {
+    if (props.checked !== state.checked) {
+      return {
+        checked: props.checked,
+      }
+    }
+    return {}
+  }
```
jo-rs, line 27: Not sure if this is frowned upon, and I should've changed the component to work with `useEffect`.

src/lib/Components/Bidding/Components/Checkbox.tsx
```diff
@@ -10,13 +10,22 @@ interface CheckboxState {
   checked: boolean
 }
 
-interface CheckboxProps extends TouchableWithoutFeedbackProperties, FlexProps {
+interface CheckboxProps extends TouchableWithoutFeedbackProps, FlexProps {
```
jo-rs, line 13: Changed because of deprecation

src/lib/Components/Bidding/Components/Checkbox.tsx
```diff
@@ -1,5 +1,5 @@
-import React, { Component } from "react"
```
zephraph, line 1: Random question not related to your changes...

I hadn't noticed we're using a checkbox in `/Components/Bidding`. Should we be? 🤔

jo-rs: I'm not sure 😬
It doesn't seem like Pallette's checkbox has made it's way to Eigen yet.

src/lib/Components/Bidding/Components/Checkbox.tsx
zephraph: Using `getDerivedStateFromProps` is fine. I don't think we necessarily have to refactor to a functional component for just this change. Given that this is a bidding button and is used in different contexts, are we sure that we're not accidentally breaking things?

jo-rs: Hmm it shouldn't, but let me double check to be safe. Stand by.

jo-rs: I checked all existing uses, and there was an issue when `checked` is not passed as a prop. All current uses are working fine after my last push. Good catch!

PR #4215, opened 2020-12-03
Bugfix: Conversation does not move to the top of the inbox when a gallery responds

lilyfromseattle:

[PURCHASE-2275]

The conversation order does not update on refresh - for example, a convo with a new reply from the gallery doesn't appear at the top of the inbox. This PR fixes that.

After struggling with this for hours, I paired with @ashleyjelks who solved it in 10 minutes!

UPDATE: Ended up changing my implementation entirely after realizing that leftover `activeBids` stuff in `Inbox` was preventing the existing refresh logic from working

[PURCHASE-2275]: https://artsyproduct.atlassian.net/browse/PURCHASE-2275

### DIFF
--- a/src/lib/Scenes/Inbox/Components/Conversations/Conversations.tsx
+++ b/src/lib/Scenes/Inbox/Components/Conversations/Conversations.tsx
@@ -22,20 +22,13 @@ interface Props {
 
 interface State {
   isLoading?: boolean
+  isRefreshing: boolean
 }
 
 export class Conversations extends Component<Props, State> {
   state = {
     isLoading: false,
-  }
-
-  componentDidUpdate() {
-    const { relay } = this.props
-    relay.refetchConnection(PAGE_SIZE, (error) => {
-      if (error) {
-        throw new Error("Conversations error: " + error.message)
-      }
-    })
+    isRefreshing: false,
   }
 
   fetchData = () => {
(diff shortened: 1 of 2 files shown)


src/lib/Scenes/Inbox/Components/Conversations/Conversations.tsx
```diff
@@ -77,6 +86,8 @@ export class Conversations extends Component<Props, State> {
       return null
     }
 
+    console.log("CONVOS12", conversations)
```
sepans, line 89: 🙈

src/lib/Scenes/Inbox/Components/Conversations/Conversations.tsx
```diff
@@ -29,6 +29,15 @@ export class Conversations extends Component<Props, State> {
     isLoading: false,
   }
 
+  componentDidUpdate() {
```
sepans, line 32: I am trying to understand why this is needed, and if this might create an infinite loop? 🤔 

So as I understood, this re-fetches the query when the props/state change. Which exact prop/state are we interested in? Can we narrow down with an if block that compares the new value to the old value for that specific prop/state item? And if the refetch updates `me` would it trigger the refetch again? (and again?)

sepans (commented):
This is a smart fix but I didn't exactly understand the underlying reason for that. Could you explain a bit?

zephraph (approved):
The are a few minor things we could touch up (like removing `isRefreshing` from the state), but I'm good to follow that with the other PR you're working on. 👍

PR #4221, opened 2020-12-04
Build Mechanism for refreshing My Bids view

lilyfromseattle:
[PURCHASE-2173]

This story was initially. a spike, but I had a pretty good idea of how to implement it so after consulting with @zephraph I just went ahead and did it.

We may want to move the refresh container beneath the upper tabs in the future, but will wait until design looks it over.

This PR was branched from my PR fixing the conversations refresh, as they both made changes to the. same function in `Inbox` and I wanted to avoid a tricky merge. This PR also includes a small update to remove some unused code I forgot to remove from that PR. 

[PURCHASE-2173]: https://artsyproduct.atlassian.net/browse/PURCHASE-2173

### DIFF
--- a/src/lib/Containers/Inbox/Inbox.tsx
+++ b/src/lib/Containers/Inbox/Inbox.tsx
@@ -29,6 +29,9 @@ export class Inbox extends React.Component<Props, State> {
   // @ts-ignore STRICTNESS_MIGRATION
   conversations: ConversationsRef
 
+  // @ts-ignore STRICTNESS_MIGRATION
+  myBids: MyBidsRef
+
   state = {
     fetchingData: false,
     inquiryTabIsSelected: false,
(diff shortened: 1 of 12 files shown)


src/lib/Containers/Inbox/Inbox.tsx
```diff
@@ -67,6 +70,10 @@ export class Inbox extends React.Component<Props, State> {
       this.conversations.refreshConversations(() => {
         this.setState({ fetchingData: false })
       })
+    } else if (this.myBids) {
+      this.myBids.refreshMyBids(() => {
+        this.setState({ fetchingData: false })
+      })
```
erikdstock, line 76: We talked about this in (zoom) person, but to reiterate here, I think we should ticket work to adapt this new inbox to a modern react style and attempt to add pull-refresh tests only at that point. What we have now is following the current patterns of what to test and it's not worth figuring out testing for a mechanism that has been in production for a long time.

src/lib/Scenes/MyBids/__tests__/MyBids-tests.tsx
```diff
@@ -91,38 +90,6 @@ describe(MyBidsQueryRenderer, () => {
     expect(closedLots[2]).toContain("Closed Aug 13")
   })
 
-  it("renders a completed lot in an ongoing live sale in the 'active' column", () => {
```
erikdstock, line 94: 🔪 duplicate tests 👯‍♂️

erikdstock (approved):
✅ from me.

PR #4228, opened 2020-12-07
fix: check user is logged in before registering token, clear out on f…

brainbicycle:

### Description

Sailthru is causing didRegister callback to be called on app launch before a user is logged in, the token is then not correctly associated with the user on our backend and we don't attempt to register again because of the newly added previous token check. 

This checks that there is a current user before attempting to save the token on the backend and also clears out the token on error saving or if the callback was called without a user to be sure the token is saved the next time around.


ashfurrow (approved):
Ah, sneaky! Good catch.

PR #4229, opened 2020-12-07
upgrade(deps): mapbox

pvinis:
The type of this PR is: **Upgrade**

This PR resolves [CX-]

### Description

When looking into the city guide crash, one of the things I tried was to upgrade the mapbox dep. Since I did the work (even though it wasn't the source of the crash), I thought I'll make a PR for it.

One thing that changes in this version is that to download the pod we need to have a `~/.netrc` file with some mapbox credentials. I created the token and added it on 1pass. Instead of forcing all devs to add this file, I added a preinstall and postinstall on the pods, to add the key in that file (or create it if it's not there), put the credentials from `.env.shared`, then remove the entry or the file again.

To test this works, edit your local `.env.shared` to have the below line from the follow-up work, and then `bundle e pod install`. Everything should work smoothly.

### Follow-up work

- [x] Update `.env.shared` file on s3 with `MAPBOX_DOWNLOAD_TOKEN=` and the token from 1Password `Eigen mapbox cocoapod pod install`


pvinis:
Actually, after trying to make the type errors go away, I am not sure if it's better to pair with someone on them (maybe someone that has touched city guide data before), or if it's not worth it and we should just keep the old version (even though sooner or later we would need to do that).

Open to suggestions 🤷.

Podfile
```diff
@@ -77,7 +116,7 @@ target 'Artsy' do
   pod 'Artsy+UILabels'
   pod 'Extraction'
 
-  pod 'Emission', path: './emission'
+  pod 'Emission', path: './emission', :inhibit_warnings => false
```
ashfurrow, line 119: 🙌

ashfurrow (approved):
I'd say that this would be a good opportunity to share some knowledge around city guides. I think I'm the only iOS engineer left who worked on it originally, and there are only 16 type errors, so this could be useful in a KS 👍 That's just my take, though.

HACKS.md
```diff
@@ -44,3 +44,9 @@
   When this is merged: https://github.com/facebook/react-native/pull/30345.
 
   For some reason CircleCI kept giving an error when running tests `TypeError: stacktraceParser.parse is not a function`. Once I moved the require higher up, things started working again.
+
+- hardcode mapbox version to at least 6.3.0 using $ReactNativeMapboxGLIOSVersion
+
+  When @react-native-mapbox-gl/maps uses mapbox-ios at least 6.3.0
+
+  We had issues with CLLocationManagerDelegate and iOS 14. Methods got deprecated and mapbox-ios supports iOS 14, so we had to manually use that version. Normally the npm package came with an older version.
```
brainbicycle, line 52: Do we still need this hard-coding since the deprecations ended up not being the cause?

brainbicycle (approved):
Looks great thanks for doing this and cleaning up the types!

HACKS.md
pvinis: 6.3 added support for xcode 12 and ios 14. Without this hardcoding we get version 5.7, so I think it's better if we keep the hardcode, at least until they give us at least that version.

brainbicycle: Got it thanks, maybe we can change explanation to something like your comment here?

emission/Pod/Classes/ViewControllers/ARMapContainerViewController.m
```diff
@@ -368,7 +368,7 @@ - (void)drawerChangedDistanceFromBottomWithDrawer:(PulleyViewController *)drawer
 
 - (void)locationManager:(CLLocationManager *)manager didChangeAuthorizationStatus:(CLAuthorizationStatus)status
 {
-    if (status == kCLAuthorizationStatusAuthorizedWhenInUse) {
+    if (status == kCLAuthorizationStatusAuthorizedWhenInUse || status == kCLAuthorizationStatusAuthorizedAlways) {
```
pvinis, line 371: No idea why this was not there already 🤔

src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -139,20 +133,13 @@ enum DrawerPosition {
   }
 })
 export class GlobalMap extends React.Component<Props, State> {
-  /** Makes sure we're consistently using { lat, lng } internally */
-  static lngLatArrayToLocation(arr: [number, number] | undefined) {
```
pvinis, line 143: unused

src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -162,36 +149,6 @@ export class GlobalMap extends React.Component<Props, State> {
   shows: { [id: string]: Show } = {}
   fairs: { [id: string]: Fair } = {}
 
-  stylesheet = Mapbox.StyleSheet.create({
```
pvinis, line 165: unused

src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -222,18 +179,6 @@ export class GlobalMap extends React.Component<Props, State> {
     this.setState({ activeIndex, activePin: null, activeShows: [] })
   }
 
-  resetZoomAndCamera = () => {
```
pvinis, line 225: absorbed by MapboxGL.Camera

src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -591,16 +537,11 @@ export class GlobalMap extends React.Component<Props, State> {
   onPressUserPositionButton = () => {
     // @ts-expect-error STRICTNESS_MIGRATION --- 🚨 Unsafe legacy code 🚨 Please delete this and fix any type errors if you have time 🙏
     const { lat, lng } = this.state.userLocation
-    this.map.moveTo([lng, lat], 500)
-  }
-
-  // @ts-expect-error STRICTNESS_MIGRATION --- 🚨 Unsafe legacy code 🚨 Please delete this and fix any type errors if you have time 🙏
-  onPressPinShapeLayer = (e) => this.handleFeaturePress(e.nativeEvent)
```
pvinis, line 598: unused

src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -591,16 +537,11 @@ export class GlobalMap extends React.Component<Props, State> {
...
-
-  storeMapRef = (c: any) => {
-    if (c) {
-      this.map = c
-    }
+    this.camera.setCamera({
```
pvinis, line 540: move and zoom at the same time

src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -591,16 +537,11 @@ export class GlobalMap extends React.Component<Props, State> {
   onPressUserPositionButton = () => {
     // @ts-expect-error STRICTNESS_MIGRATION --- 🚨 Unsafe legacy code 🚨 Please delete this and fix any type errors if you have time 🙏
     const { lat, lng } = this.state.userLocation
-    this.map.moveTo([lng, lat], 500)
-  }
-
-  // @ts-expect-error STRICTNESS_MIGRATION --- 🚨 Unsafe legacy code 🚨 Please delete this and fix any type errors if you have time 🙏
-  onPressPinShapeLayer = (e) => this.handleFeaturePress(e.nativeEvent)
-
-  storeMapRef = (c: any) => {
```
pvinis, line 600: reworked the refs

src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -677,16 +618,24 @@ export class GlobalMap extends React.Component<Props, State> {
...
-          {({ opacity }: any /* STRICTNESS_MIGRATION */) => (
+          {({ opacity }: { opacity: number }) => (
             <AnimatedView style={{ flex: 1, opacity }}>
-              <Map
+              <MapboxGL.MapView
+                ref={(r) => (this.map = r)}
+                style={{ width: "100%", height: Dimensions.get("window").height }}
                 {...mapProps}
                 onRegionIsChanging={this.onRegionIsChanging}
-                onUserLocationUpdate={this.onUserLocationUpdate}
```
pvinis, line 685: absorbed by MapboxGL.UserLocation

src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -721,16 +670,14 @@ export class GlobalMap extends React.Component<Props, State> {
    * What's happening is that we have to replicate a subset of the map's clustering algorithm to get
    * access to the shows that the user has tapped on.
    */
-  async handleFeaturePress(nativeEvent: any) {
+  async handleFeaturePress(event: OnPressEvent) {
```
pvinis, line 673: api changes

pvinis:
This is ready finally. 😰 🙌 

- [ ] I would like a quick review by @ds300 because I added some ts stuff because I didn't want to change logic or existing bugs, other than making the new version work.
- [ ] I would also like any of you to test this PR locally to make sure it works. I guess we could test on beta in QA, but I feel this change might be a bit tricky so I would like one extra vote of confidence before merging, if possible.

ds300 (approved):
amazing work figuring this all out!

src/lib/Scenes/City/CityPicker.tsx
```diff
@@ -14,76 +13,46 @@ interface Props {
...
-        Presented in partnership with BMW
-      </Sans>
-    )
-  }
+  useEffect(() => {
+    if (selectedCity === null) {
+      return
+    }
+    setSelectedCity(null)
+  }, [selectedCity])
```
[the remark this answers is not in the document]
pvinis: Apparently.. I don't know why..
https://github.com/artsy/eigen/blob/4a90fdca0bc090e2d93a535aad2f8f1823ce2869/src/lib/Scenes/City/CityPicker.tsx#L41
I tried without it and remember getting some funky data like London showed NY museums. I didn't look too much into it, but I think it might be firing events to the native part? I could investigate more, but I feel at some point we might just rework this whole thing and simplify.

Podfile
```diff
@@ -28,6 +33,40 @@ require 'json'
...
+def add_mapbox_creds
+  File.open($netrc_path, 'a+') { |f|
+    f.write("""machine api.mapbox.com
+login mapbox
+password #{ENV['MAPBOX_DOWNLOAD_TOKEN']}
+""")
+  }
+end
+def remove_mapbox_creds
+  if $netrc_exists
```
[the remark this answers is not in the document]
pvinis: renamed it to `user_already_had_netrc_file`. good idea.

ds300:
I'm only seeing a blank screen when I try to run City Guide from this branch 
image

ds300:
I just realised that the mapbox download token will be embedded in the app binary if folks put it in their `.env.shared` files. That's not ideal since it's a secret, although it is not too risky to leak. What do you think? Should we just ask people to create the `~/.netrc` file manually?

pvinis:
is it because react-native-config just grabs everything from .env.shared? I think it's ok, since they are all obfuscated, right?
The only way to access it would be by having the source code read it.

ashfurrow:
That's right. I think it's okay if we include it in the app binary, because it is obfuscated. It would be possible to attach a debugger to the running app and extract it, but that's true of all the keys. I'm okay with merging but would like to hear from @brainbicycle.

pvinis:
Trying to make the white screen go away, first. Not sure why this happens yet.

pvinis:
I figured out the white screen. @ds300, you need to give the simulator a location. In Xcode just give it some location out of the list, and the map will appear.

`MapContainer` just shows an empty view if there is no previous city slug set on the device, and it waits for the location callback, which happens automatically on a device but not on the simulator.

Let me know if it doesn't fix it for you. Also let me know if it does, and we can merge 🙇.

pvinis:
To clarify, this is not a new behaviour. I didn't change that part. I guess I was always just testing with a location set on my sim.

ashfurrow:
Oh yeah, this is a known thing. It's a quirk that only developers see, where the location callback for "approved" is called but not the one for "here's the location", if that makes sense.

ds300:
> To clarify, this is not a new behaviour. I didn't change that part. I guess I was always just testing with a location set on my sim.

I don't recall ever having to set a location before. Is there something we can do when `__DEV__` is `true` as a fallback so that other people don't run into this?

ashfurrow:
A fallback might be nice, though I would suggest we draw attention to how the simulator needs a location (rather than providing a stubbed one ourselves). That way we keep the codepaths for developers and users as close as possible.

You may have not ran into it because, after having a location set once, the simulator will remember this the next time it gets booted up. So it's the kind of thing that tends to happen only when we use new simulators for the first time (iOS/Xcode upgrades, etc).

pvinis:
I could add a quick fallback.

pvinis:
I added a fallback, I thought it's kind of a nice way, to force a default when the simulator fails if it has no location. And if it does have one, it will return that one as normal.

emission/Pod/Classes/ViewControllers/ARMapContainerViewController.m
```diff
@@ -386,12 +386,20 @@ - (void)locationManager:(CLLocationManager *)manager didUpdateLocations:(NSArray
...
 }
 
-- (BOOL)shouldAutorotate;
+- (void)locationManager:(CLLocationManager *)manager didFailWithError:(NSError *)error
+{
+#if TARGET_IPHONE_SIMULATOR
+    // if there is a failure, just pick a city on the simulator and pretend that's the city that is returned
+    [self locationManager:manager didUpdateLocations:@[[[ARCity cities] objectAtIndex:2].epicenter]];
+#endif
+}
```
ashfurrow, line 395: This looks good, but we should avoid adding this delegate method on non-simulator builds altogether. Sometimes Apple frameworks behave differently depending on if a method exists or not.

```suggestion
#if TARGET_IPHONE_SIMULATOR
- (void)locationManager:(CLLocationManager *)manager didFailWithError:(NSError *)error
{
    // if there is a failure, just pick a city on the simulator and pretend that's the city that is returned
    [self locationManager:manager didUpdateLocations:@[[[ARCity cities] objectAtIndex:2].epicenter]];
}
#endif
```

ashfurrow (approved):

emission/Pod/Classes/ViewControllers/ARMapContainerViewController.m
pvinis: Hm good catch!

PR #4231, opened 2020-12-08
feat(ci): add slack notification when betas fail

brainbicycle:

### Description

Adds a slack notification when beta lane fails for any reason based on discussion in retro and feedback at front-end-ios (now mobile) practice.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -211,6 +211,22 @@ lane :notify_if_new_license_agreement do
   end
 end
 
+lane :notify_beta_failed do |options|
+  exception = options[:exception]
+  message = ':x: :iphone:\n'\
+            'Looks like the latest beta failed to deploy!\n'\
+            'See circle job for more details.'
+  slack(
+    message: message,
+    success: false,
+    payload: {
+      'Circle Build' => ENV['CIRCLE_BUILD_URL'],
+      'Exception' => exception.message
+    },
+    default_payloads: []
+  )
+end
+
 lane :tag_and_push do |options|
   # Do a tag, we use a http git remote so we can have push access
   # as the default remote for circle is read-only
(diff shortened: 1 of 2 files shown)


fastlane/Fastfile
```diff
@@ -211,6 +211,22 @@ lane :notify_if_new_license_agreement do
   end
 end
 
+lane :notify_beta_failed do |options|
+  exception = options[:exception]
+  message = ':x: :iphone:\n'\
+            'Looks like the latest beta failed to deploy!\n'\
+            'See circle job for more details.'
```
ashfurrow, line 218: Ruby 2.3 added support for HEREDOC-style multiline strings that unindent for you. The equivalent code using them would be:

```suggestion
  message = <<~MSG
    :x: :iphone:
    Looks like the latest beta failed to deploy!
    See circle job for more details.
  MSG
```

Older Ruby versions didn't have `<<~`, only `<<-`, which would have added extra spaces at the beginning of each line.

fastlane/Fastfile
```diff
@@ -211,6 +211,22 @@ lane :notify_if_new_license_agreement do
...
+  slack(
+    message: message,
+    success: false,
+    payload: {
+      'Circle Build' => ENV['CIRCLE_BUILD_URL'],
```
ashfurrow, line 223: Yes! Good idea.

ashfurrow (approved):
Great! Just one suggestions about Ruby syntax 👍

sepans (approved):
Thanks for adding this so quickly 🙏 😍

fastlane/Fastfile
brainbicycle: ✨ Nice! I'll add that to the other slack message as well

PR #4234, opened 2020-12-09
fix: Fix check for checked property on Checkbox

jo-rs:

This PR resolves [PURCHASE-2297]

### Description

This is a small fix for the `Checkbox` component. To change the checkbox programatically, it was relying on a check of the `props.checked` value, which would return `false` if it was indeed not checked. This fix checks if it's present rather than its value.

[PURCHASE-2297]: https://artsyproduct.atlassian.net/browse/PURCHASE-2297

### DIFF
--- a/src/lib/Components/Bidding/Components/Checkbox.tsx
+++ b/src/lib/Components/Bidding/Components/Checkbox.tsx
@@ -18,7 +18,7 @@ interface CheckboxProps extends TouchableWithoutFeedbackProps, FlexProps {
 
 export class Checkbox extends Component<CheckboxProps, CheckboxState> {
   static getDerivedStateFromProps: GetDerivedStateFromProps<CheckboxProps, CheckboxState> = (props, state) => {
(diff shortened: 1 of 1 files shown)


src/lib/Components/Bidding/Components/Checkbox.tsx
```diff
@@ -18,7 +18,7 @@ interface CheckboxProps extends TouchableWithoutFeedbackProps, FlexProps {
 
 export class Checkbox extends Component<CheckboxProps, CheckboxState> {
   static getDerivedStateFromProps: GetDerivedStateFromProps<CheckboxProps, CheckboxState> = (props, state) => {
-    if (props.checked && props.checked !== state.checked) {
+    if ("checked" in props && props.checked !== state.checked) {
       return {
```
xtina-starr, line 22: Good catch. I didn't even know you could do: `"checked" in props`

#TIL

PR #4237, opened 2020-12-09
feat: Display interim status auctions in My Bids, and small UI fixes

jo-rs:

This PR resolves remainders from [PURCHASE-2167] [PURCHASE-2286]

### Description

This PR includes the ability to see auctions that have a pending registration and don't have bids on `MyBids`, a redirect to identity verification when needed (instead of faq), and small UI fixes to `MyBids`.
my_bids

[PURCHASE-2167]: https://artsyproduct.atlassian.net/browse/PURCHASE-2167
[PURCHASE-2286]: https://artsyproduct.atlassian.net/browse/PURCHASE-2286


src/lib/Scenes/MyBids/helpers/timely.ts
```diff
@@ -12,7 +12,7 @@ export const lotInActiveSale: (lotStanding: {
 export const isLotStandingComplete: (lotStanding: { lotState?: { soldStatus?: string } }) => boolean = (lotStanding) =>
   !!(lotStanding.lotState?.soldStatus && ["sold", "passed"].includes(lotStanding.lotState.soldStatus?.toLowerCase()))
 
-interface SaleWithTimes {
+export interface SaleWithTimes {
```
erikdstock, line 15: i don't think this needs to be exported any more

src/lib/Scenes/MyBids/MyBids.tsx
```diff
@@ -95,16 +101,27 @@ class MyBids extends React.Component<MyBidsProps> {
...
-                  <SaleCardFragmentContainer key={saleId} sale={sale} me={me} smallScreen={isSmallScreen}>
+                  <SaleCardFragmentContainer
+                    key={sale.internalID}
+                    sale={sale}
+                    me={me}
+                    smallScreen={isSmallScreen}
+                    hideChildren={!showNoBids && !activeLotStandings.length}
+                  >
                     <Join separator={<Separator my={1} />}>
+                      {!!showNoBids && (
```
erikdstock, line 116: I feel like this logic and its interaction with the SaleCard `hideChildren` prop is a bit difficult to understand. I'm going to take a look at it while doing some simulator QA.

src/lib/Scenes/MyBids/MyBids.tsx
```diff
@@ -67,9 +69,13 @@ class MyBids extends React.Component<MyBidsProps> {
     // sort each group of lot standings by position (lot number)
     const sortedActiveLots = mapValues(activeBySaleId, (lss) => sortBy(lss, (ls) => ls?.saleArtwork?.position!))
 
-    // sort an ordered list of sale ids by their relevant end time
-    const sortedSaleIds: string[] = sortBy(Object.keys(sortedActiveLots), (saleId) => {
-      const timelySale = TimelySale.create(sortedActiveLots[saleId][0]?.saleArtwork?.sale!)
+    // TODO: create list of all unique sales from lotStandings, bidders, and watchedLots (in the future).
+    // NOTE: bidders should include lotStandings, do we need both?
```
erikdstock, line 73: bidders should include all sales in lot standings* - our assumption is that if you have a lot standing from causality, then you have a matching sale registration in `me.bidders`, which is an assumption I suppose. As far as whether we _need_ the sale under the `lotStanding` but i don't think there is a serious performance hit in including them. We additionally might _need_ the sale, at least to check whether it is ended, but i'll take a look in the code.

The other TODO above may not be necessary for now (with the exception of watched lots which will add new sales not included in bidders).

erikdstock (commented):
Some initial comments.

src/lib/Scenes/MyBids/helpers/timely.ts
jo-rs: Removed

src/lib/Scenes/MyBids/MyBids.tsx
erikdstock: Understand better why we are doing this. @jo-rs briefly tried a refactor to move the children-rendering logic inside the SaleCard (https://github.com/artsy/eigen/pull/4237/commits/7a293decafffc99b6142188aea5ba6d0bb9500fe) but it was harder than expected. Something we could revisit but we didn't get to it today.

erikdstock:
Added test and did some minor cleanup, think i found a minor bug in the process. ✅ from me.

src/lib/Scenes/MyBids/MyBids.tsx
```diff
@@ -114,26 +134,28 @@ class MyBids extends React.Component<MyBidsProps> {
...
-                    !!ls && (
-                      <ClosedLot
-                        withTimelyInfo
-                        data-test-id="closed-sale-lot"
-                        lotStanding={ls}
-                        key={ls?.lotState?.internalID}
-                      />
+        {hasClosedBids && (
+          <>
+            <BidTitle>Closed Bids</BidTitle>
```
jo-rs, line 139: @erikdstock I'm moving this back out of the fragment. I need it to be a direct child of the `ScrollView` for the stickiness to work.

erikdstock (approved):

PR #4239, opened 2020-12-10
Fix: live lots a user is winning always appear as 'reserve not met'

erikdstock:
[PURCHASE-2293]
This pr fixes an issue where live auctions would show all winning lots as 'reserve not met.' in fact, reserve is never met in a live auction. This PR fixes the property check. The test was written in a way that it passed wrongly.

#squashongreen

[PURCHASE-2293]: https://artsyproduct.atlassian.net/browse/PURCHASE-2293

### DIFF
--- a/src/lib/Scenes/MyBids/Components/ActiveLot.tsx
+++ b/src/lib/Scenes/MyBids/Components/ActiveLot.tsx
@@ -9,7 +9,6 @@ import { LotFragmentContainer as Lot } from "./Lot"
 
 export const ActiveLot = ({ lotStanding }: { lotStanding: ActiveLot_lotStanding }, smallScreen: boolean) => {
   const timelySale = TimelySale.create(lotStanding?.saleArtwork?.sale!)
-  const isLAI = timelySale.isLiveBiddingNow()
 
   const sellingPrice = lotStanding?.lotState?.sellingPrice?.display
   const bidCount = lotStanding?.lotState?.bidCount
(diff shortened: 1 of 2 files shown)


src/lib/Scenes/MyBids/Components/ActiveLot.tsx
```diff
@@ -35,7 +34,9 @@ export const ActiveLot = ({ lotStanding }: { lotStanding: ActiveLot_lotStanding
           </Text>
         </Flex>
         <Flex flexDirection="row" alignItems="center" justifyContent="flex-end">
-          {!isLAI && lotStanding?.isHighestBidder && lotStanding.lotState.reserveStatus === "ReserveNotMet" ? (
```
erikdstock, line 38: this `isLAI` is actually wrong- the name suggests it is a live auction integration sale but it is actually testing whether the sale has entered live bidding right now. Due to the way we tested this it was incorrectly passing.

src/lib/Scenes/MyBids/__tests__/ActiveLot-tests.tsx
```diff
@@ -58,12 +58,14 @@ describe(ActiveLot, () => {
...
     it("says 'Highest bid' if the user is winning the lot but the reserveStatus is ReserveNotMet in a Live Auction", () => {
+      const date = new Date()
+      date.setDate(date.getDate() + 1)
       const tree = renderWithWrappers(
         <ActiveLot
           lotStanding={lotStandingFixture({
             isHighestBidder: true,
             lotState: { reserveStatus: "ReserveNotMet" },
-            saleArtwork: { sale: { liveStartAt: new Date().toJSON() } },
+            saleArtwork: { sale: { liveStartAt: date } },
```
erikdstock, line 68: because the date we created here was technically in the past, `isLiveBiddingNow()` was returning true when technically it would not in real life, causing the test to pass when it should have failed. now we set the date explicitly in the future to avoid this case.

erikdstock (commented):
#squashongreen

PR #4240, opened 2020-12-10
fix: multi image upload in my collection and fallback for processing photos

brainbicycle:

This PR resolves [CX-850]

### Description

This key/unique identifier we provide while uploading images doubles as a location for the uploaded image, when this was converted to a synchronous call we were only generating one key for multiple images which was causing uploading multiple images to fail.  There was another bug around uploading images that while images are processing we are getting back incomplete data from metaphysics causing artwork view to throw errors.

This PR:
- Moves key generation into for loop so each image gets a unique key
- Adds a fallback "Photo processing" state to artwork view and adds checks on image data

Notes:
There also seems to be a bug with the first photo in the iOS 14 photo picker in simulator failing to load so if that photo is selected you will get one less than what you selected. See thread here: 
https://developer.apple.com/forums/thread/658135

### Screenshots

kaws-processing

### Follow-ups

- [x] The incomplete image data seems like a bug on backend and should probably be fixed there if possible
- [ ] The image height in the artwork view is too large for some images
- [ ] There seem to be some caching issues on artwork page, ideally page would refresh when photos finish processing

[CX-850]: https://artsyproduct.atlassian.net/browse/CX-850


ds300:
> The incomplete image data seems like a bug on backend and should probably be fixed there if possible

I don't think this is a bug, but it's for sure surprising behaviour. This is just how gravity returns images that haven't been processed by gemini yet.

> The image height in the artwork view is too large for some images

@samchieng should the image section at the top of the my collection artwork page have a max height like the one on the regular artwork page? We could shrink it down to fit within the frame and add a `black5` background? something like this: before | after

image

>  There seem to be some caching issues on artwork page, ideally page would refresh when photos finish processing

We could make it a refetch container and add polling when we detect that artworks are still processing.

src/lib/Scenes/MyCollection/Screens/Artwork/Components/MyCollectionArtworkHeader.tsx
```diff
@@ -24,6 +24,55 @@ const MyCollectionArtworkHeader: React.FC<MyCollectionArtworkHeaderProps> = (pro
...
+    if (!isImage(mainImage) || imageIsProcessing(mainImage)) {
+      return true
+    }
+
+    if (!imagesToCheck) {
+      return false
+    }
+
+    const concreteImages = imagesToCheck as Image[]
+    const stillProcessing = concreteImages.filter((image) => imageIsProcessing(image)).length > 0
```
ds300, line 46: nonblocking: js has the `some(fn)` method which is the same thing as `.filter(fn).length > 1` but with better perf

```suggestion
    const stillProcessing = concreteImages.some((image) => imageIsProcessing(image))
```

src/lib/Scenes/MyCollection/Screens/Artwork/Components/MyCollectionArtworkHeader.tsx
```diff
@@ -24,6 +24,55 @@ const MyCollectionArtworkHeader: React.FC<MyCollectionArtworkHeaderProps> = (pro
...
+    return stillProcessing
+  }
+
+  const processingText = () => {
+    return images && images?.length > 0 ? "Processing photos" : "Processing photo"
+  }
+
+  const MainImageView = () => {
```
ds300, line 54: We shouldn't be creating new react component types during render, since that makes react confused about what has and hasn't changed
```suggestion
  const renderMainImageView = () => {
```

and then below

```diff
- <MainImageView />
+ {renderMainImageView()}
```

src/lib/Scenes/MyCollection/Screens/Artwork/Components/MyCollectionArtworkHeader.tsx
```diff
@@ -24,6 +24,55 @@ const MyCollectionArtworkHeader: React.FC<MyCollectionArtworkHeaderProps> = (pro
...
+    if (!isImage(defaultImage) || imageIsProcessing(defaultImage)) {
+      return (
+        <Flex
+          style={{ height: 300, alignItems: "center", justifyContent: "center", backgroundColor: color("black10") }}
+        >
+          <ArtworkIcon style={{ opacity: 0.6 }} height={100} width={100} />
+          <Text style={{ opacity: 0.6 }}>{processingText()}</Text>
```
ds300, line 61: nitpick nonblocking: I'd prefer to see this inlined since it's only used in one place and it's straightforward to read.

ds300 (approved):
Looks great! I like the processing state! just one blocking issue but otherwise 💯

bhoggard (approved):
I think it makes sense to do all of @ds300's suggestions, but otherwise LGTM.

brainbicycle:
> > The incomplete image data seems like a bug on backend and should probably be fixed there if possible
> 
> I don't think this is a bug, but it's for sure surprising behaviour. This is just how gravity returns images that haven't been processed by gemini yet.
> 
> > The image height in the artwork view is too large for some images
> 
> @samchieng should the image section at the top of the my collection artwork page have a max height like the one on the regular artwork page? We could shrink it down to fit within the frame and add a `black5` background? something like this: before | after
> 
> image
> 
> > There seem to be some caching issues on artwork page, ideally page would refresh when photos finish processing
> 
> We could make it a refetch container and add polling when we detect that artworks are still processing.

Thanks! I wonder if it makes sense to add some explicit indicator in metaphysics that photos are still processing? Maybe that already exists, not critical but I can see other people running into this in the future.

Thanks for the suggestions on the images and refetch, that all makes sense to me!

src/lib/Scenes/MyCollection/Screens/Artwork/Components/MyCollectionArtworkHeader.tsx
brainbicycle: Nice

ds300:
>  I wonder if it makes sense to add some explicit indicator in metaphysics that photos are still processing? Maybe that already exists, not critical but I can see other people running into this in the future.

As far as I know that doesn't exist. Would be a good feature though!! I'll add a ticket to our backlog.

PR #4241, opened 2020-12-10
feat: phone number input

ds300:
This PR resolves [CX-890]

### Description

In a future PR we'll be adding a step in the consignments flow where a user must confirm/provide their phone number. The consignments team wanted us to make sure there's an explicit area code on users' phone numbers, since up until now the `:phone` field has been more-or-less free text on both web and mobile. 

This PR preempts that change to the consignments flow by adding a new phone number input with an explicit area code dropdown, and using it on the `MyAccountEditPhone` screen. 

This new input, `PhoneInput`, also tries to normalize existing values from gravity to the international standard format (E.164). It uses the user's locale and whatever they've written previously as hints. I decided to do this because only 24623 out of 130702 users with a meaningful phone number use the international format. I can't easily measure the precision of my normalization function, since we don't store the user's location in gravity (maybe last IP address would work?), but it's bound to be a lot higher than 0% :p

There are some decent phone number inputs on npm, but I couldn't find a really solid one with great UX, customisable UI, and reasonable bundle footprint, so I made one using our existing `Select` and `Input` components.

new-phone-input

[CX-890]: https://artsyproduct.atlassian.net/browse/CX-890


CHANGELOG.yml
```diff
@@ -7,6 +7,19 @@ upcoming:
...
+    - Avoid shrinking the background on non-fullscreen fancy modals - david
+    - Support iOS 14 photo picker in My Collection - brian
+    - Scroll to add message input on Inquire screen - ashley, mounir
+    - Make add message input height relative to its size - ashley, mounir
+    - Avoid shrinking the background on non-fullscreen fancy modals - david
+    - removed x from inquiry sent notification - lily
+    - Minor copy changes for My Collection - pepopowitz
+    - Fix city guide hang on iOS 14 - brian
+    - Fix apns token registration issue - brian
+    - Fix regression that hids textarea in 1st inquiry modal - christina
```
starsirius, line 21: Looks like the above got carried over from 6.7.3 accidentally.

ds300: oof good catch, will fix 👍

src/lib/Components/Select.tsx
```diff
@@ -178,23 +178,32 @@ const SelectModal: React.FC<{
...
-      // try to center the option on screen
-      const initialScrollOffset = initialScrollIndex * ROW_HEIGHT - flatListHeight.current / 2 + ROW_HEIGHT
-      requestAnimationFrame(() => {
-        flatListRef.current?.scrollToOffset({ offset: initialScrollOffset, animated: false })
-      })
+      const scrollToSelectedItem = async () => {
+        let safety = 0
+        // wait for flat list to lay out
+        while (flatListHeight.current == null && safety++ < 100) {
+          await new Promise((r) => requestAnimationFrame(r))
```
MounirDhahri, line 198: Nice, I never thought about this 👏

src/lib/Components/PhoneInput/cleanUserPhoneNumber.ts
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * In gravity we have a lot of dirty data since ':phone' has been more-or-less a free text field
+ *
+ * There's huge variability in how people have entered phone numbers, so here we do our best
+ * to normalize phone numbers to a country code (iso2, e.g. 'us') and a phone number within that country code
+ * so that we can replace the country code with its international dialling prefix.
+ *
+ * e.g. if this function returns {countryCode: 'gb', phoneNumber: '7825577664'}
+ * then to call this number you'd type in '+447825577664'
+ */
```
MounirDhahri, line 14: 🙌

MounirDhahri (approved):

MounirDhahri:
Thanks for the clean git history. it made reviewing this PR really satisfying

MounirDhahri:
#squashongreen

PR #4247, opened 2020-12-10
fix: remove unused auctionsLotStandingConnection from profile

erikdstock:

This PR fixes a bug where an unused query may cause a user's profile to not load.


zephraph:
So if I'm understanding this correctly, these queries aren't used in the app but were when MyBids was in the profile tab? If so 👍

zephraph (approved):

PR #4250, opened 2020-12-11
fix: switch to preferred lot and sellingPrice fields

erikdstock:
This PR completes [PURCHASE-2294]
#minor
In short this pr switches to use `sellingPrice` rather than `floorSellingPrice` (which only respects bids recognized by the live floor auctioneer) on the my bids screen. It also switches to use `LotStanding.lot` over the deprecated `lotState` field. It's mostly a find/replace and removing some aliased field names.

To QA: Verify that artworks display with identical current bid amount on both my bids and the artwork's screen. Bid count, also included in the jira ticket, should also match.

This PR resolves [CX-]

[PURCHASE-2294]: https://artsyproduct.atlassian.net/browse/PURCHASE-2294


jo-rs (approved):
This seems fairly straight forward. Looks good! 
Merging 🚀

PR #4251, opened 2020-12-11
fix: low quality my collection images

brainbicycle:

This PR resolves [CX-889]

### Description

Fixes an issue that was causing artwork images to appear grainy in My Collection. The url we were using from metaphysics was not the gemini url with ":version" component so we were getting the wrong processed  image to display.

- Uses `imageURL` with version component rather than `url` from metaphysics
- Passes in explicit compression options to avoid unnecessary compression in native module 

#### Before

towel_before

#### After

towel_after

[CX-889]: https://artsyproduct.atlassian.net/browse/CX-889


ds300 (approved):
🙏 🙌

PR #4253, opened 2020-12-11
WIP: MyBids analytics

xtina-starr:

This PR resolves [CX-]

### DIFF
--- a/src/lib/Containers/Inbox/Inbox.tsx
+++ b/src/lib/Containers/Inbox/Inbox.tsx
@@ -120,14 +120,16 @@ export class Inbox extends React.Component<Props, State> {
   }
 
   handleNavigationTabTap = (index: number) => {
+    // we are determining which tab is clicked based on the index.
+    // The index maps to hard-coded values so if that changes the map should be updated.
+
+    // follow up on why inboxBids isn't available on ContextModule
+    const tabs = [OwnerType.inboxBids, ContextModule.inboxInquiries]
+
     track({
       action: ActionType.tappedNavigationTab,
-      // ContextModule: index === 0 ? myBids : inboxInquiries
-      context_module: ContextModule,
-      context_screen_owner_type: OwnerType.inbox,
-      // What does the owner id represent here?
-      context_screen_owner_id: index,
-      context_screen_owner_slug: "string",
+      context_module: tabs[index],
+      context_screen_owner_type: tabs[index],
     })
   }
 
--- a/src/lib/Containers/Inbox/Inbox.tsx
+++ b/src/lib/Containers/Inbox/Inbox.tsx
@@ -1,3 +1,4 @@
+import { ActionType, ContextModule, OwnerType } from "@artsy/cohesion"
 import { Inbox_me } from "__generated__/Inbox_me.graphql"
 import { InboxQuery } from "__generated__/InboxQuery.graphql"
 import { defaultEnvironment } from "lib/relay/createEnvironment"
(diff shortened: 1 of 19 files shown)


src/lib/Containers/Inbox/Inbox.tsx
```diff
@@ -128,6 +144,7 @@ export class Inbox extends React.Component<Props, State> {
           contentInset: { bottom: bottomInset },
           onLayout: this.onScrollableTabViewLayout,
         }}
+        onChangeTab={(i) => this.handleNavigationTabTap(i)}
```
xtina-starr, line 147: I think this (using the `@track` decorator) is kinda how we will want to approach tracking here seeing this is a class component. I was getting a type error that I hadn't yet addressed and I wasn't quite sure how to approach the screen owner id and slug since it doesn't really  map to a model.

I was thinking to use [`onChangeTab`](https://github.com/ptomasroos/react-native-scrollable-tab-view#props) here to be able to determine when the respective tabs were clicked.

src/lib/Scenes/Inbox/Components/Conversations/Conversations.tsx
```diff
@@ -72,6 +75,17 @@ export class Conversations extends Component<Props, State> {
     }
   }
 
+  @track((props) => {
+    action: ActionType.tappedInboxConversation,
+    context_module: ContextModule.inboxInquiries,
+    context_screen_owner_type: OwnerType.inboxInquiries,
+    destination_screen_owner_type: OwnerType.inboxConversation,
+    destination_screen_owner_id: props.id,
```
xtina-starr, line 83: I didn't correct it yet but this should actually `internalID`.

src/lib/Scenes/MyBids/Components/SaleCard.tsx
```diff
@@ -65,7 +68,25 @@ export class SaleCard extends React.Component<{ sale: SaleCard_sale; me: SaleCar
...
+              context_screen_owner_type: "home",
+              destination_screen_owner_type: "sale",
+              destination_screen_owner_id: "5e95b37a2fdcb20012a0e082",
+              destination_screen_owner_slug: "forum-auctions-colour-theory-4",
+              horizontal_slide_position: 3,
+              module_height: "double",
+              type: "thumbnail",
+            })
+            navigate(sale?.href as string)
+          }}
```
xtina-starr, line 88: These property values need to be replaced with the actual values.

xtina-starr:
[This](https://github.com/artsy/cohesion/pull/141#issuecomment-740294392) outlines the respective analytics events we are going to want to track.

eessex:
@xtina-starr made a pr in cohesion to help with these events, we should use this helper to format the arguments to `track`: https://github.com/artsy/cohesion/pull/147

src/lib/Containers/Inbox/Inbox.tsx
```diff
@@ -117,6 +119,18 @@ export class Inbox extends React.Component<Props, State> {
     this.scrollViewVerticalStart = layout.nativeEvent.layout.y
   }
 
+  handleNavigationTabTap = (index: number) => {
+    track({
```
eessex, line 123: Can use `tappedNavigationTab`: https://cohesion.artsy.net/modules/_helpers_tap_tappednavigationtab_.html

```
track(
  tappedNavigationTab({
    context_module: ContextModule.foo,
    context_screen_owner_type: OwnerType.inbox,
    context_screen_owner_id: <ID of current page if inbox if it has one>,
    context_screen_owner_slug: <slug of current page if it has one>,
    subject: <Tab text>
  })
)
```

xtina-starr: Ah okay. How do we get this helper imported into eigen in order to use it?

`import { tappedNavigationTab } from "@artsy/cohesion/src/Helpers/Tap/TappedNavigationTab"`  ??

eessex: You'll need to update the the most recent cohesion (1.67.4) and then you'll be able to access from index:
```
import { tappedNavigationTab, tappedInboxConversation } from "@artsy/cohesion"
```

PR #4254, opened 2020-12-11
chore: Remove my bids flag

lilyfromseattle:
This PR removes the feature flag for the my bids tab, so that we can use an echo flag instead (added here)[https://github.com/artsy/echo/pull/63]

May need to hold off on merging until the echo PR is merged


zephraph:
I think this is likely because echo hasn't been deployed yet.  I just shipped that out, I'll follow up once it's done.

PR #4256, opened 2020-12-13
fix(my collection):  allow deleting photos in my collection

brainbicycle:

This PR resolves [CX-851]

### Description

Calls a new deletion mutation on saving edits to an artwork with any removed image ids. 

- Adds imageURL and id fields to shared artwork image props
- Adds a function for gets removed artwork ids on saving an edited artworks
- Adds an artwork image deletion mutation 
- Fixes a bug I introduced where images for uploaded images weren't showing up in edit modal
- ~~Changes gemini upload to use a unique filename rather than a different key everytime~~

### Screenshots
deletePhoto2

### TODO

- [x] Images seem to be deleted when calling mutation but an error is being thrown by Metaphysics, believe the resolver type is not correct in metaphysics - **blocks PR merging** PR here: https://github.com/artsy/metaphysics/pull/2895
- [x] If deletion should happen on edit saving should we add this to edit mutation in Metaphysics and simplify client logic?
- [x] Found another bug while looking into this that uploaded images don't seem to be associated with artwork correctly filed here: https://artsyproduct.atlassian.net/browse/CX-904

[CX-851]: https://artsyproduct.atlassian.net/browse/CX-851


src/lib/Scenes/MyCollection/Screens/ArtworkFormModal/MyCollectionArtworkFormModal.tsx
```diff
@@ -91,6 +93,12 @@ export const MyCollectionArtworkFormModal: React.FC<MyCollectionArtworkFormModal
             ...cleanArtworkPayload(others),
             ...explicitlyClearedFields(others, dirtyFormCheckValues),
           })
+
+          // TODO: Can this logic live in metaphysics as part of edit artwork mutation?
+          const deletedIDs = deletedPhotoIDs(dirtyFormCheckValues.photos, photos)
+          for (const deletedID of deletedIDs) {
+            await deleteArtworkImage(props.artwork.internalID, deletedID)
+          }
```
ds300, line 101: i like having an explicit delete mutation, seems like a more obvious API to me.

ds300 (approved):
This is great work, thanks Brian! 🎉

PR #4260, opened 2020-12-15
POC/Spike: Convert Inbox to use a FC

erikdstock:
This PR was originally intended to fix a simple problem - refetching the active inbox view when the user exits and revisits so they always get the latest data. This was a result of testing and QA where people coming in fresh didn't intuitively pull to refresh which led to reports that new data wasn't loading. We discussed 2 possible approaches: either manually refetching data when the view reappears or doing something with relay to make it always refetch.

In the process we did a spike on converting the inbox to use a function component because the class component was very old and hard to understand. Tests are still passing and it all appears to work fine but this would be a big diff so we should be strategic if we decide to merge.

Some problems that remain with this approach:
- the `isVisible` prop passed in from Objective C code doesn't flip until you have exited the inbox by at least 2 screens, so it fundamentally doesn't work.
- Still relying on refs and class components for the underlying bids/conversations views. We don't know if this is an anti-pattern. In any case we are duck-typing our children views so we can call the same `refetchData` method on either active tab.
- even though the view refreshes immediately, in some cases we don't see the new data we expect right away. Maybe this is fine UX since the problem is coming from the backend.

I think our main question for now is whether we want to go with this big of a change or if there is a better way to make the inbox refresh automatically (maybe a relay container/renderer setting?).

Co-authored-by: Lily Pacegit <lilyfpace@gmail.com>

### DIFF
--- a/src/lib/Containers/Inbox/Inbox.tsx
+++ b/src/lib/Containers/Inbox/Inbox.tsx
@@ -1,8 +1,8 @@
 import { Inbox_me } from "__generated__/Inbox_me.graphql"
 import { InboxQuery } from "__generated__/InboxQuery.graphql"
 import { defaultEnvironment } from "lib/relay/createEnvironment"
-import { ConversationsContainer } from "lib/Scenes/Inbox/Components/Conversations/Conversations"
-import { MyBidsContainer } from "lib/Scenes/MyBids/MyBids"
+import { Conversations, ConversationsContainer } from "lib/Scenes/Inbox/Components/Conversations/Conversations"
+import { MyBids, MyBidsContainer } from "lib/Scenes/MyBids/MyBids"
 import { listenToNativeEvents } from "lib/store/NativeModel"
 import renderWithLoadProgress from "lib/utils/renderWithLoadProgress"
 import { Flex, Separator, Text } from "palette"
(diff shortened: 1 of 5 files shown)


src/lib/Containers/Inbox/Inbox.tsx
```diff
@@ -56,91 +56,81 @@ interface Props {
...
-      this.myBids.refreshMyBids(() => {
-        this.setState({ fetchingData: false })
-      })
-    } else {
-      this.props.relay.refetch({}, null, () => {
-        this.setState({ fetchingData: false })
-      })
+    } else if (!!activeTab?.current) {
+      setFetching(true)
+      activeTab?.current?.refetchData(() => setFetching(false))
```
erikdstock, line 87: I'm wondering if we could drop the refs completely and just pass `isVisible` directly to the tabs like `isVisible={props.isVisible && thisTabIsTehActiveOne}`.  Then the tabs themselves would watch for re-becoming visible and refresh when they do.

src/lib/Containers/Inbox/Inbox.tsx
```diff
@@ -56,91 +56,81 @@ interface Props {
...
+  const [isFetching, setFetching] = React.useState<boolean>(false)
+  const [notificationListener, setNotificationListener] = React.useState<EmitterSubscription | null>(null)
 
-  state = {
-    fetchingData: false,
-  }
-
-  scrollViewVerticalStart = 0
+  // TODO: Maybe this should actually be a useRef
+  const [scrollViewVerticalStart, setScrollViewVerticalStart] = React.useState(0)
```
erikdstock, line 66: i don't understand when to use `useRef()`.

src/lib/Containers/Inbox/Inbox.tsx
```diff
@@ -56,91 +56,81 @@ interface Props {
...
-  }
-
-  componentWillUnmount() {
-    this.listener?.remove()
-  }
-
-  UNSAFE_componentWillReceiveProps(newProps: Props) {
-    if (newProps.isVisible) {
-      this.fetchData()
+  const handleChangeTab = ({ i: tabIndex }: { i: number }) => {
```
erikdstock, line 73: used by our [react-native-scrollable-tab-view module](https://github.com/ptomasroos/react-native-scrollable-tab-view#props)

src/lib/Containers/Inbox/Inbox.tsx
```diff
@@ -56,91 +56,81 @@ interface Props {
...
-        this.setState({ fetchingData: false })
-      })
-    } else {
-      this.props.relay.refetch({}, null, () => {
-        this.setState({ fetchingData: false })
-      })
+    } else if (!!activeTab?.current) {
+      setFetching(true)
+      activeTab?.current?.refetchData(() => setFetching(false))
+      // also refetch from relay?
```
erikdstock, line 88: not clear what `Inbox_me` is even being used for at this point - unless refetching the entire inbox query at once in this view makes more sense (and will work for the sub-components)

erikdstock:
Closing because #4280 is a more complete draft of the same refactor.

PR #4267, opened 2020-12-16
fix(infra): make isVisible prop respect tab switches.

ds300:
### Description

This unblocks some purchase work on Inbox/MyBids. The `isVisible` prop set in objective-c is not totally reliable when switching tabs because, as an implementation detail, we sometimes have more than one tab stack VC mounted at once, even if only one is visible.

So we want to make sure that `isVisible` is true when

1. The view is the topmost view in a nav stack
2. The view is in the stack for the currently-visible tab

To do this a view needs to know three things

1. "Am I the topmost view on a nav stack?" — Already taken care of in `ARComponentViewController` by setting a prop on the root react component: `isVisible`.
2. "Which tab am I a view in?" — Not previously available, now taken care of in `ARScreenPresenterModule` by setting a new prop on the root react component: `navStackID`.
3. "Which tab is the user currently looking at?" — Already available through the `useSelectedTab()` react hook. 

We added another level of wrapping in `AppRegistry` to combine these three bits of state into a single `isVisible` prop.

### DIFF
--- a/Artsy/App/ARScreenPresenterModule.m
+++ b/Artsy/App/ARScreenPresenterModule.m
@@ -59,6 +59,21 @@ RCT_EXPORT_METHOD(pushView:(nonnull NSString *)currentTabStackID viewDescriptor:
         deepLinkVC = vc;
         return;
     }
+    
+    if ([vc isKindOfClass:ARComponentViewController.class]) {
+        ARComponentViewController *reactVC = (ARComponentViewController *)vc;
+        // if we're in a modal, get the modal id from the stack map
+        NSString *stackID = currentTabStackID;
+        if ([currentlyPresentedVC isKindOfClass:ARModalWithBottomSafeAreaViewController.class]) {
+            for(id key in [self.class cachedNavigationStacks]) {
+                if ([self.class cachedNavigationStacks][key] == stack) {
+                    stackID = key;
+                    break;
+                }
+            }
+        }
+        [reactVC setProperty:currentTabStackID forKey:@"navStackID"];
+    }
 
     [stack pushViewController:vc animated:YES];
 }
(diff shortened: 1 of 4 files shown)


Artsy/App/ARScreenPresenterModule.m
```diff
@@ -59,6 +59,21 @@ - (dispatch_queue_t)methodQueue;
...
         return;
     }
+    
+    if ([vc isKindOfClass:ARComponentViewController.class]) {
+        ARComponentViewController *reactVC = (ARComponentViewController *)vc;
+        // if we're in a modal, get the modal id from the stack map
+        NSString *stackID = currentTabStackID;
+        if ([currentlyPresentedVC isKindOfClass:ARModalWithBottomSafeAreaViewController.class]) {
+            for(id key in [self.class cachedNavigationStacks]) {
+                if ([self.class cachedNavigationStacks][key] == stack) {
```
ashfurrow, line 69: Since both `stack` and the object stored in `cachedNavigationStacks` are `NSString *`, this line will do a pointer comparison instead of comparing the string contents. But because Objective-C deduplicates all string _literals_ at compile time, this would probably work. But it would be better not to rely on it:

```suggestion
                if ([[self.class cachedNavigationStacks][key] isEqualToString:stack]) {
```

ds300: this code could be clearer 🙈 these are `UINavigationController` pointers rather than string pointers. I'll add some comments to explain what's going on.

ashfurrow: Ahhhh you're right, for some reason I read `[self.class cachedNavigationStacks][key]` as getting the _key_ and not the value from the `NSDictionary` 😅 Comments sound good 👍

brainbicycle (approved):

PR #4268, opened 2020-12-16
fix(my collection): use image carousel from standard artwork view

admbtlr:

This PR resolves [CX-894]

### Description

This PR adds the ImageCarousel component from the standard Artwork view in to the My Collection Artwork view, which solves the issue in the ticket about too-tall images, and also makes the UX more consistent.

image

Outstanding issue:
- Images are not added to the carousel until they have been processed by Gemini. One solution could be to add a default "Image still processing" image that is used in this situation
- Should I add a grey background to the images?

[CX-894]: https://artsyproduct.atlassian.net/browse/CX-894


src/lib/Scenes/Artwork/Components/ImageCarousel/ImageCarousel.tsx
```diff
@@ -30,34 +30,36 @@ export const ImageCarousel = (props: ImageCarouselProps) => {
...
-        }
+        // if (!image.height || !image.width || !image.url) {
+        //   // something is very wrong
+        //   return null
+        // }
         // @ts-expect-error STRICTNESS_MIGRATION --- 🚨 Unsafe legacy code 🚨 Please delete this and fix any type errors if you have time 🙏
-        const { width, height } = fitInside(embeddedCardBoundingBox, image)
+
+        // we want to allow for the possibility that this image is still processing
+        // in which case we should permanently show the loading spinner provided by ImageWithLoadingState
```
ds300, line 43: looks like `ImageWithLoadingState` doesn't show a loading spinner any more, despite the description in the comments of that file. It just shows a grey square until the image loads.

In the case of my collection, a permanent loading spinner is not great UX anyway. We want to let the user know that their photos are still being processed, but a loading spinner implies they're just taking a while to load and will show up at some point.

src/lib/Scenes/MyCollection/Screens/Artwork/Components/MyCollectionArtworkHeader.tsx
```diff
@@ -33,40 +30,25 @@ const MyCollectionArtworkHeader: React.FC<MyCollectionArtworkHeaderProps> = (pro
...
+        <Flex style={{ height: 300, alignItems: "center", justifyContent: "center", backgroundColor: color("black5") }}>
           <ArtworkIcon style={{ opacity: 0.6 }} height={100} width={100} />
           <Text style={{ opacity: 0.6 }}>
             {images && images?.length > 0 ? "Processing photos" : "Processing photo"}
           </Text>
         </Flex>
       )
     } else {
-      // TODO: figure out if "normalized" is the correct version
+      const screenDimensions = Dimensions.get("screen")
```
ds300, line 44:
```suggestion
      const screenDimensions = useScreenDimensions()
```

for iPad, which allows rotation

ds300 (commented):
This does solve the image height issue, but it's a pretty drastic UX change. Did you have a chance to run this by Sam? This change essentially removes the full 'Images' list view.

admbtlr:
> This does solve the image height issue, but it's a pretty drastic UX change. Did you have a chance to run this by Sam? This change essentially removes the full 'Images' list view.

Yes, we discussed it on Slack: https://artsy.slack.com/archives/C01B2P6LJUU/p1607707463344800?thread_ts=1607707356.344000&cid=C01B2P6LJUU

src/lib/Scenes/MyCollection/Screens/Artwork/Components/MyCollectionArtworkHeader.tsx
```diff
@@ -33,40 +30,25 @@ const MyCollectionArtworkHeader: React.FC<MyCollectionArtworkHeaderProps> = (pro
...
-    return stillProcessing
-  }
-
   const renderMainImageView = () => {
     if (!isImage(defaultImage) || imageIsProcessing(defaultImage)) {
       return (
-        <Flex
-          style={{ height: 300, alignItems: "center", justifyContent: "center", backgroundColor: color("black10") }}
-        >
+        <Flex style={{ height: 300, alignItems: "center", justifyContent: "center", backgroundColor: color("black5") }}>
```
brainbicycle, line 36: color change here deliberate?

brainbicycle (approved):
This looks good to me, agree with David that we should avoid long loading spinners for processing photos but it looks like this _should_ show processing state instead of carousel in that case. Thanks for proposing this and doing it!

src/lib/Scenes/Artwork/Components/ImageCarousel/ImageCarousel.tsx
```diff
@@ -30,34 +30,36 @@ export const ImageCarousel = (props: ImageCarouselProps) => {
...
     let result = props.images
       .map((image) => {
-        if (!image.height || !image.width || !image.url) {
-          // something is very wrong
-          return null
-        }
+        // if (!image.height || !image.width || !image.url) {
+        //   // something is very wrong
+        //   return null
+        // }
```
brainbicycle, line 39: ✔ Mobile Practice Meeting @done (20-12-18 10:37)
      - revisit QA script
      ✔ Prep @done (20-12-18 10:37)
        - update the instructions for

admbtlr:
The problem here is that I can't easily inject local photos into the carousel's state, in order to show an image before it's finished processing on Gemini. In fact the carousel doesn't update at all after adding an image. Doing these would require quite a lot of changes to the carousel component, so I think I'm going to close this PR for now and make a simpler fix for the initial problem, by reducing the size of portrait images. 

Maybe we can make a ticket to use the image carousel and revisit this at a later date?

PR #4274, opened 2020-12-18
chore: add jest-watch-typeahead (optional)

erikdstock:
The type of this PR is: **enhancement**
#minor

This PR includes a new dependency - totally up to the repo bosses if we want to merge it or not. It adds [jest-watch-typeahead](https://github.com/jest-community/jest-watch-typeahead) which gives a preview to the `jest --watch` command when typing a file or test name pattern. It's straightforward to use and i have found it convenient in the past.  Again, no pressure to adopt it if we don't want to.


ashfurrow (approved):
Yeah, makes sense to me 👍

ds300 (approved):
Good idea! 👍

pvinis:
o nice!

PR #4276, opened 2020-12-18
Refresh active inbox view when revisiting

lilyfromseattle:
# [PURCHASE-2307]

When the user revisits the inbox view it should refresh so they don't need to pull to update it.

Paired with @erikdstock 

The logic for refreshing the inbox when `isVisible` was true already existed, but isVisible wasn't being successfully passed to the inbox. This PR fixes that. This PR relies on [this one](https://github.com/artsy/eigen/pull/4267), which fixes `isVisible` so that it respects bottom bar tab switches.

In the future, the `Inbox` component should be refactored into a functional component, and the refresh should be done using hooks, however, I wanted to hold off on that portion so as not. to break the recently completed analytics work. 

[PURCHASE-2307]: https://artsyproduct.atlassian.net/browse/PURCHASE-2307

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -15,6 +15,7 @@ upcoming:
     - Fix bug preventing empty values from being saved (my collection) - brian
     - Add improved phone number input - david
     - Add support for deleting photos in my collection - brian. barry
+    - Inbox refreshes when tab is visited - lily
 
 releases:
   - version: 6.7.3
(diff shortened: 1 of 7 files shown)


erikdstock (commented):
These are very old comments that i guess i never submitted, but in short the linked PR #4280 contains the majority of these changes plus updates to work with analytics.

PR #4281, opened 2020-12-21
logic(bought-in): change the logic for the string in auction insights results

pvinis:

This PR resolves [CX-792]

### Description

New logic is to swap the check of bought in with the past month check. 

[CX-792]: https://artsyproduct.atlassian.net/browse/CX-792

### DIFF
--- a/src/lib/Components/Artist/ArtistInsights/ArtistInsightsAuctionResult.tsx
+++ b/src/lib/Components/Artist/ArtistInsights/ArtistInsightsAuctionResult.tsx
@@ -95,7 +95,7 @@ const ArtistInsightsAuctionResult: React.FC<Props> = ({ auctionResult }) => {
               </Flex>
             ) : (
               <Text variant="mediumText" style={{ width: 70 }} textAlign="right">
-                {isFromPastMonth ? "Awaiting results" : auctionResult.boughtIn === true ? "Bought in" : "Not available"}
+                {auctionResult.boughtIn === true ? "Bought in" : isFromPastMonth ? "Awaiting results" : "Not available"}
               </Text>
             )}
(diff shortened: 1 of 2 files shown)


src/lib/Components/Artist/ArtistInsights/ArtistInsightsAuctionResult.tsx
```diff
@@ -95,7 +95,7 @@ const ArtistInsightsAuctionResult: React.FC<Props> = ({ auctionResult }) => {
               </Flex>
             ) : (
               <Text variant="mediumText" style={{ width: 70 }} textAlign="right">
-                {isFromPastMonth ? "Awaiting results" : auctionResult.boughtIn === true ? "Bought in" : "Not available"}
+                {auctionResult.boughtIn === true ? "Bought in" : isFromPastMonth ? "Awaiting results" : "Not available"}
```
brainbicycle, line 98: This looks good 👍 This is a bit more on personal opinion side of things but I personally think this would be a bit more readable as **if, else if, else** statements rather then two combined ternary expressions

brainbicycle (approved):

src/lib/Components/Artist/ArtistInsights/ArtistInsightsAuctionResult.tsx
pvinis: I tried now, but we can't have `if`s inside the `{}` in jsx, so we would need to put that `if` for the string login somewhere above the `render return`, and I feel it's too specific a logic to be up there. Let's keep it with the ternaries?

annacarey (approved):
lgtm!

PR #4283, opened 2020-12-21
fix: workaround for iOS 14 password alert issue

brainbicycle:

This PR resolves [CX-909]

### Description

The alert to log out users after changing password was not showing up on iOS 14. For some reason the onDismiss call which presents the alert in the LoadingModal appears to be being called while the loadingModal is still in the view hierarchy causing the alert to either be presented in the loading modal and dismiss immediately as the LoadingModal dismisses or failing to present because the loadingModal has been removed from the hierarchy while trying to present the modal. This adds a slight delay to the alert presentation as a workaround. 

This is not a very good solution so If anyone has suggestions for a better solution here or some context on why this is happening would be appreciated 🙏 

[CX-909]: https://artsyproduct.atlassian.net/browse/CX-909


src/lib/Scenes/MyAccount/Components/MyAccountFieldEditScreen.tsx
```diff
@@ -103,10 +103,14 @@ export const MyAccountFieldEditScreen = React.forwardRef<
...
-                  afterLoadingAlert.current = undefined
-                }
+                // Workaround for an iOS 14 issue, modal onDismiss is being called while the
+                // view still appears to be in the hierarchy
+                setTimeout(() => {
+                  if (afterLoadingAlert.current) {
+                    Alert.alert(...afterLoadingAlert.current)
+                    afterLoadingAlert.current = undefined
+                  }
+                }, 150)
```
ashfurrow, line 113: This would be good 👍 We have also used `requestAnimationFrame()` for this kind of thing before (it's roughly analogous to `dispatch_async(dispatch_get_main_queue(), ...)`.

brainbicycle: This doesn't appear to work for this case unfortunately, I tried replacing and still got the presentation happening on the loading modal :(

pvinis:
pretty weird, considering the last rn upgrade had a fix for these modal dismissal timing issues but ok. maybe a good thing to add for HACKS, to remove when were have a better modal situation.

PR #4296, opened 2020-12-30
fix: ignore membership expiration warnings for message notifications #trivial

brainbicycle:

### Description

Don't send a slack notification when App Store connect warns us about dev account expiration date. This notification is meant to be for new license agreements that need to be reviewed.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -194,7 +194,8 @@ lane :notify_if_new_license_agreement do
   client.team_id = '479887'
   messages = Spaceship::Tunes.client.fetch_program_license_agreement_messages
 
-  if !messages.empty?
+  # ignore membership expiration warnings, auto-renew should take care of
+  if !messages.empty? and !messages[0].include? "membership expiration"
     message = <<~MSG
                 :apple: :handshake: :pencil:
                 There is a new developer program agreement that needs to be signed to continue shipping!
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -194,7 +194,8 @@ lane :notify_if_new_license_agreement do
   client.team_id = '479887'
   messages = Spaceship::Tunes.client.fetch_program_license_agreement_messages
 
-  if !messages.empty?
+  # ignore membership expiration warnings, auto-renew should take care of
+  if !messages.empty? and !messages[0].include? "membership expiration"
```
ashfurrow, line 198: This will work, but there are two things that make it feel "un-Ruby" to me. 

Two things:

- `and` (and `or`) in Ruby don't always work the way you'd expect them to, so it's preferable to use `&&` and `||` instead. [See this blog post](https://www.compassitesinc.com/difference-between-and-and-and-and-or-in-ruby/) for more info.
- This expression is `if !thing && !other_thing` is equivalent to `if !(thing || other_thing)` (by DeMorgan's Law). We can further simplify this by using the `unless` keyword, which is like `if` but negated.

```suggestion
  unless messages.empty? || messages[0].include?("membership expiration")
```

Ruby! What a weird language.

ashfurrow (approved):

PR #4299, opened 2021-01-04
feat: add support for encoded URIs

MounirDhahri:

This PR resolves [CX-925]

### Description
- Add support for encoded URIs

PS: Not sure if this is required in case we want the params to be distinguishable (https://artsy.slack.com/archives/C04KVDZU6/p1609168424086400?thread_ts=1608808952.082700&cid=C04KVDZU6)

https://user-images.githubusercontent.com/11945712/103544265-18a2fa00-4ea0-11eb-8c21-ea706223ed9b.mov

[CX-925]: https://artsyproduct.atlassian.net/browse/CX-925

### DIFF
--- a/src/lib/navigation/__tests__/routes-tests.tsx
+++ b/src/lib/navigation/__tests__/routes-tests.tsx
@@ -75,6 +75,32 @@ describe("artsy.net routes", () => {
         "type": "match",
       }
     `)
+    expect(matchRoute("artist/josef-albers?utm_medium=social&utm_source=instagram-story&utm_campaign=dp."))
+      .toMatchInlineSnapshot(`
+      Object {
+        "module": "Artist",
+        "params": Object {
+          "artistID": "josef-albers",
+          "utm_campaign": "dp.",
+          "utm_medium": "social",
+          "utm_source": "instagram-story",
+        },
+        "type": "match",
+      }
+    `)
+    expect(matchRoute("/artist/josef-albers%3Futm_medium%3Dsocial%26utm_source%3Dinstagram-story%26utm_campaign%3Ddp."))
+      .toMatchInlineSnapshot(`
+      Object {
+        "module": "Artist",
+        "params": Object {
+          "artistID": "josef-albers",
+          "utm_campaign": "dp.",
+          "utm_medium": "social",
+          "utm_source": "instagram-story",
+        },
+        "type": "match",
+      }
+    `)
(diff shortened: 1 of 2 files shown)


src/lib/navigation/__tests__/routes-tests.tsx
```diff
@@ -96,6 +122,36 @@ describe("artsy.net routes", () => {
...
+        "params": Object {
+          "artworkID": "yayoi-kusama-red-pumpkin",
+          "utm_campaign": "dp.",
+          "utm_medium": "social",
+          "utm_source": "instagram-story",
+        },
+        "type": "match",
+      }
+    `)
+    expect(
```
brainbicycle, line 139: Nice tests and thanks for looking into this! I think this probably works with our parsing but one thing would be good to make sure of is this can handle artworks or artists that have these encoded characters within them for example

`expect(
      matchRoute("/artwork/more%26more")
)`

` // assert the artwork id is the decoded value 'more&more'`

MounirDhahri: Thanks, I will add it now 🙏

brainbicycle (approved):

src/lib/navigation/__tests__/routes-tests.tsx
```diff
@@ -96,6 +131,45 @@ describe("artsy.net routes", () => {
...
+        "params": Object {
+          "artworkID": "yayoi-kusama-red-pumpkin",
+          "utm_campaign": "dp.",
+          "utm_medium": "social",
+          "utm_source": "instagram-story",
+        },
+        "type": "match",
+      }
+    `)
   })
```
ds300, line 173: I think this isn't a valid way to encode a URL query string.

It makes sense that we should decode the path parts and the query keys and values, but the query string delimiters `?` and `&` must be plain text.

So, rather than doing `url.parse(decodeURIComponents(href))` we should be doing `decode(url.parse(href))` where `decode(url)` transforms the query keys + values and the pathname using `decodeURIComponent`

ds300: added a ticket https://artsyproduct.atlassian.net/browse/CX-939

MounirDhahri: Thanks for noticing and for creating a ticket 🙏

PR #4301, opened 2021-01-04
fix: safer empty location check #trivial

brainbicycle:

This PR resolves [CX-927]

### Description

Check for undefined values rather than explicitly null values. A crash was occurring on entering city guide and giving location permission, best guess is the first location update callback was returning an  empty/undefined location object and the location.coords check was then causing a crash accessing an undefined property.

IMG_0974

[CX-927]: https://artsyproduct.atlassian.net/browse/CX-927


pvinis (approved):
huh, makes sense. good fix 👍

PR #4304, opened 2021-01-05
fix: alignment issue with fairs header #trivial

brainbicycle:

This PR resolves [CX-930]

### Description

Looks like there is an alignment issue with the list header prop for horizontal rails. This uses the same logic of the but moves header outside this prop. 

### Before:

featured_fairs_header_before

### After:
featured_fairs_header

[CX-930]: https://artsyproduct.atlassian.net/browse/CX-930


brainbicycle:
🙈 self-merging because it is a tiny PR so beta is ready for tomorrow, but if you want anything changed don't hesitate to let me know!

PR #4313, opened 2021-01-08
fix(BidFlow): App crash when two users bid simultaneously

MounirDhahri:

This PR resolves [CX-]

### Description
- Fix app crash when two users bid simultaneously on iOS 14.2
- Refactor The BidFow to use react-navigation instead of react-native-navigator-ios


src/lib/Components/Bidding/Components/PaymentInfo.tsx
```diff
@@ -29,31 +31,48 @@ export class PaymentInfo extends React.Component<PaymentInfoProps> {
...
+
+    if (this.props.navigation) {
+      this.props.navigation.navigate("BillingAddressScreen", {
         onSubmit: (address) => this.onBillingAddressAdded(address),
         billingAddress: this.props.billingAddress,
-        navigator: this.props.navigator,
-      },
-    })
+      })
+    }
```
MounirDhahri, line 75: This check is needed since `PaymentInfo` is also used in `Registration.tsx` which is still base on `react-native-navigator-ios`.

src/lib/Components/Bidding/Screens/BillingAddress2.tsx
```diff
@@ -0,0 +1,363 @@
...
+    country?: string
+    postalCode?: string
+  }
+}
+
+@screenTrack({
+  context_screen: Schema.PageNames.BidFlowBillingAddressPage,
+  context_screen_owner_type: null,
+})
+export class BillingAddressScreen extends React.Component<BillingAddressProps, BillingAddressState> {
```
MounirDhahri, line 80: Same here since `BillingAddress` is used in other instances that use `react-native-navigator-ios`

src/lib/Components/Bidding/__tests__/BidFlow-tests.tsx
```diff
@@ -1,246 +1,64 @@
...
+      render={({ props }) => {
+        if (props?.artwork?.sale_artwork && props?.me) {
+          return <BidFlowFragmentContainer sale_artwork={props.artwork.sale_artwork} me={props.me} />
+        }
+        return null
+      }}
     />
   )
 
-  screen.root.findByType(MaxBidPicker).instance.props.onValueChange(null, 2)
```
MounirDhahri, line 98: this test was no longer needed because it is testing the mutation for the credit card which is no longer part of this screen since we changed the navigation logic

src/lib/Components/Bidding/Screens/BidResult.tsx
```diff
@@ -72,9 +74,9 @@ export class BidResult extends React.Component<BidResultProps> {
   }
 
   render() {
-    const { sale_artwork, bidderPositionResult } = this.props
-    // @ts-expect-error STRICTNESS_MIGRATION --- 🚨 Unsafe legacy code 🚨 Please delete this and fix any type errors if you have time 🙏
-    const { liveStartAt, endAt } = sale_artwork.sale
+    const { sale_artwork, route } = this.props
+    const { bidderPositionResult } = route.params!
+    const { liveStartAt, endAt } = sale_artwork.sale!
```
brainbicycle, line 79: Is it safe to do `!` here?  think we should avoid where possible and have a fallback for `null` / `undefined` state

src/lib/Components/Bidding/Screens/ConfirmBid/index.tsx
```diff
@@ -315,13 +317,16 @@ export class ConfirmBid extends React.Component<ConfirmBidProps, ConfirmBidState
   }
 
   verifyBidderPosition(results: ConfirmBidCreateBidderPositionMutationResponse) {
-    // @ts-expect-error STRICTNESS_MIGRATION --- 🚨 Unsafe legacy code 🚨 Please delete this and fix any type errors if you have time 🙏
-    const { result } = results.createBidderPosition
+    const { result } = results.createBidderPosition!
```
brainbicycle, line 320: Similar to above is it safe to use `!` or should we be handling null/undefined case?

src/lib/Components/Bidding/Screens/BillingAddress2.tsx
```diff
@@ -0,0 +1,363 @@
...
+  // @ts-expect-error STRICTNESS_MIGRATION --- 🚨 Unsafe legacy code 🚨 Please delete this and fix any type errors if you have time 🙏
+  private addressLine1: StyledInputInterface
+  // @ts-expect-error STRICTNESS_MIGRATION --- 🚨 Unsafe legacy code 🚨 Please delete this and fix any type errors if you have time 🙏
+  private addressLine2: StyledInputInterface
+  // @ts-expect-error STRICTNESS_MIGRATION --- 🚨 Unsafe legacy code 🚨 Please delete this and fix any type errors if you have time 🙏
+  private city: StyledInputInterface
+  // @ts-expect-error STRICTNESS_MIGRATION --- 🚨 Unsafe legacy code 🚨 Please delete this and fix any type errors if you have time 🙏
+  private stateProvinceRegion: StyledInputInterface
```
brainbicycle, line 88: Are all these strictness migrations copied over from other BillingAddress? are they all necessary?

src/lib/Components/Bidding/Screens/ConfirmBid/index.tsx
```diff
@@ -392,29 +397,14 @@ export class ConfirmBid extends React.Component<ConfirmBidProps, ConfirmBidState
...
-    this.props.navigator.push({
-      component: SelectMaxBidEdit,
-      title: "",
-      passProps: {
-        increments: this.props.increments,
-        selectedBidIndex: this.state.selectedBidIndex,
-        updateSelectedBid: this.updateSelectedBid.bind(this),
-      },
-    })
+    this.props.navigation.goBack()
```
brainbicycle, line 400: ✨

brainbicycle (commented):
Nice work on this and the navigation looks cleaner ✨ 
I have a couple questions about use of `!` in a few places and asserting types, I think generally we should avoid the non-null assertions and handle the null case and similarly think when a type is unspecific should try to handle the case where it doesn't match the type we think it is. In other words 
```
if (thing) { 
  thing.useTheThing()
} else {
    // handle thing not existing
}
```

rather  than 
`thing!.useTheThing()`

That said I know it is not always possible so if it can't be done in some of these cases just let me know. 

Other thing I am wondering is if this PR needs to change with the removal of `react-navigator-ios` I see it is imported in a few of these files.

Thanks for looking into this!

PR #4318, opened 2021-01-12
feat: artist past shows improvements

admbtlr:

### Description

This is a kind of hybrid PR. It does three small things, two of which are kind of proposals or PoCs of UX tweaks that I think we could potentially roll out everywhere. 

The three small things:

1. Add a skeleton loading screen to Past Shows
2. Add a sticky scroll header to Past Shows
3. Add an effect that I'm going to call "springy overscroll" to Past Shows

Note that I also removed Pull to Refresh, since I don't think it makes sense in this screen.

I'm totally happy to revert (2) and/or (3), depending on what people think. I'd be interested in hearing from @samchieng and @joanna20, and maybe having a brief discussion about using these tweaks in other places.

The Skeleton:

image

The sticky scroll header and springy overscroll:

https://user-images.githubusercontent.com/1815204/104301621-296af580-54c8-11eb-868e-e429431f19d6.mov

I'll update the changelog once we decide which parts to keep.


samchieng:
Oh wow I love all these 3 small things! Curious: how easy is it to roll out these same UX improvements on more pages? We have #1 on a lot of screens and its started to become a standard pattern for us, but i've been hoping to roll out #2 as a standard pattern across the app as well. #3 also feels like a better iteration on pull to refresh in this context.

Small design note for the sticky header is that we'll want to change the text to use the `medium` weight instead of `regular`!

joanna20:
+1 love these improvements!

src/lib/Scenes/ArtistShows/ArtistShows2.tsx
```diff
@@ -17,20 +19,14 @@ interface Props {
...
-      if (error) {
-        console.log(error.message)
-      }
-      setIsRefreshing(false)
-    })
-  }
+  const { headerElement, scrollProps, scrollAnim } = useStickyScrollHeader({
+    headerText: `${artist.name} – Past Shows`,
+    fadeInStart: 50,
+  })
```
MounirDhahri, line 29: 🙏

src/lib/Scenes/ArtistShows/ArtistShows2.tsx
```diff
@@ -157,9 +169,42 @@ export const ArtistShows2QueryRenderer: React.FC<{ artistID: string }> = ({ arti
...
         Container: ArtistShows2PaginationContainer,
-        renderPlaceholder: () => <Text variant="largeTitle">Placeholder</Text>,
+        renderPlaceholder: LoadingSkeleton,
       })}
       variables={{ artistID }}
     />
   )
 }
+
+const LoadingSkeleton = () => {
```
MounirDhahri, line 179: ✨

src/lib/Scenes/ArtistShows/ArtistShows2.tsx
```diff
@@ -45,36 +41,52 @@ const ArtistShows2: React.FC<Props> = ({ artist, relay }) => {
...
-              Past Shows
-            </Text>
-          </>
-        )
-      }}
-      renderItem={({ item }) => <ArtistShowFragmentContainer show={item} styles={showStyles} />}
-      keyExtractor={({ id }) => id}
-      onEndReachedThreshold={0.2}
-      ItemSeparatorComponent={() => <Spacer height={20} />}
-      refreshControl={<RefreshControl refreshing={isRefreshing} onRefresh={handleRefresh} />}
```
MounirDhahri, line 67: I guess it's fine in this case to remove the pull to refresh in this screen since shows do not get updated that often but we may need to update the QueryRenderer cacheConfig to force refresh on screen load. `cacheConfig={{ force: true }}`

MounirDhahri (approved):

MounirDhahri:
Thanks Adam, this really looks great. We definitely need to be using these UX adjustments everywhere.

src/lib/Scenes/ArtistShows/ArtistShows2.tsx
admbtlr: Ah, good point, thanks!

MounirDhahri:
Hi Adam, I have one minor comment that I missed the last time, It looks like the back button is slightly hiding the screen title on small devices (I am using an iPhone 8)
Screenshot 2021-01-15 at 18 51 28

MounirDhahri:
Thanks Adam 🎉

src/lib/utils/useElasticOverscroll.tsx
```diff
@@ -0,0 +1,42 @@
+import React from "react"
+import { Animated } from "react-native"
+
+export const useElasticOverscroll = (header: JSX.Element, scrollAnim = new Animated.Value(0)) => {
```
MounirDhahri, line 4: Thanks for extracting this ✨

PR #4321, opened 2021-01-12
feat: confirm phone number in consignment flow

ds300:
This PR resolves [CX-607]

### Description

- Removes defunct 'welcome' screen that used to be the root view in the consignments flow
- Refreshes the typography in the consignments flow
- Adds swipe handle for dismissing the consignments modal on iOS 13+
  - Adds close button as a fallback for iOS 12 (we don't support iOS earlier than 12)
- Adds phone number confirmation step to consignments flow

#### Typography refresh

image

#### Modal closing

image

image

#### Phone number confirmation

image

[CX-607]: https://artsyproduct.atlassian.net/browse/CX-607


src/lib/Scenes/Consignments/ConsignmentsHome/ConsignmentsSubmissionForm.tsx
```diff
@@ -1,14 +1,55 @@
...
+import { dismissModal } from "lib/navigation/navigate"
+import { CloseIcon, color, Touchable } from "palette"
 import React from "react"
+import { Platform, View } from "react-native"
 import NavigatorIOS from "react-native-navigator-ios"
+import Overview from "../Screens/Overview"
 
-export const ConsignmentsSubmissionForm: React.FC = () => {
-  const initialRoute = { component: Overview }
+const isIOS13Plus = Platform.OS === "ios" && Number(Platform.Version) >= 13
```
brainbicycle, line 8: **nitpick: (non-blocking)** We have an `osMajorVersion` util in `hardware.ts` which we might be able to use here but this works fine too. Also realizing that `hardware.ts` is probably not a good place for this 😅

src/lib/Scenes/Consignments/Screens/ConfirmContactInfo.tsx
```diff
@@ -0,0 +1,143 @@
...
+import Confirmation from "./Confirmation"
+
+const ConfirmContactInfo: React.FC<{
+  me: ConfirmContactInfo_me | null
+
+  submissionRequestValidationCheck: () => boolean
+  navigator: NavigatorIOS
+}> = ({ me, submissionRequestValidationCheck, navigator }) => {
+  const { width, height } = useScreenDimensions()
+  const isPad = width > 700
```
brainbicycle, line 25: **nitpick: (non-blocking):** we have an isPad util in `hardware.ts` but what that function really is right now is iPad specific, would be good to have a standard tablet detection function and use that

brainbicycle (approved):
Looks awesome! A couple minor non-blocking comments, just thinking might be good to have standard os version checks and tablet detection ahead of Android but feel free to ignore.

PR #4323, opened 2021-01-12
fix: Display correct sale date for auction results

pepopowitz:
Collaboration with @The-Beez-Kneez!!!

This PR resolves [CX-943]

### Description

The sale date is parsed as a local timestamp, so dates were getting nudged off by 1 (for those of us west of gmt). 

In this PR we convert to UTC before displaying them.

Auction result list: 

(The Andy Warhol "Wine Label" items were previously showing Dec 14)

image

Auction result detail:
(This was previously showing Dec 14)

image

[CX-943]: https://artsyproduct.atlassian.net/browse/CX-943

### DIFF
--- a/src/lib/Components/Lists/AuctionResult.tsx
+++ b/src/lib/Components/Lists/AuctionResult.tsx
@@ -61,7 +61,7 @@ const AuctionResult: React.FC<Props> = ({ auctionResult, onPress }) => {
             )}
 
             {!!auctionResult.saleDate && (
-              <Text variant="small" color="black60" numberOfLines={1}>
+              <Text variant="small" color="black60" numberOfLines={1} testID="saleInfo">
                 {moment(auctionResult.saleDate).utc().format("MMM D, YYYY")}
                 {` ${bullet} `}
                 {auctionResult.organization}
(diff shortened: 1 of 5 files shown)


admbtlr (approved):
🎉 🎉 🎉

MounirDhahri:
👏👏👏

src/lib/Components/Lists/AuctionResult.tsx
```diff
@@ -61,8 +61,8 @@ const AuctionResult: React.FC<Props> = ({ auctionResult, onPress }) => {
             )}
 
             {!!auctionResult.saleDate && (
-              <Text variant="small" color="black60" numberOfLines={1}>
-                {moment(auctionResult.saleDate).format("MMM D, YYYY")}
+              <Text variant="small" color="black60" numberOfLines={1} testID="saleInfo">
+                {moment(auctionResult.saleDate).utc().format("MMM D, YYYY")}
```
starsirius, line 65: Out of curiosity, are we showing the sale date in the local timezone where the sale happens, or simply in UTC?

The screenshot attached in the ticket looks like it's local time zone, and in our design it's not immediately clear what time zone it is.

Phillips2

pepopowitz: I'm not exactly sure -- CollectorIQ (for the large majority of lots, including this one) is deciding what to store for the sale date that we're rendering, by scraping the phillips site (the screenshot above). We get a date without a time from them, and store it as midnight UTC.

starsirius: Got it. Then it makes sense to display as is because we don't know the time zone. 👍 

One alternative we can do here is to move this logic upstream to MP so that eigen (and other clients) doesn't need to convert to UTC in all places, and doesn't need a new release for logic changes.

PR #4331, opened 2021-01-15
feature(android): adding some build scripts

pvinis:

This PR resolves [CX-883] and [CX-975]

### Description

- tests every PR, by building the android app alongside the ios one.
https://app.circleci.com/pipelines/github/artsy/eigen/8572/workflows/98991f57-45f3-41c2-8914-f615a7f0d3bc
- deploys ios beta like before, no breakage
https://app.circleci.com/pipelines/github/artsy/eigen/8570/workflows/b99c3df1-9712-4b5d-98d5-c7b5ab93e4f9
- deploys android beta
https://app.circleci.com/pipelines/github/artsy/eigen/8578/workflows/9e61d8ad-dabd-4ea3-bc00-2c9b74317a10

### Work after
- [ ] After this is merged, we can remove the `beta` branch.
- [ ] Add lane for release promotion (alpha->production) and test it.
- [ ] Fill in `deploy_to_play_store.md`.

Main changes here:
- Added android CI that tests and deploys.
- Removed `Makefile` and replaced it with small scripts in `./scripts/`
That change was necessary to make the scripts run the same way on macos and linux/docker containers on circleci. We had problems with that while implementing the android ci.
- Added some documentation for setting up etc.
- Added `app.json` that we can use for other static shared things between platforms.
- Simplified our Gemfile setup. Now we only have one, no need to mess with `bundle install` in root and inside fastlane dir.
- A couple of tiny changes in `README.md` and `THANKS.md`.

### Notes
The new scripts all have no `.sh` extension. We shouldn't care if it's a bashscript or a js script or a binary or whatever else. They do use `set -euxo pipefail` (nice article about it [here](https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/)) to keep us all safe and sane.

I would encourage everyone to take a quick look, to see the structure of
```
hashbang
set error stuff

maybe source the source-for-bash-env file

rest of the script here
```
Also if lines are looooong, I split them with a pipe `|` at the end of the line because the alternative with
```
first part \
 | second part
```
is not as nice.
I also used `shellcheck` to make sure I didn't miss anything obvious.

I named a couple of scripts that do both ios and android as `*-both`. Like `deploy-ios`, `deploy-android`, `deploy-both`. This is to make it easy to autocomplete. With typing `de<tab>` you get to `deploy-`, ready to type `i<tab>` or `a<tab>` or `b<tab>`. With just `deploy` for both, autocomplete would still need the dev to type a `-` after the first tab.

`release.keystore` and `android-secret.json` are in the s3 citadel under `eigen/`. Locally we don't need them, CircleCI gets it from there. Also available in 1P.

[CX-883]: https://artsyproduct.atlassian.net/browse/CX-883

[CX-975]: https://artsyproduct.atlassian.net/browse/CX-975


android/app/build.gradle
```diff
@@ -224,3 +237,16 @@ task copyDownloadableDepsToLibs(type: Copy) {
...
+   def stderr = new ByteArrayOutputStream()
+   exec {
+       commandLine 'security', '-q', 'find-generic-password', '-a', currentUser, '-s', keyChain, '-w'
+       standardOutput = stdout
+       errorOutput = stderr
+       ignoreExitValue true
+   }
+   //noinspection GroovyAssignabilityCheck
+      stdout.toString().trim()
+}
```
ds300, line 252: we don't need this, right?

ds300 (commented):
Amazing work! Thank you for updating all the docs!!!

.circleci/config.yml
```diff
@@ -302,10 +316,44 @@ jobs:
...
-          branch: beta
+          branch: beta-ios
       - run:
           name: Deploy if beta
-          command: make deploy_if_beta_branch
+          command: ./scripts/deploy-if-beta-branch-ios
+
+  build-test-app-android:
+    environment:
+      UPLOAD_IOS_SNAPSHOT_BUCKET_NAME: eigen-ci
```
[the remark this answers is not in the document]
pvinis: We don't. I will remove it from ios too. In this PR I didn't try to optimize the CI stuff at all, so good catch.

(I have it written down for a future friday, to check a bit, because I think the CI is building the apps more times than it is needed 🤔.)

android/app/build.gradle
pvinis: Indeed, we didn't use that in the end. 👍

MounirDhahri:
Nice job!

PR #4337, opened 2021-01-15
fix: Add yarn install to make flip_table script

ashleyjelks:

### Description

The `make flip_table` command was failing because node modules need to be installed before calling `yarn prettier`. This PR updates the script.

https://artsy.slack.com/archives/C02BAQ5K7/p1610750140007000


pvinis:
Just to be clear, `pod install` does `yarn install` as the first step. The problem here was not that `yarn install` was missing. It was that I recently moved `update_echo` above the `pod install` step, but since it uses `prettier`, it should have stayed last. I wanted to have it above the pods step because there is no need for it to wait for that, so adding the explicit `yarn install` step helps keep the `update_echo` in a more pleasant place in the script, as well as make it clear for devs that `yarn install` will be run at some point in that script. Thanks!

PR #4344, opened 2021-01-19
WIP - Watched Lots

xtina-starr:

This PR resolves [CX-]

### DIFF
--- a/src/lib/Scenes/MyBids/Components/WatchedLot.tsx
+++ b/src/lib/Scenes/MyBids/Components/WatchedLot.tsx
@@ -1,0 +1,100 @@
+import { ActionType, ContextModule, OwnerType } from "@artsy/cohesion"
+import { WatchedLot_lotStanding } from "__generated__/WatchedLot_lotStanding.graphql"
+import { navigate } from "lib/navigation/navigate"
+import { isSmallScreen } from "lib/Scenes/MyBids/helpers/screenDimensions"
+import { Flex, Text } from "palette"
+import React from "react"
+import { TouchableOpacity } from "react-native"
+import { createFragmentContainer, graphql } from "react-relay"
+// import { useTracking } from "react-tracking"
+// import { TimelySale } from "../helpers/timely"
+// import { HighestBid, Outbid, ReserveNotMet } from "./BiddingStatuses"
+// import { LotFragmentContainer as Lot } from "./Lot"
+
+interface WatchedLotProps {
+  lotStanding: WatchedLot_lotStanding
+}
+
+export const WatchedLot: React.FC<WatchedLotProps> = ({ lotStanding }) => {
+  console.log("lotStandingggggg:", lotStanding)
+  // const timelySale = TimelySale.create(lotStanding?.saleArtwork?.sale!)
+
+  // const sellingPrice = lotStanding?.lot?.sellingPrice?.display
+  // const bidCount = lotStanding?.lot?.bidCount
+  const { saleArtwork, lot } = lotStanding
+  // const tracking = useTracking()
+
+  // const displayBidCount = (): string | undefined => {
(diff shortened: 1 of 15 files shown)


src/lib/Scenes/MyBids/Components/WatchedLot.tsx
```diff
@@ -0,0 +1,89 @@
...
+// import { TimelySale } from "../helpers/timely"
+// import { HighestBid, Outbid, ReserveNotMet } from "./BiddingStatuses"
+import { LotFragmentContainer as Lot } from "./Lot"
+
+interface WatchedLotProps {
+  lotStanding: WatchedLot_lotStanding
+}
+
+export const WatchedLot: React.FC<WatchedLotProps> = ({ lotStanding }) => {
+  console.log("lotStandingggggg:", JSON.stringify(lotStanding, null, 4))
```
erikdstock, line 20: the lot standing prop here should probably just be renamed `lot`.

The goal here is to have a new "lot" type which will add causality lot state to gravity sale artwork, and then for "lot standing" to be the lot type (in a duck typing sense) plus bidder-specific information. Lot/Lot Standing thus reflect real-world vocabulary and also unite our interfaces in a way we can use `WatchedLot` (with a `lot`) with the same interface as `ActiveLot`/`ClosedLot` (which require a `lotStanding`).

sale artwork = a lot in an auction, but an older interface derived only from gravity
lot = a lot in an auction `lot` (state from causality) plus a `saleArtwork` (gravity data about the sale/artwork)
lot standing = a lot plus the user's current status on it (bid amount, whether they are winning etc)

src/lib/Scenes/MyBids/MyBids.tsx
```diff
@@ -146,6 +153,9 @@ const MyBids: React.FC<MyBidsProps> = (props) => {
             </Join>
           </Flex>
         )}
+        {watchedLots.map((ls) => {
```
erikdstock, line 156: Related to my comment about duck typing above, according to the spec lots should be displayed in order by lot number (after grouping by sale) without regard for whether a bidder has a lot standing (active/closed lot component) or is only watching. There may be a better way to do this than the transformations I have up above but we ultimately want to have a list like:

### Artsy prints and multiples
------------
Lot 1: Winning
Lot 6: Watching
Lot 7: Outbid
Lot 99: Watching

erikdstock (commented):
Realize this is a draft PR but as we may be blocked on the metaphysics side I wanted to get some early feedback in.

src/lib/Scenes/MyBids/Components/WatchedLot.tsx
xtina-starr: Yeah this makes sense.

src/lib/Scenes/MyBids/MyBids.tsx
xtina-starr: Ahh that adds a bit more complexity. So the partition basically is determined by whether the lot is closed or not (what controls `isClosed`)?  

Could we do something like create an array that merges `lotStandings` and `watchedLots` then remove duplicate  `watchedLots` (i.e. watchedLots that already represented by a bid/lotStanding) and do the partition from there?

src/lib/Scenes/MyBids/Components/WatchedLot.tsx
ashleyjelks: @erikdstock to clarify...`sale artwork`, `lot`, and `lot standing` are the same regardless of if its a timed or live auction?

MounirDhahri: I finally understood the difference between (saleArtwork, lot and lotStanding). Thank you @erikdstock ✨✨✨ 🙇‍♂️

PR #4352, opened 2021-01-21
feat(android): configure metaphysics + relay

ds300:
This PR resolves [CX-870]

### Description

<em>Note: This was originally supposed to be part of #4346 but I decided to split it into a separate PR.</em>

This PR makes metaphysics work on android, and sets up a new GraphQL cache based on `AsyncStorage` (since our current ios one is in obj-c). We can probably already replace the obj-c one but I'll leave that until after we've migrated the admin menu since there's an option there to clear the cache which I occasionally use at least.

I added a POC query renderer to the main android page to show the user's name

image

[CX-870]: https://artsyproduct.atlassian.net/browse/CX-870

### DIFF
--- a/src/lib/NativeModules/ArtsyNativeModules.tsx
+++ b/src/lib/NativeModules/ArtsyNativeModules.tsx
@@ -1,4 +1,5 @@
 import { NativeModules as AllNativeModules, Platform } from "react-native"
+import { getLocales, getTimeZone } from "react-native-localize"
 
 const noop: any = (name: string) => () => console.warn(`method ${name} doesn't exist on android yet`)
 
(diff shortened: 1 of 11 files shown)


src/lib/NativeModules/ArtsyNativeModules.tsx
```diff
@@ -18,47 +19,12 @@ export const ArtsyNativeModules = Platform.select({
     ARCocoaConstantsModule: {
       UIApplicationOpenSettingsURLString: "UIApplicationOpenSettingsURLString",
       AREnabled: true,
-      CurrentLocale: "en_US",
-      LocalTimeZone: "",
+      CurrentLocale: getLocales()[0].languageTag,
+      LocalTimeZone: getTimeZone(),
```
pvinis, line 23: Are these two different on ios and android? 🤔

src/lib/NativeModules/ArtsyNativeModules.tsx
```diff
@@ -18,47 +19,12 @@ export const ArtsyNativeModules = Platform.select({
...
-          AROptionsNewFirstInquiry: false,
-          AROptionsUseReactNativeWebView: false,
-          AROptionsNewFairPage: false,
-          AROptionsNewInsightsPage: false,
-          AROptionsInquiryCheckout: false,
-        },
-        legacyFairSlugs: ["some-fairs-slug", "some-other-fair-slug"],
-        legacyFairProfileSlugs: [],
-      },
+      nativeState: null as any,
```
pvinis, line 27: Just making sure, I guess this stays null for now, and when we have some kind of way to do notifications on android, this will be added in an `android` key in this `platform.select`?

pvinis (approved):
goooood stuuff!

src/lib/NativeModules/ArtsyNativeModules.tsx
ds300: `CurrentLocale` on iOS uses an underscore while `react-native-localize` uses the standard hyphen. (e.g. 'en-US' vs 'en_US'). `LocalTimeZone` should be the same

src/lib/NativeModules/ArtsyNativeModules.tsx
ds300: i think this will stay null until we decide we need to store some state as a source of truth in java/kotlin. Hopefully we won't need to do that.

src/lib/relay/RelayCache.ts
```diff
@@ -0,0 +1,67 @@
...
+    return `{${Object.keys(val as object)
+      .sort()
+      .map((k) => `${JSON.stringify(k)}:${stableSerialize(val[k])}`)
+      .join(",")}}`
+  } else if (_.isArray(val)) {
+    return `[${val.map(stableSerialize).join(",")}]`
+  } else {
+    return JSON.stringify(val)
+  }
+}
```
MounirDhahri, line 17: ✨

MounirDhahri (commented):
👏👏👏

PR #4353, opened 2021-01-21
fix(Artist): Hide artist insights tab when the user has no auction results

MounirDhahri:

This PR resolves [CX-]

### Description
- Hide artist insights tab when the user has no auction results

**When the user has no auction results**
Screenshot 2021-01-21 at 14 59 33
Screenshot 2021-01-21 at 14 58 58
____
**When the user has auction results**
Screenshot 2021-01-21 at 14 58 47


src/lib/Components/HeaderTabGridPlaceholder.tsx
```diff
@@ -28,7 +25,6 @@ export const HeaderTabsGridPlaceholder: React.FC = () => {
           <PlaceholderText width={40} />
           <PlaceholderText width={50} />
           <PlaceholderText width={40} />
-          {showInsights ? <PlaceholderText width={40} /> : null}
```
MounirDhahri, line 31: This is no longer needed since we only have 2/3 tabs now in all cases

src/lib/Scenes/Artist/Artist.tsx
```diff
@@ -122,6 +123,13 @@ export const ArtistQueryRenderer: React.FC<ArtistQueryRendererProps> = ({ artist
               }
               ...ArtistHeader_artist
               ...ArtistArtworks_artist
+              auctionResultsConnection(first: 1) {
+                edges {
+                  node {
+                    internalID
+                  }
+                }
+              }
```
pepopowitz, line 132: I wouldn't hold up a merge for this, but it looks like there is a `totalCount` field on `auctionResultsConnection` that could be used to identify if the auction has results. I'd prefer that because it's a little more direct about what we're querying. (Associated changes would have to be made if you decide to go this route.)
 
```suggestion
              auctionResultsConnection {
                totalCount
              }
```

pepopowitz (commented):
I'm going to reassign to you to decide if you want to incorporate my suggestion. Aside from this, it looks great!

src/lib/Scenes/Artist/Artist.tsx
MounirDhahri: You're absolutely right. I will change it 🙏

PR #4356, opened 2021-01-22
(feat) connect market stats to real data

pepopowitz:
This PR connects the market stats module to real data. It currently shows only the stats for the first medium, but in the future we'll be able to pick which medium we want to see.

My Progress: 
- [x] MarketStats extracted to its own component & wrapped with a FragmentContainer
- [x] Relay hooked up to provide data
- [x] MarketStats emitting real data just to prove connectivity. 
- [x] See what it looks like when I waterfall/nest QueryRenderers, so that the priceInsights query isn't fetched until the artist returns with internalID. It looks fine because the nested QueryRenderer is on a different tab when it renders!
- [x] Fill in all the correct data.
- [x] Refactor the `formatText` util to use the new `formatLargeNumber` util introduced in this PR

For followup PRs: 
- [ ] Add the ability to select a medium from a `<Picker>`. 
- [ ] Add a loading skeleton
- [ ] Tests

This PR partially addresses [CX-738]

### Description

Connects real data to the market stats module. 

How it looks:

simulator_screenshot_491F4899-4F33-4E70-82D6-B057EC5CB760

[CX-738]: https://artsyproduct.atlassian.net/browse/CX-738


src/lib/Components/Artist/ArtistInsights/MarketStats.tsx
```diff
@@ -0,0 +1,66 @@
...
+        </Text>
+        <TouchableOpacity hitSlop={{ top: 20, right: 20, bottom: 20, left: 20 }}>
+          <Image source={require("@images/info.png")} />
+        </TouchableOpacity>
+      </Flex>
+      <Text variant="small" color="black60">
+        Based on the last 12 months of auction data
+      </Text>
+      {priceInsights.edges?.map((e) => {
+        return (
```
pepopowitz, line 26: This is just to prove connectivity.

pepopowitz:
I've got this PR rebased on master, and I intend to follow up with the last remaining tasks in future PRs. 

I've got a broken test for the artist page - the error is pretty cryptic, but I suspect it is due to the fact that I've added a nested QueryRenderer: 

```
FAIL  src/lib/Scenes/Artist/__tests__/Artist-tests.tsx
  ● availableTabs › returns two tabs if artist has metadata, works, and shows when AROptionsNewInsightsPage is true

    Failed: "TypeError: Cannot read property '_responseForQueryIDWithVariables' of undefined"

```

I'll take a look at this later today or Monday, but if anyone from @artsy/collector-experience has thoughts about that error, I'm all ears.

brainbicycle:
Hmmm the `yarn bundle-for-native-ci` command seems to not like the underscores in the numbers in `formatLargeNumber` wonder if this is typescript version delta? removing the underscores seems to fix it

src/lib/utils/formatLargeNumber.ts
```diff
@@ -0,0 +1,11 @@
+export function formatLargeNumber(number: number, decimalPlaces: number = 0) {
+  if (number < 1_000) {
```
brainbicycle, line 2: the `yarn bundle-for-native-ci` command on circle seems to not like the underscores in the numbers in this function wonder if this is typescript version delta? removing the underscores seems to fix it

src/lib/utils/__tests__/formatLargeNumberOfItems-tests.ts
```diff
@@ -0,0 +1,39 @@
...
+        `"850 works for sale"`
+      )
+    })
+    it("formats more than a thousand less than a million", () => {
+      expect(formatLargeNumberOfItems(151000, "work for sale", "works for sale")).toMatchInlineSnapshot(
+        `"151.0k works for sale"`
+      )
+    })
+    it("formats more than a million", () => {
+      expect(formatLargeNumberOfItems(2200000, "work for sale", "works for sale")).toMatchInlineSnapshot(
```
brainbicycle, line 34: Nice tests!

src/lib/utils/formatLargeNumber.ts
pepopowitz: ah, cool, I was surprised I didn't get a linting error anyway. I'll remove them!

brainbicycle (approved):
Looks great 📈 !

src/lib/utils/formatLargeNumber.ts
pepopowitz: though this code really shows the value of `_` separators in large numbers.

PR #4357, opened 2021-01-22
CX-938: Fix empty auction

The-Beez-Kneez:
The type of this PR is: **Bug Fix**

This PR resolves [CX-938]

### Description
This is PR to hide the `Clear Filters` button in the case of an empty auction, as there is nothing to filter.

What has changed:
Screen Shot 2021-01-22 at 1 14 45 PM

### TODO:
- [ ] Follow up with @samchieng for potential designs for empty auction copy

[CX-938]: https://artsyproduct.atlassian.net/browse/CX-938


src/lib/Scenes/Sale/Components/SaleLotsList.tsx
```diff
@@ -16,6 +17,7 @@ import { SaleArtworkListContainer } from "./SaleArtworkList"
 
 interface Props {
   saleArtworksConnection: SaleLotsList_saleArtworksConnection
+  unfilteredSaleArtworksConnection: SaleLotsList_unfilteredSaleArtworksConnection | null
```
damassi, line 20: The `| null` is unnecessary, I think. If data can be null that's defined in the Metaphysics schema, and Relay, which automatically generates these types,  [responds accordingly](https://github.com/artsy/eigen/pull/4357/files#diff-11dfe8f645f4657a00b1a4cf68341a0e57b86f382a5adb526c8d8ba795d70efdR10).

src/lib/Scenes/Sale/Components/SaleLotsList.tsx
```diff
@@ -51,7 +53,14 @@ export const SaleLotsListSortMode = ({
   )
 }
 
-export const SaleLotsList: React.FC<Props> = ({ saleArtworksConnection, relay, saleID, saleSlug, scrollToTop }) => {
+export const SaleLotsList: React.FC<Props> = ({
+  saleArtworksConnection,
+  unfilteredSaleArtworksConnection,
```
MounirDhahri, line 58: **issue:** I think you don't need to pass the `unfilteredSaleArtworksConnection` connection and instead what you can do is assigning `saleArtworksConnection ` to `unfilteredSaleArtworksConnection` at the initial render.

I suggest you try out the following
`const unfilteredSaleArtworksConnection = useRef<SaleLotsList_saleArtworksConnection>(saleArtworksConnection)`

src/lib/Scenes/Sale/Components/SaleLotsList.tsx
```diff
@@ -124,6 +133,10 @@ export const SaleLotsList: React.FC<Props> = ({ saleArtworksConnection, relay, s
     })
   }
 
+  if (unfilteredSaleArtworksConnection?.counts?.total === 0) {
```
MounirDhahri, line 136: This becomes then `unfilteredSaleArtworksConnection.current?.counts?.total`

src/lib/Scenes/Sale/Components/SaleLotsList.tsx
pepopowitz: I don't think this is true. The `saleArtworksConnection` will always include the filtered set of artworks, which means we can't tell if there are zero results because they filtered them down too far or there are truly zero artworks in the sale. `unfilteredSaleArtworksConnection` was added to identify the true count of artworks in the sale, regardless of filters applied by the user.

src/lib/Scenes/Sale/Components/SaleLotsList.tsx
The-Beez-Kneez: The null was added in because I was receiving a type check error, when attempting to push up. 

Was getting the following errors in my terminal:
Screen Shot 2021-01-25 at 9 02 18 AM

Podfile.lock
```diff
@@ -416,6 +416,8 @@ PODS:
     - Mapbox-iOS-SDK (~> 6.3)
     - React
     - React-Core
+  - react-native-navigator-ios (1.0.0):
```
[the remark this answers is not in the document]
The-Beez-Kneez: Removed them!

src/lib/Scenes/Sale/Components/SaleLotsList.tsx
damassi: These failures cascade from the top failure, which is from data stubbed for a test. In that case one can pass `{null as any}`

src/lib/Scenes/Sale/Components/SaleLotsList.tsx
MounirDhahri: maybe I am missing something, but I thought the first time the screen mounts,no filters are applied and therefore `saleArtworksConnection` is equal to `unfilteredSaleArtworksConnection`

Podfile.lock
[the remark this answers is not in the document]
MounirDhahri: Thanks 🙏

src/lib/Scenes/Sale/Components/SaleLotsList.tsx
The-Beez-Kneez: So passing in `{null as any}` from the top level test in SaleLotsList-test should resolve the error in Sale and Sale Lot List?

damassi: Ah, I was just assuming that would fix it. Not too sure! (But don't let this hold things up as yr transitioning off of CX, very minor.)

The-Beez-Kneez: Mmkay, thanks for the clarification.

src/lib/Scenes/Sale/Components/SaleLotsList.tsx
The-Beez-Kneez: @pepopowitz we determined this was not the case correct?

pepopowitz: Ah @MounirDhahri I missed the words "at the initial render". Yes, the first time this component renders, `unfilteredSaleArtworksConnection` would be the same as `saleArtworksConnection`. We could capture it then and store it; we could also rely on relay to tell us how many are in the full set. This feels like a difference in preference, and I don't see any objective benefits to one method over the other -- do you?

pepopowitz: Oh, and I noticed this PR doesn't include anything [in `Sale.tsx`to prevent the animated filter button from showing when there are no results](https://github.com/artsy/eigen/blob/f75ea7e82e27ec6d1490eb043b7c75d785f6119c/src/lib/Scenes/Sale/Sale.tsx#L226) -- @The-Beez-Kneez and I talked about using the `unfilteredArtworksConnection` count there, too. (That might be another reason not to centralize the logic in SaleLotsList.tsx.)

The-Beez-Kneez: Should that maybe be made into a JIRA ticket @pepopowitz so that it can be tracked? We had mentioned it would need to contain a copy message for an empty auction, and that was needing a follow up with @samchieng

MounirDhahri: I agree with @pepopowitz. Using Relay to get the number of lots in the full set looks good to me.

Podfile.lock
```diff
@@ -630,6 +630,7 @@ DEPENDENCIES:
   - react-native-config (from `node_modules/react-native-config`)
   - "react-native-geolocation (from `node_modules/@react-native-community/geolocation`)"
   - "react-native-mapbox-gl (from `node_modules/@react-native-mapbox-gl/maps`)"
+  - react-native-navigator-ios (from `node_modules/react-native-navigator-ios`)
```
brainbicycle, line 633: we should this remove this dependency here as well, sorry for churn on this and thanks for following up!

The-Beez-Kneez: Removed!

src/lib/Scenes/Sale/Components/SaleLotsList.tsx
pepopowitz: @The-Beez-Kneez good call, I'll add a subtask to that ticket for hiding the animated filter button!

brainbicycle (approved):

PR #4358, opened 2021-01-25
fix: auction results list QA

MounirDhahri:

This PR resolves [CX-969]

### Description
- Make **Not available** show in one line
- Fix issue with filter button not visible when there are less than 2 auction results available
Screenshot 2021-01-25 at 12 14 04
Screenshot 2021-01-25 at 12 13 41

[CX-969]: https://artsyproduct.atlassian.net/browse/CX-969


src/lib/Components/Artist/ArtistInsights/ArtistInsights.tsx
```diff
@@ -26,7 +26,7 @@ interface ViewToken {
   section?: any
 }
 
-const FILTER_BUTTON_OFFSET = 100
```
pepopowitz, line 29: Will this solve the problem for all devices, or is this fix dependent on the height of the device's screen?

MounirDhahri: Yes, this will solve the issue on all devices since the new offset is smaller than the `AuctionResultRow` height and therefore the user can reach it by scrolling contrarily to when it was (100)

pepopowitz (approved):
🎉

PR #4362, opened 2021-01-26
feat: Android nav

ds300:
This PR resolves [CX-873] [CX-872] [CX-874]

### Description

This PR introduces navigation stacks for Android (and eventually iOS) using `react-navigation` + `react-native-screens`. This means we are using platform native navigation primitives under the hood, for speed 🚗 

This PR also puts Artsy's iOS-only native modules (e.g. `ARTemporaryAPIModule`) behind a `LegacyNativeModules` object to emphasise that they are old, and many of them are ripe for removal as we get Android up and running.

https://user-images.githubusercontent.com/1242537/106463137-ea104300-648e-11eb-88fd-bea55435da59.mp4

https://user-images.githubusercontent.com/1242537/106463148-f1375100-648e-11eb-8120-25e04a27c801.mp4

### Follow up work

- add tests for nav infra (deferring this to keep velocity high and avoid testing things which are liable to completely change when we adopt react-navigation on iOS too) https://artsyproduct.atlassian.net/browse/CX-982
- add scroll-to-top support on bottom tab press https://artsyproduct.atlassian.net/browse/CX-983

[CX-873]: https://artsyproduct.atlassian.net/browse/CX-873
[CX-872]: https://artsyproduct.atlassian.net/browse/CX-872
[CX-874]: https://artsyproduct.atlassian.net/browse/CX-874


package.json
```diff
@@ -84,7 +84,9 @@
     "@react-native-community/masked-view": "^0.1.10",
     "@react-native-community/netinfo": "4.6.1",
     "@react-native-community/picker": "1.3.0",
+    "@react-native-community/viewpager": "^4.2.2",
```
ds300, line 87: this was a transitive dependency of `react-native-scrollable-tab-view` that we had to make explicit, otherwise the android autolinking doesn't work.

src/lib/Components/ArtworkGrids/ArtworkGridItem.tsx
```diff
@@ -85,19 +85,25 @@ export const Artwork: React.FC<ArtworkProps> = ({
...
+                borderRadius={2}
+                alignSelf="flex-start"
+              >
                 <Sans size="2" color="black100" numberOfLines={1}>
                   {urgencyTag}
                 </Sans>
               </Flex>
             )}
-          </OpaqueImageView>
+          </View>
```
ds300, line 106: Lots of screens were causing the app to crash because a few commonly-used components were putting children inside an `OpaqueImageView`. On android `OpaqueImageView` currently uses the normal react-native `Image` component under the hood, which doesn't support children like our iOS native image component, so we decided to just refactor the few places where this was being done.

src/lib/Scenes/BottomTabs/BottomTabsNavigator.tsx
```diff
@@ -40,37 +51,40 @@ export const BottomTabsNavigator = () => {
...
-          )
+          <View
+            key={index}
+            pointerEvents={index === activeIndex ? undefined : "none"}
+            style={{ position: "absolute", width: "100%", height: "100%" }}
+          >
+            <Animated.View style={{ opacity: opacities[index], flex: 1, backgroundColor: "white" }}>
+              {!!(hasLoaded[index] || index === activeIndex) && v}
+            </Animated.View>
+          </View>
```
ds300, line 87: We refactored this component because there seemed to be an `Animated` bug on android during fade transitions. When animating the next view from opacity 0 to 1 the first frame would have opacity 1 already for some reason. e.g. if there were 10 frames in the animation, the opacity at each frame should be like `[0, 0.1, 0.2, ..., 1]` but we were seeing `[1, 0.1, 0.2, ..., 1]`.

Most confusing.

Anyway we made the problem disappear by moving things around.

src/lib/navigation/BackButton.tsx
```diff
@@ -0,0 +1,27 @@
...
+        width: 40,
+        height: 40,
+        borderRadius: 20,
+        alignItems: "center",
+        justifyContent: "center",
+      }}
+    >
+      <TouchableOpacity onPress={onPress}>
+        <ChevronIcon direction="left" />
+      </TouchableOpacity>
```
ds300, line 24: 💡  At first this was using our custom `Touchable` but when debugging on android the onPress event was not being fired, no clue why. Most bizarre.

patches/@react-navigation+core+5.14.0.patch
```diff
@@ -0,0 +1,41 @@
...
+     const { addOptionsGetter, getCurrentOptions } = useOptionsGetters({});
+@@ -202,7 +212,7 @@ const BaseNavigationContainer = React.forwardRef(
+       dispatch,
+       canGoBack,
+       getRootState,
+-      dangerouslyGetState: () => state,
++      dangerouslyGetState: getRootState,
+       dangerouslyGetParent: () => undefined,
+       getCurrentRoute,
+       getCurrentOptions,
```
damassi, line 41: Can we add these changes to [hacks.md](https://github.com/artsy/eigen/blob/master/HACKS.md)?

android/app/src/main/java/net/artsy/app/ArtsyNativeModule.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.HashMap;
+
+public class ArtsyNativeModule extends ReactContextBaseJavaModule {
+    // this is called on application launch by MainApplication#onCreate
+    private static String LAUNCH_COUNT = "launchCount";
+    public static void didLaunch(SharedPreferences prefs) {
+        launchCount = prefs.getInt(LAUNCH_COUNT, 0) + 1;
+        prefs.edit().putInt(LAUNCH_COUNT, launchCount).commit();
+        System.out.println("LaunchCount:" + launchCount);
+    }
```
ds300, line 22: We need the `launchCount` at dev time for maintaining navigation state across reloads.

We want the behaviour to be:

- If you relaunch the app, the navigation state should reset
- If you reload the js bundle, the navigation state should kept

We couldn't find a built-in way to tell the difference between relaunch and reload, so we added this `launchCount` property that is incremented only when the app launches. We already had it available on iOS.

src/lib/Components/FilterModal/FilterModal.tsx
```diff
@@ -176,7 +176,7 @@ export const FilterModalNavigator: React.FC<FilterModalProps> = (props) => {
     state.selectedFilters.length > 0 || (state.previouslyAppliedFilters.length === 0 && state.appliedFilters.length > 0)
 
   return (
-    <NavigationContainer>
+    <NavigationContainer independent>
```
ds300, line 179: We now have a `NavigationContainer` wrapping the main part of the app (after onboarding). So any places where we use react-navigation in modals, e.g. here, we need to set this `independent` prop to `true`, which just silences a `console.error` that is intended for people who are using react-navigation in the normal way.

src/lib/utils/AdminMenuWrapper.tsx
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        state.lastTapTimestamp = now
+
+        if (state.numTaps >= 5) {
+          state.numTaps = 0
+          setIsShowingAdminMenu(true)
+        }
+        return false
+      }}
```
ds300, line 31: I added a placeholder admin menu already, to allow us to clear `AsyncStorage` easily during development. For now I just made it trigger on five taps anywhere, with less than 0.5 seconds between each tap. We can iterate on this.

ds300:
hmm QAing this and it seems to have affected touch handling on the `FilterModal` on iOS, I am investigating, don't merge yet.

ds300:
OK @pvinis this is good to go

typings/emission.d.ts
```diff
@@ -1,67 +0,0 @@
-import type { ViewDescriptor } from "lib/navigation/navigate"
```
MounirDhahri, line 1: 👋

src/lib/Components/FilterModal/FilterModal.tsx
```diff
@@ -180,6 +180,8 @@ export const FilterModalNavigator: React.FC<FilterModalProps> = (props) => {
       <FancyModal visible={props.isFilterArtworksModalVisible} onBackgroundPressed={handleClosingModal} maxHeight={550}>
         <View style={{ flex: 1 }}>
           <Stack.Navigator
+            // force it to not use react-native-screens, which is broken inside a react-native Modal for some reason
+            detachInactiveScreens={false}
```
MounirDhahri, line 187: this is an interesting prop!

MounirDhahri (approved):

MounirDhahri:
🔥🔥🔥

android/app/src/main/java/net/artsy/app/ArtsyNativeModule.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.Map;
+import java.util.HashMap;
+
+public class ArtsyNativeModule extends ReactContextBaseJavaModule {
+    // this is called on application launch by MainApplication#onCreate
+    private static String LAUNCH_COUNT = "launchCount";
+    public static void didLaunch(SharedPreferences prefs) {
+        launchCount = prefs.getInt(LAUNCH_COUNT, 0) + 1;
+        prefs.edit().putInt(LAUNCH_COUNT, launchCount).commit();
+        System.out.println("LaunchCount:" + launchCount);
```
pvinis, line 21:
```suggestion
```

android/app/src/main/java/net/artsy/app/ArtsyNativeModule.java
pvinis: Smort! Nice one.

src/lib/NativeModules/ARScreenPresenterModule.tsx
```diff
@@ -0,0 +1,130 @@
+import { NavigationContainerRef, StackActions } from "@react-navigation/native"
+import { ViewDescriptor } from "lib/navigation/navigate"
+import { BottomTabType } from "lib/Scenes/BottomTabs/BottomTabType"
+import { last } from "lodash"
+import { NativeModules } from "react-native"
+
+/**
+ * Here we maintain references to all the navigators in the main app navigation hierarchy, which are:
```
pvinis, line 8: 👌

src/lib/navigation/BackButton.tsx
pvinis: Hm, we could take a look at a KS.

src/lib/navigation/ModalStack.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { useReloadedDevNavigationState } from "./useReloadedDevNavigationState"
+
+const Stack = createNativeStackNavigator()
+
+/**
+ * ModalStack is the root navigation stack in the app. The root screen in this stack is
+ * the main app (with bottom tabs, etc), and then whenever we present a modal it gets
+ * pushed on the top of this stack. We use react-native-screens to get native modal presetation
+ * transitions etc.
+ */
```
pvinis, line 16: These comments are very nice and helpful! I'm sure they will help make everything clear to every dev that comes in here.

src/lib/utils/AdminMenuWrapper.tsx
```diff
@@ -0,0 +1,103 @@
...
+import { MenuItem } from "lib/Components/MenuItem"
+import { CloseIcon, Flex, Text } from "palette"
+import React, { useRef, useState } from "react"
+import { DevSettings, ScrollView, TouchableOpacity, View } from "react-native"
+import { useScreenDimensions } from "./useScreenDimensions"
+
+export const AdminMenuWrapper: React.FC = ({ children }) => {
+  const [isShowingAdminMenu, setIsShowingAdminMenu] = useState(false)
+  const gestureState = useRef({ lastTapTimestamp: 0, numTaps: 0 })
+  // TODO: if public release, just return <>{children}</>
```
pvinis, line 11: I would say we keep it in release too, still useful as it is now.

src/lib/utils/AdminMenuWrapper.tsx
pvinis: It might be good to have like a three finger hold or something, but yea, that's for the admin jira ticket 👍.

pvinis (approved):
Very very cool work from both of you!

src/lib/utils/AdminMenuWrapper.tsx
ds300: only problem with a three finger hold is, how to do that on a simulator?

pvinis: Hm.. true.. ok we will figure it out in the other ticket. I only don't like the spam tapping because it's easy to keep tapping and then tap something on the admin menu 😬.

PR #4368, opened 2021-01-27
fix: explicitly tag releases to fix sourcemaps for sentry errors

brainbicycle:

This PR resolves [CX-956]

### Description

We are seeing an issue with sentry errors javascript code not being deobfuscated. The problem seems to be that the releaseName we are specifying for sentry artifacts (including sourcemaps) when uploading betas does not match the releaseName we are sending with sentry errors. 

This PR attempts to resolve this by:
- Explicitly passing a release name in our sentry init including the buildVersion
- Sending a release name of same format when uploading beta + sentry artifacts
- Adding a dev option to trigger an error in sentry for testing
- Reenabling sentry releases + sourcemapping for beta builds
    - We disabled this at some point because we were getting duplicate file upload errors, I believe the cause of this is that releases are org wide, so attempting to resolved this by adding a postfix "-dev" to releases for beta builds

Co-authored-by @artsy/collector-experience during Knowledge Share!

### Note 

I haven't tested this much locally aside from logging out expected release names and stepping through code with debugger because most of this sentry artifact uploading happens as part of our beta deploy and seems like the easiest way to test would be to get this into a beta and trigger an error via the dev option and then see if the js code is being deobfuscated correctly. I will follow up and or revert if anything doesn't work or is broken. 

[CX-956]: https://artsyproduct.atlassian.net/browse/CX-956


brainbicycle:
Reviewed in knowledge share, going to merge and test in beta! Will follow up with any changes necessary!

MounirDhahri:
Sorry I forgot to approve the changes here 🙇‍♂️

PR #4387, opened 2021-02-03
fix: Conversation layout and pagination issues

jo-rs:

### Description

This PR fixes an issue with Conversations on Inbox that prevented pagination to work correctly, and that blocked the bottom part of the container.
scroll-fix

@artsy/purchase-devs 

### DIFF
--- a/src/lib/Scenes/Inbox/Components/Conversations/Conversations.tsx
+++ b/src/lib/Scenes/Inbox/Components/Conversations/Conversations.tsx
@@ -1,7 +1,7 @@
 import React from "react"
 import { createPaginationContainer, graphql, RelayPaginationProp } from "react-relay"
 
-import { ActivityIndicator, FlatList, RefreshControl, View } from "react-native"
+import { ActivityIndicator, FlatList, RefreshControl } from "react-native"
 
 import { navigate } from "lib/navigation/navigate"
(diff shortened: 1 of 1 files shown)


erikdstock (approved):

src/lib/Scenes/Inbox/Components/Conversations/Conversations.tsx
```diff
@@ -93,7 +93,7 @@ export const Conversations: React.FC<Props> = (props) => {
   const shouldDisplayMyBids = getCurrentEmissionState().options.AROptionsBidManagement
 
   return (
-    <View style={{ flexGrow: 1 }}>
+    <>
```
erikdstock, line 96: I wondered if there was a chance changing the view to a fragment would fix our other refreshing issue... strange though. Do you have thoughts on what was going on here?

jo-rs: The `grow` prop was redundant, as we have one for the `FlatList`'s content, and was causing some strange behaviour resulting in the container rendering behind the nav which, with the inset on the parent, would have been ok but was now causing some issues. After removing the `grow`, there was no need for the `View` really, or at least everything seems to be working fine with the fragment.

PR #4390, opened 2021-02-03
fix: use api key for beta deploy #trivial

brainbicycle:

### Description

Apple has started enforcing 2fa on all accounts in App Store connect, and this broke our beta deploys. This PR uses an App Store connect api key rather than user password login. 

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -22,7 +22,15 @@ app_json = JSON.parse(File.read(app_json_path))
 app_plist = "../Artsy/App_Resources/Artsy-Info.plist"
 sticker_plist = "../Artsy Stickers/Info.plist"
 
+
 lane :ship_beta_ios do
+  api_key = app_store_connect_api_key(
+    key_id: "B95P9K5S5V",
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -22,7 +22,15 @@ app_json = JSON.parse(File.read(app_json_path))
 app_plist = "../Artsy/App_Resources/Artsy-Info.plist"
 sticker_plist = "../Artsy Stickers/Info.plist"
 
+
 lane :ship_beta_ios do
+  api_key = app_store_connect_api_key(
+    key_id: "B95P9K5S5V",
+    issuer_id: "69a6de72-31ba-47e3-e053-5b8c7c11a4d1",
+    key_content: ENV['ARTSY_APP_STORE_CONNECT_API_KEY_CONTENT'],
```
iskounen, line 30: I'm not familiar with this setup so I just want to make sure that `key_content` is the only sensitive piece of information, and that it's safe to expose the `key_id` and `issuer_id`.

brainbicycle: These aren't sensitive values but it would probably be nicer to have them as environment vars as well,  I can update

PR #4407, opened 2021-02-05
fix: year created design updates #trivial

MounirDhahri:

This PR resolves [CX-999]

### Description
**Design adjustment for year created**
- Change hyphen between years to an "en-dash"
- Make toggle switches border radius black instead of grey
- Add haptic feedback on value change for slider ✨ (the only way to test this is on real device)
- Update double divider to single divider (1px) between the slider and the "Include lots without year created info"
- Decrease left-padding to be 20px when the slider is active

https://user-images.githubusercontent.com/11945712/107047596-472a3280-67c8-11eb-98b2-a817ac2f126d.mov

[CX-999]: https://artsyproduct.atlassian.net/browse/CX-999

### DIFF
--- a/src/lib/Components/CircleWithBorder/CircleWithBorder.tsx
+++ b/src/lib/Components/CircleWithBorder/CircleWithBorder.tsx
@@ -1,0 +1,38 @@
+import { Flex, FlexProps } from "palette"
+import React from "react"
+
+interface CircleWithBorderProps extends FlexProps {
+  backgroundColor: string
+  borderColor: string
+  borderWidth: number
+  diameter: number
+}
+
+//  It seems like React Native has a bug on iOS that causes some of the background to bleed
+//  so using a circle within a circle as a workaround
+export const CircleWithBorder: React.FC<CircleWithBorderProps> = ({
+  backgroundColor,
+  diameter,
+  borderColor,
+  borderWidth,
+  ...wrapperProps
+}) => (
+  <Flex flex={1} justifyContent="center" alignItems="center" backgroundColor={borderColor} {...wrapperProps}>
+    <Flex
+      justifyContent="center"
+      alignItems="center"
+      borderRadius={diameter / 2}
+      width={diameter}
+      height={diameter}
+      backgroundColor={borderColor}
+    >
+      <Flex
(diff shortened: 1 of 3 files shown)


brainbicycle (commented):
This looks good but I think the bug with the border was that the background color black was showing slightly outside the white border, as far as I know white border should still be present we just want to get rid of the outline if we can:

black-border


I think this might be a react bug: 
https://stackoverflow.com/questions/35030758/react-native-border-radius-with-background-color

I looked into a bit and couldn't get it to go away when using border-width and border-radius so I was thinking of working around by doing two concentric circles instead like color-swatches here:
https://github.com/artsy/eigen/blob/a1712279c0848c37b069f9e149c19974e16827b4/src/lib/Components/ArtworkFilterOptions/ColorSwatch.tsx#L36

samchieng (approved):
looks great!

src/lib/Components/CircleWithBorder/CircleWithBorder.tsx
```diff
@@ -0,0 +1,38 @@
...
+interface CircleWithBorderProps extends FlexProps {
+  backgroundColor: string
+  borderColor: string
+  borderWidth: number
+  diameter: number
+}
+
+//  It seems like React Native has a bug on iOS that causes some of the background to bleed
+//  so using a circle within a circle as a workaround
+export const CircleWithBorder: React.FC<CircleWithBorderProps> = ({
```
MounirDhahri, line 13: Now that you mentioned it, I remember we had the same issue with the background bleeding few times before. Therefore I created this component to avoid recreating it in the future. I wanted to refactor the `ColorSwatch` to use it as well, but it seemed like it doesn't make sense since it has some logic that is different

MounirDhahri: here is how it looks like now
Screenshot 2021-02-09 at 12 27 37

brainbicycle: 🤩 Looks great! Thanks for doing that!

samchieng: love it!

PR #4414, opened 2021-02-08
Adding direnv rc file

pvinis:

This PR resolves [CX-]

### Description

Adding [direnv](https://direnv.net/#quick-demo) which you can see the use in the link.

I added it locally because I keep getting a flood of warnings with the fastlane lanes like:
Screenshot 2021-02-08 at 11 29 43

and now with direnv:
Screenshot 2021-02-08 at 11 33 35

bundler so clean ✨

If we don't want this in the repo, I'm totally fine keeping it local-only. If others have this flood of warnings in bundler from gem deps, maybe try this.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+export RUBYOPT="-W0"


brainbicycle (commented):
Hmm, this will silence all ruby warnings which I am not sure we want although I agree if they are getting out of hand we should do something. Oddly I am not seeing all the same warnings although I am seeing the json last parameter warnings.

pvinis:
Hm, ok. I will keep this local for now then. Maybe it's my asdf setup that's give the extra ones.

PR #4420, opened 2021-02-08
fix: Dedupe currencies on auction result (CX-973)

pepopowitz:

This PR resolves [CX-973]

### Description

Currencies are being duplicated for lots in currencies other than USD, GBP, EUR, and MYR. This PR removes the explicit currency as a suffix, allowing the `display` property to handle rendering the currency. 

This applies to both the Sale Price & the Estimate Range.

#### Why were currencies only *sometimes* duplicated?

The `display` property for the prices are conditionally including the currency as a prefix. The decision to show the prefix is made here: https://github.com/artsy/metaphysics/blob/master/src/schema/v2/auction_result.ts#L201

The list of currencies which suppress the prefix is here: https://github.com/artsy/metaphysics/blob/master/src/schema/v2/auction_result.ts#L21

#### What's the tradeoff?

By removing the suffix, we're assuming that the symbol alone is adequate for USD, GBP, EUR, and MYR currencies. For the first three this seems reasonable; the 4th (MYR which renders as RM) I'm unsure of.

### What it looks like after these changes

image

### What it looked like before

image

[CX-973]: https://artsyproduct.atlassian.net/browse/CX-973


samchieng (approved):
looks good!

MounirDhahri (approved):

MounirDhahri:
great, thanks!

PR #4425, opened 2021-02-09
[RFC] enchancement(styled-system): make all uses of `ml` and other space friends consistent and predictable

pvinis:

This PR resolves [CX-]

## Marked as draft because it needs some work before merging.

`ml` and friends are inconsistent in our code.

- How big is the left margin of `<Box width={10} height={10} ml={3} />`?
- How big is the left margin of `<Box width={10} height={10} ml="3" />`?
- How big is the left margin of `<Box width={10} height={10} ml="3px" />`?
- How big is the left margin of `<Box width={10} height={10} ml={7} />`?
- How big is the left margin of `<Box width={10} height={10} ml="7" />`?
- How big is the left margin of `<Box width={10} height={10} ml="7px" />`?

<details>
  <summary>Guess first, look after.</summary>  

  - 30 pixels
  - 30 pixels
  - 3 pixels
  - 7 pixels
  - 7 pixels
  - 7 pixels
</details>

After guessing and seeing, we see that `"Xpx"` is consistently pixels. But `{X}` and `"X"` give mixed results. Since our design system is using numbers as SpaceUnit keys, if `X` exists, `ml` will take the value of that, if it doesn't exist, it will be pixels. That's why `{3}` gives 30 pixels (`3` exists in our design system) and `{7}` gives 7 pixels (`7` doesn't exist in our design system).

That's how we ended up with this funkiness:
Screenshot 2021-02-08 at 12 59 03
There was a `<Spacer ml={5} />` expecting a 5 pixel gap, but we got 50. Yes, `0.5` is in our system and that could be used, but that doesn't stop cases where we want 15 pixels or specifically 5 pixels (independent from our design system) and not "our small design system space".

My wish is for devs to differentiate between a number being a number of pixels and a string being a key of our design system. `{3}` is 3 pixels. `"3"` is 30 pixels, coming from `3: 30` mapping in palette.

I have tried to fix this by patching `styled-system` with a very tiny change. Before anything else, it checks if we are passing a number. If yes, then it's always pixels. `{3}` is 3 pixels. From there it continues as normal, and by itself, it handles `"3px"` and `"3"`.

Additionally, I have made some improvements to the types, so now even in VSCode we can get nice autocompletion. This way we (1) know that it's `number` for pixels or `"1" | "2" | ..` for design system, and (2) we know which keys are in the design system. Ever wondered if 8 is in there? How about 10?

Exhibit A:
Screenshot 2021-02-09 at 17 57 45
Exhibit B:
Screenshot 2021-02-09 at 17 57 57

### Before merging
- [x] Go through the app and look for things like `ml={0.5}` that probably need to become `ml="0.5"`?


HACKS.md
```diff
@@ -82,3 +82,10 @@
 
   - Stop using nested navigation containers.
   - Fix `@react-navigation/core` properly upstream.
+
+- @styled-system/space patch-package
+
+  Never?
+
+  We have this patch, so that we can use `ml={1}` for 1 pixel, `ml="1"` for our design system `1`, which is 10 pixels.
+  We have the tests in `Box.tests.tsx` to make sure we keep it the way we want it.
```
damassi, line 91: Rather than patching styled-system, can we just use our own `space` function mixin?

pvinis: I did that at first but thought a patch is smaller and easier to understand the change than recreating that file 🤷‍♂️. I'd be ok with either way.

damassi: Def not a blocker, just wondering yr thoughts 👍

ds300:
❤️ this has always mildly bugged me about palette, so 👍🏽 💯 

I'm also not crazy about the idea of making this kind of fundamental change via a patch file. A custom `space` fn would definitely be preferable imo, even if it's more code, for the sake of maintainability + portability.

pvinis:
Alright, you had me at maintainablility. I will do this (I'm cleaning up PRs today) and then I'll try to do a quick sweep of the app and see if we have funky spacings that need fixing after this change. If it's few I'll fix them. If it's a lot I'll schedule a time or maybe during KS for the team to do it together.

brainbicycle:
Looks good to me! Agree with David that if it is not a huge amount of extra effort would be nice to have a custom space function instead of a patch. Thanks for doing this! 👍

pvinis:
This is ready. I'll merge and announce once tests are green!

MounirDhahri:
🙌

PR #4427, opened 2021-02-09
fix: update support email, fix optional bridge param issue

brainbicycle:

This PR resolves [CX-680]

### Description

Updates old support email to new support email. There seems to be an issue with passing optional params over the bridge that was causing an error when running on the simulator. This PR works around this by instead creating two email bridge methods, one with a body and one without and choosing which one to call via a helper function on the js side.

[CX-680]: https://artsyproduct.atlassian.net/browse/CX-680


pvinis:
what do you think about using https://github.com/artsy/eigen/blob/master/src/lib/utils/sendEmail.ts and completely getting rid of the native code? 🤔

brainbicycle:
> what do you think about using https://github.com/artsy/eigen/blob/master/src/lib/utils/sendEmail.ts and completely getting rid of the native code? 🤔

We talked async about this but the UX is slightly different for these two functions. The native function shows up as a compose email modal and cancelling takes you back to the Artsy app, the typescript function opens the users mail app. The former seems like slightly better UX to me but would be good to unify what we do. @samchieng do you have thoughts?


https://user-images.githubusercontent.com/49686530/107545036-d2287580-6b98-11eb-903a-90494558c9bf.mp4


https://user-images.githubusercontent.com/49686530/107545045-d5236600-6b98-11eb-800f-99498450a2f2.mp4

brainbicycle:
Realized after I wrote this Sam is out for week, if it is cool with you @pvinis  could we leave as is for now and discuss when she gets back?

pvinis:
totally

samchieng:
@brainbicycle @pvinis definitely prefer the former suggestion to use the native mail function!

PR #4429, opened 2021-02-11
enhancement(analytics): use cohesion

pvinis:

This PR resolves [CX-]

### Description

Small PR to use cohesion for tracking on share events.

### DIFF
--- a/src/lib/Scenes/Artwork/Components/ArtworkHeader.tsx
+++ b/src/lib/Scenes/Artwork/Components/ArtworkHeader.tsx
@@ -63,7 +63,7 @@ export const ArtworkHeader: React.FC<ArtworkHeaderProps> = (props) => {
         urls: [base64Data, details.url],
         message: details.message,
       })
-      trackEvent(share(tracks.iosShare(res.app, artwork.id, artwork.slug)))
+      trackEvent(share(tracks.iosShare(res.app, artwork.internalID, artwork.slug)))
     } catch (err) {
       console.log({ err })
     } finally {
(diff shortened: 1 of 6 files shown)


src/lib/Scenes/Artwork/Components/ArtworkHeader.tsx
```diff
@@ -97,11 +95,13 @@ export const ArtworkHeader: React.FC<ArtworkHeaderProps> = (props) => {
       backgroundBottomColor: "#ffffff",
       backgroundImage: base64Data,
     })
+    trackEvent(share(tracks.customShare(CustomService.instagram_stories, artwork.id, artwork.slug)))
```
MounirDhahri, line 98: I think this needs to be the `internalID` instead of the `id`. 
Few days ago I implemented something similar using the id and mike didn't recognize them. (here https://artsyproduct.atlassian.net/browse/CX-1045). Only then I figured out we are actually using the internalID (here https://github.com/artsy/eigen/blob/2621bb317871dabfe6b48638468c71553a42f4d3/src/lib/Scenes/Artwork/Components/CommercialInformation.tsx#L123)

brainbicycle (approved):
Looks good to me! Mounir's call out is a good one but other than that merge when ready!

src/lib/Scenes/Artwork/Components/ArtworkHeader.tsx
pvinis: gooood catch. doing it.

pvinis:
#squashongreen

PR #4434, opened 2021-02-11
feat(fx-2641): converts medium => additionalGeneIDs filter

dzucconi:

This PR resolves [FX-2641]

### Description

So this... works. It just feels obviously incorrect. I want to be able to dispatch a filter select without having to shovel everything into a single value; rather support multiple selected filters on the same param. Right now, if I wanted to merge this I'd have to write a humanization function to turn the selected values into readable labels (currently they are the values; which are slugs). This is, of course, insane. Send help.

[FX-2641]: https://artsyproduct.atlassian.net/browse/FX-2641

### DIFF
--- a/src/lib/Components/FilterModal/FilterModal.tsx
+++ b/src/lib/Components/FilterModal/FilterModal.tsx
@@ -24,6 +24,7 @@ import { FlatList, TouchableOpacity, View, ViewProperties } from "react-native"
 import { useTracking } from "react-tracking"
 import styled from "styled-components/native"
 import { AnimatedBottomButton } from "../AnimatedBottomButton"
+import { AdditionalGeneIDsOptionsScreen } from "../ArtworkFilterOptions/AdditionalGeneIDsOptions"
 import { ArtistIDsOptionsScreen } from "../ArtworkFilterOptions/ArtistIDsOptionsScreen"
 import { AttributionClassOptionsScreen } from "../ArtworkFilterOptions/AttributionClassOptions"
 import { CategoriesOptionsScreen } from "../ArtworkFilterOptions/CategoriesOptions"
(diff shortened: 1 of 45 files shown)


src/lib/Components/FilterModal/FilterModal.tsx
```diff
@@ -46,20 +47,21 @@ export type FilterScreen =
   | "artistIDs"
   | "artistsIFollow"
   | "attributionClass"
-  | "color"
   | "categories"
+  | "color"
   | "dimensionRange"
   | "estimateRange"
   | "gallery"
+  | "geneIDs"
```
mzikherman, line 55: FYI we actually had to change this, as `geneIDs` param can only be used by gravity in an AND style, whereas I believe the goal here is OR. For the OR style list, its `additionalGeneIDs`: https://github.com/artsy/metaphysics/blob/486eddce74c6ece92e2b4e7ba64d7b1741d8dd58/src/schema/v2/filterArtworksConnection.ts#L102

dzucconi:
Updated geneIDs => additionalGeneIDs. Will have to just update all the `medium` arguments call sites if this looks fine.

src/lib/Components/FilteredInfiniteScrollGrid/FilteredInfiniteScrollGrid.tsx
```diff
@@ -1,3 +1,4 @@
+// TODO
```
dzucconi, line 1: OOOOK so there's two instances where I have no idea whether or not the `additionalGeneIDs` argument is present on the connection. Not sure what this component is; but it doesn't look like it's used anywhere? Safe to delete?

src/lib/Containers/Gene.tsx
```diff
@@ -1,3 +1,4 @@
+// TODO
```
dzucconi, line 1: OK, I *think* this one is the old filter and can be ignored

src/lib/utils/ArtworkFilter/FilterArtworksHelpers.ts
```diff
@@ -170,6 +172,19 @@ export const changedFiltersParams = (currentFilterParams: FilterParams, selected
   return changedFilters
 }
 
+/**
+ * NOTE: I am not happy this exists — but right now we can only dispatch arrays of paramValues
```
dzucconi, line 176: 😭

dblandin: 😢 but glad you put an inline comment in 👍

dblandin (approved):
Sorry that this was such a slog! Looks good to me 👍

PR #4438, opened 2021-02-12
fix: simplify watched lots refreshing

erikdstock:
The type of this PR is: bugfix

cc @artsy/purchase-devs 

This PR resolves [PURCHASE-2431]

### Description
This PR resolves an issue where the user's watched lots would not update after re-visiting the inbox. We believe It was caused by having two connections in a pagination container - the pagination container's refetch functionality could only keep track of one and so the newer watchedLotConnection was ignored for updates. This PR fixes it by downgrading it to a refetch container. I investigated other fixes with @damassi and we agreed this is fine for our use case until other limitations in these queries unblock querying a user's bid history.

[PURCHASE-2431]: https://artsyproduct.atlassian.net/browse/PURCHASE-2431


damassi (approved):
Nice and simple fix 👍

jo-rs (approved):
Looks great! 💯

PR #4444, opened 2021-02-16
Revert "chore: update artist insights flag"

MounirDhahri:
Reverts artsy/eigen#4443


ds300:
I don't think we need to revert this now that https://github.com/artsy/echo/pull/74 keeps the old echo flag. shall we close this PR?

PR #4452, opened 2021-02-17
feat: migrate LoadFailureView to typescript

MounirDhahri:

This PR resolves [CX-1055]

### Description
- Migrate LoadFailureView to typescript
- Remove old native `LoadFailureView`

[CX-1055]: https://artsyproduct.atlassian.net/browse/CX-1055


Artsy.xcodeproj/project.pbxproj
```diff
@@ -3325,7 +3325,6 @@
 				"${PODS_ROOT}/Artsy+UIFonts/Pod/Assets/Unica77LL-Regular.otf",
 				"${PODS_ROOT}/Artsy+UILabels/Pod/Assets/Chevron_Black.png",
 				"${PODS_ROOT}/Artsy+UILabels/Pod/Assets/Chevron_Black@2x.png",
-				"${PODS_ROOT}/Extraction/Extraction/Assets/ARLoadFailureRetryIcon@2x.png",
```
pvinis, line 3328: Even though I love removing these, I see now that this comes back once we run `bundle e pod install`. This is normal, since these come from https://github.com/artsy/extraction. I see that we need this pod for other things like the onboarding for now. I wouldn't bother making a change there and making a new pod version etc, since this pod dependency will go away soon with the android rework.

I would suggest we revert the changes in this file (or just run `bundle e pod install`), so that they don't unexpectedly come back in the next PR that needs to pod install. I guess one extra image in the bundle is not too bad for now, until we ditch the pod altogether.

Gemfile.lock
```diff
@@ -311,6 +311,7 @@ GEM
       xcpretty (~> 0.2, >= 0.0.7)
 
 PLATFORMS
+  ruby
```
pvinis, line 314: Not sure where this came from. I don't think it's a problem, but I also don't see it added when I run stuff like `bundle install` 🤔.

pvinis (changes requested):
Just a change for the xcodeproj file that I think is important, otherwise this seems to work well 👍.

Gemfile.lock
MounirDhahri: yes, this came using bundle install

MounirDhahri: pardon -- I meant bundle exec pod install

Artsy.xcodeproj/project.pbxproj
MounirDhahri: thanks for noticing this. I will update it now

Gemfile.lock
pvinis: 🤷 All good then. Thanks.

pvinis (approved):

Gemfile.lock
MounirDhahri: you're welcome

MounirDhahri:
Please hold on merging this until we release

MounirDhahri:
@pvinis  since we submitted the app for release already, I will merge this now

PR #4454, opened 2021-02-17
fix(android): send email

admbtlr:

This PR resolves [CX-1062]

### Description

Open default mail client in Android when the user taps on _Send Feedback_.

[CX-1062]: https://artsyproduct.atlassian.net/browse/CX-1062

### DIFF
--- a/src/lib/NativeModules/presentEmailComposer.ts
+++ b/src/lib/NativeModules/presentEmailComposer.ts
@@ -1,9 +1,10 @@
+import { stringify } from "qs"
 import { Linking, Platform } from "react-native"
 import { LegacyNativeModules } from "./LegacyNativeModules"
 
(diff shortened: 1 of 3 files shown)


src/lib/NativeModules/presentEmailComposer.ts
```diff
@@ -1,7 +1,10 @@
+import { Linking, Platform } from "react-native"
 import { LegacyNativeModules } from "./LegacyNativeModules"
 
 export function presentEmailComposer(toAddress: string, subject: string, body?: string) {
-  if (body) {
+  if (Platform.OS !== "ios") {
+    Linking.openURL(`mailto:${toAddress}?subject=${subject}${body ? "&body=" + body : ""}`)
```
ds300, line 6: we should probably be encoding the query string properly. We can use `import {stringify} from 'qs'` and then `stringify({body, subject})`

PR #4457, opened 2021-02-18
[WIP] chore: use cohesion screen tracking

MounirDhahri:

### Description
- This PR prepares to use cohesion everywhere to track screen changes instead of our local schema.
- [x] Make ProvideScreenTracking behave like ProvideScreenTrackingWith Cohesion and remove the latter
- [x] Make the changes in all files where we are using ProvideScreenTrackingWithCohesion
- [x] Change `lib/track/schema` to rely on cohesion instead of our local enums (for `PageNames` and `OwnerEntityType`)
- [x] Add missing PageNames/OwnerTypes to cohesion https://github.com/artsy/cohesion/pull/166
- [x] Communicate the changes with the data team
- [x] Upgrade cohesion version on eigen
- [x] Fix typescript issues and formatting issues (we are using snake case for screen names in cohesion)

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
     "cheerio": "0.22.0"
   },
   "dependencies": {
-    "@artsy/cohesion": "^1.73.0",
+    "@artsy/cohesion": "^1.74.0",
     "@artsy/palette-tokens": "^1.3.1",
     "@ptomasroos/react-native-multi-slider": "^2.2.2",
     "@react-native-community/async-storage": "1.6.3",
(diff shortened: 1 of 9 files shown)


mikehrom (commented):
Excited for this. thanks! Not sure if I need to contribute here, but once we merge changes, I'll have to watch traffic to make sure we're still seeing screen types in `eigen_production` schema tables (i.e. Segment is reading our analytics code correctly). 

Is there a way we can test for a small set of owner types? City guide owners would be perfect, since it's in the app, uses legacy screen tracking, and is not particularly important traffic these days.

PR #4458, opened 2021-02-18
fix fast reload

ds300:
### Description

This fixes the long-standing issue of metro's Fast Reload feature not working properly.

It turns out that relay rewrites all generated files every time you touch _any_ file in `src`, which in turn triggers a cascade of 'file updated' notifications to metro's Fast Reload service. On iOS this sudden influx seems to raise a flag, and instead of doing hundreds of Fast Reload operations it bails out and does one normal full reload. On android it doesn't bail out that way and instead just applies all the Fast Reload operations sequentially, which takes like 20s most of the time.

Anyway we can prevent the whole mess by making it so that relay only rewrites files that have actually changed.

### DIFF
--- a/HACKS.md
+++ b/HACKS.md
@@ -88,3 +88,13 @@
 
   - Stop using nested navigation containers.
   - Fix `@react-navigation/core` properly upstream.
+
+- relay-compiler
+
+  There are two hacks here:
+
+  - We hack the output of the compiler to provide clickable links for error messages. Relay assumes that you put your
+    `__genearted__` folder in the root of your project, but we put it in `src`.
+  - We make sure that files which do not change are not overwritten. This prevents excessive reloading by metro.
+
+  We can remove these hacks when they don't matter anymore. Neither are liley to be fixed by facebook.
--- a/HACKS.md
+++ b/HACKS.md
@@ -94,7 +94,7 @@
   There are two hacks here:
 
   - We hack the output of the compiler to provide clickable links for error messages. Relay assumes that you put your
-    `__genearted__` folder in the root of your project, but we put it in `src`.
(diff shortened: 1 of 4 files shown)


pvinis:
🤦

Such a good PR. It must have been a fun investigation! Good job to everyone involved.

HACKS.md
```diff
@@ -88,3 +88,13 @@
 
   - Stop using nested navigation containers.
   - Fix `@react-navigation/core` properly upstream.
+
+- relay-compiler
+
+  There are two hacks here:
+
+  - We hack the output of the compiler to provide clickable links for error messages. Relay assumes that you put your
+    `__genearted__` folder in the root of your project, but we put it in `src`.
```
pvinis, line 97:
```suggestion
    `__generated__` folder in the root of your project, but we put it in `src`.
```

HACKS.md
```diff
@@ -88,3 +88,13 @@
...
+  - We make sure that files which do not change are not overwritten. This prevents excessive reloading by metro.
+
+  We can remove these hacks when they don't matter anymore. Neither are liley to be fixed by facebook.
```
pvinis, line 100:
```suggestion
  We can remove these hacks when they don't matter anymore. Neither are likely to be fixed by facebook.
```

pvinis (approved):

MounirDhahri (approved):

MounirDhahri:
Goodbye infinite refreshes, you won't be missed 
https://media4.giphy.com/media/W3fRi6uCcjsTQKEGcr/giphy.gif

MounirDhahri:
Looks good to me. The edits Pavlos reported make sense, please feel free to merge once they are addressed ⭐️

ds300:
hmm is squash on green broken again?

PR #4462, opened 2021-02-19
MP query freshness checker prototype

pvinis:

This PR resolves [CX-]

### Description

This PR is ready. It will take a bunch of our eigen queries and run them against MP and complain if any fails.

Next steps:
- Use users with funky accounts


metaphysics-queries-freshness-tests/test-freshness.ts
```diff
@@ -0,0 +1,153 @@
...
+      count: 4,
+    },
+  })
+
+  // end game
+  if (errors.length !== 0) {
+    console.error(`Oh noes! ${errors.length} queries failed!`)
+    errors.map(({ request, error }) => {
+      console.warn("- The following query:")
+      console.log(request)
```
ds300, line 141: does this log out the whole query text? That's gonna make the error reports super noisy. I wonder if we could just quickly parse the query text and log out only the query name? Then people can find it in the query map if they want to debug via graphiql or insomnia.

pvinis: Ah. It was useful for me to print the whole thing because I could just copy-paste and run it manually myself using the playground, without the need to go find the query.

In the json file the query is in one line all messy, and in our source files we have things like `...SomeFragment`. Logging te whole thing expands these too, which is very useful I think.

ds300 (approved):
🚢

PR #4463, opened 2021-02-19
fix(tooling): ensure relay runs once before packager

ds300:
### Description

This PR fixes the following bug I'm sure you've all seen

image

#### How

At dev time relay sends the full query text to metaphysics (in beta and production it only sends the query ID). The full query text is stored in `data/complete.queryMap.json` which relay-compiler overwrites every time files change in watch mode. This file is not checked into git.

Until now `yarn relay` would first do `cat '{}' > data/complete.queryMap.json` to make sure the file exists for people who are running eigen for the first time (if the file is not there when the initial bundle begins, metro just quits).

Sometimes there seems to be a race condition where metro starts bundling before the first time relay writes a hydrated `data/complete.queryMap.json`, and from then on no updates to `data/complete.queryMap.json` trigger rebuilds. Thus the bundled query map is just an empty json object and we get the error about missing query text in the screenshot.

So in this PR we just make sure that there's a fully hydrated version of `data/complete.queryMap.json` before even starting metro. It will cost people a couple of seconds each time a person runs `yarn start` but that's better than facing this bug every now and then I reckon.


MounirDhahri (approved):

brainbicycle (approved):
Nice!

PR #4464, opened 2021-02-22
enhancement(sentry): enable release health tracking

pvinis:

This PR is related to [CX-867]

### Description

This PR just updated to latest sentry and enables the release health tracking. After this is merged I will release a beta to test it out.

Also, a call to `setRelease` is removed because it is not deprecated. The `release` key in the `init` config should be enough to set it.

[CX-867]: https://artsyproduct.atlassian.net/browse/CX-867

### DIFF
--- a/src/lib/ErrorReporting.ts
+++ b/src/lib/ErrorReporting.ts
@@ -1,4 +1,4 @@
-import { init, setRelease } from "@sentry/react-native"
+import Sentry from "@sentry/react-native"
 import Config from "react-native-config"
 import { LegacyNativeModules } from "./NativeModules/LegacyNativeModules"
 import { getCurrentEmissionState } from "./store/GlobalStore"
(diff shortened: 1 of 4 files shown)


ds300 (approved):

src/lib/ErrorReporting.ts
```diff
@@ -16,9 +16,10 @@ if (getCurrentEmissionState().sentryDSN) {
...
   }
 
-  init({
+  Sentry.init({
     dsn: getCurrentEmissionState().sentryDSN,
     release: sentryReleaseName,
+    enableAutoSessionTracking: true,
+    autoSessionTracking: true,
   })
-  setRelease(sentryReleaseName)
```
ds300, line 23: cc @brainbicycle was this line required for the release linking to work?

brainbicycle: This was a a workaround for a bug but in sentry but it might be fixed in newer versions, I'll put a comment on the ticket but maybe whoever QAs this can trigger the sentry test error and make sure it is deobfuscated in beta.

brainbicycle: That is to say I think we can leave it out!

pvinis: I can take a look at this and make sure it doesn't break the behavior we want.

brainbicycle (approved):
🎉

src/lib/ErrorReporting.ts
pvinis: Everything works still. Thanks!

PR #4470, opened 2021-02-22
feat: allow adding/editing pricePaid in myCollection

brainbicycle:

This PR resolves [CX-907]

### Description

This PR allows for specifying and editing pricePaid information on MyCollection artworks.

### Screenshots

<details>
<summary>Add Price Paid</summary>

https://user-images.githubusercontent.com/49686530/108908326-27c93d00-75f1-11eb-8607-9286def6a6ad.mp4

</details>

<details>
<summary>Remove Price Paid</summary>

https://user-images.githubusercontent.com/49686530/108908336-2a2b9700-75f1-11eb-9e9e-1621ebfba760.mp4

</details>

## TODO:

- [ ] Allow removing price paid information from MyCollection artworks

PR here will need to be merged before this:  https://github.com/artsy/metaphysics/pull/3013

Currently there is an issue with types in metaphysics which is preventing removing pricePaid information from artworks, believe the behavior we want is to allow specifying `null` explicitly for pricePaidCents field in the mutation and creation and this causing the entire pricePaid object to be removed from the artwork.

[CX-907]: https://artsyproduct.atlassian.net/browse/CX-907


src/lib/Scenes/MyCollection/Screens/ArtworkFormModal/MyCollectionArtworkFormModal.tsx
```diff
@@ -76,23 +76,38 @@ export const MyCollectionArtworkFormModal: React.FC<MyCollectionArtworkFormModal
...
+      ...others
+    }) => {
       setLoading(true)
       try {
         const externalImageUrls = await uploadPhotos(photos)
+
+        let pricePaidCentsClean
+        if (pricePaidCents && !isNaN(Number(pricePaidCents))) {
+          pricePaidCentsClean = Number(pricePaidCents) * 100
+        }
```
ds300, line 95: should `pricePaidCents` here be called just `pricePaid` since it's not actually in cents?

src/lib/Scenes/MyCollection/State/MyCollectionArtworkModel.tsx
```diff
@@ -42,8 +43,8 @@ export const initialFormValues: ArtworkFormValues = {
   artistIds: [],
   artistSearchResult: null,
   category: "",
-  costMinor: "", // in cents
-  costCurrencyCode: "",
+  pricePaidCents: "",
```
ds300, line 46: again it seems like this field is not actually representing cents and should probably be renamed

ds300 (commented):
Looks good so far. I think it would be good to clean up the cents vs dollars naming situation, but otherwise 👍🏼

src/lib/Scenes/MyCollection/Screens/ArtworkFormModal/MyCollectionArtworkFormModal.tsx
brainbicycle: very good point 🤦

brainbicycle: Alright I changed to `pricePaidDollars` and `pricePaidCents` where applicable, even though technically these aren't always dollars I find it more readable than something like `pricePaidMinorUnit` `pricePaidMajorUnit` but open to suggestions. Thanks for the feedback!

PR #4473, opened 2021-02-24
chore: disable hermes, enable JSC with intl

ds300:
### Description

I didn't check thoroughly, it turns out Hermes doesn't support `Intl` yet 😭 but it's being developed now so maybe in a couple of RN releases we'll be able to switch. We could also switch earlier by replacing our current (very limited) `Intl` usage with lightweight npm packages for number/date formatting until Hermes `Intl` support drops. I'll bring that up at Knowledge Share on thursday.


MounirDhahri (approved):

MounirDhahri:
That's sad to hear, but as you said, we may be able to use something else to replace it

PR #4479, opened 2021-02-25
Update Eigen to call the MP mutation

lilyfromseattle:
[PURCHASE-2382]

# Update Eigen to call the MP mutation

When a collector clicks the confirm button, we’ll call the mutation(s) to create a pending offer order with proper associations.

This is a WIP because I hope to add tracking tomorrow when the schema is ready, but I wanted to get a first found of feedback now. Yes, I know its a class component - I used the existing MO button as a template, but if people feel strongly that this should be a functional component I can convert it.

[PURCHASE-2382]: https://artsyproduct.atlassian.net/browse/PURCHASE-2382


src/lib/Scenes/Inbox/Components/Conversations/InquiryMakeOfferButton.tsx
```diff
@@ -1,35 +1,148 @@
...
+        style: "cancel",
+      },
+      {
+        text: "Retry",
+        onPress: () => {
+          this.handleCreateInquiryOfferOrder()
+        },
+      },
+    ])
+    console.log("src/lib/Scenes/Artwork/Components/InquiryMakeOfferButton.tsx", error)
```
xtina-starr, line 51: Did you mean to leave this console.log in?

src/lib/Scenes/Inbox/Components/Conversations/InquiryMakeOfferButton.tsx
```diff
@@ -1,35 +1,148 @@
...
+  //   action_type: Schema.ActionTypes.Tap,
+  //   context_module: Schema.ContextModules.Conversation,
+  // })
+  handleCreateInquiryOfferOrder() {
+    const { relay, artwork, editionSetID, impulseConversationId } = this.props
+    const { isCommittingCreateOfferOrderMutation } = this.state
+    const { internalID } = artwork
+
+    if (isCommittingCreateOfferOrderMutation) {
+      return
```
xtina-starr, line 65: Do we want an empty return here? I'd imagine we'd show some kind of loading indicator. Might be worth checking with Jonathan for what the loading state should be.

src/lib/Scenes/Artwork/Components/CommercialButtons/MakeOfferButton.tsx
```diff
@@ -106,7 +106,7 @@ export class MakeOfferButton extends React.Component<MakeOfferButtonProps, State
               } else if (orderOrError.__typename === "CommerceOrderWithMutationSuccess") {
                 navigate(`/orders/${orderOrError.order.internalID}`, {
                   modal: true,
-                  passProps: { artworkID: orderOrError.order.internalID },
+                  passProps: { orderID: orderOrError.order.internalID },
```
starsirius, line 109: Oops nice catch!

src/lib/Scenes/Inbox/Components/Conversations/InquiryMakeOfferButton.tsx
```diff
@@ -1,35 +1,148 @@
...
-      </Flex>
-    </>
-  )
+export interface State {
+  isCommittingCreateOfferOrderMutation: boolean
+  showCheckoutFlowModal: boolean
+  orderUrl: string | null
 }
+
+// @ts-expect-error STRICTNESS_MIGRATION --- 🚨 Unsafe legacy code 🚨 Please delete this and fix any type errors if you have time 🙏
```
starsirius, line 26: I assume these are copied over from existing implementation?

src/lib/Scenes/Inbox/Components/Conversations/InquiryMakeOfferButton.tsx
starsirius, line 51: I'm not familiar with error reporting in general in Eigen. Curious what would be a good way to track this for produciton.

src/lib/Scenes/Inbox/Components/Conversations/InquiryMakeOfferButton.tsx
```diff
@@ -1,35 +1,148 @@
...
+                      data
+                    }
+                  }
+                }
+              }
+            }
+          `,
+          variables: {
+            input: {
+              artworkId: internalID,
```
starsirius, line 95: Can alias `internalID` to `artworkID` and make it clear.

src/lib/Scenes/Inbox/Components/Conversations/InquiryMakeOfferButton.tsx
```diff
@@ -1,35 +1,148 @@
...
+              editionSetId: editionSetID,
+              impulseConversationId,
+            },
+          },
+          onCompleted: (data) => {
+            this.setState({ isCommittingCreateOfferOrderMutation: false }, () => {
+              const {
+                // @ts-expect-error STRICTNESS_MIGRATION --- 🚨 Unsafe legacy code 🚨 Please delete this and fix any type errors if you have time 🙏
```
xtina-starr, line 103: I'd be curious as to what the type errors would be here and if they are simply fixes. Not a blocker though. Just thinking it might be a good opportunity to clean some of them up.

src/lib/Scenes/Inbox/Components/Conversations/InquiryMakeOfferButton.tsx
```diff
@@ -1,35 +1,148 @@
...
+      <Button
+        onPress={() => this.handleCreateInquiryOfferOrder()}
+        loading={isCommittingCreateOfferOrderMutation}
+        size="large"
+        block
+        width={100}
+        variant={this.props.variant}
+        haptic
+      >
+        {this.props.buttonText ? this.props.buttonText : "Make offer"}
```
xtina-starr, line 136: What else could the text be other than Make offer?

starsirius (approved):
Have some minor comments/questions and overall looks great to me! 🥇  The test failure seems to related to the prop name changes.

src/lib/Scenes/Inbox/Components/Conversations/OpenInquiryModalButton.tsx
```diff
@@ -0,0 +1,39 @@
+import { navigate } from "lib/navigation/navigate"
+import { Button, Flex, Separator } from "palette"
+import React from "react"
+import styled from "styled-components/native"
+
+export interface OpenInquiryModalButtonProps {
+  artworkID: string
+  impulseConversationId: string | null | undefined
```
xtina-starr, line 8: This is a little nit-pick and feel free to use your discretion, but I'd propose changing this to simply `conversationID`. Knowing the services that this comes from makes sense in the context exchange, impulse, and metaphysics, but I don't think the client needs to bother with all of that. At this layer it's just a conversation id and could come from anywhere.

xtina-starr (commented):
This is looking good. I had one or two suggestions (that aren't blockers). And well-done on the mutation specs  🎉 

It looks like there is a failing spec related to the regular make offer flow. I wonder if it is some how point to the wrong mutation. I'd be happy to look at it with you.

src/lib/Scenes/Artwork/Components/CommercialButtons/MakeOfferButton.tsx
xtina-starr: Looks like you made the update here but not in the [spec here](https://github.com/artsy/eigen/blob/96d21a375b388ee1e0058dd1371a4d07136513dc/src/lib/Scenes/Artwork/Components/CommercialButtons/__tests__/CommercialButtons-tests.tsx#L225). Updating this should fix the test and get the build green.

src/lib/Scenes/Inbox/Components/Conversations/InquiryMakeOfferButton.tsx
lilyfromseattle: Yes I used the regular MO button as a template

src/lib/Scenes/Inbox/Components/Conversations/InquiryMakeOfferButton.tsx
lilyfromseattle: In the make offer modal the text is actually "Submit". The button that opens the modal says "make offer"

src/lib/Scenes/Inbox/Components/Conversations/InquiryMakeOfferButton.tsx
lilyfromseattle: This was there in the other MO button component so I left it in here although it looks like I didn't fully update the path

src/lib/Scenes/Inbox/Components/Conversations/InquiryMakeOfferButton.tsx
lilyfromseattle: There actually is a loading indicator it just comes from the web view component

lilyfromseattle: I wasn't sure and had to double check to verify though so thanks for that

src/lib/Scenes/Inbox/Components/Conversations/InquiryMakeOfferButton.tsx
lilyfromseattle: This is also copied over from the other MO button so I'm not sure

src/lib/Scenes/Inbox/Components/Conversations/InquiryMakeOfferButton.tsx
lilyfromseattle:
```const orderOrError: any
Property 'orderOrError' does not exist on type '{ readonly orderOrError: { readonly __typename: "CommerceOrderWithMutationSuccess"; readonly order: { readonly internalID: string; readonly mode: "BUY" | "OFFER" | "%future added value" | null; }; } | { ...; } | { ...; }; } | null'.ts(2339)

starsirius:
Merging this for now and let's follow up with the strictness migration!

PR #4484, opened 2021-02-26
fix zero appearing

pvinis:

This PR resolves [CX-]

### DIFF
--- a/src/lib/Scenes/AuctionResult/AuctionResult.tsx
+++ b/src/lib/Scenes/AuctionResult/AuctionResult.tsx
@@ -18,6 +18,7 @@ import { graphql, QueryRenderer } from "react-relay"
 import { useTracking } from "react-tracking"
 import { RelayModernEnvironment } from "relay-runtime/lib/store/RelayModernEnvironment"
 import { getImageDimensions } from "../Sale/Components/SaleArtworkListItem"
+import { auctionResultHasPrice, auctionResultText } from "./helpers"
 
 const CONTAINER_HEIGHT = 80
 
@@ -126,13 +127,8 @@ const AuctionResult: React.FC<Props> = ({ artist, auctionResult }) => {
     details.push(makeRow("Description", auctionResult.description, { fullWidth: true }))
   }
 
-  const hasSalePrice = !!auctionResult.priceRealized?.display
-  const now = moment()
-  const isFromPastMonth = auctionResult.saleDate
-    ? moment(auctionResult.saleDate).isAfter(now.subtract(1, "month"))
-    : false
-  const salePriceMessage =
-    auctionResult.boughtIn === true ? "Bought in" : isFromPastMonth ? "Awaiting results" : "Not available"
+  const hasSalePrice = auctionResultHasPrice(auctionResult)
+  const salePriceMessage = auctionResultText(auctionResult)
 
   const renderRealizedPriceModal = () => (
     <>
@@ -235,6 +231,7 @@ export const AuctionResultQueryRenderer: React.FC<{
       query={graphql`
         query AuctionResultQuery($auctionResultInternalID: String!, $artistID: String!) {
           auctionResult(id: $auctionResultInternalID) {
+            ...helpers_auctionResultHelpers
             internalID
             artistID
             boughtIn
--- a/src/lib/Scenes/AuctionResult/__tests__/helpers-tests.ts
+++ b/src/lib/Scenes/AuctionResult/__tests__/helpers-tests.ts
@@ -1,0 +1,57 @@
+import { helpers_auctionResultHelpers } from "__generated__/helpers_auctionResultHelpers.graphql"
+import moment from "moment"
+import { auctionResultHasPrice, auctionResultText } from "../helpers"
+
+describe("auction result helpers", () => {
+  it("works for existing price", () => {
+    const auctionResultWithPrice: helpers_auctionResultHelpers = {
+      " $refType": "helpers_auctionResultHelpers",
+      currency: "USD",
+      saleDate: moment().toISOString(),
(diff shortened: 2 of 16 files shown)


src/lib/Scenes/AuctionResult/AuctionResult.tsx
```diff
@@ -126,13 +127,8 @@ const AuctionResult: React.FC<Props> = ({ artist, auctionResult }) => {
...
   }
 
-  const hasSalePrice = !!auctionResult.priceRealized?.display
-  const now = moment()
-  const isFromPastMonth = auctionResult.saleDate
-    ? moment(auctionResult.saleDate).isAfter(now.subtract(1, "month"))
-    : false
-  const salePriceMessage =
-    auctionResult.boughtIn === true ? "Bought in" : isFromPastMonth ? "Awaiting results" : "Not available"
+  const hasSalePrice = auctionResultHasPrice(auctionResult)
```
pepopowitz, line 130: 💪 💪 💪

src/lib/Scenes/AuctionResult/__tests__/helpers-tests.ts
```diff
@@ -0,0 +1,52 @@
+import moment from "moment"
+import { auctionResultHasPrice, AuctionResultHelperData, auctionResultText } from "../helpers"
+
+describe("auction result helpers", () => {
```
pepopowitz, line 4: These tests are beautiful.

src/lib/Scenes/AuctionResult/helpers.ts
```diff
@@ -0,0 +1,46 @@
...
+// If tests or types break, the fix should be to just update this interface with the new metaphysics types.
+export interface AuctionResultHelperData {
+  currency: string | null
+  boughtIn: boolean | null
+  priceRealized: {
+    display: string | null
+    cents: number | null
+  } | null
+  saleDate: string | null
+}
```
pepopowitz, line 13: What do you think about using `Pick<whatever_the_matching_MP_type_is, "currency" | "boughtIn" | ...>` to eliminate a bit of duplication? It might reduce the chance of "If tests or types break, the fix should be to just update this interface with the new metaphysics types."

src/lib/Scenes/AuctionResult/helpers.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export const auctionResultHasPrice = (auctionResult: AuctionResultHelperData): boolean => {
+  if (
```
pepopowitz, line 16: These helpers are very clear and easy to read. ❤️ ❤️ ❤️

pepopowitz (approved):
💪 ❤️ 

This looks great. I made one suggestion about picking types from the relay-generated source....I wouldn't hold up a merge for it, though.

src/lib/Scenes/AuctionResult/helpers.ts
```diff
@@ -0,0 +1,46 @@
...
+    auctionResult.priceRealized === null ||
+    auctionResult.priceRealized.cents === null ||
```
MounirDhahri, line 18: I think we don't need this

src/lib/Scenes/AuctionResult/helpers.ts
MounirDhahri: Absolutely!

src/lib/Scenes/AuctionResult/helpers.ts
```diff
@@ -0,0 +1,46 @@
...
+    auctionResult.priceRealized.display === null ||
+    auctionResult.currency === null
+  ) {
+    return false
+  }
+
+  if (auctionResult.priceRealized.cents === 0) {
```
MounirDhahri, line 25: **nitpick:** Maybe you can move this to the previous check as well

pvinis: I wanted to keep the null checks and the actual value checks separate. Like, if it's null, then we have no price. But if we do have a price but it's zero, then we also actually have no price.

src/lib/Scenes/AuctionResult/helpers.ts
pvinis: Huh thats a good suggestion. I'lll try it first thing monday.

pvinis: I had to rename some things because we had `AuctionResult` for a list item and for the screen, so there was a name conflict on fragments.

Anyway, I changed the types to grab from the MP types. I liked the idea, I think it works better!

pvinis:
This should be ready to go, after the last changes to make the types a bit more robust. We pick from the used queries and fragments, rather than manually mirror the types.

src/lib/Scenes/AuctionResult/helpers.ts
MounirDhahri: You marked this as solved although it's not solved yet I guess. We don't do the conversion to cents (or whatever subunit) to all lots we get since we are showing here also showing very old lots that were sold in old currencies that we do not support yet (french francs for example) that we do not have a subunit for (similar to cents).

MounirDhahri: We have a ticket in our backlog to discuss those unit hopefully soon. https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=77&projectKey=CX&view=planning&selectedIssue=CX-1120&issueLimit=100&search=currenc

pvinis:
#squashongreen

src/lib/Scenes/AuctionResult/helpers.ts
MounirDhahri: @pvinis I think you missed my comments here

PR #4489, opened 2021-03-02
fix: relay environment

ds300:
### Description

This fixes a long-standing issue that you may have seen where you're developing the app where occasionally you'll get errors like `cannot read property 'me.canRequestEmailConfirmation' of null`.

That error is caused when the the query in `BottomTabsModel.tsx` fails. That query fetches `me { unreadConversationCount }` from metaphysics. Because we have `noThrow` set to `true` in our default relay environment, when the query fails and `me` comes back as `null`, relay thinks the new value of `me` everywhere should be `null` so it updates the cache and any mounted components that actively reference `me` throw errors.

To fix that I'm isolating the `unreadConversationCount` query in a separate environment that doesn't use the cache. At the same time I needed to clean up the cache middleware by moving the logic for handling persisted queries to its own middleware.


MounirDhahri:
Thanks for looking into this 👍

brainbicycle (approved):
Nice!

PR #4490, opened 2021-03-02
feat: support deep links on android

MounirDhahri:

This PR resolves [CX-1015]

### Description
- Extend Deep links URI Schemes on android
- Support deep links when the app is closed
- Store launch URL

**PS**:
- I created a follow-up ticket to look into the Deny list 
_https://artsyproduct.atlassian.net/browse/CX-1150_

[CX-1015]: https://artsyproduct.atlassian.net/browse/CX-1015


src/lib/AndroidApp.tsx
```diff
@@ -18,18 +19,53 @@ import { useStripeConfig } from "./utils/useStripeConfig"
 const Main: React.FC<{}> = track()(({}) => {
   const isHydrated = GlobalStore.useAppState((state) => state.sessionState.isHydrated)
   const isLoggedIn = GlobalStore.useAppState((state) => !!state.auth.userAccessToken)
+  const launchURL = GlobalStore.useAppState((state) => state.auth.sessionState.url)
+
+  const setLaunchURLAction = GlobalStore.actions.auth.setState
```
ds300, line 24: nitpick, nonblocking: `setState` can be used directly below, rather than being aliased here first.

src/lib/AndroidApp.tsx
```diff
@@ -18,18 +19,53 @@ import { useStripeConfig } from "./utils/useStripeConfig"
...
+      // If the state is hydrated and the user is logged in
+      // We navigate them to the the deep link
+
+      if (isHydrated && isLoggedIn) {
+        navigate(url)
+      }
+
+      // Otherwise, we save the deep link url to the global store/asynchstorage
+      // Then we redirect them to the login screen
+      setLaunchURLAction({ sessionState: { url } })
```
ds300, line 55: this should be in an `else` clause. Otherwise it will run whenever any deep link is evaluated, not just the initial one. And I think that could lead to duplicated calls to `navigate`. Admittedly we debounce `navigate` to avoid this kind of thing but it would be best to make the intention clear here I think.

src/lib/store/AuthModel.ts
```diff
@@ -5,11 +5,17 @@ import type { GlobalStoreModel } from "./GlobalStoreModel"
...
 export interface AuthModel {
+  // State
   userID: string | null
   userAccessToken: string | null
   userAccessTokenExpiresIn: string | null
   xAppToken: string | null
   xApptokenExpiresIn: string | null
+  sessionState: {
+    url: string | null
+  }
```
ds300, line 16: We should rename this to `launchURL`.

nonblocking: Might also make sense to remove it from easy-peasy and keep it in a `useState` or  maybe even `useRef` in the `Main` component. Seems like the only benefit of keeping it here is that we can see the `setState` actions in the dev log.

android/app/src/main/AndroidManifest.xml
```diff
@@ -26,6 +26,10 @@
           <category android:name="android.intent.category.DEFAULT" />
           <category android:name="android.intent.category.BROWSABLE" />
           <data android:scheme="artsy" />
+          <data android:scheme="https"
+              android:host="www.artsy.net" />
+              <data android:scheme="https"
+              android:host="www.m.artsy.net" />
```
ds300, line 32: should this be?
```suggestion
              android:host="m.artsy.net" />
```

Admittedly they both work but I feel like it's super unlikely we've ever used `www.m.artsy.net` in the wild.

CHANGELOG.yml
```diff
@@ -7,6 +7,7 @@ upcoming:
     - Improve bottom tabs dev state management  - david, mounir
     - Use adjustPan for windowSoftInputMode  - mounir
     - Make android app full-bleed - david
+    - Suppot deep links on android - adam, barry, brian, david, mounir, steven
```
brainbicycle, line 10:
```suggestion
    - Support deep links on android - adam, barry, brian, david, mounir, steven
``` 
Small typo

brainbicycle (approved):

android/app/src/main/AndroidManifest.xml
MounirDhahri: makes more sense, yes

src/lib/store/AuthModel.ts
MounirDhahri: makes sense

PR #4500, opened 2021-03-05
feat: Support echo kill switches

MounirDhahri:

This PR resolves [CX-1169]

### Description
- [x] Add react-native support to kill specific app versions
- [x] Add react-native support to kill versions below a specific app version

**What's Remaining**
- [x] Add tests
- [x] Update echo
- [x] Create force update app screen
- [x] Add ios support to kill specific app versions (native side)

https://user-images.githubusercontent.com/11945712/111307018-34254080-8659-11eb-9bc7-41c87683dcac.mov

[CX-1169]: https://artsyproduct.atlassian.net/browse/CX-1169


src/lib/store/config/__tests__/EchoModel-tests.tsx
```diff
@@ -56,4 +64,82 @@ describe("Echo", () => {
...
+            "1.9.0": {
+              message: "Please update the app",
+            },
+          },
+          android: {
+            "1.9.0": {
+              message: "Please update the app",
+            },
+          },
+        },
```
pvinis, line 88: Is this part needed in this test and the one below this? If not, maybe we can remove it so it's simple and obvious how this works? Because I am a bit confused 😅.

Also, where is the "current" version of the app in this test here?

pvinis:
Definitely let's change the name of `excluded`, it's very confusing. Maybe `badVersions` or `alwaysForceUpdateOnVersions` or something?

Very nice thought!

src/lib/store/config/__tests__/EchoModel-tests.tsx
MounirDhahri: yes it is needed to make sure that when the excluded versions array includes a version that doesn't match the version we have in the `app.json` we still tell the user that their version is valid

MounirDhahri:
you can maybe try `killedAppVersions` to have the same naming convention with the minimum app version

pvinis:
Added the hooks for devtoggles and in our new devtoggle (disable fetching remote echo config), when it becomes true, it will load the bundled echo config, and when it becomes false again, it will fetch fresh.

I tested it with the bundled echo config with viewing rooms disabled, and you can see that they are there at start, then it goes away, then it's there again.
https://f.p0l.co/file/dropshare-public-pavlos/2021-03-25-at-12.02.01-chocolate-cicada.gif

pvinis:
Since I worked on this too, I'll assign this to brian. I guess this is ready to go.

pvinis:
Since https://github.com/artsy/eigen/pull/4563 is ready, maybe we can merge that one onto this (it's already pointing to here), and we merge the resulting one from here to master?

MounirDhahri:
Great, thanks for leading this Pavlos. Now that echo is using `killedVersions` instead of `excludedVersions` can you rename it here and we merge this

pvinis:
It's renamed in #4563. I could cherry pick, but I think it's easier to just merge that one. It's pointing to this branch here, not master. I kinda made it a new branch so that we could review it on last KS.

MounirDhahri:
aha, I get you now, I will merge this then.

PR #4506, opened 2021-03-08
feat: only show make offer button in conversation when there is no active order (PURCHASE-2463)

starsirius:

This PR resolves [PURCHASE-2463] and depends on https://github.com/artsy/metaphysics/pull/3034.

### Description

We still need to address some loose ends in the recording (such as after submission we should go straight back to the conversation instead of to the artwork confirmation screen), but it demonstrates when we navigate back to the conversation, the "Make Offer" button is gone.

one-active-order

[PURCHASE-2463]: https://artsyproduct.atlassian.net/browse/PURCHASE-2463


src/lib/Scenes/Inbox/Screens/Conversation.tsx
```diff
@@ -267,6 +277,7 @@ export const ConversationQueryRenderer: React.FC<{
       variables={{
         conversationID,
       }}
+      cacheConfig={{ force: true }}
```
xtina-starr, line 280: Why add the force cache true here?

starsirius: When navigating from the conversation list screen to a particular conversation, requests are not fired. I had to disable caching for it to work. I'm curious if there is a better way to address it. 🤔

src/lib/Scenes/Inbox/Screens/Conversation.tsx
```diff
@@ -234,6 +234,16 @@ export const ConversationFragmentContainer = createFragmentContainer(Conversatio
             }
           }
         }
+        submittedOrderConnection: orderConnection(participantType: BUYER, state: SUBMITTED) {
```
sepans, line 237: 💯 alias

src/lib/Scenes/Inbox/Screens/Conversation.tsx
sepans: Very useful to know. Thanks for noticing @xtina-starr and the explanation @starsirius

sepans (approved):
🚀

lilyfromseattle:
Looks good! Just fix that `CHANGELOG` conflict and we can #mergeongreen

src/lib/Scenes/Inbox/Screens/Conversation.tsx
erikdstock: I'm not sure if it would work here but we have some kind of refresh-on-revisit logic in the Inbox component.

Issue #4499, opened 2021-03-05
[RFC] New changelog process


ds300:
## Problem

With the Android launch just around the corner, we need to figure out a way to manage independent release notes for iOS and Android. 

For iOS we currently generate release notes based on the `user_facing` portion of a changelog entry in [CHANGELOG.yml](https://github.com/artsy/eigen/tree/master/CHANGELOG.yml).

Here's the relevant code in our fastlane script https://github.com/artsy/eigen/blob/d88d47b5a6f7c315b7a1ad63089df766a425ed04/fastlane/Fastfile#L43-L44

Also I think many of us are frustrated by the current process's tendency to cause merge conflicts, so it would be nice to simultaneously find a way around that.

### Extending the current process

We can't simply replace the `user_facing` field with two fields: `ios_user_facing` and `android_user_facing` because we need to be able to release the app independently on both platforms.

But what we could do is replace `CHANGELOG.yml` with `CHANGELOG_IOS.yml` and `CHANGELOG_ANDROID.yml`.

Pros:

- Quick change to make
- Close to what we already do, so not much learning involved.

Cons:

- Ultimately most changes product engineers make will be cross-platform, so they will often need to touch two files instead of one. Double merge conflicts!
- What to do with dev-only changes? Duplicate them on both platforms or move them to their own changelog file? I personally find it useful to cross-reference dev-only changes with user-facing changes, so duplicating would be helpful for me at least. But, again, who wants to do all that copy-pasting and merge conflict resolution?

### Building a new system based on PR descriptions (proposed solution)
 
We could use PR descriptions as a source of truth to avoid merge conflicts, manual duplication, and making extra 'update changelog' commits.

How would that work? 

#### PR template section + danger validation
We'd add a new section to the PR template. Something along the lines of:

```markdown
### Changelog updates
<!-- 📝 Please fill out at least one of these sections. -->
<!-- ⓘ 'User-facing' changes will be published as release notes. -->
<!-- ⌫ Feel free to remove sections that don't apply. -->
<!-- • Write a markdown list or just a single paragraph, but stick to plain text. -->
<!-- 🤷‍♂️ Replace this entire block with the hashtag `#nochangelog` to avoid updating the changelog. -->
#### Cross-platform user-facing changes
- 
#### iOS user-facing changes
- 
#### Android user-facing changes
- 
#### Dev changes
- 
```
And then we'd add a danger rule to validate the requirements outlined in the comments.

#### Log collation tooling

[...]

PR #4507, opened 2021-03-09
feat: multi-selection time period filters

iskounen:

This PR resolves [FX-2624]

This PR is the work of @dzucconi, @dblandin, and @anandaroop 

### Description

- Allows multiple selection of time period (a.k.a. major period) filters on artwork grids.

### Screenshots

<details><summary>Before</summary>

before_time_options

</details>
<details><summary>After</summary>

after_time_options

</details>

[FX-2624]: https://artsyproduct.atlassian.net/browse/FX-2624

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -15,6 +15,7 @@ upcoming:
     - Add feature flag for new artwork filters - iskounen
     - Enable feature flags on android - mounir
     - Enable new onboarding feature flag - mounir
+    - Adds improved time period filter behind feature flag - iskounen, damon, devon, roop
   user_facing:
     - Adds inquiry checkout offer status CTA - lily
     - Update consignment photo selection flow for iOS 14 - brian
--- a/src/lib/utils/ArtworkFilter/ArtworkFiltersStore.tsx
+++ b/src/lib/utils/ArtworkFilter/ArtworkFiltersStore.tsx
@@ -262,6 +262,7 @@ const getSortDefaultValueByFilterType = (filterType: FilterType) => {
 
 export const ParamDefaultValues = {
   acquireable: false,
+  additionalGeneIDs: [],
   allowEmptyCreatedDates: true,
   artistIDs: [],
   atAuction: false,
@@ -271,11 +272,10 @@ export const ParamDefaultValues = {
   dimensionRange: "*-*",
   earliestCreatedYear: undefined,
   estimateRange: "",
-  additionalGeneIDs: [],
   includeArtworksByFollowedArtists: false,
   inquireableOnly: false,
   latestCreatedYear: undefined,
-  majorPeriods: undefined,
+  majorPeriods: [],
   medium: "*",
   offerable: false,
   partnerID: undefined,
@@ -356,7 +356,7 @@ export const selectedOptionsUnion = ({
       displayText: "All",
     },
     { paramName: FilterParamName.color, displayText: "All" },
-    { paramName: FilterParamName.timePeriod, paramValue: "*-*", displayText: "All" },
+    { paramName: FilterParamName.timePeriod, paramValue: [], displayText: "All" },
     {
       paramName: FilterParamName.waysToBuyBuy,
       paramValue: false,
--- a/src/lib/Components/ArtworkFilterOptions/TimePeriodMultiOptions.tsx
+++ b/src/lib/Components/ArtworkFilterOptions/TimePeriodMultiOptions.tsx
@@ -1,0 +1,126 @@
+import { StackScreenProps } from "@react-navigation/stack"
+import {
+  ArtworkFilterContext,
+  FilterData,
+  ParamDefaultValues,
+  useSelectedOptionsDisplay,
+} from "lib/utils/ArtworkFilter/ArtworkFiltersStore"
+import {
+  FilterDisplayName,
+  FilterParamName,
+  getDisplayNameForTimePeriod,
+} from "lib/utils/ArtworkFilter/FilterArtworksHelpers"
+import { useArtworkFiltersAggregation } from "lib/utils/ArtworkFilter/useArtworkFilters"
+import _ from "lodash"
+import React, { useContext, useState } from "react"
+import { FilterModalNavigationStack } from "../FilterModal"
+import { MultiSelectOptionScreen } from "./MultiSelectOption"
+
+interface TimePeriodOptionsScreenProps
+  extends StackScreenProps<FilterModalNavigationStack, "TimePeriodOptionsScreen"> {}
+
+export const TimePeriodMultiOptionsScreen: React.FC<TimePeriodOptionsScreenProps> = ({ navigation }) => {
+  const { dispatch } = useContext(ArtworkFilterContext)
+  const { aggregation } = useArtworkFiltersAggregation({ paramName: FilterParamName.timePeriod })
+
+  const DEFAULT_OPTION: FilterData = {
+    displayText: "All",
+    paramName: FilterParamName.timePeriod,
+    paramValue: ParamDefaultValues.majorPeriods,
+  }
+
+  const TIME_PERIOD_OPTIONS: FilterData[] = [
+    DEFAULT_OPTION,
+    ...(aggregation?.counts ?? []).map((c) => {
+      return {
+        displayText: getDisplayNameForTimePeriod(c.name),
+        paramName: FilterParamName.timePeriod,
+        paramValue: c.value,
+      }
+    }),
+  ]
+
+  const selectedOptions = useSelectedOptionsDisplay()
+
+  const selectedTimePeriodOptions = selectedOptions.find((option) => {
+    return option.paramName === FilterParamName.timePeriod
+  })
+
+  const [nextOptions, setNextOptions] = useState<string[]>((selectedTimePeriodOptions?.paramValue as string[]) ?? [])
+
+  const handleSelect = (option: FilterData, updatedValue: boolean) => {
+    if (updatedValue) {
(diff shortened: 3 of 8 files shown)


iskounen:
[The Apply button](https://github.com/artsy/eigen/blob/d8a9e7601d968ee34396bc9a6df30bf513aba38d/src/lib/Components/FilterModal/FilterModal.tsx#L162) is currently displaying "Apply (1)" for any combination of time periods because we pass all selected options as one array of values like so:

```
[{"displayText": "2010-2019", "paramName": "majorPeriods", "paramValue": ["2020", "2010"]}]
```

The "Ways to buy" filter, however, can display "Apply (2)" when two options are selected because it passes each selected option as a separate filter like so:

```
[{"displayText": "Bid", "paramName": "atAuction", "paramValue": true}, {"displayText": "Make offer", "paramName": "offerable", "paramValue": true}]
```

MounirDhahri:
This looks good.

About the designs, I thought we were trying to push more into the direction of using checkmarks instead of toggle switches to have more consistent behavior in all filter types. It offers a better UX (since the user will expect to check an item by pressing anywhere in the row instead of only selecting the toggle switch). But I see that the filters specs made by CX are designed differently than the ones made by FX. @samchieng do you mind if you clear this out to make sure we don't have more different type of filters in the future.

> The Apply button is currently displaying "Apply (1)" for any combination of time periods because we pass all selected options as one array of values like so:

That's great, that's the behavior we want.

> The "Ways to buy" filter, however, can display "Apply (2)" when two options are selected because it passes each selected option as a separate filter like so:

Ideally we should be showing this as 1 as well. Take a look at this https://github.com/artsy/eigen/blob/ac5ac0ab8c2ad354e3ab6aee1b7882ce78dced13/src/lib/Components/FilterModal/FilterModal.tsx#L177-L187 and this https://github.com/artsy/eigen/blob/ac5ac0ab8c2ad354e3ab6aee1b7882ce78dced13/src/lib/Components/FilterModal/FilterModal.tsx#L603-L614

src/lib/Components/ArtworkFilterOptions/__tests__/TimePeriodOptions-tests.tsx
```diff
@@ -1,71 +1,152 @@
...
+    }
+
+    it("displays a comma-separated list of the selected filters on the filter modal screen", () => {
+      const tree = renderWithWrappers(<MockFilterScreen initialState={state} />)
+      const items = tree.root.findAllByType(FilterModalOptionListItem)
+      const item = items.find((i) => extractText(i).startsWith("Time period"))
+
+      expect(item).not.toBeUndefined()
+      if (item) {
+        expect(extractText(item)).toContain("2020, 2000")
```
anandaroop, line 104: Based on real-world data I was expecting to see these as ranges ("2000–2010" etc) — but this is fine bc they match the mock data, yeah?

anandaroop (approved):
Nice, and thanks for leading this mob 🙏

CHANGELOG.yml
```diff
@@ -14,6 +14,7 @@ upcoming:
     - Update consignment photo selection flow for iOS 14 - brian
     - Allow adding pricePaid info in my collection - barry, brian
     - Adds inquiry make offer button - lily
+    - Adds multi-select time period filters to artwork grid - adam, damon, devon, roop
```
dblandin, line 17: Might not be a big deal, this first name pattern is bound to create conflicts. I think we already have two adams in the changelog now. Maybe we should switch to GitHub usernames at some point?

dblandin (approved):
LGTM 👍 💯

samchieng:
@MounirDhahri yes, apologies for the inconsistencies right now! Currently collaborating with Tricia to refine the final filter specs.

CHANGELOG.yml
iskounen: I'm happy to use my last name for now. There's some discussion about the future of developer names in public release notes as a part of [this RFC](https://github.com/artsy/eigen/issues/4499) (whether to pull them from GitHub or discontinue them altogether).

src/lib/Components/ArtworkFilterOptions/__tests__/TimePeriodOptions-tests.tsx
iskounen: That's correct. I confirmed by changing the mocked data to `["Roger", "Robert"]` and observing that the result was a comma-separated string `"Roger, Robert"`.

iskounen:
This work will be placed behind a [new feature flag](https://github.com/artsy/eigen/pull/4515) before the PR can come out of its draft state.

src/lib/Components/ArtworkFilterOptions/MultiSelectOption.tsx
```diff
@@ -50,7 +50,7 @@ export const MultiSelectOptionScreen: React.FC<MultiSelectOptionScreenProps> = (
       <FancyModalHeader onLeftButtonPress={handleBackNavigation}>{filterHeaderText}</FancyModalHeader>
       <Flex flexGrow={1}>
         <FlatList<FilterData>
-          initialNumToRender={4}
```
iskounen, line 53: This was causing a visual glitch where the first 4 filter options were displayed for a split second before the rest of the options.

multi-select-opion-screen_lag

src/lib/Components/ArtworkFilterOptions/MultiSelectOption.tsx
```diff
@@ -50,7 +50,7 @@ export const MultiSelectOptionScreen: React.FC<MultiSelectOptionScreenProps> = (
...
+          style={{ flex: 1 }}
```
iskounen, line 53: This resolved a bug where the last 2 filter options would snap behind the "Apply" button and prevent users from selecting them.

multi-select-opion-screen_scroll

src/lib/Components/ArtworkFilterOptions/TimePeriodOptions.tsx
```diff
@@ -10,11 +11,18 @@ import _ from "lodash"
...
 interface TimePeriodOptionsScreenProps
   extends StackScreenProps<FilterModalNavigationStack, "TimePeriodOptionsScreen"> {}
 
-export const TimePeriodOptionsScreen: React.FC<TimePeriodOptionsScreenProps> = ({ navigation }) => {
+export const TimePeriodOptionsScreen: React.FC<TimePeriodOptionsScreenProps> = ({ navigation, route }) => {
+  const useImprovedArtworkFilters = useFeatureFlag("ARUseImprovedArtworkFilters")
+
+  if (useImprovedArtworkFilters) {
+    return <TimePeriodMultiOptionsScreen navigation={navigation} route={route} />
+  }
```
iskounen, line 24: @pvinis I could use your advice with how I'm putting these changes behind a feature flag.

My intention was to keep the existing and new filter components separate to make development on the new filters cleaner, and then replace the old components with the new components when we're ready to launch. (We will be repeating this work for 5 more filters this month.)

Does that match how other teams use feature flags for components?

src/lib/utils/ArtworkFilter/ArtworkFiltersStore.tsx
```diff
@@ -271,11 +272,10 @@ export const ParamDefaultValues = {
   dimensionRange: "*-*",
   earliestCreatedYear: undefined,
   estimateRange: "",
-  additionalGeneIDs: [],
   includeArtworksByFollowedArtists: false,
   inquireableOnly: false,
   latestCreatedYear: undefined,
-  majorPeriods: undefined,
+  majorPeriods: [],
```
iskounen, line 278: I didn't put this behind a feature flag because I don't think it matters for the current implementation what the default value is, as long as there is one.

I tested the All option on the time periods filter with and without the feature flag to confirm this, and both the results and the filters screen behaved as expected.

src/lib/Components/ArtworkFilterOptions/MultiSelectOption.tsx
dblandin: Should we pull this out into a separate commit? I wonder if this should also have a separate changelog entry.

src/lib/Components/ArtworkFilterOptions/MultiSelectOption.tsx
dblandin: Same comment as above. Should we pull this out into a separate commit? I wonder if this should also have a separate changelog entry.

src/lib/Components/ArtworkFilterOptions/TimePeriodOptions.tsx
dblandin: This might work well enough for this purpose but I would expect the feature flag check to be one level up. Possible in FilterModal.tsx where we're defining the available filters.

src/lib/Components/ArtworkFilterOptions/MultiSelectOption.tsx
iskounen: I've removed this change and will open a new PR for that.

src/lib/Components/ArtworkFilterOptions/MultiSelectOption.tsx
iskounen: Same response as above :)

src/lib/Components/ArtworkFilterOptions/TimePeriodOptions.tsx
iskounen: Good call. I moved it up one level.

dzucconi (approved):
Looks good to me. Feel free to merge after the changelog conflict is resolved 😞

PR #4509, opened 2021-03-10
fix: add safety valve in extractNodes

ds300:
We had some bug reports about favourites disappearing, and we saw that metaphysics was returning `null` as node values for a shows connection, which should not be happening. We'll fix that separately, and try to add some tooling in metaphyiscs to prevent it happening again. But in case it does happen again in the future we can add this extra helper in `extractNodes` to filter out null/undefined.

Slack context: https://artsy.slack.com/archives/C01B2P6LJUU/p1615324641014600

### DIFF
--- a/src/lib/utils/__tests__/extractNodes-tests.ts
+++ b/src/lib/utils/__tests__/extractNodes-tests.ts
@@ -16,5 +16,6 @@ describe(extractNodes, () => {
     expect(extractNodes(null)).toEqual([])
     expect(extractNodes({ edges: null })).toEqual([])
     expect(extractNodes({ edges: [] })).toEqual([])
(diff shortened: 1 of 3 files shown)


src/lib/utils/__tests__/extractNodes-tests.ts
```diff
@@ -16,5 +16,6 @@ describe(extractNodes, () => {
     expect(extractNodes(null)).toEqual([])
     expect(extractNodes({ edges: null })).toEqual([])
     expect(extractNodes({ edges: [] })).toEqual([])
+    expect(extractNodes({ edges: [null] })).toEqual([])
```
pvinis, line 19: sweet!

pvinis (approved):

PR #4465, opened 2021-02-22
chore: Move environment config to TS

ds300:
This PR resolves [CX-871]

### Description

This PR moves the app's [environment configuration](https://github.com/artsy/eigen/pull/4465/files#diff-13f9506e86aecda94da8d225c6be033bf731681310428cee9b039ee612c10b15) to TS.

This was quite a significant change that had a lot of knock-on effects and related fixes:

- Consolidated sentry projects into one (previously 'eigen-beta' and 'eigen'. Now just 'eigen'. See [slack context](https://artsy.slack.com/archives/C02BAQ5K7/p1613498325159700))
- Moved [auth validation check](https://github.com/artsy/eigen/pull/4465/files#diff-62a4f62f32cb84f82db8a186424bd7cce489a9575d2888db316cb90fdf25e995) to TS. (this check is responsible for logging the user out with a friendly Alert if the user's session expires due to password changes etc)
- Refactored the `ConfigModel` into separate models for feature flags, echo state management, and env state management.
- Simplified sign out control flow.
  - It used to be a cyclical thing where we'd dispatch an event requesting a logout through ARNotificationsManager, then the native user state would be cleared and it would dispatch an event back over the bridge to clear the easy peasy state. 
  - Now [we clear the easy peasy state directly](https://github.com/artsy/eigen/blob/902a8c74c8d0ac0ef0f3680aa084cd9d6c92cfb2/src/lib/store/GlobalStoreModel.ts#L54-L63) (without clearing admin settings) at the same time as asking `ARUserManager` to clear its state
- Added a `TS -> objc` [state bridge for env strings](https://github.com/artsy/eigen/blob/902a8c74c8d0ac0ef0f3680aa084cd9d6c92cfb2/emission/Pod/Classes/EigenCommunications/ARNotificationsManager.m#L25-L32). (now the only obj-c controlled state is auth/onboarding-related, so we'll be fully TS when the new onboarding stuff is deployed on iOS)
- Added basic admin menu functionality for toggling between staging and production, and for using local URLs in dev mode.

https://user-images.githubusercontent.com/1242537/108872197-d5d4e700-75f1-11eb-82d1-9666d3f2044b.mp4

https://user-images.githubusercontent.com/1242537/108873055-ba1e1080-75f2-11eb-9671-f0635b631f9d.mp4

Things to QA (Completed 💆‍♀️)

- [x] Sign out (iOS and android)
- [x] Switch between staging and production (iOS and android)
- [x] Inquiry (iOS only)
[...]

PR #4511, opened 2021-03-10
fix: setup router before emission bootstraps #trivial

brainbicycle:

This PR resolves [CX-1176]

### Description

As part of https://github.com/artsy/eigen/pull/4465 we moved some app setup code to an afterBootstrap function that fires after the packaged js is loaded. One of these setup calls was `[ARRouter setup]` which among other things sets up an http client for http requests that still live in Obj-C. When a user taps a Sailthru link we use this http client to get redirected to our own artsy links, then in turn route the user to that page in the app. When the app is in a killed state and a user taps one of these links the bootstrap function has not yet been called and so the httpclient is not setup causing a crash. Since we are using NSOperationQueue for this http request the operation was being created but never completed causing it to get stuck in the queue and blocking all subsequent Sailthru requests. Moving the router setup back to outside this function fixes the issue.

[CX-1176]: https://artsyproduct.atlassian.net/browse/CX-1176


bhoggard (approved):

Artsy/App/ARAppDelegate.m
```diff
@@ -154,6 +154,7 @@ - (void)setupForAppLaunch
     _landingURLRepresentation = self.landingURLRepresentation ?: @"https://artsy.net";
 
     [[ARLogger sharedLogger] startLogging];
+    [ARRouter setup];
```
ds300, line 157: this causes the app to crash on launch because it tries to use emission before emission has been initialized.

Instead of this we should make sure that we use `afterBootstrap` to delay evaluating the sailthru link.

Something like this: 
image

I should have caught this when working on #4465 :( — will take some time today to do a more thorough audit of all usages of ARRouter to make sure there are no other places where we might be using it before the app has finished launching.

ds300: 📝 audit complete. There are no other places where ARRouter might be used before the app has initialized.

brainbicycle: Thanks David! I guess I didn't notice the crash because I was running on device and the sharedInstance was already instantiated, I'll update

ds300 (approved):
🙏🏼

PR #4524, opened 2021-03-12
fix: allow users to scroll to bottom of multis-elect filters

iskounen:

### Description

When a user scrolls to the bottom of a long multi-select filter (like the categories filter on an artwork grid) they are not able to keep the bottom 2 options in view long enough to select them before the snap to hide behind the "Apply" button.

multi-select-opion-screen_scroll


iskounen:
@dzucconi I assigned this ticket to you because we spoke about it during our mob session on Monday.

src/lib/Components/ArtworkFilterOptions/MultiSelectOption.tsx
```diff
@@ -51,6 +51,7 @@ export const MultiSelectOptionScreen: React.FC<MultiSelectOptionScreenProps> = (
       <Flex flexGrow={1}>
         <FlatList<FilterData>
           initialNumToRender={4}
+          style={{ flex: 1 }}
```
dzucconi, line 54: 😗👌

PR #4528, opened 2021-03-15
feat: interleaved conversation messages w/ order updates

erikdstock:

This PR resolves [PURCHASE-2450]

### Description
This PR interleaves messages with the new `orderHistory` field. It is ready for QA and merge with the exception of the `Messages.tsx` file which still needs a little bit of type cleanup. We've added tests and fixed a bug on the way.

paired on this with @lilyfromseattle .

cc @artsy/purchase-devs 

Simulator Screen Shot - iPhone X - 2021-03-22 at 16 19 33

[PURCHASE-2450]: https://artsyproduct.atlassian.net/browse/PURCHASE-2450


src/lib/Scenes/Inbox/Components/Conversations/Messages.tsx
```diff
@@ -24,15 +29,31 @@ const LoadingIndicator = styled.ActivityIndicator`
...
   const [reloadingData, setReloadingData] = useState(false)
 
-  const [messages, setMessages] = useState<MessageGroupType[]>()
+  const [messages, setMessages] = useState<MessageGroup[]>()
+
+  const orders: Order[] = extractNodes(conversation?.orderConnection)
+
+  const allOrderEvents: OrderHistoryEvent[] = orders
+    .reduce<OrderHistoryEvent[]>((prev, order) => prev.concat(order.orderHistory), [])
+    .map((event, index) => ({ ...event, key: `event-${index}` }))
```
erikdstock, line 53: here we flatmap all orders to their events (still figuring out best way to handle the keys bit)...

src/lib/Scenes/Inbox/Components/Conversations/Messages.tsx
```diff
@@ -42,8 +63,14 @@ export const Messages: React.FC<Props> = forwardRef((props, ref) => {
...
         return { key: node.id, ...node }
       })
-      .reverse()
-    setMessages(groupMessages(nodes))
+
+    const sortedMessages = sortBy([...allOrderEvents, ...allMessages], (message) =>
+      DateTime.fromISO(message.createdAt!)
+    )
+    // TODO: Check ordering and reversing around here
+    const groupedMessages = groupMessages(sortedMessages as any)
```
erikdstock, line 71: Here we combine and group messages+events...

src/lib/Scenes/Inbox/Components/Conversations/Messages.tsx
```diff
@@ -117,13 +144,13 @@ export const Messages: React.FC<Props> = forwardRef((props, ref) => {
             group={item}
             conversationId={conversation.internalID!}
             subjectItem={conversation.items?.[0]?.item!}
-            key={`group-${index}-${item[0]?.key}`}
+            key={`group-${index}-${(item[0] as any)?.key}`}
```
erikdstock, line 147: here we pass our groups in...

src/lib/Scenes/Inbox/Components/Conversations/TimeSince.tsx
```diff
@@ -8,10 +8,13 @@ const exactDate = (time: string) => {
...
   const date = moment(time)
+  const isToday = moment().isSame(date, "day")
+  const isYesterday = moment().subtract(1, "days").isSame(date, "day")
   const daysSince = moment().diff(date, "days")
-  if (daysSince === 0) {
+
+  if (isToday) {
     return date.format("[Today] h:mm A")
-  } else if (daysSince === 1) {
+  } else if (isYesterday) {
```
erikdstock, line 17: minor bugfix

src/lib/Scenes/Inbox/Components/Conversations/MessageGroup.tsx
```diff
@@ -19,51 +21,89 @@ const SubjectContainer = styled(Flex)`
...
+    const { group } = this.props
+    const firstItem = group[0]
+    if (!firstItem) {
+      return null
+    }
+    // Events come as a single item in an array
+    if (isRelevantEvent(firstItem)) {
+      return <OrderUpdateFragmentContainer event={firstItem as any} />
+    }
+    if (isMessageArray(group)) {
```
erikdstock, line 86: in our MessageGroup mapping function we now just check whether the group is an event (in which case it appears alone as a single-item array) or a message group (in which case it is a group of 1+ messages).

src/lib/Scenes/Inbox/Components/Conversations/OrderUpdate.tsx
```diff
@@ -0,0 +1,86 @@
...
+export interface OrderUpdateProps {
+  event: OrderUpdate_event
+}
+
+export const OrderUpdate: React.FC<OrderUpdateProps> = ({ event }) => {
+  let color: Color
+  let message: string
+  let Icon: React.FC<IconProps> = MoneyFillIcon
+
+  if (event.__typename === "CommerceOfferSubmittedEvent") {
```
erikdstock, line 16: `OrderUpdate` renders an event inline with the conversation.

erikdstock (commented):
basic tour of changes here

src/lib/Scenes/Inbox/Components/Conversations/Messages.tsx
```diff
@@ -42,8 +63,14 @@ export const Messages: React.FC<Props> = forwardRef((props, ref) => {
       .map((node) => {
         return { key: node.id, ...node }
       })
-      .reverse()
-    setMessages(groupMessages(nodes))
+
+    const sortedMessages = sortBy([...allOrderEvents, ...allMessages], (message) =>
+      DateTime.fromISO(message.createdAt!)
```
starsirius, line 68: If the `createdAt` is already an ISO string, I think itself should be sufficient for sorting and doesn't need to be converted to `DateTime`.

Also is the `!` in `createdAt!` a typo or is it a valid syntax?

starsirius: Ohh looks like it's the [Non-null assertion operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator)! #TIL

src/lib/Scenes/Inbox/Components/Conversations/OrderUpdate.tsx
```diff
@@ -0,0 +1,86 @@
...
+    const { offer } = event
+    const isCounter = offer.respondsTo !== null
+    if (offer.fromParticipant === "BUYER") {
+      color = "black100"
+      message = `You sent ${isCounter ? "a counteroffer" : "an offer"} for ${event.offer.amount}`
+    } else if (offer.fromParticipant === "SELLER") {
+      color = "copper100"
+      Icon = AlertCircleFillIcon
+      message = `You received ${isCounter ? "a counteroffer" : "an offer"} for ${event.offer.amount}`
```
starsirius, line 25: Currently partners would not be able to send an offer but this logic makes it flexible for future use cases! 💯

src/lib/Scenes/Inbox/Components/Conversations/OrderUpdate.tsx
```diff
@@ -0,0 +1,86 @@
...
+    }
+  } else if (event.__typename === "CommerceOrderStateChangedEvent") {
+    const { state, stateReason } = event
+    if (state === "APPROVED") {
+      color = "green100"
+      message = `Offer Accepted`
+    } else if (state === "CANCELED" && stateReason?.includes("_rejected")) {
+      color = "red100"
+      message = `Offer Declined`
+    } else if (state === "CANCELED" && stateReason?.includes("_lapsed")) {
```
starsirius, line 38: Nice! The logic here is pretty self-explanatory!

src/lib/Scenes/Inbox/Components/Conversations/MessageGroup.tsx
starsirius: Love this approach!

src/lib/Scenes/Inbox/Components/Conversations/Messages.tsx
starsirius: I think we are paginating messages. In case where order events are older than the oldest message we fetch so far, will we then still display order events above the messages?
```
event
event
event
event
oldest message fetched so far
event
...
latest message
```
Or maybe we want to filter out order events that's older than the oldest message fetched so far. So we would display this instead
```
oldest message fetched so far
event
...
latest message
```
And when we fetch the next page of messages, the older order events will then be displayed and interleaved with messages.

starsirius (approved):
This is looking very good! Love the progress here and test coverage! 🥇  @erikdstock @lilyfromseattle

src/lib/Scenes/Inbox/Components/Conversations/Messages.tsx
erikdstock: yeah, this has been in the back of my head as a case to handle. good callout

src/lib/Scenes/Inbox/Components/Conversations/MessageGroup.tsx
erikdstock: Looking at this I'm wishing i was testing an item array in both cases instead of a single item in one case.

PR #4532, opened 2021-03-16
fix: search input layout animation

MounirDhahri:

### Description

- Fix buggy search input animation
- This serves as an example on how we might solve some of the issues here https://artsyproduct.atlassian.net/browse/CX-1096

**Before**

https://user-images.githubusercontent.com/11945712/111333117-a7877c00-8672-11eb-8d12-7dfb71d71e01.mov

___
**After**

https://user-images.githubusercontent.com/11945712/111333108-a5bdb880-8672-11eb-922f-dd9055e6b66a.mov

### DIFF
--- a/src/lib/Components/SearchInput.tsx
+++ b/src/lib/Components/SearchInput.tsx
@@ -1,15 +1,22 @@
 import SearchIcon from "lib/Icons/SearchIcon"
 import { Flex, Sans } from "palette"
 import React, { RefObject, useState } from "react"
-import { LayoutAnimation, TextInput, TouchableOpacity } from "react-native"
+import { LayoutAnimation, Platform, TextInput, TouchableOpacity, UIManager } from "react-native"
 import { Input, InputProps } from "./Input/Input"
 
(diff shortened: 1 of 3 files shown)


src/lib/Components/SearchInput.tsx
```diff
@@ -1,15 +1,22 @@
...
 interface SearchInputProps extends InputProps {
   enableCancelButton?: boolean
 }
 
+if (Platform.OS === "android") {
+  if (UIManager.setLayoutAnimationEnabledExperimental) {
+    UIManager.setLayoutAnimationEnabledExperimental(true)
+  }
+}
+
```
ds300, line 16: should we just set this globally in `index.android.js`?

MounirDhahri: yes, yes, yes

MounirDhahri: I will set it inside AndroidApp.tsx

ds300:
great catch! ❤️

PR #4534, opened 2021-03-16
feat(fx-2623): Colors multi-select filter options screen

dzucconi:

This PR resolves [FX-2623]

### Description

One minor thing you'll notice is that the selected options overview stays at "All" this is just because there's a mis-match between `color` and `colors` as filters. This works correctly once we swap this out properly without the feature flag. I can explore this more if it's a blocker for now?

The other thing to note is that there really isn't a separate white and black color filter. This was just misdirection in the original layout (remember the existence of 'black-and-white-2'?) and I think a weird UI choice since it doesn't do what you'd expect it to do. I believe Tricia is still meditating on how to resolve this paradox.

[View screen recording](https://static.damonzucconi.com/_capture/PrKNB4JQ.gif)

[FX-2623]: https://artsyproduct.atlassian.net/browse/FX-2623


src/lib/Components/ArtworkFilterOptions/ColorsOptions.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+const SWATCHES_PER_ROW = 4
+
+interface ColorsOptionsScreenProps extends StackScreenProps<FilterModalNavigationStack, "ColorsOptionsScreen"> {}
+
+export const ColorsOptionsScreen: React.FC<ColorsOptionsScreenProps> = ({ navigation }) => {
+  const { layout, Sentinel } = useLayout()
+
+  // Pull out aggregations for 'color' not 'colors'
+  const { aggregation } = useArtworkFiltersAggregation({
```
dzucconi, line 45: It reeeeeally seems like we should wise up and always accept multiple param values in all filters. I don't see why Metaphysics can't accept a union of a string/array of strings value for these without breaking anything?

src/lib/Components/ArtworkFilterOptions/useMultiSelect.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  useEffect(() => {
+    if (didMountRef.current) {
+      dispatch({
+        type: "selectFilters",
+        payload: {
+          paramName,
+          paramValue: nextOptions,
+          displayText: "ignore",
```
dzucconi, line 61: This was intentional but now I'm realizing the obvious: we could use this to return a reasonable display value (in this case it's not going to help much since we probably want to emit color swatches, but for other multi-select values it'd be easy to avoid the humanization of slugs with this.

src/lib/Components/ArtworkFilterOptions/ColorsSwatch.tsx
```diff
@@ -0,0 +1,44 @@
...
+  name,
+  backgroundColor,
+  foregroundColor,
+  selected,
+  ...rest
+}) => {
+  return (
+    <Touchable {...rest}>
+      <Box width={width} py={1}>
+        <Box position="relative" mx="auto" width={34} height={34} borderRadius={99} bg={backgroundColor}>
```
anandaroop, line 23: **nonblocking:** O... nbd but why 99?

src/lib/Scenes/Partner/Components/PartnerArtwork.tsx
```diff
@@ -114,6 +116,7 @@ export const PartnerArtworkFragmentContainer = createPaginationContainer(
       return props.partner && props.partner.artworks
     },
     getVariables(props, { count, cursor }, fragmentVariables) {
+      console.log({ fragmentVariables })
```
anandaroop, line 119: Intentional?

src/lib/Components/ArtworkFilterOptions/ColorsOptions.tsx
anandaroop: So — whether a widget is a single-valued / radio input -vs- a multi-valued / checkbox input would just be an implementation detail of the client, and their selections would always be serialized as `…&foo[]=bar…` even if `foo` is just a one-element array. Is that what you're picturing?

anandaroop (approved):

src/lib/Components/ArtworkFilterOptions/ColorsSwatch.tsx
dzucconi: Oh! Yeah this is super annoying: react-native doesn't support percentages. And when I was playing with the sizing of this I just set it to an arbitrarily large value since it never goes over 50% anyway. Will just replace this with 17 now.

src/lib/Components/ArtworkFilterOptions/ColorsOptions.tsx
dzucconi: Yeah basically. Though we don't need to serialize anything. In GraphQL any array input automatically supports a single value non array input: `colors: ["red"]` and `colors: "red"` are equivalent. But ultimately I don't know if we support arrays universally for all filter types in Gravity.

iskounen (approved):
LGTM 🎉 I'm excited to use your new useMultiSelect in the time period filter.

src/lib/Components/ArtworkFilterOptions/useMultiSelect.ts
```diff
@@ -0,0 +1,75 @@
...
+        return String(option.paramValue)
+    }
+  }
+
+  const handleSelect = useCallback(
+    (option: FilterData, updatedValue: boolean) => {
+      if (updatedValue) {
+        setNextOptions((prev) => {
+          return [...prev, getParamValue(option)]
+        })
```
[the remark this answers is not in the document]
dzucconi: In this case there's no All option — All is just nothing selected. After talking with Tricia, it seems like we're going to switch from having all as an explicit option to just having a persistent clear button for all filters to handle this in a uniform manner.

src/lib/Components/ArtworkFilterOptions/useMultiSelect.ts
```diff
@@ -88,7 +91,8 @@ export const useMultiSelect = ({ options, paramName }: { options: FilterData[];
         payload: {
           paramName,
           // Label which appears on the filter overview screen
-          displayText: nextOptions.map(({ displayText }) => displayText).join(", "),
+          displayText:
+            nextOptions.map(({ displayText }) => displayText).join(", ") || defaultFilter?.displayText || "All",
```
dzucconi, line 95: Restores the default display text now instead of being blank

dblandin (approved):

PR #4535, opened 2021-03-16
fix: android consign photos action sheet

brainbicycle:

This PR resolves [CX-1154]

### Description

Our current method for showing action sheets is iOS only, this PR replaces that dependency with `@expo/react-native-action-sheet` which is cross platform and has the same API as the old `ActionSheetIOS` dependency. It requires wrapping our app in an `ActionSheetProvider` object. 

#### Screenshots

**Android**

https://user-images.githubusercontent.com/49686530/111371631-11eaed00-8670-11eb-9fb0-c48723021469.mp4

**iOS**

- UX is the same as previously

[CX-1154]: https://artsyproduct.atlassian.net/browse/CX-1154


src/lib/Scenes/Consignments/Screens/Overview.tsx
```diff
@@ -312,3 +314,6 @@ export default class Overview extends React.Component<Props, State> {
     )
   }
 }
+
+const ConnectedOverview = connectActionSheet<Props & State>(Overview)
+export default ConnectedOverview
```
ds300, line 319: nitpick: let's avoid default exports. i think the reasons why are well documented and they've definitely cost me a fair amount of time over the years https://www.google.com/search?q=javascript+why+default+export+is+bad

ds300 (approved):
great work! 👍🏼

MounirDhahri (approved):

MounirDhahri:
📦

PR #4537, opened 2021-03-16
enhancement(DX): add easy "Quick Access" info

pvinis:

### Description

Often I need to find some `internalID` or some specific identifier or info for a screen. Usually it's either for:
- getting the info so I can query MP in the separate playground application, or
- simply for debugging.

I added a simple way to add panels for "Quick Access" info.

My idea is that eventually we could have this work similarly to the analytics debug banner, where each screen/list item can show the basic info we might need, in order to debug or query in MP or web etc.

This is how it looks, for my use today:
- Auction results list items, they all have their id.
Screenshot 2021-03-16 at 20 58 35
- Auction results screens, they all have their id and bought in and cents. (I only wanted id, but I added the other two as examples.)
Screenshot 2021-03-16 at 20 58 43

To see them, just enable the flag "Show Quick Access Info"! When you tap on a value, it will be copied to the clipboard! That's how I use it, I find the item I need, tap, and I have the id. Then I go to the graphql playground and paste the id and boom. (The simulator uses the same clipboard as macos, so that's easy, and I have this enabled for my phone and mac as well, so everything is supereasy to copy and paste from/to anywhere.)

To use them:
- Make a component called `QAInfo`
- Use `QAInfoPanel`
- List the things you want to show
- Tell the panel where to be
- Render it

For example, in the auctions result screen, I have `<QAInfoPanel position="absolute" top={0} left={95}>` to that it shows up at the top, below the title. It doesn't have to be super accurate, since this is just for us devs. Usually `position="absolute"` is all you need, plus some basic `top`/`bottom` and `left`/`right`.
Then `info` is a list of `[name, value]: [string, string]` tuples.

If we need a bit more control, we can use the `QAInfoManualPanel` which is just a Flex + the flag check, and then we can render whatever we want in there, or use the `QAInfoRow` to reuse the text rendering + clipboard functionality, like `<QAInfoRow i={{theKey: "theValue"}} />`.

### DIFF
--- a/src/lib/utils/QAInfo.tsx
+++ b/src/lib/utils/QAInfo.tsx
@@ -45,7 +45,7 @@ export const QAInfoRow: React.FC<{ i: { [key: string]: string } }> = ({ i }) => 
           toast.show("Copied", "middle")
         }}
       >
-        <Text>{value}</Text>
+        <Text fontWeight="bold">{value}</Text>
       </Touchable>
     </Flex>
   )
--- a/src/lib/utils/QAInfo.tsx
+++ b/src/lib/utils/QAInfo.tsx
@@ -1,0 +1,37 @@
+import Clipboard from "@react-native-community/clipboard"
+import { useToast } from "lib/Components/Toast/toastHook"
+import { useFeatureFlag } from "lib/store/GlobalStore"
+import { color, Flex, FlexProps, Text, Touchable } from "palette"
+import React from "react"
+
+export const QAInfoFlex: React.FC<FlexProps> = (props) => {
+  const enabled = useFeatureFlag("ARShowQuickAccessInfo")
+  if (!enabled) {
+    return null
+  }
+
+  return <Flex borderColor="purple100" borderWidth={1} {...props} />
+}
+
+export const QAInfoRow: React.FC<{ i: { [key: string]: string } }> = ({ i }) => {
+  const toast = useToast()
+
+  const key = Object.keys(i)[0]
+  const value = i[key]
+
+  return (
+    <Flex flexDirection="row">
+      <Text>{key}: </Text>
+      <Touchable
+        underlayColor={color("black5")}
+        haptic
+        onPress={() => {
+          Clipboard.setString(value)
+          toast.show("Copied", "middle")
+        }}
+      >
+        <Text>{value}</Text>
+      </Touchable>
+    </Flex>
+  )
+}
--- a/src/lib/utils/QAInfo.tsx
+++ b/src/lib/utils/QAInfo.tsx
@@ -4,7 +4,7 @@ import { useFeatureFlag } from "lib/store/GlobalStore"
 import { color, Flex, FlexProps, Text, Touchable } from "palette"
 import React from "react"
 
-export const QAInfoFlex: React.FC<FlexProps> = (props) => {
+export const QAInfoPanel: React.FC<FlexProps> = (props) => {
   const enabled = useFeatureFlag("ARShowQuickAccessInfo")
   if (!enabled) {
     return null
--- a/src/lib/utils/QAInfo.tsx
+++ b/src/lib/utils/QAInfo.tsx
@@ -4,7 +4,22 @@ import { useFeatureFlag } from "lib/store/GlobalStore"
 import { color, Flex, FlexProps, Text, Touchable } from "palette"
 import React from "react"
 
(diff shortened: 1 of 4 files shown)


brainbicycle:
👀 Ooo, I like this idea 👍

brainbicycle (approved):

ds300 (approved):
yeah this is awesome 👍🏼 

I think I would prefer if it's moved to the 'tools' section of the admin menu, but I won't block merging over that.

src/lib/utils/QAInfo.tsx
```diff
@@ -0,0 +1,52 @@
...
+export const QAInfoPanel: React.FC<Omit<FlexProps, "children"> & { info: Array<[string, string]> }> = (props) => {
+  const enabled = useFeatureFlag("ARShowQuickAccessInfo")
+  if (!enabled) {
+    return null
+  }
+
+  return (
+    <QAInfoManualPanel {...props}>
+      {props.info.map(([key, value]) => (
+        <QAInfoRow i={{ [key]: value }} />
```
ds300, line 16: doesn't this show the 'childen in an array must have a unique key' warning?

pvinis:
I can do this in a few min and we merge after.

src/lib/utils/QAInfo.tsx
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  return <Flex borderColor="purple100" borderWidth={1} {...props} />
+}
+
+export const QAInfoRow: React.FC<{ i: { [key: string]: string } }> = ({ i }) => {
+  const toast = useToast()
+
+  const key = Object.keys(i)[0]
+  const value = i[key]
```
ds300, line 35: nonblocking: this is an odd way to pass the key and value in. why not just two separate props like `name` and `value`?

pvinis: I kinda wanted them to be close to each other, and less extra typing. So I thought `i=` is the smallest I can do.

src/lib/utils/QAInfo.tsx
pvinis: Ah true. fixing this and moving them and ready.

src/lib/utils/QAInfo.tsx
ds300: again nonblocking but I think readability is more important than writeability. I mean, if we measure how many keystrokes it saves vs how many minutes it costs developers to figure out what `i` means and how to use it, I think it's more valuable to have readable, unsurprising prop names.

pvinis:
Screenshot 2021-03-17 at 16 27 30
Now in the tools. This PR should be good to go, I'll wait for last comments from @ds300 and then I'll merge. Thanks!

ds300 (approved):
Nice! I had some comments on the tool refactor, but this is looking great. I can't wait to use it!

src/lib/utils/AdminMenu.tsx
```diff
@@ -129,16 +136,18 @@ const Buttons: React.FC<{ onClose(): void }> = ({ onClose }) => {
...
   const valText = currentValue ? "Yes" : "No"
+  const description = isTool(flagKey) ? tools[flagKey].description : features[flagKey].description ?? flagKey
+
   return (
     <MenuItem
-      title={features[flagKey].description ?? flagKey}
+      title={description}
       onPress={() => {
-        Alert.alert(features[flagKey].description ?? flagKey, undefined, [
+        Alert.alert(description, undefined, [
```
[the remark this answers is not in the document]
pvinis: I think I'll keep it the same for now. I don't know how we could have them `true` by default, but I could see us having something like `DTShowImages` that is true by default and in the dev toggles we could turn it off. Of course we could have `DTDontShowImages` and flip the behaviour, but naming without negatives is easier. 🤔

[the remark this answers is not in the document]
ds300: Fair enough 👍🏼  For what it's worth I think it's normally better to optimize for today's use-case then tomorrow's hypothetical use-case. But at the end of the day people can still flip this toggle even if the UX is a little wonky, so it doesn't matter much.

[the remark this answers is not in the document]
pvinis: Actually, I'll make a new thing.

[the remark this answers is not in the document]
pvinis: Done.

pvinis:
Thanks for taking a good look to both! #squashongreen

PR #4540, opened 2021-03-17
fix: modal presentation on iOS

ds300:
This PR contributes towards [CX-436]

### Description

This is, counterintuitively, the first step towards making react web views work properly. I'm moving slow and not breaking things 🐌 

This also fixes an issue where the bottom safe area was not being avoided in most of our modals.

The problem is that we were adding safe area insets in both react and native code based on different criteria, and we don't need to be doing any of the native-side stuff. So I removed it and then handled a couple of edge cases:

- not adding bottom safe area on Main, since it's added in the bottom tabs navigator
- fixing back button position in modals.

I've QAd all the modal screens and they either got better or didn't change with one exception: the consignments modal's back button on iOS 12, which will be fixed by a ticket we have in our sprint anyway.

Here's a selection of screenshots

#### Pushing an artist view into a modal's inner stack (this is very hard to achieve right now without cheating, but will happen more readily after we have web view link interception)
image

#### The consignments flow (back button on iOS 12 got moved a bit, but will be replaced this sprint anyway)
image

#### Consignments artist input w/o keyboard

image

#### Consignments artist input with keyboard (this inconsistent padding is due to naive use of KeyboardAvoidingView, will add a ticket as a follow up to fix)

image

#### Auction registration flow
image

#### Live auction view
image

[CX-436]: https://artsyproduct.atlassian.net/browse/CX-436


Artsy/App/ARModalViewController.m
```diff
@@ -0,0 +1,21 @@
...
+    _stack = stack;
+    return self;
+}
+
+- (void)viewDidLoad
+{
+    [super viewDidLoad];
+    [self addChildViewController:self.stack];
+    [self.view addSubview:self.stack.view];
+}
```
brainbicycle, line 19: nice much simpler

brainbicycle (approved):

brainbicycle:
Looks good to me! merge when ready!

PR #4541, opened 2021-03-17
fix: don't clear launch count on logout

brainbicycle:

This PR resolves [CX-1223]

### Description

We are overly aggressively clearing out all user defaults on logout, including launchCount. This is causing the first install analytics event to be fired more than once per device install. This PR saves the value and rewrites it to defaults on logout. This has been broken for a while we were just made aware of it recently due to the staging issue causing many logouts.

[CX-1223]: https://artsyproduct.atlassian.net/browse/CX-1223

### DIFF
--- a/Artsy/Constants/ARDefaults.m
+++ b/Artsy/Constants/ARDefaults.m
@@ -1,6 +1,7 @@
 #import "ARDefaults+SiteFeatures.h"
 #import "ARDefaults.h"
 #import "AROptions.h"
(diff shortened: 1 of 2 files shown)


pvinis:
Are we doing the right thing in android too? 🤔 Just so that we don't have to check it later.

ds300 (approved):
👍🏼 @pvinis The android one is not cleared on logout.

PR #4545, opened 2021-03-19
add analytics tracking for make offer and view offer buttons

xtina-starr:

This PR resolves [PURCHASE-2492]

### Description

This adds analytics tracking for the Make Offer and View Offer buttons.

[PURCHASE-2492]: https://artsyproduct.atlassian.net/browse/PURCHASE-2492


xtina-starr:
Should I be checking in `Podfile.lock`?

iskounen:
> Should I be checking in Podfile.lock?

I haven't been, but I'm curious too. I keep needing to run `pod install --repo-update` for the app to build. @MounirDhahri do you know what the best-practice is for this?

MounirDhahri:
It's fine if you check in `Podfile.lock`.

> Since we are making sure that everyone is using the same version of `cocoapods` using the Gemfile we know for sure that it's going to change only when necessary.

I do however recommend spending some time to just test if whatever got updated did not break something in the app, just like updating any regular npm package.

MounirDhahri:
You could as well create a PR with only the pod changes like I did here https://github.com/artsy/eigen/pull/4547

iskounen:
Oh, I see. Thanks for sharing the best-practice, Mounir 🙌

lilyfromseattle (approved):
🎉

xtina-starr:
> You could as well create a PR with only the pod changes like I did here #4547

Ahh this is good to know. Thanks all. 

@erikdstock you okay with merging this?

src/lib/Scenes/Inbox/Components/Conversations/OpenInquiryModalButton.tsx
```diff
@@ -15,12 +17,15 @@ const ShadowSeparator = styled(Separator)`
...
 export const OpenInquiryModalButton: React.FC<OpenInquiryModalButtonProps> = ({ artworkID, conversationID }) => {
+  const { trackEvent } = useTracking()
+
   return (
     <>
       <ShadowSeparator />
       <Flex p={1.5}>
         <Button
           onPress={() => {
+            trackEvent(tappedMakeOffer(conversationID as string))
```
erikdstock, line 28: #minor smell to me that `conversationID` isn't guaranteed to be available in this button. I don't think it needs to be a part of this PR but assuming it is nullable i wonder if we couldn't just assert it is present `(!)` higher up the tree. Again doesn't have to be a part of this PR.

erikdstock (approved):
Seems good here!

PR #4548, opened 2021-03-22
feat: add splash screen to android

MounirDhahri:

This PR resolves [CX-]

### Description
- Add splash screen to android
- Add native method to set the navigation bar color

RFC for **react-native-bootsplash**: https://github.com/artsy/eigen/issues/4549

https://user-images.githubusercontent.com/11945712/111991222-5b29b980-8b14-11eb-8cf5-8f923779df65.mov

Testing deep links again since i moved the intent to the splash activity

https://user-images.githubusercontent.com/11945712/111991239-5ebd4080-8b14-11eb-924c-59b0459efb24.mov


android/app/src/main/java/net/artsy/app/MainActivity.java
```diff
@@ -36,19 +38,6 @@ protected ReactRootView createRootView() {
...
-      decorView.setOnApplyWindowInsetsListener(
-        (v, insets) -> {
-          WindowInsets defaultInsets = v.onApplyWindowInsets(insets);
-          return defaultInsets.replaceSystemWindowInsets(
-                  defaultInsets.getSystemWindowInsetLeft(),
-                  0,
-                  defaultInsets.getSystemWindowInsetRight(),
-                  defaultInsets.getSystemWindowInsetBottom());
-        });
-    }
```
MounirDhahri, line 52: we only need this styling on the MainActivity when the splash screen hides, so i had to move it to a method that we call from the native side

android/app/src/main/java/net/artsy/app/MainActivity.java
```diff
@@ -36,19 +38,6 @@ protected ReactRootView createRootView() {
...
-      View decorView = getWindow().getDecorView();
-      decorView.setOnApplyWindowInsetsListener(
-        (v, insets) -> {
-          WindowInsets defaultInsets = v.onApplyWindowInsets(insets);
-          return defaultInsets.replaceSystemWindowInsets(
-                  defaultInsets.getSystemWindowInsetLeft(),
-                  0,
-                  defaultInsets.getSystemWindowInsetRight(),
-                  defaultInsets.getSystemWindowInsetBottom());
-        });
```
MounirDhahri, line 51: This part of the code was causing the UI to jump to fill the entire screen, so I made the app full screen globally using `styles.xml` and `decorView.setSystemUiVisibility`

package.json
```diff
@@ -104,6 +104,7 @@
     "query-string": "4.3.4",
     "react": "16.13.1",
     "react-native": "0.63.3",
+    "react-native-bootsplash": "https://github.com/artsy/react-native-bootsplash",
```
MounirDhahri, line 107: I am using our own fork from `react-native-bootsplash` which has no iOS stuff in it.

ds300: does something break if we use the official package?

MounirDhahri: no not at all

ds300 (approved):
Nice, this lib seems to work really nicely, good choice 🙇🏼 

My one concern here is that it looks odd to have the navigation bar switching from black to white instantly while the rest of the screen fades in nicely. Is there any way to fix that? If it's not easy this is fine for an MVP I think.

Finally, it looks like the commit history is messed up, I'll take another look at the diff once that's fixed but from what I saw yesterday it should be ok.

MounirDhahri:
Thanks for checking this out, GitHub messed up the commit history somehow. I will also check the navigation bar issue as well

MounirDhahri:
I spent a while the last few days to get the bottom navigation bar to be part of the fade animation without success, so I removed the fade animation for now. I also added a timeout of 500ms to make sure that the UI finished drawing behind the splash screen so that when we hide the splash screen, the MainActivity UI is already there

https://user-images.githubusercontent.com/11945712/112846834-c2acaf80-90a6-11eb-90bc-c93d9ff0202e.mov

Without timeout

https://user-images.githubusercontent.com/11945712/112846999-e8d24f80-90a6-11eb-918a-efbdcca575fa.mov

PR #4558, opened 2021-03-25
fix: fix os version check in modal header

brainbicycle:

This PR resolves [CX-1268]

### Description

Our os version check in the consignments flow was faulty and trying to coerce a version string "14.4.1" into a number which resulted in NaN. This uses an osMajorVersion util instead that parse the version correctly, also moved this to platfromUtil file because it feels like it fits there a bit better.

### Screenshots

consignments

[CX-1268]: https://artsyproduct.atlassian.net/browse/CX-1268

### DIFF
--- a/src/lib/utils/__tests__/platformUtil-tests.ts
+++ b/src/lib/utils/__tests__/platformUtil-tests.ts
@@ -1,16 +1,33 @@
+jest.mock("react-native", () => ({
+  Platform: {
+    isPad: true,
+    OS: "ios",
+  },
+}))
+
+jest.mock("react-native-safe-area-context", () => ({
+  // workaround import issue in tests
+  useSafeAreaInsets: jest.fn(),
+}))
+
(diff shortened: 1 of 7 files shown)


src/lib/utils/__tests__/platformUtil-tests.ts
```diff
@@ -0,0 +1,16 @@
+import { Platform } from "react-native"
+import { osMajorVersion } from "../platformUtil"
+
+describe(osMajorVersion, () => {
+  it("returns the correct version when version is a string", () => {
+    ;(Platform as any).Version = "12"
```
bhoggard, line 6: Should we have a test for strings like "12.2.1" here instead/also?

brainbicycle: Very good point 👍 I'll add

MounirDhahri:
nice catch!

PR #4563, opened 2021-03-29
enhancement(json files and tests): rework how we load json files and how we test with them

pvinis:

This PR resolves [CX-]

### Description

After our last two KSs, this is the end result.

In short, importing json files makes some things hard to test, like changing values in them. We now use loader functions for each json file (just app.json and EchoNew.json for now). This way we can test the killed versions for example, by "changing" the app.json version of the app inside each test. Way more control over our tests' setup.


src/lib/store/__tests__/GlobalStoreModel-tests.ts
```diff
@@ -27,6 +28,65 @@ describe("GlobalStoreModel", () => {
 
     expect(__globalStoreTestUtils__?.getCurrentState().search.recentSearches[0].props.displayLabel).toEqual("Banksy")
   })
+
+  it("can be manipulated", () => {
+    GlobalStore.actions.manipulate((store) => {
```
MounirDhahri, line 33: I think we can also use `__injectState__` here instead of manipulate

src/lib/store/config/EchoModel.tsx
```diff
@@ -1,13 +1,24 @@
...
+  id: number // 1
+  name: string // "eigen"
+  created_at: string // iso string date
+  updated_at: string // iso string date
+  features: Array<{ name: string; value: boolean }>
+  messages: Array<{ name: string; content: string }>
+  routes: Array<{ name: string; path: string }>
+  killedVersions: {
+    ios: { [version: string]: { message: string } | undefined }
+    android: { [version: string]: { message: string } | undefined }
```
MounirDhahri, line 19: I don't this is the right type for `killedVersions` since we will need an object with a message always for a killed version

src/lib/store/config/__tests__/EchoModel-tests.tsx
```diff
@@ -1,145 +1,169 @@
...
-          },
-          android: {
-            "1.9.0": {
-              message: "Please update the app",
-            },
-          },
-        },
-      }
-
-      GlobalStore.actions.rehydrate({ config: { echo: { state: validVersionEcho } } })
```
MounirDhahri, line 91: I think we still need the rehydrate to make sure that with a valid echo we will be updating our local version of echo

src/lib/store/__tests__/GlobalStoreModel-tests.ts
MounirDhahri: we maybe don't need `__manipulate__`

scripts/update-echo
```diff
@@ -4,3 +4,8 @@ set -euxo pipefail
 
 curl https://echo.artsy.net/Echo.json > Artsy/App/EchoNew.json
 yarn prettier -w Artsy/App/EchoNew.json
+
+
+# Update an echo test file
+updatedAt=$(jq -r '.updated_at' ./Artsy/App/EchoNew.json)
+perl -i -pe "s/(moment\(\").*(\"\) \/\/ Updated by)/\${1}${updatedAt}\${2}/g" ./src/lib/store/config/__tests__/EchoModel-tests.tsx
```
ds300, line 11: hmm is `jq` universally available? do we have it in our various CI images?

I only ask because i kinda remember having to replace a `jq` usage with a `node -e` usage one time.

e.g. in this case it'd be `node -e 'console.log(require("./Artsy/App/EchoNew.json").updated_at)'`

but it might be fine these days.

MounirDhahri:
Thanks for following up on this! I left some comments open for discussion and would be interested that should be quick to fix

src/lib/store/__tests__/GlobalStoreModel-tests.ts
pvinis: I tried inject first. it also uses deepmerge so it can't delete things :(

src/lib/store/GlobalStore.tsx
```diff
@@ -39,6 +39,9 @@ function createGlobalStore() {
     ;(GlobalStoreModel as any).__injectState = action((state, injectedState) => {
       assignDeep(state, injectedState)
     })
+    ;(GlobalStoreModel as any).__manipulateState = action((state, theEdits) => {
+      theEdits(state)
+    })
   }
```
ds300, line 45: nice 👍🏼

src/lib/store/config/__tests__/EchoModel-tests.tsx
pvinis: I'm testing this on line 165 👍

src/lib/utils/jsonFiles.ts
```diff
@@ -0,0 +1,6 @@
+import { Echo } from "lib/store/config/EchoModel"
+import appJsonFile from "../../../app.json"
+import echoLaunchJsonFile from "../../../Artsy/App/EchoNew.json"
+
+export const appJson = () => appJsonFile
+export const echoLaunchJson = (): Echo => echoLaunchJsonFile
```
MounirDhahri, line 6: nice!

scripts/update-echo
pvinis: oh that's a good point. I'll check that.

src/lib/store/config/__tests__/EchoModel-tests.tsx
```diff
@@ -1,145 +1,169 @@
...
-  ...jest.requireActual("../../../../../Artsy/App/EchoNew.json"),
-  updated_at: "2021-02-04T11:10:33.768Z",
-}))
+import * as loads from "lib/utils/jsonFiles"
+const appJsonSpy = jest.spyOn(loads, "appJson")
 
-jest.mock("../../../../../app.json", () => ({
-  version: "2.0.0",
-}))
+const originalDate = moment("2021-03-29T10:41:42.845Z") // Updated by `update-echo`, and is in sync with `EchoNew.json`.
```
ds300, line 10: why does this need to be kept in-sync with the actual json file? Feels like one of those things that could easily cause merge conflicts etc.

pvinis: yea.. that's why I made jq thing to update this file, to make it easier. I guess now we could skip the syncing, but I kept the tests that test if the date is older etc to not update the config. But now they use this date. there is the originalDate, laterDate etc. 🤔

scripts/update-echo
pvinis: `jq` is indeed preinstalled in circle, so all good!

src/lib/store/config/EchoModel.tsx
pvinis: You are correct, good catch. To make it clear for me too, I made this:

Screenshot 2021-03-29 at 17 02 22

I see here that killed1, killed3 and killed4 are invalid, and killed2 and killed5 are valid, as they should.

I'll fix it.

src/lib/store/config/__tests__/EchoModel-tests.tsx
pvinis: Done. Detached 🙌.

src/lib/Scenes/BottomTabs/BottomTabsModel.ts
```diff
@@ -21,7 +21,7 @@ export interface BottomTabsModel {
   setTabProps: Action<BottomTabsModel, { tab: BottomTabType; props: object | undefined }>
 }
 
-export const BottomTabsModel: BottomTabsModel = {
+export const getBottomTabsModel = (): BottomTabsModel => ({
```
pvinis, line 24: All models are not a function that return the actual model.

src/lib/store/config/EchoModel.tsx
```diff
@@ -20,10 +31,10 @@ export interface EchoModel {
   forceUpdateMessage: Computed<EchoModel, string | undefined>
 }
 
-export const EchoModel: EchoModel = {
-  state: echoLaunchJSON,
-  setEchoState: action((state, echoJSON) => {
-    state.state = echoJSON
+export const getEchoModel = (): EchoModel => ({
+  state: echoLaunchJson(),
```
pvinis, line 35: This was the reason for turning all models to funcs. With a func, now this will respect the mocks during tests, so it will load the json file we want and mock. Before it would evaluate during import of this file, which is not what we want when we test.

src/lib/utils/jsonFiles.ts
pvinis, line 6: Instead of importing the json files directly, we now have these "json loader" functions for each file, so we can mock them in tests.

src/lib/store/config/EchoModel.tsx
```diff
@@ -63,31 +74,33 @@ export const EchoModel: EchoModel = {
...
-    const minimumRequiredVersion = state.state.messages.find((message) => {
-      if (Platform.OS === "ios") {
-        return message.name === "KillSwitchBuildMinimum"
-      }
-      if (Platform.OS === "android") {
-        return message.name === "KillSwitchBuildMinimumAndroid"
-      }
-    })?.content
+    const minimumRequiredVersion = state.state.messages
+      .filter((message) => message.name.startsWith("KillSwitchBuildMinimum"))
```
pvinis, line 90: Just adding a tiny filter here to make it skip checking all the other messages in the echo config.

src/lib/store/config/EchoModel.tsx
```diff
@@ -1,13 +1,24 @@
...
 import moment from "moment-timezone"
 import { Platform } from "react-native"
 import { lt as lessThan } from "semver"
-import appJson from "../../../../app.json"
-import echoLaunchJSON from "../../../../Artsy/App/EchoNew.json"
 import { unsafe_getDevToggle } from "../GlobalStore"
 import { GlobalStoreModel } from "../GlobalStoreModel"
 
-type Echo = typeof echoLaunchJSON
+export interface Echo {
```
pvinis, line 9: 🌟 Now with actual types, fancy fancy 🌟

src/lib/store/config/__tests__/EchoModel-tests.tsx
```diff
@@ -26,120 +42,138 @@ const forceUpdateMessage = () => {
...
+        killedVersions: { ios: {}, android: {} },
+      })
+      GlobalStore.actions.__manipulate((state) => {
+        state.config.echo.state = echoConfig
+      })
+      expect(forceUpdateMessage()).toEqual("New app version required. Please update your Artsy app to continue.")
+    })
+
+    it("when the version number is exactly the minumum required version", () => {
+      appJsonSpy.mockReturnValue({ version: "1.5.0" })
```
pvinis, line 116: here we can mock the json files!

src/lib/store/GlobalStore.tsx
```diff
@@ -35,13 +35,7 @@ function createGlobalStore() {
     })
   }
 
-  if (__TEST__) {
-    ;(GlobalStoreModel as any).__injectState = action((state, injectedState) => {
```
pvinis, line 39: Moved this in the actual model. Since now its a function, here is not possible, and also in the model it seems more consistent.

src/lib/store/GlobalStore.tsx
```diff
@@ -92,19 +86,25 @@ const hooks = createTypedHooks<GlobalStoreModel>()
...
+  return <StoreProvider store={globalStoreInstance()}>{children}</StoreProvider>
 }
 
 export function useSelectedTab() {
   return hooks.useStoreState((state) => state.bottomTabs.sessionState.selectedTab)
 }
 
-let globalStoreInstance = createGlobalStore()
+let _globalStoreInstance: ReturnType<typeof createGlobalStore> | undefined
+const globalStoreInstance = (): ReturnType<typeof createGlobalStore> => {
```
pvinis, line 102: like the models, this is a function too, that always returns the instance, like a typical singleton, which is the same as we had before, but now we don't evaluate on file import, so tests work with mocks.

src/lib/store/GlobalStoreModel.ts
```diff
@@ -75,14 +79,26 @@ export const GlobalStoreModel: GlobalStoreModel = {
...
-  auth: AuthModel,
-}
+  bottomTabs: getBottomTabsModel(),
+  search: getSearchModel(),
+  myCollection: getMyCollectionModel(),
+  config: getConfigModel(),
+  auth: getAuthModel(),
+
+  // for testing only. noop otherwise.
+  __inject: __TEST__
```
pvinis, line 92: Moved over here. I thought of having a new type like `GlobalStoreModel_Testing` that extends the normal one but it complicated things. I think the `__` prefix is enough to not accidentaly use it, and even in that case, it is a noop when __TEST__ is false.

src/lib/store/__tests__/GlobalStoreModel-tests.ts
```diff
@@ -27,6 +28,65 @@ describe("GlobalStoreModel", () => {
 
     expect(__globalStoreTestUtils__?.getCurrentState().search.recentSearches[0].props.displayLabel).toEqual("Banksy")
   })
+
+  it("can be manipulated", () => {
```
pvinis, line 32: some extra tests to make sure all the global model stuff work well.

pvinis:
This is ready. I would appreciate a re-review! I feel way nicer about this now, and I really like how the code keeps getting nicer around the store and features etc with every pass on it we do.

ds300 (approved):
great stuff 💯

pvinis:
I'll go ahead and merge this one.

PR #4566, opened 2021-03-29
fix: use artistAuctionResults instead of artistInsights for contextScreenOwnerType

MounirDhahri:

### Description
- Use `artistAuctionResults` instead of `artistInsights` for artist insights `contextScreenOwnerType`

Related cohesion changes PR: https://github.com/artsy/cohesion/pull/175

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "cheerio": "0.22.0"
   },
   "dependencies": {
-    "@artsy/cohesion": "^1.73.0",
+    "@artsy/cohesion": "^1.80.1",
     "@artsy/palette-tokens": "^1.3.1",
     "@expo/react-native-action-sheet": "^3.8.0",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -77,7 +77,7 @@
     "cheerio": "0.22.0"
   },
   "dependencies": {
-    "@artsy/cohesion": "^1.73.0",
+    "@artsy/cohesion": "^1.80.1",
```
brainbicycle, line 80: **(question, non-blocking)**: Maybe we should pin this? looks like type got changed in a .7x version causing the break?

brainbicycle (approved):

MounirDhahri:
#squashongreen

package.json
MounirDhahri: I just saw this and I think yes, we should maybe pin this to avoid the build breaking in the future.

PR #4575, opened 2021-03-30
Avoid multiple-modal stack in the inquiry checkout/offer flow

erikdstock:
Closes [PURCHASE-2515]
Currently the inquiry/make offer flow involves confirming the artwork in a modal, then completing the offer flow in another web view modal. At completion the user must X out of two modals. This PR introduces a 'replace' functionality when navigating to a modal that does the following in quick succession
- remove current view (the confirm artwork modal) from its parent.
- add new view
- thats all
Paired with @ds300 on this. We started out trying to be more clever to avoid the additional animation (put new modal up and dismiss the one behind it) but the crux of it seems to be that dismissing the one behind it includes the new modal we want on top- so we have to manually remove and add them.

cc @artsy/purchase-devs 

- [x] I have documented any follow-up work that this PR will require, or it does not require any. _does this require followup for other cases? I also added a TODO in the native code where it think there may be a bug, but am hesitant to fix it since it would affect a commonly-used path_

[PURCHASE-2515]: https://artsyproduct.atlassian.net/browse/PURCHASE-2515


src/lib/navigation/navigate.ts
```diff
@@ -7,12 +7,14 @@ import { matchRoute } from "./routes"
...
   moduleName: AppModule
+  // Whether the new view should replace the previous (modal only)
+  replace?: boolean
   props: object
 }
 
 let lastInvocation = { url: "", timestamp: 0 }
 
-export async function navigate(url: string, options: { modal?: boolean; passProps?: object } = {}) {
+export async function navigate(url: string, options: { modal?: boolean; passProps?: object; replace?: boolean } = {}) {
```
brainbicycle, line 17: **(nitpick, question)** can this **replace** have a default value of false rather than doing the nil coalescing below?

brainbicycle (approved):

brainbicycle:
Looks good to me, one question but definitely not blocking, merge when ready!

erikdstock:
#squashongreen

erikdstock:
Going to let this merge and follow up with #4586 to bring this new bit of API over to android.

PR #4576, opened 2021-03-30
upgrades to bring all relay things in the same versions

pvinis:

This PR resolves [CX-]

### Description

I was reading the docs and saw that all these packages except the relay language typescript, are released together, so they should be in the same versions.


package.json
```diff
@@ -127,7 +127,6 @@
     "react-use": "^17.1.0",
     "redux-logger": "3.0.6",
     "relay-hooks": "3.5.2",
-    "relay-runtime": "^10.0.1",
```
damassi, line 130: This not needed?

damassi (approved):
One q about removing `relay-runtime`, but otherwise LGTM 👍

package.json
pvinis: it is a dep of `react-relay`, so we might as well remove it. we don't use it directly.

damassi:
#mergeOnGreen

PR #4577, opened 2021-03-30
fix: android ci install sentry cli #trivial

brainbicycle:

### Description

Installs sentry-cli via curl on linux/android ci machines, our android beta builds are failing after addition of sentry uploads to fastfile, this adds a sentry install step to our box setup as per instructions here: https://docs.sentry.io/product/cli/installation/


brainbicycle:
Going to self merge this one 🙈 to hopefully get a beta out ahead of tomorrows QA session

PR #4580, opened 2021-03-31
chore: use radio buttons in single-select filters

iskounen:

This PR resolves [FX-2760](https://artsyproduct.atlassian.net/browse/FX-2760)

### Description

The single-select filter is being changed to show radio buttons instead of checkmarks. 

#### ~ Before ~

before

#### ~ After ~

after

#### ~ iPad ~

ipad


src/palette/elements/Radio/Radio.tsx
```diff
@@ -0,0 +1,192 @@
+import debounce from "debounce"
```
MounirDhahri, line 1: I think you can instead use use denounce from lodash like we're doing here
https://github.com/artsy/eigen/blob/134f1f5c3e6b54f717a609f4af5c96dcc7c9c528/src/lib/Scenes/MyProfile/MyProfilePushNotifications.tsx#L7

src/lib/Components/ArtworkFilterOptions/SingleSelectOption.tsx
```diff
@@ -65,28 +66,37 @@ const ListItem = ({
...
-            </Sans>
+}) => {
+  const useImprovedArtworkFilters = useFeatureFlag("ARUseImprovedArtworkFilters")
+  const selected = item.displayText === selectedOption.displayText
+
+  return (
+    <SingleSelectOptionListItemRow onPress={() => onSelect(item)}>
+      <OptionListItem>
+        <InnerOptionListItem px={withExtraPadding && item.displayText !== "All" ? 3 : 2}>
+          <Sans color="black100" size="3t">
```
dzucconi, line 77: Should use `Text` instead of `Sans`

package.json
```diff
@@ -95,6 +95,7 @@
     "@react-navigation/native": "^5.9.2",
     "@react-navigation/stack": "^5.14.2",
     "@sentry/react-native": "2.2.1",
+    "debounce": "^1.2.1",
```
dzucconi, line 98: This shouldn't be necessary (wasn't really necessary in Palette either)

src/palette/elements/Radio/Radio.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { BorderProps, borders, SizeProps, space as styledSpace, SpaceProps } from "styled-system"
+
+export interface RadioProps extends FlexProps {
+  selected?: boolean
+}
+
+export const Radio: React.FC<RadioProps> = ({ selected }) => {
+  return (
+    <>
+      <RadioButton border={1} mr={1} mt="2px" mb="-2px" selected={selected}>
```
dzucconi, line 15: These mt/mb margins are unnecessary

src/palette/elements/Radio/Radio.tsx
iskounen: I didn't need to use debounce in the end, but thanks for letting me know!

src/palette/elements/Radio/RadioDot.tsx
```diff
@@ -0,0 +1,125 @@
+import { color } from "@artsy/palette-tokens"
+import React from "react"
+import styled, { css } from "styled-components/native"
+import { Box } from "../Box"
+
+const RADIO_DOT_MODES = {
```
iskounen, line 6: We're only using the `default` mode at the moment, but I am keeping these in here so that the other modes are available if we need them.

I'm happy to remove the other modes (and props) if we want to keep the palette components lean.

src/lib/Components/ArtworkFilterOptions/SingleSelectOption.tsx
```diff
@@ -65,28 +66,37 @@ const ListItem = ({
...
-              {" "}
-              ({item.count})
-            </Sans>
+}) => {
+  const useImprovedArtworkFilters = useFeatureFlag("ARUseImprovedArtworkFilters")
+  const selected = item.displayText === selectedOption.displayText
+
+  return (
+    <SingleSelectOptionListItemRow onPress={() => onSelect(item)}>
+      <OptionListItem>
```
iskounen, line 75: @dzucconi I noticed that you wrapped the `OptionListItem` component in a `Touchable` in your PR #4596. I didn't explore that in my PR, but do you think it would be useful here?

dzucconi: I mean if we want some sort of consistent feedback on touch, then yeah. There's no depressed state to the checks so I think it makes sense to use a touchable (which also has the option of some kind of tactile feedback if we want).

This is absolute the kind of thing that should be specified in the designs; but just using the `Touchable` consistently 
as one would a `<button>` doesn't seem like a bad idea.

MounirDhahri: `Touchable` also supports a prop called `haptic` which we use to return a haptic feedback when the user presses on a button.

MounirDhahri (approved):

PR #4583, opened 2021-03-31
Pull to refresh refreshes CTA

lilyfromseattle:
# PURCHASE-2536

Existing pull to refresh in conversation should refresh cta state in addition to messages

I couldn't figure out how to test this but am open to suggestions

### DIFF
--- a/src/lib/Scenes/Inbox/Screens/Conversation.tsx
+++ b/src/lib/Scenes/Inbox/Screens/Conversation.tsx
@@ -201,6 +201,8 @@ export class Conversation extends React.Component<Props, State> {
             conversation={conversation as any}
             onDataFetching={(loading) => {
               this.setState({ fetchingData: loading })
+            }}
+            onRefresh={() => {
               this.props.relay.refetch(
                 { conversationID: conversation?.internalID },
                 null,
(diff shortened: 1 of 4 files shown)


src/lib/Scenes/Inbox/Screens/Conversation.tsx
```diff
@@ -201,6 +201,16 @@ export class Conversation extends React.Component<Props, State> {
             conversation={conversation as any}
             onDataFetching={(loading) => {
               this.setState({ fetchingData: loading })
+              this.props.relay.refetch(
```
starsirius, line 204: Nice! This is so exciting! 😍  I ran this branch locally and it worked great on pull to refresh. The only thing I noticed was when [loading the next page](https://github.com/artsy/eigen/blob/7ae0c7e8aaa37ad78b4e3ffe9388cc3a15740e27/src/lib/Scenes/Inbox/Components/Conversations/Messages.tsx#L85-L105) it becomes an infinite loop. I think as @zephraph suggested, we can introduce a new prop instead of using this one.

infinite-loop

zephraph: Yep, putting it in `onDataFetching` was a good way to test the functionality, but isn't how we want it to be long term. Having a new prop like `onPullToRefresh` or something where this logic could be injected would be better.

starsirius:
Going to merge this and let's iterate! 💪

PR #4589, opened 2021-04-01
enhancement(admin menu): hide admin menu behind 7 taps on "Version"

pvinis:

This PR resolves [CX-]

### Description

We've had this problem where we wanted to hide the admin menu from the regular user, but still have it accessible in some way in all builds of the app.

It used to be that shake on ios or 5 taps anywhere on android would open the admin menu. (I think the 5 taps is still a placeholder until we find a better way.)

After this PR, this stays the same, but you need to enable "developer mode" in the app, similar to the way android devices need to, before you develop on them.

To enable "developer mode", go to Profile > About, and tap "Version" 7 times in quick succession. You should get a toast saying it was enabled, and a prompt with how to disable it, in case a user does this accidentally.

https://user-images.githubusercontent.com/100233/113286815-9c745300-92e4-11eb-9502-bbae6dfd806b.mov

Also, you see that when you are in "developer mode", there is a one-pixel wide purple line next to "Version", so we know it's on.

Now you can shake or 5-tap to bring up the admin menu like before.

To disable "developer mode", tap "Version" 3 times (easier on the way out of that mode, since only accidental usage or instructed users would do this. We, artsy people, can handle dev mode :p.)

When it's disabled, a shake or 5-tap will not bring up the admin menu.


yarn.lock
```diff
@@ -12243,11 +12248,12 @@ react-universal-interface@^0.6.2:
...
-react-use@^17.1.0:
-  version "17.1.0"
-  resolved "https://registry.yarnpkg.com/react-use/-/react-use-17.1.0.tgz#a2db05f7cca2f09a6b90b2151cd4c620e9582fe8"
-  integrity sha512-BVVG5FWXYhfkQcGhtSYNx3wPh5bvOZieH2iq1h6/WYTDCScade1JqoM5XRHPkYXfFzAzhb8QlAtt9Suj8yAumA==
+react-use@17.2.1:
+  version "17.2.1"
+  resolved "https://registry.yarnpkg.com/react-use/-/react-use-17.2.1.tgz#c81e12544115ed049c7deba1e3bb3d977dfee9b8"
+  integrity sha512-9r51/at7/Nr/nEP4CsHz+pl800EAqhIY9R6O68m68kaWc8slDAfx1UrIedQqpsb4ImddFYb+6hF1i5Vj4u4Cnw==
   dependencies:
+    "@types/js-cookie" "^2.2.6"
```
ds300, line 12256: 👀

emission/Pod/Classes/EigenCommunications/ARNotificationsManager.m
```diff
@@ -107,6 +108,11 @@ - (void)updateState:(NSDictionary *)state
     }
 }
 
+RCT_EXPORT_METHOD(stateUpdated:(nonnull NSDictionary *)state)
+{
+    [self updateState:state];
+}
+
```
ds300, line 115: This should not exist. 

We already have two state values in this module:

- `state` for state whose source of truth is objective-c
- `reactState` for state whose source of truth is TS

This `userIsDev` value should be stored in the `reactState` value and not the `state` value. We already have a method to update it called `updateReactState` below.

emission/Pod/Classes/EigenCommunications/ARNotificationsManager.h
```diff
@@ -10,6 +10,7 @@
 + (NSString *)onboardingState;
 + (NSString *)userAgent;
 + (NSString *)deviceId;
++ (NSString *)userIsDev;
```
ds300, line 13: this should be moved to ARReactStateKey below

emission/Pod/Classes/Core/AREmission.h
```diff
@@ -31,6 +31,7 @@ extern NSString *const AREnvTest;
 
 - (void)updateState:(NSDictionary *)state;
 - (NSString *)stateStringForKey:(NSString *)stateKey;
+- (BOOL)stateBoolForKey:(NSString *)stateKey;
```
ds300, line 34: this should be called `reactStateBoolForKey`

src/lib/Scenes/About/About.tsx
```diff
@@ -1,17 +1,58 @@
...
+      updateTapCount((_) => 0)
+      GlobalStore.actions.config.setUserIsDev({
+        nextValue: !userIsDev,
+        callback: (newValue) => {
+          if (newValue) {
+            toast.show('Developer mode enabled.\nTap "Version" three times to disable it.', "bottom")
+          } else {
+            toast.show("Developer mode disabled.", "bottom")
+          }
+        },
```
ds300, line 29: nitpick: This callback option seems unnecessary. You could just put the `if` statement after calling `setUserIsDev`.

ds300 (commented):
Nice! 🙏🏼 

This was using the wrong state value in obj-c. But we don't really need this on iOS anyway, since the admin menu is already disabled in the app store release. I'd suggest just undoing the obj-c stuff.

Another suggestion I'd make would be to enable dev mode by default for anyone with @artsymail.com emails, and also when `__DEV__` is true.

emission/Pod/Classes/EigenCommunications/ARNotificationsManager.h
pvinis: aaah ok that makes sense.

src/lib/Scenes/About/About.tsx
pvinis: I had that at first, but it was called maybe before the state was changed, and I was getting the old value. I'll check again, or maybe I could use await? 🤔

pvinis:
I was about to remove the check for "@artsymail" actually. I think that's the only thing that's disabling it on iOS currently. But I do like the idea of making it `true` for these emails plus `__DEV__`.

Hm, I do like having the consistency for both, so I would prefer to keep the objc stuff (after I fix them).

yarn.lock
pvinis: Apparently that's a thing! https://github.com/streamich/react-use/pull/1859 and https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#dependencies have info. Weird and cool.

pvinis:
Reworked.

Now it's using a feature flag. I added the functionality to have a function to decide the value of `readyToRelease`.

For ios it saves the userEmail and uses that to decide if it's true or not by default. For android it uses userEmail as well, but in the auth model.

docs/using_the_beta.md
```diff
@@ -6,13 +6,16 @@ While on staging, the bottom tab bar of the app is purple. This lets us know whi
 
 ### The admin menu
 
-You can access the admin menu in two ways:
-
-- If logged in as an admin, shake your device.
-- Otherwise, go to Safari and type `artsy:///admin` into the address bar (note the triple-slash).
-
```
ds300, line 13: I think we should keep the note about `artsy:///admin` but update it to `artsy:///admin2`

src/lib/store/config/features.ts
```diff
@@ -102,6 +104,19 @@ export const features = defineFeatures({
...
+        return true
+      }
+      if (isArtsyUser(userEmail(store) ?? "")) {
+        return true
+      }
+      return false
+    },
+    description: "Is user a developer?",
+    showInAdminMenu: true,
+  },
```
ds300, line 119: I would prefer not to abuse the features flag infrastructure like this. I'm worried it will confuse people.

Can we change this to just a boolean property in the AuthModel? I think it was something like that before.

pvinis: We could, but then we can't dynamically set this depending on the email or if it's __DEV__ 🤔. Do you think it's too complex? I feel like we will end up adding this at some point, just to have the flexibility to enable or disable by default depending on something.

ds300: you could use a computed property to handle __DEV__ etc

image

pvinis: oh snap! I haven't used the computed before, I totally missed that. I'll try this now!

pvinis:
Ugh. I think now it's done. I feel like both ways, flag and state, are a bit nice and funky. Flags/devtoggles are good but like you said it becomes a bit complex to have dynamic starting value.

The config way it better in that case, but it needs four things. Also I found out that ezpz apparently doesn't save `undefined` as a starting value.
```
userIsDevFlipValue: boolean
userIsDev: Computed<this, boolean, GlobalStoreModel>
setUserIsDevFlipValue: Action<this, ConfigModel["userIsDevFlipValue"]>
onSetUserIsDevFlipValue: ThunkOn<this>
```
- `userIsDev` is the "final" value that we read. That defines if the shake or 5tap will do something or not.
- `userIsDevFlipValue` is basically to flip the starting value. The starting value is `(if __DEV__, true; if artsyEmail, true; otherwise false)`. Then we allow the 7tap on "Version" to "flip" that value, and then "flip" back. So there is no concept of `default value | on | off` like the flags. It's rather a "should be on | should be off", and then flip and flip-flip and flip-flip-flip etc.
- `setUserIsDevFlipValue` is to flip the flip value.
- `onSetUserIsDevFlipValue` is to notify the native ios code, can be removed as soon as we move the onboarding in RN.


(Also, I noticed that on iOS we can keep shaking to get more and more admin modals 😆, but that will go away soon so I didn't bother fixing it.)

src/lib/utils/general.ts
```diff
@@ -0,0 +1,4 @@
+// tslint:disable-next-line:variable-name
+export const is__DEV__ = () => __DEV__
```
ds300, line 2: nonblocking: is this needed? we just use `__DEV__` directly everywhere else, which I think is standard practice. It also allows metro to remove unused code from production bundles, which this helper might prevent.

ds300 (approved):
👍🏼 One nonblocking comment but feel free to merge when ready.

Also remember to socialise this change throughout the company since it affects people's existing workflows on iOS. (Still an option to remove the iOS native bits though, which I'd recommend, but it's your decision!)

Artsy/App/ARAppDelegate.m
```diff
@@ -336,6 +333,10 @@ - (BOOL)application:(UIApplication *)app openURL:(NSURL *)url options:(NSDiction
 
 - (void)rageShakeNotificationRecieved
 {
+    if (![[AREmission sharedInstance] reactStateBoolForKey:[ARReactStateKey userIsDev]]) {
+        return;
+    }
+
```
MounirDhahri, line 339: can we also maybe enable the rageShareNotification on the simulator to avoid having developers going to the user profile every time they want to enable a feature flag? I am not sure exactly how to do it but it seems to be possible here https://stackoverflow.com/questions/9718037/how-to-detect-app-is-running-on-simulator-or-device/9718063

docs/debugging.md
```diff
@@ -3,9 +3,10 @@
 React Native relies on the debugger that’s part of the Chrome Developer Tools. To enable it:
 
 1. Run the app in the simulator.
-2. Simulate a ‘shake’ event (<kbd>^⌘Z</kbd>), which will bring up the dev menu.
-3. Choose the ‘Debug JS Remotely’ option, which should open Chrome.
-4. In the new Chrome window, open the Developer Tools (<kbd>⌘⌥J</kbd>).
+2. Go to Profile > About, and tap "Version" 7 times quickly, to enable developer mode.
```
MounirDhahri, line 6: I added a comment above about this, and if you decide to do it. we will have to change this to be if `if you are in a real device, Go to Profile > About, and tap "Version" 7 times quickly, to enable developer mode`

src/lib/store/ConfigModel.ts
```diff
@@ -1,15 +1,46 @@
...
+  }),
+  setUserIsDevFlipValue: action((state, nextValue) => {
+    state.userIsDevFlipValue = nextValue
+  }),
+  onSetUserIsDevFlipValue: thunkOn(
+    (actions) => actions.setUserIsDevFlipValue,
+    () => {
+      LegacyNativeModules.ARNotificationsManager.reactStateUpdated(unsafe__getEnvironment())
+    }
+  ),
```
MounirDhahri, line 45: I think we should move these to a separate model same as we do for `echo` `features` and `environment`

src/lib/utils/general.ts
pvinis: Oh interesting about metro. I added this so that we can mock it, but I see I can do it without it. I'll remove this and add the tests back.

Artsy/App/ARAppDelegate.m
pvinis: __DEV__ is on in the simulator, so it should be all good.

docs/debugging.md
pvinis: True, I'll make that clearer.

Artsy/App/ARAppDelegate.m
MounirDhahri: aha cool then :)

MounirDhahri (approved):

src/lib/store/ConfigModel.ts
pvinis: I didn't want to go too deep with this and have a whole other model, but it might be good, to make it kind of "contained".

pvinis: done

pvinis:
Good call on the announcement of this. I will on #dev. I will keep the iOS part too, since it basically doesn't change anything for devs, and only "hides" the shake gesture for regular users, so it's a pretty non-side-effects situation.

pvinis:
Let's go. Thank you both!

PR #4591, opened 2021-04-01
feat: add login screen

MounirDhahri:

This PR resolves [CX-1083]

### Description
- Add login screen
- The bottom navigation bar and status bar will be worked on separately

https://user-images.githubusercontent.com/11945712/113322225-b7f75200-9314-11eb-989f-8cda1746e0d9.mov

[CX-1083]: https://artsyproduct.atlassian.net/browse/CX-1083

### DIFF
--- a/src/lib/Components/Input/Input.tsx
+++ b/src/lib/Components/Input/Input.tsx
@@ -1,11 +1,20 @@
-import { color, Color, Flex, Sans, XCircleIcon } from "palette"
+import { color, Color, EyeOpenedIcon, Flex, Sans, TEXT_FONTS, XCircleIcon } from "palette"
 import { fontFamily } from "palette/platform/fonts/fontFamily"
 import React, { useImperativeHandle, useRef, useState } from "react"
-import { TextInput, TextInputProps, TouchableOpacity, TouchableWithoutFeedback, View } from "react-native"
+import {
+  LayoutAnimation,
+  Platform,
+  TextInput,
+  TextInputProps,
+  TouchableOpacity,
+  TouchableWithoutFeedback,
+  View,
+} from "react-native"
 import styled from "styled-components/native"
+import { EyeClosedIcon } from "../../../palette/svgs/EyeClosedIcon"
 import { InputTitle } from "./InputTitle"
 
-export const INPUT_HEIGHT = 40
+export const INPUT_HEIGHT = 43
 
 export interface InputProps extends TextInputProps {
   containerStyle?: React.ComponentProps<typeof Flex>["style"]
@@ -16,6 +25,7 @@ export interface InputProps extends TextInputProps {
   required?: boolean
   title?: string
   enableClearButton?: boolean
+  canHidePassword?: boolean
   onClear?(): void
   renderLeftHandSection?(): JSX.Element
 }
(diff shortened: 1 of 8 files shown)


src/lib/Components/Input/Input.tsx
```diff
@@ -76,22 +108,31 @@ export const Input = React.forwardRef<TextInput, InputProps>(
...
+                textAlignVertical="center"
                 {...(rest as any)}
                 onChangeText={(text) => {
                   setValue(text)
                   rest.onChangeText?.(text)
                 }}
                 onFocus={(e) => {
+                  if (Platform.OS === "android") {
+                    LayoutAnimation.configureNext(LayoutAnimation.create(60, "easeInEaseOut", "opacity"))
+                  }
```
MounirDhahri, line 122: I didn't want to break anything for iOS so I only added this animation to android

src/lib/Components/Input/Input.tsx
```diff
@@ -152,8 +193,8 @@ const StyledInput = styled(TextInput)`
   left: 10;
   right: 10;
   top: 12; /* to center the text nicely */
-  bottom: -20;
-  padding-bottom: 30;
+  bottom: -30;
```
MounirDhahri, line 196: This is to allow more padding between the keyboard and the error message

src/lib/Scenes/Onboarding/OnboardingLogin.tsx
```diff
@@ -1,10 +1,155 @@
...
+      const res = await GlobalStore.actions.auth.signIn({
+        email,
+        password,
+      })
+
+      if (!res) {
+        setShowErrorMessage(true)
+      }
+    } finally {
+      passwordInputRef.current?.focus()
```
MounirDhahri, line 44: Usually if the login is wrong it's due to a wrong password, therefore I tried to focus on the password input when the login fails

src/lib/Scenes/Onboarding/OnboardingLogin.tsx
```diff
@@ -1,10 +1,155 @@
...
+    try {
+      setIsLoading(true)
+      const res = await GlobalStore.actions.auth.signIn({
+        email,
+        password,
+      })
+
+      if (!res) {
+        setShowErrorMessage(true)
+      }
```
MounirDhahri, line 42: we are intentionally showing a generic error message for security purposes.

src/palette/elements/Button/Button.tsx
```diff
@@ -180,24 +181,20 @@ export const Button: React.FC<ButtonProps> = (props) => {
...
 
-    const { black100 } = themeProps.colors
-
     return {
-      backgroundColor: black100,
-      borderColor: black100,
-      color: "rgba(0, 0, 0, 0)",
+      backgroundColor: `rgba(0, 0, 0, ${opacity})`,
+      borderColor: `rgba(0, 0, 0, 0)`,
+      color: color("white100"),
```
MounirDhahri, line 197: They way android handles opacities is different than ios.
If you set a parent component opacity to 0.5, on iOS only the parent component opacity drops and all children opacity remains the 1, however on android, the opacity gets inherited to children component and all their opacity drops to 0.5 which we don't want to have in this case because it would mean that the button text won't be visible if it's disabled

src/lib/NativeModules/ArtsyNativeModule.tsx
```diff
@@ -11,6 +11,20 @@ export const ArtsyNativeModule = {
     Platform.OS === "ios"
       ? LegacyNativeModules.ARNotificationsManager.nativeState.launchCount
       : (NativeModules.ArtsyNativeModule.getConstants().launchCount as number),
-  setAppStyling: Platform.OS === "ios" ? null : NativeModules.ArtsyNativeModule.setAppStyling,
-  setNavigationBarColor: Platform.OS === "ios" ? null : NativeModules.ArtsyNativeModule.setNavigationBarColor,
+  /**
+   * Set android app theme
+   */
+  setAppStyling: Platform.OS === "ios" ? undefined : NativeModules.ArtsyNativeModule.setAppStyling,
```
brainbicycle, line 17: **nitpick**: maybe we should have a function that just logs out unsupported function calls on iOS rather than undefined?

src/palette/elements/Button/Button.tsx
```diff
@@ -237,7 +234,8 @@ export const Button: React.FC<ButtonProps> = (props) => {
   const { children, loading, disabled, inline, longestText, ...rest } = props
   const s = getSize()
   const variantColors = getColorsForVariant(variant)
-  const opacity = props.disabled ? 0.1 : 1.0
+  // const opacity = 1
+  // const opacity = props.disabled ? 0.1 : 1.0
```
brainbicycle: can we remove?

brainbicycle (approved):

brainbicycle:
Looks good! It would be nice to remove some of the commented out code and I had a couple questions but they are non-blocking. Merge when ready.

src/lib/Components/Input/Input.tsx
```diff
@@ -36,14 +46,36 @@ export const Input = React.forwardRef<TextInput, InputProps>(
       enableClearButton,
       title,
       renderLeftHandSection,
+      secureTextEntry = false,
+      textContentType,
+      canHidePassword,
```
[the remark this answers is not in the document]
MounirDhahri: that's right, I should have deleted it.

src/lib/NativeModules/ArtsyNativeModule.tsx
MounirDhahri: that definitely makes more sense

PR #4593, opened 2021-04-01
Add missing overlay page title to MO modal and remove cancel button in modal header

lilyfromseattle:
# PURCHASE-2526

Make Offer (Initial View) - Title "Make Offer with Artsy Pay" is missing / Cancel link is in nav and should be ONLY the CTA (below "Confirm')

__Before:__
Screen Shot 2021-04-01 at 3 56 32 PM

__After:__
Screen Shot 2021-04-01 at 3 55 44 PM


jo-rs (approved):
Looks good! 🥇

PR #4595, opened 2021-04-01
fix: fix incorrect message order when content is shorter than the scroll view (PURCHASE-2535)

starsirius:

This PR resolves [PURCHASE-2535]

### Description

We noticed in some cases the messages (including order events) would be displayed in the wrong order. It turns out we were only inverting the `FlatList` when the content is taller than the scroll view, and in cases the content is shorter, the messages were displayed in the wrong order.

This fixes it by always inverting the `FlatList` and uses flexbox `justifyContent` to align content to the top when the content is shorter than the scroll view. More details inline.

**Before**

Screen Shot 2021-04-01 at 6 07 22 PM

**After**

Screen Shot 2021-04-01 at 6 08 04 PM

[PURCHASE-2535]: https://artsyproduct.atlassian.net/browse/PURCHASE-2535

### DIFF
--- a/src/lib/Scenes/Inbox/Components/Conversations/Messages.tsx
+++ b/src/lib/Scenes/Inbox/Components/Conversations/Messages.tsx
@@ -74,13 +74,6 @@ export const Messages: React.FC<Props> = forwardRef((props, ref) => {
   }, [allOrderEvents.length, allMessages.length])
 
   const flatList = useRef<FlatList>(null)
-  const [flatListHeight, setFlatListHeight] = useState(0)
-  const [contentHeight, setContentHeight] = useState(0)
-  const [shouldStickFirstMessageToTop, setShouldStickFirstMessageToTop] = useState(false)
-
-  useEffect(() => {
-    setShouldStickFirstMessageToTop(contentHeight < flatListHeight)
-  }, [contentHeight || flatListHeight])
 
   const loadMore = () => {
     if (!relay.hasMore() || relay.isLoading()) {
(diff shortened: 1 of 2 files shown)


src/lib/Scenes/Inbox/Components/Conversations/Messages.tsx
```diff
@@ -147,26 +140,17 @@ export const Messages: React.FC<Props> = forwardRef((props, ref) => {
           />
         )
       }}
-      inverted={!shouldStickFirstMessageToTop}
+      inverted={true}
```
starsirius, line 143: Looks like what we were doing was:
- inverting the `FlatList` so we displayed messages from bottom to top, and
- when there were less than one page of messages, instead of aligning messages to the bottom, we wanted to align them to the top (and hence the `shouldStickFirstMessageToTop` logic).

However, setting `inverted` based on `shouldStickFirstMessageToTop` seems wrong, and that caused the incorrect message order when we have less than 1 page of content.

This sets `inverted` to be always `true` and uses `justifyContent: "flex-end"` to achieve the alignment. And we can retire `shouldStickFirstMessageToTop` and relevant logic.

starsirius: I'm not sure how to test style changes in this case and open to suggestions!

erikdstock: i don't think style changes are worth testing in this case. LGTM!

erikdstock (approved):
Great simplification - every time i looked at this something felt a little sus but a mental block stopped me from looking any deeper or considering how this may be *not* working as intended. Good catch!

PR #4596, opened 2021-04-02
feat(fx-2761): Converts multi-select to use checkboxes

dzucconi:

This PR resolves [FX-2761]

[FX-2761]: https://artsyproduct.atlassian.net/browse/FX-2761


iskounen (approved):
Thanks for sharing this PR to help me navigate this work for single-select options. I have two (non-blocking) questions, but otherwise LGTM.

1. Why did you decide to wrap the `<OptionListItem>` in a `<Touchable>`? 
2. Is it useful to keep the `error` and `hover` props in `CheckProps` around if we're not currently using them?

dzucconi:
> Why did you decide to wrap the <OptionListItem> in a <Touchable>?
Mentioned it the other PR but: user feedback

> Is it useful to keep the error and hover props in CheckProps around if we're not currently using them?
I think so — it's part of the design system so these things are generally useful outside of just this PR, ideally

dzucconi:
Build was just failing because I was importing something from the wrong package. Should be good now.

PR #4613, opened 2021-04-08
CTA Inquiry Checkout improvements

erikdstock:
Blocked by https://github.com/artsy/metaphysics/pull/3087 (need to sync schema) (done)
See comments for screenshots.

This PR Resolves [PURCHASE-2501].

### Description
This PR makes it possible to submit multiple offers in a single conversation. Implementation notes:
- `ConversationCTA` - a new component that houses all logic around which (if any) CTA should appear.
- Querying a second order connection, `conversation.activeOrders` which filters only by active states. The presence of an active order prevents the 'make offer' button from appearing, and when there is no active order (expires, cancels etc) the button will be back. The previous `conversation.orderConnection` is still used for interlaced messages.
- `ReviewOfferButton` which renders the colored banners is now more static, unconnected from relay and mainly responds to props received from `ConversationCTA`.

There is also an animation added to the CTA when it shows and hides. There might be better ways to do this- @MounirDhahri mentioned #4619 may be a hint for a way forward.

[PURCHASE-2501]: https://artsyproduct.atlassian.net/browse/PURCHASE-2501


src/lib/Scenes/Inbox/Components/Conversations/CTAPopUp.tsx
```diff
@@ -0,0 +1,49 @@
...
+    } else {
+      setTimeout(() => setHidden(true), 225)
+    }
+    Animated.timing(animationProgress, {
+      toValue: doShow ? 0 : 1,
+      useNativeDriver: true,
+      // ease-out to blend nicely with keyboard animation
+      duration: 250,
+      easing: Easing.in(Easing.linear),
+      // easing: Easing.out(Easing.linear),
```
erikdstock, line 21: Some cleanup to do here - this component just animates the appear/disappear in a kind of hacky way. Worked a bit with @MounirDhahri on a better solution but I hit a wall and decided to punt in order to complete the business logic first.

src/lib/Scenes/Inbox/Components/Conversations/Composer.tsx
```diff
@@ -148,6 +134,7 @@ export default class Composer extends React.Component<Props, State> {
                   </Button>
                 </TouchableWithoutFeedback>
               </Container>
+              {/* <Box height={50} bg="green" /> */}
```
erikdstock, line 137: Some diff in this file is left over from when i was trying to make the CTA hide nicely. the point of this was to prevent small timing differences to make the button peek out from the gap between the text input and keyboard while sliding up and down.

src/lib/Scenes/Inbox/Components/Conversations/Composer.tsx
```diff
@@ -80,35 +77,11 @@ export default class Composer extends React.Component<Props, State> {
...
-    const conversationID = conversation.conversationID!
-
-    const firstItem = conversation?.items?.[0]?.item
-    const artwork = firstItem?.__typename === "Artwork" ? firstItem : null
-    const { artworkID, isOfferableFromInquiry } = { ...artwork }
-
     // TODO: assumption is that there will be only 0/1 active (not pending or abandoned) order
     // and we will take the first without worrying about sort/order
-    const orders = extractNodes(conversation.orderConnection)
-    const inactiveOrderStates = ["PENDING", "ABANDONED", "CANCELED"]
```
erikdstock, line 93: Lots of this logic was able to either move into `ConversationCTA` or, in this case, go away entirely because we are specifically not filtering these states.

src/lib/Scenes/Inbox/Components/Conversations/Composer.tsx
```diff
@@ -127,14 +100,27 @@ export default class Composer extends React.Component<Props, State> {
...
-              {!this.state.active && CTA}
-              <Container active={this.state.active}>
+            <Flex
+              //
+              style={
+                {
+                  // alignSelf: "flex-end",
+                  // position: "relative",
+                  // bottom: -50,
+                }
```
erikdstock, line 110: cleanup

src/lib/Scenes/Inbox/Components/Conversations/ConversationCTA.tsx
```diff
@@ -0,0 +1,93 @@
...
+      let ctaKind: ReviewOfferCTAKind | null = null
+      const { lastTransactionFailed, state, lastOffer } = activeOrder
+
+      if (lastTransactionFailed) {
+        ctaKind = "PAYMENT_FAILED"
+      } else if (state === "SUBMITTED" && lastOffer?.fromParticipant === "SELLER") {
+        ctaKind = "OFFER_RECEIVED"
+      } else if (state === "APPROVED" && lastOffer?.fromParticipant === "BUYER") {
+        ctaKind = "OFFER_ACCEPTED"
+      }
```
erikdstock, line 41: I extracted the `returnButtonMessaging` helper logic into two places:
- What kind of banner to render (if any): here , specified by the `ctaKind` string union type.
- In the simplified `ReviewOfferButton` which uses the same `kind` prop string union to select its colors. I'll just copy-paste this note there I guess.

src/lib/Scenes/Inbox/Components/Conversations/ConversationCTA.tsx
```diff
@@ -0,0 +1,93 @@
...
+      if (ctaKind) {
+        CTA = <ReviewOfferButton kind={ctaKind} activeOrder={activeOrder} conversationID={conversationID} />
+      }
+    }
+  }
+  if (!CTA) {
+    return null
+  } else {
+    return <CTAPopUp show={show}>{CTA}</CTAPopUp>
```
erikdstock, line 50: After all the logic above, we either have a CTA to show or we don't. All logic to contend over this space can live in this file (maybe in the future there is a CTA to follow the gallery you are talking to or something).

src/lib/Scenes/Inbox/Components/Conversations/ConversationCTA.tsx
```diff
@@ -0,0 +1,93 @@
...
+      items {
+        item {
+          __typename
+          ... on Artwork {
+            artworkID: internalID
+            isOfferableFromInquiry
+          }
+        }
+      }
+      activeOrders: orderConnection(first: 10, states: [APPROVED, FULFILLED, SUBMITTED, REFUNDED]) {
```
erikdstock, line 67: Using new `states` filter- effectively presence of any of these orders will prevent the user from making a new offer in this conversation.

src/lib/Scenes/Inbox/Components/Conversations/ReviewOfferButton.tsx
```diff
@@ -1,65 +1,80 @@
...
-      lastOfferFromParticipant: order.lastOffer?.fromParticipant,
-      hoursTillExpiration: hours,
-    })
+  switch (kind) {
+    case "PAYMENT_FAILED": {
+      ctaAttributes = {
+        backgroundColor: "red100",
+        message: "Payment Failed",
+        subMessage: "Please update payment method",
+        Icon: AlertCircleFillIcon,
```
erikdstock, line 42: The switch statement above handles the second bullet point here:
I extracted the `returnButtonMessaging` helper logic into two places:
- What kind of banner to render (if any): in `ConversationCTA`, specified by the `ctaKind` string union type.
- In the simplified `ReviewOfferButton` which uses the same `kind` prop string union to select its colors.

src/lib/Scenes/Inbox/Components/Conversations/ReviewOfferButton.tsx
```diff
@@ -1,65 +1,80 @@
...
+      }
+      break
+    }
+    case "OFFER_RECEIVED": {
+      ctaAttributes = {
+        backgroundColor: "copper100",
+        message: `${offerType} Received`,
+        // TODO: what about <1 hr?
+        subMessage: `Expires in ${hours}hr`,
```
erikdstock, line 51: possible QA item.

src/lib/Scenes/Inbox/Components/Conversations/__tests__/ConversationCTA-tests.tsx
```diff
@@ -0,0 +1,156 @@
...
+    })
+    const artworkItem = {
+      items: [{ item: { __typename: "Artwork", isOfferableFromInquiry: true } }],
+    }
+    const getWrapperWithOrders = (...orders: any[]) => {
+      return getWrapper({
+        Conversation: () => ({
+          ...artworkItem,
+          activeOrders: {
+            edges: orders.map((order) => ({ node: order })),
```
erikdstock, line 81: The tests here contain most logic about whether a banner will be rendered. Some tests remain in `Composer-tests`, especially specifying color, and maybe would be better here. I think it's important we have clear, readable specs for the CTA logic since we will probably be continuing to revisit it with design folks... though I guess you could say this about anything.

src/lib/Scenes/Inbox/Components/Conversations/__tests__/ReviewOfferButton-tests.tsx
```diff
@@ -1,211 +1,91 @@
...
-    const tree = renderWithWrappers(<TestRenderer />)
-    act(() => {
-      env.mock.resolveMostRecentOperation((operation) => MockPayloadGenerator.generate(operation, mockResolvers))
-    })
-    return tree
+  const getWrapper = (kind: ReviewOfferCTAKind, activeOrder: Partial<ReviewOfferButtonProps["activeOrder"]> = {}) => {
+    const props = {
+      conversationID: "conversation-id",
+      kind,
+      activeOrder: {
```
erikdstock, line 27: these are now strictly presentational tests with no relay stuff.

src/lib/Scenes/Inbox/Components/Conversations/__tests__/ReviewOfferButton-tests.tsx
```diff
@@ -1,211 +1,91 @@
...
-    })
+    const wrapper = getWrapper("OFFER_RECEIVED", { offers: { edges: [{}, {}] } })
 
     wrapper.root.findByType(TouchableWithoutFeedback).props.onPress()
 
     expect(trackEvent).toHaveBeenCalledWith({
       action: "tappedViewOffer",
       context_owner_type: "conversation",
-      impulse_conversation_id: "1234",
+      impulse_conversation_id: "conversation-id",
```
erikdstock, line 83: wish i'd structured the getWrapper function in a way i could easily pass these props in rather than rely on defaults, but i didn't and didn't bother fixing it either.

src/lib/Scenes/Inbox/Screens/Checkout.tsx
```diff
@@ -19,9 +24,11 @@ export const Checkout: React.FC<{ orderID: string; title: string }> = ({ orderID
           onLeftButtonPress={() => {
             dismissModal()
           }}
-          leftButtonText="Cancel"
-          rightButtonDisabled
-          rightButtonText=" "
+          leftButtonText={inquiryCheckout ? " " : "Cancel"}
+          // leftButtonDisabled={!inquiryCheckout} // prop doesnt exist so just hide?
```
erikdstock, line 28: i dont remember writing this.

src/lib/navigation/navigate.ts
```diff
@@ -50,6 +50,7 @@ export async function navigate(url: string, options: { modal?: boolean; passProp
     ...module.options,
   }
 
+  console.log({ screenDescriptor })
```
erikdstock, line 53: this really seems unrelated

src/lib/Scenes/Inbox/Components/Conversations/ConversationCTA.tsx
erikdstock: I want to call out that there will be no CTA rendered if there is an active order but no `ctaKind` assigned within this block, for example `APPROVED &&  last offer not from buyer`.

erikdstock (commented):
Bit of a code tour.

Podfile.lock
```diff
@@ -498,8 +498,6 @@ PODS:
     - React-Core
```
lilyfromseattle, line 498: Are these `Podfile` changes intentional?

src/lib/Scenes/Inbox/Components/Conversations/ReviewOfferButton.tsx
lilyfromseattle: Nice switch usage!

lilyfromseattle (approved):
Looks good! We should manually test this a ton

Podfile.lock
erikdstock: I don't understand where they came from and cannot make them go away if i check out this file from master and rerun `bundle exec pod install` i wind up right back here.

src/lib/Scenes/Inbox/Components/Conversations/ReviewOfferButton.tsx
```diff
@@ -104,28 +115,3 @@ export const ReviewOfferButton: React.FC<ReviewOfferButtonProps> = ({ conversati
     </TouchableWithoutFeedback>
   )
 }
```
sepans, line 117: Seems this component no longer queries for the data it needs and gets it from its parent (`ConversationCTA`). Is there a reason for that? Its all the same but nice to have each component to fetch the data they care about.

sepans (approved):
Looks great 🚀 

Nice comprehensive tests 🥇 

Just one (totally non-blocking) question about the removed `ReviewOfferButton`fragment container.

src/lib/Scenes/Inbox/Components/Conversations/ReviewOfferButton.tsx
erikdstock: My reasoning here was that at this point the CTA is purely presentational and which colors and words to render are almost entirely determined at the `ConversationCTA` level and the `kind` key passed in. Good callout though, and @lilyfromseattle's work on the payment failed stuff may change this again so I'll check in with her.

erikdstock:
# Make offer after declined offer
Simulator Screen Shot - iPhone X - 2021-04-12 at 17 47 43

# Offer accepted by seller
Simulator Screen Shot - iPhone X - 2021-04-12 at 17 30 24

# Offer fulfilled (still no CTA or make offer button)
Simulator Screen Shot - iPhone X - 2021-04-12 at 17 30 37

# Payment failed
Simulator Screen Shot - iPhone X - 2021-04-12 at 17 37 34

src/lib/Scenes/Inbox/Components/Conversations/ReviewOfferButton.tsx
sepans: 💯 Not a big deal, fetching its own data makes the component more portable.

PR #4616, opened 2021-04-08
fix: fix toast test, don't use button

brainbicycle:

### Description

We are seeing intermittent test failures in Toast component.

Worked on in knowledge share with @artsy/collector-experience it looks like the problem is not with the Toast component but with some recent changes to the Button component here: https://github.com/artsy/eigen/pull/4591, 
Only theory which is kind of vague right now is the loading style now being passed is causing button component to animate which is causing the timer recursion issue but beyond that I am still unclear exactly what is going on. I didn't want to sink too much time into investigation so I just used a different component in this test.

### DIFF
--- a/src/lib/Components/Toast/__tests__/Toast-tests.tsx
+++ b/src/lib/Components/Toast/__tests__/Toast-tests.tsx
@@ -1,6 +1,7 @@
 import { renderWithWrappers } from "lib/tests/renderWithWrappers"
-import { Button, Flex } from "palette"
+import { Touchable } from "palette"
 import React from "react"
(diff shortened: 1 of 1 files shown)


MounirDhahri:
Thanks for figuring this out!

MounirDhahri:
@pvinis I will merge this since we have few open PRs blocked because of this and need to be merged. I hope you don't mind

MounirDhahri (approved):

pvinis:
I was about to, but yea :) nice.

wow, such a funky problem.

PR #4619, opened 2021-04-09
fix: keyboard avoiding

ds300:
This PR resolves [CX-1280] [CX-1295] [CX-1281]

### Description

This PR takes care of our keyboard avoidance woes by

- Replacing `android:windowSoftInputMode="adjustPan"` with `android:windowSoftInputMode="adjustResize"` in AndroidManifest.xml

  This was our main problem on Android. `"adjustPan"` causes the main activity to scroll away when the keyboard opens, which is not how any of our keyboard avoidance interactions should work.

- Creating a two-mode version of `KeyboardAvoidingView` which handles iOS 13+ modals and requires less babysitting.

  The react-native built-in `KeyboardAvoidingView` calculates the keyboard displacement based on the assumption that the top of the view is at the top of the screen. To account for situations where this is not true it provides the `keyboardVerticalOffset` prop. Unfortunately, it can be difficult or impossible to provide an accurate value for that prop, since it relies on the view hierarchy and whether or not the view is inside a ios 13 modal. We fix this in a couple of ways:

  - for modals we add a new mode that makes the assumption that the view's bottom is at the bottom of the screen, which is always true in modals, or at least easy to make true.
  - for non-modals we use `measureInWindow` to calculate the y-offset of the view relative to the window, rather than relative to the parent view. Then you should never need to provide an explicit `keyboardVerticalOffset` value.
  
  Our custom `KeyboardAvoidingView` also takes care of the fact that android view measurement coordinates are calculated relative to the status bar bottom, rather than the top of the screen, while the keyboard measurement coordinates are calculated relative to the top of the screen. 

  All of this is to say that, it should 'just work' wherever you put it, as long as in modals it touches the bottom of the screen. If we ever build UI where that's not the case we can add an 'bottomOffset' override prop.

- MyAccountFieldEditScreen

   image

- BottomAlignedButtonWrapper (used only in CreditCardForm in registration flow)
 
  image

- BottomAlignedButton (used in the Consignments flow)

  image

- ArtsyReactWebView

  image

- BillingAddress (registration flow)

  image

- FancyModal (affects InquiryModal, MyCollectionArtworkForm)

  image

  image

- Composer

  image

- Search
 
  image

- Onboarding

  image

[CX-1280]: https://artsyproduct.atlassian.net/browse/CX-1280
[CX-1295]: https://artsyproduct.atlassian.net/browse/CX-1295
[CX-1281]: https://artsyproduct.atlassian.net/browse/CX-1281

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -11,6 +11,7 @@ upcoming:
     - Hide AdminMenu behind 7 taps of "Version" - pavlos
     - Use radio buttons for single-select filters behind feature flag - iskounen
     - Specify tracking screen owner in tab bar tracking - ole
+    - Fix keyboard avoidance - david
   user_facing:
     - Use toolbar style modal headers in consignments flow - brian
     - Add Onboarding welcome screen - mounir
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -21,7 +21,7 @@
         android:configChanges="keyboard|keyboardHidden|orientation|screenSize|uiMode"
         android:launchMode="singleTask"
         android:exported="true"
-        android:windowSoftInputMode="adjustPan">
+        android:windowSoftInputMode="adjustResize">
       </activity>
 
       <activity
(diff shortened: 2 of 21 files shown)


MounirDhahri:
This is awesome ❤️

MounirDhahri (approved):

PR #4624, opened 2021-04-12
enhancement: Set action type for tracking filter params change

olerichter00:

This PR resolves [CX-1258]

[CX-1258]: https://artsyproduct.atlassian.net/browse/CX-1258

### DIFF
--- a/src/lib/Components/FilterModal/FilterModal.tsx
+++ b/src/lib/Components/FilterModal/FilterModal.tsx
@@ -1,4 +1,4 @@
-import { ContextModule } from "@artsy/cohesion"
+import { ActionType, ContextModule } from "@artsy/cohesion"
 import { NavigationContainer } from "@react-navigation/native"
 import { createStackNavigator, StackScreenProps, TransitionPresets } from "@react-navigation/stack"
 import { useFeatureFlag } from "lib/store/GlobalStore"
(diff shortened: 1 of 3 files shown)


MounirDhahri (approved):

MounirDhahri:
LGTM! 🎉🎉🎉

mikehrom:
Nice 👍  thank you

src/lib/Components/FilterModal/FilterModal.tsx
```diff
@@ -296,6 +303,7 @@ export const FilterModalNavigator: React.FC<FilterModalProps> = (props) => {
                     break
                   case FilterModalMode.AuctionResults:
                     trackChangeFilters({
+                      actionType: ActionType.auctionResultsFilterParamsChanged,
                       screenName: PageNames.ArtistPage,
```
mikehrom, line 307: @MounirDhahri @olerichter00  I should have noticed this earlier, but was examining the new filter event in staging and saw that it's firing a screen owner type of 'Artist' but i think the screen owner type is actually 'artistAuctionResults' since we're on that Insights tab. 

I don't think this is a major blocking issue for metrics, since this is the only auction results filtering context in the Artist screen ecosystem, but something to keep in mind if we end up being able to filter auctions results on places other than the Insights tab.

olerichter00: Yeah, that makes sense. Shall I create a ticket for it or shall we just leave it as it is and try to keep in mind to change it when we touch artwork filter tracking again?

PR #4628, opened 2021-04-13
Link CTA to failed payment page

lilyfromseattle:
# PURCHASE-2547

CTA should link to {orderID}/payment/new only in case of failed payment. Also changed modal header title in this case to be "Retry Payment" as per this discussion https://artsy.slack.com/archives/C9YNS4X32/p1618337020223300

Screen Shot 2021-04-13 at 11 23 53 AM


erikdstock (approved):
Looks great, very readable!

PR #4639, opened 2021-04-15
Failed payment notification

lilyfromseattle:
# PURCHASE-2542

_From the Jira ticket:_
Failed payment displayed as "offer accepted" in inline order history

When testing a failed payment scenario, failed transaction got displayed as “Offer Accepted” in the inline history. See the screenshot below for an example: I submitted the offer with a test card that’ll fail when capturing. I accepted partner’s counteroffer 2 times (both were failed) and they were displayed as accepted in the history.

------
In this PR I am filtering out `APPROVED` state change updates when they are followed by `SUBMITTED` updates because they only occur in this order when payment fails (order goes to approved state then is reverted to submitted state when payment fails). While this solution may not be ideal, it solves the problem, and after discussing with team members I couldn't come up with another way to filter these notifications without major changes (adding a reference to the order object to `CommerceOrderStateChangedEvent` so we could access `lastPaymentFailed`)

Github is being weird when I try and attach screenshots (connectivity issue?) - will try again later

### DIFF
--- a/src/lib/Scenes/Inbox/Components/Conversations/Messages.tsx
+++ b/src/lib/Scenes/Inbox/Components/Conversations/Messages.tsx
@@ -60,9 +60,15 @@ export const Messages: React.FC<Props> = forwardRef((props, ref) => {
     .reduce<OrderEvent[]>((prev, order) => prev.concat(order.orderHistory), [])
     .map<OrderEventWithKey>((event, index) => ({ ...event, key: `event-${index}` }))
 
+  const orderEventsWithoutFailedPayment = allOrderEvents.filter((event, index) => {
+    if (!(event.state === "APPROVED" && allOrderEvents[index + 1] && allOrderEvents[index + 1].state === "SUBMITTED")) {
+      return event
+    }
+  })
+
   // Combine and group events/messages
   useEffect(() => {
-    const sortedMessages = sortBy([...allOrderEvents, ...allMessages], (message) =>
+    const sortedMessages = sortBy([...orderEventsWithoutFailedPayment, ...allMessages], (message) =>
       DateTime.fromISO(message.createdAt!)
     )
 
--- a/src/lib/Scenes/Inbox/Components/Conversations/__tests__/Messages-tests.tsx
+++ b/src/lib/Scenes/Inbox/Components/Conversations/__tests__/Messages-tests.tsx
@@ -174,6 +174,70 @@ describe("messages with order updates", () => {
     expect(messagesAndUpdates[0]).toContain("You received a counteroffer")
   })
 
+  it("removes 'Offer accepted' events when payment fails", () => {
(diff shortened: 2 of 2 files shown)


src/lib/Scenes/Inbox/Components/Conversations/Messages.tsx
```diff
@@ -60,9 +60,15 @@ export const Messages: React.FC<Props> = forwardRef((props, ref) => {
     .reduce<OrderEvent[]>((prev, order) => prev.concat(order.orderHistory), [])
     .map<OrderEventWithKey>((event, index) => ({ ...event, key: `event-${index}` }))
 
+  const orderEventsWithoutFailedPayment = allOrderEvents.filter((event, index) => {
+    if (!(event.state === "APPROVED" && allOrderEvents[index + 1].state === "SUBMITTED")) {
```
starsirius, line 64: @lilyfromseattle Apologies for the slowness. I was thinking we want to filter out _both_ the approved and submitted events in this case. Looks like we only care about [certain state change events](https://github.com/artsy/eigen/blob/c7c6ea506563a4835a4f4391d996a4bfaf3ef839/src/lib/Scenes/Inbox/Components/Conversations/OrderUpdate.tsx#L30-L43), so this might be fine.

I do wonder if this would error out when "approved" is the last event and `allOrderEvents[index + 1]` does not exist. Maybe we can have a test case for it to verify?

lilyfromseattle: If I'm not mistaken those submitted events don't actually appear in the view, but I can verify this/filter them anyway. I will also test that case

src/lib/Scenes/Inbox/Components/Conversations/__tests__/Messages-tests.tsx
```diff
@@ -174,6 +174,34 @@ describe("messages with order updates", () => {
     expect(messagesAndUpdates[0]).toContain("You received a counteroffer")
   })
 
+  fit("sorts interleaved items by date", () => {
```
lilyfromseattle, line 177: whoops!

src/lib/Scenes/Inbox/Components/Conversations/__tests__/Messages-tests.tsx
```diff
@@ -174,6 +174,70 @@ describe("messages with order updates", () => {
...
+    // get all elements and remove timestamps
+    const messagesAndUpdates = tree.root
+      .findAllByType(Text)
+      .filter((element) => element.props.color !== "black30")
+      .map((element) => extractText(element))
+
+    expect(messagesAndUpdates).not.toContain("Offer Accepted")
+  })
+
+  it("does not remove 'Offer accepted' events when payment succeeds", () => {
```
starsirius, line 205: 👍  very clear!

starsirius:
💯  Merging!

PR #4643, opened 2021-04-19
fix(android): fix bg image on welcome screen

pvinis:

This PR resolves [CX-1293]

### Description
Screenshot 2021-04-19 at 16 39 02

The main issue was that the image was given the "screenHeight", but that value did not include the safe area top inset, and since we need to have it, to draw the image from top to bottom, we need to add it.

I went ahead and changed the hook we use a bit. I simplified it to combine the hook from RN and the hook from safe-area-context to give us screen dimensions and insets respectively. I then also added `fullWidth` and `fullHeight` to help with these situations like above.

Also for two class components that we used the context, I just replaced it with the direct `Dimensions.get` call, since the context is removed.

I ended up improving some old tests that were testing using `toJson` to now use `extractText` or whatever made sense. One test I removed as it didn't seem to offer much, `BidFlow`. And lastly, some ImageCarousel tests are now being skipped.

[CX-1293]: https://artsyproduct.atlassian.net/browse/CX-1293


src/lib/utils/useScreenDimensions.tsx
```diff
@@ -14,53 +11,30 @@ export interface ScreenDimensions {
...
+    safeAreaInsets: insets,
+    width: frame.width,
+    height: frame.height,
+
+    get fullWidth() {
+      return this.width + this.safeAreaInsets.left + this.safeAreaInsets.right
+    },
+    get fullHeight() {
+      return this.height + this.safeAreaInsets.top + this.safeAreaInsets.bottom
+    },
```
ds300, line 35: I think we should rename these values to both make it more obvious what they are, and to maintain backwards compatibility

```suggestion
    safeAreaWidth: frame.width,
    safeAreaHeight: frame.height,

    get width() {
      return this.safeAreaWidth + this.safeAreaInsets.left + this.safeAreaInsets.right
    },
    get height() {
      return this.safeAreaHeight + this.safeAreaInsets.top + this.safeAreaInsets.bottom
    },
```

and also to maintain backwards compatibility. i noticed e.g. the fancy modal is messed up now because it's still using `height` which no longer includes the safe area

image

src/lib/types/SafeAreaInsets.ts
```diff
@@ -1,6 +1,4 @@
-export interface SafeAreaInsets {
-  top: number
-  bottom: number
-  left: number
-  right: number
-}
+import { EdgeInsets } from "react-native-safe-area-context"
+
+// tslint:disable-next-line:no-empty-interface
+export interface SafeAreaInsets extends EdgeInsets {}
```
ds300, line 4: 👍🏼

ds300 (approved):
Great work! I think we should reconfigure the useScreenDimensions values but otherwise this is great stuff. Thanks for cleaning up the tests 💆🏼

src/lib/utils/useScreenDimensions.tsx
pvinis: ah. I'll fix this. good catch.

pvinis: Here's how it is now:
Screenshot 2021-04-22 at 11 14 46

pvinis:
Updated, should be ready to go.

src/lib/utils/useScreenDimensions.tsx
ds300: ah that's still wrong. Maybe it would be better to go back to having the `Dimensions.get('screen')`

pvinis: Huh, ok. Should be even higher, closer to the top? I'll take a look.

brainbicycle (approved):

brainbicycle:
Nice work, thanks for cleaning up the hooks and the tests!  Feel free to merge when ready!

PR #4646, opened 2021-04-20
feat: release improved colors filter (FX-2868)

iskounen:

This PR resolves [FX-2868]

### Description

- Releases the new **colors** filter and replaces the old **color** filter.
- Fixes the display of the applied **colors** filter in the filter overview (see screenshot 3).

|image|image|image|
|---|---|---|

[FX-2868]: https://artsyproduct.atlassian.net/browse/FX-2868


src/lib/Components/ArtworkFilter/ArtworkFilterStore.tsx
```diff
@@ -246,7 +246,6 @@ export const DEFAULT_FILTERS: FilterArray = [
   { paramName: FilterParamName.size, paramValue: "*-*", displayText: "All" },
   { paramName: FilterParamName.gallery, displayText: "All" },
   { paramName: FilterParamName.institution, displayText: "All" },
-  { paramName: FilterParamName.color, displayText: "All" },
```
olerichter00, line 249: Note: I'm not sure whether to remove the `color` filter from the store or not. I guess we can remove it here because we don't use this filter anymore.

olerichter00: If we do remove the color filter in the store, we need to add a migration.

olerichter00: cc @MounirDhahri

MounirDhahri: As long as it's not used anywhere else, we should be good, there is no need to write a migration here.

dzucconi (approved):

MounirDhahri (approved):

PR #4647, opened 2021-04-20
feat: Onboarding sign up screen (CX-1081)

MounirDhahri:

This PR resolves [CX-1081]

### Description
- User can Create Account with Email + Password

https://user-images.githubusercontent.com/11945712/115420523-7e22b880-a1fb-11eb-94fc-df19352bf6a3.mov

[CX-1081]: https://artsyproduct.atlassian.net/browse/CX-1081


src/lib/Scenes/Onboarding/Onboarding.tsx
```diff
@@ -1,20 +1,20 @@
...
-import { OnboardingCreateAccount } from "./OnboardingCreateAccount"
+import { OnboardingCreateAccount } from "./OnboardingCreateAccount/OnboardingCreateAccount"
 import { OnboardingLogin } from "./OnboardingLogin"
 import { OnboardingWelcome } from "./OnboardingWelcome"
 
 // tslint:disable-next-line:interface-over-type-literal
 export type OnboardingNavigationStack = {
   OnboardingWelcome: undefined
-  OnboardingLogin: undefined
+  OnboardingLogin: { withFadeAnimation: boolean; email: string } | undefined
```
MounirDhahri, line 17: I want the login screen to fade instead of getting pushed from the right.
**Fade**

https://user-images.githubusercontent.com/11945712/115433373-c7790500-a207-11eb-9260-0daeaceb8b2f.mov


**Horizontal transition**

https://user-images.githubusercontent.com/11945712/115433359-c1832400-a207-11eb-8159-b5a93dcc8ffc.mov

src/lib/Scenes/Onboarding/OnboardingCreateAccount/__tests__/OnboardingCreateAccount-tests.tsx
```diff
@@ -0,0 +1,148 @@
...
+      id: "my-user-id",
+    })
+    const isLoggedIn = !!__globalStoreTestUtils__?.getCurrentState().auth.userAccessToken
+    expect(isLoggedIn).toEqual(false)
+
+    signUpButton.props.onPress()
+
+    setTimeout(() => {
+      expect(isLoggedIn).toEqual(true)
+    }, 2000)
```
MounirDhahri, line 118: I didn't know how to work around this

MounirDhahri: await flushPromiseQueue didn't do the trick for me

src/palette/elements/Button/Button.tsx
```diff
@@ -68,88 +68,92 @@ export interface ButtonBaseProps extends BoxProps {
...
-    colors: { black100, black10, black30, white100, purple100, red100 },
-  } = themeProps
+export function getColorsForVariant(variant: ButtonVariant, disabled: boolean = false) {
+  const opacity = disabled ? "0.1" : "1"
+  const black100WithOpacity = `rgba(0, 0, 0, ${opacity})`
+  const black10WithOpacity = `rgba(229, 229, 229, ${opacity})`
+  const whiteWithOpacity = `rgba(255, 255, 255, ${opacity})`
+  const purple100WithOpaicty = `rgba(110, 30, 255, ${opacity})`
+  const black30WithOpacity = `rgba(194, 194, 194, ${opacity})`
+  const red100WithOpacity = `rgba(232, 46, 29, ${opacity})`
```
MounirDhahri, line 78: view inherit their flex opacity in android and it makes the text in with a 0.1 opacity not show up, to avoid that i had to set the opacity as a color instead of a flex prop

src/palette/elements/Button/Button.tsx
```diff
@@ -183,6 +187,10 @@ export const Button: React.FC<ButtonProps> = (props) => {
   const loadingStyles = (() => {
     const opacity = props.disabled ? "0.1" : "1"
 
+    if (!props.loading) {
+      return {}
+    }
+
```
MounirDhahri, line 193: remove this on this [PR](https://github.com/artsy/eigen/commit/2a7e1fa6893041bc6859fa46ead537ba01a173c3#diff-c3d3201efe476e1ae269898a28f45d4c831bd8850a95a6074c088568611b30bbL183-L185) was a mistake, I am just reverting it here

src/lib/store/AuthModel.ts
```diff
@@ -164,4 +165,27 @@ export const getAuthModel = (): AuthModel => ({
...
+      },
+    })
+
+    // The user account has been successfully created
+    if (result.status === 201) {
+      await actions.signIn({ email, password })
+      return true
+    }
+    return false
+  }),
```
MounirDhahri, line 190: This seems a bit weird for me but it's how we do sign up on iOS now, please me time know if I should change it

src/palette/elements/Button/Button.tsx
```diff
@@ -68,88 +68,92 @@ export interface ButtonBaseProps extends BoxProps {
...
-export function getColorsForVariant(variant: ButtonVariant) {
-  const {
-    colors: { black100, black10, black30, white100, purple100, red100 },
-  } = themeProps
+export function getColorsForVariant(variant: ButtonVariant, disabled: boolean = false) {
+  const opacity = disabled ? "0.1" : "1"
+  const black100WithOpacity = `rgba(0, 0, 0, ${opacity})`
+  const black10WithOpacity = `rgba(229, 229, 229, ${opacity})`
+  const whiteWithOpacity = `rgba(255, 255, 255, ${opacity})`
+  const purple100WithOpaicty = `rgba(110, 30, 255, ${opacity})`
```
ds300, line 76:
```suggestion
  const purple100WithOpacity = `rgba(110, 30, 255, ${opacity})`
```

src/lib/store/AuthModel.ts
ds300: seems fine to me, what seems weird to you?

src/lib/Scenes/Onboarding/OnboardingLogin.tsx
```diff
@@ -39,86 +39,107 @@ export const OnboardingLoginForm: React.FC<OnboardingLoginProps> = ({ navigation
...
+   * on the back button to navigate to the welcome screen, the screen
+   * fades instead of translating horizontally. To avoid that we need
+   * to overwrite withFadeAnimation param once the screen shows up
+   */
+  useEffect(() => {
+    setTimeout(() => {
+      navigation.setParams({ withFadeAnimation: false })
+    }, 1000)
+    if (route.params?.email) {
+      handleChange("email")(route.params?.email || "")
```
ds300, line 54:
```suggestion
      handleChange("email")(route.param.email)
```
the truthiness was already verified in the if condition above

src/lib/Scenes/Onboarding/OnboardingCreateAccount/__tests__/OnboardingCreateAccount-tests.tsx
ds300: might need to do it multiple times, depending on what the async logic is doing. Also might need to wrap the call to `.onPress` in `act`

src/lib/Scenes/Onboarding/OnboardingCreateAccount/__tests__/OnboardingCreateAccount-tests.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+describe("OnboardingCreateAccount", () => {
+  beforeEach(() => {
+    mockFetch.mockClear()
+  })
+
+  it("form validation works properly", async () => {
+    const tree = renderWithWrappers(<OnboardingCreateAccount navigation={navigationMock as any} route={null as any} />)
+
+    const signUpButton = tree.root.findAllByProps({ testID: "signUpButton" })[0]
```
ds300, line 39: nonblocking:
```suggestion
    const signUpButton = tree.root.findByProps({ testID: "signUpButton" })
```
i prefer this pattern since you get better error messages if the button is not present for some reason, or if there's multiple instances of the same button

src/lib/Scenes/Onboarding/OnboardingCreateAccount/OnboardingCreateAccountPassword.tsx
```diff
@@ -0,0 +1,69 @@
...
+  extends StackScreenProps<OnboardingCreateAccountNavigationStack, "OnboardingCreateAccountPassword"> {}
+
+export const OnboardingCreateAccountPassword: React.FC<OnboardingCreateAccountPasswordProps> = ({ navigation }) => {
+  const { values, handleSubmit, handleChange, validateForm, errors, setErrors } = useFormikContext<UserSchema>()
+
+  return (
+    <Flex backgroundColor="white" flexGrow={1}>
+      <ScrollView
+        contentContainerStyle={{
+          paddingHorizontal: 15,
```
ds300, line 21: is this `15` on every page deliberate? the whole app uses 20px margins and it seems odd to suddenly change for this onboarding flow.

src/lib/Scenes/Onboarding/OnboardingCreateAccount/OnboardingCreateAccountName.tsx
```diff
@@ -0,0 +1,60 @@
...
+        contentContainerStyle={{
+          paddingHorizontal: 15,
+          paddingTop: useScreenDimensions().safeAreaInsets.top,
+          justifyContent: "flex-start",
+        }}
+        showsVerticalScrollIndicator={false}
+        keyboardDismissMode="on-drag"
+        keyboardShouldPersistTaps="handled"
+      >
+        <BackButton onPress={navigation.goBack} />
```
ds300, line 29: nonblocking: might be nice to pull this stuff out into a separate component, kinda like we do for the MyAcccountFieldEditScreen.

src/lib/Scenes/Onboarding/OnboardingCreateAccount/OnboardingCreateAccount.tsx
```diff
@@ -0,0 +1,228 @@
...
+})
+
+const getCurrentRoute = () =>
+  __unsafe__createAccountNavigationRef.current?.getCurrentRoute()?.name as
+    | keyof OnboardingCreateAccountNavigationStack
+    | undefined
+
+const EMAIL_EXISTS_ERROR_MESSAGE = "We found an account with this email"
+
+export const OnboardingCreateAccount: React.FC<OnboardingCreateAccountProps> = ({ navigation }) => {
```
ds300, line 58: nonblocking:

I think this onboarding flow would be more maintainable if the individual pages were more independent. e.g. they each have their own formik context and their own next/submit buttons. Like this it takes extra effort to figure out how things work on a particular page.

src/lib/Scenes/Onboarding/OnboardingCreateAccount/__tests__/OnboardingCreateAccount-tests.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+    signUpButton.props.onPress()
+
+    const emailInput = tree.root.findAllByType(Input)[0]
```
ds300, line 43: nonblocking: I'm a bit worried about the `Input` selections in this test, and that they might be quite brittle as people make changes to the onboarding flow. It would probably be better to use `testID` props to select them.

src/lib/Scenes/Onboarding/OnboardingLogin.tsx
```diff
@@ -39,86 +39,107 @@ export const OnboardingLoginForm: React.FC<OnboardingLoginProps> = ({ navigation
...
+            error={errors.email}
+          />
+          <Spacer mt={2} />
+          <Input
+            autoCapitalize="none"
+            autoCompleteType="password"
+            autoCorrect={false}
+            // If we have the email already prefilled from the navigation params
+            // we want the autoFocus to be on the password
+            autoFocus={!!route.params?.email}
```
ds300, line 107: 👍🏼

ds300 (approved):
This is looking great!

src/lib/store/AuthModel.ts
MounirDhahri: nothing really, but I wanted to highlight this since the flow is a bit different on web

src/lib/Scenes/Onboarding/Onboarding.tsx
```diff
@@ -31,27 +31,34 @@ export const Onboarding = () => {
...
-            <StackNavigator.Screen
-              name="OnboardingLogin"
-              component={OnboardingLogin}
-              options={{ headerShown: false }}
-            />
-            <StackNavigator.Screen name="OnboardingCreateAccount" component={OnboardingCreateAccount} />
-            <StackNavigator.Screen name="ForgotPassword" component={ForgotPassword} options={{ headerShown: false }} />
-          </StackNavigator.Navigator>
-        </ArtsyKeyboardAvoidingView>
+        <ArtsyKeyboardAvoidingViewContext.Provider value={{ isVisible: true, isPresentedModally: false }}>
```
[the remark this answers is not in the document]
MounirDhahri: oh thank you, I will adjust it then

src/lib/Scenes/Onboarding/Onboarding.tsx
```diff
@@ -31,27 +31,34 @@ export const Onboarding = () => {
...
+                  cardStyleInterpolator: params?.withFadeAnimation
+                    ? CardStyleInterpolators.forFadeFromBottomAndroid
+                    : CardStyleInterpolators.forHorizontalIOS,
+                })}
+              />
+              <StackNavigator.Screen name="OnboardingCreateAccount" component={OnboardingCreateAccount} />
+              <StackNavigator.Screen
+                name="ForgotPassword"
+                component={ForgotPassword}
+                options={{ headerShown: false }}
```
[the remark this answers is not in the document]
MounirDhahri: oh makes sense actually

src/lib/Scenes/Onboarding/OnboardingCreateAccount/__tests__/OnboardingCreateAccount-tests.tsx
MounirDhahri: yes, that makes sense actually. I will change it

src/lib/Scenes/Onboarding/OnboardingCreateAccount/OnboardingCreateAccountPassword.tsx
MounirDhahri: Yes it was deliberate actually, but I think it makes sense to change it to be here as well

src/lib/Scenes/Onboarding/OnboardingCreateAccount/OnboardingCreateAccountName.tsx
MounirDhahri: 👌👌👌

src/lib/Scenes/Onboarding/OnboardingCreateAccount/OnboardingCreateAccount.tsx
MounirDhahri: yes, I wanted actually to make them separate but wanted to have the button shared between them all. but i guess yes, it should be possible

src/lib/Scenes/Onboarding/OnboardingCreateAccount/__tests__/OnboardingCreateAccount-tests.tsx
MounirDhahri: that's smart, oki I will adjust it

src/lib/Scenes/Onboarding/Onboarding.tsx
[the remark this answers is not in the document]
MounirDhahri: it looks like it's required to be added 😬, I will see if I can change that

src/palette/elements/Button/Button.tsx
```diff
@@ -68,88 +68,92 @@ export interface ButtonBaseProps extends BoxProps {
...
  * @param variant
  */
-export function getColorsForVariant(variant: ButtonVariant) {
-  const {
-    colors: { black100, black10, black30, white100, purple100, red100 },
-  } = themeProps
+export function getColorsForVariant(variant: ButtonVariant, disabled: boolean = false) {
+  const opacity = disabled ? "0.1" : "1"
+  const black100WithOpacity = `rgba(0, 0, 0, ${opacity})`
+  const black10WithOpacity = `rgba(229, 229, 229, ${opacity})`
```
dzucconi, line 74: `black10` with opacity would be `rgba(0,0,0,0.1)` this is a very light grey which is then barely visible once the opacity is there.

src/palette/elements/Button/Button.tsx
```diff
@@ -68,88 +68,92 @@ export interface ButtonBaseProps extends BoxProps {
...
+  const whiteWithOpacity = `rgba(255, 255, 255, ${opacity})`
+  const purple100WithOpacity = `rgba(110, 30, 255, ${opacity})`
+  const black30WithOpacity = `rgba(194, 194, 194, ${opacity})`
```
dzucconi, line 77: Ditto

src/palette/elements/Button/Button.tsx
```diff
@@ -68,88 +68,92 @@ export interface ButtonBaseProps extends BoxProps {
  * Returns various colors for each state given a button variant
  * @param variant
  */
-export function getColorsForVariant(variant: ButtonVariant) {
-  const {
-    colors: { black100, black10, black30, white100, purple100, red100 },
-  } = themeProps
+export function getColorsForVariant(variant: ButtonVariant, disabled: boolean = false) {
+  const opacity = disabled ? "0.1" : "1"
```
dzucconi, line 72: Rather than applying opacity to the RGBA values to denote disabled state; the colors should be full opacity and the outer container is the thing that should have opacity. (`0.1` is also quite aggressive here I suspect)

src/palette/elements/Button/Button.tsx
MounirDhahri: good catch, I will update it 🙌

src/palette/elements/Button/Button.tsx
MounirDhahri: The issue was that on android, the child components inherit the opacity from the outer container and it makes them all almost invisible compared to iOS where setting the parent component opacity sets it only for the outer container.

PR #4652, opened 2021-04-20
feat: live auctions android

brainbicycle:

This PR resolves [CX-1279]

### Description

Shows live auctions in a webview for Android builds.

~~Prediction (Live Auctions) Auth:~~
~~Prediction uses gravity oauth authentication and so sessions are not automatically shared with artsy.net and without extra work user would have to login to live auctions again in the webview. To workaround this on user login I am logging into gravity with the same user. Then when showing prediction initially we redirect to '{original_url}/login' to kick off gravity oauth and give the user a valid session. Inspired by this PR: https://github.com/artsy/force/pull/4379  thanks @dblandin for pointing me in the right direction!~~

[CX-1279]: https://artsyproduct.atlassian.net/browse/CX-1279


ds300:
@brainbicycle I think all the auth stuff can be done with just this one line change in ArtsyReactWebView.tsx

image

This is my bad, I didn't check the prediction auth status thoroughly when doing the original web view infra work.

As for the routes.tsx bit I think we could also simplify that to

ds300:
proof that the simplified auth change works:

image

ds300:
yay 🎉

PR #4656, opened 2021-04-21
fix: fix android share

brainbicycle:

This PR resolves [CX-1276]

### Description

Fixes sharing on Android, we had 2 issues. 1 sharing was very slow because we were fetching a large image, 2 we were passing a non-base64 encoded url to urls array which was causing the base64 string to be shared instead of the image. 

- Fixes a small image instead
- Moves details url out of urls array and adds it to end of share message

### Screenshots

https://user-images.githubusercontent.com/49686530/115553878-e5804d00-a27b-11eb-8c86-eb96f58136ac.mp4

[CX-1276]: https://artsyproduct.atlassian.net/browse/CX-1276

### DIFF
--- a/src/lib/Scenes/Artwork/Components/ArtworkActions.tsx
+++ b/src/lib/Scenes/Artwork/Components/ArtworkActions.tsx
@@ -133,7 +133,7 @@ export class ArtworkActions extends React.Component<ArtworkActionsProps> {
               </UtilButton>
             </TouchableWithoutFeedback>
           )}
-          <Touchable noFeedback haptic onPress={() => this.props.shareOnPress()}>
+          <Touchable haptic onPress={() => this.props.shareOnPress()}>
             <UtilButton>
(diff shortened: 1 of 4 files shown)


src/lib/Scenes/Artwork/Components/ArtworkActions.tsx
```diff
@@ -133,7 +133,7 @@ export class ArtworkActions extends React.Component<ArtworkActionsProps> {
               </UtilButton>
             </TouchableWithoutFeedback>
           )}
-          <Touchable noFeedback haptic onPress={() => this.props.shareOnPress()}>
+          <Touchable haptic onPress={() => this.props.shareOnPress()}>
```
pvinis, line 136: Good call. I check why that was there, and it was because we had `TouchableWithoutFeedback`, but after introducing `Touchable` I guess it's better to have all three consistent.

PR #4658, opened 2021-04-21
fix: keyboard avoidance for iOS13 modals

ds300:
### Description

This is a follow-up to #4619

This fixes an issue where iOS 13+ pageSheet modals had too much bottom padding added when the keyboard is visible

image

No idea how I missed this before, I even posted a screenshot that showed this same screen working as expected in the original PR. 🤷🏼

Anyway this change makes perfect sense to me so I'm also confused how it was ever working for that screenshot. Maybe I had some weird simulator state.

I checked this on iOS 12 and android and everything there is still 👍🏼 


pvinis (approved):
Nice!
I only noticed that the bottom button disappears when I scroll even a little bit. But this behavior was there already, so all good. Just raising it in case you know it's a bug or correct.


PS: That is some good "aviodance" 🕺 🥁 😛

PR #4665, opened 2021-04-21
chore: move .ruby-version declaration to project root

anandaroop:

## Description

This moves the `.ruby-version` declaration from the `/emission` folder up to the root `/` .

That way, as soon as you `cd` into Eigen your Ruby version manager will actually pick up the correct version and thus your bundle installs etc will target the correct version of Ruby.

Maybe worth a team head's up since this will enforce a specific version of ruby and potentially require some reinstalling of ruby versions and/or gems. Probably worth it though, to ensure a more consistent dev environment across the team (which was probably the intent to begin with 😄 )


MounirDhahri:
Thank you for fixing this 👌

PR #4674, opened 2021-04-22
feat(FX-2625): combine galleries and institutions filters to a single filter

mdole:

This PR resolves [FX-2625]

### Description

Using the patterns established by other newly-updated filters like rarity and medium, we're combining the existing galleries and institutions filters into a single filter with multi-select capabilities.

### Demo

gall-inst-filter

[FX-2625]: https://artsyproduct.atlassian.net/browse/FX-2625


iskounen:
I converted this ticket to be a draft because I believe that this should be behind a feature flag until we add the search interface. Happy to discuss if people feel differently.

src/lib/Components/ArtworkFilter/ArtworkFilterHelpers.ts
```diff
@@ -344,13 +346,23 @@ export const selectedOption = ({
...
+  } else if (filterScreen === "institution") {
+    const multiSelectedOptions = selectedOptions.filter((filter) => filter.paramName === FilterParamName.institution)
+
+    const institutionOptions = multiSelectedOptions.map((option) => option.displayText)
+
+    if (institutionOptions.length === 0) {
+      return "All"
+    }
+
+    return institutionOptions.join(", ")
```
dzucconi, line 352: So the hook takes care of this for now and it's no longer needed.

src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
```diff
@@ -282,13 +283,14 @@ export default createPaginationContainer(
...
           partnerID: $partnerID
+          partnerIDs: $partnerIDs
           dimensionRange: $dimensionRange
           majorPeriods: $majorPeriods
           acquireable: $acquireable
           inquireableOnly: $inquireableOnly
           atAuction: $atAuction
           offerable: $offerable
-          aggregations: [COLOR, DIMENSION_RANGE, GALLERY, INSTITUTION, MAJOR_PERIOD, MEDIUM, PRICE_RANGE]
+          aggregations: [COLOR, DIMENSION_RANGE, PARTNER, MAJOR_PERIOD, MEDIUM, PRICE_RANGE]
```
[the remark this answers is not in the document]
dzucconi: [Presumably] this returns aggregations for both galleries **and** institutions.

dzucconi:
@iskounen I'm in favor of not using feature flags unless we absolutely have to. We can follow up with the search functionality rapidly; but this is mergeable and useable as is IMO.

dzucconi:
OK; I think this looks good. Fixed a minor bug with the labels and will merge when green. I'll just pick up the search ticket for this right now which is a minor thing.

src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
```diff
@@ -370,6 +373,7 @@ export default createPaginationContainer(
                 color: $color
                 colors: $colors
                 partnerID: $partnerID
+                partnerIDs: $partnerIDs
```
dzucconi, line 376: OK so whenever making these changes; right now they have to be changed in literally every query. I just wrote a ticket and set its priority to high to convert these to a single structured $input; which will completely eliminate this problem.

For now I'll just try to bulk find and replace the remaining of these to get this merged.

dzucconi:
OK, now with 9a8cd1c9ed0797a9495d7fda75620e92c8ed3cad this is working in all the places it shows up

PR #4679, opened 2021-04-22
feat: release improved artwork filter style

iskounen:
This PR resolves [FX-2871]

### Description

This PR removes style changes to the artwork filter modal. (For Artsy engineers, the design specs can be found [here](https://www.figma.com/file/zG9cduJdOaTQAw886QcOwn/Filters?node-id=1376%3A15695).)

- The "Apply (N)" button now simply shows "Show results"
- The separator lines between rows have been removed
- Rows are highlighted when a user taps on them
- Default values "All" and "Default" are no longer shown
- Multi-select filters have changed from toggle switches to checkboxes
- Single-select filters have changed from check marks to radio buttons

#### Before

before

#### After

after

[FX-2871]: https://artsyproduct.atlassian.net/browse/FX-2871

### DIFF
--- a/src/lib/Components/ArtworkFilter/ArtworkFilter.tsx
+++ b/src/lib/Components/ArtworkFilter/ArtworkFilter.tsx
@@ -6,7 +6,6 @@ import {
   changedFiltersParams,
   FilterArray,
   filterArtworksParams,
-  FilterParamName,
   FilterParams,
 } from "lib/Components/ArtworkFilter/ArtworkFilterHelpers"
 import { ArtworksFiltersStore } from "lib/Components/ArtworkFilter/ArtworkFilterStore"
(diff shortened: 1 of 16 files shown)


src/lib/Components/ArtworkFilter/ArtworkFilter.tsx
```diff
@@ -136,36 +133,6 @@ export const ArtworkFilterNavigator: React.FC<ArtworkFilterProps> = (props) => {
...
-      )
-      const hasArtistIDs = !!selectedFiltersState.find((filter) => filter.paramName === FilterParamName.artistIDs)
-
-      if (hasArtistIDs && hasArtistsIFollow) {
-        --selectedFiltersSum
-      }
-    }
-    return selectedFiltersSum > 0 ? `Apply (${selectedFiltersSum})` : "Apply"
-  }
-
```
dzucconi, line 168: Nice

src/lib/Components/ArtworkFilter/ArtworkFilterOptionsScreen.tsx
```diff
@@ -170,12 +167,8 @@ export const ArtworkFilterOptionsScreen: React.FC<
             aggregations: aggregationsState,
           })
 
-          // TODO: When unwinding the `ARUseImprovedArtworkFilters` flag; simply return `null`
-          // instead of `"All"` in the `selectedOption` function
```
dzucconi, line 175: I think it's fine to leave this as is for now but my comment here was stating that we're still returning "All" in places only to check for it and then ignore it; which is kinda weird. Where as we can just return nothing in those cases. It's probably an update in a few places and I know we want to get this released so maybe just leave this TODO comment so we know we can safely delete those at some point.

dzucconi (approved):
👍

PR #4684, opened 2021-04-26
fix: deactivate filter apply button when toggling unapplied filters

iskounen:

This PR resolves [FX-2870]

### Description

When a user selects options in a multi-select filter (ie. Medium, Rarity, Time period, Color) and un-selects those options without applying them, the "Show results" button should be de-activated. However, it currently remains active so this PR addresses that behavior.

#### Before

before

#### After

after

[FX-2870]: https://artsyproduct.atlassian.net/browse/FX-2870


src/lib/Components/ArtworkFilter/ArtworkFilterStore.tsx
```diff
@@ -166,7 +166,7 @@ export const ArtworkFiltersModel: ArtworkFiltersModel = {
         ) {
           return true
         }
-        return defaultFilterOptions[paramName] !== paramValue
+        return !isEqual(defaultFilterOptions[paramName], paramValue)
```
iskounen, line 169: This was needed because `[] === []` is `false`.

MounirDhahri: ⭐️

MounirDhahri (approved):

src/lib/Components/ArtworkFilter/ArtworkFilterStore.tsx
iskounen: @MounirDhahri just for your information, I made the same change to `applyFiltersAction` to fix the following bug where the count on the "Sort & Filter" pill was not being updated when applied filters were deselected.

pill

dzucconi (approved):
🙏

PR #4690, opened 2021-04-27
enhancement(cocoapods): Remove version pin

pvinis:

### Description

Also came out of the RN upgrade attempt.

Cocoapods stays stable mostly, and every now and then we need to update, as with fastlane. Both upgrade more easily when there is no pin, with just `bundle update cocoapods`, whenever there is the need for it. For RN 0.64 there was.

Pinning doesn't really help us for fastlane and cocoapods, where we just want the latest version that works. If we ever get to the point where we don't want some specific new version, then we can pin again.


Artsy.xcodeproj/project.pbxproj
```diff
@@ -3311,6 +3311,8 @@
 				"${PODS_ROOT}/Artsy+UIFonts/Pod/Assets/Unica77LL-Regular.otf",
 				"${PODS_ROOT}/Artsy+UILabels/Pod/Assets/Chevron_Black.png",
 				"${PODS_ROOT}/Artsy+UILabels/Pod/Assets/Chevron_Black@2x.png",
+				"${PODS_ROOT}/../emission/Pod/Assets/Emission.js",
+				"${PODS_ROOT}/../emission/Pod/Assets/assets",
```
ds300, line 3315: I think the changes in this file shouldn't be staged because it makes building the app fail for people who haven't run `react-native bundle` locally. I could be wrong 🤔

Are these changes intentional? If so can you try checking out this branch, removing `emission/Pod/Assets`, and building the app for dev mode?

pvinis: It appeared after running `bundle e pod install`, so I guessed it was needed. I can remove them and if they appear for others too, we commit them then? 🤔

pvinis: I'll try what you said now

pvinis: oh yea. interesting. ok removing this file change then. can't wait till we fix that thing with emission.

PR #4705, opened 2021-05-03
[WIP] feat(FX-2907): Convert artist series filter to static button

mdole:

This PR resolves [FX-2907]

### Description

This PR converts the artist series screen from the floating "pill" filter button to a static button in the header.

I converted the ArtistSeries screen from a Flatlist to a ScrollView to deal with a weird sticky header bug I was having (see gif). Is this a terrible idea? Open to input! My thinking was that this page isn't massive so it wasn't a big deal if we render all of it (though it could get quite long if the user hits "load more artworks" a bunch - does the Flatlist in `InfiniteScrollArtworksGridContainer` handle that?). If converting is a bad idea, I'm sure there's a different way to handle this bug.

### Sticky header w/ flatlist (buggy):
scroll-bug

[FX-2907]: https://artsyproduct.atlassian.net/browse/FX-2907

### DIFF
--- a/src/lib/Scenes/ArtistSeries/ArtistSeries.tsx
+++ b/src/lib/Scenes/ArtistSeries/ArtistSeries.tsx
@@ -7,43 +7,28 @@ import { ArtistSeriesHeaderFragmentContainer } from "lib/Scenes/ArtistSeries/Art
 import { ArtistSeriesMetaFragmentContainer } from "lib/Scenes/ArtistSeries/ArtistSeriesMeta"
 import { ArtistSeriesMoreSeriesFragmentContainer } from "lib/Scenes/ArtistSeries/ArtistSeriesMoreSeries"
 import { ProvideScreenTracking, Schema } from "lib/utils/track"
-import { Box, Separator, Spacer, Theme } from "palette"
-import React, { useRef, useState } from "react"
+import { Box, Flex, Separator, Spacer, Theme } from "palette"
+import React, { useState } from "react"
 
-import { AnimatedArtworkFilterButton, ArtworkFilterNavigator, FilterModalMode } from "lib/Components/ArtworkFilter"
+import { ArtworkFilterNavigator, FilterModalMode } from "lib/Components/ArtworkFilter"
 import { ArtworkFiltersStoreProvider } from "lib/Components/ArtworkFilter/ArtworkFilterStore"
+import { useHideBackButtonOnScroll } from "lib/utils/hideBackButtonOnScroll"
 import { PlaceholderBox, PlaceholderGrid, PlaceholderText } from "lib/utils/placeholders"
 import { renderWithPlaceholder } from "lib/utils/renderWithPlaceholder"
 import { OwnerEntityTypes, PageNames } from "lib/utils/track/schema"
-import { FlatList, View, ViewToken } from "react-native"
+import { ScrollView } from "react-native"
 import { createFragmentContainer, graphql, QueryRenderer } from "react-relay"
 import { useTracking } from "react-tracking"
 interface ArtistSeriesProps {
   artistSeries: ArtistSeries_artistSeries
 }
 
-interface ViewableItems {
-  viewableItems?: ViewToken[]
-}
-
 export const ArtistSeries: React.FC<ArtistSeriesProps> = (props) => {
   const { artistSeries } = props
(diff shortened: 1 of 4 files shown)


src/lib/Components/ArtworkFilter/ArtworkFilterOptionsScreen.tsx
```diff
@@ -351,6 +351,70 @@ export const AnimatedArtworkFilterButton: React.FC<AnimatedArtworkFilterButtonPr
...
+        {getFiltersCount() > 0 && (
+          <>
+            <Sans size="3t" pl={0.5} color="black100" weight="medium">
+              {"\u2022"}
+            </Sans>
+            <Sans size="3t" pl={0.5} color="black100" weight="medium">
+              {getFiltersCount()}
+            </Sans>
+          </>
+        )}
```
iskounen, line 412: The FX team decided to not include the applied filter count in the new sort & filter button because the count that we were applying wasn't always clear, and we were spending too much time in QA sessions just addressing count issues.

src/lib/Components/ArtworkFilter/ArtworkFilterOptionsScreen.tsx
```diff
@@ -351,6 +351,70 @@ export const AnimatedArtworkFilterButton: React.FC<AnimatedArtworkFilterButtonPr
...
+  }
+
+  return (
+    <Touchable haptic onPress={onPress}>
+      <Flex flexDirection="row">
+        <FilterIcon fill="black100" />
+
+        <Sans size="3t" pl="1" color="black100" weight="medium">
+          {text}
+        </Sans>
```
iskounen, line 402: We should refactor any `Sans` or `Serif` elements that we find to `Text` elements.

src/lib/Components/ArtworkFilter/ArtworkFilterOptionsScreen.tsx
```diff
@@ -351,6 +351,70 @@ export const AnimatedArtworkFilterButton: React.FC<AnimatedArtworkFilterButtonPr
   )
 }
 
+type StaticArtworkFilterButtonProps = Omit<AnimatedArtworkFilterButtonProps, "isVisible">
+
+export const StaticArtworkFilterButton: React.FC<StaticArtworkFilterButtonProps> = ({
```
iskounen, line 356: **suggestion (non-blocking):** We should move this into its own file because this one is getting unwieldy.

src/lib/Scenes/ArtistSeries/ArtistSeries.tsx
```diff
@@ -109,6 +100,13 @@ export const ArtistSeries: React.FC<ArtistSeriesProps> = (props) => {
                   case "artistSeriesArtworks":
                     return (
                       <Box px={2}>
+                        <Separator mb={2} />
+                        <Flex mb={2} flexDirection="row" justifyContent="space-between">
+                          <Text color="black60">
+                            Showing {totalArtworksCount} of {totalArtworksCount}
+                          </Text>
+                          <StaticArtworkFilterButton onPress={openFilterArtworksModal} />
+                        </Flex>
```
iskounen, line 109: **suggestion:** This header will be used by all artwork rails, so let's make it its own component (and maybe `StaticArtworkFilterButton` can live inside the same file as this component).

src/lib/Scenes/ArtistSeries/ArtistSeries.tsx
```diff
@@ -109,6 +100,13 @@ export const ArtistSeries: React.FC<ArtistSeriesProps> = (props) => {
                   case "artistSeriesArtworks":
                     return (
                       <Box px={2}>
+                        <Separator mb={2} />
+                        <Flex mb={2} flexDirection="row" justifyContent="space-between">
+                          <Text color="black60">
+                            Showing {totalArtworksCount} of {totalArtworksCount}
```
iskounen, line 106: **thought:** This caught my attention because the designs call for `"Showing {totalArtworksCount} works"`. I think that this will give users an indication that they are seeing a filtered view, but I am not sure that it is clear enough. (If I saw "Showing 87 or 105" I would think "where are the other results?")

mdole:
Closing - deprioritized for saved search v1.0

PR #4718, opened 2021-05-06
[WIP] feat: convert artist insights filter button to new style

mdole:
The type of this PR is: **FEAT**

This PR resolves [FX-2905]

### Description

Following up on #4703, this PR does the same for the Insights tab

Still to resolve:
- [ ] Stickiness

[FX-2905]: https://artsyproduct.atlassian.net/browse/FX-2905

### DIFF
--- a/src/lib/Components/Artist/ArtistInsights/ArtistInsightsAuctionResults.tsx
+++ b/src/lib/Components/Artist/ArtistInsights/ArtistInsightsAuctionResults.tsx
@@ -2,15 +2,14 @@ import { ActionType, ContextModule, OwnerType, tappedInfoBubble, TappedInfoBubbl
 import { ArtistInsightsAuctionResults_artist } from "__generated__/ArtistInsightsAuctionResults_artist.graphql"
 import { filterArtworksParams } from "lib/Components/ArtworkFilter/ArtworkFilterHelpers"
 import { ArtworksFiltersStore } from "lib/Components/ArtworkFilter/ArtworkFilterStore"
-import { ORDERED_AUCTION_RESULTS_SORTS } from "lib/Components/ArtworkFilter/Filters/SortOptions"
 import { FilteredArtworkGridZeroState } from "lib/Components/ArtworkGrids/FilteredArtworkGridZeroState"
 import { InfoButton } from "lib/Components/Buttons/InfoButton"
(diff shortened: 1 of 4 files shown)


src/lib/Components/Artist/ArtistInsights/ArtistInsightsAuctionResults.tsx
```diff
@@ -140,28 +133,39 @@ const ArtistInsightsAuctionResults: React.FC<Props> = ({ artist, relay, scrollTo
           }}
         />
       )}
+      stickyHeaderIndices={[0]}
```
mdole, line 136: I'm confused as to why this isn't working. Seems like it has something to do with this already being inside another scrollview ([this one](https://github.com/artsy/eigen/blob/63821892c5ceb8c19493f28fd865495c814abede/src/lib/Components/Artist/ArtistInsights/ArtistInsights.tsx#L75)), but I'm failing to understand the actual problem. @iskounen I think you alluded to this as a potential problem; open to ideas if you have 'em.

iskounen: @MounirDhahri explained why this isn't working and laid out some solutions.

In this screen, we have a `FlatList` (auction results) nested inside of a `ScrollView`, but the scroll events in the `ScrollView` aren't propagated to the `FlatList`, so it never knows that it needs to stick the header component.

The artist artwork screen doesn't have this problem because it has a [`ScrollView`](https://github.com/artsy/eigen/blob/dc8075d3569b0e6b786ba836ffc8a41e69d6aea9/src/lib/Components/ArtworkGrids/InfiniteScrollArtworksGrid.tsx#L285-L317) (artwork grid) nested inside of a `ScrollView`, and we have a [monkey patch](https://github.com/artsy/eigen/blob/master/emission/Pod/Classes/Core/RCTScrollView+Artsy.m#L1) in the repo to detect a nested `ScrollView` and propagate scroll events to it.

The desired solution to this is to update that monkey-patch to do the same thing for a nested `FlatList`. An interim solution is to wrap the `FlatList` in a `ScrollView` and make the header be a sibling component to the `FlatList`.

mdole:
Closing - deprioritized for saved search v1.0

PR #4721, opened 2021-05-06
feature(analytics): adding analytics to android

pvinis:

This PR resolves [CX-1259]

@mikehrom 

### Description

Slightly more than 2 days but all good. Basically the main changes here are that we add the segment react native lib, we disable it for ios since we have the native one still. I made our `util/track` support more "providers", so one is the native analytics, and another one is the segment RN lib.
Then I had to fix a few tests to work, and I added some tests that unfortunately I have to `skip`, because to make them work I need to rework a bunch of things, so for now the will be there skipped 🤷.

You can try it here https://app.segment.com/artsy-engineering/sources/eigendroid_staging/debugger

[CX-1259]: https://artsyproduct.atlassian.net/browse/CX-1259


src/lib/utils/track/providers.tsx
```diff
@@ -0,0 +1,43 @@
...
+}
+export const isCohesionScreen = (info: CohesionAction | CohesionScreen): info is CohesionScreen =>
+  info.action === ActionType.screen
+
+interface LegacyNameAction {
+  // TODO: This can be removed once we remove these uses. Currently there is one in `WorksForYou` and one in `ArtistRail`.
+  name: string
+}
+
+type InfoType = Schema.PageView | Schema.Entity | CohesionAction | CohesionScreen | LegacyNameAction
```
pvinis, line 22: some temporary typings until cohesion can give these to us.

src/lib/AppRegistry.tsx
```diff
@@ -222,21 +221,9 @@ const InnerPageWrapper: React.FC<PageWrapperProps> = ({ fullBleed, isMainView, V
...
-                <ToastProvider>
-                  <_FancyModalPageWrapper>
-                    <InnerPageWrapper {...this.props} />
-                  </_FancyModalPageWrapper>
-                </ToastProvider>
-              </Theme>
-            </GlobalStoreProvider>
-          </RelayEnvironmentProvider>
-        </ActionSheetProvider>
-      </ProvideScreenDimensions>
```
pvinis, line 239: no change. just rearranging to we have the same for both platforms.

src/lib/utils/track/index.tsx
```diff
@@ -90,7 +88,7 @@ export interface Track<P = any, S = null, T extends Schema.Global = Schema.Entit
 export const track: Track = (trackingInfo, options) => {
   return _track(trackingInfo, {
     ...options,
-    dispatch: (data) => postEvent(data),
+    dispatch: (data) => postEventToProviders(data),
```
pvinis, line 91: here's the main change. we don't just post to native anymore. we post to providers. on ios we add the native analytics as a provider, on android we add segment-rn. we could make the logging into a provider as well, etc.

src/lib/utils/track/__tests__/tracking-tests.tsx
```diff
@@ -0,0 +1,72 @@
...
+            onPress={() => {
+              trackEvent({ wow: "yes!" })
+            }}
+          />
+        </View>
+      </ProvideScreenTracking>
+    )
+  }
+
+  it.skip("works for tracking events with hooks", () => {
```
pvinis, line 29: skipped because its so hard to fix all our tests right now.

ds300: nonblocking: This file explicitly tests the infrastructure you've refactored so it seems a little odd to me that the tests are just being disabled. If you don't think the tests are valuable any more maybe we could delete the file?

src/lib/utils/track/index.tsx
ds300: nonblocking: I feel like the word 'provider' here is a bit overloaded with react providers. A more standard pattern for this kind of thing might be to have a single EventEmitter for analytics events, and then to add eventListeners for each thing you  currently call a 'provider'.

```ts
const analyticsEvents = new EventEmitter<TypesLol>()
if (Platform.OS === 'ios') {
  analyticsEvents.addEventListener('event', (data) => NativeModules.Blah.postEvent(data))
} else {
  analyticsEvents.addEventListener('event', (data) => trackSegmentEvent(data))
  analyticsEvents.addEventListener('identify', (data) => trackSegmentIdentify(data))
}
if (__DEV__) {
  analyticsEvents.addEventListener('event', (data) => console.log('[Event tracked]', data))
}
// .. etc
```

and then this line would be 
```suggestion
    dispatch: (data) => analyticsEvents.emit('event', data),
```

It's OK how it is though

tsconfig.json
```diff
@@ -17,6 +17,9 @@
     "resolveJsonModule": true,
     "strict": true,
     "strictBindCallApply": true,
+    "paths": {
+      "*": ["*", "*.ios", "*.android"]
+    },
```
ds300, line 22: nonblocking: again, I think these platform-specific files make it harder to navigate and maintain the codebase and I would caution against using them.

patches/@segment+analytics-react-native+1.4.4.patch
```diff
@@ -0,0 +1,32 @@
+diff --git a/node_modules/@segment/analytics-react-native/RNAnalytics.podspec b/node_modules/@segment/analytics-react-native/RNAnalytics.podspec
```
ds300, line 1: nonblocking: would it be worth adding a HACKS.md entry for this? might also be pretty obvious 🤔

ds300 (approved):
This is great work! 🎉 💯  I think all my comments are nonblocking so feel free to merge when ready 👍🏼

scripts/ci-ios
```diff
@@ -1,5 +1,4 @@
 #!/usr/bin/env bash
-set -euxo pipefail
```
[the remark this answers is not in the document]
pvinis: it was just me testing on ci :D all good

pvinis:
ah, I had these in automerge, ill make a quick pr to address these, i like the ideas.

patches/@segment+analytics-react-native+1.4.4.patch
pvinis: ill make a ticket for this one, so i skipped the hacks.md.

tsconfig.json
pvinis: I agree. I did start with one file, but I ended up having a bunch of if guards, and also the ios build was complaining for the import. Then I moved the import in an if, and it just got messy. Let me see if i can find it.

src/lib/store/AuthModel.ts
```diff
@@ -194,4 +199,13 @@ export const getAuthModel = (): AuthModel => ({
...
   }),
+  notifyTracking: thunk((_, { userId }) => {
+    SegmentTrackingProvider.identify?.(userId, { is_temporary_user: userId === null ? 1 : 0 })
+  }),
+  didRehydrate: thunkOn(
+    (_, storeActions) => storeActions.rehydrate,
+    (actions, __, store) => {
+      actions.notifyTracking({ userId: store.getState().userID })
+    }
+  ),
```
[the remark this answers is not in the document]
pvinis: They are separate 🤔. So, `identify` is additive, whatever you pass, it will be added, and it will not remove anything from before. So I decided to make the identify call for user interface here, and passing `null` means `no change`, so if it' before login, it's the anonId, if it's after login it's the userId.

Then the one in android app is doing the call only for the id and the temp user key. So they call identify for different "concerns", which I think makes sense. Maybe I could add a comment to both.

src/lib/utils/track/index.tsx
pvinis: huh. interesting. the reverse way. emit events, and whoever cares can listen.

pvinis:
https://github.com/artsy/eigen/pull/4728 next one here. I kept the tests, because i'll make a ticket for it. I kept the patch without hacks.md because i'll make a ticket too. I didn't add the emitter even though i like the idea. It kinda makes things a bit more verbose so I left it for now. Could be a future improvement if it needs more.

PR #4740, opened 2021-05-10
Fix(cx 1371)/phone input misaligned

lordkiz:

This PR resolves [CX-1371]

### Description
TextInput on android by default has underline bar beneath which can cause text to appear slightly padded below. Even though this underline bar can be hidden by passing underlineColorAndroid='transparent', you'd still need to design around it.
Fixed by defining a new prop: `inputTextStyle: TextStyle?` with which you can use to apply textStyling to text inputed - in this case `{paddingTop: UNDERLINE_BAR_HEIGHT}`

| Before |
|----|
|Screenshot 2021-05-10 at 11 06 42|

| After |
|----|
|Screenshot 2021-05-10 at 11 30 30|

[CX-1371]: https://artsyproduct.atlassian.net/browse/CX-1371


src/lib/Components/Input/Input.tsx
```diff
@@ -108,7 +111,7 @@ export const Input = React.forwardRef<TextInput, InputProps>(
               <StyledInput
                 ref={input}
                 placeholderTextColor={color("black60")}
-                style={{ flex: 1, fontFamily: TEXT_FONTS.sans, fontSize: 15 }}
+                style={[{ flex: 1, fontFamily: TEXT_FONTS.sans, fontSize: 15 }, inputTextStyle && inputTextStyle]}
```
MounirDhahri, line 114: **thought(non-blocking):** we prefer to use the following pattern here `{...inputTextStyle}`

MounirDhahri (approved):

src/lib/Components/Input/Input.tsx
lordkiz: Thanks 🙌 . Changed

PR #4741, opened 2021-05-10
refactor: simplify react-navigation infra & add isVisible prop

ds300:
### Description

This PR refactors the react-navigation infra (currently android-only but will hopefully be adopted on iOS soon).

The goal is to make the navigation infrastructure less idiosyncratic so that we can leverage common APIs and patterns in react-navigation, e.g. `useIsFocused()` and updating the navigation state manually for custom navigation logic.

The main architectural changes are:

- We no longer use nested independent `NavigationContainer` instances for tab stacks or modal stacks (it's still used for the `NavigatorIOS` shim though). There is one `NavigationContainer` at the root of the app, and all the screens that can be navigated to by `navigate` are presented within that container.
- We no longer use the custom bottom tabs navigator on android, although it is still used on iOS. On android we now use the standard `@react-navigation/bottom-tabs` navigator. This means that the `selectedTab` state is no longer stored in `GlobalStore` on android. Instead it is derived from the main `NavigationContainer`'s state.

The main user-facing impacts are:

- No more fade transition on the bottom tabs navigator. We could potentially bring it back but it's more trouble than it's worth I think.
- The bottom tabs buttons are a tiny bit less responsive, at least in dev mode, because the react-navigation state change seems to propagate a bit slower than the GlobalStore state change. I think this probably won't be an issue in beta/production builds. If it is, and we think it's significant enough to warrant fixing, it should be easy enough to work around by having some optimistic state somewhere.

### Behaviour to QA

- [x] pushing/popping in a tab stack
- [x] pushing/popping in a modal stack
- [x] presenting/unpresenting modal views
- [x] switching tabs
- [x] dev reload state
- [x] replacing modals
- [x] pop to root on tab press
- [x] goBack
  - [x] inside modal with > 1 view
  - [x] inside modal with 1 view
  - [x] inside tab stack with > 1 view
  - [x] inside non-home tab stack with 1 view (should take you to home)
  - [x] inside home tab stack with 1 view (should background the app)
- [x] dismissModal
  - [x] with > 1 presented modals
  - [x] inside modal with > 1 view 
- [x] popParentViewController (when called from within a modal, should dismiss the topmost view on the selected tab's stack, behind the scenes)

@MounirDhahri do you mind taking a look while I do a final round of QA?


MounirDhahri (approved):

src/lib/store/GlobalStore.tsx
```diff
@@ -97,8 +100,18 @@ export const GlobalStoreProvider: React.FC<{}> = ({ children }) => {
...
+export function useSelectedTab(): BottomTabType {
+  if (Platform.OS === "ios") {
+    return hooks.useStoreState((state) => state.bottomTabs.sessionState.selectedTab)
+  }
+
+  const blah = useNavigationState((state) => state.routes.find((r) => r.state?.type === "tab")?.state)
+  if (!blah) {
+    return "home"
+  } else {
+    const { index, routes } = blah
```
ds300, line 112: whoops I forgot to rename these, will fix

ds300:
@MounirDhahri I finished QAing this, is it good to merge?

MounirDhahri:
Yes, I will merge this in 10 minutes after creating a new build

PR #4753, opened 2021-05-10
feat: allow sticky tabs to add additional sticky header content

mdole:

This PR resolves: [FX-2922] and [CX-1452]

### Description

Following up on https://github.com/artsy/eigen/pull/4703, this PR solves an issue where scrolling down and then back up will result in the "sticky filter bar" hiding behind the tabs.

With this PR, each sticky tab can access a `setJSX` function from `StickyTabPageFlatListContext`, which can be used to add additional sticky content below the existing sticky tabs.

I also took this opportunity to clean up the ArtistShows code. Because the Insights tab is now fully released and shows were moved to the About tab, that code was completely unused.

Still to be resolved:
- [x] Because this can result in headers being different heights, there is an odd "bounce" effect when tapping between tabs. Ideally, we want any additional content below the tabs to behave as though it is part of the content and not the header. Instead of disappearing and reappearing, causing the rest of the content to move downward, it would be pre-rendered (or there's probably another solution here)

tab-updates

cc @iskounen 

[FX-2922]: https://artsyproduct.atlassian.net/browse/FX-2922

[CX-1452]: https://artsyproduct.atlassian.net/browse/CX-1452


mdole:
I'm still having difficulty logging in on Android simulator 😞 I'll try to resolve on Monday, but if anyone wants to poke around sooner, would appreciate the extra eyes!

src/lib/Components/StickyTabPage/StickyTabPage.tsx
```diff
@@ -44,7 +44,15 @@ export const StickyTabPage: React.FC<{
...
   const activeTabIndexNative = useAnimatedValue(initialTabIndex)
   const [activeTabIndex, setActiveTabIndex] = useGlobalState(initialTabIndex)
+  const [tabSpecificStickyHeaderContent, setTabSpecificStickyHeaderContent] = useState<JSX.Element[]>([])
   const { jsx: staticHeader, nativeHeight: staticHeaderHeight } = useAutoCollapsingMeasuredView(staticHeaderContent)
+
+  const stickyRailRef = useRef<SnappyHorizontalRail>(null)
+
+  const tabSpecificStickyHeaderContentArray = tabs.map((_, i) => {
+    return useAutoCollapsingMeasuredView(tabSpecificStickyHeaderContent[i])
+  })
```
mdole, line 54: This breaks the [rules of hooks](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level). It was causing a test failure with the `ArtistShows` tab, which I paired with @dzucconi on. We confirmed that moving this out of a map (temporarily hardcoding it as if we knew how many tabs we were rendering in every case) and adding some `act()` calls to the test got us to a passing state.

However, we weren't sure how to avoid using `.map` in this case. We _think_ it's not actually going to break anything, and I removed the failing test because it was dead code anyway. But I'm open to feedback if there's a better way to resolve this while not breaking the rules of hooks!

src/lib/store/config/features.ts
```diff
@@ -45,10 +45,6 @@ export const features = defineFeatures({
     readyForRelease: true,
     echoFlagKey: "AROptionsNewFirstInquiry",
   },
-  AROptionsNewArtistInsightsPage: {
-    readyForRelease: true,
-    echoFlagKey: "AROptionsNewArtistInsightsPage",
-  },
```
mdole, line 51: Is there anything else that needs to be done to deprecate a feature flag?

pvinis: Remove this, and all its uses. I would say as the last step, we could also do this https://github.com/artsy/echo/pull/86. That way, after like a few months, we could delete that flag from echo too.

pvinis:
I'm happy to pair for the android issue.

src/lib/store/config/features.ts
mdole: Nice, thanks @pvinis!

src/lib/Components/StickyTabPage/StickyTabPage.tsx
pvinis: I couldn't run the code myself, but what if you move the map inside the hook? 🤔

mdole: hm interesting idea - wouldn't that defeat the point though? we use this to get `tabSpecificContentHeight` for each individual tab, which is why we need an array of values so we can look up tab height. If we passed all of the tabs into `useAutoCollapsingMeasuredView`, we would end up with a single React element. And actually, I think that's what we had beforehand, which was causing the "bounce" problem :/

mdole: @damassi I think of you as something of a hooks expert - got any opinions here?

pvinis: Hm. That makes sense. Maybe we could (all?) do a little 🍐ing on this. It would be easier to have the code and run it and play around to find the way to go.

src/lib/Components/StickyTabPage/StickyTabPageFlatList.tsx
```diff
@@ -39,7 +43,9 @@ export const StickyTabPageFlatList: React.FC<StickyTabFlatListProps> = (props) =
   if (!stickyHeaderHeight) {
     throw new Error("invalid state, mounted flat list before stickyHeaderHeight was determined")
   }
-  const { tabIsActive } = useContext(StickyTabPageFlatListContext)
+  const { tabIsActive, tabSpecificContentHeight } = useContext(StickyTabPageFlatListContext)
+
+  const totalStickyHeaderHeight = Animated.add(stickyHeaderHeight, tabSpecificContentHeight ?? 0)
```
MounirDhahri, line 48: I think this should better be
```suggestion
  const totalStickyHeaderHeight = Animated.add(stickyHeaderHeight, Animated.cond(Animated.defined(tabSpecificContentHeight),tabSpecificContentHeight,0) )
```

pvinis: is this `Animated.cond` and friends coming from RN or reanimated?

mdole: @pvinis those are from reanimated! and thanks @MounirDhahri, let's go with that

mdole:
Going ahead and marking #squashongreen since I discussed this with both Mounir and Pavlos earlier!

PR #4763, opened 2021-05-12
fix(CX-1379): “cancel” button doesn’t clear terms and results

JenqaaaBY:
The type of this PR is: fix

This PR resolves [CX-1379]

### Description

This PR fixes the behavior of the cancel button. Now when you press the "cancel" button, text inputs and search results are cleared. Also in this PR, the component was refactored by using a prettier.
https://user-images.githubusercontent.com/30176348/117949037-63cf8c80-b31a-11eb-98a6-f5d50e875939.mov

- [x]  I have included screenshots or videos to illustrate my changes, or I have not changed anything that impacts the UI.

[CX-1379]: https://artsyproduct.atlassian.net/browse/CX-1379


MounirDhahri (approved):

MounirDhahri:
Congratulations on your first contribution to Eigen 🎉. Looks good to me except few minor remarks.
- The first one is about the PR title. Ideally, it should be **fix(CX-1379):  “cancel” button doesn’t clear terms and results** (the ticket between brackets is not mandatory). This helps us have a clean git history with a unified format (see https://github.com/artsy/eigen/commits/master)
- The second one is about the last item on the PR checklist, the changelog update. I think you forgot to add it. We try to keep track of all the changes we do for every version in this changelog file https://github.com/artsy/eigen/blob/master/CHANGELOG.yml that we use also to generate release notes when we submit the app for review.

JenqaaaBY:
> Congratulations on your first contribution to Eigen 🎉. Looks good to me except few minor remarks.
> 
> * The first one is about the PR title. Ideally, it should be **fix(CX-1379):  “cancel” button doesn’t clear terms and results** (the ticket between brackets is not mandatory). This helps us have a clean git history with a unified format (see https://github.com/artsy/eigen/commits/master)
> * The second one is about the last item on the PR checklist, the changelog update. I think you forgot to add it. We try to keep track of all the changes we do for every version in this changelog file https://github.com/artsy/eigen/blob/master/CHANGELOG.yml that we use also to generate release notes when we submit the app for review.

Thanks for your review.
I've changed the PR name and added my changes to the CHANGELOG.yml file

CHANGELOG.yml
```diff
@@ -8,6 +8,7 @@ upcoming:
     - Add staging support for deep links on android - mounir, ole
     - Refactor filter event payloads to send JSON - roop
     - Fix issue with ARSpinner being registered more than once - mounir, kizito
+    - Fix cancel button doesnt clear terms and results - yauheni
```
MounirDhahri, line 11: **thought:** I think this is not needed here since you already included it in the `user_facing` section

MounirDhahri:
Great, thanks for making the changes. I left a minor comment about an extra addition to the changelog. Other than that, this is clear to go. Also there seem to be a changelog conflict

MounirDhahri:
Awesome thank you
#squashongreen

PR #4765, opened 2021-05-12
enhancement(country select): replace the old one with the new one

pvinis:

This PR resolves [CX-1448]

### Description

The old country select was pretty janky. No initial list, the user had to type something first. The search was pretty bad too because it used the google places autocomplete API that returns "regions", which means countries as well as other things. So typing `gree` gave no results in the selector, because google returned `greenville` and other non-country things, that we then filter out.

I removed it completely and replaced it with the way nicer one we already use in Payments.

The only issue here is that the fonts are slightly off on the RegisterToBid and Bid screens now, but it's an ok tradeoff, since we will redesign that one at some point anyway.

Some screenshots:

old and bad
Screenshot 2021-05-11 at 12 51 03

old and bad
Screenshot 2021-05-11 at 12 51 09

new and good
Screenshot 2021-05-11 at 12 51 17

new and good
Screenshot 2021-05-11 at 12 51 26

Here they are together. Top is new, bottom is old. The font used in the new one is not taking into account the `Theme`. I started fixing that, but then the PR would have grown, so I kept it simple. It's a fair tradeoff.
Screenshot 2021-05-11 at 12 53 09

(I have stopped adding things to changelog since like 2 weeks ago. The bright automatic future is almost here, and I can't wait 😄!)

[CX-1448]: https://artsyproduct.atlassian.net/browse/CX-1448


ds300 (approved):
nice nice

MounirDhahri:
nice 🎉
@pvinis do you mind if you create a follow-up ticket and add it to the Jira board

PR #4780, opened 2021-05-14
feat(FX-2899): convert filter queries to input

dimatretyak:

This PR resolves [FX-2899]

### Description
Convert filter queries to use a single structured input for all `filterArtworksConnection` requests. This should be **carefully QA'ed**.

Web version of this change [artsy/force#7151](https://github.com/artsy/force/pull/7151)

[FX-2899]: https://artsyproduct.atlassian.net/browse/FX-2899

### DIFF
--- a/src/lib/Components/ArtworkFilter/ArtworkFilterHelpers.ts
+++ b/src/lib/Components/ArtworkFilter/ArtworkFilterHelpers.ts
@@ -261,7 +261,6 @@ export const changedFiltersParams = (currentFilterParams: FilterParams, selected
 }
 
 export const filterArtworksParams = (appliedFilters: FilterArray, filterType: FilterType = "artwork") => {
-  // console.log("[variables] filterArtworksParams", appliedFilters, filterType)
   const defaultFilterParams = getDefaultParamsByType(filterType)
   return paramsFromAppliedFilters(appliedFilters, { ...defaultFilterParams }, filterType)
 }
--- a/src/lib/Components/ArtworkFilter/ArtworkFilterHelpers.ts
+++ b/src/lib/Components/ArtworkFilter/ArtworkFilterHelpers.ts
@@ -1,5 +1,5 @@
 import { FilterScreen } from "lib/Components/ArtworkFilter"
-import { capitalize, compact, forOwn, groupBy, sortBy } from "lodash"
+import { capitalize, compact, forOwn, groupBy, pick, sortBy } from "lodash"
 
 export enum FilterDisplayName {
   // artist = "Artists",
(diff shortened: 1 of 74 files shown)


src/lib/Components/ArtworkFilter/ArtworkFilterHelpers.ts
```diff
@@ -453,3 +454,55 @@ export const getDisplayNameForTimePeriod = (aggregationName: string) => {
 
   return DISPLAY_TEXT[aggregationName] ?? aggregationName
 }
+
+export const prepareFilterArtworksParamsForInput = (filters: FilterParams) => {
```
dimatretyak, line 457: The default filter params may contain params that cannot be passed to `FilterArtworksInput` (for example `estimateRange` from `DEFAULT_ARTWORKS_PARAMS`)

src/lib/Scenes/Artist/Artist.tsx
```diff
@@ -133,14 +133,20 @@ export const ArtistQueryRenderer: React.FC<ArtistQueryRendererProps> = ({ artist
...
               auctionResultsConnection {
                 totalCount
               }
             }
           }
         `,
-        variables: { artistID },
+        variables: {
+          artistID,
+          input: {
```
dimatretyak, line 145: Can we pass default arguments this way?

dimatretyak:
Should I convert filter params to `$input` in all places of `filterArtworksConnection` call? For example [this](https://github.com/artsy/eigen/blob/dimatretyak/convert-filter-queries-to-input/src/lib/Containers/Gene.tsx#L325)

src/lib/Scenes/Artist/Artist.tsx
dzucconi: Hm, I know you can pass default values for individual arguments within the query/fragment itself but I suppose that might be weird with the structured input — I'm having a bit of trouble finding examples of that. No problem mixing in defaults via variables though.

src/lib/Components/ArtworkFilter/ArtworkFilterHelpers.ts
dzucconi: 👍

dzucconi:
> Should I convert filter params to $input in all places of filterArtworksConnection call? For example this

Yeah, ideally any place where possible we convert to the input type.

brainbicycle (commented):
✨ much cleaner, nice work!

src/lib/Components/ArtworkFilter/ArtworkFilterHelpers.ts
brainbicycle: **thought (non-blocking)**: 🤔 are DEFAULT_ARTWORKS_PARAMS combining multiple potentially unrelated types? estimateRange is for saleArtworks but these don't use same FilterArtworksInput, should defaults just be separated? Ignore me if that makes no sense, thanks for cleaning up this code.

src/lib/Scenes/ArtistSeries/ArtistSeries.tsx
```diff
@@ -160,7 +160,10 @@ export const ArtistSeriesFragmentContainer = createFragmentContainer(ArtistSerie
 
       ...ArtistSeriesHeader_artistSeries
       ...ArtistSeriesMeta_artistSeries
-      ...ArtistSeriesArtworks_artistSeries
+      ...ArtistSeriesArtworks_artistSeries @arguments(input: {
+        sort: "-decayed_merch",
+        dimensionRange: "*-*"
+      })
```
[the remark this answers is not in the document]
dzucconi: Totally fine to inline arguments where they don't change

src/lib/Scenes/Fair/Fair.tsx
```diff
@@ -316,13 +317,19 @@ export const FairQueryRenderer: React.FC<FairQueryRendererProps> = ({ fairID })
...
           }
         }
       `}
-      variables={{ fairID }}
+      variables={{
+        fairID,
+        input: {
+          sort: "-decayed_merch",
+          dimensionRange: "*-*",
+        }
```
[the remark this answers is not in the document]
dzucconi: Either way works. I think for things where the arguments are hardcoded it makes more sense to just inline them. But for things like that we don't even really need to worry about it right now. This ticket is mainly concerned with making future filter updates simple (so when adding a new filter we don't need to redeclare arguments everywhere) — on things that already just have hardcoded args, this won't make a difference.

dzucconi (approved):
Looks great — thanks! Resolve the (annoying) changelog conflict and we'll get this merged.

PR #4781, opened 2021-05-14
fix: add webViewRoute for buyer guarantee page

xtina-starr:

This PR resolves [PURCHASE-2665]

### Description

This adds an add webViewRoute for the buyer guarantee page.

[PURCHASE-2665]: https://artsyproduct.atlassian.net/browse/PURCHASE-2665

### DIFF
--- a/src/lib/Scenes/Inbox/Components/Conversations/OpenInquiryModalButton.tsx
+++ b/src/lib/Scenes/Inbox/Components/Conversations/OpenInquiryModalButton.tsx
@@ -30,7 +30,7 @@ export const OpenInquiryModalButton: React.FC<OpenInquiryModalButtonProps> = ({ 
                 color="black100"
                 variant="small"
                 onPress={() => {
-                  navigate(`buyer-guarantee`)
+                  navigate(`/buyer-guarantee`)
                 }}
               >
                 The Artsy Guarantee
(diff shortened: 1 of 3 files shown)


xtina-starr:
This is difficult to verify because the admin page is currently behind an admin only flag and you can't make an inquiry as an admin. Any ideas on how we could get around this?

starsirius:
Hmm the buyer guarantee page renders the footer when logged out. Do we see that in the webview with this PR? If so, we can at least make sure it renders the page and can iterate on the Force view independently. And I think we have a back button on the modal that users can navigate back, right?

src/lib/Scenes/Inbox/Components/Conversations/OpenInquiryModalButton.tsx
```diff
@@ -30,7 +30,7 @@ export const OpenInquiryModalButton: React.FC<OpenInquiryModalButtonProps> = ({
                 color="black100"
                 variant="small"
                 onPress={() => {
-                  navigate(`buyer-guarantee`)
+                  navigate(`/buyer-guarantee`)
```
starsirius, line 33: It seems to still work without the leading slash, but adding it to be consistent.

starsirius (approved):
I managed to verify it by submitting an inquiry from a non-admin user and then making the user an admin to view the buyer guarantee webview page. There is some extra padding on the top but I assume that can be fixed in Force directly.

I also committed a [minor change to the route](https://github.com/artsy/eigen/pull/4781/commits/89db592948677d452aa01c4299c1b348bc324537) and make it more consistent. Hope that makes sense.

LGTM and will let @brainbicycle 🕹️ ! Thanks @xtina-starr 

buyer-guarantee

sepans (approved):
Thanks, @xtina-starr, and @starsirius for verifying. 💯

xtina-starr:
Thanks @starsirius!!! @sepans @brainbicycle can we go ahead and merge this? That way it will be available to QA in the beta for tomorrow.

sepans:
Makes sense. merging. @brainbicycle if you have any comments or objections we can do a quick follow-up tomorrow.

PR #4782, opened 2021-05-15
feat: enable Inquiry Checkout (PURCHASE-2684)

starsirius:

This enables Inquiry Checkout. The corresponding Echo PR is [here](https://github.com/artsy/echo/pull/85).

https://artsyproduct.atlassian.net/browse/PURCHASE-2684

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -31,6 +31,7 @@ upcoming:
     - Fix spaces between items in SaleList - yauheni
     - Inquiry checkout copy additions from legal - lily
     - Fix artwork unavailable issue on the conversation screen - starsirius
+    - Enable Inquiry Checkout - starsirius
 
 releases:
   - version: 6.9.0
--- a/src/lib/store/config/features.ts
+++ b/src/lib/store/config/features.ts
@@ -50,7 +50,8 @@ export const features = defineFeatures({
     echoFlagKey: "AROptionsNewArtistInsightsPage",
   },
   AROptionsInquiryCheckout: {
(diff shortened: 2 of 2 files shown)


starsirius:
CI failed because Echo hasn't had [the new feature key](https://github.com/artsy/echo/pull/85). This makes me wonder if we should

1. Set the new feature to `false` in [the Echo PR](https://github.com/artsy/echo/pull/85), merge and deploy that.
2. Rebuild this PR, merge and submit for approval
3. When we are ready to release Inquiry Checkout, update the Echo flag to `true`.

Would love to pair on this and learn more about the sequencing!

Screen Shot 2021-05-16 at 10 46 28 PM

pvinis:
But in general, I think merging the echo PR should make this PR here pass, and it will be good to go.

ds300:
yeah this test is to make sure that the echo flag exists before deploying.

@starsirius the steps you outlined are exactly how we suggest to do things for _timed releases_. If you don't care about whether the feature becomes available on a particular day, you can also set the echo flag to `true` and then the feature will be switched on as soon as people update their apps to the version that this PR gets deployed in.

starsirius:
Thanks @pvinis and @ds300! I submitted another [Echo PR](https://github.com/artsy/echo/pull/91) to disable the feature for now. Will rebuild this PR again once it's deployed! 🙏

pvinis:
Could you make this PR into a draft, instead of using the [don't merge] title? It makes it clearer for us 🙏.

PR #4783, opened 2021-05-17
docs(flags): removing a feature flag

pvinis:
The type of this PR is: **Documentation**

This PR resolves [CX-]

### Description

This came up in https://github.com/artsy/eigen/pull/4753#discussion_r633201657, so I made a quick addition in docs to have this for the future.

### DIFF
--- a/docs/developing_a_feature.md
+++ b/docs/developing_a_feature.md
@@ -102,6 +102,10 @@ You can find documentation about how to do QA on the new shiny feature [here](ht
 
 There is also the general QA page [here](https://www.notion.so/artsy/QA-decba0c3a57a4508b726f3a8624ceca3).
 
+## Removing a Feature Flag
+
+Once a feature flag is no longer needed, it can be removed from the `features.ts` file. If that feature has no `echoFlagKey`, then removing it from `features.ts` and all the uses of the flag is all you need to do. If it **does** have an `echoFlagKey`, then a small PR on [echo](https://github.com/artsy/echo) is needed after your PR on eigen is done. You can find an [example here](https://github.com/artsy/echo/pull/86). This is needed so we can remove that flag from echo as well, after giving it a few months buffer, for all users to get the new version of the app with the flag removed.
+
 ## Still Need Help?
 
 Ask for help in the #practice-mobile slack channel, we will be happy to assist!


docs/developing_a_feature.md
```diff
@@ -102,6 +102,10 @@ You can find documentation about how to do QA on the new shiny feature [here](ht
 
 There is also the general QA page [here](https://www.notion.so/artsy/QA-decba0c3a57a4508b726f3a8624ceca3).
 
+## Removing a Feature Flag
+
+Once a feature flag is no longer needed, it can be removed from the `features.ts` file. If that feature has no `echoFlagKey`, then removing it from `features.ts` and all the uses of the flag is all you need to do. If it **does** have an `echoFlagKey`, then a small PR on [echo](https://github.com/artsy/echo) is needed after your PR on eigen is done. You can find an [example here](https://github.com/artsy/echo/pull/86). This is needed so we can remove that flag from echo as well, after giving it a few months buffer, for all users to get the new version of the app with the flag removed.
+
```
ds300, line 108: I think it's not safe to remove echo flags that existed before the TS feature flag migration, so we should probably mention that.

Otherwise it's totally safe to remove the echo flag at any time, because it is the same as setting it to `true` forever. You don't need to wait a few months or anything.

It might also be worth mentioning that after deleting the flag in features.ts they will probably need to go and remove any branching logic based on the feature flag.

pvinis: For the last part, that's what the "all the uses of the flag" meant, and I'm sure the builds will fail if that's not done, so I guess that's alright?

For the second, hm, that makes sense. I just checked the code https://github.com/artsy/eigen/blob/7ae0c7e8aaa37ad78b4e3ffe9388cc3a15740e27/src/lib/store/config/FeaturesModel.tsx#L39-L42 and yep, makes sense. If the flag on echo was true, should be safe to remove whenever.

For the first part, that's because these old flags could still be used with that old version of the app? If that's the case, I could go look for the version of the app that did the migration, and I can add some comments/docs. I could also check how many users still use the version before the migration. Does that make sense?

ds300: > For the last part, that's what the "all the uses of the flag" meant, and I'm sure the builds will fail if that's not done, so I guess that's alright?

Ah yeah sorry, my bad, I missed that part of the sentence.

> that's because these old flags could still be used with that old version of the app?

Yeah, and until we stop supporting those older versions of the app we need to keep those echo flags around.

pvinis: Based on https://github.com/artsy/eigen/pull/4400, seems like the latest "old version" is `6.7.7`, and the first version with ts flags is `6.7.8`. I'll take a look to see how many users we have on that version first.

pvinis: https://artsy.looker.com/explore/frontend_analytics/eigen_tracks?qid=SZXeeacYh6Hq2CnE9ooGDl&origin_space=712&toggle=fil,vis shows that there are 900 people in the last 30 days, some of which could have updated by now. I can put a reminder on slack to check this again in another month. I will check also which flags where used in that version and mark them on echo.

pvinis: I wonder if we had some default handling of missing echo flags back then..

PR #4785, opened 2021-05-17
cleanup(old echo): removed some methods and code that were not used

pvinis:

This PR resolves [CX-]

### Description

In order to remove aerodramus completely, I started by removing some code that was not used anywhere in the app. It was an easy quick step. The rest is now used in a few places. Once the old admin menu is migrated, we could probably do another step of cleanup.

### DIFF
--- a/Artsy/App/ArtsyEcho.h
+++ b/Artsy/App/ArtsyEcho.h
@@ -6,7 +6,5 @@
 - (instancetype)init;
 - (NSDictionary *)featuresMap;
 - (BOOL)isFeatureEnabled:(NSString *)featureFlag;
-- (NSArray *)legacyFairSlugs;
-- (NSArray *)legacyFairProfileSlugs;
 
 @end
--- a/Artsy/App/ArtsyEcho.h
+++ b/Artsy/App/ArtsyEcho.h
@@ -4,7 +4,6 @@
 @interface ArtsyEcho : Aerodramus
 
 - (instancetype)init;
-- (NSDictionary *)featuresMap;
 - (BOOL)isFeatureEnabled:(NSString *)featureFlag;
 
 @end
--- a/src/lib/store/config/EchoModel.tsx
+++ b/src/lib/store/config/EchoModel.tsx
@@ -26,8 +26,6 @@ export interface EchoModel {
   fetchRemoteEcho: Thunk<EchoModel>
   didRehydrate: ThunkOn<EchoModel, {}, GlobalStoreModel>
   stripePublishableKey: Computed<EchoModel, string, GlobalStoreModel>
-  legacyFairSlugs: Computed<EchoModel, string[]>
-  legacyFairProfileSlugs: Computed<EchoModel, string[]>
   forceUpdateMessage: Computed<EchoModel, string | undefined>
 }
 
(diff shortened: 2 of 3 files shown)


Artsy/App/ArtsyEcho.h
```diff
@@ -4,9 +4,5 @@
 @interface ArtsyEcho : Aerodramus
 
 - (instancetype)init;
-- (NSDictionary *)featuresMap;
```
pvinis, line 7: not used anywhere

Artsy/App/ArtsyEcho.h
```diff
@@ -4,9 +4,5 @@
...
-- (BOOL)isFeatureEnabled:(NSString *)featureFlag;
-- (NSArray *)legacyFairSlugs;
-- (NSArray *)legacyFairProfileSlugs;
```
pvinis, line 10: not used anywhere

src/lib/store/config/EchoModel.tsx
```diff
@@ -67,12 +65,6 @@ export const getEchoModel = (): EchoModel => ({
     const key = env === "production" ? "StripeProductionPublishableKey" : "StripeStagingPublishableKey"
     return state.state.messages.find((e) => e.name === key)?.content!
   }),
-  legacyFairProfileSlugs: computed((state) => {
-    return state.state.messages.find((e) => e.name === "LegacyFairProfileSlugs")?.content.split(",")!
-  }),
-  legacyFairSlugs: computed((state) => {
-    return state.state.messages.find((e) => e.name === "LegacyFairSlugs")?.content.split(",")!
-  }),
```
pvinis, line 75: I will follow up with an echo PR to mark these as deprecated, and we can remove them in the future.

Artsy/App/ARScreenPresenterModule.m
```diff
@@ -234,13 +234,7 @@ + (UIViewController *)currentlyPresentedVC
 
 + (UIViewController *)loadAuctionWithID:(NSString *)saleID
 {
-    if ([[[ARAppDelegate sharedInstance] echo] isFeatureEnabled:@"DisableNativeAuctions"] == YES) {
```
pvinis, line 237: this is the only usage of `isFeatureEnabled`. The string `DisableNativeAuctions` was not found anywhere else in the codebase or echo, and `isFeatureEnabled` is defaulting to `NO`, so I went ahead and removed the if part and kept the else part, since I guess that's the flow the app uses right now.

brainbicycle: I believe the reason this was kept so long is that Eigen's native auctions feature needs to stay in lockstep with causality's version because it uses web sockets. This way if we have a breaking change we can always fall back to the webview. I am not 100% on how versioning in causality works so I could be wrong but think we probably want to do some investigation before we remove this and might want to keep for now that there are changes on horizon for bidding engine.

brainbicycle (changes requested):
I am a little bit paranoid about removing the "DisableNativeAuctions" flag, so would be in favor of keeping and removing the rest.

pvinis:
Makes sense. Ok I will only remove the smaller chunk and keep that auctions part. I didn't see the string anywhere so I thought we don't even have a way to change it. Well, I guess we could add it in echo, but I thought if we didn't have it already there, then probably it's fine to remove.

pvinis:
This shoudl be fine now I think.

brainbicycle (approved):

PR #4794, opened 2021-05-18
[POC] react navigation ios

ds300:
### Description

This PR is a demonstration of the steps needed to consolidate the navigation infrastructure across iOS and Android.

This gets it about 90% of the way there, there's some tricky infra issues to overcome:

- if we use `createNativeStackNavigator` (i.e. react-native-screens) for the outer modal stack, then we run into a couple of serious issues:
  - touches don't work on full screen modals (they do if you set the modal type to 'transparent full screen' but that's not an option on iOS 12). I debugged this for a while but failed to pinpoint where the touches were being lost.
  - safe area insets were a bit glitchy.
  
   Meanwhile, using the regular `createStackNavigator` for the outer modal stack (as I left it in this PR) has a couple of issues

  - safe area insets are not propagated to native view controllers embedded in the modal (e.g. as needed by live auctions)
  - there is no 'page sheet' style modal, so all modals would be full screen.

Also I just noticed that this PR fails to retain the `scrollToTop` functionality, but that should be easy enough to reinstate.


src/lib/utils/useDidRegainFocus.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export const OnDidRegainFocus: React.FC<{ onRegainFocus(): void }> = ({ onRegainFocus }) => {
+  const didRegainFocus = useDidRegainFocus()
+  useEffect(() => {
+    if (didRegainFocus) {
+      onRegainFocus()
+    }
+  }, [didRegainFocus, onRegainFocus])
+  return null
+}
```
ds300, line 27: our primary use-case for the `isVisible` prop has been to refetch data when screens regain focus. We also had the same requirement recently for the inquiry checkout flow. So I mad this to take care of that use-case.

src/lib/utils/useDidRegainFocus.tsx
```diff
@@ -0,0 +1,27 @@
...
+  const isVisible = useIsFocused()
+  const lastIsVisible = usePrevious(isVisible) ?? isVisible
+  const hasBeenVisibleAtLeastOnce = useRef(false)
+  const didJustBecomeVisible = isVisible && !lastIsVisible
+  const didRegainFocus = didJustBecomeVisible && hasBeenVisibleAtLeastOnce.current
+
+  if (isVisible) {
+    hasBeenVisibleAtLeastOnce.current = true
+  }
+
```
ds300, line 15: this could also be simplified to
```suggestion
  const isVisible = useIsFocused()
  const lastIsVisible = usePrevious(isVisible) ?? true
  const didRegainFocus = isVisible && !lastIsVisible
```

but it's a little less obvious how it works like this.

MounirDhahri:
Closing PR for inactivity - please feel free to reopen it if needed 👌 Thanks David and Brian for working on this

PR #4797, opened 2021-05-18
fix: don't rerender home when logged out

brainbicycle:

This PR resolves [CX-1311]

### Description

All of our tabs are rerendered any time we switch tabs, on logging out we reset the tab state to home causing a rerender of the home tab which fails because at this point the users access token is gone leading to the unable to load screen. This checks for presence of access token and doesn't render the home screen if it is absent.

[CX-1311]: https://artsyproduct.atlassian.net/browse/CX-1311


MounirDhahri (approved):

MounirDhahri:
Finally we fixed this 🎉

PR #4803, opened 2021-05-19
feat: add new route for order history & render order's id PX-4062

alexj105:
Please,  check my tests especially the mocking data request. If I'm wrong, please give me a hint about it!

The type of this PR is Feature

This PR resolves PX-4062
https://artsyproduct.atlassian.net/browse/PX-4062

ezgif com-gif-maker


brainbicycle (changes requested):
Code looks great, excited to get order history in the app! I think we will need to add a feature flag for this to hide it until it is ready to ship, there are docs here but let me know if anything is unclear!
https://github.com/artsy/eigen/blob/master/docs/developing_a_feature.md

alexj105:
> Code looks great, excited to get order history in the app! I think we will need to add a feature flag for this to hide it until it is ready to ship, there are docs here but let me know if anything is unclear!
> https://github.com/artsy/eigen/blob/master/docs/developing_a_feature.md

https://artsyproduct.atlassian.net/browse/PX-4061
It's already done and merged. See https://github.com/artsy/eigen/pull/4795

brainbicycle:
> > Code looks great, excited to get order history in the app! I think we will need to add a feature flag for this to hide it until it is ready to ship, there are docs here but let me know if anything is unclear!
> > https://github.com/artsy/eigen/blob/master/docs/developing_a_feature.md
> 
> https://artsyproduct.atlassian.net/browse/PX-4061
> It's already done and merged. See #4795

Great! but it doesn't like this code is behind the flag unless I am missing it somewhere? For example I would expect the MenuItem in the profile to be only present if the feature is turned on.

src/lib/Scenes/MyProfile/MyProfile.tsx
```diff
@@ -47,6 +47,7 @@ const MyProfile: React.FC<{ me: MyProfile_me; relay: RelayRefetchProp }> = ({ me
       <Separator mt={3} mb={2} />
       <SectionHeading title="Account Settings" />
       <MenuItem title="Account" onPress={() => navigate("my-account")} />
+      <MenuItem title="Order History" onPress={() => navigate("/orders")} />
```
sweir27, line 50: @alexj105 this is where we need to make sure we're checking the value on the feature flag.

It'll be something like this--
```
const showOrderHistory = useFeatureFlag("AREnableOrderHistoryOption")
....
{ showOrderHistory && <MenuItem title="Order History" onPress={() => navigate("/orders")} /> }
```

(cc @brainbicycle in case I messed up!)

alexj105: Good point! It's clear now!

brainbicycle (approved):
Thanks for adding the feature flag, nice work!

src/lib/store/config/features.ts
```diff
@@ -33,6 +33,10 @@ function defineFeatures<T extends string>(featureMap: { readonly [featureName in
 export type FeatureName = keyof typeof features
 
 export const features = defineFeatures({
+  AREnableOrderHistoryOption: {
+    readyForRelease: false,
+    echoFlagKey: "AREnableOrderHistoryOption",
```
brainbicycle, line 38: You can leave out the echoFlagKey until this is ready for release and the tests should pass! Echo is our config server so if we have the key set here it is expected to be in echo as well!
```suggestion
```

kiryl-zubarau (approved):

brainbicycle:
> Thanks for adding the feature flag, nice work! One last request, could you add an entry to CHANGELOG.yml for this?

Went ahead and added an entry here: https://github.com/artsy/eigen/pull/4803/commits/19458c98951f2de8b9cdcf78f18c1aae516e865e

PR #4814, opened 2021-05-19
fix: ensure make offer butten refetches when conversation is pulled to refetched

xtina-starr:

This PR resolves [PURCHASE-2616]

### Description

Because the Make Offer button uses a separate QueryRenderer, when the <Conversation> is refetched via pull-to-refresh the button isn't being updated when an artwork is marked sold. This PR wraps the <OpenInquiryModalButton> in a RefetchContainer and triggers a refetch through an event emitter inside the Conversation component.

https://user-images.githubusercontent.com/5201004/118890201-49725180-b8cc-11eb-8293-d653dc986dd4.mp4

[PURCHASE-2616]: https://artsyproduct.atlassian.net/browse/PURCHASE-2616

### DIFF
--- a/src/lib/Scenes/Inbox/Components/Conversations/OpenInquiryModalButton.tsx
+++ b/src/lib/Scenes/Inbox/Components/Conversations/OpenInquiryModalButton.tsx
@@ -1,21 +1,47 @@
 import { tappedMakeOffer } from "@artsy/cohesion"
 import { captureMessage } from "@sentry/react-native"
 import { OpenInquiryModalButtonQuery } from "__generated__/OpenInquiryModalButtonQuery.graphql"
-import { navigate } from "lib/navigation/navigate"
+import { navigate, navigationEvents } from "lib/navigation/navigate"
 import { defaultEnvironment } from "lib/relay/createEnvironment"
 import { Button, CheckCircleIcon, Flex, Text } from "palette"
-import React from "react"
-import { graphql, QueryRenderer } from "react-relay"
+import React, { useEffect } from "react"
+import { createRefetchContainer, graphql, QueryRenderer, RelayRefetchProp } from "react-relay"
 import { useTracking } from "react-tracking"
 import { ShadowSeparator } from "../ShadowSeparator"
 
 export interface OpenInquiryModalButtonProps {
   artworkID: string
   conversationID: string | null | undefined
+  relay: RelayRefetchProp
 }
 
-export const OpenInquiryModalButton: React.FC<OpenInquiryModalButtonProps> = ({ artworkID, conversationID }) => {
+export const OpenInquiryModalButton: React.FC<OpenInquiryModalButtonProps> = ({ artworkID, conversationID, relay }) => {
   const { trackEvent } = useTracking()
+
+  useEffect(() => {
+    navigationEvents.addListener("refetchConversation", handleConversationRefresh)
+    return () => {
+      navigationEvents.removeListener("refetchConversation", handleConversationRefresh)
+    }
+  }, [])
+
+  const handleConversationRefresh = () => {
+    refetch()
+  }
+
+  const refetch = () => {
+    // console.warn("refetch got called")
+    relay.refetch(
+      { artworkID },
+      null,
+      (error) => {
+        if (error) {
+          console.error("OpenInquiryMoodalButton.tsx", error.message)
+        }
+      },
+      { force: true }
+    )
+  }
 
   return (
     <>
@@ -60,6 +86,24 @@ export const OpenInquiryModalButton: React.FC<OpenInquiryModalButtonProps> = ({ 
   )
 }
 
+export const OpenInquiryModalButtonContainer = createRefetchContainer(
+  OpenInquiryModalButton,
+  {
+    artwork: graphql`
+      fragment OpenInquiryModalButton_artwork on Artwork {
+        isOfferableFromInquiry
+      }
+    `,
+  },
+  graphql`
+    query OpenInquiryModalButtonRefetchQuery($artworkID: String!) {
+      artwork(id: $artworkID) {
+        isOfferableFromInquiry
+      }
+    }
+  `
+)
+
 export const OpenInquiryModalButtonQueryRenderer: React.FC<{
   artworkID: string
   conversationID: string
@@ -88,7 +132,7 @@ export const OpenInquiryModalButtonQueryRenderer: React.FC<{
         }
 
         if (props?.artwork?.isOfferableFromInquiry) {
-          return <OpenInquiryModalButton artworkID={artworkID} conversationID={conversationID} />
+          return <OpenInquiryModalButtonContainer artworkID={artworkID} conversationID={conversationID} />
         } else {
           return null
         }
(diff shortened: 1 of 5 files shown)


src/lib/Scenes/Inbox/Components/Conversations/__tests__/OpenInquiryModalButton-tests.tsx
```diff
@@ -87,7 +87,7 @@ describe("OpenInquiryModalButtonQueryRenderer", () => {
       const tree = getWrapper()
 
       expect(extractText(tree.root)).toContain("Only purchases completed with our secure checkout are protected")
-      expect(tree.root.findAllByType(OpenInquiryModalButton)).toHaveLength(1)
+      expect(tree.root.findAllByType(OpenInquiryModalButtonContainer)).toHaveLength(1)
```
xtina-starr, line 90: I'm trying to figure out a good testing pattern for testing this refetch mechanism. I was thinking something the following but I was running into issues when spying on the method inside the functional component. Open to idea for this.

```tsx
it("handles button refetch when conversation is refetch", () => {
    const makeOfferButton = getWrapper()
    const componentInstance = (makeOfferButton.root.findByType(OpenInquiryModalButtonContainer)
      .children[0] as ReactTestInstance).instance

    console.log("componentInstance:", componentInstance.refetch)
    console.log("jest.spyOn:", jest.spyOn)
    jest.spyOn(componentInstance, "refetch")
    navigationEvents.emit("refetchConversation")

    expect(componentInstance.refetch).toHaveBeenCalled()
})
```

xtina-starr: @damassi do you have any ideas on how to possibly test this?

src/lib/Scenes/Inbox/Screens/Conversation.tsx
```diff
@@ -233,6 +235,7 @@ export class Conversation extends React.Component<Props, State> {
                 },
                 { force: true }
               )
+              navigationEvents.emit("refetchConversation")
```
erikdstock, line 238: it looks like L228-238 is identical to calling `this.refetch()` but i'm not sure. Can we dry this up?

src/lib/Scenes/Inbox/Screens/Conversation.tsx
```diff
@@ -108,6 +108,8 @@ export class Conversation extends React.Component<Props, State> {
       },
       { force: true }
     )
+    navigationEvents.emit("refetchConversation")
```
erikdstock, line 111: _Thinking through this notes:_ So in this file we are emitting the new event when a modal is closed or when the `Messages` component refreshes (using pull>refresh).

src/lib/Scenes/Inbox/Components/Conversations/OpenInquiryModalButton.tsx
```diff
@@ -1,22 +1,48 @@
...
   conversationID: string | null | undefined
+  relay: RelayRefetchProp
 }
 
-export const OpenInquiryModalButton: React.FC<OpenInquiryModalButtonProps> = ({ artworkID, conversationID }) => {
+export const OpenInquiryModalButton: React.FC<OpenInquiryModalButtonProps> = ({ artworkID, conversationID, relay }) => {
   const { trackEvent } = useTracking()
 
+  useEffect(() => {
+    navigationEvents.addListener("refetchConversation", handleConversationRefresh)
```
erikdstock, line 22: _Thinking through this notes pt 2_: And here we listen for that new event and refresh the button data. OK, I get it now.

src/lib/Scenes/Inbox/Components/Conversations/OpenInquiryModalButton.tsx
```diff
@@ -1,22 +1,48 @@
...
+    return () => {
+      navigationEvents.removeListener("refetchConversation", handleConversationRefresh)
+    }
+  }, [])
+
+  const handleConversationRefresh = () => {
+    refetch()
```
erikdstock, line 29: This function that calls another function seems unnecessary

erikdstock (commented):
I had a little trouble remembering the various flows that can cause a conversation to refresh so stepped through the code to jog my memory. The conversation screen looks like this (simplified):
```tsx
<Composer>
  <Conversation>
    <Messages>
  </Conversation
  <ConversationCTA />
  <Text Input />
```

Basically the changes in `Conversation.tsx` capture the points where refreshes happen - it's either on (a) a navigation event (the `modalDismissed` event) or (b) when `Messages` are pulled to refresh. We want to make sure that the current CTA is also refreshed  for both (a) & (b), and are addressing the issue by adding a second event, `refetchConversation`, to both of these call sites, so now we refetch the `Conversation` query for latest Messages _and_ dispatch the new event `refetchConversation` to let the `OpenInquiryModalButton` know it should refetch. A few thoughts based on this:
- It seems like the second event may be unnecessary. Can the CTA just listen for the same `modalDismissed` event? Ah, maybe not, because the messages can also be pulled-to-refresh. If that is the case I'd prefer a more event-style name like `messagesRefetched`, which the CTA can listen for.
- The `OpenInquiryModalButton` (which says 'Make Offer') is only one of several possible CTAs. Do we want to refetch the entire `ConversationCTA` which can render other banners as well? I remember us starting on this change a week or so before I was out so this may be wrong.

Based on these assumptions I would
- emit `messagesRefetched` after refetching messages, possibly DRYing out the code where we call it in 2 different places (I think it is exactly the same, see comment)
- consider moving the second query renderer specific to the CTA from `OpenInquiryModalButton` into its enclosing `ConversationCTA` which controls for all  possible CTAs.

src/lib/Scenes/Inbox/Screens/Conversation.tsx
xtina-starr: Yeah I think we can dry this up by using the `this.refetch()` here.

src/lib/Scenes/Inbox/Screens/Conversation.tsx
xtina-starr: Right, these are the two instances where the conversation gets updated and we need to know about that in order to update the make offer button.

xtina-starr:
> consider moving the second query renderer specific to the CTA from OpenInquiryModalButton into its enclosing ConversationCTA which controls for all possible CTAs.

These two different queries were separated intentionally. There is more context on that in this [PR](https://github.com/artsy/eigen/pull/4637).

src/lib/Scenes/Inbox/Screens/Conversation.tsx
starsirius: 👍  Let's consolidate so we emit the new event from a single place!

starsirius (approved):
Overall looks good to me! @erikdstock has some suggestions on the naming and consolidation that would be good to include.

And wow the conversation component structure is getting complex. 🤯

xtina-starr:
Closing this PR in favor of [this](https://github.com/artsy/eigen/pull/5165) implementation.

PR #4817, opened 2021-05-20
fix: android deep links

admbtlr:

This PR partly resolves [CX-1454]

### Description

Fixes the Android manifest file to register intents so that links to artsy.net, staging.artsy.net and www.artsy.net all open in the app. But it seems like this only works sometimes, in particular:

- opening links from Chrome never seems to work
- opening links from e.g. mail sometimes works, but sometimes they just open the app without going to the linked page

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -42,6 +42,26 @@
                     android:scheme="https"
                     android:host="staging.artsy.net" />
             </intent-filter>
+            <intent-filter android:autoVerify="true">
+                <action android:name="android.intent.action.VIEW" />
+
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+
+                <data
+                    android:scheme="https"
+                    android:host="www.artsy.net" />
+            </intent-filter>
+            <intent-filter android:autoVerify="true">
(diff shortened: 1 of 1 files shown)


android/app/src/main/AndroidManifest.xml
```diff
@@ -1,53 +1,75 @@
...
+            android:configChanges="keyboard|keyboardHidden|orientation|screenSize|uiMode"
+            android:launchMode="singleTask"
+            android:exported="true"
+            android:windowSoftInputMode="adjustResize">
+            <!-- Handle Deep Links -->
+            <intent-filter>
+                <action android:name="android.intent.action.MAIN" />
+                <category android:name="android.intent.category.LAUNCHER" />
+            </intent-filter>
+            <intent-filter android:autoVerify="true">
```
lordkiz, line 35: Hello Adam,
I think we can use a single intent-filter to for multiple domains instead of creating many intent-filters since they all open into the same activity. Combining them will make the manifest slimmer.

admbtlr: I used separate filters because of this SO comment: https://stackoverflow.com/a/66579456/1788521. Not sure whether's actually necessary, but it seems like there are a lot of small ways in which this won't work, so I wanted to be extra careful.

brainbicycle (approved):
Looks good to me, going to go ahead and merge this so we can test on devices!

PR #4827, opened 2021-05-21
fix(CX-1226): lock screen orientation to portrait on android

MounirDhahri:

This PR resolves [CX-1226]

### Description
**Approach:** Whenever we create a new activity, we need to make to lock its screen orientation
**Issue:** The way we were locking the screen orientation on new activities was wrong, since we were not waiting for previous activities to be closed
**Solution:** Make sure we only lock device orientation once the previous activity is closed (in our case, once the splash screen is hidden)
**Note:** I tested this on my personal phone and it's working properly. I would like to merge this to make sure it's also working for release builds

[CX-1226]: https://artsyproduct.atlassian.net/browse/CX-1226


pvinis (changes requested):
I'm still able to make it go landscape with the same steps as the jira ticket 😢

pvinis:
https://user-images.githubusercontent.com/100233/119338606-1e9b4b00-bc88-11eb-801b-1c00bda479ca.mov

MounirDhahri:
oh thanks, I think I now know what is breaking the orientation. it's navigating using native events. I just pushed a fix 🙌

yarn.lock
```diff
@@ -11841,10 +11841,10 @@ react-native-safe-area-context@^3.1.9:
   resolved "https://registry.yarnpkg.com/react-native-safe-area-context/-/react-native-safe-area-context-3.1.9.tgz#48864ea976b0fa57142a2cc523e1fd3314e7247e"
   integrity sha512-wmcGbdyE/vBSL5IjDPReoJUEqxkZsywZw5gPwsVUV1NBpw5eTIdnL6Y0uNKHE25Z661moxPHQz6kwAkYQyorxA==
 
-react-native-screens@^2.17.1:
-  version "2.17.1"
-  resolved "https://registry.yarnpkg.com/react-native-screens/-/react-native-screens-2.17.1.tgz#c3c0ac750af48741c5b1635511e6af2a27b74309"
-  integrity sha512-B4gD5e4csvlVwlhf+RNqjQZ9mHTwe/iL3rXondgZxnKz4oW0QAmtLnLRKOrYVxoaJaF9Fy7jhjo//24/472APQ==
+react-native-screens@^3.2.0:
```
pvinis, line 11844: I feel kinda 😬 updating this one day before release. I will review this PR and write what I think after review.

HACKS.md
```diff
@@ -117,3 +117,10 @@
   - this also required a patch to fix types
   - See issues here: https://github.com/CocoaPods/CocoaPods/issues/10385, https://github.com/react-native-mapbox-gl/maps/issues/1097
   - we should update to a non-beta version ASAP
+
+- react-native-screens - 3.2.0 postinstall script
+
+  We had issues on the android app where whenever we navigate from a screen to an other screen by dispatching a native event action, the default orientation gets overwritten. This fix makes sure that we are maintaining the default orientation.
+
+  - react-native-screens already created a fix for this that can be found here and should be released in the next build. See https://github.com/software-mansion/react-native-screens/issues/836
+  - Once a new react-native-screens version is released, we can remove our patch and use it instead.
```
pvinis, line 126:
```suggestion
  - Once a new react-native-screens version is released (after 3.2.0), we can remove our patch and use it instead.
```

patches/react-native-screens+3.2.0.patch
```diff
@@ -0,0 +1,6828 @@
...
++id accessibility_custom_action_21
++id accessibility_custom_action_22
++id accessibility_custom_action_23
++id accessibility_custom_action_24
++id accessibility_custom_action_25
++id accessibility_custom_action_26
++id accessibility_custom_action_27
++id accessibility_custom_action_28
++id accessibility_custom_action_29
+
```
pvinis: I think we can remove everything from this line and above.

pvinis, line 13: oooooo nice, nice find. Is that line on the old version of that lib too? If yes, I would much prefer making the patch for that version. If not, then I will test the app now a bit, but I would feel safer with a big big QA since this lib kinda touching things in nav and everything. modals, etc.

pvinis (changes requested):

patches/react-native-screens+3.2.0.patch
```diff
@@ -0,0 +1,16 @@
+diff --git a/node_modules/react-native-screens/android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.java b/node_modules/react-native-screens/android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.java
+index 941c2c6..9b821fb 100644
+--- a/node_modules/react-native-screens/android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.java
++++ b/node_modules/react-native-screens/android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.java
+@@ -46,7 +46,10 @@ public class ScreenStackHeaderConfig extends ViewGroup {
+   private boolean mIsTranslucent;
+   private int mTintColor;
+   private final Toolbar mToolbar;
+-  private int mScreenOrientation = ActivityInfo.SCREEN_ORIENTATION_UNSPECIFIED;
```
pvinis, line 9: oooooo nice, nice find. Is that line on the old version of that lib too? If yes, I would much prefer making the patch for that version. If not, then I will test the app now a bit, but I would feel safer with a big big QA since this lib kinda touching things in nav and everything. modals, etc.

patches/react-native-screens+3.2.0.patch
pvinis: thanks for removing.

brainbicycle (approved):
Thanks both for all the investigation here! I am going to go ahead and merge so we can get in a build and I will test on device later today!

pvinis (approved):
I'll approve this even though I made changes to it so I'm kinda biased 😅. @MounirDhahri check it again when you get the change tomorrow. I also added brian.

MounirDhahri:
Thanks @pvinis this looks good to me

PR #4851, opened 2021-05-24
fix: select filter item on swipe

dimatretyak:

This PR resolves [FX-2885]

### Description
Users can swipe on rows in the filters modal to enter a filter facet or return to the filters overview. When a user does this, it selects the filter facet that was pressed to swipe. The problem is that the filter navigation container is rendered [inside React modal component](https://github.com/artsy/eigen/blob/dimatretyak/fix-select-filter-item-on-swipe/src/lib/Components/ArtworkFilter/ArtworkFilter.tsx#L142) (see attached videos).

https://user-images.githubusercontent.com/3513494/119384422-5aa4cf00-bccd-11eb-91b4-02baffd55bc1.mp4

https://user-images.githubusercontent.com/3513494/119385657-1dd9d780-bccf-11eb-9eeb-bcc0c67770ec.mp4

### There are two possible solutions
1. [Disable the swipe back gesture](https://reactnavigation.org/docs/stack-navigator/#gestureenabled) to go to the previous screen
2. Add extra space on the left and make the swipe back zone equal to the extra space ([This solution](https://github.com/artsy/eigen/blob/master/src/lib/Components/ArtworkFilter/Filters/MultiSelectOption.tsx#L89) is already used in the application)
3. Import `TouchableHighlight` and `TouchableWithoutFeedback` from `react-native-gesture-handler` (thanks @MounirDhahri for this hint) – **on iOS this works perfectly, on Android there are problems with tapping**

#### Fixed demos

https://user-images.githubusercontent.com/3513494/119663680-3c5ce180-be3b-11eb-8670-bdb2301569a7.mp4

https://user-images.githubusercontent.com/3513494/119809034-5f47ce00-beed-11eb-8602-31575639ab59.mp4

[FX-2885]: https://artsyproduct.atlassian.net/browse/FX-2885

### DIFF
--- a/src/setupJest.ts
+++ b/src/setupJest.ts
@@ -425,6 +425,8 @@ jest.mock("react-native/Libraries/LayoutAnimation/LayoutAnimation", () => ({
 
 jest.mock("react-native-gesture-handler", () => {
   const View = require("react-native/Libraries/Components/View/View")
+  const TouchableWithoutFeedback = require("react-native/Libraries/Components/Touchable/TouchableWithoutFeedback")
+  const TouchableHighlight = require("react-native/Libraries/Components/Touchable/TouchableHighlight")
   return {
     Swipeable: View,
     DrawerLayout: View,
@@ -454,6 +456,8 @@ jest.mock("react-native-gesture-handler", () => {
     FlatList: View,
     gestureHandlerRootHOC: jest.fn(),
     Directions: {},
+    TouchableHighlight,
+    TouchableWithoutFeedback,
   }
 })
 
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -32,6 +32,7 @@ upcoming:
     - Fix info button in conversation is hard to press - dzmitry
     - Update order of filters - dzmitry tratsiak
     - Fix time mismatch between auction list and individual auction page - matt
+    - Fix swiping on filter options selects them - dzmitry tratsiak
 
 releases:
   - version: 6.9.2
@@ -57,7 +58,6 @@ releases:
       - Add order history to user profile (behind feature flag) - alexj105
       - Enable lotsByFollowedArtists - ole
       - Add ability to unselected "ways to buy" filter - dzmitry tratsiak
-
 
   - version: 6.9.1
     date: May 18, 2021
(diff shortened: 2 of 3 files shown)


MounirDhahri:
Hey @dimatretyak, I think there might be an other way to fix this as well. Can you try to use `TouchableHighlight` from `react-native-gesture-handler` inside `Touchable.tsx` instead of the react-native one. This should fix the conflict with the button flex wrapper PanResponder and the modal PanResponder

dimatretyak:
> Hey @dimatretyak, I think there might be an other way to fix this as well. Can you try to use `TouchableHighlight` from `react-native-gesture-handler` inside `Touchable.tsx` instead of the react-native one. This should fix the conflict with the button flex wrapper PanResponder and the modal PanResponder

on iOS, this works perfectly, on android, there are problems with selecting the filter element (I can't click on the button)

MounirDhahri:
Since the swipe gesture is not supported on android, let's keep the existing behaviour using `react-native` `TouchableHighlight` on Android and use the one from `react-native-gesture-handler` on iOS. We will be fixing it on android soon, it seems like we are missing some steps from the setup here https://docs.swmansion.com/react-native-gesture-handler/docs/.
It would be probably a good idea not to update the `Touchable` button and instead maybe have a custom button to be used inside the modal

MounirDhahri:
If you want to follow up the progress of the implementation here is it https://artsyproduct.atlassian.net/browse/CX-1519

dimatretyak:
> It would be probably a good idea not to update the `Touchable` button and instead maybe have a custom button to be used inside the modal

In this situation, we will have to create separate `Touchable` and `TouchableRow` components for modals, as well as support them in the future. Maybe it would be better to update only the `Touchable` component?

MounirDhahri:
That's good forward-thinking, it makes sense. Let's go ahead with your suggestion then 👍

MounirDhahri (approved):

src/setupJest.ts
```diff
@@ -425,6 +425,8 @@ jest.mock("react-native/Libraries/LayoutAnimation/LayoutAnimation", () => ({
 
 jest.mock("react-native-gesture-handler", () => {
   const View = require("react-native/Libraries/Components/View/View")
+  const TouchableWithoutFeedback = require("react-native/Libraries/Components/Touchable/TouchableWithoutFeedback")
+  const TouchableHighlight = require("react-native/Libraries/Components/Touchable/TouchableHighlight")
```
MounirDhahri, line 429: 👏

MounirDhahri (approved):

MounirDhahri:
🛳🛳🛳

PR #4852, opened 2021-05-24
enhancement(admin menu): show version, build number, commit short hash

pvinis:

This PR is part of whatever result came after [CX-871].

### Description

Old Admin menu had version and build number and commit hash. I added the same in the new Admin menu.

To add it, I added an ArtsyNativeModule on ios, to keep, same as we have for android. kind of a general place to put things and expose them to TS. For android I added an extra constant to be exported, that reads the git hash.

I skipped the email because it's easy enough to see it in profile. I also renamed some fields to make more sense. I also removed the old swift playground we had on the ios app.

old admin
Screenshot 2021-05-24 at 19 50 07

new admin ios
Screenshot 2021-05-24 at 19 50 01

new admin android
Screenshot 2021-05-24 at 20 14 38

[CX-871]: https://artsyproduct.atlassian.net/browse/CX-871

### DIFF
--- a/Artsy/App/ARTArtsyNativeModule.m
+++ b/Artsy/App/ARTArtsyNativeModule.m
@@ -1,0 +1,15 @@
+#import "ARTArtsyNativeModule.h"
+
+
+@implementation ARTArtsyNativeModule
+
+RCT_EXPORT_MODULE(ArtsyNativeModule);
+
+- (NSDictionary *)constantsToExport
+{
+    return @{
+        @"gitCommitShortHash": [[[NSBundle mainBundle] infoDictionary] objectForKey:@"GITCommitShortHash"],
+    };
+}
+
+@end
--- a/Artsy/App/ARTArtsyNativeModule.m
+++ b/Artsy/App/ARTArtsyNativeModule.m
@@ -5,6 +5,11 @@
 
 RCT_EXPORT_MODULE(ArtsyNativeModule);
 
++ (BOOL)requiresMainQueueSetup
+{
+    return YES;
+}
+
 - (NSDictionary *)constantsToExport
 {
     return @{
--- a/.gitignore
+++ b/.gitignore
@@ -44,8 +44,6 @@ Preview.html
 fastlane/report.xml
 fastlane/Preview.html
 
-Artsy/Eigen.playground/
-
 iOSInjectionProject/
 .vscode/.browse.VC.db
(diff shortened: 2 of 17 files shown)


Artsy/App/ARTArtsyNativeModule.m
```diff
@@ -0,0 +1,20 @@
+#import "ARTArtsyNativeModule.h"
+
+
+@implementation ARTArtsyNativeModule
```
brainbicycle, line 4: I am liking the ART prefix 🤩

Artsy/App/ARTArtsyNativeModule.m
```diff
@@ -0,0 +1,20 @@
...
+{
+    return YES;
+}
+
+- (NSDictionary *)constantsToExport
+{
+    return @{
+        @"gitCommitShortHash": [[[NSBundle mainBundle] infoDictionary] objectForKey:@"GITCommitShortHash"],
+    };
+}
```
brainbicycle, line 18: **thought** We also have the file: https://github.com/artsy/eigen/blob/master/emission/Pod/Classes/TemporaryAPI/ARTemporaryAPIModule.m which seems to serve a similar purpose although I guess some are iOS only, I wonder if we should consolidate?

brainbicycle (approved):
One suggestion on the native module file but very excited to get this back in the admin menu 🙏

Artsy/App/ARTArtsyNativeModule.m
pvinis: I did try to find a place for it at first, before making a new class. everything seemed to have a different purpose, or going away soon. the idea for the ArtsyNativeModule for both platforms is to be there forever to allow us to do this kind of thing, for things that don't necessarily need a whole module for themselves. we had this for android but not iOS.
specifically for the temp one here, I feel like it is best to migrate things from temp to ArtsyNativeModule as necessary once the temp module goes away.

does that make sense?

lordkiz (approved):

Artsy/App/ARTArtsyNativeModule.m
brainbicycle: Yeah I was always a bit on fence for the temp module name because some stuff in there is likely to stick around, I am okay with migrating the permanent stuff over eventually to the new module 👍

brainbicycle:
Merge when ready! Thanks!

PR #4853, opened 2021-05-24
fix: app links on android

MounirDhahri:

This PR resolves [CX-1454]

### Description
- Enable app links on android. This allows us to open the artsy app without asking for which app to handle the deep link.
- The issue with deep links was that we were handling the deep links in the wrong activity (MainActivity instead of the SplashActivity)
- The app links were broken because they were configured in the wrong way as well
- We were missing some deep links so I added them as well

**When the app is closed**

https://user-images.githubusercontent.com/11945712/119411208-4fac6780-bcea-11eb-821e-3e2a4e2a2afb.mov

**When the app is open**

https://user-images.githubusercontent.com/11945712/119411121-32779900-bcea-11eb-964b-629baec763a9.mov

[CX-1454]: https://artsyproduct.atlassian.net/browse/CX-1454

### DIFF
--- a/src/lib/utils/useDeepLinks.ts
+++ b/src/lib/utils/useDeepLinks.ts
@@ -17,7 +17,7 @@ export function useDeepLinks() {
         handleDeepLink(url)
       }
     })
-  }, [])
+  }, [isHydrated])
 
   useEffect(() => {
     Linking.addEventListener("url", ({ url }) => {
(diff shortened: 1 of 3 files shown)


src/lib/utils/useDeepLinks.ts
```diff
@@ -17,7 +17,7 @@ export function useDeepLinks() {
         handleDeepLink(url)
       }
     })
-  }, [])
+  }, [isHydrated])
```
brainbicycle, line 20: Ahh, that was what was missing

brainbicycle (approved):
🙌 🔗

PR #4856, opened 2021-05-25
fix: apply color filter

dimatretyak:

This PR resolves [FX-2396]

### Description
Apply previously selected filters when navigating between Artworks and Exhibitors sections on Fair screen

https://user-images.githubusercontent.com/3513494/119484860-19f59600-bd5f-11eb-87ae-46c0b1824dec.mp4

[FX-2396]: https://artsyproduct.atlassian.net/browse/FX-2396

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -34,8 +34,8 @@ upcoming:
     - Add artwork location filter to artwork grids - iskounen
     - Fix back button in android onboarding is flying off in smaller screens - dzmitry
     - Prevent multiline input placeholder on Android - ole
-
     - Apply selected filters on Fair screen - dzmitry tratsiak
+
 releases:
   - version: 6.9.1
     date: May 18, 2021
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -35,6 +35,7 @@ upcoming:
     - Fix back button in android onboarding is flying off in smaller screens - dzmitry
     - Prevent multiline input placeholder on Android - ole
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.yml
```diff
@@ -27,7 +27,7 @@ upcoming:
     - Add order history to user profile (behind feature flag) - alexj105
     - Enable lotsByFollowedArtists - ole
     - Add ability to unselected "ways to buy" filter - dzmitry tratsiak
-
+    - Apply selected filters on Fair screen - dzmitry tratsiak
```
dblandin, line 30: **minor**: It looks the this extra newline before `releases` might be intentional

dblandin (approved):
This looks good to me but I wonder as a potential follow-up if we might explore a way to set the filters _before_ the initial artwork grid load.

With this PR it looks like we:
- Load the unfiltered grid
- Apply the previously selected filters
- Load the filtered grid

This causes a brief flash between the two grids when you're navigating back to the Artworks tab. Feels a bit jarring.

cc/ @MounirDhahri What do you think?

CHANGELOG.yml
dimatretyak: Fixed

MounirDhahri:
@dblandin I agree, the screen itself has already been a bit slow when switching between tabs, and with the delay to see the already applied filters I think we will have to refactor it soon to be based on the StickTabPage like the Artist page.  (the StickTabPage does not support such a long static header so far and we will have to implement it as well)

PR #4874, opened 2021-05-26
cleanup(metadata): dont upload metadata for android beta, and remove images

pvinis:
The type of this PR is: **cleanup**

This PR resolves [CX-]

### Description

Other teams are free to change our appstore and playstore images, so we should not mess with their changes.

On ios we don't, on android we also don't now.

With that PR I also removed the images we had for android in the repo. We don't need them here, same as with ios, so they are removed.


fastlane/Fastfile
```diff
@@ -172,8 +172,11 @@ lane :ship_beta_android do
 
   upload_sentry_artifacts(sentry_release_name: sentry_release_name, dist_version: "#{vcode}", platform: "android")
 
-  upload_to_play_store(
+  supply(
```
pvinis, line 175: I changed the name to the alias, so that we use the same alias everywhere.

MounirDhahri: 🎉

MounirDhahri (approved):

MounirDhahri:
🛳🛳🛳
#squashongreen

PR #4877, opened 2021-05-27
fix: show error modal on all registration errors

olerichter00:

This PR resolves [CX-1444]

### Description

Right now we show an error result screen for almost every error that happens in the registration flow in the **Register to bid** screen. 

There are two problems:
1. The error messages are not accurate (e.g. when the phone number is invalid the page shows a Network error)
2. When the user closes the error screen the app navigates back to the auction screen and the user has to click on "Register to bid" and enter all the data again.

This PR changes the error handling so that the user sees a simple error modal whenever an error happens.

| Before (generic) | Before (network) | After (generic) | After (phone) |
|---|---|---|---|
|image |image|image|image|

https://user-images.githubusercontent.com/4691889/119827262-3af1ee80-bef9-11eb-88fc-4ae289682cc7.mp4

[CX-1444]: https://artsyproduct.atlassian.net/browse/CX-1444


src/lib/Components/Bidding/Screens/Registration.tsx
```diff
@@ -125,7 +125,7 @@ export class Registration extends React.Component<RegistrationProps, Registratio
       await this.createBidder()
     } catch (error) {
       if (!this.state.errorModalVisible) {
-        this.presentRegistrationError(error, RegistrationStatus.RegistrationStatusError)
+        this.presentErrorModal(error, null)
```
olerichter00, line 128: Note: this presents an error modal with the default message.

src/lib/Components/Bidding/Screens/Registration.tsx
```diff
@@ -135,20 +135,22 @@ export class Registration extends React.Component<RegistrationProps, Registratio
...
-            this.presentErrorModal(errors, null)
+            this.presentErrorModal(errors, errorMessage)
             reject(errors)
           } else {
             done()
           }
         },
         onError: (error) => {
-          this.presentRegistrationError(error, RegistrationStatus.RegistrationStatusNetworkError)
+          this.presentErrorModal(error, errorMessage)
```
MounirDhahri, line 153: **praise:** much better like this

src/lib/Components/Bidding/Screens/Registration.tsx
```diff
@@ -201,7 +203,7 @@ export class Registration extends React.Component<RegistrationProps, Registratio
             }
           }
         },
-        onError: (errors) => this.presentRegistrationError(errors, RegistrationStatus.RegistrationStatusNetworkError),
+        onError: (errors) => this.presentErrorModal(errors, null),
```
MounirDhahri, line 206: too bad we don't get any information onError here

MounirDhahri (approved):

MounirDhahri (approved):

MounirDhahri:
LGTM 🎉

brainbicycle (approved):
Nice! Thanks for doing this, this is much better UX. **thought:** is there any extra phone number validation we should be doing in the form to catch this before the user submits?

MounirDhahri:
@brainbicycle good point. I think we can maybe use our existing phone input here as well

olerichter00:
@MounirDhahri Good idea, I can create a ticket for it. Should be a #good-first-issue.

https://github.com/artsy/eigen/blob/90816d76ea263bfb790a061210a45fb16b1db64e/src/lib/Components/PhoneInput/PhoneInput.tsx#L102

MounirDhahri:
That's the one, thank you for linking it @olerichter00. It can definitely a be a good first issue 🙌

src/lib/Components/Bidding/Screens/Registration.tsx
```diff
@@ -275,12 +273,6 @@ export class Registration extends React.Component<RegistrationProps, Registratio
     }
   }
 
-  // @ts-expect-error STRICTNESS_MIGRATION --- 🚨 Unsafe legacy code 🚨 Please delete this and fix any type errors if you have time 🙏
-  presentRegistrationError(error, status) {
-    console.error("Registration.tsx", error)
-    this.presentRegistrationResult(status)
-  }
-
```
pvinis, line 283: One less thing to care about 👌👌

pvinis:
Nice PR!

(Oops, instead of adding just a comment, I made it as a review. Oh well.
Also, I think you have the wrong video on the PR desc?)

PR #4879, opened 2021-05-27
fix disabled button bug for works with only one edition

lilyfromseattle:

[PURCHASE-2723]

### Description
As per [this discussion]("https://artsy.slack.com/archives/C9YNS4X32/p1622137949282200"), it is clear this is happening for edition works with only 1 edition. For editioned works, we hide the edition selector when `editionSets` has a length of 1,  but we also disable the button unless an edition has been selected. To fix this, I set `selectedEdition` to be the only edition in `editionSets`.

__Before:__
Screen Shot 2021-05-27 at 1 51 21 PM

__After:__
Screen Shot 2021-05-27 at 2 48 54 PM

[PURCHASE-2723]: https://artsyproduct.atlassian.net/browse/PURCHASE-2723

### DIFF
--- a/src/lib/Scenes/Inbox/Components/Conversations/MakeOfferModal.tsx
+++ b/src/lib/Scenes/Inbox/Components/Conversations/MakeOfferModal.tsx
@@ -8,7 +8,7 @@ import { CollapsibleArtworkDetailsFragmentContainer as CollapsibleArtworkDetails
 import renderWithLoadProgress from "lib/utils/renderWithLoadProgress"
 import { ProvideScreenTrackingWithCohesionSchema } from "lib/utils/track"
 import { BorderBox, Button, Flex, Text } from "palette"
-import React, { useState } from "react"
+import React, { useEffect, useState } from "react"
 import { View } from "react-native"
 import { createFragmentContainer, graphql, QueryRenderer } from "react-relay"
 import { InquiryMakeOfferButtonFragmentContainer as InquiryMakeOfferButton } from "./InquiryMakeOfferButton"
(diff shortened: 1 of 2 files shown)


src/lib/Scenes/Inbox/Components/Conversations/MakeOfferModal.tsx
```diff
@@ -30,6 +31,12 @@ export const MakeOfferModal: React.FC<MakeOfferModalProps> = ({ ...props }) => {
     }
   }
 
+  useEffect(() => {
+    if (editionSets?.length === 1) {
+      setSelectedEdition(editionSets[0]?.internalID)
+    }
+  })
```
sepans, line 38:
```suggestion
  }, editionSets)
```
Wondering if it is possible to just run the effect only the `editionSets` change.

sepans (approved):
💯 Great catch!

Not sure if with👇🏽 change it still works as expected but I if it does not a bad idea not to run the effect on every re-render for future-proofing it.

src/lib/Scenes/Inbox/Components/Conversations/MakeOfferModal.tsx
erikdstock: should be `[editionSets]`, no? Or `[]` if we only want it to run once on mount- probably no difference in this case and one is more explicit.

erikdstock (commented):
Looks good from here with one comment on the same area as @sepans - since @lilyfromseattle will be out I can pick this up tomorrow.

src/lib/Scenes/Inbox/Components/Conversations/MakeOfferModal.tsx
jo-rs: Yes! Good catch, @erikdstock. I think, in this case, either solution would work.

sepans:
Did some QA and it looks good.

PR #4901, opened 2021-05-31
feat: Add latest articles rail to home screen

olerichter00:

This PR resolves [CX-1471]

### Description

Adds a "Latest from Artsy Editorial" rail to the home screen.

- Updates Cohesion to 2.1.0
- Adds articles rail to home screen
- Updates design of articles rail in artist overview (e.g. https://www.artsy.net/artist/ana-teresa-fernandez)

#### Home Screen
| iOS | Android | 
|---|---|
| image|image|

#### Artist About Page

| (Before)| iOS | Android | 
|---|---|---|
| image|image|image|

### Todo

- [x] Find out if we want to have the new layout in the about artist screen as well
- [x] Find out how many lines the title has
- [x] Move to correct position on home screen
- [x] Find out if vertical name is needed (not always present on staging)
- [x] Check if image fit is correct
- [x] Track event
- [x] Do not render when list is empty
- [x] Check if we need `inEditorialFeed: true`

[CX-1471]: https://artsyproduct.atlassian.net/browse/CX-1471


src/lib/Components/Artist/Articles/Articles.tsx
```diff
@@ -24,7 +25,7 @@ class Articles extends Component<Props> {
           initialNumToRender={2}
           data={articles}
           keyExtractor={(item) => item.id}
-          renderItem={({ item }) => <Article article={item} />}
+          renderItem={({ item }) => <ArticleCardContainer article={item} />}
```
olerichter00, line 28: Note: We use the new `ArticleCard` component in both rails. I kept both article rail components (home, artist) as two separate components because the title and the used list components differ. Since there is not much else in these components they look identical if we use `ArticleCard` in both.

brainbicycle (approved):
🎉 Nice! Didn't expect to see this in the app so soon! **(thought)** this probably a relatively safe rail because it isn't dependent on user state, for future rails that are we probably want to release initially with a feature flag so we can shut them off if they cause loading problems.

olerichter00:
I was just thinking that it is probably a good idea to release it with a feature flag. We'll introduce the rail and very soon the new articles screen that can be accessed by the rail. What do you think?

PR #4913, opened 2021-06-02
fix: send only changed filter params in the analytics event

dimatretyak:

This PR resolves [FX-2888]

### Description
When either “Ways to buy”, “size” or “price range” filter is selected, selecting another filter resets their value in the analytics event

#### Demo

https://user-images.githubusercontent.com/3513494/120470555-556e1100-c3ac-11eb-9f57-b62a0e798cb2.mp4

[FX-2888]: https://artsyproduct.atlassian.net/browse/FX-2888

### DIFF
--- a/src/lib/Components/ArtworkFilter/ArtworkFilterHelpers.ts
+++ b/src/lib/Components/ArtworkFilter/ArtworkFilterHelpers.ts
@@ -1,5 +1,5 @@
 import { FilterScreen } from "lib/Components/ArtworkFilter"
-import { capitalize, compact, forOwn, groupBy, pick, sortBy } from "lodash"
+import { capitalize, compact, groupBy, isEqual, pick, sortBy } from "lodash"
 import { LOCALIZED_UNIT } from "./Filters/helpers"
 
 export enum FilterDisplayName {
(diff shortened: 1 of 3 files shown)


src/lib/Components/ArtworkFilter/ArtworkFilterHelpers.ts
```diff
@@ -257,27 +257,18 @@ const getDefaultParamsByType = (filterType: FilterType) => {
...
-  forOwn(getChangedParams(selectedFilterOptions), (_value, paramName) => {
-    const filterParamName = paramName as FilterParamName
-    if (currentFilterParams[filterParamName] !== selectedFilterParams[filterParamName]) {
-      changedFilters[filterParamName] = selectedFilterParams[filterParamName]
+  selectedFilterOptions.forEach((selectedFilterOption) => {
+    const { paramName, paramValue } = selectedFilterOption
+    const currentFilterParamValue = currentFilterParams[paramName]
+
+    if (currentFilterParamValue && isEqual(paramValue, currentFilterParamValue)) {
+      return
```
dblandin, line 268: **minor / non-blocking**: Instead of the early return, what do you think of negating the condition and bringing the action into this if block?

```typescript
if (currentFilterParamValue && !isEqual(paramValue, currentFilterParamValue)) {
  changedFilters[paramName] = paramValue
}
```

dblandin (approved):
This LGTM! Left a minor non-blocking suggestion.

src/lib/Components/ArtworkFilter/ArtworkFilterHelpers.ts
dimatretyak: Unfortunately, such a check will not cover all use cases (for example, when a user specifies criteria for a material filter, `currentFilterParams` will not contain this key and, accordingly, `currentFilterParamValue` will return `false`)

dblandin: Ahh, I see. That makes sense 👍🏼

PR #4919, opened 2021-06-02
fix android native go back from sale tab to search

JenqaaaBY:

This PR resolves [CX-1289]

### Description

I fixed an issue where pressing the android native button back would return us to the home tab. Now, if we switched to the sales tab from the search tab when we click the android native back button, we return to the search tab:

https://user-images.githubusercontent.com/30176348/120526543-f75c2080-c3e1-11eb-81be-89dc211f38b3.mov

[CX-1289]: https://artsyproduct.atlassian.net/browse/CX-1289

### DIFF
--- a/src/lib/Components/Artist/ArtistConsignButton.tsx
+++ b/src/lib/Components/Artist/ArtistConsignButton.tsx
@@ -44,7 +44,7 @@ export const ArtistConsignButton: React.FC<ArtistConsignButtonProps> = (props) =
           destination_path: destination,
         })
 
-        navigate(destination, { passProps: { initialGoBackPath: "search" } })
+        isSalesTab ? navigate(destination) : navigate(destination, { passProps: { initialGoBackPath: "search" } })
       }}
     >
       <BorderBox p={0}>
--- a/src/lib/Components/Artist/ArtistConsignButton.tsx
+++ b/src/lib/Components/Artist/ArtistConsignButton.tsx
@@ -44,7 +44,7 @@ export const ArtistConsignButton: React.FC<ArtistConsignButtonProps> = (props) =
           destination_path: destination,
         })
 
-        navigate(destination)
+        navigate(destination, { passProps: { initialGoBackPath: "search" } })
       }}
     >
       <BorderBox p={0}>
--- a/src/lib/Components/Artist/ArtistConsignButton.tsx
+++ b/src/lib/Components/Artist/ArtistConsignButton.tsx
@@ -44,7 +44,9 @@ export const ArtistConsignButton: React.FC<ArtistConsignButtonProps> = (props) =
           destination_path: destination,
         })
 
-        isSalesTab ? navigate(destination) : navigate(destination, { passProps: { initialGoBackPath: "search" } })
+        isSalesTab
+          ? navigate(destination)
+          : navigate(destination, { passProps: { overwriteHardwareBackButtonPath: "search" } })
       }}
     >
       <BorderBox p={0}>
--- a/src/lib/Components/Artist/__tests__/ArtistConsignButton-tests.tsx
+++ b/src/lib/Components/Artist/__tests__/ArtistConsignButton-tests.tsx
@@ -220,7 +220,7 @@ describe("ArtistConsignButton", () => {
       })
       tree.root.findByType(TouchableOpacity).props.onPress()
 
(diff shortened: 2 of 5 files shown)


src/lib/Components/Artist/__tests__/ArtistConsignButton-tests.tsx
```diff
@@ -237,7 +237,9 @@ describe("ArtistConsignButton", () => {
       })
       tree.root.findByType(TouchableOpacity).props.onPress()
 
-      expect(navigate).toHaveBeenCalledWith("/collections/my-collection/marketing-landing")
+      expect(navigate).toHaveBeenCalledWith("/collections/my-collection/marketing-landing", {
+        passProps: { initialGoBackPath: "search" },
+      })
```
MounirDhahri, line 242: **thought:** I am not sure why this needs to be updated? shouldn't only landing to the `/sales` tab scenario be updated

src/lib/Components/Artist/ArtistConsignButton.tsx
```diff
@@ -44,7 +44,7 @@ export const ArtistConsignButton: React.FC<ArtistConsignButtonProps> = (props) =
           destination_path: destination,
         })
 
-        navigate(destination)
+        navigate(destination, { passProps: { initialGoBackPath: "search" } })
```
MounirDhahri, line 47: **issue(blocker):** the naming seems off to me, might be better if we change this to be `overwriteHardwareBackButton` or something else that can easily be understood (it's fine if the string is too long )

src/lib/Scenes/Consignments/index.tsx
```diff
@@ -58,6 +62,41 @@ export interface ConsignmentSetup {
...
+    return state.bottomTabs.sessionState.tabProps.sell ?? {}
+  }) as GlobalStoreData
+
+  const initialGoPackPathString = globalStoreData?.initialGoBackPath ?? null
+
+  const globalStoreDataRef = useRef<BottomTabType | null>(null)
+
+  useEffect(() => {
+    globalStoreDataRef.current = initialGoPackPathString
+  }, [initialGoPackPathString])
```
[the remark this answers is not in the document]
JenqaaaBY: This is necessary in order for the data to be updated, otherwise, the initialGoPackPathString will always take the same value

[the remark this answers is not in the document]
JenqaaaBY: Screen Shot 2021-06-03 at 2 09 06 PM

src/lib/Scenes/Consignments/index.tsx
```diff
@@ -58,6 +62,41 @@ export interface ConsignmentSetup {
...
+  const handleBackButton = useCallback(() => {
+    if (sellTabPropsRef.current) {
+      switchTab(sellTabPropsRef.current)
+      GlobalStore.actions.bottomTabs.setTabProps({ tab: "sell", props: {} })
+    } else {
+      goBack()
+    }
+
+    return true
+  }, [sellTabPropsRef.current])
```
MounirDhahri, line 99: **thought:** do we really need the reference here? would something close to what's down here not work?
```suggestion
  useEffect( () => {
      BackHandler.addEventListener("hardwareBackPress", handleBackButton)
      return () => BackHandler.removeEventListener("hardwareBackPress", handleBackButton)
      }, [])

  const handleBackButton = useCallback(() => {
    if (sellTabPropsRef.current) {
      switchTab(sellTabPropsRef.current)
      GlobalStore.actions.bottomTabs.setTabProps({ tab: "sell", props: {} })
    } else {
      goBack()
    }

    return true
  }, [overwriteHardwareBackButtonPath])
```

JenqaaaBY: Yes, it's needed. It doesn't work otherwise. We talked on the call with @dimatretyak and came to this decision

JenqaaaBY: if you do not use ref, then you have to update the subscription to BackHandler every time

MounirDhahri: makes sense, probably it would make sense then to migrate the handleBackButton inside the useEffect to fix that. This is clear to go then

MounirDhahri:
@JenqaaaBY I forgot to mention something, I think your Github handle is misconfigured, you might need to fix it

PR #4921, opened 2021-06-03
fix: prevent home screen placeholders from rerendering

olerichter00:

This PR resolves [CX-1027]

### Description

Prevent home screen placeholders from rerendering and changing the size before the real content is visible

https://user-images.githubusercontent.com/4691889/120608205-38d7e480-c451-11eb-9577-86d6a53364b3.mp4

[CX-1027]: https://artsyproduct.atlassian.net/browse/CX-1027

### DIFF
--- a/src/lib/Scenes/Home/Home.tsx
+++ b/src/lib/Scenes/Home/Home.tsx
@@ -22,7 +22,7 @@ import { Home_featured } from "__generated__/Home_featured.graphql"
 import { AboveTheFoldFlatList } from "lib/Components/AboveTheFoldFlatList"
 import { GlobalStore, useFeatureFlag } from "lib/store/GlobalStore"
 import { isPad } from "lib/utils/hardware"
-import { PlaceholderBox, PlaceholderText } from "lib/utils/placeholders"
+import { PlaceholderBox, PlaceholderText, RandomWidthPlaceholderText, useMemoizedRandom } from "lib/utils/placeholders"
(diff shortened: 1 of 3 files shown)


MounirDhahri:
Awesome. Thank youuuu 🎉

MounirDhahri:
#squashongreen

src/lib/Scenes/Home/Home.tsx
```diff
@@ -270,15 +267,15 @@ const HomePlaceholder: React.FC<{}> = () => {
           times(2).map((r) => (
             <Box key={r} ml={2} mr={2}>
               <Spacer mb={3} />
-              <PlaceholderText width={100 + Math.random() * 100} />
+              <RandomWidthPlaceholderText minWidth={100} maxWidth={200} />
```
MounirDhahri, line 270: Although using `RandomWidthPlaceholderText` fixes the re-rendering issues in this case, I think the best way to fix the re-rerendering issue would be to identify why is the home screen actually re-rerendering and that seems still unclear to me.

olerichter00: Yep, it would be nice to find out. Didn't someone mentioned a PR that could have introduced the rerendering?

MounirDhahri: Yes, I think it's because of the flash message on the home screen, I worked on it a bit but couldn't get it fixed :(

olerichter00: Thanks for having a look!

PR #4926, opened 2021-06-03
feat: create save search mutation

dimatretyak:

This PR resolves [FX-2971]

### Description
As an admin,
When I toggle the on/off switch on the saved search banner
A new searchCriteria and userSearchCriteria are created in Gravity
With the exact combination of filters I have selected in the UI

#### Acceptance criteria:
* Toggling the switch "on" _executes the mutation_
* For the purposes of this ticket, it's sufficient to check that the `searchCriteria` and `userSearchCriteria` were created using a console
* _Not in scope_: handling failure
* _Not in scope_: preserving banner "on" state on reload/reapplication of same filters
* _Not in scope_: deletion of previously saved searchCritera on toggle off

#### Demo

https://user-images.githubusercontent.com/3513494/120678920-4917b000-c4a1-11eb-8042-5fb55b7ee6b3.mp4

https://user-images.githubusercontent.com/3513494/120678945-50d75480-c4a1-11eb-9afb-42cd2a8044d5.mp4

[FX-2971]: https://artsyproduct.atlassian.net/browse/FX-2971

### DIFF
--- a/src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
+++ b/src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
@@ -142,7 +142,11 @@ const ArtistArtworksContainer: React.FC<ArtworksGridProps & ArtistArtworksContai
       mutation: graphql`
       mutation ArtistArtworksContainerCreateSavedSearchMutation($input: CreateSavedSearchInput!) {
         createSavedSearch(input: $input) {
-          clientMutationId
+          savedSearchOrErrors {
+            ... on SearchCriteria {
+              internalID
+            }
+          }
         }
       }
     `,
--- a/src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
+++ b/src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
@@ -152,7 +152,7 @@ const ArtistArtworksContainer: React.FC<ArtworksGridProps & ArtistArtworksContai
     `,
       variables: {
         input: {
-          artistID: artist.slug,
+          artistID: artist.internalID,
           ...input,
         }
       },
--- a/src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
+++ b/src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
@@ -1,9 +1,11 @@
 import { OwnerType } from "@artsy/cohesion"
 import { ArtistArtworks_artist } from "__generated__/ArtistArtworks_artist.graphql"
+import { ArtistArtworksContainerCreateSavedSearchMutation } from "__generated__/ArtistArtworksContainerCreateSavedSearchMutation.graphql"
 import { ArtworkFilterNavigator, FilterModalMode } from "lib/Components/ArtworkFilter"
 import {
   filterArtworksParams,
   prepareFilterArtworksParamsForInput,
+  prepareFilterParamsForSaveSearchInput,
 } from "lib/Components/ArtworkFilter/ArtworkFilterHelpers"
 import { ArtworkFiltersStoreProvider, ArtworksFiltersStore } from "lib/Components/ArtworkFilter/ArtworkFilterStore"
 import { FilteredArtworkGridZeroState } from "lib/Components/ArtworkGrids/FilteredArtworkGridZeroState"
@@ -19,7 +21,7 @@ import { Schema } from "lib/utils/track"
 import { useScreenDimensions } from "lib/utils/useScreenDimensions"
 import { Box, FilterIcon, Flex, Separator, Spacer, Text, Touchable } from "palette"
 import React, { useContext, useEffect, useState } from "react"
-import { createPaginationContainer, graphql, RelayPaginationProp } from "react-relay"
+import { commitMutation, createPaginationContainer, graphql, RelayPaginationProp } from "react-relay"
 import { useTracking } from "react-tracking"
 import { SavedSearchBanner } from "./SavedSearchBanner"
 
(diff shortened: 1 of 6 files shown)


dblandin (commented):
Looking good! Left a comment about the mutation input type.

Do you think we should convert the input to the new format (nested under attributes / attribution > attributionClasses) within this PR?

src/lib/Components/ArtworkFilter/Filters/helpers.ts
```diff
@@ -98,3 +105,15 @@ export const parseRange = (range: string): Range => {
 
   return { min: enforceNumeric(min), max: enforceNumeric(max) }
 }
+
+export const getFilterArtworkSizeName = (filterArtworkSize: string): string | null => {
```
dblandin, line 109: It feels a bit odd to need this additional mapping when we have something similar in the SizeOptions filter component: https://github.com/artsy/eigen/blob/master/src/lib/Components/ArtworkFilter/Filters/SizeOptions.tsx#L32-L46

I might be missing something in how Eigen actually sends the `size` value up to MP/Gravity when a user is filtering the artwork grid. Is the value actually `small` or is it `*-16.0`?

cc/ @MounirDhahri

src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
```diff
@@ -131,8 +134,49 @@ const ArtistArtworksContainer: React.FC<ArtworksGridProps & ArtistArtworksContai
...
+            ... on SearchCriteria {
+              internalID
+            }
+          }
+        }
+      }
+    `,
+      variables: {
+        input: {
+          artistID: artist.slug,
```
MounirDhahri, line 155: **thought:** why do we need to send the artist slug instead of the id? if there mutation is still not yet public we can probably fix that already on metaphysics

src/lib/Components/ArtworkFilter/Filters/helpers.ts
MounirDhahri: @dblandin the value is actually `*-16.0` 
I also think we better use the ones defined in `SizeOptions`

src/lib/Components/ArtworkFilter/__tests__/FilterArtworksHelpers-tests.tsx
```diff
@@ -825,4 +827,113 @@ describe("prepareFilterArtworksParamsForInput", () => {
...
+        displayText: "Paper",
+        paramName: FilterParamName.materialsTerms,
+        paramValue: ["paper"],
+      }
+    ]);
+
+    expect(prepareFilterParamsForSaveSearchInput(filters)).toEqual({
+      priceMin: 5000,
+      priceMax: 10000,
+      size: "large",
```
MounirDhahri, line 891: **thought:** the paramValue should probably be `40.0-*` here

src/lib/Components/ArtworkFilter/__tests__/FilterArtworksHelpers-tests.tsx
```diff
@@ -825,4 +827,113 @@ describe("prepareFilterArtworksParamsForInput", () => {
       sort: "-decayed_merch",
     })
   })
-})
\ No newline at end of file
+})
+
+describe("getFilterArtworkSizeName", () => {
```
MounirDhahri, line 832: **praise:** nice tests

src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
dblandin: This was my fault - I had the slug in an example request but I actually think this is better as `internalID`.

src/lib/Components/ArtworkFilter/__tests__/FilterArtworksHelpers-tests.tsx
dblandin: This GraphQL mutation isn't backed by a Gravity REST API endpoint and we're still on the fence about whether we also want to support this query param-friendly range syntax in the Gravity GraphQL mutation resolvers. Instead of `40.0-*` we could also support `minSize` and `maxSize` arguments:

```graphql
createSavedSearch(attributes: { minSize: 40 })
```

src/lib/Components/ArtworkFilter/Filters/helpers.ts
dblandin: Gotcha. Thank you, that's helpful. Very tricky to track these values.

@anandaroop has even started [a Google Sheet](https://docs.google.com/spreadsheets/d/1yyVcBLFLn9394ksg3LH5-4DZRUhceVq20-0qPEw_cbo/edit?usp=sharing) to get organized :)

MounirDhahri (approved):

dimatretyak:
> Do you think we should convert the input to the new format (nested under attributes / attribution > attributionClasses) within this PR?

Converted input to a new format

src/lib/Components/ArtworkFilter/Filters/helpers.ts
```diff
@@ -3,6 +3,13 @@ import { isNull, isUndefined, round as __round__ } from "lodash"
 
 type Numeric = "*" | number
 type Unit = "in" | "cm"
+export enum FilterArtworkSize {
```
dblandin, line 6: I think we should remove this new enum as well until we figure out how to best handle the size input within the new mutation.

dblandin (approved):
Nice! So exciting!! 💯 

totoro

dblandin:
#squashongreen

dimatretyak:
Thank you 😀✌️

PR #4932, opened 2021-06-04
Feat: (FX-2910) - filter art in collection screen

lordkiz:

This PR resolves [FX-2910]

### Description
Tested on devices with screen notch and those without.

#### ANDROID

https://user-images.githubusercontent.com/18648835/120812858-ac5f1c00-c54d-11eb-9604-ace3427c34a6.mp4

https://user-images.githubusercontent.com/18648835/120810124-0f02e880-c54b-11eb-8ba3-af8b76033c52.mp4

#### IOS

https://user-images.githubusercontent.com/18648835/120815715-6192d380-c550-11eb-93c4-e889fdf2f2fb.mp4

https://user-images.githubusercontent.com/18648835/120817975-78d2c080-c552-11eb-85f1-8e1e36f1371a.mp4

[FX-2910]: https://artsyproduct.atlassian.net/browse/FX-2910


src/lib/Scenes/Collection/Components/CollectionArtworksFilter.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+export const CollectionArtworksFilter: React.FC<FilterProps> = ({ collection, animationValue }) => {
+  const tracking = useTracking()
+
+  const filteredTotal = ArtworksFiltersStore.useStoreState((state) => state.counts.total) || 0
+  const filtersPresent = ArtworksFiltersStore.useStoreState((state) => state.appliedFilters.length > 0)
+
+  const screenWidth = useScreenDimensions().width
+
+  const artworksTotal = filtersPresent
```
dimatretyak, line 39: We can read `artworksTotal` from `ArtworksFiltersStore`

```javascript
const artworksTotal = ArtworksFiltersStore.useStoreState((state) => state.counts.total) || 0
```

src/lib/Scenes/Collection/Components/CollectionArtworksFilter.tsx
```diff
@@ -0,0 +1,194 @@
...
+            mode={FilterModalMode.Collection}
+          />
+        </Box>
+      </Animated.View>
+    </Box>
+  ) : null
+}
+
+export const CollectionArtworksFilterFragmentContainer = createFragmentContainer(CollectionArtworksFilter, {
+  collection: graphql`
```
dimatretyak, line 182: I think the query should look something like this.
1. We will only get `id` and `slug` for analytics.
2. We will get artwork counts from `ArtworksFiltersStore`

```
export const CollectionArtworksFilterFragmentContainer = createFragmentContainer(CollectionArtworksFilter, {
  collection: graphql`
    fragment CollectionArtworksFilter_collection on MarketingCollection
    @argumentDefinitions(input: { type: "FilterArtworksInput" }) {
      slug
      id
    }
  `,
})
```

src/lib/Scenes/Collection/Screens/CollectionArtworks.tsx
```diff
@@ -55,6 +57,24 @@ export const CollectionArtworks: React.FC<CollectionArtworksProps> = ({ collecti
...
 
+  useEffect(() => {
+    // for use by CollectionArtworksFilter to keep count
+    setFiltersCountAction({
+      total: artworksTotal ?? null,
+      followedArtists,
+    })
+  }, [artworksTotal])
+
+  useEffect(() => {
```
dimatretyak, line 68: We can delete this `useEffect`, since we will store the artwork counts for the `CollectionArtworksFilter` component inside the `ArtworksFiltersStore`

src/lib/Scenes/Collection/Screens/CollectionArtworks.tsx
```diff
@@ -55,6 +57,24 @@ export const CollectionArtworks: React.FC<CollectionArtworksProps> = ({ collecti
     setAggregationsAction(collection.collectionArtworks!.aggregations)
   }, [])
 
+  useEffect(() => {
+    // for use by CollectionArtworksFilter to keep count
+    setFiltersCountAction({
+      total: artworksTotal ?? null,
+      followedArtists,
+    })
+  }, [artworksTotal])
```
dimatretyak, line 66: We should pass `followedArtists` in dependencies

```javascript
useEffect(() => {
    // for use by CollectionArtworksFilter to keep count
    setFiltersCountAction({
      total: artworksTotal ?? null,
      followedArtists,
    })
  }, [artworksTotal, followedArtists])
```

dimatretyak:
@MounirDhahri for the current problem, will such an algorithm be considered normal?
1. Save the artworks count in the `CollectionArtworks` component using `setFiltersCountAction` (We also update the artwork counts every time after applying the filter)
2. Read the artworks count from the `ArtworksFiltersStore` inside the `Collection` component

src/lib/Scenes/Collection/Components/CollectionArtworksFilter.tsx
lordkiz: Well, I will need the initial counts total before filters are applied. Before filters are applied the count in ArtworksFiltersStore is null

src/lib/Scenes/Collection/Screens/CollectionArtworks.tsx
lordkiz: You are correct. Actually, I think this entire use of ArtworksFiltersStore can be avoided in this scenario. I am finding it difficult syncing the the updated count when connections are refetched in the CollectionArtworks component and the count in the filter component.

src/lib/Scenes/Collection/Components/CollectionArtworksFilter.tsx
dimatretyak: It will be installed as soon as the list of artworks is loaded

lordkiz: Oh yeah. True that! How did I miss this?? 🤦

src/lib/Scenes/Collection/Screens/CollectionArtworks.tsx
lordkiz: I see what you mean, Dima but that wouldn't be ideal since it is a store that other components may be reading from too. Have to watch out for stale data being used elsewhere. I think this updating followedArtist and total is just messy especially as there is no business btw the count and followedArtist in the context of this ticket. I think I will either separate this whole count sync thing to its own separate store or I will add a new state/action for it on ArtworksFiltersStore. Thanks for spotting these things. 👍

src/lib/Scenes/Collection/Screens/CollectionArtworks.tsx
lordkiz: Since followedArtist isn't in this context, final fix doesn't update it alongside artworkstotal

dblandin (approved):
Overall, this looks great! Nice work 💯 

Looks like we have a very minor fix to the changelog pending but other than that, I think we're about ready to merge this in 👍🏼

PR #4937, opened 2021-06-07
feat: check savedSearch query field

dimatretyak:

This PR resolves [FX-2967]

### Description
As an admin with the saved search feature flag enabled
_When I_ apply filters on the artist artwork screen
_A query_ is executed that checks if I have a saved search with those filters

##### Acceptance criteria:
* Every time a user with the saved search flag applies filters, a query is executed that checks if that user has an existing saved search with those filters
* End result of this ticket could just be logging out the result of that query

#### Demo

https://user-images.githubusercontent.com/3513494/121049697-9e650180-c7c0-11eb-9c11-e93e46673b2e.mp4

https://user-images.githubusercontent.com/3513494/121050503-4c70ab80-c7c1-11eb-8feb-832316e97550.mp4

[FX-2967]: https://artsyproduct.atlassian.net/browse/FX-2967

### DIFF
--- a/src/lib/Components/Artist/ArtistArtworks/SavedSearchBanner.tsx
+++ b/src/lib/Components/Artist/ArtistArtworks/SavedSearchBanner.tsx
@@ -87,7 +87,7 @@ export const SavedSearchBanner: React.FC<SavedSearchBannerProps> = ({ me, attrib
   )
 }
 
-export const SavedSearchBannerContainer = createFragmentContainer(
+export const SavedSearchBannerFragmentContainer = createFragmentContainer(
   SavedSearchBanner,
   {
     me: graphql`
@@ -119,7 +119,7 @@ export const SavedSearchBannerQueryRender: React.FC<{ filters: FilterParams, art
           }
         }
       `}
-      render={({ props }) => <SavedSearchBannerContainer {...props} loading={props === null} attributes={attributes} artistId={artistId} />}
+      render={({ props }) => <SavedSearchBannerFragmentContainer {...props} loading={props === null} attributes={attributes} artistId={artistId} />}
       variables={{
         criteria: attributes
       }}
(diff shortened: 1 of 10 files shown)


src/lib/Components/Artist/ArtistArtworks/SavedSearchBanner.tsx
```diff
@@ -34,3 +115,48 @@ export const SavedSearchBanner: React.FC<SavedSearchBannerProps> = ({ enabled, l
...
+    <QueryRenderer<SavedSearchBannerQuery>
+      environment={defaultEnvironment}
+      query={graphql`
+        query SavedSearchBannerQuery($criteria: SearchCriteriaAttributes!) {
+          me {
+            ...SavedSearchBanner_me @arguments(criteria: $criteria)
+          }
+        }
+      `}
+      render={({ props, error }) => (
```
dimatretyak, line 149: Should the banner be displayed if an error occurred during the request?

dblandin: Sometimes you can have an error but still have the props you need to render.

There's this pattern which will either console.log the error or send it to Sentry, which seems reasonable:

https://github.com/artsy/eigen/blob/master/src/lib/Scenes/Search/AutosuggestResults.tsx#L206-L211

dblandin:
> > We should eventually add specs that exercise both GraphQL mutations
> 
> @dblandin If it is not difficult, can you describe in more detail what is meant by this?

Sure! I'm thinking of something like this which sets up some mock mutation data, does something in the UI (like tap the enable/disable) button, and asserts on some kind of response like a toast appearing or the banner switching to the other state.

There should be some examples in Eigen but here's one example in Force that I'm familiar with: https://github.com/artsy/force/blob/8aa26fc8215cbb679ba7d14f81d25d23fc6337e9/src/v2/Components/UserSettings/TwoFactorAuthentication/__tests__/TwoFactorAuthentication.jest.tsx#L124

dimatretyak:
@dblandin

* [createSavedSearch mutation spec](https://artsyproduct.atlassian.net/browse/FX-3006)
* [deleteSavedSearch mutation spec](https://artsyproduct.atlassian.net/browse/FX-3006)

PR #4939, opened 2021-06-07
fix(icons): small fix in decrease and increase icons

pvinis:

This PR resolves []

### Description

tiny fix but you can see the change if you zoom in.

before:
Screenshot 2021-06-07 at 17 00 35

after:
Screenshot 2021-06-07 at 17 00 15

If you cant see it, the difference is on the down arrow. before, it was not pointy like the up arrow. it was a bit smushed, because the rotation was at the wrong point.

I re-exported from figma and now its nice and pointy like the up arrow.

### DIFF
--- a/src/lib/AndroidApp.tsx
+++ b/src/lib/AndroidApp.tsx
@@ -80,7 +80,7 @@ const Main: React.FC<{}> = track()(({}) => {
   }, [isHydrated])
 
   if (!isHydrated) {
-    return <View></View>
+    return <View />
   }
 
   if (forceUpdateMessage) {
--- a/src/palette/svgs/DecreaseIcon.tsx
+++ b/src/palette/svgs/DecreaseIcon.tsx
@@ -1,13 +1,13 @@
-import React from "react"
+import React, { FC } from "react"
 import { color } from "../Theme"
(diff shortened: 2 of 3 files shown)


edez1t:
nice, but how did u notice that? xd

pvinis:
hahaha well, thats an interesting story.

- i had to add these two icons from eigen to palette to be used in force.
- I noticed that decrease is basically increase but flipped, so I tried to do something like
```
<Icon>
  <G rotate(..)>
    <IncreaseIcon />
```
but it didn't work well.
- I still thought that's the right way, so I put the svg on an svg editor, played around, and I saw that the size was not correct for the rotation point
- I tried to make the rotation right manually, while learning how svg rotation works, and got a bit lost haha
- I went to figma to export, and I saw the size was changed, so I tested the stuff above again and it worked
- I noticed it was missing the point while I was in palette storybook. I thought I was cutting of something with padding or whatever, but it was just the rotation I saw before.
- I decided to keep them separate but re-export

🤷 😅 

I still think we could have decrease be a rotation of increase and not just the increase `path`, but I'll keep it simple for now :D

MounirDhahri:
nice catch!

PR #4945, opened 2021-06-08
feat: don’t close modal when “clear all” tapped

dimatretyak:

This PR resolves [FX-3000]

### Description
User Story:As a collector, when I click "clear all" when filtering I expect the filter options to clear without closing the modal so I can immediately choose a new set of them. 

#### Demo
https://user-images.githubusercontent.com/3513494/121177984-b1321180-c866-11eb-9b98-cb830b943694.mp4

[FX-3000]: https://artsyproduct.atlassian.net/browse/FX-3000

### DIFF
--- a/src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
+++ b/src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
@@ -32,9 +32,8 @@ const ArtworksGrid: React.FC<ArtworksGridProps> = ({ artist, relay, ...props }) 
   const tracking = useTracking()
   const [isFilterArtworksModalVisible, setFilterArtworkModalVisible] = useState(false)
 
-  const handleFilterArtworksModal = () => {
-    setFilterArtworkModalVisible(!isFilterArtworksModalVisible)
-  }
+  const handleCloseFilterArtworksModal = () => setFilterArtworkModalVisible(false)
+  const handleOpenFilterArtworksModal = () => setFilterArtworkModalVisible(true)
 
   const openFilterArtworksModal = () => {
     tracking.trackEvent({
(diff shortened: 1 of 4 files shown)


src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
```diff
@@ -32,9 +32,8 @@ const ArtworksGrid: React.FC<ArtworksGridProps> = ({ artist, relay, ...props })
   const tracking = useTracking()
   const [isFilterArtworksModalVisible, setFilterArtworkModalVisible] = useState(false)
 
-  const handleFilterArtworksModal = () => {
-    setFilterArtworkModalVisible(!isFilterArtworksModalVisible)
-  }
+  const handleCloseFilterArtworksModal = () => setFilterArtworkModalVisible(false)
+  const handleOpenFilterArtworksModal = () => setFilterArtworkModalVisible(true)
```
mdole, line 36: I suggest that we get rid of `handleCloseFilterArtworksModal` and `handleOpenFilterArtworksModal` because they do not seem to make the code easier to understand compared to `setFilterArtworkModalVisible()`

mdole (commented):
One small suggestion, otherwise looks good!

mdole:
Assigning back to you to merge when ready @dimatretyak

src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
dimatretyak: Should I return how it was [implemented before](https://github.com/artsy/eigen/blob/master/src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx#L35-L37)? Or use `setFilterArtworkModalVisible` without helper wrappers??

mdole: Ah, I see now that this way gives us a callback so that we don't have to write `() => ...` repeatedly. Let's keep it as-is! I will resolve the changelog conflict and merge.

dimatretyak: Thanks!

PR #4951, opened 2021-06-09
feat: artist series sticky filter

dimatretyak:

This PR resolves []

### Description
[Initial implementation](https://github.com/artsy/eigen/pull/4705)

#### Demo
https://user-images.githubusercontent.com/3513494/121366762-684f8b00-c942-11eb-9ade-d47ebf5ba1c0.mp4

### DIFF
--- a/src/lib/Components/StickyHeaderPage.tsx
+++ b/src/lib/Components/StickyHeaderPage.tsx
@@ -2,7 +2,7 @@ import { useAnimatedValue } from "lib/Components/StickyTabPage/reanimatedHelpers
 import { useUpdadeShouldHideBackButton } from "lib/utils/hideBackButtonOnScroll"
 import { useAutoCollapsingMeasuredView } from "lib/utils/useAutoCollapsingMeasuredView"
 import { useScreenDimensions } from "lib/utils/useScreenDimensions"
-import React, { useState } from "react"
+import React, { useRef } from "react"
 import { View } from "react-native"
 import Animated from "react-native-reanimated"
 
@@ -15,7 +15,7 @@ interface StickyHeaderPageProps {
 export const StickyHeaderPage: React.FC<StickyHeaderPageProps> = (props) => {
   const { headerContent, stickyHeaderContent, footerContent, children } = props
 
-  const [footerOffsetY, setFooterOffsetY] = useState<Animated.Value<number>>(new Animated.Value(-1))
+  const footerOffsetY = useRef<Animated.Value<number>>(new Animated.Value(-1))
   const { jsx: staticHeader, nativeHeight: headerHeight } = useAutoCollapsingMeasuredView(headerContent)
   const { jsx: stickyHeader, nativeHeight: stickyHeaderHeight } = useAutoCollapsingMeasuredView(stickyHeaderContent)
   const scrollOffsetY = useAnimatedValue(0)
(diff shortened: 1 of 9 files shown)


src/lib/Scenes/ArtistSeries/ArtistSeries.tsx
```diff
@@ -84,66 +70,67 @@ export const ArtistSeries: React.FC<ArtistSeriesProps> = (props) => {
...
+                    currentArtistSeriesExcluded
+                  />
+                </Box>
+              </>
+            ) : undefined}
+            stickyHeaderContent={(
+              <Box backgroundColor="white100" py={1}>
+                <Flex flexDirection="row" px={2} justifyContent="space-between" alignItems="center">
+                  <Text variant="subtitle" color="black60">
+                    Showing 100 works
```
mdole, line 103: We should update this hardcoded value before merge :)

src/lib/Components/StickyHeaderPage.tsx
```diff
@@ -0,0 +1,90 @@
...
+    Animated.cond(
+      Animated.greaterThan(scrollStickyHeaderOffsetY, footerOffsetY),
+      Animated.max(
+        Animated.sub(footerOffsetY, scrollStickyHeaderOffsetY),
+        Animated.multiply(-1, stickyHeaderHeight ?? 0)
+      ),
+      stickyHeaderOffsetY
+    ),
+    stickyHeaderOffsetY
+  )
```
mdole, line 39: @dimatretyak can you help me understand what this does? I'm guessing it's to allow the sticky header to scroll out of view if we have a large-enough footer, but I'm getting a bit lost trying to read it!

mdole (commented):
@dimatretyak I think this is a really good approach! I like the way you created a pattern that could be used elsewhere in the app where we want a sticky header without tabs. I'll tag Pavlos for review as well because he's a more experienced mobile dev than I am, but what you have here is looking good to me.

mdole:
@dimatretyak you mentioned via Slack that you were seeing some small lags when scrolling on Android, which you were also seeing with StickyTabPage. I don't see any lags when scrolling in the app on my device - can you perhaps upload a video of the issue? Does it happen on a physical device, on a simulator, or both?

src/lib/Components/StickyHeaderPage.tsx
dimatretyak: @mdole
line 29 – if the user passed the footer component, then we check whether the coordinate of the beginning of this component was read
Line 31 - if the user has started scrolling the footer area, then run a smooth translateY reduction for the sticky header (line 33) until the sticky header is completely hidden (lines 33-34)

dimatretyak: If I have time, I will try to refactor the component a little

src/lib/Scenes/ArtistSeries/ArtistSeries.tsx
dimatretyak: No problems. I've left hardcoded for the demo

dimatretyak:
> @dimatretyak you mentioned via Slack that you were seeing some small lags when scrolling on Android, which you were also seeing with StickyTabPage. I don't see any lags when scrolling in the app on my device - can you perhaps upload a video of the issue? Does it happen on a physical device, on a simulator, or both?

@mdole  I tested it on a physical device. If you look closely, you may notice a slight twitching of the header while scrolling. Perhaps this problem will not be in the release build

https://user-images.githubusercontent.com/3513494/121727154-33833580-caf4-11eb-909d-fb0af0b2841e.mp4

src/lib/Components/StickyHeaderPage.tsx
dimatretyak: Made a small refactoring of the `StickyHeaderPage`

mdole: Thanks for the explanation and refactor, that helps!

mdole (approved):
@dimatretyak nice work, and thanks for the video demo! The code is looking good to me. I'm not sure what optimizations we do differently on Production builds and whether they would fix the scrolling animation. I'll start a thread in #dev-help and see if anyone else knows!

dimatretyak:
@mdole @MounirDhahri unfortunately, the problem still remains

https://user-images.githubusercontent.com/3513494/123660257-2bd5c780-d83c-11eb-8157-d16fb2eca187.mp4

mdole:
Yeah, that sounds good to me as well. It's not all that noticeable unless you're really looking for it

PR #4958, opened 2021-06-11
feat: show toast for saved search

dimatretyak:

This PR resolves [FX-3009]

### Description
As an admin,
When I toggle the on/off switch on the saved search banner
A new searchCriteria and userSearchCriteria are created in Gravity
Show toast message about the result of mutation execution

#### Acceptance criteria:
* `createSavedSearch` mutation shows the following
    * Title: `Your alert has been set.`
    * Description: `We will send you a push notification once new works are added.`
* `deleteSavedSearch` mutation shows the following
    * Title: `Your alert has been removed.`
    * Description: `Don’t worry, you can always create a new one.`

#### Demo

https://user-images.githubusercontent.com/3513494/121697482-c9f42e80-cad5-11eb-88c9-c08a37f7dae5.mp4

https://user-images.githubusercontent.com/3513494/121697469-c5c81100-cad5-11eb-8f50-64b122678f49.mp4

[FX-3009]: https://artsyproduct.atlassian.net/browse/FX-3009


dblandin (approved):
This looks fantastic 💯

pvinis:
not for this PR to fix, but we have 3 different toast styles, each with different plumbing to show and trigger. we should at least make that 1 way, and we should ask designers for the styles.

src/lib/AppProviders.tsx
```diff
@@ -3,6 +3,7 @@ import { Theme } from "palette"
 import React, { ReactNode } from "react"
 import { RelayEnvironmentProvider } from "relay-hooks"
 import { _FancyModalPageWrapper } from "./Components/FancyModal/FancyModalContext"
+import { NotificationProvider } from "./Components/Notification/NotificationProvider"
 import { ToastProvider } from "./Components/Toast/toastHook"
```
brainbicycle, line 7: I see we have `ToastProvider` which looks like it serves a similar function, is it possible to reuse this or are the 2 significantly different? 
https://github.com/artsy/eigen/blob/master/src/lib/Components/Toast/Toast.tsx

src/lib/AppProviders.tsx
```diff
@@ -15,7 +16,11 @@ export const AppProviders = ({ children }: { children: ReactNode }) => (
         <ActionSheetProvider>
           <GlobalStoreProvider>
             <ToastProvider>
-              <_FancyModalPageWrapper>{children}</_FancyModalPageWrapper>
+              <NotificationProvider>
```
brainbicycle, line 19: **(blocking)** Can we name this something else, like PopoverMessageProvider or something similar? Notification in the mobile context means push notifications for me and I suspect will be kind of confusing for others.

src/lib/Components/Notification/Notification.tsx
```diff
@@ -0,0 +1,175 @@
...
+        useNativeDriver: true,
+        friction: FRICTION,
+      }),
+      Animated.timing(opacityAnim, {
+        toValue: 1,
+        useNativeDriver: true,
+        duration: 450,
+      }),
+    ]).start()
+  }, [])
```
brainbicycle, line 78: Nice I need to do some reading on Animated, this is cool

src/lib/AppProviders.tsx
dimatretyak: Renamed

src/lib/AppProviders.tsx
dimatretyak: Renamed

src/lib/AppProviders.tsx
dimatretyak: Unfortunately, too different designs, appearance animation, custom appearance of the Toast component for the middle placement. We can combine these two components (Toast and PopoverMessage), but then there will be too many options to pass different component props to customize the appearance. And I'm afraid that this may confuse the developer

brainbicycle: I think rather than allowing devs to customize all the different props we could have a mode option and just have 'centerToast' | 'popoverMessage' for example, but out of scope of this PR.

brainbicycle:
Thanks for the rename, looks good! I agree with Pavlos that we should consolidate these Toasts but don't think it should block this PR.

PR #4962, opened 2021-06-14
fix: Normalize market news rail spacing

olerichter00:

This PR resolves [CX-1600]

### Description

simulator_screenshot_E68130E3-ED62-4911-9892-CA93D6820260

[CX-1600]: https://artsyproduct.atlassian.net/browse/CX-1600


src/lib/Scenes/Home/Components/ArticlesRail.tsx
```diff
@@ -41,6 +41,7 @@ export const ArticlesRail: React.FC<ArticlesRailProps> = ({ articlesConnection }
           showsHorizontalScrollIndicator={false}
           ListHeaderComponent={() => <Spacer ml="2" />}
           ListFooterComponent={() => <Spacer ml="2" />}
+          ItemSeparatorComponent={() => <Spacer ml={15} />}
```
pvinis, line 44: I think here it would be better to use the design system values, so instead of `{15}`, if you write `"1.5"`, it should give you 15 pixels.

src/lib/Components/Artist/Articles/Articles.tsx
```diff
@@ -20,6 +21,7 @@ class Articles extends Component<Props> {
         <AboveTheFoldFlatList<Articles_articles[number]>
           horizontal
           showsHorizontalScrollIndicator={false}
+          ItemSeparatorComponent={() => <Spacer ml={20} />}
```
pvinis, line 24: Same here, with `"2"`.

src/lib/Components/ArticleCard.tsx
```diff
@@ -21,16 +24,16 @@ export const ArticleCard: React.FC<ArticleCardProps> = ({ article, onPress, isFl
   }
 
   return (
-    <Flex width={isFluid ? "100%" : 320}>
+    <Flex width={isFluid ? "100%" : WIDTH}>
```
pvinis, line 27: YES! random values without a name are always confusing. Thanks for fixing that.

pvinis (approved):
I'll approve, because my comments are not really blocking, but I'd say take a look and try it. That's one other thing to add to the preferred practices ;) I'll add right now.

src/lib/Scenes/Home/Components/ArticlesRail.tsx
olerichter00: Nice, that's good to know :)

olerichter00:
#squashongreen

PR #4969, opened 2021-06-15
feat: Connect auction results for you to metaphysics endpoint

JenqaaaBY:

This PR resolves [CX-1558]

### Description

Implemented graphQL queries for auctionResultForYou components and wrote tests

How it looks on android:
https://user-images.githubusercontent.com/30176348/122578159-48177e80-d05c-11eb-94dd-d0a906d55b29.mov

How it looks on iOS:

https://user-images.githubusercontent.com/30176348/122578569-aba1ac00-d05c-11eb-941c-1ce30f0668a4.mov

[CX-1558]: https://artsyproduct.atlassian.net/browse/CX-1558


src/lib/Scenes/AuctionResultForYou/AuctionResultForYou.tsx
```diff
@@ -1,239 +1,84 @@
...
+      }
+      setLoadingMoreData(false)
+    })
+  }
 
-          return (
+  return (
+    <PageWithSimpleHeader title="Auction Results for You">
+      <ArtworkFiltersStoreProvider>
+        <Sans size="3" textAlign="left" color="black60" style={{ marginHorizontal: 20, marginVertical: 17 }}>
```
olerichter00, line 52: Afaik we try to use the palette component `Text` everywhere instead of `Sans` or `Serif`. @MounirDhahri can you confirm?

src/lib/Scenes/AuctionResultForYou/AuctionResultForYouContainer.tsx
```diff
@@ -0,0 +1,87 @@
...
+      query AuctionResultForYouContainerQuery {
+        me {
+          ...AuctionResultForYouContainer_me
+        }
+      }
+    `}
+    variables={{}}
+    cacheConfig={{
+      force: true,
+    }}
```
olerichter00, line 84: (question) I'm wondering if it is necessary to set `force: true` for this query. Is there any specific reason? (https://relay.dev/docs/api-reference/types/CacheConfig/)

src/lib/Scenes/AuctionResultForYou/AuctionResultForYouListItem.tsx
```diff
@@ -8,19 +10,25 @@ import { bullet, color, Flex, NoArtworkIcon, Text, Touchable } from "palette"
 import React from "react"
 
 interface Props {
-  auctionResult: any // add correct type after implementing actionResult query
-  onPress: () => void
+  auctionResult: NonNullable<
+    NonNullable<
+      NonNullable<NonNullable<AuctionResultForYouContainer_me["auctionResultsByFollowedArtists"]>["edges"]>[0]
+    >["node"]
+  >
```
olerichter00, line 17: I guess you can simply create another fragment container for `AuctionResultForYouListItem` and specify the data requirements for auction result items in there. Then you can use the type from the new fragment, which should be `AuctionResultForYouContainer_auctionResult`

You can follow the pattern from here:
https://github.com/artsy/eigen/blob/8c33f569d02b7e71388b6f08f4fc312a80cc060e/src/lib/Components/ArtistListItem.tsx#L199

src/lib/Scenes/AuctionResultForYou/AuctionResultForYouListItem.tsx
```diff
@@ -82,7 +93,7 @@ export const AuctionResultForYouListItem: React.FC<Props> = ({ auctionResult, on
               </Flex>
             ) : (
               <Flex alignItems="flex-end">
-                <Text variant="caption" fontWeight="bold" style={{ width: 100 }} textAlign="right" testID="price">
+                <Text variant="text" fontWeight="bold" style={{ width: 100 }} textAlign="right" testID="price">
```
olerichter00, line 96: (non-blocking) I'm wondering if we really need `style={{ width: 100 }}` here. I guess it is ok to override styles but we should not do it if we don't have to.

src/lib/Scenes/Home/Components/AuctionResultsRail.tsx
```diff
@@ -153,37 +37,53 @@ const AuctionResultsRail: React.FC<RailScrollProps> = (props) => {
 
       <CardRailFlatList
         listRef={listRef}
-        data={mockAuctionResults}
+        data={auctionResultsByFollowedArtists?.edges}
```
olerichter00, line 40: We can use `extractNodes` here as well to extract the data.

https://github.com/artsy/eigen/blob/8c33f569d02b7e71388b6f08f4fc312a80cc060e/src/lib/utils/extractNodes.ts#L1

src/lib/Scenes/Home/Components/AuctionResultsRail.tsx
```diff
@@ -153,37 +37,53 @@ const AuctionResultsRail: React.FC<RailScrollProps> = (props) => {
...
         keyExtractor={(_, index) => String(index)}
         horizontal={false}
         initialNumToRender={3}
-        renderItem={({ item: result, index }) => {
-          if (index >= 3) {
+        renderItem={({ item, index }) => {
+          if (index >= 3 || !item?.node) {
```
olerichter00, line 45: (question) Why do we need `index >= 3` here? Shouldn't the query specify how many items we are retrieving?

olerichter00 (commented):
That looks good so far. 

I just added some comments and questions and I'm wondering if we can improve the file structure to make it more consistent with other modules. The file `AuctionResultForYouContainer.tsx` exports the container and the query renderer but the first one isn't used anywhere else. Usually, the container and the query render live inside of the same file (in your case `AuctionResultForYou.tsx`)

You can have a look at `ArtistSeries` for an example: 
https://github.com/artsy/eigen/blob/8c33f569d02b7e71388b6f08f4fc312a80cc060e/src/lib/Scenes/ArtistSeries/ArtistSeries.tsx#L29

src/lib/Scenes/AuctionResultForYou/AuctionResultForYouListItem.tsx
```diff
@@ -8,19 +10,25 @@ import { bullet, color, Flex, NoArtworkIcon, Text, Touchable } from "palette"
...
-  onPress: () => void
+  auctionResult: NonNullable<
+    NonNullable<
+      NonNullable<NonNullable<AuctionResultForYouContainer_me["auctionResultsByFollowedArtists"]>["edges"]>[0]
+    >["node"]
+  >
 }
 
-export const AuctionResultForYouListItem: React.FC<Props> = ({ auctionResult, onPress }) => {
+export const AuctionResultForYouListItem: React.FC<Props> = ({ auctionResult }) => {
```
olerichter00, line 20: Can't we just used the already existing [AuctionResultListItem](https://github.com/artsy/eigen/blob/593f860d35e92358ab63e99f39942d3b16812a60/src/lib/Components/Lists/AuctionResultListItem.tsx#L17) here?

src/lib/Scenes/AuctionResultForYou/AuctionResultForYou.tsx
```diff
@@ -1,239 +1,84 @@
...
-]
+interface Props {
+  me: AuctionResultForYouContainer_me | null
+  relay: RelayPaginationProp
+}
+
+export const AuctionResultForYou: React.FC<Props> = ({ me, relay }) => {
+  const { hasMore, isLoading, loadMore } = relay
+  const [loadingMoreData, setLoadingMoreData] = useState(false)
+  const auctionResultData = me?.auctionResultsByFollowedArtists?.edges
```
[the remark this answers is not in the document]
JenqaaaBY: I tried to do this
This data format does not suit me

Screen Shot 2021-06-21 at 3 14 11 PM
Screen Shot 2021-06-21 at 3 14 21 PM

src/lib/Scenes/AuctionResultForYou/AuctionResultForYouContainer.tsx
JenqaaaBY: You're right. It doesn't have to be used here

src/lib/Scenes/AuctionResultForYou/AuctionResultForYou.tsx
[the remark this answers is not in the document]
olerichter00: I think it should work. I don't see that `cursor` is used anywhere else in the component and you could replace `item?.node` and `item.node` everywhere with `item`.

src/lib/Scenes/Home/Components/AuctionResultsRail.tsx
JenqaaaBY: Yes, you are right, we can not use this check, since we indicate the number of elements in the request itself

src/lib/Scenes/AuctionResultForYou/AuctionResultForYou.tsx
```diff
@@ -1,236 +1,184 @@
...
+  {
+    me: graphql`
+      fragment AuctionResultForYou_me on Me
+      @argumentDefinitions(first: { type: "Int", defaultValue: 10 }, after: { type: "String" }) {
+        auctionResultsByFollowedArtists(first: $first, after: $after)
+          @connection(key: "AuctionResultForYouContainer_auctionResultsByFollowedArtists") {
+          totalCount
+          edges {
+            node {
+              ...AuctionResultListItem_auctionResult
```
olerichter00, line 115: Can we either remove the fragment or the fields below the fragment? Now we have all of them here and in the fragment as well.

olerichter00: I guess it should be enough to just use the fragment here.

olerichter00 (commented):
That looks really good! Just some small comments.

src/lib/Components/Lists/AuctionResultListItem.tsx
```diff
@@ -53,7 +53,10 @@ const AuctionResultListItem: React.FC<Props> = ({ auctionResult, onPress }) => {
         {/* Sale Artwork Details */}
         <Flex pl={15} flex={1} flexDirection="row" justifyContent="space-between">
           <Flex flex={3}>
-            <Flex flexDirection="row" mb={"3px"}>
+            <Flex mb={"3px"}>
```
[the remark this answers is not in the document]
JenqaaaBY: In this case, the indents are different.
I could not find the value in such a format so that the indent was 3px

src/lib/Scenes/AuctionResultForYou/AuctionResultForYou.tsx
```diff
@@ -1,236 +1,184 @@
...
-      ...FeaturedArtists_collection
-      ...CollectionHubsRails_collection
+          )}
+          ItemSeparatorComponent={() => (
+            <Flex px={2}>
+              <Separator />
+            </Flex>
+          )}
+          ListFooterComponent={loadingMoreData ? <Spinner style={{ marginTop: 20, marginBottom: 20 }} /> : null}
+          style={{ width: useScreenDimensions().width }}
```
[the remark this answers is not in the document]
JenqaaaBY: These properties do not exist here (width, mt and mb) and therefore I have to use styles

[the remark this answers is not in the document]
olerichter00: Ah I see - then let's just use `style` here :)

olerichter00:
#squashongreen

src/lib/Components/Lists/AuctionResultListItem.tsx
```diff
@@ -107,6 +110,10 @@ export const AuctionResultFragmentContainer = createFragmentContainer(AuctionRes
       dateText
       id
       internalID
+      artist {
+        name
+        slug
```
MounirDhahri, line 115: **thought:** I don't see where you are using the slug here

src/lib/Scenes/AuctionResultsForYou/AuctionResultsForYou.tsx
```diff
@@ -0,0 +1,162 @@
...
+interface Props {
+  me: AuctionResultsForYou_me | null
+  relay: RelayPaginationProp
+}
+
+export const AuctionResultsForYou: React.FC<Props> = ({ me, relay }) => {
+  const { hasMore, isLoading, loadMore } = relay
+  const [loadingMoreData, setLoadingMoreData] = useState(false)
+  const auctionResultData = extractNodes(me?.auctionResultsByFollowedArtists)
+  const groupedAuctionResultData = groupBy(auctionResultData, (item) => moment(item!.saleDate!).format("MMM"))
```
MounirDhahri, line 32: **issue:** I am not sure this is the proper way to be formatting the auction results. This way, auction results from august 2020 and august 2021 will be both in the same group I guess.

src/lib/Scenes/AuctionResultsForYou/AuctionResultsForYou.tsx
```diff
@@ -0,0 +1,162 @@
...
+  const groupedAuctionResultSectionData = Object.keys(groupedAuctionResultData).map((key) => {
+    return { date: key, data: groupedAuctionResultData[key] }
```
MounirDhahri, line 34: **issue:** I think the date should be the full date and not the `MMM` formatted date

src/lib/Scenes/AuctionResultsForYou/AuctionResultsForYou.tsx
```diff
@@ -0,0 +1,162 @@
...
+          <LinkText onPress={() => navigate("/favorites", { passProps: { initialTab: Tab.artists } })}>
+            artists you follow
+          </LinkText>
+          . You can also look up more auction results on the insights tab on any artist’s page.
+        </Text>
+        <SectionList
+          sections={groupedAuctionResultSectionData}
+          onEndReachedThreshold={0.5}
+          onEndReached={loadMoreArtworks}
+          keyExtractor={(item, index) => item.internalID + index.toString()}
```
MounirDhahri, line 65: **issue:** Since this is a list that can get more items when refreshed (if we implement pull to refresh), we should not be using the index as a `keyExtractor` since it will change whenever new auction results are added

src/lib/Scenes/AuctionResultsForYou/AuctionResultsForYou.tsx
```diff
@@ -0,0 +1,162 @@
...
+            ) : (
+              <></>
+            )
+          }
+          renderSectionHeader={({ section: { date } }) => (
+            <Text
+              textAlign="left"
+              color="black"
+              style={{ fontFamily: Fonts.Unica77LLRegular, fontSize: 18, marginLeft: 20 }}
+            >
```
MounirDhahri, line 81: **issue:** were none of the text variants useful here?

src/lib/navigation/routes.tsx
```diff
@@ -147,7 +147,7 @@ function getDomainMap(): Record<string, RouteMatcher[] | null> {
 
     new RouteMatcher("/inquiry/:artworkID", "Inquiry"),
     new RouteMatcher("/viewing-rooms", "ViewingRooms"),
-    new RouteMatcher("/auction-result-for-you", "AuctionResultForYou"),
+    new RouteMatcher("/auction-results-for-you", "AuctionResultsForYou"),
```
MounirDhahri, line 150: **issue:** do you mind if you add this route to the `AndroidManifest.xml` as well

MounirDhahri (commented):
Nice work here @JenqaaaBY ! I left few comments a bit late. Would be great if you address them

olerichter00:
@MounirDhahri I already added the #squashongreen label before I saw your comments. I'll create a separate story to address them.

src/lib/Scenes/AuctionResultsForYou/AuctionResultsForYou.tsx
olerichter00: Nice catch!

src/lib/Scenes/AuctionResultsForYou/AuctionResultsForYou.tsx
olerichter00: 👍

MounirDhahri:
That would be great, thank you @olerichter00

PR #4971, opened 2021-06-15
feat (CX-1590): implement share button for articles

kalshannikava:

This PR resolves [CX-1590]

### Description

Share button added to article pages. 

https://user-images.githubusercontent.com/27921300/122058130-9d008e00-cdeb-11eb-9436-ccee1665726e.mov

[CX-1590]: https://artsyproduct.atlassian.net/browse/CX-1590

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -8,21 +8,6 @@ upcoming:
   user_facing:
     - Fix unfollowing categories - katsiaryna alshannikava
     - Remove "Enable push notification" guard for disabling saved search - dzmitry tratsiak
-    - OrderDetails with SoldBy section - alexj105
-    - OrderDetails with headInfo - alexj105
-    - OrderDetails with ArtInfo and ShipsTo - alexj105
-    - fix inaccurate validators for artwork submission - yauheni
-    - Check saved search when the user applied the filter - dzmitry tratsiak
-    - Don't close filter artwork modal when "clear all" tapped - dzmitry tratsiak
-    - Continue to show the offer approved banner in an inquiry offer conversation after fulfillment - erikdstock
-    - Add link to Conditions of Sale in sign up - dzmitry
-    - Add saved search banner update banner state automatically on save/filter actions - kizito
-    - Show visual feedback when the "Sort & Filter" button is pressed - dzmitry tratsiak
-    - Don't show saved search banner when only sort filter is selected - dzmitry tratsiak
-    - Add articles (market news) screen - ole
-    - Add Filter Component to Collections Page - kizito
-    - Check push notification permissions before saving search - kizito
-    - Implement infinite scroll on android - kizito
     - Implement share button for articles - katsiaryna alshannikava, mounir
 
 releases:
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -8,6 +8,22 @@ upcoming:
   user_facing:
     - Fix unfollowing categories - katsiaryna alshannikava
     - Remove "Enable push notification" guard for disabling saved search - dzmitry tratsiak
+    - OrderDetails with SoldBy section - alexj105
+    - OrderDetails with headInfo - alexj105
+    - OrderDetails with ArtInfo and ShipsTo - alexj105
+    - fix inaccurate validators for artwork submission - yauheni
+    - Check saved search when the user applied the filter - dzmitry tratsiak
+    - Don't close filter artwork modal when "clear all" tapped - dzmitry tratsiak
+    - Continue to show the offer approved banner in an inquiry offer conversation after fulfillment - erikdstock
+    - Add link to Conditions of Sale in sign up - dzmitry
+    - Add saved search banner update banner state automatically on save/filter actions - kizito
+    - Show visual feedback when the "Sort & Filter" button is pressed - dzmitry tratsiak
+    - Don't show saved search banner when only sort filter is selected - dzmitry tratsiak
+    - Add articles (market news) screen - ole
+    - Add Filter Component to Collections Page - kizito
+    - Check push notification permissions before saving search - kizito
+    - Implement infinite scroll on android - kizito
+    - Implement share button for articles - katsiaryna alshannikava, mounir
 
 releases:
   - version: 6.9.5
--- a/src/lib/Components/ArtsyReactWebView.tsx
+++ b/src/lib/Components/ArtsyReactWebView.tsx
@@ -1,13 +1,16 @@
+import { OwnerType } from "@artsy/cohesion"
 import { color } from "@artsy/palette-tokens"
 import { addBreadcrumb } from "@sentry/react-native"
 import { goBack, navigate } from "lib/navigation/navigate"
 import { matchRoute } from "lib/navigation/routes"
 import { getCurrentEmissionState, GlobalStore, useEnvironment } from "lib/store/GlobalStore"
+import { Schema } from "lib/utils/track"
 import { useScreenDimensions } from "lib/utils/useScreenDimensions"
 import { parse as parseQueryString } from "query-string"
 import React, { useEffect, useRef, useState } from "react"
-import { Platform, View } from "react-native"
+import { Platform, Share, View } from "react-native"
 import WebView, { WebViewProps } from "react-native-webview"
+import { useTracking } from "react-tracking"
 import { parse as parseURL } from "url"
 import { ArtsyKeyboardAvoidingView } from "./ArtsyKeyboardAvoidingView"
 import { FancyModalHeader } from "./FancyModal/FancyModalHeader"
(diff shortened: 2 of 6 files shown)


CHANGELOG.yml
```diff
@@ -20,64 +20,64 @@ upcoming:
...
-     - Remove feature flag check for artist nationality filter - devon
-     - Remove feature flag check for custom size filter - devon
-     - Fix iOS app blank screen - ole, mounir
-     - Replace "color" filter with "colors" filter - dzmitry tratsiak
-     - Add Saved Search feature flag - kizito
-     - Send only changed filter params in the analytics event - dzmitry tratsiak
-     - Update PR template to not automatically attach an unrelated Jira ticket - erikdstock
-     - Add createSavedSearch mutation - dzmitry tratsiak
-     - Add deleteSavedSearch mutation - dzmitry tratsiak
+      - Improve admin menu - dzmitry
```
olerichter00, line 29: 👀 Is this intended?

src/lib/Components/ArtsyReactWebView.tsx
```diff
@@ -24,18 +27,50 @@ export interface ArtsyWebViewConfig {
...
-> = ({ url, title, isPresentedModally, allowWebViewInnerNavigation = true, mimicBrowserBackButton = true }) => {
+> = ({
+  url,
+  title,
+  isPresentedModally,
+  allowWebViewInnerNavigation = true,
+  mimicBrowserBackButton = true,
+  showShareButton,
+}) => {
+  console.log("showShareButton", showShareButton)
```
olerichter00, line 51: I guess we can remove this `console.log`.

src/lib/navigation/routes.tsx
```diff
@@ -128,7 +128,7 @@ function getDomainMap(): Record<string, RouteMatcher[] | null> {
     })),
     new RouteMatcher("/artist/:artistID/auction-result/:auctionResultInternalID", "AuctionResult"),
     new RouteMatcher("/artist/:artistID/artist-series", "FullArtistSeriesList"),
-    webViewRoute("/artist/:artistID/articles"),
+    webViewRoute("/artist/:artistID/articles", { showShareButton: true }),
```
olerichter00, line 131: I'm not sure if we want to show the share button here as well. Isn't this a route for a list of articles (actually I haven't seen this one yet)?

src/lib/navigation/routes.tsx
```diff
@@ -219,7 +219,7 @@ function getDomainMap(): Record<string, RouteMatcher[] | null> {
       : new RouteMatcher("/orders/:orderID", "Checkout"),
 
     new RouteMatcher("/:slug", "VanityURLEntity"),
-    webViewRoute("/*"),
+    webViewRoute("/*", { showShareButton: true }),
```
olerichter00, line 222: I don't think we need a share button here as well.

src/lib/navigation/routes.tsx
kalshannikava: which route is for it then? I can't find another one

kalshannikava: we need to use a route for a single article page. actually, when I pass showShareButton here it doesn't appear which makes me think that you're right. but I still can't find another route for an article

CHANGELOG.yml
kalshannikava: @olerichter00 well vscode adds these spaces automatically on save action - I thought it's part of project linting configuration

olerichter00: Interestingly, I had two conflicting formatter in vscode that prevented the file from being automatically formated. When I set a default it's doing the same formatting :)

olerichter00:
#squashongreen

CHANGELOG.yml
kalshannikava: @olerichter00 so, does it require fixing?

olerichter00: @kalshannikava sorry, I should have marked my comment as non-blocking. It doesn't require fixing and is probably intended :)

src/lib/navigation/routes.tsx
kalshannikava: fixed

src/lib/Components/ArtsyReactWebView.tsx
kalshannikava: fixed

CHANGELOG.yml
kalshannikava: @olerichter00 thank you! :)
could this PR be merged then?

src/lib/Components/FancyModal/FancyModalHeader.tsx
```diff
@@ -31,6 +33,14 @@ export const FancyModalHeader: React.FC<FancyModalHeaderProps> = ({
     }
   }
 
+  const rightButton = () => {
+    if (!useShareButton) {
+      return <ArrowRightIcon fill="black100" />
```
olerichter00, line 38: Hmm, do we really need an `ArrowRightIcon` here?

olerichter00 (approved):
Looks great!

I just added one question. When we resolve this, we can merge this PR 🚀

src/lib/Components/FancyModal/FancyModalHeader.tsx
kalshannikava: It was the initial implementation of the header component which I didn't want to remove. But taking a look now, I can't find any place in the code where this arrow would be expected. So I guess it can be removed?

olerichter00: Ah yes, now I see it as well. I guess we can leave the `ArrowRightIcon` then and merge it 🚀

olerichter00 (approved):

CHANGELOG.yml
```diff
@@ -20,64 +20,64 @@ upcoming:
...
+      - Add saved search banner (behind feature flag) - dzmitry tratsiak
+      - Add articles rail to home screen - ole
+      - Update articles rail design on artist overview screen - ole
+      - Fix home screen placeholder changing with bug 🎉 - ole
+      - Fix android native go back from sale tab to search - yauheni
+      - Vertically aligned 'Follow/Following' artist button text - stas hanchar
+      - Fix card input on android - dzmitry
+      - Add articles rail to home screen - ole
+      - Change articles rail design in artist overview page - ole
+      - Specified a more specific error for logging in with Apple or Facebook - stas hanchar
```
MounirDhahri, line 80: **issue:** do you mind removing those changes? on vs code you can save without formatting if you do it through the command menu

src/lib/Components/ArtsyReactWebView.tsx
```diff
@@ -24,18 +27,49 @@ export interface ArtsyWebViewConfig {
...
+    const shareUrl = ref.current?.shareTitleUrl || uri
+    tracking.trackEvent(tracks.share(shareUrl))
+    try {
+      await Share.share({
+        url: shareUrl,
+      })
+    } catch (error) {
+      console.error("ArtsyReactWebView.tsx", error)
+    }
+  }
```
MounirDhahri, line 72:
```suggestion
  const handleArticleShare = async () =>  {
    const uri = url.startsWith("/") ? webURL + url : url
    /*
     * We only set shareTitleUrl if we navigate to a different URL within the same WebView
     */
    const shareUrl = ref.current?.shareTitleUrl || uri
    tracking.trackEvent(tracks.share(shareUrl))
    try {
      await Share.share({
        url: shareUrl,
      })
    } catch (error) {
      console.error("ArtsyReactWebView.tsx", error)
    }
  }
```

src/lib/Components/ArtsyReactWebView.tsx
```diff
@@ -49,6 +83,8 @@ export const ArtsyReactWebViewPage: React.FC<
               ref.current?.goBack()
             }
           }}
+          useShareButton={showShareButton}
+          onRightButtonPress={() => handleArticleShare()}
```
MounirDhahri, line 87:
```suggestion
          onRightButtonPress={handleArticleShare}
```

src/lib/Components/ArtsyReactWebView.tsx
```diff
@@ -218,3 +257,12 @@ export const __webViewTestUtils__ = __TEST__
       expandGoogleAdLink,
     }
   : null
+
+export const tracks = {
+  share: (slug: string) => ({
+    action: Schema.ActionNames.Share,
+    action_type: Schema.ActionTypes.Tap,
+    context_screen_owner_type: OwnerType.articles,
+    context_screen_owner_slug: slug,
```
MounirDhahri, line 266: @mikehrom we had to not pass the article id here because it wasn't possible with our current infra and it would may be take days to implement it if even possible. So for now we decided to only send the slug. Please let me know if you can manage to get the tracking to work properly with the article slug only.

src/lib/Components/FancyModal/FancyModalHeader.tsx
```diff
@@ -31,6 +33,14 @@ export const FancyModalHeader: React.FC<FancyModalHeaderProps> = ({
...
+    } else {
+      return <ShareIcon fill="black100" height="25px" width="25px" />
+    }
+  }
```
MounirDhahri, line 42: **issue(blocker):** Since the share button is the exception and not the rule, I think this would better be
```suggestion
    if (useShareButton) {
      return <ShareIcon fill="black100" height="25px" width="25px" />
    } else {
      return <ArrowRightIcon fill="black100" />
    }
  }
```

src/lib/navigation/routes.tsx
```diff
@@ -118,6 +118,7 @@ function getDomainMap(): Record<string, RouteMatcher[] | null> {
     new RouteMatcher("/my-profile", "MyProfile"),
     new RouteMatcher("/articles", "Articles"),
     webViewRoute("/articles/:articleID"),
+    webViewRoute("/article/:articleID", { showShareButton: true }),
```
MounirDhahri, line 121: issue(blocker): Would you mind adding a test for this to make sure we get the right param

src/lib/Components/ArtsyReactWebView.tsx
```diff
@@ -24,18 +27,49 @@ export interface ArtsyWebViewConfig {
...
   allowWebViewInnerNavigation?: boolean
+  /**
+   * Show the share URL button
+   */
+  showShareButton?: boolean
 }
 
+type CustomWebView = WebView & { shareTitleUrl: string }
+
 export const ArtsyReactWebViewPage: React.FC<
```
MounirDhahri, line 38: **issue(blocker):** Do you mind adding some tests for the new behaviour you added

MounirDhahri (changes requested):

CHANGELOG.yml
kalshannikava: @MounirDhahri I've checked - this was already merged to master as part of adcc799 commit so I think there is no need to remove it?

src/lib/Components/ArtsyReactWebView.tsx
olerichter00: @MounirDhahri Shall we include this in `preferred_practices.md`?

MounirDhahri: Yes we can definitely!

src/lib/Components/ArtsyReactWebView.tsx
mikehrom: this is fine for now! it's most important that we have the 'share' event implemented and we know an article was the object shared, plus the slug tells us which article it is. `id` is only preferable because sometimes the slug changes, plus the `id` matches the `root_id` in our pageviews/sessions transforms.

MounirDhahri: Great, thank you Mike

olerichter00:
#squashongreen

PR #4988, opened 2021-06-17
feat: convert saved search criteria to filter params

dimatretyak:

This PR resolves [FX-3018]

### Description
Convert [saved search attributes](https://github.com/artsy/eigen/blob/master/data/schema.graphql#L10582-L10603) to the artwork store [filter format](https://github.com/artsy/eigen/blob/master/src/lib/Components/ArtworkFilter/ArtworkFilterHelpers.ts#L172-L178)

#### Demo

https://user-images.githubusercontent.com/3513494/122418625-02db4a00-cf93-11eb-8bba-43bd04d7fc7d.mp4

https://user-images.githubusercontent.com/3513494/122419677-c3f9c400-cf93-11eb-8b7e-987e46417e37.mp4

[FX-3018]: https://artsyproduct.atlassian.net/browse/FX-3018

### DIFF
--- a/src/lib/Components/ArtworkFilter/Filters/helpers.ts
+++ b/src/lib/Components/ArtworkFilter/Filters/helpers.ts
@@ -1,7 +1,7 @@
 import { LegacyNativeModules } from "lib/NativeModules/LegacyNativeModules"
 import { isNull, isUndefined, round as __round__ } from "lodash"
 
-export type Numeric = "*" | number
+type Numeric = "*" | number
 type Unit = "in" | "cm"
 
 export interface Range {
--- a/src/lib/Components/ArtworkFilter/Filters/helpers.ts
+++ b/src/lib/Components/ArtworkFilter/Filters/helpers.ts
@@ -1,7 +1,7 @@
 import { LegacyNativeModules } from "lib/NativeModules/LegacyNativeModules"
 import { isNull, isUndefined, round as __round__ } from "lodash"
 
-type Numeric = "*" | number
+export type Numeric = "*" | number
 type Unit = "in" | "cm"
 
 export interface Range {
@@ -104,4 +104,18 @@ export const parseRange = (range: string): Range => {
   })
 
   return { min: enforceNumeric(min), max: enforceNumeric(max) }
+}
+
+export const parsePriceRangeLabel = (min: Numeric, max: Numeric) => {
+  let parsedMin = "$0"
+  let parsedMax = "+"
+
+  if (min !== "*") {
+    parsedMin = `$${min.toLocaleString("en-US", { maximumFractionDigits: 2 })}`
+  }
+  if (max !== "*") {
+    parsedMax = `–${max.toLocaleString("en-US", { maximumFractionDigits: 2 })}`
+  }
+
+  return `${parsedMin}${parsedMax}`
 }
--- a/src/lib/Components/ArtworkFilter/SavedSearch/__tests__/convertersToFilterParams-tests.ts
+++ b/src/lib/Components/ArtworkFilter/SavedSearch/__tests__/convertersToFilterParams-tests.ts
@@ -60,17 +60,11 @@ describe("convertPriceToFilterParam", () => {
 
 describe("convertSizeToFilterParams", () => {
   it("returns the default size filter value", () => {
-    expect(convertSizeToFilterParams({})).toEqual([
-      {
-        displayText: "All",
-        paramValue: "*-*",
-        paramName: FilterParamName.dimensionRange,
-      },
-    ])
+    expect(convertSizeToFilterParams({})).toEqual(null)
   })
 
   it("returns the small size filter value", () => {
-    expect(convertSizeToFilterParams({ dimensionScoreMax: 16 })).toEqual([
+    expect(convertSizeToFilterParams({ dimensionRange: "*-16.0" })).toEqual([
       {
         displayText: "Small (under 16in)",
         paramValue: "*-16.0",
@@ -80,7 +74,7 @@ describe("convertSizeToFilterParams", () => {
   })
 
   it("returns the medium size filter value", () => {
-    expect(convertSizeToFilterParams({ dimensionScoreMin: 16, dimensionScoreMax: 40 })).toEqual([
+    expect(convertSizeToFilterParams({ dimensionRange: "16.0-40.0" })).toEqual([
       {
         displayText: "Medium (under 16in – 40in)",
         paramValue: "16.0-40.0",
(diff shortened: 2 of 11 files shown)


src/lib/Components/ArtworkFilter/SavedSearch/convertersToFilterParams.ts
```diff
@@ -0,0 +1,231 @@
...
+    convertPriceToFilterParam,
+    convertSizeToFilterParams,
+    convertColorsToFilterParam,
+    convertAttributionToFilterParam,
+    convertWaysToBuyToFilterParams,
+    convertMajorPeriodToFilterParam,
+  ]
+
+  converters.forEach((converter) => {
+    const filterParamItem = converter(criteria, aggregationValueByFilterParamName)
```
mdole, line 205: I am a bit surprised that we're passing all of the criteria to each converter - my tendency would have been to pass only the criteria we wanted to convert. That being said, I'm not sure if it's a bad thing necessarily. @dblandin @MounirDhahri do you have opinions on this pattern?

mdole (approved):
nice work @dimatretyak! the thorough specs are especially nice 👏 left one question to get some other opinions on, but mostly I think this is good to go

src/lib/Components/ArtworkFilter/SavedSearch/convertersToFilterParams.ts
```diff
@@ -0,0 +1,265 @@
...
+
+export const convertCustomSizeToFilterParamByName = (
+  paramName: FilterParamName,
+  min: number | null,
+  max: number | null
+) => {
+  const minSizeValue = min ?? "*"
+  const maxSizeValue = max ?? "*"
+  const widthMinLocalized = localizeDimension(minSizeValue, "in")
+  const widthMaxLocalized = localizeDimension(maxSizeValue, "in")
```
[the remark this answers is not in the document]
dimatretyak: Not really. The second argument specifies the basic unit of the value that was passed in the first argument (the values on the backend are stored in inches)

src/lib/Components/ArtworkFilter/Filters/helpers.ts
```diff
@@ -1,7 +1,7 @@
 import { LegacyNativeModules } from "lib/NativeModules/LegacyNativeModules"
 import { isNull, isUndefined, round as __round__ } from "lodash"
 
-type Numeric = "*" | number
+export type Numeric = "*" | number
```
dblandin, line 4: I'm not seeing this type used outside of this file. Do we need the export?

src/lib/Components/ArtworkFilter/SavedSearch/convertersToFilterParams.ts
dimatretyak: I can do this option without any problems. But unfortunately, this will add a lot of repetitive code (such as [availability checks](https://github.com/artsy/eigen/blob/470e4e7e98dfc8e2d14ca7d865f8dcd52f299126/src/lib/Components/ArtworkFilter/SavedSearch/convertersToFilterParams.ts#L207-L209)).

dblandin (approved):
This is looking good!

Still feels a bit odd that we have to prune out elements of the criteria that aren't currently present in the artwork aggregations.

If I remember correctly, I don't think the filter menu would support setting filters that would not be present in the aggregation because you couldn't format or deselect them. Is that right?

Acknowledging that potential confusion, I think this should work for most use cases and in subsequent iterations I don't think we'll be navigating to this artist screen's artwork grid so hopefully these edge cases will be short-lived.

src/lib/Components/ArtworkFilter/SavedSearch/convertersToFilterParams.ts
dblandin: > I am a bit surprised that we're passing all of the criteria to each converter

If this simplifies the code, I don't mind passing the whole criteria and letting each converter determine which attributes are relevant. We actually take a similar approach in gravity when processing API request params ([example](https://github.com/artsy/gravity/pull/14195/files#diff-d42c69934591dd0072fadb9dea3deea0437379901f3b0c7f92880d8f8a5c9d01)).

dimatretyak:
> If I remember correctly, I don't think the filter menu would support setting filters that would not be present in the aggregation because you couldn't format or deselect them. Is that right?

Yes, that's right. Only the data that is in the aggregations will be displayed. 

Could there potentially be a use case where the user made a saved search long ago (for example, a year ago) and some of the data (for example, some materials) in the aggregations are no longer available?

src/lib/Components/ArtworkFilter/Filters/helpers.ts
dimatretyak: My bad. After yesterday's changes, I forgot to remove the export

src/lib/Components/ArtworkFilter/SavedSearch/convertersToFilterParams.ts
mdole: Ok, then let's keep it as-is! Thanks for chatting about it :)

src/lib/Components/ArtworkFilter/SavedSearch/convertersToFilterParams.ts
```diff
@@ -0,0 +1,231 @@
...
+    }
+  }
+
+  return null
+}
+
+export const convertSavedSearchCriteriaToFilterParams = (
+  criteria: SearchCriteriaAttributes,
+  aggregations: Aggregations
+) => {
```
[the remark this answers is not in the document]
dimatretyak: Done

src/lib/Components/ArtworkFilter/Filters/helpers.ts
dimatretyak: Done

src/lib/Components/ArtworkFilter/SavedSearch/convertersToFilterParams.ts
```diff
@@ -0,0 +1,231 @@
...
+export const convertColorsToFilterParam = (
+  criteria: SearchCriteriaAttributes,
+  aggregation: AggregationByFilterParamName
+): FilterData | null => {
+  const colorsValue = criteria[FilterParamName.colors]
+
+  if (Array.isArray(colorsValue) && colorsValue.length > 0) {
+    const colorFromAggregationByValue = keyBy(aggregation[FilterParamName.colors], "value")
+    const availableColors = colorsValue.filter((color) => !!colorFromAggregationByValue[color])
+    const colorNames = availableColors.map((color) => COLORS_INDEXED_BY_VALUE[color].name)
```
MounirDhahri, line 91: **thought:** we may be adjusting gemini to contain new colors that do not exist in `COLORS_INDEXED_BY_VALUE` in the future, I think we should probably make sure first that the colors do exist in the `COLORS_INDEXED_BY_VALUE`

src/lib/Components/ArtworkFilter/SavedSearch/convertersToFilterParams.ts
```diff
@@ -0,0 +1,231 @@
...
+
+    if (filterParamItem) {
+      filterParams = filterParams.concat(filterParamItem)
+    }
+  })
+
+  // Extract value names from aggregation
+  shouldExtractValueNamesFromAggregation.forEach((filterParamName) => {
+    const aggregationValue = aggregationValueByFilterParamName[filterParamName]
+    const criteriaValue = criteria[filterParamName as SearchCriteriaAttributeKeys] as string[]
```
MounirDhahri, line 215: **praise:** 🙌

src/lib/Components/ArtworkFilter/SavedSearch/convertersToFilterParams.ts
```diff
@@ -0,0 +1,231 @@
...
+
+    if (aggregationValue) {
```
MounirDhahri, line 217: **thought:** shouldn't this be `if (aggregationValue !== undefined)` just in case the `aggregationValue` is a falsy value?

src/lib/Components/ArtworkFilter/SavedSearch/__tests__/convertersToFilterParams-tests.ts
```diff
@@ -0,0 +1,668 @@
+import { SearchCriteriaAttributes } from "__generated__/SavedSearchBannerQuery.graphql"
```
MounirDhahri, line 1: **praise:** nice tests

MounirDhahri (commented):
Great work here. I left few questions that are fine to be addressed in a future PR.

src/lib/Components/ArtworkFilter/SavedSearch/convertersToFilterParams.ts
dimatretyak: @MounirDhahri  I think it would even be better to use `isNill` from `lodash`

dimatretyak: In this use case, it should always be an array of strings. It might be better to add an `Array. isArray` check to improve the code readability

MounirDhahri: perfect then, thanks for the quick answer

PR #4993, opened 2021-06-18
fix: correct round cm values during inch convert

dimatretyak:

This PR resolves []

### Description
Rounding the value when converting from inches to centimeters globally


dzucconi:
I'm fine with this, just wondering how this cropped up in a bug? We'd never want to keep the precision?

dimatretyak:
> I'm fine with this, just wondering how this cropped up in a bug? We'd never want to keep the precision?

I agree with you. Perhaps I was too hasty with the changes. I was thinking of doing the rounding globally so that we don't have to manually round each time

dzucconi:
Could also add an option to round and default it to true

PR #4995, opened 2021-06-21
feat: Sticky header on show screen

anastasiapyzhik:

[FX-2908]

This PR resolves []

### Description

I incapsulated logic of sticky header from [here](https://github.com/artsy/eigen/pull/4932) and applied for Show screen

**Before**
Simulator Screen Shot - iPhone 8 - 2021-06-21 at 13 20 03

**After**
after

[FX-2908]: https://artsyproduct.atlassian.net/browse/FX-2908

### DIFF
--- a/src/lib/Components/HeaderArtworksFilter.tsx
+++ b/src/lib/Components/HeaderArtworksFilter.tsx
@@ -64,8 +64,8 @@ export const HeaderArtworksFilter: React.FC<FilterProps> = ({ total, animationVa
     BACK_BUTTON_SIZE.top -
     PixelRatio.getPixelSizeForLayoutSize(extraOffset)
 
-  return (
-    <Box backgroundColor="white" onLayout={(e) => _onLayout(e)}>
+  const SeparatorWithSmoothOpacity = () => {
+    return (
       <Animated.View
         style={{
           opacity: animationValue.interpolate({
@@ -77,6 +77,16 @@ export const HeaderArtworksFilter: React.FC<FilterProps> = ({ total, animationVa
       >
         <Separator mt={2} width={screenWidth * 2} />
       </Animated.View>
+    )
+  }
+
+  if (!total) {
+    return <SeparatorWithSmoothOpacity />
+  }
+
+  return (
+    <Box backgroundColor="white" onLayout={(e) => _onLayout(e)}>
+      <SeparatorWithSmoothOpacity />
       <Animated.View
         style={{
           transform: [
--- a/src/lib/Components/HeaderArtworksFilter.tsx
+++ b/src/lib/Components/HeaderArtworksFilter.tsx
@@ -1,0 +1,133 @@
+import { isPad } from "lib/utils/hardware"
+import { useScreenDimensions } from "lib/utils/useScreenDimensions"
+import { Box, FilterIcon, Flex, Separator, Text, TouchableHighlightColor } from "palette"
+import React, { useEffect, useState } from "react"
+import { Animated, Dimensions, LayoutChangeEvent, PixelRatio } from "react-native"
+
+interface FilterProps {
+  total: number
+  animationValue: Animated.Value
+  onPress: () => void
+}
+
+const pixelRatio = PixelRatio.get()
+// values based on px used in <BackButton>
+const BACK_BUTTON_SIZE = {
+  top: isPad() ? 10 / pixelRatio : 14 / pixelRatio,
+  left: isPad() ? 20 / pixelRatio : 10 / pixelRatio,
+  right: 0,
+  bottom: 0,
+  image: {
+    height: 40,
+    width: 40,
+  },
+}
+
+export const HeaderArtworksFilter: React.FC<FilterProps> = ({ total, animationValue, onPress }) => {
+  const screenWidth = useScreenDimensions().width
+
+  const [onLayoutCalled, setOnLayoutCalled] = useState(false)
+  const [filterPageY, setPageY] = useState(0)
+
+  useEffect(() => {
+    // orientation changed, allow for recalculation of pageY
+    Dimensions.addEventListener("change", orientationChanged)
+    return () => {
+      Dimensions.removeEventListener("change", orientationChanged)
+    }
+  }, [])
+
+  const orientationChanged = () => {
+    setOnLayoutCalled(false)
+  }
+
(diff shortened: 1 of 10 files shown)


src/lib/Scenes/Collection/Components/CollectionArtworksFilter.tsx
```diff
@@ -75,107 +48,18 @@ export const CollectionArtworksFilter: React.FC<FilterProps> = ({ collection, an
...
+      <HeaderArtworksFilter total={artworksTotal} animationValue={animationValue} onPress={openFilterArtworksModal} />
+      <ArtworkFilterNavigator
+        id={collection.id}
+        slug={collection.slug}
+        isFilterArtworksModalVisible={isFilterArtworksModalVisible}
+        exitModal={toggleFilterArtworksModal}
+        closeModal={closeFilterArtworksModal}
+        mode={FilterModalMode.Collection}
+      />
+    </>
```
dimatretyak, line 62: 👍

src/lib/Scenes/Show/Components/ShowArtworksFilter.tsx
```diff
@@ -0,0 +1,9 @@
+import { ArtworksFiltersStore } from "lib/Components/ArtworkFilter/ArtworkFilterStore"
+import { HeaderArtworksFilter } from "lib/Components/HeaderArtworksFilter"
+import React from "react"
+
+export const ShowArtworksFilter: React.FC<any> = (props) => {
```
mdole, line 5: We should avoid using `any` if possible! Based on how this is used, I would expect the props to look something like:
```typescript
interface ShowArtworksFilterProps {
  animationValue: Animated.Value
  onPress: () => void
}
```
How does that sound to you @anastasiapyzhik ?

src/lib/Components/HeaderArtworksFilter.tsx
```diff
@@ -0,0 +1,144 @@
...
+                      inputRange: filterPageY > 0 ? [0, filterPageY - ANIM_START, filterPageY] : [0, 0, 0],
+                      outputRange: filterPageY > 0 ? [0, 0, TRANSLATE_X_VALUE] : [0, 0, 0],
+                      extrapolate: "clamp",
+                    }),
+                  },
+                ],
+              }}
+            >
+              <Text variant="subtitle" color="black60">
+                Showing {total} works
```
lordkiz, line 124: Nice work @anastasiapyzhik . Do you want to show "Showing 0 artworks" and the entire header alongside the empty state when artworks total is 0?

src/lib/Scenes/Show/Show.tsx
```diff
@@ -70,19 +65,17 @@ export const Show: React.FC<ShowProps> = ({ show }) => {
...
       element: (
-        <Box mx={2}>
-          <Separator />
-        </Box>
+        <ShowArtworksFilter animationValue={filterComponentAnimationValue} onPress={toggleFilterArtworksModal} />
       ),
     },
 
+    ...(Boolean(show.viewingRoomIDs.length)
+      ? [{ key: "viewing-room", element: <ShowViewingRoom show={show} mx={2} /> }]
```
mdole, line 76: I believe this displays an additional element if the show is associated with a viewing room - we should make sure to look at one as a test to make sure there are no weird UI interactions between the artworks filter and the `<ShowViewingRoom>` component. Here's an example of a VR with a show associated: https://staging.artsy.net/show/xavier-hufkens-cathy-wilkes-2

mdole (changes requested):
Nice work @anastasiapyzhik ! Couple comments to resolve and then I think this is good to go :)

src/lib/Scenes/Show/Components/ShowArtworks.tsx
```diff
@@ -86,7 +87,14 @@ const ShowArtworks: React.FC<Props> = ({ show, relay, initiallyAppliedFilter })
     setAggregationsAction(artworkAggregations)
   }, [])
 
-  if ((artworks?.counts?.total ?? 0) === 0) {
+  useEffect(() => {
+    setFiltersCountAction({
```
lordkiz, line 91: Looks good. However, are we certain we always want to reset the store's `followedArtists` to null? Especially since this PR did not change that property. If we use another component at the the same level with this component and nested within the same StoreProvider, if that component makes a change to followedArtist, this component will overwrite it. 
In CollectionArtworks we did something like:
```
const counts = ArtworksFiltersStore.useStoreState((state) => state.counts)
useEffect(() => {
    // for use by CollectionArtworksFilter to keep count
    const filterCount = { ...counts, ...artworks?.counts }
    setFiltersCountAction(filterCount)
  }, [artworksTotal])
```

src/lib/Scenes/Show/Components/ShowArtworksFilter.tsx
anastasiapyzhik: Ahh I just put `any` temporarily and forgot about it. Thanks for the note!

src/lib/Components/HeaderArtworksFilter.tsx
anastasiapyzhik: In case artworks total is 0 the component returns separator only https://github.com/artsy/eigen/blob/320ed16f6085c596514547344077417ae9065437/src/lib/Components/HeaderArtworksFilter.tsx#L88-L90

src/lib/Scenes/Show/Show.tsx
anastasiapyzhik: There are no any works so I added the header with 0 works in order to show you what it looks like

for

Please don't mind the UI freezing, I just have a slow mac and can't help it. It happens on demo only, not in simulator

src/lib/Components/HeaderArtworksFilter.tsx
lordkiz: Oh. I totally missed that line. 👍🏾👍🏾👍🏾

src/lib/Scenes/Show/Components/ShowArtworks.tsx
anastasiapyzhik: That's a good point, thanks!

anastasiapyzhik:
Please don't merge yet, I have to add a small update

mdole (approved):
Nice work! Let's ship it!

PR #5005, opened 2021-06-21
fix: Order History: "Unable to load" page is opened (exception occurs) when artwork is deleted/unpublished from order (PX-4253)

Serge0n:

This PR resolves [PX-4253]

image
[PX-4253]: https://artsyproduct.atlassian.net/browse/PX-4253

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -7,6 +7,7 @@ upcoming:
     - Update open simulator command in getting started docs - ole
     - Update debugging docs - mounir
   user_facing:
+    - Fixed "Unable to load" error on OrderHistory page when artwork is deleted/unpublished from order
     - Pick up should be displayed in Sold by section - Serge0n
     - Country should be written with full name - Serge0n
     - Name from shipping address should be displayed instead of account name - Serge0n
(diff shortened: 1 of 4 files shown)


sweir27:
@Serge0n Can you screenshot what the "Order Details" page looks like in the case where the artwork is deleted? I'm wondering if we need to gracefully degrade that view as well.

brainbicycle (approved):
Nice!

Serge0n:
> @Serge0n Can you screenshot what the "Order Details" page looks like in the case where the artwork is deleted? I'm wondering if we need to gracefully degrade that view as well.

Yes, there is the same bug in OrderDetails. I'll fix it too

MounirDhahri:
This is great work but I think this won't be fixing the issue as adding the @principalField directive only suppresses the warning on dev but the issue remains there for production builds. I guess we have 2 ways to fix this for now: First way, would be probably from metaphysics level, I think that's the impact over perfection approach. The second way, would be probably adjust the way we create orders and instead of referring to the original artwork, we will need to be referring to a snapshot of the artwork that was ordered. 
I will be pairing with @Serge0n on the metaphysics approach, but I strongly suggest that we consider the second approach for a more robust solution. Please correct me if I am wrong.

PR #5015, opened 2021-06-23
feat(CX-1603): Add new onboarding feature flag

edez1t:

This PR resolves [CX-1603]

### Description

https://user-images.githubusercontent.com/73339470/123070452-714d5b80-d41c-11eb-8247-c28534b2099d.mov

[CX-1603]: https://artsyproduct.atlassian.net/browse/CX-1603


src/lib/store/AuthModel.ts
```diff
@@ -101,7 +100,7 @@ export const getAuthModel = (): AuthModel => ({
     })}`
     const result = await fetch(tokenURL, {
       headers: {
-        "User-Agent": context.getStoreState().native.sessionState.userAgent,
+        "User-Agent": getCurrentEmissionState().userAgent,
```
edez1t, line 103: context.getStoreState().native.sessionState.userAgent returned undefined for android and wasn't used on ios

src/lib/store/AuthModel.ts
```diff
@@ -127,7 +126,7 @@ export const getAuthModel = (): AuthModel => ({
       headers: {
         "X-Xapp-Token": xAppToken,
         Accept: "application/json",
-        "User-Agent": context.getStoreState().native.sessionState.userAgent,
+        "User-Agent": getCurrentEmissionState().userAgent,
```
edez1t, line 129: same here

src/lib/store/AuthModel.ts
```diff
@@ -78,9 +79,7 @@ export const getAuthModel = (): AuthModel => ({
   xApptokenExpiresIn: null,
   onboardingState: "none",
   userEmail: computed([(_, store) => store], (store) => {
-    if (Platform.OS === "ios") {
-      return store.native.sessionState.userEmail
-    } else if (Platform.OS === "android") {
+    if (Platform.OS === "android" || Platform.OS === "ios") {
       return store.auth.androidUserEmail
     }
     return null
```
edez1t, line 85: these logic wasn't used on ios, so we don't need feat flag

src/lib/store/config/features.ts
```diff
@@ -105,6 +105,11 @@ export const features = defineFeatures({
     description: "Enable Saved Search",
     showInAdminMenu: true,
   },
+  AREnableNewOnboardingFlow: {
+    readyForRelease: false,
+    description: "Enable new onboarding flow",
+    showInAdminMenu: Platform.OS === "ios",
```
pvinis, line 111: 👨‍🍳👌

src/lib/store/GlobalStore.tsx
```diff
@@ -164,13 +164,13 @@ export function unsafe_getDevToggle(key: DevToggleName) {
 
 export function getCurrentEmissionState() {
   const state = globalStoreInstance().getState() ?? null
-  if (Platform.OS === "ios") {
+  if (Platform.OS === "ios" && !unsafe_getFeatureFlag("AREnableNewOnboardingFlow")) {
```
pvinis, line 167: I think it makes it cleaner to have clear ifs (only with the flag check) that wrap code, so we can easily get rid of them when we remove the flag too.

Something like
```
if (!flag(AREnable) {
  if(Platform === "ios") {
    ..
  }
}
```

What do you think?

src/lib/store/AuthModel.ts
```diff
@@ -78,9 +79,7 @@ export const getAuthModel = (): AuthModel => ({
   xApptokenExpiresIn: null,
   onboardingState: "none",
   userEmail: computed([(_, store) => store], (store) => {
-    if (Platform.OS === "ios") {
-      return store.native.sessionState.userEmail
-    } else if (Platform.OS === "android") {
+    if (Platform.OS === "android" || Platform.OS === "ios") {
       return store.auth.androidUserEmail
```
pvinis, line 83: Could you rename this from `androidUserEmail` to just `userEmail`? It was named `android*` so it's clear it's only used there. Now we are slowly going for both platforms, we can use a better name :)

src/lib/store/config/features.ts
```diff
@@ -105,6 +105,11 @@ export const features = defineFeatures({
     description: "Enable Saved Search",
     showInAdminMenu: true,
   },
+  AREnableNewOnboardingFlow: {
+    readyForRelease: false,
```
pvinis, line 109: Just a thought: I wonder if instead of true, something like `Platform.OS === "android"` would be helpful here. So basically to be `on` by default.

Actually, while I'm writing this, I kinda answered myself.. It would be a problem if we turn that flag to `false` in echo, and it just overrides the check here 😅.

src/lib/Scenes/VanityURL/VanityURLPossibleRedirect.tsx
```diff
@@ -14,10 +14,11 @@ function join(...parts: string[]) {
 export const VanityURLPossibleRedirect: React.FC<{ slug: string }> = ({ slug }) => {
   const [jsx, setJSX] = useState(<Loading />)
 
+  const showNewOnboarding = useFeatureFlag("AREnableNewOnboardingFlow")
   const authenticationToken =
-    Platform.OS === "ios"
-      ? GlobalStore.useAppState((store) => store.native.sessionState.authenticationToken)
-      : GlobalStore.useAppState((store) => store.auth.userAccessToken!)
+    Platform.OS === "android" || showNewOnboarding
```
pvinis, line 19: Just curious, but why didn't you go for `Platform.OS === "ios" && !showNewOnboarding`?

pvinis (commented):
Just some comments :)

src/lib/Scenes/VanityURL/VanityURLPossibleRedirect.tsx
edez1t: i'll do it, just thought that maybe it will be more readable

pvinis: Both work. I was just curious. For me both read kinda the same, so if you prefer one over the other, do the one you prefer.

src/lib/store/GlobalStore.tsx
edez1t: done

src/lib/store/AuthModel.ts
edez1t: done

pvinis (changes requested):
We should not merge this yet!

Maybe I was unclear with the migration instructions. We need to make a new migration in order to rename.

Also we should check if it works if we turn the flag off again 🤔.

Let's do it together maybe?

src/lib/Scenes/Onboarding/OnboardingCreateAccount/OnboardingCreateAccountName.tsx
```diff
@@ -35,7 +35,6 @@ export const OnboardingCreateAccountName: React.FC<OnboardingCreateAccountNamePr
           handleChange("name")(text)
         }}
         onSubmitEditing={handleSubmit}
-        onBlur={() => validateForm()}
```
edez1t, line 38: We don't need this here. We only validate the form when pressing "next" button. (CX-1629) I guess creating a ticket wan't necessary

edez1t:
> We should not merge this yet!
> 
> Maybe I was unclear with the migration instructions. We need to make a new migration in order to rename.
> 
> Also we should check if it works if we turn the flag off again 🤔.
> 
> Let's do it together maybe?

edez1t:
Screenshot 2021-06-25 at 13 04 39

the only place where we use userEmail. is it good this way? if no it would be better to pair

src/lib/Scenes/Onboarding/OnboardingWelcome.tsx
```diff
@@ -72,7 +78,7 @@ export const OnboardingWelcome: React.FC<OnboardingWelcomeProps> = ({ navigation
   }, [navigation])
 
   return (
-    <Flex flex={1}>
+    <Flex flex={1} style={{ paddingBottom: useScreenDimensions().safeAreaInsets.bottom }}>
```
edez1t, line 81: Yauheni'es changes fixing white bottom padding [CX-1651]

src/lib/store/AuthModel.ts
```diff
@@ -71,20 +71,12 @@ export interface AuthModel {
 
 export const getAuthModel = (): AuthModel => ({
   userID: null,
-  androidUserEmail: null,
   userAccessToken: null,
   userAccessTokenExpiresIn: null,
   xAppToken: null,
   xApptokenExpiresIn: null,
   onboardingState: "none",
-  userEmail: computed([(_, store) => store], (store) => {
```
pvinis, line 80: This here is an important change. the `userEmail` is no longer computed. For android its basically what `androidUserEmail` was a bit further up the file, so just a regular state thing. For ios, we make sure to manually update it from native state, whenever it changes, while the flag is `false`. When the flag is done and `true` and/or when we remove it, we can also remove the `getUserEmail` function and function call as well, as it will no longer use the native state for auth. We will go through the new TS way, like android. That's what ios does in this PR when the flag is `true`.

pvinis:
I think we can merge this now and get it ready for QA.

pvinis (approved):
Let's go :D

PR #5021, opened 2021-06-23
[WIP]feat(CX-882): Reimplement OpaqueImageView

gkartalis:

This PR resolves [CX-882]

### Description

The problem that this ticket solves is to get Images **to load** on Android in a similar and **more performant** way as they do in iOS.

For this reason we decided to go with `react-native-fast-image` library.

This PR reimplements OpaqueImageView (which is the component that renders most of our Images) and also:
- introduces react-native-fast-image
- removes the native code and `AROpaqueImageView` and replaces that with `<OpaqueImageView />`
- updates jest tests to cover new components
- removes native tests related to `AROpaqueImageView`

FastImage component **requires** height and width in contrast with the Image component from react-native. This is why I added the height/width props in some cases like this: src/lib/Components/ArtworkGrids/ArtworkGridItem.tsx

#### Need to be fixed
- native tests iOS
- testing deepzoom
- tablet testing
- videos/screenshots
- cleanup ( I would love to do that once everything is done)

- [] I have included screenshots or videos to illustrate my changes, or I have not changed anything that impacts the UI.

[CX-882]: https://artsyproduct.atlassian.net/browse/CX-882


src/lib/Components/Artist/ArtistShows/ArtistShow.tsx
```diff
@@ -32,7 +33,7 @@ const Show: React.FC<Props> = ({ styles, show }) => {
       <View style={[styles?.container]}>
         <OpaqueImageView
           imageURL={imageURL}
-          style={[styles?.image, { overflow: "hidden", borderRadius: 2, flex: 0 }]}
+          style={[styles?.image, { overflow: "hidden", borderRadius: 2, flex: 0 }] as ImageStyle}
```
gkartalis, line 36: I would love to not use `as ImageStyle` here so any suggestions to avoid that are welcome.
When removing I'm getting the following error:
```
Type 'ImageStyle | undefined' is not assignable to type 'ImageStyle'.
Type 'undefined' is not assignable to type 'ImageStyle'.
```

src/lib/Components/ArtworkGrids/GenericGrid.tsx
```diff
@@ -82,10 +82,9 @@ export class GenericArtworksGrid extends React.Component<Props, State> {
   }
 
   sectionedArtworks() {
-    // @ts-expect-error STRICTNESS_MIGRATION --- 🚨 Unsafe legacy code 🚨 Please delete this and fix any type errors if you have time 🙏
-    const sectionedArtworks = []
-    // @ts-expect-error STRICTNESS_MIGRATION --- 🚨 Unsafe legacy code 🚨 Please delete this and fix any type errors if you have time 🙏
-    const sectionRatioSums = []
+    const sectionedArtworks: GenericGrid_artworks[] = []
```
gkartalis, line 85: removed some of the `// @ts-expect-error`

src/lib/Components/ArtworkGrids/GenericGrid.tsx
```diff
@@ -107,13 +104,12 @@ export class GenericArtworksGrid extends React.Component<Props, State> {
         }
 
         if (sectionIndex != null) {
-          // @ts-expect-error STRICTNESS_MIGRATION --- 🚨 Unsafe legacy code 🚨 Please delete this and fix any type errors if you have time 🙏
           const section = sectionedArtworks[sectionIndex]
+          // @ts-expect-error STRICTNESS_MIGRATION --- 🚨 Unsafe legacy code 🚨 Please delete this and fix any type errors if you have time 🙏
```
gkartalis, line 108: any ideas here about how to remove this as well are welcome 🤓

src/lib/Components/Artist/ArtistShows/ArtistShow.tsx
pvinis: The problem is the `styles?.image`, which could be undefined, and the style array doesn't like it. Doing it like `styles?.image ?? {}` makes sure there is "something" there and not just undefined. Not perfectly clean, but I think nicer than having a typecast. What do you think?

src/lib/Components/ArtworkGrids/InfiniteScrollArtworksGrid.tsx
```diff
@@ -155,7 +155,7 @@ class InfiniteScrollArtworksGrid extends React.Component<Props & PrivateProps, S
       const { shouldAddPadding } = this.props
       const artworkPadding = shouldAddPadding ? space(4) : 0
 
-      return (width - sectionMargins) / (this.props.sectionCount! - artworkPadding)
```
pvinis, line 158: No idea how this calculation worked so far.. 🤔

src/lib/Components/Artist/ArtistShows/ArtistShow.tsx
gkartalis: looks great!

PR #5022, opened 2021-06-23
fix[CX-1604] Facebook login is failing after signing up for a new account

Gon4ar5:
The type of this PR is: Fix

This PR resolves [CX-1604](https://artsyproduct.atlassian.net/browse/CX-1604)

### Steps to Reproduce:

1. Create an account in staging using Facebook auth
2. Log out
3. Log back in using Facebook auth
4. Observe you get error ‘Couldn’t link Facebook account'

### Description

If we login via Facebook for the first time, we don’t have the gravity_session_id cookie (1 screenshot), but before the second login it is (2 screenshots), and because of its presence when sending a request for gravity -> device will authorize us by session ID and sets the value for current_user -> returns an error that the user already exists (3 screenshots).
success
fail
Screenshot 2021-06-22 at 17 02 35

Since the failed login is only affected by the session cookies, I decided to delete them before logging in with facebook.

I'm not sure if the solution with deleting cookies before logging in via facebook is the best solution, but it seemed to me the most quickly implemented, in any case I'm open to suggestions :)

### Result

https://user-images.githubusercontent.com/21379857/123115851-18e08300-d449-11eb-9a45-9a8afb9c09b0.mov

### DIFF
--- a/Artsy/View_Controllers/Login_and_Onboarding/AROnboardingViewController.m
+++ b/Artsy/View_Controllers/Login_and_Onboarding/AROnboardingViewController.m
@@ -623,6 +623,7 @@
 {
     __weak typeof(self) wself = self;
     if (email && ![email isEqualToString:@""]) {
+        [wself deleteGravitySecureSessionCookieToken];
         [[ARUserManager sharedManager] createUserViaFacebookWithToken:token
                                                                 email:email
                                                                  name:name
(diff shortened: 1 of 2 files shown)


MounirDhahri:
I am very sorrry that I didn't get to review this earlier Stas, thanks for this fix.

MounirDhahri (approved):

PR #5023, opened 2021-06-23
feat: Add optionalField directive

olerichter00:

This PR resolves [CX-1132]

🚨 Is blocked by https://github.com/artsy/metaphysics/pull/3242

### Description

- Changes the relay error middleware to not throw when all errors from the response are also included in the `optionalFields` extension.
- Adds the `@optionalField` directive to the `articlesConnection` query of the home screen.

## Context

Adding error handling logic to not throw when a field is marked as optional solves the problem of "Unable to load" screens. We can just annotating fields that are not essential for the page as optional.

We decided to add an `@optionalField` directive and to not use the already existing `@principalField` directive because the `@principalField` only includes errors that have a status code (it's used in Force which has different error handling logic) and the `@optionalField` directive gives us more flexibility. 

All fields can be annotated with the directive `@optionalField` which means that our custom Relay error middleware won't raise an error if the annotated field fails to load. This means that the component has to handle a `null` value to not run into runtime errors.

## Example

**Query:**
```
{ 
  artwork(id: "banksy-love-rat-signed-16")   {
    id
    articles(size: 1)  @optionalField {
      title
    }
  }
}
```

**Response:**
```
{
  "data": {
    "artwork": {
      "id": "QXJ0d29yazo2MDYyMTUxN2RiM2FiMTAwMTJlN2QwOWI=",
      "articles": null
    },
    "errors": [
       {
           "message": "Unexpected error value: [function Error]",
           "path": [
               "artwork",
               "articles"
           ],
        }
    ],
    "extensions": {
      "optionalFields": [
          {
            "httpStatusCode": 500,
            "path": [
              "artwork",
              "articles"
            ]
          },
      ]
    }
 }
```

[CX-1132]: https://artsyproduct.atlassian.net/browse/CX-1132


src/lib/relay/middlewares/errorMiddleware.ts
```diff
@@ -40,7 +40,13 @@ export const principalFieldErrorMiddleware = () => {
       return res
     }
 
-    // at this point, we have errors
+    const allErrorsAreOptional = resJson.extensions?.optionalFields?.length === resJson.errors?.length
```
olerichter00, line 43: Note: We just check if we have as many errors in the response as errors included in the `optionalFields` extension. This should be enough to check if we need to throw or not.

src/lib/relay/middlewares/metaphysicsMiddleware.ts
```diff
@@ -14,9 +14,10 @@ export function metaphysicsExtensionsLoggerMiddleware() {
   return (next) => (req) => {
     // @ts-expect-error STRICTNESS_MIGRATION --- 🚨 Unsafe legacy code 🚨 Please delete this and fix any type errors if you have time 🙏
     return next(req).then((res) => {
-      if (res.json.extensions && console.groupCollapsed) {
+      const requests = res.json.extensions?.requests
+
+      if (requests && console.groupCollapsed) {
```
olerichter00, line 19: We need to check if we have the **request** extension instead of just checking if there are any extensions.

src/lib/relay/middlewares/__tests__/errorMiddleware-tests.ts
```diff
@@ -48,8 +47,65 @@ describe(principalFieldErrorMiddleware, () => {
...
+
+        const res = await middleware(next)(request)
+        expect(res).toBe(relayResponse)
+      })
+    })
+  })
+
   describe("using principalField", () => {
-    // @ts-ignore
     const request: GraphQLRequest = {
```
brainbicycle, line 109: 🧹

src/lib/relay/middlewares/__tests__/errorMiddleware-tests.ts
```diff
@@ -48,8 +47,65 @@ describe(principalFieldErrorMiddleware, () => {
...
+
+        expect.assertions(1)
+        try {
+          await middleware(next)(request)
+        } catch (err) {
+          expect(err.message).toContain("Relay request for `xxx` failed")
+        }
+      })
+
+      it.only("passes through when the optionalField is involved in the errors", async () => {
```
MounirDhahri, line 91: **issue:** do you mind removing the `only` here

src/lib/relay/middlewares/metaphysicsMiddleware.ts
MounirDhahri: **praise:** good catch

src/lib/relay/middlewares/errorMiddleware.ts
MounirDhahri: I agree this should be enough for now. It will be interesting though to see how we will adapt this in the future if we add other directives.

MounirDhahri (approved):
Great work 🎉 I left a small comment that we should be ready to merge once you address it

src/lib/relay/middlewares/__tests__/errorMiddleware-tests.ts
olerichter00: Good catch! Thanks 🙇

olerichter00: ✅

PR #5026, opened 2021-06-23
feat: Add Braze iOS sdk GRO-261

jonallured:
This PR adds the Braze SDK which is required for our integration with them. We're winding down our relationship with Sailthru so in a follow up we'll be able to yank that one out but for now I'm leaving it.

Note: I'm starting off this PR in draft mode because I want to get these keys into Citadel prior to any merge that would cause a runtime error.

https://artsyproduct.atlassian.net/browse/GRO-261

/cc @artsy/grow-devs


jonallured:
Update: I _think_ updating the shared ENV is as easy as:

```
$ ./scripts/setup-env-for-artsy
# edit the .env.shared file to include new keys
$ aws s3 cp .env.shared s3://artsy-citadel/dev/.env.eigen
```

So that's what I did - can you confirm or deny that @pvinis or maybe @MounirDhahri?

pvinis:
Yep, and we also add it on 1pass, if you could edit it there as well, that would be nice 🙏.

pvinis:
huh, this worked? I would expect we would need to touch the two segment patches. cool if we can skip that for now, until we move ios analytics to ts.

jonallured:
Ok I have updated 1Password with the latest version of that file, thanks for the confirmation @pvinis!! 🙌 

I moved this off draft state and I'm thinking it's ready. One thing I'm not clear about is the CI failure here:

https://app.circleci.com/pipelines/github/artsy/eigen/13605/workflows/296ec1a4-5edb-493f-be8e-b9a655405dc5/jobs/38281

Going to cross my fingers that it's unrelated. 🤷

brainbicycle (approved):

PR #5034, opened 2021-06-24
feat: saved search notification

dimatretyak:

This PR resolves [FX-2979]

### Description
Initial development start: #5000
[A small discussion in Slack](https://artsy.slack.com/archives/C9SATFLUU/p1624526284260900?thread_ts=1624440986.218100&cid=C9SATFLUU)

#### Algorithm
- Wrap the artist’s screen in a `QueryRender`, which will receive a list of criteria by ID
- If there is no search criteria id (the user just opened the screen, and did not click on the notification), then QueryRender will not be rendered
- After executing the request, we will get saved search criteria
- We can immediately send this search criteria to `artistArtworks` request (we will get the filtered artworks in one request)

##### With this implementation, we will get the advantages
- We will get a list of filtered artworks in one request
- There will be less need to do [such checks](https://github.com/artsy/eigen/blob/21164d759402cf8b5b42544aa235a8034e4d68a3/src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx#L125-L126) in the code

#### Demo

https://user-images.githubusercontent.com/3513494/123406163-49dcc700-d5b3-11eb-89ff-2e6dc6e9bc16.mp4

- User tapping saved search push notification sees filtered artworks and applied filters - dzmitry tratsiak

[FX-2979]: https://artsyproduct.atlassian.net/browse/FX-2979

### DIFF
--- a/src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
+++ b/src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
@@ -9,6 +9,7 @@ import {
 } from "lib/Components/ArtworkFilter/ArtworkFilterHelpers"
 import { ArtworkFiltersStoreProvider, ArtworksFiltersStore } from "lib/Components/ArtworkFilter/ArtworkFilterStore"
 import { convertSavedSearchCriteriaToFilterParams } from "lib/Components/ArtworkFilter/SavedSearch/convertersToFilterParams"
+import { SearchCriteriaAttributes } from 'lib/Components/ArtworkFilter/SavedSearch/types'
 import { FilteredArtworkGridZeroState } from "lib/Components/ArtworkGrids/FilteredArtworkGridZeroState"
 import {
   InfiniteScrollArtworksGridContainer as InfiniteScrollArtworksGrid,
@@ -28,7 +29,7 @@ import { SavedSearchBannerQueryRender } from "./SavedSearchBanner"
 
 interface ArtworksGridProps extends InfiniteScrollGridProps {
   artist: ArtistArtworks_artist
-  searchCriteriaAttributes?: {}
+  searchCriteria: SearchCriteriaAttributes | null
   relay: RelayPaginationProp
 }
 
@@ -87,7 +88,7 @@ interface ArtistArtworksContainerProps {
 const ArtistArtworksContainer: React.FC<ArtworksGridProps & ArtistArtworksContainerProps> = ({
   artist,
   relay,
-  searchCriteriaAttributes,
+  searchCriteria,
   openFilterModal,
   ...props
 }) => {
@@ -96,7 +97,6 @@ const ArtistArtworksContainer: React.FC<ArtworksGridProps & ArtistArtworksContai
   const appliedFilters = ArtworksFiltersStore.useStoreState((state) => state.appliedFilters)
 
   const setInitialFilterStateAction = ArtworksFiltersStore.useStoreActions((state) => state.setInitialFilterStateAction)
-  const applyFiltersAction = ArtworksFiltersStore.useStoreActions((action) => action.applyFiltersAction)
 
   const applyFilters = ArtworksFiltersStore.useStoreState((state) => state.applyFilters)
   const relevantFiltersForSavedSearch = appliedFilters.filter((filter) => !(filter.paramName === FilterParamName.sort))
@@ -127,13 +127,12 @@ const ArtistArtworksContainer: React.FC<ArtworksGridProps & ArtistArtworksContai
   useEffect(() => {
     setAggregationsAction(artworks?.aggregations)
 
-    if (searchCriteriaAttributes && artworks?.aggregations) {
+    if (searchCriteria && artworks?.aggregations) {
       const params = convertSavedSearchCriteriaToFilterParams(
-        searchCriteriaAttributes,
+        searchCriteria,
         artworks.aggregations as Aggregations
       )
       setInitialFilterStateAction(params)
-      applyFiltersAction()
     }
   }, [])
 
--- a/src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
+++ b/src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
@@ -2,11 +2,13 @@ import { OwnerType } from "@artsy/cohesion"
 import { ArtistArtworks_artist } from "__generated__/ArtistArtworks_artist.graphql"
 import { ArtworkFilterNavigator, FilterModalMode } from "lib/Components/ArtworkFilter"
 import {
+  Aggregations,
   filterArtworksParams,
   FilterParamName,
   prepareFilterArtworksParamsForInput,
 } from "lib/Components/ArtworkFilter/ArtworkFilterHelpers"
 import { ArtworkFiltersStoreProvider, ArtworksFiltersStore } from "lib/Components/ArtworkFilter/ArtworkFilterStore"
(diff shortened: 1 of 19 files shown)


src/lib/Scenes/Artist/SearchCriteria.tsx
```diff
@@ -0,0 +1,83 @@
...
+  render: {
+    renderPlaceholder: () => React.ReactElement,
+    renderComponent: (args: {
+      fetchCriteriaError: Error | null,
+      savedSearchCriteria: SearchCriteriaAttributes | null
+    }) => React.ReactElement
+  }
+}
+
+export const SearchCriteriaQueryRender: React.FC<SearchCriteriaQueryRenderProps> = (props) => {
```
mdole, line 23: **Minor:** suggest renaming `SearchCriteriaQueryRender` -> `SearchCriteriaQueryRenderer` (just the extra `er` at the very end) to match our normal naming conventions

mdole (changes requested):
Awesome work @dimatretyak! There are a couple small adjustments I think we should make. And because https://github.com/artsy/eigen/pull/5032 is now merged in, this PR should be rebased and updated (we changed the format of the props object we were passing from notifications into React Native).

Let me know if you want to pair or have questions!

src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
```diff
@@ -120,6 +126,14 @@ const ArtistArtworksContainer: React.FC<ArtworksGridProps & ArtistArtworksContai
 
   useEffect(() => {
     setAggregationsAction(artworks?.aggregations)
+
+    if (searchCriteria && artworks?.aggregations) {
+      const params = convertSavedSearchCriteriaToFilterParams(
+        searchCriteria,
+        artworks.aggregations as Aggregations
+      )
+      setInitialFilterStateAction(params)
```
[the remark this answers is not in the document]
dimatretyak: Yep, you're right. Now we pass the search criteria immediately at [the first request](https://github.com/artsy/eigen/blob/8d1f80a0d6808439a33a0ef33dc018728f94b58c/src/lib/Scenes/Artist/Artist.tsx#L143-L148)

src/lib/Scenes/Artist/Artist.tsx
```diff
@@ -10,31 +10,56 @@ import { ArtistAboutContainer } from "lib/Components/Artist/ArtistAbout/ArtistAb
...
   const displayAboutSection =
     artistAboveTheFold.has_metadata ||
     (artistAboveTheFold.counts?.articles ?? 0) > 0 ||
     (artistAboveTheFold.counts?.related_artists ?? 0) > 0
 
+  useEffect(() => {
+    if (!!fetchCriteriaError) {
+      popoverMessage.show({
+        title: "Sorry, an error occured",
+        message: "Failed to get saved search criteria",
```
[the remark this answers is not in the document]
dimatretyak: We [discussed](https://artsy.slack.com/archives/C9SATFLUU/p1624542547280100?thread_ts=1624440986.218100&cid=C9SATFLUU) this with @dblandin yesterday

src/lib/Scenes/Artist/Artist.tsx
```diff
@@ -101,53 +126,83 @@ export const Artist: React.FC<{
...
-              ...ArtistAbout_artist
-              ...ArtistInsights_artist
-            }
-          }
-        `,
-        variables: { artistID },
-      }}
+    <SearchCriteriaQueryRender
+      searchCriteriaId={notificationPayload?.searchCriteriaID}
+      environment={environment}
```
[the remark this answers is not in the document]
dimatretyak: I don't think there will be an impact on performance since we will never have a query/render query render if the [following conditions](https://github.com/artsy/eigen/blob/8d1f80a0d6808439a33a0ef33dc018728f94b58c/src/lib/Scenes/Artist/SearchCriteria.tsx#L28) are not met

src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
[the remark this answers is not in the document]
mdole: ah right! thanks for clarifying

src/lib/Scenes/Artist/Artist.tsx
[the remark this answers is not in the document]
mdole: Ah yes, good point. Makes sense to me!

mdole (approved):

src/lib/Scenes/Artist/Artist.tsx
```diff
@@ -109,61 +129,78 @@ interface ArtistQueryRendererProps extends ArtistAboveTheFoldQueryVariables, Art
...
-              ...ArtistAbout_artist
-              ...ArtistInsights_artist
-            }
-          }
-        `,
-        variables: { artistID },
-      }}
+    <SearchCriteriaQueryRenderer
+      searchCriteriaId={searchCriteriaID}
+      environment={environment}
```
MounirDhahri, line 138: **issue:** I am not sure if it's a good idea to be nesting QueryRenderers here. We did some optimisations recently to the Artist screen to improve its TTI and the changes here break those optimizations since we will only be rendering the artist screen after we get the search criteria related fields. I also believe that if any updates to the search criteria, will lead to 2 additional unnecessary queries at the `AboveTheFoldQueryRenderer` in line 151.

Did you maybe considered having the `SearchCriteriaQuery` as part of the below the fold Query and if this is not possible to have the nested QueryRenderer (which we should avoid) only at the Artworks tab

dimatretyak: @MounirDhahri  I'll see what I can do. But, unfortunately, this is the best solution to the problem that I managed to come up with. With this solution, we can immediately get filtered works according to the saved criteria in one request. Otherwise, we would first have to make [a request without any criteria](https://github.com/artsy/eigen/blob/f07a71eae224cda84298c56480a27cd2396a50f2/src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx#L226-L239), after we receive the saved search criteria, we need to send another request for receiving artworks, but with the specified filters (there is an example of [such an implementation](https://github.com/artsy/eigen/pull/5000/files), but it complicates the code)

The main idea of `SearchCriteriaQueryRenderer` is that it will execute only when the `searchCriteriaId` was passed (in our case, this is when the user clicked on the PUSH notification about the saved search) and it will not change (respectively, the two queries below will also not be re-executed)

MounirDhahri: Thanks for the explanation @dimatretyak. I have a better understanding now of what's happening. I think it's fine this way then and you or @mdole please feel free to merge this.
I do however want to suggest a possible solution for the future. Maybe instead of getting the search criteria on Eigen then querying for artworks using those criteria, we can do that query on metaphysics and even better probably on gravity.
Using sequence diagrams, what I mean is, instead of the following (may not be too accurate and it's definitely very simplified):
diagram
I would suggest that we try to do something like this:
Untitled (3)
_What's in the red rectangle is not necessary for now because we can also just move stuff to metaphysics._

dimatretyak: @MounirDhahri Thank you for the diagrams and detailed discussion! Yes, this option was also discussed with the team. I think that in the future this is exactly the solution that will be added

mdole: +1 thanks for the thoughts and diagrams! we just chatted as a team and we agreed to move forward with this version and make the change to filter artworks by search criteria ID as you suggested in the future.

MounirDhahri: sure 🙇‍♂️

PR #5039, opened 2021-06-25
fix (CX-1538): improve welcome bg img quality

kalshannikava:

This PR resolves [CX-1538]

### Description

The welcome image in the background seems to be blurry and this might be because we are using a small size image.

Screenshot 2021-06-25 at 11 26 23

So the goal was to replace the images with higher resolution ones.  

- [] I have tested my changes on **iOS** and **Android**.

-

#### iOS user-facing changes

-

-

#### Dev changes

-

[CX-1538]: https://artsyproduct.atlassian.net/browse/CX-1538

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -44,6 +44,7 @@ upcoming:
     - Show price with currency prefix in auction result list - ole
     - Add auction house filter for auction results - ole
     - Remove "Enable push notification" guard for disabling saved search - dzmitry tratsiak
+    - Improve background image quality - katsiaryna alshannikava
 
 releases:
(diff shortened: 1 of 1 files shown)


kalshannikava:
@MounirDhahri I've replaced the images but not sure that managed to achieve the desired result

MounirDhahri (approved):

MounirDhahri (changes requested):
This gets the desired effect, I only have a concern about the @2x image size (4mb), that's roughly 20% of our app size now!. I don't think we should include it in the project at this size. Can you check with Jonathan if any smaller size versions are available

PR #5044, opened 2021-06-25
fix(CX-1650): tapping artist opens its page after passing new onboarding on ios

edez1t:

This PR resolves [CX-1650]

### Description

It'll be lot of small ui bugs after adding new onboarding, here's one of them fixed

Before:

https://user-images.githubusercontent.com/73339470/123431304-c8e0f800-d5d1-11eb-9587-d38858128263.mov

After:

https://user-images.githubusercontent.com/73339470/123430124-7e12b080-d5d0-11eb-9775-d7906e2fcc87.mov

-

#### iOS user-facing changes

-

-

#### Dev changes

-

[CX-1650]: https://artsyproduct.atlassian.net/browse/CX-1650


src/lib/Scenes/Onboarding/OnboardingPersonalization/OnboardingPersonalizationArtistListItem.tsx
```diff
@@ -126,7 +127,7 @@ export class OnboardingPersonalizationArtistListItem extends React.Component<Pro
     return (
       <TouchableComponent
         onPress={() => {
-          if (href) {
+          if (href && !disableNavigation) {
             this.handleTap(href)
```
pvinis, line 131: Sorry, I'm way late here, but is this component using `handleTap`? Do we even need it? I think during onboarding, there is no tap navigation. Maybe we should not add the prop here and just remove the tap functionality?

edez1t: Yeah, you're right, those two artistlistitem components look almost exactly the same, if we look closer navigation isn't the only useless item here

pvinis: Hmmm ok. I guess we will get to cleaning that up in the future 🤷.

PR #5045, opened 2021-06-25
feat: add diagnose script

MounirDhahri:

This PR adds a new command `diagnose` to our node scripts. The goal from this command is to check your environment file and packages if they're not up to date.

This is the first iteration of this script and more follow-up work will follow. Some of the checks that I have in mind include but are not limited to:
- Checking cocopods
- Check native android code
- Checking Dev Environment (node version, xcode version...)
- Check if the node port is used
- Suggest resolutions on the flip table script

### Description

https://user-images.githubusercontent.com/11945712/123451788-1f543380-d5de-11eb-89ab-1dcfd693a671.mov


scripts/diagnose.js
```diff
@@ -0,0 +1,49 @@
+const { spawnSync } = require("child_process")
```
olerichter00, line 1: Can we add a small comment here to explain what the script does?

olerichter00 (approved):
Awesome!

scripts/diagnose.js
```diff
@@ -0,0 +1,49 @@
...
+  const task = spawnSync(command, { shell: true, cwd })
+  if (task.status != 0) {
+    throw new Error(task.stderr.toString())
+  }
+  return task.stdout.toString()
+}
+
+const checkEnvVariables = () => {
+  exec("touch .env.temp")
+  exec("aws s3 cp s3://artsy-citadel/dev/.env.eigen .env.temp")
```
damassi, line 16: cc @olerichter00 - one more example

ovasdi: @damassi thanks for the cc on this. Shall we adopt this to force as well?

This is awesome. @MounirDhahri

damassi: Not sure about the full script, but checking to see if .env vars are in sync is pretty sweet, since thats about 100% of the problems we (consistently) face when booting force at the moment ([eg this, from yesterday](https://artsy.slack.com/archives/CP9P4KR35/p1624550328273200)). 

Not sure what the final force flow would be, but yeah we can kick ideas around 👍

ovasdi: Great. I noted this [in the force turn-key ticket](https://artsyproduct.atlassian.net/browse/PLATFORM-3135) which i'll likely be picking up when done with rotations.

brainbicycle (approved):
🩺 Nice one thanks for adding! Looks good to me, if you want to add the comment Ole suggested good to merge!

scripts/diagnose.js
MounirDhahri: I will 🙌

MounirDhahri:
#squashongreen

PR #5048, opened 2021-06-25
chore: update convert to filter params tests

dimatretyak:
Update convert to filter params tests


dimatretyak:
#squashongreen

PR #5054, opened 2021-06-28
refactor: category and refine logic

dimatretyak:
The type of this PR is: **Refactor**

This PR resolves [FX-3044]

### Description
* Remove refine native module changes (was reviewed) #5061
* The “Refine” button on the category screen (i.e. Black and White) in Eigen does not work for Android users. It is expected to make a filter modal appear, but it currently doesn’t because that modal is implemented natively for iOS. The purpose of this ticket is to refactor the filter modal so that it can be opened by Android users and enable the “Refine” button. This can be done by rebuilding the existing native filter modal in React Native or by reusing the same filter modal components used by other artwork grids (such as the artist artwork grid).

#### iOS

https://user-images.githubusercontent.com/3513494/124098559-4e96f480-da65-11eb-8eea-d8f8e434f263.mp4

#### Android

https://user-images.githubusercontent.com/3513494/124098613-5a82b680-da65-11eb-914a-30d927ab367c.mp4

- Refactor category screen header and filters to match other artwork grids - dzmitry tratsiak

#### iOS user-facing changes

-

-

#### Dev changes

-

[FX-3044]: https://artsyproduct.atlassian.net/browse/FX-3044


src/lib/Components/Gene/GeneArtworks.tsx
```diff
@@ -0,0 +1,224 @@
...
+interface GeneArtworksContainerProps {
+  gene: GeneArtworks_gene
+  relay: RelayPaginationProp
+}
+
+interface GeneArtworsProps extends GeneArtworksContainerProps {
+  openFilterModal: () => void
+}
+
+export const GeneArtwors: React.FC<GeneArtworsProps> = (props) => {
```
brainbicycle, line 31:
```suggestion
export const GeneArtworks: React.FC<GeneArtworsProps> = (props) => {
```
small typo here!

src/lib/Components/Gene/GeneArtworks.tsx
```diff
@@ -0,0 +1,224 @@
...
+import { useEffect } from "react"
+import { createPaginationContainer, graphql, RelayPaginationProp } from "react-relay"
+import { useTracking } from "react-tracking"
+
+interface GeneArtworksContainerProps {
+  gene: GeneArtworks_gene
+  relay: RelayPaginationProp
+}
+
+interface GeneArtworsProps extends GeneArtworksContainerProps {
```
brainbicycle, line 27:
```suggestion
interface GeneArtworksProps extends GeneArtworksContainerProps {
```
small typo here!

src/lib/Components/Gene/GeneArtworks.tsx
dimatretyak: Solved 👍

src/lib/Components/Gene/GeneArtworks.tsx
dimatretyak: Solved 👍

src/lib/Components/Gene/GeneArtworks.tsx
```diff
@@ -0,0 +1,222 @@
...
+          hasMore={props.relay.hasMore}
+          loadMore={props.relay.loadMore}
+        />
+      </Box>
+    )
+  }
+
+  return (
+    <Box mt={1}>
+      <Message>There aren’t any works available by the category at this time.</Message>
```
brainbicycle, line 104: Not sure about copy here, can you ask @thomasmeerschwam if this is correct?

brainbicycle (changes requested):
Fantastic work! Code looks great and thank you for removing the old Objective-C code, I have one question about empty state copy I think we should address but otherwise looks great to me 🙇

src/lib/Components/Gene/GeneArtworks.tsx
```diff
@@ -0,0 +1,222 @@
...
+        PAGE_SIZE,
+        (error) => {
+          if (error) {
+            throw new Error("Gene/GeneArtworks filter error: " + error.message)
+          }
+        },
+        { input: prepareFilterArtworksParamsForInput(filterParams) }
+      )
+    }
+  }, [appliedFilters])
```
MounirDhahri, line 65: **thought:** shouldn't this be `applyFilters` ?

src/lib/Components/Gene/GeneArtworks.tsx
```diff
@@ -0,0 +1,222 @@
...
+    )
+  }, [artworksTotal, openFilterModal])
+
+  if (gene.artworks) {
+    if (artworksTotal === 0) {
+      return (
+        <Box pt={1}>
+          <FilteredArtworkGridZeroState id={gene.id} slug={gene.slug} trackClear={trackClear} />
+        </Box>
+      )
```
MounirDhahri, line 88: **issue(blocker):** I think in the case we have no artworks available for a gene (with no filters enabled) this would show `FilteredArtworkGridZeroState` instead of `no artworks available for this gene`

src/lib/Components/Gene/GeneArtworks.tsx
```diff
@@ -0,0 +1,222 @@
...
+    tracking.trackEvent({
+      action_name: "closeFilterWindow",
+      context_screen_owner_type: Schema.OwnerEntityTypes.Gene,
+      context_screen: Schema.PageNames.GenePage,
+      context_screen_owner_id: gene.id,
+      context_screen_owner_slug: gene.slug,
+      action_type: Schema.ActionTypes.Tap,
+    })
+    handleCloseFilterArtworksModal()
+  }
```
MounirDhahri, line 139: **issue:** same comment here about analyics as above

src/lib/Components/Gene/GeneArtworks.tsx
MounirDhahri: I just saw the test you made, so the assumption you have I guess is that `gene.artworks` is null when no artworks are available but I think we actually return an empty array and not null from metaphysics.

src/lib/Components/Gene/GeneArtworks.tsx
dimatretyak: @MounirDhahri Can you please clarify?

src/lib/Components/Gene/GeneArtworks.tsx
dimatretyak: Yes, you're right. I don't think there is a possible case where `gene` is `null`

src/lib/Components/Gene/GeneArtworks.tsx
```diff
@@ -0,0 +1,222 @@
...
+  const trackClear = () => {
+    tracking.trackEvent({
+      action_name: "clearFilters",
+      context_screen: Schema.ContextModules.ArtworkGrid,
+      context_screen_owner_type: Schema.OwnerEntityTypes.Gene,
+      context_screen_owner_id: gene.id,
+      context_screen_owner_slug: gene.slug,
+      action_type: Schema.ActionTypes.Tap,
+    })
+  }
```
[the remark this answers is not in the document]
dimatretyak: If I understood you correctly, then this is solved

src/lib/Components/Gene/GeneArtworks.tsx
```diff
@@ -72,29 +72,21 @@ export const GeneArtworks: React.FC<GeneArtworksProps> = (props) => {
...
-  if (gene.artworks) {
-    if (artworksTotal === 0) {
-      return (
-        <Box pt={1}>
-          <FilteredArtworkGridZeroState id={gene.id} slug={gene.slug} trackClear={trackClear} />
-        </Box>
-      )
-    }
-
+  if (artworksTotal === 0) {
```
MounirDhahri, line 75: **issue:** unfortunately here the artworks total will be changing whenever the filters change and it's hard to tell later if there are no artworks at all or just with the applied filter. I think you can solve that by saving the initial value as a reference or in the state (I prefer as a reference since it won't be changing)
We did something like that here (but probably with a reference) https://github.com/artsy/eigen/blob/7d6e9ca3eecce96eaf4534b07aa53bf56cdf8d78/src/lib/Scenes/Sale/Components/SaleLotsList.tsx#L85-L88

dimatretyak: @MounirDhahri done

src/lib/Components/Gene/GeneArtworks.tsx
MounirDhahri: Usually we should be applying the filters only when the applyFitlers boolean is set to true but I think it's fine to have it this way as well

MounirDhahri (approved):
🙌

src/lib/Components/Gene/GeneArtworks.tsx
dimatretyak: In our case, filters will also be applied when `applyFilters` is `true`. I looked at how it is implemented in other places and did it by analogy ([ArtistArtworks](https://github.com/artsy/eigen/blob/master/src/lib/Components/Artist/ArtistArtworks/ArtistArtworks.tsx#L113-L125), [ArtistInsightsAuctionResults](https://github.com/artsy/eigen/blob/master/src/lib/Components/Artist/ArtistInsights/ArtistInsightsAuctionResults.tsx#L41-L54), [ArtistSeriesArtworks](https://github.com/artsy/eigen/blob/master/src/lib/Scenes/ArtistSeries/ArtistSeriesArtworks.tsx#L47-L59), etc)

dimatretyak:
@brainbicycle The text of the empty message is agreed with Thomas

brainbicycle (approved):

PR #5057, opened 2021-06-29
fix: use palette touchable /w underlayColor & consistency style changes in ArtistSeriesListItem

gkartalis:

This PR resolves [FX-2254]

### Description

ArtistSeriesListItem was using `TouchableOpacity` from `react-native` and not `Touchable` from palette.

Furthermore the `underlayColor` wasn't the same as other parts of the app so added "black5" to match the other Touchables.

The screenshots below show the touchable behaviour when the user is pressing the button.

Also made the `<ArtistSeriesListItem />` to take 100% width of the list and adjusted margins accordingly for consistensy as suggested from design [here](https://artsy.slack.com/archives/C9SATFLUU/p1624960749370100)

|Android|iOS|
|---|---|
|android|ios|

#### More screenshots with all affected screens

|ArtistSeries|ArtistAbout|
|---|---|
|Simulator Screen Shot - iPhone 8 - 2021-06-30 at 09 44 13|Simulator Screen Shot - iPhone 8 - 2021-06-30 at 09 45 47|

|ArtistSeriesMoreSeries|Artwork|
|---|---|
|Simulator Screen Shot - iPhone 8 - 2021-06-30 at 09 37 17|Simulator Screen Shot - iPhone 8 - 2021-06-30 at 09 51 29|

- Fix add touchable underlay color in ArtistSeriesListItem

[FX-2254]: https://artsyproduct.atlassian.net/browse/FX-2254

### DIFF
--- a/src/lib/Scenes/ArtistSeries/ArtistSeriesListItem.tsx
+++ b/src/lib/Scenes/ArtistSeries/ArtistSeriesListItem.tsx
@@ -52,12 +52,13 @@ export const ArtistSeriesListItem: React.FC<ArtistSeriesListItemProps> = ({
   return (
     <Touchable
       underlayColor={color("black5")}
+      style={{ marginHorizontal: -20 }}
       onPress={() => {
         trackArtworkClick()
         navigate(`/artist-series/${listItem?.node?.slug}`)
       }}
     >
-      <Flex flexDirection="row" mb={1} justifyContent="space-between">
+      <Flex px={2} my={1} flexDirection="row" justifyContent="space-between">
         <Flex flexDirection="row" justifyContent="space-between" width="100%">
           <Flex flexDirection="row">
             <OpaqueImageView
--- a/src/lib/Scenes/ArtistSeries/ArtistSeriesListItem.tsx
+++ b/src/lib/Scenes/ArtistSeries/ArtistSeriesListItem.tsx
@@ -52,6 +52,8 @@ export const ArtistSeriesListItem: React.FC<ArtistSeriesListItemProps> = ({
   return (
     <Touchable
       underlayColor={color("black5")}
+      // the negative margin here is for resetting padding of 20 that all the parent components of this instance
+      // have and to avoid changing the component tree in multiple spots.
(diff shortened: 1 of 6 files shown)


mdole (changes requested):
Nice! Did notice one thing from the screenshots: it looks like the bottom padding between each item also has the opacity on tap. @gkartalis would you be up for tweaking it so that only the item and not the padding has the background color on tap?
123755168-f07add80-d8bb-11eb-9d73-f881bc9aca61

gkartalis:
Good observation. Sure thing!

I also noticed another inconsistency that was discussed on [this slack thread](https://artsy.slack.com/archives/C9SATFLUU/p1624960749370100) with Johnathan as well and will address everything from the reply in this pr.

src/lib/Scenes/ArtistSeries/ArtistSeriesListItem.tsx
```diff
@@ -51,13 +50,15 @@ export const ArtistSeriesListItem: React.FC<ArtistSeriesListItemProps> = ({
   }
 
   return (
-    <TouchableOpacity
+    <Touchable
+      underlayColor={color("black5")}
+      style={{ marginHorizontal: -20 }}
```
gkartalis, line 55: Not sure this is the best approach here. The reason I used this negative margin is because in all instances of `<ArtistSeriesListItem />` and `<ArtistSeriesMoreSeries />` we have a parent component padding of 20 and thought that it would be better to change the margin in one spot than in multiple and also mess with the component tree.

Any thoughts or suggestions to change that are welcome ofc!

mdole: Yeah, this is a tricky decision! I think your choice here is reasonable. One small thing you could do would be to add a comment noting the purpose of the `marginHorizontal`; while it's not a particularly complex piece of code, sometimes it's helpful to have the intent clearly spelled out.

mdole (approved):
Nice work! Let's ship it

gkartalis:
#squashongreen

PR #5059, opened 2021-06-29
fix white line below in new onboarding screen

JenqaaaBY:

This PR resolves [CX-1651]

### Description

NOTE: Can be QAed after merging this PR https://github.com/artsy/eigen/pull/5015

How it looks now on iPhone 11:
Screen Shot 2021-06-29 at 2 52 49 PM

https://user-images.githubusercontent.com/30176348/123793271-32227d00-d8ea-11eb-8f9f-5c4fb9608352.mov

#### iOS user-facing changes

- Now on devices with safeArea there will be no white line below in the new onboarding screen
- For other devices, the change will not be visible

- For android devices, the change will not be visible

[CX-1651]: https://artsyproduct.atlassian.net/browse/CX-1651


src/lib/Scenes/Onboarding/OnboardingPersonalization/OnboardingPersonalization.tsx
```diff
@@ -147,6 +148,7 @@ export const OnboardingPersonalizationList: React.FC<OnboardingPersonalizationLi
           variant="primaryBlack"
           block
           testID="doneButton"
+          mb={useScreenDimensions().safeAreaInsets.bottom}
```
pvinis, line 151: correct me if im wrong, but i think if that component goes in an `if` or becomes in any way conditional, we would break the hooks rules.

it might be better to do `const { safeAreaInsets } = useScreenDimensions()` and `mb={safeAreaInsets.bottom}`, if the above is correct.

pvinis (approved):

edez1t (approved):

PR #5063, opened 2021-06-30
feat: Prevent users from submit consignments for non target supply artists

olerichter00:

This PR resolves [CX-630]

### Description

- Removes artists that are not marked as target supply from the search autosuggest.
- Increases the number of results for the search query (because we have to filter out a lot of the results).
- Does not add tracking - I'm waiting for some input and will implement it in another PR to get this change out as soon as possible.

| Target Supply | No Target Supply |
|---|---|
| Simulator Screen Shot - iPhone 8 - 2021-06-30 at 13 51 22 | Simulator Screen Shot - iPhone 8 - 2021-06-30 at 13 51 16 |

### QA

- Click on "Submit a work"
- Click on "Artist/Designer"
- Search for "Gong" -> shouldn't display artists
- Search for "Banksy" -> should only display Banksy

-

#### iOS user-facing changes

-

-

#### Dev changes

-

[CX-630]: https://artsyproduct.atlassian.net/browse/CX-630

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -19,6 +19,7 @@ upcoming:
     - Fix opening artist page after passing new onboarding on ios - dzmitry
     - Fix crash when opening app from background on android - brian
     - Change error handling middleware & include optionalField graphql directive to home screen - ole
+    - Permit users to submit consignments for non target supply artists - ole
 
 releases:
   - version: 6.9.5
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -19,7 +19,7 @@ upcoming:
     - Fix opening artist page after passing new onboarding on ios - dzmitry
     - Fix crash when opening app from background on android - brian
     - Change error handling middleware & include optionalField graphql directive to home screen - ole
-    - Permit users to submit consignments for non target supply artists - ole
+    - Prevent users from submitting consignments for non target supply artists - ole
 
 releases:
   - version: 6.9.5
--- a/src/lib/Scenes/Consignments/Screens/ConsignmentsArtist.tsx
+++ b/src/lib/Scenes/Consignments/Screens/ConsignmentsArtist.tsx
@@ -53,7 +53,7 @@ export default class Artist extends React.Component<Props, State> {
       environment,
       graphql`
         query ConsignmentsArtistQuery($query: String!) {
-          searchConnection(query: $query, first: 10, entities: [ARTIST], mode: AUTOSUGGEST) {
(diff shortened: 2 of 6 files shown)


src/lib/Scenes/Consignments/Screens/ConsignmentsArtist.tsx
```diff
@@ -76,6 +78,11 @@ export default class Artist extends React.Component<Props, State> {
     this.setState({ results, searching: false })
   }, 1000)
 
+  filteredResults = () => {
+    // filter for artists that are marked as target supply
+    return this.state.results?.filter((artist) => artist.targetSupply?.isTargetSupply) || null
+  }
```
olerichter00, line 84: **Note:** Here we filter for the results.

src/lib/Scenes/Consignments/Screens/ConsignmentsArtist.tsx
```diff
@@ -54,7 +53,7 @@ export default class Artist extends React.Component<Props, State> {
       environment,
       graphql`
         query ConsignmentsArtistQuery($query: String!) {
-          searchConnection(query: $query, first: 10, entities: [ARTIST], mode: AUTOSUGGEST) {
+          searchConnection(query: $query, first: 30, entities: [ARTIST], mode: AUTOSUGGEST) {
```
olerichter00, line 56: **Note**: I increased the number to 30 to get more results because we have to filter out all artists that are not marked as target supply [slack thread](https://artsy.slack.com/archives/CP9P4KR35/p1625054724376500).

CHANGELOG.yml
```diff
@@ -19,6 +19,7 @@ upcoming:
     - Fix opening artist page after passing new onboarding on ios - dzmitry
     - Fix crash when opening app from background on android - brian
     - Change error handling middleware & include optionalField graphql directive to home screen - ole
+    - Permit users to submit consignments for non target supply artists - ole
```
brainbicycle, line 22:
```suggestion
    - Prevent users from submitting consignments for non target supply artists - ole
```
This is actually the opposite right?

brainbicycle (commented):
Looks good! One comment about the changelog and the title, it looks like this is actually preventing users from submitting with non-target supply which I believe is intentional but if that is case may want to update the title + changelog message to reflect. Thanks for cleaning up the typescript comments!

CHANGELOG.yml
olerichter00: Thanks 💯

PR #5064, opened 2021-06-30
fix (CX-958): refresh artwork after purchase

kalshannikava:

This PR resolves [CX-958]

### Description

**The issue:**
As a collector, when I buy an artwork, the artwork screen should automatically refresh to show I've purchased it so that I'm sure I bought it.
Without refresh "Buy now" button is still available and we don't display "Sold" status.

**The fix:**

https://user-images.githubusercontent.com/27921300/123959866-f00f3f00-d9ae-11eb-8527-6c626b8d80c1.mov

-

#### iOS user-facing changes

-

-

#### Dev changes

-

[CX-958]: https://artsyproduct.atlassian.net/browse/CX-958

### DIFF
--- a/src/lib/Components/ArtsyReactWebView.tsx
+++ b/src/lib/Components/ArtsyReactWebView.tsx
@@ -1,7 +1,7 @@
 import { OwnerType } from "@artsy/cohesion"
 import { color } from "@artsy/palette-tokens"
 import { addBreadcrumb } from "@sentry/react-native"
-import { goBack, navigate } from "lib/navigation/navigate"
+import { dismissModal, goBack, navigate } from "lib/navigation/navigate"
 import { matchRoute } from "lib/navigation/routes"
 import { getCurrentEmissionState, GlobalStore, useEnvironment } from "lib/store/GlobalStore"
 import { Schema } from "lib/utils/track"
@@ -79,7 +79,9 @@ export const ArtsyReactWebViewPage: React.FC<
         <FancyModalHeader
           useXButton={isPresentedModally && !canGoBack}
           onLeftButtonPress={() => {
-            if (!canGoBack) {
+            if (isPresentedModally && !canGoBack) {
+              dismissModal()
+            } else if (!canGoBack) {
               goBack()
             } else {
               ref.current?.goBack()
--- a/src/lib/Scenes/Artwork/Artwork.tsx
+++ b/src/lib/Scenes/Artwork/Artwork.tsx
@@ -6,6 +6,7 @@ import { ArtworkAboveTheFoldQuery } from "__generated__/ArtworkAboveTheFoldQuery
 import { ArtworkBelowTheFoldQuery } from "__generated__/ArtworkBelowTheFoldQuery.graphql"
 import { ArtworkMarkAsRecentlyViewedQuery } from "__generated__/ArtworkMarkAsRecentlyViewedQuery.graphql"
 import { RetryErrorBoundary } from "lib/Components/RetryErrorBoundary"
+import { navigationEvents } from "lib/navigation/navigate"
 import { defaultEnvironment } from "lib/relay/createEnvironment"
 import { ArtistSeriesMoreSeriesFragmentContainer as ArtistSeriesMoreSeries } from "lib/Scenes/ArtistSeries/ArtistSeriesMoreSeries"
 import { unsafe_getFeatureFlag } from "lib/store/GlobalStore"
@@ -98,6 +99,11 @@ export class Artwork extends React.Component<Props, State> {
 
   componentDidMount() {
     this.markArtworkAsRecentlyViewed()
+    navigationEvents.addListener("modalDismissed", this.handleModalDismissed)
+  }
+
+  componentWillUnmount() {
+    navigationEvents.removeListener("modalDismissed", this.handleModalDismissed)
   }
 
   // @ts-expect-error STRICTNESS_MIGRATION --- 🚨 Unsafe legacy code 🚨 Please delete this and fix any type errors if you have time 🙏
@@ -170,6 +176,11 @@ export class Artwork extends React.Component<Props, State> {
         force: true,
       }
     )
+  }
+
+  handleModalDismissed = () => {
+    this.onRefresh()
+    return true
   }
 
   markArtworkAsRecentlyViewed = () => {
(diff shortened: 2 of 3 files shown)


src/lib/Scenes/Artwork/Artwork.tsx
```diff
@@ -299,18 +310,22 @@ export class Artwork extends React.Component<Props, State> {
...
-            <Box mx={2} my={3}>
-              <Separator />
-            </Box>
-          )}
-          refreshControl={<RefreshControl refreshing={this.state.refreshing} onRefresh={this.onRefresh} />}
-          contentContainerStyle={{ paddingBottom: 40 }}
-          renderItem={({ item }) => (item.excludePadding ? item.element : <Box px={2}>{item.element}</Box>)}
-        />
+        {this.state.refreshing ? (
+          <ActivityIndicator />
```
kalshannikava, line 314: I'm not sure if ActivityIndicator is the correct component to use here but by a quick look at the project it seems like so. It could be changed if it's needed.

src/lib/Components/ArtsyReactWebView.tsx
```diff
@@ -79,7 +79,9 @@ export const ArtsyReactWebViewPage: React.FC<
         <FancyModalHeader
           useXButton={isPresentedModally && !canGoBack}
           onLeftButtonPress={() => {
-            if (!canGoBack) {
+            if (isPresentedModally && !canGoBack) {
+              dismissModal()
```
kalshannikava, line 83: Had to add it becuse for some reason we were using 'goBack' event even for the modal close which doesn't seem to be correct.

src/lib/Scenes/Artwork/Artwork.tsx
JenqaaaBY: I think you should use a spinner from palette
`import { Spinner } from "palette"`

JenqaaaBY (approved):
Now it looks good to me

src/lib/Scenes/Artwork/Artwork.tsx
```diff
@@ -299,18 +311,24 @@ export class Artwork extends React.Component<Props, State> {
...
-          ItemSeparatorComponent={() => (
-            <Box mx={2} my={3}>
-              <Separator />
-            </Box>
-          )}
-          refreshControl={<RefreshControl refreshing={this.state.refreshing} onRefresh={this.onRefresh} />}
-          contentContainerStyle={{ paddingBottom: 40 }}
-          renderItem={({ item }) => (item.excludePadding ? item.element : <Box px={2}>{item.element}</Box>)}
-        />
+        {this.state.refreshing ? (
```
olerichter00, line 314: This change actually breaks the RefreshControl UI (scrolling down from top to refresh the page). I'd suggest to use the already existing logic we have in [componentDidUpdate](https://github.com/artsy/eigen/blob/62b1bbe8cf3d4e7e8618d14e4a8c3fce08bf1e38/src/lib/Scenes/Artwork/Artwork.tsx#L106), extract it into a method (e.g. `refresh`) and call it in `componentDidUpdate` and `handleModalDismissed`.

olerichter00 (changes requested):
The change refreshes the page but it also breaks the RefreshControl UI. I think we can accomplish the same easier by just the logic in `componentDidUpdate`.

src/lib/Scenes/Artwork/Artwork.tsx
```diff
@@ -64,6 +67,7 @@ interface State {
 export class Artwork extends React.Component<Props, State> {
   state = {
     refreshing: false,
+    fetchingData: false,
```
olerichter00, line 70: I think we don't even need to add an indicator that we reload the data when we come from the modal.

src/lib/Scenes/Artwork/Artwork.tsx
```diff
@@ -299,18 +318,24 @@ export class Artwork extends React.Component<Props, State> {
...
+            ItemSeparatorComponent={() => (
+              <Box mx={2} my={3}>
+                <Separator />
+              </Box>
+            )}
+            refreshControl={<RefreshControl refreshing={this.state.refreshing} onRefresh={this.onRefresh} />}
+            contentContainerStyle={{ paddingBottom: 40 }}
+            renderItem={({ item }) => (item.excludePadding ? item.element : <Box px={2}>{item.element}</Box>)}
+          />
+        )}
```
olerichter00, line 338: I'd just remove this change completely and don't indicate when we reload because we come back from the because the spinner we introduce in this block is breaking the UI of the <RefreshControl /> component

kalshannikava: @olerichter00 hmmm, but I'd say this switch from 'buy now' to 'sold' doesn't look good
https://user-images.githubusercontent.com/27921300/124146787-6767d000-da8e-11eb-8bc4-de4e24fb107b.mov

olerichter00: Sorry, I haven't seen that you actually changed the refresh indicator to a spinner. That's actually way better than not showing any loading indicator 👍

src/lib/Scenes/Artwork/Artwork.tsx
```diff
@@ -299,18 +318,24 @@ export class Artwork extends React.Component<Props, State> {
...
-            </Box>
-          )}
-          refreshControl={<RefreshControl refreshing={this.state.refreshing} onRefresh={this.onRefresh} />}
-          contentContainerStyle={{ paddingBottom: 40 }}
-          renderItem={({ item }) => (item.excludePadding ? item.element : <Box px={2}>{item.element}</Box>)}
-        />
+        {this.state.fetchingData ? (
+          <Flex style={{ flex: 1, height: "100%" }} flexDirection="row" justifyContent="center" alignItems="center">
+            <Spinner />
+          </Flex>
```
olerichter00, line 324:
```suggestion
          <ProvidePlaceholderContext>
            <AboveTheFoldPlaceholder />
          </ProvidePlaceholderContext>
```

**Suggestion:** Let's use the same placeholder/loading indicator we use when entering the screen.

olerichter00 (approved):
Looks good. What do you think of using the placeholder we use when we initially loading the screen (see comment)?

src/lib/Scenes/Artwork/Artwork.tsx
kalshannikava: Good idea to use the placeholder! Done

PR #5068, opened 2021-06-30
fix(CX-1577): featured fair show more button doesn't work

kalshannikava:

This PR resolves [CX-1577]

### Description

Original report is that 'Show more' button doesn't work. This button should be removed and changed to infinite scroll. Discussion: https://artsy.slack.com/archives/C01B2P6LJUU/p1625139230356700
**The issue:**
https://user-images.githubusercontent.com/27921300/123990113-c9123680-d9c9-11eb-8a88-ac7f0328ba9b.mov

**The fix for the case when we need to fetch more data:** 
https://user-images.githubusercontent.com/27921300/124141825-02aa7680-da8a-11eb-9691-6c7f596d3131.mov

**The fix for the case when the end of the list is reached:** 
https://user-images.githubusercontent.com/27921300/124141666-e0b0f400-da89-11eb-80ff-42b07b5178ac.mov

Note: my simulator works super slow, so videos are kinda glitchy, sorry for that.

-

#### iOS user-facing changes

-

-

#### Dev changes

-

[CX-1577]: https://artsyproduct.atlassian.net/browse/CX-1577


src/lib/Scenes/Fair/Components/FairExhibitors.tsx
```diff
@@ -61,22 +45,19 @@ const FairExhibitors: React.FC<FairExhibitorsProps> = ({ fair, relay }) => {
...
-        </Col>
-      </Row>
-    </>
+      }}
+      keyExtractor={(item) => String(item?.node?.id)}
+      onEndReached={loadMoreExhibitors}
+      ListFooterComponent={
+        isLoading ? (
+          <Box p={2} style={{ height: 50 }}>
+            <Flex style={{ flex: 1 }} flexDirection="row" justifyContent="center">
```
olerichter00, line 54:
```suggestion
          <Box p={2}>
            <Flex flex={1} flexDirection="row" justifyContent="center">
```

olerichter00 (approved):
Nice, looks very good!

src/lib/Scenes/Fair/Components/FairExhibitorRail.tsx
```diff
@@ -74,6 +74,7 @@ const FairExhibitorRail: React.FC<FairExhibitorRailProps> = ({ show }) => {
       </Box>
       <FlatList
         horizontal
+        scrollEventThrottle={1000}
```
MounirDhahri, line 77: **issue:** why do we have to set this 1000?

src/lib/Scenes/Fair/Components/FairExhibitors.tsx
```diff
@@ -61,22 +45,19 @@ const FairExhibitors: React.FC<FairExhibitorsProps> = ({ fair, relay }) => {
...
-            onPress={handlePress}
-            disabled={!relay.hasMore()}
-          >
-            Show more
-          </Button>
-        </Col>
-      </Row>
-    </>
+      }}
+      keyExtractor={(item) => String(item?.node?.id)}
```
MounirDhahri, line 49: **issue(blocker):** I am not sure why would we be getting an item with no `id` here 🤔 usually if you need to use conditional chaining in this case, then there must be a problem with the data we're passing. I recommend you check for the data first probably to fix this issue

MounirDhahri (commented):
Thanks for improving this component greatly! I left a small comment that I hope you get to address

src/lib/Scenes/Fair/Components/FairExhibitors.tsx
olerichter00: It might be a good idea to use the helper method `extractNodes` (https://github.com/artsy/eigen/blob/5fc768317095f3cb18c2a140521d7d70b50de93c/src/lib/utils/extractNodes.ts#L1) in this component to extract the shows. The code will look cleaner and we get a safe way to, well...., extract the nodes.

src/lib/Scenes/Fair/Components/FairExhibitorRail.tsx
olerichter00: I guess there is no need to, I removed it.

olerichter00:
#squashongreen

PR #5105, opened 2021-07-08
feat: add tracking to unable to load screen

MounirDhahri:

This PR resolves [CX-1668]

### Description
- Add Tracking to "unable to load" screen

https://user-images.githubusercontent.com/11945712/124904995-dccf2580-dfe5-11eb-85ac-30e303660757.mov

-

#### iOS user-facing changes

-

-

#### Dev changes

- Add tracking to unable to load screen

</details>

[CX-1668]: https://artsyproduct.atlassian.net/browse/CX-1668

### DIFF
--- a/src/lib/Components/LoadFailureView.tsx
+++ b/src/lib/Components/LoadFailureView.tsx
@@ -1,7 +1,9 @@
+import { screen } from "lib/utils/track/helpers"
 import { debounce } from "lodash"
 import { color, Flex, Text, Touchable } from "palette"
-import React, { useRef, useState } from "react"
+import React, { useEffect, useRef, useState } from "react"
 import { Animated, Easing } from "react-native"
+import { useTracking } from "react-tracking"
 import { ReloadIcon } from "../../palette/svgs/ReloadIcon"
 
 interface LoadFailureViewProps {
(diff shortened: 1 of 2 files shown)


src/lib/Components/LoadFailureView.tsx
```diff
@@ -71,3 +80,11 @@ export const LoadFailureView: React.FC<LoadFailureViewProps> = (props) => {
     </Flex>
   )
 }
+
+const tracks = {
+  screen: () =>
+    screen({
+      // @ts-ignore
+      context_screen_owner_type: "unableToLoad",
```
MounirDhahri, line 88: note: I am using a `ts-ignore` and hard coding the screen here instead of adding it to cohesion because so far we are using cohesion for actual screens in the app

olerichter00 (commented):
Can we track which screen failed to load and why (error message, query etc.). With this information, we could identify screens and queries that fail.

Would be helpful for this [ticket](https://artsyproduct.atlassian.net/browse/CX-1690)

MounirDhahri:
Yes, that should be possible but I think it will require a bit of work to implement. I think we better have this as part of a separate spike because I don't see a straightforward implementation for it

MounirDhahri:
I will merge this since we will be creating a follow-up ticket for it on the backlog.

PR #5143, opened 2021-07-19
feat(cx-1636): send token to gravity

lordkiz:

This PR resolves [CX-1636]

### Description
- Allows retrieved push notification token on android to be sent to gravity. This is saved as part of current user's `Device`(s).
- #### TODO:
- `production` param is sent as `!__DEV__`. This means beta builds will send `production: true` to gravity. We must refactor this in Eigen asap as soon as we can determine betas on android.

-

#### iOS user-facing changes

-

-

#### Dev changes

- Save android push notification tokens on gravity - kizito

</details>

[CX-1636]: https://artsyproduct.atlassian.net/browse/CX-1636


ArtsyOpenSource:

- Dev changes (Save android push notification tokens on gravity - kizito)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against f5e9287bcc4bf07e0ff3aadbdd3a93f6e3c079ac
</p>

pvinis (approved):
I'll approve. Code seems 👌👌, but I have not run this PR on my device to test it.

src/lib/utils/PushNotification.ts
```diff
@@ -17,6 +25,69 @@ export const CHANNELS = [
...
+    const pendingStorageKey =
+      os === "android" ? PENDING_ANDROID_PUSH_NOTIFICATION_TOKEN : PENDING_IOS_PUSH_NOTIFICATION_TOKEN
+
+    const previousToken = await AsyncStorage.getItem(storageKey)
+    if (token !== previousToken) {
+      const { authenticationToken, userAgent } = getCurrentEmissionState()
+      if (!authenticationToken) {
+        // user is not logged in. The first time a user opens the app, expect token to be gotten before the global store is initialised
+        // save the token and send to gravity when they log in
+        await AsyncStorage.setItem(pendingStorageKey, token)
```
MounirDhahri, line 53: **nitpick:** I would have used a boolean to save the state of the key (pending or not) to avoid having to manage two tokens

src/lib/utils/PushNotification.ts
```diff
@@ -17,6 +25,69 @@ export const CHANNELS = [
 
 type TypedNotification = Omit<ReceivedNotification, "userInfo"> & { title?: string }
 
+export const savePendingToken = async () => {
+  const pendingStorageKey =
+    Platform.OS === "android" ? PENDING_ANDROID_PUSH_NOTIFICATION_TOKEN : PENDING_IOS_PUSH_NOTIFICATION_TOKEN
+  const token = await AsyncStorage.getItem(pendingStorageKey)
+  if (token) {
+    const saved = await saveToken({ os: Platform.OS, token })
```
MounirDhahri, line 33: **issue(non-blocker):** I don't think there is a need to send the os here since you can get it from `Platform.OS` inside `saveToken` method. `saveToken` will then become simply `saveToken(token)`

src/lib/utils/PushNotification.ts
```diff
@@ -17,6 +25,69 @@ export const CHANNELS = [
...
+          if (__DEV__) {
+            console.warn(`New Push Token ${token} was NOT saved`, response?.error)
+          }
+          reject("Push Notification: Failed to save new push notification token")
+          return
+        }
+        if (__DEV__) {
+          console.log(`New Push Token ${token} saved!`)
+        }
+        await AsyncStorage.setItem(storageKey, token)
```
MounirDhahri, line 84: **thought:** is there a reason to save the fcm token if we already sent it to gravity? I don't think it will be used anywhere on the client side.

MounirDhahri (approved):
I left few minor comments that I believe we should be able to merge once they got addressed.

src/lib/utils/PushNotification.ts
lordkiz: Yes. For comparison purpose. Fcm tokens is usually regenerated on fresh install but there is no guarantee that the tokens cannot be regenerated spontaneously. If we do not compare we will be sending the same token to gravity every time the user opens the app, because onRegister is triggered everytime the app starts.

src/lib/utils/PushNotification.ts
lordkiz: I opted for this because I wanted to pass the entire token object gotten from onRegister callback to saveToken as is. There is no added advantage. I will refactor to use Platform.OS.

src/lib/utils/PushNotification.ts
MounirDhahri: perfect. Thanks Kizito

src/lib/utils/PushNotification.ts
MounirDhahri: If we opt with this way, we will be only saving one key I guess.

lordkiz: Makes sense. It will be a lot cleaner. And also there will be no need having separate keys for each platform, especially if saveToken uses Platform.OS. I will refactor.

MounirDhahri (approved):
GG 🎉🎉🎉

PR #5166, opened 2021-07-24
feat(CX-1778): Relay query prefetching

olerichter00:

This PR resolves [CX-1778]

This Feature is behind a feature switch and won't be released with the PR.

### Description

This PR implements **Relay Query Prefetching**, which allows us to prefetch the query for any screen via URL. A prefetched query is cached by Relay and is immediately available when the user navigates to the screen.

```JSX

prefetchUrl("/artist/wolfgang-tillmans")
```
To automatically prefetch in-viewport links, this PR adds `PrefetchFlatList` which wraps `FlatList` and provides the functionality to prefetch URLs for each element once the element is visible to the user and the user has interacted (e.g. scrolled) with the `FlatList`.

```JSX
// Calls `prefetchUrl(item.href, item.slug)` for each item

<PrefetchFlatList
  prefetchUrlExtractor={(item) => item.href}
  prefetchVariablesExtractor={(item) => ({ saleSlug: item.slug })}
  data={data}
  renderItem={renderItem}
 />
```

**Changes:**

- [x] Adds logic to prefetch Relay queries for a given URL ([code](https://github.com/artsy/eigen/pull/5166/files#diff-c396b9a5ba93a067facd296a43809afaf8b908d3457ba9fc239d7ea1169e8191R1)).
- [x] Introduces a [PrefetchFlatlist](https://github.com/artsy/eigen/pull/5166/files#diff-bbddf9b75bd35091fb322115540a306ad9f1ffbef791787753534eef6cbc45f2R1) component which is able to prefetch elements as soon as they are visible on the screen.
- [x] Adds screen queries to the **app registry** in order to prefetch queries by URL ([example](https://github.com/artsy/eigen/pull/5166/files#diff-757fcd98f2962f4e88d3ef7810c702024da8d0f39a22efd11921db149ccdde39R267)).
- [x] Uses `PrefetchFlatlist` within [AboveTheFoldFlatList](https://github.com/artsy/eigen/pull/5166/files#diff-54433a1ad73674cb3a70f9a34712ef3e2201fc4fa1fe905713dd699f95f775deR1) to be able to prefetch elements within the `AboveTheFoldFlatList`.
- [x] Prefetches artworks in the [ArtistRail](https://github.com/artsy/eigen/pull/5166/files#diff-5b2f88630a64476d1e86492328a7d8e19be249061abac04f00400968e1cd9616R183) and some other places.
- [x] Uses rate limiting to avoid clients from sending too many requests ([code](https://github.com/artsy/eigen/pull/5166/files#diff-c396b9a5ba93a067facd296a43809afaf8b908d3457ba9fc239d7ea1169e8191R16)).
- [x] Adds a feature flag to disable query prefetching for now ([code](https://github.com/artsy/eigen/pull/5166/files#diff-69e8dce13d31b9746033fb9b2e5de2672f4041627b223c0ad3b17ed8452b8084R204)).
## Demo

https://user-images.githubusercontent.com/4691889/143563793-41c1eece-41cf-4247-b25b-a27363fd84e4.mov

- Introduce query prefetching (behind feature flag) - ole

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-1778]: https://artsyproduct.atlassian.net/browse/CX-1778?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 5167a049a12dbbf056d244c5140626f6760f2f48
</p>

src/lib/AppRegistry.tsx
```diff
@@ -277,7 +283,7 @@ export const modules = defineModules({
   Artist: reactModule(ArtistQueryRenderer),
   ArtistShows: reactModule(ArtistShows2QueryRenderer),
   ArtistSeries: reactModule(ArtistSeriesQueryRenderer),
-  Artwork: reactModule(Artwork),
+  Artwork: reactModule(Artwork, {}, ArtworkScreenQuery),
```
olerichter00, line 286: Adding the screen query to the module to [later](https://github.com/artsy/eigen/pull/5166/files#diff-c396b9a5ba93a067facd296a43809afaf8b908d3457ba9fc239d7ea1169e8191R29) map the url to the query.

src/lib/utils/prefetchQuery.ts
```diff
@@ -0,0 +1,31 @@
...
+  createOperationDescriptor,
+  Environment,
+  fetchQuery,
+  getRequest,
+  GraphQLTaggedNode,
+  Variables,
+} from "relay-runtime"
+import { RelayModernEnvironment } from "relay-runtime/lib/store/RelayModernEnvironment"
+
+export const prefetchQuery = async (environment: Environment, query: GraphQLTaggedNode, variables: Variables = {}) => {
```
olerichter00, line 14: We should probably throttle this function or limit the maximum number of invocations.

src/lib/Components/Home/ArtistRails/ArtistRail.tsx
```diff
@@ -183,6 +183,13 @@ const ArtistRail: React.FC<Props & RailScrollProps> = (props) => {
         <SectionTitle title={props.title} subtitle={props.subtitle} />
       </Flex>
       <CardRailFlatList<SuggestedArtist>
+        prefetchUrlExtractor={(item) => item?.href!}
+        prefetchVariablesExtractor={() => ({
+          input: {
+            dimensionRange: "*-*",
+            sort: "-decayed_merch",
+          },
```
MounirDhahri, line 191: **nitpick:** in order to keep parity with the `Artist` screen initial input and avoid the need to maintain the input in both places, maybe you can just extract the initial input and use it for the `prefetchVariablesExtractor` and also here https://github.com/artsy/eigen/blob/54780cf608b241b0ba68471e52730cf59cdf5621/src/lib/Scenes/Artist/Artist.tsx#L146-L150

src/lib/Components/PrefetchFlatList.tsx
```diff
@@ -0,0 +1,70 @@
+import { usePrefetch } from "lib/utils/prefetchQuery"
+import React, { Ref, useCallback, useReducer, useRef } from "react"
+import { FlatList, FlatListProps, ViewToken } from "react-native"
+
+export type PrefetchFlatListProps<ItemType> = {
+  prefetchUrlExtractor?: (item?: ItemType) => string | undefined
+  prefetchVariablesExtractor?: (item?: ItemType) => object
```
MounirDhahri, line 7:
```suggestion
  prefetchQueryUrlExtractor?: (item?: ItemType) => string | undefined
  prefetchQueryVariablesExtractor?: (item?: ItemType) => object
```
**thought:** This is just a personal preference, I don't care much about how long variable names are as long as they are descriptive

src/lib/Components/PrefetchFlatList.tsx
```diff
@@ -0,0 +1,70 @@
...
+      // Prefetch all changed items that are not already prefetched
+      changed.forEach((item) => {
+        const url = prefetchUrlExtractor?.(item.item)
+        const variables = prefetchVariablesExtractor?.(item.item)
+
+        if (!url || viewedUrls[url]) {
+          return
+        }
+
+        addViewedUrl(url)
```
MounirDhahri, line 55: **praise:** brilliant 👏

src/lib/Scenes/BottomTabs/BottomTabs.tsx
```diff
@@ -7,6 +8,12 @@ import { ICON_HEIGHT } from "./BottomTabsIcon"
 
 export const BottomTabs: React.FC = () => {
   const { color } = useTheme()
+
+  const prefetchUrl = usePrefetch()
+
+  useEffect(() => {
+    prefetchUrl("inbox")
+  }, [])
```
MounirDhahri, line 16: **question:** is this because the inbox is slow? if so maybe we can delay this with a second or so

src/lib/Scenes/Home/Components/SalesRail.tsx
```diff
@@ -73,6 +73,8 @@ const SalesRail: React.FC<Props & RailScrollProps> = (props) => {
         />
       </Flex>
       <CardRailFlatList<Sale>
+        prefetchUrlExtractor={(item) => item?.href!}
+        prefetchVariablesExtractor={(item) => ({ saleSlug: item?.slug })}
```
MounirDhahri, line 77: **issue:** I think you should probably adjust the typings here to be like in `renderItem` in line 80 so that we get the proper types. I know in this case they will be always there but I am afraid someone with less experience with the product would be confident that a property exists when it doesn't

src/lib/store/config/features.ts
```diff
@@ -201,6 +201,12 @@ export const features = defineFeatures({
     showInAdminMenu: true,
     echoFlagKey: "AREnableVisualProfileIconAndBio",
   },
+  ARPrefetching: {
```
MounirDhahri, line 204: **nipitck:** 
```suggestion
  AREnableQueriesPrefetching: {
```

src/lib/utils/prefetchQuery.ts
```diff
@@ -0,0 +1,84 @@
...
+  const result = matchRoute(url)
+
+  if (result.type !== "match") {
+    return
+  }
+
+  const module = modules[result.module]
+
+  if (module.type !== "react") {
+    console.error(`Failed to prefetch ${url} (cannot prefetch native module).`)
```
MounirDhahri, line 55:
```suggestion
    console.error(`Failed to prefetch ${url} (cannot prefetch react module).`)
```

src/lib/utils/prefetchQuery.ts
```diff
@@ -0,0 +1,84 @@
...
+  createOperationDescriptor,
+  Environment,
+  fetchQuery,
+  getRequest,
+  GraphQLTaggedNode,
+  Variables,
+} from "relay-runtime"
+import { RelayModernEnvironment } from "relay-runtime/lib/store/RelayModernEnvironment"
+
+const limiter = new RateLimiter({ tokensPerInterval: 60, interval: "minute", fireImmediately: true })
```
MounirDhahri, line 16: **thought:** I am interested in making this dynamic actually and getting it from somewhere and adjusting it to see its effect. In echo, we store some properties in messages, we can do it there as well https://github.com/artsy/echo/blob/eaf4eddee50e719f3fed53a0d87428202a220585/Echo.json5#L57

src/lib/utils/prefetchQuery.ts
```diff
@@ -0,0 +1,84 @@
...
+}
+
+export const usePrefetch = () => {
+  const enablePrefetching = useFeatureFlag("ARPrefetching")
+
+  if (!enablePrefetching || __TEST__) {
+    return () => null
+  }
+
+  const environment = useRelayEnvironment()
```
MounirDhahri, line 81: **thought:** is there a reason why you didn't use `defaultEnvironment`

MounirDhahri (commented):
Great work Ole, I enjoyed reviewing this PR. I left few small comments

src/lib/utils/prefetchQuery.ts
olerichter00: Good catch!

src/lib/Components/Home/ArtistRails/ArtistRail.tsx
olerichter00: Good idea! I extracted the default input like this:

```
export const defaultArtworksInput = () => ({
  input: {
    dimensionRange: "*-*",
    sort: DEFAULT_ARTWORK_SORT.paramValue,
  },
})
```

src/lib/Components/PrefetchFlatList.tsx
olerichter00: I think both versions would work. I used the version without `Query` because `PrefetchFlatList` could also work with another implementation of `prefetchUrl` that doesn't use the query to prefetch (e.g. preloading the component). Does that make sense?

src/lib/Scenes/BottomTabs/BottomTabs.tsx
olerichter00: I plan to prefetch all bottom tabs at some point because they are very likely to be clicked soon after the user opens the app. The query doesn't take too much time, so I don't think delaying the prefetching is necessary and just adds more complexity.

src/lib/Scenes/Home/Components/SalesRail.tsx
olerichter00: I don't think I completely understand. Aren't the types the same (except for `(item) => {}` vs. `({item}) => {}`)? Maybe you can write down how you would change it (`prefetchUrlExtractor` is specified [here](https://github.com/artsy/eigen/blob/be8f28c63da087d62ed90875c2497afaf99e506d/src/lib/Components/PrefetchFlatList.tsx#L6)).

src/lib/store/config/features.ts
olerichter00: 👍

src/lib/utils/prefetchQuery.ts
olerichter00: I changed it to `console.error(`Failed to prefetch ${url} (cannot prefetch ${module.type} module).`)`.

src/lib/Components/Home/ArtistRails/ArtistRail.tsx
MounirDhahri: nice, thanks

src/lib/Components/PrefetchFlatList.tsx
MounirDhahri: Yes, of course it makes sense. It was just a personal preference

src/lib/Scenes/Home/Components/SalesRail.tsx
MounirDhahri: what I meant is, would it be possible for example to have `prefetchUrlExtractor={(item) => item.href}` instead of `prefetchUrlExtractor={(item) => item?.href!}`

MounirDhahri: probably there is a typing mistake somewhere that is causing that

src/lib/utils/prefetchQuery.ts
MounirDhahri: even better!

src/lib/Scenes/BottomTabs/BottomTabs.tsx
MounirDhahri: ah oki then, I got you

Podfile.lock
```diff
@@ -825,14 +825,7 @@ DEPENDENCIES:
...
 SPEC REPOS:
-  https://github.com/artsy/Specs.git:
-    - Aerodramus
-    - "Artsy+UIColors"
-    - "Artsy+UIFonts"
-    - "Artsy+UILabels"
-    - Artsy-UIButtons
-    - Extraction
-  trunk:
+  https://cdn.cocoapods.org/:
```
MounirDhahri, line 828: I think this change is intentional as we still want to get our specs for the artsy specs repo. @brainbicycle referred me to a way to fix that a few months ago and I forgot how. I hope he still remembers it

MounirDhahri (approved):
this is great work. We need to enable the feature flag here asap to get enough time to test this

src/lib/Scenes/Home/Components/SalesRail.tsx
olerichter00: The issue is that `salesModule.results` (the data that is passed into FlatList) has the type `(Result | null)[]`. Every item could be `null` and we have to use `item?.href` (in `prefetchUrlExtractor` and in `renderItem`).

MounirDhahri: aha I get you now, thanks for explaining

MounirDhahri (approved):
🚀

MounirDhahri:
@olerichter00 Would you mind notifying people in the #practice-mobile channel once we merge this PR to report any possible regressions

MounirDhahri (dismissed):

src/lib/Scenes/Home/Home.tsx
```diff
@@ -180,6 +180,12 @@ const Home = (props: Props) => {
           data={modules}
           initialNumToRender={5}
           refreshControl={<RefreshControl refreshing={isRefreshing} onRefresh={handleRefresh} />}
+          prefetchUrlExtractor={(item) => {
+            switch (item?.type) {
+              case "newWorksForYou":
```
pvinis, line 185: question: what about every other case? maybe just an `if` works? how are you thinking of this?

src/lib/Scenes/MyProfile/MyProfile.tsx
```diff
@@ -146,16 +146,19 @@ export const OldMyProfileContainer = createRefetchContainer(
...
   <QueryRenderer<MyProfileQuery>
     environment={defaultEnvironment}
-    query={graphql`
-      query MyProfileQuery {
-        me @optionalField {
-          ...MyProfile_me
-        }
-      }
-    `}
+    /* tslint:disable relay-operation-generics */
```
pvinis, line 160: whats this about? if this rule doesnt give us anything, should we remove it?

src/lib/Scenes/Home/Components/SalesRail.tsx
```diff
@@ -73,6 +73,8 @@ const SalesRail: React.FC<Props & RailScrollProps> = (props) => {
         />
       </Flex>
       <CardRailFlatList<Sale>
+        prefetchUrlExtractor={(item) => item?.href!}
```
pvinis, line 76: it would be so cool to have something like ruby's `&:href`. ts could have something like `&?.href!` instead of `item => item?.href!`. (or some other symbol if `&` is not possible.

pvinis (dismissed):

src/lib/Scenes/Home/Home.tsx
olerichter00: I just used the same logic as in [`renderItem`](https://github.com/artsy/eigen/blob/fc504169eaf54c31f5d4516b6fa1999f48e48324/src/lib/Scenes/Home/Home.tsx#L193) and plan to implement prefetching for other modules at a later point.

src/lib/Scenes/MyProfile/MyProfile.tsx
olerichter00: Very good question! [relay-operation-generics](https://github.com/relay-tools/tslint-plugin-relay#relay-operation-generics) affects `mutations` as well and is probably helpful in general. What do you think? Remove it or keep it?

src/lib/Scenes/Home/Components/SalesRail.tsx
olerichter00: Something like [Ramda's prop](https://ramdajs.com/docs/#prop) or just:

```
const ˆ = (attribute) => (item) => item && item[attribute]

prefetchUrlExtractor={^("href")}
```

src/lib/Scenes/MyProfile/MyProfile.tsx
pvinis: I guess we are violating this rule here `QueryRenderer components must use graphql tagged template strings for their query prop.`.

Honestly, I prefer the query being outside. This is the way it's done with hooks too, so it's good preparation.

I would vote to completely remove this rule (which is the only rule that this package provides?). I would get rid of that rule, get rid of the tslint comments, and then if that package gives no other rules, get rid of the package too.

src/lib/Scenes/Home/Components/SalesRail.tsx
pvinis: ~~yea.. I played around with that and I kinda don't like that there is basically zero type-checking with this way. Because of this I still prefer the old school way.~~

LOOK! I basically took what I had, combined what you had, BOOM!
https://www.typescriptlang.org/play?#code/FAYw9gdgzgLgBAUwB4EMC2AHANguBeAbWDhLgG84ALAJwQDMAuOAIhUpQEZmBfAGmNIUa9Jq3YAmHvwC6wOeGjwAJPjgAeANJRESGAggATbbGoBLCAHNecAEoIQOvYe13w1A5qjWUEAJ4A+fwAKFBgYMwAjAFc9Ji0ASnx-OCDTPTQmV0S8ZLSENDgAMkK4PLQCUPDTaL1ZeUhYKlo6bTwddGwEADo0FAwgpSDmYTpmePjQBrAcLqwwCyCRqAmgA

Oh this is so good. Ok, not for this PR, but let's talk about this in the show and tell or next KS? I love it!

olerichter00: That looks great! Going [point-free](https://medium.com/dailyjs/functional-js-7-point-free-style-b21a1416ac6a)

src/lib/Scenes/MyProfile/MyProfile.tsx
olerichter00: Sounds good, I'll remove the `relay-operation-generics` rule from tslint.

MounirDhahri (approved):
Let's gooo 🎉

PR #5186, opened 2021-07-28
fix(palette flag): separated the temporary palette flag

pvinis:

This PR resolves the issue here https://artsy.slack.com/archives/C02BAQ5K7/p1627387196273400.

### Description

easy-peasy/redux cause a lot of require cycles, but usually they don't cause problems. The latest flag for palette v3 just broke that limit. iOS was ok with that but not Android.

The main problem was that, starting from `Theme`, we have a hook that needs to check a flag. That flag lives in GlobalStore, that requires NativeModel. NativeModel needs `navigate`, and that needs `AppRegistry`. That file imports all screens, and of course a lot of them use Theme. And this is a nice require cycle that breaks our stuff.

One of the two possible solutions was to remove the flag, but I didn't want that. Code listens to me, not the other way around 😆. So the other solution was to not use redux in the "global way" it works, which causes the cycles. I thought I'll try `zustand`, which is kinda redux-but-only-the-good-parts, and also it allows/prefers many small stores, instead of one humongous one, which helps alleviate cycles.

So yea, that's what this PR here is, plus two other small things:
- `DTDShowAnalyticsVisualiser` -> `DTShowAnalyticsVisualiser`
- removing `styledWrapper` to just `styled`

-

#### iOS user-facing changes

-

-

#### Dev changes

- use a separate store for the temporary palette flag

</details>

### DIFF
--- a/package.json
+++ b/package.json
@@ -154,7 +154,8 @@
     "supercluster": "6.0.1",
     "tipsi-stripe": "https://github.com/artsy/tipsi-stripe.git#fix-infinite-loop",
     "url": "^0.11.0",
-    "yup": "0.29.1"
+    "yup": "0.29.1",
+    "zustand": "^3.5.7"
   },
   "devDependencies": {
     "@artsy/update-repo": "0.2.1",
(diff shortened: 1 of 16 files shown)


ArtsyOpenSource:

- Dev changes (use a separate store for the temporary palette flag)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 4293a782d96a420d17b8b5ed7aea944c82e8ae9e
</p>

dimatretyak (approved):
LGTM 👍

olerichter00:
Looks good!

PR #5195, opened 2021-07-29
feat(FX-3109): tag page

tr-ann:
Jira ticket: [FX-3109](https://artsyproduct.atlassian.net/browse/FX-3109)

This PR resolves []

### Description

- `tag/:slug` route render Tag page
- Added artworks grid, filters modal and sticky filter bar
- Added tests for the page

### Demo

https://user-images.githubusercontent.com/56556580/127502537-95079861-e624-460f-85cd-20b45009fd87.mov

- Tag scene

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>


ArtsyOpenSource:

- Cross-platform user-facing changes (Tag scene)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 37736c91e11a33bd8a1f94ac467d453d01c6a704
</p>

src/lib/Scenes/Tag/Tag.tsx
```diff
@@ -0,0 +1,113 @@
...
+    <ProvideScreenTracking
+      info={{
+        context_screen: Schema.PageNames.TagPage,
+        context_screen_owner_type: Schema.OwnerEntityTypes.Tag,
+        context_screen_owner_id: tagID,
+        context_screen_owner_slug: tag.slug,
+      }}
+    >
+      <View style={styles.container}>
+        <StickyTabPage staticHeaderContent={headerContent} stickyHeaderContent={<></>} tabs={tabs} />
```
tr-ann, line 54: Set `stickyHeaderContent={<></>}` to hide the tab title. In this case only filter bar are sticky.

src/lib/Components/Tag/Header.tsx
```diff
@@ -0,0 +1,30 @@
...
+interface HeaderProps {
+  tag: Header_tag
+  relay: RelayProp
+}
+
+const Header: React.FC<HeaderProps> = ({ tag }) => {
+  return (
+    <Box marginTop={60} justifyContent="center">
+      <Sans size="8" numberOfLines={2}>
+        {tag.name || ""}
```
mdole, line 15: **minor:** I don't think we need the `|| ""` here; all tags have names (though looking at our data, many names are things like IDs or URLs, it's definitely not clean data)

src/lib/Components/Tag/Header.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+interface HeaderProps {
+  tag: Header_tag
+  relay: RelayProp
+}
+
+const Header: React.FC<HeaderProps> = ({ tag }) => {
+  return (
+    <Box marginTop={60} justifyContent="center">
+      <Sans size="8" numberOfLines={2}>
```
mdole, line 14: **minor:** we should use `<Text>` instead of `<Sans>` - both work, but `<Text>` is newer and is our preferred method :)

src/lib/Components/Tag/Header.tsx
```diff
@@ -0,0 +1,30 @@
...
+    </Box>
+  )
+}
+
+export default createFragmentContainer(Header, {
+  tag: graphql`
+    fragment Header_tag on Tag {
+      internalID
+      slug
+      id
```
mdole, line 26: **minor:** do we need these three? Seems like we only display `name` in this component - does something else break if we don't fetch them?

src/lib/Components/Tag/__tests__/About-tests.tsx
```diff
@@ -0,0 +1,15 @@
+import React from "react"
+import "react-native"
+
+// Note: test renderer must be required after react-native.
```
mdole, line 4: **very minor:** we probably don't need to keep this one 😄

src/lib/Components/Tag/__tests__/Header-tests.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+it("renders without throwing a error", () => {
+  const tag = {
+    id: "handmade-paper",
+    internalID: "gravity-id",
+    gravityID: "handmade-paper",
+    name: "Handmade Paper",
+  }
+
+  renderWithWrappers(<Header tag={tag as any} />)
```
mdole, line 14: **Curiosity:** why is this `as any` required?

src/lib/Components/Tag/__tests__/TagArtworks-tests.tsx
```diff
@@ -0,0 +1,156 @@
...
+          }
+
+          return null
+        }}
+        variables={{
+          tagID,
+          // input: {
+          //   medium: "*",
+          //   priceRange: "*-*",
+          // },
```
mdole, line 72: **Minor:** could get rid of these

src/lib/Scenes/Tag/Tag.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+interface TagProps {
+  tagID?: string
+  tag: NonNullable<TagQueryResponse["tag"]>
+}
+
+interface TagQueryRendererProps {
+  tagID: string
+  medium?: string
+  price_range?: string
```
mdole, line 25: Maybe we don't need these as well?

src/lib/Scenes/Tag/Tag.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+export const Tag: React.FC<TagProps> = (props) => {
+  const { tag, tagID } = props
+
+  const tabs: TabProps[] = [
+    {
+      title: "Artworks",
+      content: <TagArtworksPaginationContainer tag={tag} />,
+    },
+  ]
```
mdole, line 36: Could we use `<StickyHeaderPage />` instead of `<StickyTabPage />` since we have no tabs for this screen?

src/lib/Scenes/Tag/Tag.tsx
```diff
@@ -0,0 +1,113 @@
...
+    paddingLeft: commonPadding,
+    paddingRight: commonPadding,
+    ...(isPad
+      ? {
+          width: 330,
+          alignSelf: "center",
+        }
+      : {}),
+  },
+})
```
mdole, line 113: We generally prefer [styled-components](https://styled-components.com/) over StyleSheet.create these days; this is left over from early code in the app. Suggest refactoring if you have time!

mdole (changes requested):
Awesome work, @tr-ann! I left several comments but they're all very small. Thanks for taking this on, and welcome to Eigen!

src/lib/Scenes/Tag/Tag.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+export const TagQueryRenderer: React.FC<TagQueryRendererProps> = (props) => {
+  const { tagID, medium, price_range } = props
+  const input = getParamsForInputByFilterType(
+    {
+      medium,
+      priceRange: price_range,
+    },
+    "tagArtwork"
+  ) as FilterArtworksInput
```
dimatretyak, line 68: I think that passing `medium`, `price_range` and also passing `input` to `QueryRender` is not necessary for `Tag`

src/lib/Components/Tag/TagArtworks.tsx
```diff
@@ -0,0 +1,236 @@
...
+            LOCATION_CITY
+            PRICE_RANGE
+            MATERIALS_TERMS
+            PARTNER
+            ARTIST_NATIONALITY
+            MAJOR_PERIOD
+            ARTIST
+            TOTAL
+          ]
+          forSale: true
```
dimatretyak, line 165: Do we have to pass the `forSale` parameter?

src/lib/Components/Tag/TagArtworks.tsx
```diff
@@ -0,0 +1,236 @@
...
+              value
+              name
+              count
+            }
+          }
+          edges {
+            node {
+              id
+            }
+          }
```
dimatretyak, line 183: Perhaps there is also no need to query `edges` here

src/lib/Components/Tag/About.tsx
```diff
@@ -0,0 +1,34 @@
+import { About_tag } from "__generated__/About_tag.graphql"
```
dimatretyak, line 1: Is the `About `component used on the `Tag` page?

src/lib/Components/Tag/Header.tsx
```diff
@@ -0,0 +1,30 @@
+import { Header_tag } from "__generated__/Header_tag.graphql"
```
dimatretyak, line 1: I think it would be better to move the `Header` component to the `src/lib/Scenes/Components` folder

src/lib/Components/Tag/About.tsx
tr-ann: Hmm, it's seems like description for tags is empty (checked for some tags). I'm not sure description should be displayed for tag. If not I just remove `About` component.

@mdole what do you think?

src/lib/Components/Tag/__tests__/Header-tests.tsx
tr-ann: The `tag` prop for `Header` has `Header_tag` type which contains `name`, `$fragmentRefs` and other fields received from relay. So to avoid errors like "missing property from type" I just send custom object with `any` type

src/lib/Scenes/Tag/Tag.tsx
tr-ann: I think we can use `<StickyHeaderPage />`, but in this case we should move filters tab to `Tag.tsx` and change the structure of the page (I think there will be a lot of changes here). So use  `<StickyTabPage />` is the easiest way to use sticky filter tab it tag page.

Maybe @dimatretyak can add some more description for this

src/lib/Components/Tag/About.tsx
tr-ann: Update: Tag description is requested for tag page in Force. So I leave `About` component in Tag page

mdole (approved):
Woohoo, looking good! Let's get it in the beta and see how it feels 😎

PR #5208, opened 2021-08-02
Feat (CX-1768): handle data in notification

lordkiz:

This PR resolves [CX-1768]

### Description
Handles the display and proper routing of users based on custom data in push notification payload.
##### Behaviour
- if user is logged in, tapped push notification is handled immediately
- If user is not logged in, tapped push notification is temporarily saved and handled if user logs in within 90 seconds else it is discarded.

https://user-images.githubusercontent.com/18648835/127865744-c113a3f8-a340-481a-9fa5-befdcbdf6b65.mov

 

-

#### iOS user-facing changes

-

-Handles the display and proper routing of users based on custom data in push notification payload.

#### Dev changes

-

</details>

[CX-1768]: https://artsyproduct.atlassian.net/browse/CX-1768

### DIFF
--- a/src/lib/AndroidApp.tsx
+++ b/src/lib/AndroidApp.tsx
@@ -17,6 +17,7 @@ import { ModalStack } from "./navigation/ModalStack"
 import { BottomTabsNavigator } from "./Scenes/BottomTabs/BottomTabsNavigator"
 import { ForceUpdate } from "./Scenes/ForceUpdate/ForceUpdate"
 import { Onboarding } from "./Scenes/Onboarding/Onboarding"
+import { handlePendingNotification } from "./utils/PushNotification"
 import { ConsoleTrackingProvider } from "./utils/track/ConsoleTrackingProvider"
 import { AnalyticsConstants } from "./utils/track/constants"
 
@@ -37,6 +38,7 @@ const Main: React.FC<{}> = track()(({}) => {
   const isLoggedIn = GlobalStore.useAppState((state) => !!state.auth.userAccessToken)
   const onboardingState = GlobalStore.useAppState((state) => state.auth.onboardingState)
   const forceUpdateMessage = GlobalStore.useAppState((state) => state.config.echo.forceUpdateMessage)
+  const pendingNotification = GlobalStore.useAppState((state) => state.pendingPushNotification.android)
 
   useSentryConfig()
   useStripeConfig()
@@ -84,6 +86,14 @@ const Main: React.FC<{}> = track()(({}) => {
       }, 500)
     }
   }, [isHydrated])
+
+  useEffect(() => {
+    if (isLoggedIn) {
+      if (pendingNotification) {
+        handlePendingNotification(pendingNotification)
+      }
+    }
+  }, [isLoggedIn])
 
   if (!isHydrated) {
     return <View />
--- a/src/lib/AndroidApp.tsx
+++ b/src/lib/AndroidApp.tsx
@@ -5,9 +5,10 @@ import { addTrackingProvider, track } from "lib/utils/track"
 import { SEGMENT_TRACKING_PROVIDER, SegmentTrackingProvider } from "lib/utils/track/SegmentTrackingProvider"
 import { useDeepLinks } from "lib/utils/useDeepLinks"
 import { useStripeConfig } from "lib/utils/useStripeConfig"
-import React, { useEffect } from "react"
+import React, { useEffect, useState } from "react"
 import { Appearance, UIManager, View } from "react-native"
 import RNBootSplash from "react-native-bootsplash"
+import PushNotification from "react-native-push-notification"
 import { AppProviders } from "./AppProviders"
 import { useWebViewCookies } from "./Components/ArtsyReactWebView"
 import { _FancyModalPageWrapper } from "./Components/FancyModal/FancyModalContext"
@@ -17,7 +18,7 @@ import { ModalStack } from "./navigation/ModalStack"
 import { BottomTabsNavigator } from "./Scenes/BottomTabs/BottomTabsNavigator"
 import { ForceUpdate } from "./Scenes/ForceUpdate/ForceUpdate"
 import { Onboarding } from "./Scenes/Onboarding/Onboarding"
-import { handlePendingNotification } from "./utils/PushNotification"
+import { handlePendingNotification, handleReceivedNotification } from "./utils/PushNotification"
 import { ConsoleTrackingProvider } from "./utils/track/ConsoleTrackingProvider"
 import { AnalyticsConstants } from "./utils/track/constants"
 
@@ -34,6 +35,7 @@ const Main: React.FC<{}> = track()(({}) => {
       webClientId: "673710093763-hbj813nj4h3h183c4ildmu8vvqc0ek4h.apps.googleusercontent.com",
     })
   }, [])
+  const [hasHandledInitialNotification, setHasHandledInitialNotification] = useState(false)
   const isHydrated = GlobalStore.useAppState((state) => state.sessionState.isHydrated)
   const isLoggedIn = GlobalStore.useAppState((state) => !!state.auth.userAccessToken)
   const onboardingState = GlobalStore.useAppState((state) => state.auth.onboardingState)
@@ -89,7 +91,19 @@ const Main: React.FC<{}> = track()(({}) => {
 
   useEffect(() => {
     if (isLoggedIn) {
-      if (pendingNotification) {
+      if (!hasHandledInitialNotification) {
+        // initial notification is most recent and should be prioritised
+        PushNotification.popInitialNotification((notification) => {
+          if (notification) {
+            handleReceivedNotification(notification)
+            // prevent loop where user logs out and logs back in and previously
+            // handled initial push notification is rehandled
+            setHasHandledInitialNotification(true)
+            return
+          }
+          handlePendingNotification(pendingNotification)
+        })
+      } else {
         handlePendingNotification(pendingNotification)
       }
     }
--- a/src/lib/store/GlobalStore.tsx
+++ b/src/lib/store/GlobalStore.tsx
@@ -120,7 +120,7 @@ export function useSelectedTab(): BottomTabType {
 }
 
 let _globalStoreInstance: ReturnType<typeof createGlobalStore> | undefined
-export const globalStoreInstance = (): ReturnType<typeof createGlobalStore> => {
+const globalStoreInstance = (): ReturnType<typeof createGlobalStore> => {
   if (_globalStoreInstance === undefined) {
     _globalStoreInstance = createGlobalStore()
   }
@@ -160,6 +160,17 @@ export function unsafe_getDevToggle(key: DevToggleName) {
     throw new Error(`Unable to access ${key} before GlobalStore bootstraps`)
   }
   return false
+}
+
+export function unsafe_getUserAccessToken() {
+  const state = globalStoreInstance().getState() ?? null
+  if (state) {
+    return state.auth.userAccessToken
+  }
+  if (__DEV__) {
+    throw new Error(`Unable to access userAccessToken before GlobalStore bootstraps`)
+  }
+  return null
 }
 
 export function getCurrentEmissionState() {
--- a/src/lib/store/GlobalStore.tsx
+++ b/src/lib/store/GlobalStore.tsx
@@ -120,7 +120,7 @@ export function useSelectedTab(): BottomTabType {
 }
 
 let _globalStoreInstance: ReturnType<typeof createGlobalStore> | undefined
-const globalStoreInstance = (): ReturnType<typeof createGlobalStore> => {
+export const globalStoreInstance = (): ReturnType<typeof createGlobalStore> => {
   if (_globalStoreInstance === undefined) {
     _globalStoreInstance = createGlobalStore()
   }
--- a/src/lib/store/GlobalStoreModel.ts
+++ b/src/lib/store/GlobalStoreModel.ts
@@ -13,6 +13,7 @@ import { ConfigModel, getConfigModel } from "./ConfigModel"
 import { unsafe__getEnvironment } from "./GlobalStore"
 import { CURRENT_APP_VERSION } from "./migration"
 import { getNativeModel, NativeModel } from "./NativeModel"
+import { getPendingPushNotificationModel, PendingPushNotificationModel } from "./PendingPushNotificationModel"
 import { assignDeep, sanitize } from "./persistence"
 import { getToastModel, ToastModel } from "./ToastModel"
 
@@ -29,6 +30,7 @@ interface GlobalStoreStateModel {
   config: ConfigModel
   auth: AuthModel
   toast: ToastModel
+  pendingPushNotification: PendingPushNotificationModel
 }
 export interface GlobalStoreModel extends GlobalStoreStateModel {
   rehydrate: Action<GlobalStoreModel, DeepPartial<State<GlobalStoreStateModel>>>
@@ -93,6 +95,7 @@ export const getGlobalStoreModel = (): GlobalStoreModel => ({
   config: getConfigModel(),
   auth: getAuthModel(),
   toast: getToastModel(),
+  pendingPushNotification: getPendingPushNotificationModel(),
 
   // for testing only. noop otherwise.
   __inject: __TEST__
--- a/src/lib/store/PendingPushNotificationModel.ts
+++ b/src/lib/store/PendingPushNotificationModel.ts
@@ -1,0 +1,24 @@
+import { action, Action } from "easy-peasy"
+import { ReceivedNotification } from "react-native-push-notification"
+
+export interface PendingPushNotification extends Omit<ReceivedNotification, "userInfo" | "finish"> {
+  tappedAt: number
+  finish?: () => {}
+}
+
+export interface PendingPushNotificationModel {
+  ios: PendingPushNotification | null
+  android: PendingPushNotification | null
+  setPendingPushNotification: Action<
+    this,
+    { platform: "ios" | "android"; notification: PendingPushNotification | null }
+  >
+}
+
+export const getPendingPushNotificationModel = (): PendingPushNotificationModel => ({
+  ios: null,
+  android: null,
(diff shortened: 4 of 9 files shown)


ArtsyOpenSource:

- Android user-facing changes (Handles the display and proper routing of users based on custom data in push notification payload.)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against a3627930f82ee257d598aa27054cb6d33957ef34
</p>

src/lib/store/GlobalStore.tsx
```diff
@@ -120,7 +120,7 @@ export function useSelectedTab(): BottomTabType {
 }
 
 let _globalStoreInstance: ReturnType<typeof createGlobalStore> | undefined
-const globalStoreInstance = (): ReturnType<typeof createGlobalStore> => {
+export const globalStoreInstance = (): ReturnType<typeof createGlobalStore> => {
```
pvinis, line 123: I don't think we should export this here. We have wrappers to avoid this. If they are enough for this PR, then try that way. Something like `GlobalStore.useAppState` and `GlobalStore.actions`

pvinis:
Also I'm pretty sure this needs a migration, since we are changing the store structure.

src/lib/store/GlobalStore.tsx
lordkiz: Sure. But we can't use hooks in PushNotifications.ts

src/lib/utils/PushNotification.ts
```diff
@@ -1,7 +1,75 @@
...
+
+export const handlePendingNotification = (notification: PendingPushNotification) => {
+  const elapsedTimeInSecs = Math.floor((Date.now() - notification.tappedAt) / 1000)
+  if (elapsedTimeInSecs <= MAX_ELAPSED_TAPPED_NOTIFICATION_TIME && !!notification.data.url) {
+    navigate(notification.data.url, { passProps: { ...notification.data, url: undefined } })
+  }
+  globalStoreInstance()
+    .getActions()
+    .pendingPushNotification.setPendingPushNotification({ platform: "android", notification: null })
+}
```
pvinis, line 18: Basically here. I think `GlobalStore.actions.pendingPushNotif..` should work instead.

lordkiz: Absolutely! Thanks

src/lib/AndroidApp.tsx
```diff
@@ -85,6 +87,14 @@ const Main: React.FC<{}> = track()(({}) => {
     }
   }, [isHydrated])
 
+  useEffect(() => {
+    if (isLoggedIn) {
+      if (pendingNotification) {
+        handlePendingNotification(pendingNotification)
+      }
+    }
```
olerichter00, line 95: **(suggestion):** This might be easier to read.
```suggestion
    if (isLoggedIn && pendingNotification) {
      handlePendingNotification(pendingNotification)
    }
```

src/lib/utils/PushNotification.ts
```diff
@@ -1,7 +1,75 @@
...
+import { PendingPushNotification } from "lib/store/PendingPushNotificationModel"
+import PushNotification, { ReceivedNotification } from "react-native-push-notification"
 import { ASYNC_STORAGE_PUSH_NOTIFICATIONS_KEY } from "./AdminMenu"
 
+const MAX_ELAPSED_TAPPED_NOTIFICATION_TIME = 90 // seconds
+
+export const handlePendingNotification = (notification: PendingPushNotification) => {
+  const elapsedTimeInSecs = Math.floor((Date.now() - notification.tappedAt) / 1000)
+  if (elapsedTimeInSecs <= MAX_ELAPSED_TAPPED_NOTIFICATION_TIME && !!notification.data.url) {
+    navigate(notification.data.url, { passProps: { ...notification.data, url: undefined } })
```
olerichter00, line 13: No need to set `url` to undefined:

```suggestion
    navigate(notification.data.url, { passProps: { ...notification.data } })
```

src/lib/utils/PushNotification.ts
```diff
@@ -1,7 +1,75 @@
...
+
+export const handleReceivedNotification = (notification: Omit<ReceivedNotification, "userInfo">) => {
+  if (__DEV__ && !__TEST__) {
+    console.log("RECIEVED NOTIFICATION", notification)
+  }
+  const isLoggedIn = !!globalStoreInstance().getState().auth.userAccessToken
+  if (notification.userInteraction) {
+    const hasUrl = !!notification.data.url
+    if (isLoggedIn && hasUrl) {
+      navigate(notification.data.url as string, { passProps: { ...notification.data, url: undefined } })
```
olerichter00, line 28:
```suggestion
      navigate(notification.data.url as string, { passProps: { ...notification.data } })
```

src/lib/AndroidApp.tsx
lordkiz: Well, there are still other stuffs to go into that block from other PRs. such as:
```
if (isLoggedIn) {
    savePendingToken()
}
```

olerichter00: 👍

src/lib/utils/PushNotification.ts
lordkiz: I wouldn't want to pass `url` as prop to any component unintentionally.

src/lib/utils/PushNotification.ts
lordkiz: Same as above. The url prop is such a defacto prop for rerouting that I would not want to pass it to any component unintentionally.

src/lib/utils/PushNotification.ts
olerichter00: Ah I see, so you intentionally unsetting it here, right? That's a pretty good idea 👍

olerichter00: Thanks for the clarification :)

src/lib/store/GlobalStore.tsx
lordkiz: @pvinis I actually need to access store state outside of a react component. What is wrong with calling a singleton instance directly like this? I've always thought it safe.

pvinis (approved):
I'll approve. Code seems 👌👌, but I have not run this PR on my device to test it.

lordkiz:
Thanks @pvinis . Undid export of globalStoreInstance. + added handling initial notification on login (handy for when initial notification fails to pop automatically)

src/lib/AndroidApp.tsx
```diff
@@ -85,6 +89,26 @@ const Main: React.FC<{}> = track()(({}) => {
     }
   }, [isHydrated])
 
+  useEffect(() => {
+    if (isLoggedIn) {
+      if (!hasHandledInitialNotification) {
```
olerichter00, line 94: **Idea:** Just want to mention the [Early Return Pattern](https://szymonkrajewski.pl/why-should-you-return-early/) here to prevent nested if/else structures. Not sure though if we prefer one or the other in our code base.

olerichter00 (approved):
Looks very good 👌👌👌

src/lib/AndroidApp.tsx
lordkiz: I actually prefer the early return pattern. I tend to use a mix of both depending on what I deem easily understandable. I feel having something like this is just a pain to read
```
              /////
             return
          }
          handlePendingNotification(pendingNotification)
        })
        return
      }
      handlePendingNotification(pendingNotification)
  }
```

src/lib/store/GlobalStoreModel.ts
```diff
@@ -93,6 +95,7 @@ export const getGlobalStoreModel = (): GlobalStoreModel => ({
   config: getConfigModel(),
   auth: getAuthModel(),
   toast: getToastModel(),
+  pendingPushNotification: getPendingPushNotificationModel(),
```
MounirDhahri, line 98: **thought:** I think it makes more sense to name `pendingPushNotifications`

src/lib/store/PendingPushNotificationModel.ts
```diff
@@ -0,0 +1,24 @@
...
+  android: PendingPushNotification | null
+  setPendingPushNotification: Action<
+    this,
+    { platform: "ios" | "android"; notification: PendingPushNotification | null }
+  >
+}
+
+export const getPendingPushNotificationModel = (): PendingPushNotificationModel => ({
+  ios: null,
+  android: null,
```
MounirDhahri, line 20: **question:** Is there a reason why we have separate push notifications per platform. I though ideally the goal is to unify both

src/lib/utils/PushNotification.ts
MounirDhahri: good thinking but I don't think that would cause problems since I don't see [navigate](https://github.com/artsy/eigen/blob/ea6754df375ffe4d8cc180fb08781602ddba93d6/src/lib/navigation/navigate.ts) using it anywhere. Is it being used somewhere maybe where I am not aware of it?

src/lib/utils/PushNotification.ts
MounirDhahri: same comment here

src/lib/utils/PushNotification.ts
```diff
@@ -1,7 +1,82 @@
...
+    if (!isLoggedIn) {
+      // removing finish because we do not use it on android and we don't want to serialise functions at this time
+      const newNotification = { ...notification, finish: undefined, tappedAt: Date.now() }
+      delete newNotification.finish
+      GlobalStore.actions.pendingPushNotification.setPendingPushNotification({
+        platform: "android",
+        notification: newNotification,
+      })
+      return
+    }
```
MounirDhahri, line 46: **thought:** I think it would have made more sense to have this before `if (isLoggedIn)`

src/lib/utils/PushNotification.ts
```diff
@@ -1,7 +1,82 @@
...
+    return
+  }
+  if (notification.foreground) {
+    // flash notification and put it in Tray
+    const typedNotification: Omit<ReceivedNotification, "userInfo"> & { title?: string } = { ...notification }
+    const channelId = "net.artsy.artsy"
```
MounirDhahri, line 52: **issue(non-blocker):** I think it would make more sense to define this at the beginning of the file with an uppercase maybe

src/lib/utils/PushNotification.ts
```diff
@@ -1,7 +1,82 @@
...
+          onlyAlertOnce: false, // (optional) alert will open only once with sound and notify, default: false
+          userInfo: notification.data,
+
+          /* iOS and Android properties */
+          id: 0,
+          message: typedNotification.title ?? "Artsy", // (required)
+        })
+      }
+    )
+  }
```
MounirDhahri, line 77: **issue:(blocker)** I think this is not the right place where we should be creating the notifications channel as it will be created every time we interact with a push notification and not the recommended approach from the docs [here](https://github.com/zo0r/react-native-push-notification#channel-management-android). We should probably move this to AndroidApp when the app first starts

src/lib/utils/PushNotification.ts
```diff
@@ -1,7 +1,82 @@
...
+    return
+  }
+  if (notification.foreground) {
+    // flash notification and put it in Tray
+    const typedNotification: Omit<ReceivedNotification, "userInfo"> & { title?: string } = { ...notification }
+    const channelId = "net.artsy.artsy"
+    PushNotification.createChannel(
+      {
+        channelId, // (required)
+        channelName: "default", // (required)
```
MounirDhahri, line 56: I added a ticket to investigate more about push notifications channels [here](https://artsyproduct.atlassian.net/browse/CX-1788). Feel free to watch its progress

src/lib/utils/PushNotification.ts
```diff
@@ -1,7 +1,82 @@
...
+          /* Android Only Properties */
+          channelId, // (required) channelId, if the channel doesn't exist, notification will not trigger.
+          subText: notification.subText,
+          ignoreInForeground: false, // (optional) if true, the notification will not be visible when the app is in the foreground (useful for parity with how iOS notifications appear). should be used in combine with `com.dieam.reactnativepushnotification.notification_foreground` setting
+          onlyAlertOnce: false, // (optional) alert will open only once with sound and notify, default: false
+          userInfo: notification.data,
+
+          /* iOS and Android properties */
+          id: 0,
+          message: typedNotification.title ?? "Artsy", // (required)
```
MounirDhahri, line 73: **issue(blocker):** I think this should be the title instead of the message and the message will be `typedNotification.message`. See [here](https://github.com/zo0r/react-native-push-notification#local-notifications) for examples

src/lib/utils/PushNotification.ts
```diff
@@ -1,7 +1,82 @@
...
+      delete newNotification.finish
+      GlobalStore.actions.pendingPushNotification.setPendingPushNotification({
+        platform: "android",
+        notification: newNotification,
+      })
+      return
+    }
+    return
+  }
+  if (notification.foreground) {
```
MounirDhahri, line 49: **praise:** 🙌 Thanks for covering this case. 
Would you mind if you add a comment here saying something like: In order to have a consistent behaviour in Android & iOS with the most flexibility, it is best to handle it manually by prompting a local notification when `onNotification` is triggered by a remote push notification on foreground

src/lib/utils/PushNotification.ts
```diff
@@ -1,7 +1,82 @@
...
+}
+
+export const handleReceivedNotification = (notification: Omit<ReceivedNotification, "userInfo">) => {
+  if (__DEV__ && !__TEST__) {
+    console.log("RECIEVED NOTIFICATION", notification)
+  }
+  const isLoggedIn = !!unsafe_getUserAccessToken()
+  if (notification.userInteraction) {
+    const hasUrl = !!notification.data.url
+    if (isLoggedIn && hasUrl) {
```
MounirDhahri, line 28: **thought:** In case of the `isLoggedIn && !hasUrl` shouldn't we also `setPendingPushNotification({
        platform: "android",
        notification: null,
      })` ?

MounirDhahri (commented):
Nice work here! I left few comments that can be addressed in a different PR. I apologize for the late review

src/lib/AndroidApp.tsx
```diff
@@ -85,6 +89,26 @@ const Main: React.FC<{}> = track()(({}) => {
...
+            return
+          }
+          handlePendingNotification(pendingNotification)
+        })
+      } else {
+        handlePendingNotification(pendingNotification)
+      }
+    }
+  }, [isLoggedIn])
+
```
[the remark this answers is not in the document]
lordkiz: Super! That will be better!

src/lib/store/GlobalStoreModel.ts
lordkiz: This will only hold one notification at a time per platform. If pluralised, it might convey wrongly that there can be more than one notification there.

src/lib/store/PendingPushNotificationModel.ts
lordkiz: There are lots of properties you might find in iOS's that isn't in android's. But we can unify them and put in checks when accessing the properties. Another reason I started with separating them is because I did not want to serialise `finish function` since it isn't necessary for android now. So using pendingPushNotification.android one need not worry about calling finish. Not separating them might be better.

src/lib/utils/PushNotification.ts
lordkiz: @MounirDhahri navigate will pass it as a prop to the next component. I was just afraid that it is easy for any Dev to just choose url as a prop name (without thinking twice) to pass whatever urls including external urls that might pop open webviews. If this notification lands on such, we'd have a very weird bug.

src/lib/utils/PushNotification.ts
lordkiz: There is no point because we have not set it in the first place

src/lib/utils/PushNotification.ts
lordkiz: I see. It would be more understandable, yes?

src/lib/utils/PushNotification.ts
lordkiz: It would appear so, but message is used by PushNotification.localNotification for both ios and android and it is used in place of title when none is defined. However, fcm notifications will not have message property but instead have title property. In this case typedNotification will not have message property.

src/lib/utils/PushNotification.ts
MounirDhahri: Yes, I understand that's true but I don't believe it will be triggering opening webviews since the navigate is based on the first param (the URL param) and does not take in count merging both params (URL and passProps). This is not a blocker at all but It's something I don't see as necessary, we can of course leave it as it is with no issues

src/lib/store/GlobalStoreModel.ts
MounirDhahri: ah it's just one notification, my bad. makes sense then

src/lib/utils/PushNotification.ts
MounirDhahri: It's really about being more understandable, but as a pattern, usually starting by the non-common case and short-circuiting from there is advised. it's a personal preference though

src/lib/utils/PushNotification.ts
MounirDhahri: I see now! thanks for explaining. Is there a way though for us to maybe get the message assuming that we already received it?

src/lib/utils/PushNotification.ts
MounirDhahri: clear! thanks for explaining

src/lib/store/PendingPushNotificationModel.ts
MounirDhahri: I understand that the push notification payload is different for both platforms but I don't see the harm in that since it will be still coherent within the same platform and it would allow us to unify where push notifications leave in the future. 

> Another reason I started with separating them is because I did not want to serialise finish function since it isn't necessary for android now.
I get that but since `notification.finish` will be called in only one block (`onNotification` method at the configuration step), I feel we could have just made a platform check and included it there. A platform check won't even be necessary maybe since it has no effect on Android

PR #5230, opened 2021-08-06
feature(CX-1796): Introduce AboveTheFoldQueryRenderer to Home screen

olerichter00:

This PR resolves [CX-1796]

### Description

This PR splits the home screen query into a below-the-fold and an above-the-fold query to improve performance.

All modules below "Market News" will be loaded with a delay via the below-the-fold query. Like this, even a fresh account without any personalized modules will see at least two modules that cover the whole screen as soon as the above-the-fold query loads.

### Performance Improvements

**First Query:**

Screenshot 2021-08-06 at 16 43 59

**Second Query:**
Screenshot 2021-08-06 at 16 33 42

## Screenshots

**below-the-fold placeholder:**

Simulator Screen Shot - iPhone 8 - 2021-08-18 at 10 11 27

- Introduce AboveTheFoldQueryRenderer to Home screen - ole

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-1796]: https://artsyproduct.atlassian.net/browse/CX-1796


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 67c1e79e0b7d408c1418cbaabc3625f76c4ab060
</p>

src/lib/store/config/EnvironmentModel.tsx
```diff
@@ -32,7 +32,7 @@ export const environment = defineEnvironmentOptions({
     description: "Metaphysics URL",
     presets: {
       local: "http://localhost:3000/v2",
-      staging: "https://metaphysics-staging.artsy.net/v2",
+      staging: "http://localhost:3000/v2",
```
olerichter00, line 35:
```suggestion
      staging: "https://metaphysics-staging.artsy.net/v2",
```

src/lib/Scenes/Home/Home.tsx
```diff
@@ -14,34 +17,44 @@ import { FairsRailFragmentContainer } from "lib/Scenes/Home/Components/FairsRail
...
 import { RailScrollRef } from "./Components/types"
 
 interface Props extends ViewProps {
-  articlesConnection: Home_articlesConnection
-  homePage: Home_homePage
-  me: Home_me
-  featured: Home_featured
+  articlesConnection: Home_articlesConnection | null
+  homePage: Home_homePage | null
+  homePageAbove: Home_homePageAbove | null
```
olerichter00, line 43: **Note:** Props without prefix are from the `below` query.

src/lib/Scenes/Home/Home.tsx
```diff
@@ -14,34 +17,44 @@ import { FairsRailFragmentContainer } from "lib/Scenes/Home/Components/FairsRail
...
+  loading: boolean
 }
 
 const Home = (props: Props) => {
-  const { articlesConnection, homePage, me, featured } = props
-  const artworkModules = homePage?.artworkModules || []
-  const salesModule = homePage?.salesModule
+  const { articlesConnection, homePage, homePageAbove, me, meAbove, featured, loading } = props
+
+  const artworkModules = (homePageAbove?.artworkModules || []).concat(homePage?.artworkModules || [])
```
olerichter00, line 54: **Note:** Concatenating artwork modules from the above and below the fold query here.

src/lib/Scenes/Home/Home.tsx
```diff
@@ -189,15 +215,60 @@ const Home = (props: Props) => {
...
+            GENERIC_GENES
+            LIVE_AUCTIONS
+            CURRENT_FAIRS
+            RELATED_ARTISTS
+            FOLLOWED_GENES
+            RECOMMENDED_WORKS
+            FOLLOWED_GALLERIES
+            SIMILAR_TO_SAVED_WORKS
+            SIMILAR_TO_RECENTLY_VIEWED
+          ]
```
olerichter00, line 240: **Note:** We should use something like `include` to make sure to not load a module in both queries.

src/lib/Scenes/Home/Components/FairsRail.tsx
```diff
@@ -41,6 +41,10 @@ const FairsRail: React.FC<Props & RailScrollProps> = (props) => {
     </Flex>
   )
 
+  if (!props.fairsModule?.results) {
+    return null
+  }
```
MounirDhahri, line 46: **thought:** shouldn't this be `props.fairsModule?.results.length`

src/lib/utils/AboveTheFoldQueryRenderer.tsx
```diff
@@ -75,6 +78,15 @@ export function AboveTheFoldQueryRenderer<AboveQuery extends OperationType, Belo
...
   }
 
+  const [renderBelowTheFold, setRenderBelowTheFold] = useState(false)
+
+  useEffect(() => {
+    if (props.belowTheFoldTimeout === undefined) {
+      return
+    }
+    setImmediate(() => setRenderBelowTheFold(true), props.belowTheFoldTimeout)
+  }, [])
```
MounirDhahri, line 88: **nitpick:** I think we can also return `clearImmediate` here to clear the pending immediate actions when the component unmounts

src/lib/utils/AboveTheFoldQueryRenderer.tsx
```diff
@@ -25,6 +25,9 @@ interface AboveTheFoldQueryRendererProps<AboveQuery extends OperationType, Below
         renderPlaceholder: () => React.ReactChild
       }
   cacheConfig?: CacheConfig | null
+  /** Fire below-the-fold query after the given timeout or when the above-the-fold query returns. */
+  /** If 'belowTheFoldTimeout' is not set, the below-the-fold query will be fired when the above-the-fold query returns */
+  belowTheFoldTimeout?: number
```
MounirDhahri, line 30: **thought:** what's the reasoning behind this props `belowTheFoldTimeout`. I don't see you using it anywhere. Is there a problem with running both queries at the same time?

MounirDhahri (commented):
This will be a great improvement! looking forward to getting this out!

src/lib/Scenes/Home/Components/FairsRail.tsx
olerichter00: That's actually a good idea. Just checking for `props.fairsModule?.results` will render an empty `FlatList` without a header but `props.fairsModule?.results?.length` seems more clear. Was that your point?

src/lib/utils/AboveTheFoldQueryRenderer.tsx
olerichter00: Good catch - thanks! We definitely want to return `clearImmediate`!

src/lib/utils/AboveTheFoldQueryRenderer.tsx
olerichter00: It's used [here](https://github.com/artsy/eigen/blob/03de5cc946bce649c3b283088bd90265311930c1/src/lib/Scenes/Home/Home.tsx#L539) in the Home componen. I introduced the timeout to speed up the below-the-fold query because it can take up to 7 or 8 seconds. Before it was fired after the above-the-fold query returned. Do you think it's a good idea?

src/lib/utils/AboveTheFoldQueryRenderer.tsx
olerichter00: Done

src/lib/Scenes/Home/Components/FairsRail.tsx
olerichter00: Done ✅

olerichter00:
> This will be a great improvement! looking forward to getting this out!

Yeah, I can't wait to see how this affects the average load time for real users 🤩

src/lib/utils/AboveTheFoldQueryRenderer.tsx
MounirDhahri: I got what you mean now, Is there a way maybe we can trick this to make both queries execute at the same time? I mean something like this without any delay
Untitled (4)

olerichter00: To make both queries at the same time we can set `belowTheFoldTimeout` to `0` on the Home screen. I've set it to `100` (milliseconds) to not let both queries interfere and to make sure the above-the-fold query returns as soon as possible. This 100 ms shouldn't be noticeable by the user.

MounirDhahri: Oki, after a short call, I got what you mean now and this makes sense. thanks

MounirDhahri (approved):

PR #5232, opened 2021-08-09
feat(cx-1768): fix concerns raised in handling of push notifications

lordkiz:

This PR resolves [CX-1768]

### Description
Fixes general concerns raised regarding handling of push notifications in [this PR](https://github.com/artsy/eigen/pull/5208)
- create channels on app start
- stop unsetting url when passing push notif props (as eigen rerouting logic is not based on url prop, thus making the url unsetting confusing to devs)
- unify android & ios pending notification objects

-

#### iOS user-facing changes

-

-

#### Dev changes

- Update general android push notification handling logic - kizito

</details>

[CX-1768]: https://artsyproduct.atlassian.net/browse/CX-1768

### DIFF
--- a/src/lib/AndroidApp.tsx
+++ b/src/lib/AndroidApp.tsx
@@ -18,7 +18,6 @@ import { ModalStack } from "./navigation/ModalStack"
 import { BottomTabsNavigator } from "./Scenes/BottomTabs/BottomTabsNavigator"
 import { ForceUpdate } from "./Scenes/ForceUpdate/ForceUpdate"
 import { Onboarding } from "./Scenes/Onboarding/Onboarding"
-import { handlePendingNotification, handleReceivedNotification } from "./utils/PushNotification"
 import { ConsoleTrackingProvider } from "./utils/track/ConsoleTrackingProvider"
 import { AnalyticsConstants } from "./utils/track/constants"
 
@@ -35,12 +34,10 @@ const Main: React.FC<{}> = track()(({}) => {
       webClientId: "673710093763-hbj813nj4h3h183c4ildmu8vvqc0ek4h.apps.googleusercontent.com",
     })
   }, [])
-  const [hasHandledInitialNotification, setHasHandledInitialNotification] = useState(false)
   const isHydrated = GlobalStore.useAppState((state) => state.sessionState.isHydrated)
   const isLoggedIn = GlobalStore.useAppState((state) => !!state.auth.userAccessToken)
   const onboardingState = GlobalStore.useAppState((state) => state.auth.onboardingState)
   const forceUpdateMessage = GlobalStore.useAppState((state) => state.config.echo.forceUpdateMessage)
-  const pendingNotification = GlobalStore.useAppState((state) => state.pendingPushNotification.android)
 
   useSentryConfig()
   useStripeConfig()
@@ -88,26 +85,6 @@ const Main: React.FC<{}> = track()(({}) => {
       }, 500)
     }
   }, [isHydrated])
-
-  useEffect(() => {
-    if (isLoggedIn) {
-      if (!hasHandledInitialNotification) {
-        // initial notification is most recent and should be prioritised
-        PushNotification.popInitialNotification((notification) => {
-          if (notification) {
-            handleReceivedNotification(notification)
-            // prevent loop where user logs out and logs back in and previously
-            // handled initial push notification is rehandled
-            setHasHandledInitialNotification(true)
-            return
-          }
-          handlePendingNotification(pendingNotification)
-        })
-      } else {
-        handlePendingNotification(pendingNotification)
-      }
-    }
-  }, [isLoggedIn])
 
   if (!isHydrated) {
     return <View />
--- a/src/lib/AndroidApp.tsx
+++ b/src/lib/AndroidApp.tsx
@@ -20,6 +20,7 @@ import { Onboarding } from "./Scenes/Onboarding/Onboarding"
 import { createAllChannels } from "./utils/PushNotification"
 import { ConsoleTrackingProvider } from "./utils/track/ConsoleTrackingProvider"
 import { AnalyticsConstants } from "./utils/track/constants"
+import { useInitialNotification } from "./utils/useInitialNotification"
 
 addTrackingProvider(SEGMENT_TRACKING_PROVIDER, SegmentTrackingProvider)
 addTrackingProvider("console", ConsoleTrackingProvider)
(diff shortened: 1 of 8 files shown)


ArtsyOpenSource:

- Dev changes (Update general android push notification handling logic - kizito)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 84b40d13230bc2f3f0045df9217e550fd10b603b
</p>

src/lib/utils/PushNotification.ts
```diff
@@ -7,15 +7,70 @@ import { ASYNC_STORAGE_PUSH_NOTIFICATIONS_KEY } from "./AdminMenu"
...
+      userInfo: notification.data,
+
+      /* iOS and Android properties */
+      id: 0,
+      message: notification.title ?? "Artsy", // (required)
+    })
+  }
+  if (CHANNELS.some((channel) => channel.id === channelId)) {
+    // we can safely assume that these channels were created on App start
+    create()
```
olerichter00, line 58: Is there another way to check if the channel was created? Assuming that creating a channel could fail on the app start, this check wouldn't detect it.

olerichter00 (commented):
Looks great 🎉

src/lib/AndroidApp.tsx
```diff
@@ -35,19 +35,19 @@ const Main: React.FC<{}> = track()(({}) => {
...
-  const pendingNotification = GlobalStore.useAppState((state) => state.pendingPushNotification.android)
 
   useSentryConfig()
   useStripeConfig()
   useWebViewCookies()
   useDeepLinks()
+  useInitialNotification()
 
   useEffect(() => {
+    createAllChannels()
```
MounirDhahri, line 50: **issue(blocker)**: I think we should move this to call to its own `useEffect` to make sure it does not get executed again when `isHydrated` changes

src/lib/AndroidApp.tsx
```diff
@@ -35,19 +35,19 @@ const Main: React.FC<{}> = track()(({}) => {
...
   const isLoggedIn = GlobalStore.useAppState((state) => !!state.auth.userAccessToken)
   const onboardingState = GlobalStore.useAppState((state) => state.auth.onboardingState)
   const forceUpdateMessage = GlobalStore.useAppState((state) => state.config.echo.forceUpdateMessage)
-  const pendingNotification = GlobalStore.useAppState((state) => state.pendingPushNotification.android)
 
   useSentryConfig()
   useStripeConfig()
   useWebViewCookies()
   useDeepLinks()
+  useInitialNotification()
```
MounirDhahri, line 47: **praise:** nice!

src/lib/utils/PushNotification.ts
```diff
@@ -7,15 +7,70 @@ import { ASYNC_STORAGE_PUSH_NOTIFICATIONS_KEY } from "./AdminMenu"
 
 const MAX_ELAPSED_TAPPED_NOTIFICATION_TIME = 90 // seconds
 
+export const CHANNELS = [
+  { name: "net.artsy.artsy", id: "net.artsy.artsy", properties: {} },
```
MounirDhahri, line 11: **thought:** since the channel names will be visible in the notifications settings of the app. I think we should probably adjust the channel name to be in a more readable format. Maybe "Default" or something like this.
93714fe-android_o_channel_list_example

src/lib/utils/PushNotification.ts
```diff
@@ -7,15 +7,70 @@ import { ASYNC_STORAGE_PUSH_NOTIFICATIONS_KEY } from "./AdminMenu"
...
+      /* iOS and Android properties */
+      id: 0,
+      message: notification.title ?? "Artsy", // (required)
+    })
+  }
+  if (CHANNELS.some((channel) => channel.id === channelId)) {
+    // we can safely assume that these channels were created on App start
+    create()
+  } else {
+    createChannel(channelId, channelName)
```
MounirDhahri, line 60: **issue:** There is no reason to be creating the channel here when we receive the local push notification since we already created the channel at the app startup 👍

src/lib/utils/PushNotification.ts
```diff
@@ -24,56 +79,29 @@ export const handleReceivedNotification = (notification: Omit<ReceivedNotificati
...
-          id: 0,
-          message: typedNotification.title ?? "Artsy", // (required)
-        })
-      }
-    )
+    // In order to have a consistent behaviour in Android & iOS with the most flexibility,
+    // it is best to handle it manually by prompting a local notification when onNotification
+    // is triggered by a remote push notification on foreground
+    const typedNotification: TypedNotification = { ...notification }
+    createLocalNotification(typedNotification)
```
MounirDhahri, line 104: **praise:** 🙌

MounirDhahri (commented):
Great work! Thanks for following up! I left few minor comments but overall this should be clear to go

src/lib/utils/PushNotification.ts
```diff
@@ -7,15 +7,70 @@ import { ASYNC_STORAGE_PUSH_NOTIFICATIONS_KEY } from "./AdminMenu"
...
+  PushNotification.createChannel(
+    {
+      channelId,
+      channelName,
+      ...properties,
+    },
+    (created) => {
+      if (created && __DEV__) {
+        console.log(`NEW CHANNEL ${channelName} CREATED`)
+      }
```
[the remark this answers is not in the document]
lordkiz: False just means channel already exists.

src/lib/utils/PushNotification.ts
lordkiz: Good catch! An extra layer of check won't hurt.

src/lib/AndroidApp.tsx
lordkiz: @MounirDhahri this is not in the useEffect that deps on isHydrated

src/lib/utils/PushNotification.ts
lordkiz: Dang! How did I not think of this? 🤦

src/lib/utils/PushNotification.ts
lordkiz: The problem is if suddenly any of our BE services starts sending notifications targeting a new channel that is not in the current list of channels, app users will not need to update their app in order to get them

MounirDhahri: that's a good point. Although I believe that the app must create a channel with this ID before any notification with this key is received and that's how FCM knows how to target them properly. I am not sure about it though, because I remember reading it somewhere in the docs in the past but can't find it now. 

Let's keep the code as you did it now and check it later once we finalize all the work.

MounirDhahri (approved):

src/lib/utils/PushNotification.ts
lordkiz: @MounirDhahri You are absolutely correct. That is why for channel id that does not exist yet or wasn't created on app start will be created first before creating local notification.

Issue #4549, opened 2021-03-22
[RFC] Add react-native-bootsplash


MounirDhahri:
**Status:** Open

**Type:** New Dependency

**Name: react-native-bootsplash**

**Github URL:** https://github.com/zoontek/react-native-bootsplash/
**Npmjs URL:** https://www.npmjs.com/package/react-native-bootsplash

**Description**
- React-native-bootsplash allows us to show a splash screen during app startup and hide it when you from js level when ready.
- We need it, or something like it, for this PR: https://github.com/artsy/eigen/pull/4548

**Checklist**
- [x] Have read over the source code, and we can maintain it
- [x] Has had a release in the last year, or looks done and stable

**Alternatives**
The main alternative is [react-native-splash-screen](https://github.com/crazycodeboy/react-native-splash-screen)
I went with react-native-bootsplash because:
- It is more maintained (last update on react-native-splash-screen was back in 2019)
-  The latest version of react-native-splash-screen seem to be drawing on top of the main activity instead of having the splash screen as a separate activity.
- I didn't want to have the entire splash screen as an image and wanted instead to be able to easily customize the layout of it.

MounirDhahri:
We will adopt this
[Resolved]

PR #5254, opened 2021-08-11
feat: add improved search input and algolia dependencies

gkartalis:

This PR resolves [FX-3180]

### Description

- Introduces algolia dependencies (algoliasearch, react-instantsearch-native, @types/react-instantsearch-native)
- Add a search bar that doesn't do anything other than being able to change its text on improved search screen

-

#### iOS user-facing changes

-

-

#### Dev changes

- Add algoliasearch and react-instantsearch-native dependencies - gkartalis

</details>


dblandin:
@gkartalis We have a process for "New dependency RFCs" ([example](https://github.com/artsy/eigen/issues/4549)).

I think we should create a new issue to briefly explain why we want to pull these dependencies in, mention alternate approaches, and solicit some feedback from others (particularly the mobile practice).

Some example alternative approaches which we might list:

- Don't use Instantsearch.JS and use the `algolia` api client directly
- Don't query algolia directly but instead expose a GraphQL interface which will proxy requests to Algolia

package.json
```diff
@@ -108,6 +108,7 @@
     "@segment/analytics-react-native-appboy": "^1.5.0",
     "@sentry/react-native": "2.2.1",
     "@styled-system/theme-get": "^5.1.2",
+    "algoliasearch": "^4.10.3",
```
mdole, line 111: **suggest, non-blocking:** It's a little odd to me that this PR introduces new dependencies and then doesn't use them. I don't feel strongly about this, but I wonder if this PR should either ONLY introduce the deps and do nothing else, or if it should both add these deps and then actually put them to use.

mdole (approved):

package.json
gkartalis: Good point! I will introduce these on the next ticket that is going to actually use them

iskounen (approved):

dblandin (approved):
Looks great!! 💯

src/lib/Scenes/Search2/Search2.tsx
```diff
@@ -1,10 +1,168 @@
...
 
-export const Search2: React.FC = () => {
   return (
-    <Flex flexGrow={1} justifyContent="center" alignItems="center">
-      <Text>Improved Search Screen</Text>
-    </Flex>
+    <Sans size="3t" weight="regular">
+      {parsedHit.map(({ isHighlighted, value }) =>
+        isHighlighted ? (
+          <Sans size="3t" weight="medium" style={{ padding: 0, margin: 0 }}>
```
dblandin, line 41: fyi: I believe these components can take `padding` and `margin` directly though I'm a bit surprised that they would have initial non-zero values.

```suggestion
          <Sans size="3t" weight="medium" padding={0} margin={0} >
```

src/lib/Scenes/Search2/Search2.tsx
```diff
@@ -1,10 +1,168 @@
...
+const Highlight = connectHighlight(({ highlight, attribute, hit, highlightProperty = "_highlightResult" }) => {
+  const parsedHit = highlight({ attribute, hit, highlightProperty })
 
-export const Search2: React.FC = () => {
   return (
-    <Flex flexGrow={1} justifyContent="center" alignItems="center">
-      <Text>Improved Search Screen</Text>
-    </Flex>
+    <Sans size="3t" weight="regular">
+      {parsedHit.map(({ isHighlighted, value }) =>
```
dblandin, line 39: **question / non-blocking**: Do we need any `key={index}` assignment here to avoid the react unique key warning?

gkartalis: indeed we do

src/lib/Scenes/Search2/Search2.tsx
```diff
@@ -1,10 +1,168 @@
...
+  if (!searchClient) {
+    // error handling in case appID or apiKey is not set ??
+    return null
+  }
+
+  return (
+    <ArtsyKeyboardAvoidingView>
+      <InstantSearch
+        searchClient={searchClient}
+        indexName="POC_Artists"
```
anandaroop, line 110: This hard-coding presumably gets replaced eventually — is that what you were getting at with [this Slack comment](https://artsy.slack.com/archives/C9SATFLUU/p1629122073019000?thread_ts=1629120615.018100&cid=C9SATFLUU) about serving indices' metadata from MP too, @dblandin?

anandaroop (approved):
Looks amazing, nice work @gkartalis ! 🤩 

**question, non-blocking**: One general thought about naming — I wonder if the `ImprovedFoo` convention will age well. At some point this just becomes *the* implementation instead of the the _improved_ one. Should the names be simplified to reflect that?

~Improved~SearchInput
~Improved~SearchResults
etc…

src/lib/Scenes/Search2/Search2.tsx
dblandin: >  is that what you were getting at with this Slack comment about serving indices' metadata from MP too

Exactly 👍🏼 I think we'll want to serve these index names/keys from Metaphysics instead of hardcoding them within Eigen.

src/lib/Scenes/Search2/Search2.tsx
```diff
@@ -1,10 +1,168 @@
...
+
+const Highlight = connectHighlight(({ highlight, attribute, hit, highlightProperty = "_highlightResult" }) => {
+  const parsedHit = highlight({ attribute, hit, highlightProperty })
 
-export const Search2: React.FC = () => {
   return (
-    <Flex flexGrow={1} justifyContent="center" alignItems="center">
-      <Text>Improved Search Screen</Text>
-    </Flex>
+    <Sans size="3t" weight="regular">
```
mdole, line 38: **minor, non-blocking:** we currently prefer `<Text>` to `<Sans>`. we used to have both Sans and Serif text components, but some time ago the decision was made to only use sans-serif fonts going forward so a single `<Text>` component was created that should be used for all text.

src/lib/Scenes/Search2/Search2.tsx
```diff
@@ -1,10 +1,168 @@
...
+          <Sans key={index} size="3t" weight="regular">
+            {value}
+          </Sans>
+        )
+      )}
+    </Sans>
+  )
+})
+
+const ImprovedSearchResults: React.FC<{ hits: any }> = ({ hits }) => {
```
mdole, line 54: **non-blocking:** I know this PR is mostly proof-of concept, but I would still encourage avoiding `any` types because it's always easier to change at the very beginning than to come back to later and make explicit types 🙂 and hopefully, any time invested in typing right now will actually speed up development in the future!

mdole (approved):
Awesome work @gkartalis!

mdole:
> Looks amazing, nice work @gkartalis ! 🤩
> 
> **question, non-blocking**: One general thought about naming — I wonder if the `ImprovedFoo` convention will age well. At some point this just becomes _the_ implementation instead of the the _improved_ one. Should the names be simplified to reflect that?
> 
> ~Improved~SearchInput
> ~Improved~SearchResults
> etc…

+1 to this - great example is Artsy's use of "New Selling Options" (AKA BNMO) for like, oh, 3 years after buy now and make offer were launched 😀 

If we want to have some way of differentiating them from the old names, we could consider `AlgoliaSearchInput` etc. I think that would age better but it still tells us very clearly that this is not the old components.

src/lib/Scenes/Search2/Search2.tsx
gkartalis: Thanks for pointing that out! I just copied the old highlight components 😅 will change em to be `<Text />`

src/lib/Scenes/Search2/Search2.tsx
gkartalis: Great point, addressed that on a new commit! Let me know what you think 🙂

src/lib/Scenes/Search2/Search2.tsx
```diff
@@ -1,10 +1,183 @@
...
+          </Text>
+        ) : (
+          <Text key={index}>{value}</Text>
+        )
+      )}
+    </Text>
   )
+})
+
+interface Artist {
```
mdole, line 52: @gkartalis one more small suggestion, sorry! Let's make this something like `SearchResultArtist` or `ArtistForSearch`, just so that future devs searching around the code base know this isn't meant to be a type for every place we refer to artists :)

PR #5280, opened 2021-08-13
fix: migrate current auth to ts

MounirDhahri:

### Description

- The new onboarding flow lives currently on the react-native side of the world and gets stuff from the `authModel`. By enabling the new flag on users devices, they will be signed off since the `authModel` is not populated for them. 

This PR does two things:
- Migrate the `nativeState` to the `authModel` to make sure users do no get signed off when they get the new flag.
- Populate the `authModel` already when the users log in to make sure both states are filled

-

#### iOS user-facing changes

-

-

#### Dev changes

- Migrate nativeState to authModel - mounir

</details>


ArtsyOpenSource:

- Dev changes (Migrate nativeState to authModel - mounir)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 66927d6a5619990762dc1e1a0667610f6cd00fb4
</p>

src/lib/store/NativeModel.ts
```diff
@@ -88,8 +88,19 @@ listenToNativeEvents((event: NativeEvent) => {
         }
       }
       GlobalStore.actions.native.setLocalState(event.payload)
-      if (!newOnboardingFlow && event.payload.userEmail !== null) {
-        GlobalStore.actions.auth.setState({ userEmail: event.payload.userEmail })
+      // We need to set the values we get from the native state on iOS to the global store
+      // to have parity between the auth on native and react-native
+      if (
+        event.payload.userEmail &&
+        event.payload.onboardingState &&
```
olerichter00, line 95: **(question):** Just to understand things better, why do we check for `event.payload.onboardingState` here?

src/lib/store/__tests__/migration-tests.ts
```diff
@@ -311,3 +311,22 @@ describe("PendingPushNotification migration", () => {
     expect(migratedState.pendingPushNotification).toEqual({ notification: null })
   })
 })
+
+describe("CopyIOSNativeSessionAuthToTS migration", () => {
+  const migrationToTest = Versions.CopyIOSNativeSessionAuthToTS
+  it.only("populates authentication details into the auth model", () => {
```
olerichter00, line 317: 👀 

```suggestion
  it("populates authentication details into the auth model", () => {
```

olerichter00 (approved):
LGTM 🚀

src/lib/store/NativeModel.ts
MounirDhahri: this is not needed actually because it will be always a truthy string it can be removed without affecting anything

src/lib/store/__tests__/migration-tests.ts
```diff
@@ -311,3 +330,31 @@ describe("PendingPushNotification migration", () => {
     expect(migratedState.pendingPushNotification).toEqual({ notification: null })
   })
 })
+
+describe("CopyIOSNativeSessionAuthToTS migration", () => {
+  beforeAll(() => {
```
pvinis, line 335: I would suggest not doing it this way, since this messes with stuff in previous migrations when going from v0 to latest in this test.

it's better and safer to do it like some others tests in this file, doing to migration version -1, then injecting your stuff to the legacy native module, then migrate to the latest version.

what do you think?

src/lib/store/__tests__/migration-tests.ts
```diff
@@ -5,6 +5,25 @@ import { __globalStoreTestUtils__ } from "../GlobalStore"
 import { CURRENT_APP_VERSION, migrate, Versions } from "../migration"
 import { sanitize } from "../persistence"
 
+jest.mock("lib/NativeModules/LegacyNativeModules", () => ({
```
pvinis, line 8: related to the other comment, mocking in the whole file could lead to weird problems too, with other migrations expecting different things.

pvinis (changes requested):

Artsy/Networking/API_Modules/ARUserManager.m
```diff
@@ -463,6 +463,27 @@ - (void)loginWithAppleUID:(NSString *)appleUID
     [op start];
 }
 
+// When the user authenticates from react-native, we need to make sure that native side
+// is synchronized with react-native
+- (void)handleAuthState:(NSString *)token
```
MounirDhahri, line 468: **note:** this is almost exactly what we do on the native side when the user authenticates and it's how we populate the auth state in the native side

Artsy/App/ARTArtsyNativeModule.m
```diff
@@ -1,11 +1,25 @@
...
+#import "User.h"
+#import "ARUserManager.h"
 
 
 @implementation ARTArtsyNativeModule
 
 RCT_EXPORT_MODULE(ArtsyNativeModule);
 
+
+RCT_EXPORT_METHOD(updateAuthState:(NSString *) token
```
MounirDhahri, line 15: **note:** I am expecting us to still need this in the long term, that's why I added it to **ArtsyNativeModule** instead of the **ARTemporaryModule**

src/lib/AppProviders.tsx
```diff
@@ -11,10 +11,10 @@ import { GlobalStoreProvider } from "./store/GlobalStore"
 import { ProvideScreenDimensions } from "./utils/useScreenDimensions"
 
 export const AppProviders = ({ children }: { children: ReactNode }) => (
-  <RelayEnvironmentProvider environment={defaultEnvironment}>
```
MounirDhahri, line 14: **note:** This has been causing us many of the **unable to load** issues that users have been reporting recently.
In few sensitive areas, we are using the GlobalStore instance inside relay and that's mainly inside our middleware. When the globalstore instance is still not ready yet, calls like this one
https://github.com/artsy/eigen/blob/1254002b9c62dddc6bb264c1adc2d65110b84336/src/lib/relay/middlewares/metaphysicsMiddleware.ts#L92-L92
will return the `authenticationToken` as `undefined`/`null` (same for the `userID` and all other fields) and will make all queries break since no token is available, the user won't be logged out using by the `checkAuthenticationMiddleware` as well since that middleware requires a token to exist https://github.com/artsy/eigen/blob/84f004984b9018ac1502266b7fe3d7400eb9c7ba/src/lib/relay/middlewares/checkAuthenticationMiddleware.ts#L13-L13 
and the user won't be logged out since by the time react starts the reconciliation (that's the name React uses for calculating + applying changes from (re-)renders), and we do the check to if there is a logged in user to render either the personalization flow or the home screen, the global store had already been initiated and there is a token.
https://github.com/artsy/eigen/blob/d2a11fa2207281e146d7569087621a3fa1d6e043/src/lib/AppRegistry.tsx#L432-L432

src/lib/AppRegistry.tsx
```diff
@@ -429,12 +429,9 @@ const Main: React.FC<{}> = track()(({}) => {
...
   const isHydrated = GlobalStore.useAppState((state) => state.sessionState.isHydrated)
-  const isLoggedIn = GlobalStore.useAppState((state) =>
-    showNewOnboarding ? !!state.auth.userID : !!state.native.sessionState.userID
-  )
-  const onboardingState = GlobalStore.useAppState((state) =>
-    showNewOnboarding ? state.auth.onboardingState : state.native.sessionState.onboardingState
-  )
+  const isLoggedIn = GlobalStore.useAppState((store) => store.auth.userAccessToken)
+
+  const onboardingState = GlobalStore.useAppState((state) => state.auth.onboardingState)
```
MounirDhahri, line 434: **note:** this should be safe to do since we already did the migration

src/lib/Scenes/MyProfile/MyCollectionAndSavedWorks.tsx
```diff
@@ -14,33 +17,28 @@ export enum Tab {
   savedWorks = "Saved Works",
 }
 
-interface Props extends ViewProps {
-  initialTab: Tab
-  me: MyProfile_me
-}
-
-export const MyCollectionAndSavedWorks: React.FC<Props> = ({ initialTab = Tab.collection, me }) => {
```
MounirDhahri, line 22: **note:** no need to send the `initialTab` since it's always `Tab.collection` for now. We can update it when we need it

src/lib/Scenes/MyProfile/MyCollectionAndSavedWorks.tsx
```diff
@@ -54,12 +52,36 @@ export const MyProfileHeader: React.FC<{ name: string | null; createdAt: string
...
+  <QueryRenderer<MyCollectionAndSavedWorksQuery>
+    environment={defaultEnvironment}
+    query={graphql`
+      query MyCollectionAndSavedWorksQuery {
+        me @optionalField {
+          ...MyCollectionAndSavedWorks_me
+        }
+      }
+    `}
+    render={renderWithLoadProgress(MyCollectionAndSavedWorksFragmentContainer)}
```
MounirDhahri, line 84: **note:** I didn't need to return a skeleton placeholder here because that's being taking of somewhere else

src/lib/Scenes/MyProfile/MyCollectionAndSavedWorks.tsx
```diff
@@ -54,12 +52,36 @@ export const MyProfileHeader: React.FC<{ name: string | null; createdAt: string
...
+const MyCollectionAndSavedWorksFragmentContainer = createFragmentContainer(MyCollectionAndSavedWorks, {
+  me: graphql`
+    fragment MyCollectionAndSavedWorks_me on Me {
+      name
+      createdAt
+    }
+  `,
+})
+
+export const MyCollectionAndSavedWorksQueryRenderer: React.FC<{}> = ({}) => (
```
MounirDhahri, line 74: **note:** I had to convert `MyCollectionAndSavedWorks` into a `QueryRenderer` since we need the user full name. this was broken because we were only querying for the full name as part of the old profile query but not with the new collection.
This is how it was looking like
Screenshot 2021-10-01 at 18 30 57

src/lib/Scenes/MyProfile/MyProfile.tsx
```diff
@@ -143,27 +142,40 @@ export const MyProfileContainer = createRefetchContainer(
...
-      }
-    `}
-    render={renderWithPlaceholder({
-      Container: MyProfileContainer,
-      renderPlaceholder: () => <MyProfilePlaceholder />,
-    })}
-    variables={{}}
-  />
-)
+export const MyProfileQueryRenderer: React.FC<{}> = ({}) => {
```
MounirDhahri, line 152: **note:** I made this change because there is no need to query for `MyProfile_me` if we are not going to be showing `MyProfile` to the user

src/lib/store/AuthModel.ts
```diff
@@ -239,96 +244,112 @@ export const getAuthModel = (): AuthModel => ({
...
+        method: "POST",
+        headers: {
+          "Content-Type": "application/json",
+        },
+        body,
+      })
 
-    if (result.status === 201) {
-      const { expires_in, access_token } = await result.json()
-      const { id } = await (
```
MounirDhahri, line 290: **note:** I noticed that the id here is always `null` and changed this call to be the same as we are doing on the native side to get the full user details that we are going to need to fill the auth state in the native side.
https://github.com/artsy/eigen/blob/9f4ef5241c9541b328542143398f97cb551f9ca3/Artsy/Networking/API_Modules/ARUserManager.m#L186-L187
https://github.com/artsy/eigen/blob/385e94eb003bd605f205adf34da1bd179ddbe819/Artsy/Networking/ARRouter.m#L384-L386
https://github.com/artsy/eigen/blob/ab66a301fefbeac4c98922925da1fa48ed053ddb/Artsy/Networking/ARNetworkConstants.m#L23-L23

src/lib/store/GlobalStoreModel.ts
```diff
@@ -77,10 +73,13 @@ export const getGlobalStoreModel = (): GlobalStoreModel => ({
...
-    CookieManager.clearAll()
-    RelayCache.clearAll()
-    actions.reset({ config })
+    await Promise.all([
+      Platform.OS === "ios" ? await LegacyNativeModules.ARTemporaryAPIModule.clearUserData() : Promise.resolve(),
+      await signOutGoogle(),
+      CookieManager.clearAll(),
+      RelayCache.clearAll(),
+      actions.reset({ config }),
+    ])
```
MounirDhahri, line 82: **note:** there was no need to wait for every individual request to finish to launch the other one, so instead we are now making all those request at the same time

PR #5286, opened 2021-08-16
feat: delay navigate to screen

dimatretyak:

This PR resolves []

#### Problem
When user click on the toast message, when the saved search alert was created, we need to navigate to the page with saved search alerts, which is located on the `Profile` tab. But the main problem is that the Profile tab is a lazy load, and for this reason there is no navigation to the desired page.

#### Solution
Implement delayed navigation, which will be performed at the moment when the Profile tab has been initialized and loaded

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 6a5dbcdec8be263198d2be768c8490b934a01f85
</p>

mdole (approved):
Good catch! Looks great to me

MounirDhahri:
This seems to not work for me the first time the app loads and the profile tab is still not rendered. 

https://user-images.githubusercontent.com/11945712/130221224-0c1ebc11-516b-427e-9006-805d25603ffc.mov

It's working though after that. It might have something to do with a race condition between saving the `delayNavigate` to the store and the reading it in my profile tab. I might be wrong though

dimatretyak:
> It's working though after that. It might have something to do with a race condition between saving the `delayNavigate` to the store and reading it in my profile tab. I might be wrong though

Maybe something else overwrites the `tabProps` for the profile tab. I'll try to reproduce your environment and check how it works

MounirDhahri (commented):
I think there may be a better way to solve this. In `navigate.ts` we can adjust the `navigate` method to be smart and behave like the delayed hook you made. The problem is with the native thread and the js thread having a race condition initially. My idea is to check if we are going to be going to the same route as the `selectedTab` and if not, we wait for the native thread to continue it's execution and then navigate.

Screenshot 2021-08-20 at 18 03 51

dimatretyak:
> I think there may be a better way to solve this. In `navigate.ts` we can adjust the `navigate` method to be smart and behave like the delayed hook you made. The problem is with the native thread and the js thread having a race condition initially. My idea is to check if we are going to be going to the same route as the `selectedTab` and if not, we wait for the native thread to continue it's execution and then navigate.

Thank you for the idea. Everything works fine!

MounirDhahri (approved):
Awesome thanks  🙌 
Please feel free to merge 🛳🛳🛳

PR #5294, opened 2021-08-16
feat: Track separate event for auction screen view GRO-363

jonallured:
This is likely the wrong way to do this because it's not using Cohesion, but what I'm trying to match is this event in Force:

https://github.com/artsy/force/blob/master/src/lib/analytics/trackPageView.ts#L52

https://artsyproduct.atlassian.net/browse/GRO-363

/cc @artsy/grow-devs

### DIFF
--- a/src/lib/Scenes/Sale/Sale.tsx
+++ b/src/lib/Scenes/Sale/Sale.tsx
@@ -237,6 +237,12 @@ export const tracks = {
       context_screen_owner_slug: slug,
     }
   },
+  pageView: (slug: string) => {
+    return {
+      auction_slug: slug,
+      name: "Auctions Pageview",
+    }
+  },
   openFilter: (id: string, slug: string) => {
     return {
       action_name: "filter",
@@ -328,6 +334,12 @@ export const SaleQueryRenderer: React.FC<{ saleID: string; environment?: RelayMo
   saleID,
   environment,
 }) => {
+  const { trackEvent } = useTracking()
+
+  useEffect(() => {
+    trackEvent(tracks.pageView(saleID))
+  }, [])
+
   return (
     <RetryErrorBoundary
       render={({ isRetry }) => {


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 51d60f2c0ec33b8b5b874e54f6cf8b586ed01da5
</p>

pvinis:
for screens, we usually do something like https://github.com/artsy/eigen/blob/fe5163bdc0a6d0d7923f061ca516e08b87848862/src/lib/Scenes/MyCollection/MyCollection.tsx#L70-L75.

if it's a rogue track event that we just want to match, then cool to add it just like that and make a ticket to add it to cohesion and do it right.

src/lib/Scenes/Sale/Components/SaleHeader.tsx
```diff
@@ -19,6 +20,12 @@ interface Props {
 
 export const SaleHeader: React.FC<Props> = ({ sale, scrollAnim }) => {
   const saleTimeDetails = saleTime(sale)
+  const { trackEvent } = useTracking()
+
+  trackEvent({
+    auction_slug: sale.slug,
+    name: "Auctions Pageview",
+  })
```
MounirDhahri, line 28: **thought:** I think you better track this in `Sale.tsx` here https://github.com/artsy/eigen/blob/7d6e9ca3eecce96eaf4534b07aa53bf56cdf8d78/src/lib/Scenes/Sale/Sale.tsx#L332 and add the tracking event as in here https://github.com/artsy/eigen/blob/7d6e9ca3eecce96eaf4534b07aa53bf56cdf8d78/src/lib/Scenes/Sale/Sale.tsx#L231-L239

If for some reason, it has to be inside the `SaleHeader` then it would need to be wrapped with `useEffect` to avoid triggering it on screen re-rerenders
```suggestion
useEffect(() => {
  trackEvent({
    auction_slug: sale.slug,
    name: "Auctions Pageview",
  })
},[])
  
```

jonallured: Oh good call @MounirDhahri !! 🏆

jonallured:
Ok I sorta started over based on feedback from @MounirDhahri and moved the track call with b7ecdfe. So now I'm doing the call inside the query renderer wrapped with a `useEffect` so I think this is better but lmk if I missed anything! Oh I also moved the hard-coded bits into that `tracks` object to follow the existing pattern in that file.

src/lib/Scenes/Sale/Sale.tsx
```diff
@@ -328,6 +334,12 @@ export const SaleQueryRenderer: React.FC<{ saleID: string; environment?: RelayMo
   saleID,
   environment,
 }) => {
+  const { trackEvent } = useTracking()
+
+  useEffect(() => {
+    trackEvent(tracks.pageView(saleID))
```
jonallured, line 340: The saleID here is a slug, right? 🤔

MounirDhahri: Yes 🙌

MounirDhahri (approved):

MounirDhahri:
Thanks for making the changes 🙌

jonallured:
Rebased this on latest master and it's green so going to merge for now.

PR #5311, opened 2021-08-20
spec: update tests to use RNTL and add test for delete address

gkartalis:

This PR resolves []

### Description

This PR adds a test to make sure that the delete mutation is called with the correct input after the user presses delete address button.

Also updated the test file to use `@testing-library/react-native`

-

#### iOS user-facing changes

-

-

#### Dev changes

- add tests for delete address - gkartalis

</details>

### DIFF
--- a/src/lib/Scenes/SavedAddresses/__tests__/SavedAddresses-tests.tsx
+++ b/src/lib/Scenes/SavedAddresses/__tests__/SavedAddresses-tests.tsx
@@ -1,8 +1,8 @@
+import { act, fireEvent } from "@testing-library/react-native"
 import { SavedAddressesTestsQuery } from "__generated__/SavedAddressesTestsQuery.graphql"
 import { navigate, navigationEvents } from "lib/navigation/navigate"
-import { extractText } from "lib/tests/extractText"
-import { renderWithWrappers } from "lib/tests/renderWithWrappers"
-import { Button, Flex } from "palette"
+import { defaultEnvironment } from "lib/relay/createEnvironment"
+import { renderWithWrappersTL } from "lib/tests/renderWithWrappers"
 import React from "react"
 import { graphql, QueryRenderer } from "react-relay"
 import { createMockEnvironment, MockPayloadGenerator } from "relay-test-utils"
@@ -11,10 +11,11 @@ import { SavedAddressesContainer, SavedAddressesQueryRenderer, util } from "../S
 jest.unmock("react-relay")
 
 describe(SavedAddressesQueryRenderer, () => {
-  let mockEnvironment: ReturnType<typeof createMockEnvironment>
+  type MockEnvironment = ReturnType<typeof createMockEnvironment>
+  const mockEnvironment = defaultEnvironment as MockEnvironment
   const TestRenderer = () => (
     <QueryRenderer<SavedAddressesTestsQuery>
-      environment={mockEnvironment}
+      environment={defaultEnvironment}
       query={graphql`
         query SavedAddressesTestsQuery {
           me {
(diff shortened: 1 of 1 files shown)


ArtsyOpenSource:

- Dev changes (add tests for delete address - gkartalis)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 8710e21d892fdada90405544a2559d77a76689db
</p>

src/lib/Scenes/SavedAddresses/__tests__/SavedAddresses-tests.tsx
```diff
@@ -95,24 +93,23 @@ describe(SavedAddressesQueryRenderer, () => {
...
     })
-    const text = extractText(tree.root)
 
-    expect(text).toContain("George Tester")
-    expect(text).toContain("Stallschreiberstr 21, apt 61, 1st Floor")
-    expect(text).toContain("Berlin, 10179")
-    expect(text).toContain("015904832846")
+    expect(queryByText("George Tester")).toBeTruthy()
+    expect(queryByText("Stallschreiberstr 21, apt 61, 1st Floor")).toBeTruthy()
+    expect(queryByText("Berlin, 10179")).toBeTruthy()
```
pvinis, line 99: I think this will always pass?

If this text is not found, query will give `[]`, which is truth, right?

Maybe expect a nonempty array or some other way of testing this.

src/lib/Scenes/SavedAddresses/__tests__/SavedAddresses-tests.tsx
```diff
@@ -186,4 +184,54 @@ describe(SavedAddressesQueryRenderer, () => {
...
+          },
+        }),
+      })
+      return result
+    })
+
+    act(() => fireEvent.press(getAllByText("Delete")[0]))
+
+    const createOperation = mockEnvironment.mock.getMostRecentOperation()
+    expect(createOperation.request.node.operation.name).toEqual("deleteSavedAddressDeleteUserAddressMutation")
```
pvinis, line 234: nice

pvinis (changes requested):

src/lib/Scenes/SavedAddresses/__tests__/SavedAddresses-tests.tsx
gkartalis: Just checked it [here](https://callstack.github.io/react-native-testing-library/docs/api-queries/#queryby) `queryByText` returns null and doesn't pass if it doesn't find the result.

`queryAllByText` is the one that returns the array 🤘

pvinis: o snap. i misread. sorry!

pvinis (approved):

src/lib/Scenes/SavedAddresses/__tests__/SavedAddresses-tests.tsx
```diff
@@ -124,12 +121,13 @@ describe(SavedAddressesQueryRenderer, () => {
       })
       return result
     })
-    tree.findByType(Button).props.onPress()
+
+    fireEvent.press(getAllByText("Add New Address")[0])
```
gkartalis, line 125: This was the only weird part of using `testing-library/react-native`.

For some reason (even when I used debug to see the actual component tree) the content of ListEmptyComponent was displayed x2 times and this is why I used  `    fireEvent.press(getAllByText("Add New Address")[0])` to select the first `Add New Address` button.

It looks like it is a bug that I didn't find a related issue on their repo. Maybe it is worth opening an issue for that.

PR #5351, opened 2021-08-30
fix(FX-3220): Previously selected artists are not reset

dimatretyak:

This PR resolves [FX-3220]

### Description
When a user select artists
And after that select some other filter (for example, ways to buy)
All previously selected artists should be displayed
And previously selected artists should not be reset

#### iOS
https://user-images.githubusercontent.com/3513494/131350247-e86a50dc-2a99-49d3-b22a-efe423a77145.mp4

#### Android
https://user-images.githubusercontent.com/3513494/131352093-2b7e08fd-62ce-4be5-975d-1433f31ece7b.mp4

- Previously selected artists are not reset - dimatretyak

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[FX-3220]: https://artsyproduct.atlassian.net/browse/FX-3220

### DIFF
--- a/src/lib/Components/ArtworkFilter/ArtworkFilterStore.tsx
+++ b/src/lib/Components/ArtworkFilter/ArtworkFilterStore.tsx
@@ -94,7 +94,15 @@ export const ArtworkFiltersModel: ArtworkFiltersModel = {
         filtersToSelect = [...state.selectedFilters, payload]
       }
     } else {
-      filtersToSelect = unionBy([payload], state.selectedFilters, "paramName")
+      filtersToSelect = unionBy([payload], state.selectedFilters, ({ paramValue, paramName }) => {
+        // We don't want to union the artistID params, as each entry corresponds to a
+        // different artist that may be selected. Instead we de-dupe based on the paramValue.
+        if (paramName === FilterParamName.artistIDs && state.filterType === "artwork") {
+          return paramValue
+        } else {
+          return paramName
+        }
+      })
     }
 
     // Then we have to remove any "invalid" choices.
--- a/src/lib/Components/ArtworkFilter/__tests__/ArtworkFiltersStore-tests.tsx
+++ b/src/lib/Components/ArtworkFilter/__tests__/ArtworkFiltersStore-tests.tsx
@@ -1684,6 +1684,135 @@ describe("selectedOptionsUnion", () => {
         },
       ])
     })
+
+    it("correctly unions params when an artistID has been previously applied", () => {
+      const previouslyAppliedFilters = [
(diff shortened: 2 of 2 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (Previously selected artists are not reset - dimatretyak)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against a7ce4d275673aa0ae1983fd337c28a0576c2c2ea
</p>

iskounen (commented):
This change does fix the behavior that was observed, but I am concerned about the amount of code in this method that handles the special case of the `artistIDs` filter. It would take more time, but would it make sense to refactor the artists filter to work like other multi-select filters do?

dimatretyak:
> This change does fix the behavior that was observed, but I am concerned about the amount of code in this method that handles the special case of the `artistIDs` filter. It would take more time, but would it make sense to refactor the artists filter to work like other multi-select filters do?

Yes, it definitely makes sense. Should I do this as part of the current task, or is it better to create a new task, discuss it on grooming session and only then start working?

iskounen:
I agree that it's better to create a new task for the refactor. I imagine that implementing the "All artists I follow" option will be especially difficult.

This work looks good to merge. One suggestion before I do would be to add a test [here](https://github.com/artsy/eigen/blob/50bad0b9dc88856fbbb27861975afdd927d194e3/src/lib/Components/ArtworkFilter/__tests__/ArtworkFiltersStore-tests.tsx#L1069) to assert that any previously applied `artistIDs` are still present in the union after applying other filters.

dimatretyak:
> This work looks good to merge. One suggestion before I do would be to add a test here to assert that any previously applied artistIDs are still present in the union after applying other filters.

There is already such a check [there](https://github.com/artsy/eigen/blob/ea613c50291789119bc28d52e9a7deb783bb3b09/src/lib/Components/ArtworkFilter/__tests__/ArtworkFiltersStore-tests.tsx#L1420)

iskounen:
That test case asserts that selecting additional artists will union correctly, which is valuable, but it wasn't able to prevent this bug because it doesn't assert that selecting additional mediums or time periods will union correctly.

What I'm suggesting is a similar test to the one that you linked to, except that `selectedFilters` would be something like:

```
[{
  displayText: "2010-2020",
  paramName: "majorPeriods",
  paramValue: ["2010-2020"],
 },
 {
   displayText: "Buy Now",
   paramName: "acquireable",
   paramValue: true,
}]
```

iskounen (approved):

PR #5387, opened 2021-09-03
feat: Enable Palette V3 for all users

olerichter00:

This PR resolves [CX-1760]

### Description

Enables Palette V3 for all users by setting `allowV3` to `true`.

- Enable Palette V3 for all users - ole

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-1760]: https://artsyproduct.atlassian.net/browse/CX-1760

### DIFF
--- a/src/palette/PaletteFlag.tsx
+++ b/src/palette/PaletteFlag.tsx
@@ -15,7 +15,7 @@ interface PaletteFlagState {
 export const usePaletteFlagStore = create<PaletteFlagState>(
   persist(
     (set) => ({
-      allowV3: __TEST__ || __STORYBOOK__,
+      allowV3: true,
       setAllowV3: (value) => set((_state) => ({ allowV3: value })),
       toggleAllowV3: () => set((state) => ({ allowV3: !state.allowV3 })),
     }),
--- a/src/palette/PaletteFlag.tsx
+++ b/src/palette/PaletteFlag.tsx
@@ -1,10 +1,8 @@
-import AsyncStorage from "@react-native-community/async-storage"
 import { MenuItem } from "lib/Components/MenuItem"
 import { Text } from "palette"
 import React from "react"
 import { Alert } from "react-native"
 import create from "zustand"
-import { persist } from "zustand/middleware"
 
 interface PaletteFlagState {
   allowV3: boolean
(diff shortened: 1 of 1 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 760bfd084a1ed9247bb8b9c4b9abd3ec2367b9b8
</p>

src/palette/PaletteFlag.tsx
```diff
@@ -15,7 +15,7 @@ interface PaletteFlagState {
 export const usePaletteFlagStore = create<PaletteFlagState>(
   persist(
     (set) => ({
-      allowV3: __TEST__ || __STORYBOOK__,
+      allowV3: true,
```
olerichter00, line 18: Will this be enough to set the value to `true` when the app gets updated?

pvinis: it will not actually 🤔.

the way to do this here would be one of two:
- change the name on line 22 to anything else
- remove the persist wrapper

or a third way could be to add the value change in the migrations but this is not set up yet for rogue flags so it might be 2 second or way more work than is needed.

the explanation is this: we had this flag outside out features.ts file, and that flag is persisted and we have shipped this already. that means that if anyone updates and didn't have that flag, they will get true, which is what we want. if they did have the flag already then the persisted value will not change with this line change, which is not what we want.
so the two solutions basically will "hard reset" the flag. one is basically a new flag key, so the initial value is added fresh, the other is to basically not care about the persisted value and only use a fresh value each time we start the app.

since this flag is temporary, we don't need to set up too much infrastructure with migrations, so I would go for the "remove the persist wrapper". this allows us to change the initial value, and to also be able to flip it back with a new appstore release in the worst case. then in the cleanup ticket we can just remove the flag altogether.

does all that make sense?

pvinis (changes requested):

src/palette/PaletteFlag.tsx
olerichter00: Thanks a lot for the explanation, @pvinis 🌟 

I completely agree, removing the persistent wrapper from Zustand seems to be the best solution. I updated the PR 🚀

pvinis (approved):
awesome! looking forward to QA and release of this, and then we can start the cleanup :)

PR #5392, opened 2021-09-03
fix(FX-3107): display artworks in correct order

dimatretyak:

This PR resolves [FX-3107]

### Description
Breakdown into sections was performed according to `aspectRatio`, thus the sorting order was incorrect (for example, if the section on the left had a lower aspectRation value, then artwork was added to that section)

For example, we received a list of artworks in this order (prices are sorted in descending order). The order of displaying prices: `["$20.000", "$10.000", "$15.000", "$5.000"]`
```javascript
[
  {
    aspectRatio: 1,
    saleMessage: "$20.000",
  },
  {
    aspectRatio: 0.99,
    saleMessage: "$15.000",
  },
  {
    aspectRatio: 1,
    saleMessage: "$10.000",
  },
  {
    aspectRatio: 1,
    saleMessage: "$5.000",
  },
]
```

after splitting into sections, we received the following order (pay attention to `saleMessage`)
```javascript
[
  // section 1
  [
    {
      aspectRatio: 1,
      saleMessage: "$20.000",
    },
    {
      aspectRatio: 1,
      saleMessage: "$5.000",
    }
  ],

  // section 2
  [
    {
      aspectRatio: 0.99,
      saleMessage: "$15.000",
    },
    {
      aspectRatio: 1,
      saleMessage: "$10.000",
    }
  ]
]
```

Expected result. The order of displaying prices: `["$20.000", "$15.000", "$10.000", "$5.000"]`
```javascript
[
  // section 1
  [
    {
      aspectRatio: 1,
      saleMessage: "$20.000",
    },
    {
      aspectRatio: 1,
      saleMessage: "$10.000",
    }
  ],

  // section 2
  [
    {
      aspectRatio: 0.99,
      saleMessage: "$15.000",
    },
    {
      aspectRatio: 1,
      saleMessage: "$5.000",
    }
  ]
]

```

#### iOS
https://user-images.githubusercontent.com/3513494/132208599-07847d7c-a349-46fe-ad1f-ccb08f29631d.mp4

#### Android
https://user-images.githubusercontent.com/3513494/132233352-9d3ca45c-283c-47a0-9e65-19de8ed1fc7f.mp4

- display artworks in correct order when sorting by price - dimatretyak

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[FX-3107]: https://artsyproduct.atlassian.net/browse/FX-3107

### DIFF
--- a/src/lib/Components/ArtworkGrids/InfiniteScrollArtworksGrid.tsx
+++ b/src/lib/Components/ArtworkGrids/InfiniteScrollArtworksGrid.tsx
@@ -1,11 +1,3 @@
-// 1. Get first layout pass of grid view so we have a total width and calculate the column width (componentDidMount?).
-// 2. Possibly do artwork column layout now, as we can do so based just on the aspect ratio, assuming the text height
-//    won't be too different between artworks.
-// 3. Get artwork heights by either:
-//    - calculating the item size upfront with aspect ratio and a static height for the text labels.
-//    - leting the artwork component do a layout pass and calculate its own height based on the column width.
-// 4. Update height of grid to encompass all items.
-
 import React from "react"
 import { ActivityIndicator, Dimensions, LayoutChangeEvent, Platform, StyleSheet, View, ViewStyle } from "react-native"
 import { createFragmentContainer, RelayPaginationProp } from "react-relay"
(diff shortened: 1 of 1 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 3da770567fd52f201c4a4318af0291720477ccf7
</p>

MounirDhahri:
I spent the last bit of time thinking about your implementation and the previous implementation and try to come up with an implementation on my own as well and I don't think there is any possible solution to sort the artworks correctly. Your approach might be the most accurate when the artworks heights are close to each other (almost the same aspect ratio in case we normalize the width), but it will diverge pretty quickly if you use it on an artist that has artworks that are very different in size. In that case, the current approach will give us better results, but it's still not too accurate as well. 
I tried to design a solution based on your approach and the approach we are using now using a stack for each section but I realized it's also not going to be accurate because we can either order images with height, or with price but not both at the same time.
Screenshot 2021-09-08 at 11 01 25
I honestly think we should leave this as it is or discuss with designers that it might make more sense to show artworks as a list as we do at the sale screen. (I already implemented the logic there and it should be just a matter of setting a flag to show the artworks as a list)
Screenshot 2021-09-08 at 11 12 13

dimatretyak:
> I honestly think we should leave this as it is or discuss with designers that it might make more sense to show artworks as a list as we do at the sale screen

@MounirDhahri  Yes, I completely agree with you that there is no 100% working solution if we render artworks as grid view

PR #5397, opened 2021-09-06
Feature(CX-1696): Tabs V3 [StepTabs, ContentTabs, NavigationalTabs]

lordkiz:

This PR resolves [CX-1696]

#### Phone ios
https://user-images.githubusercontent.com/18648835/132239868-f6420ab1-c5cc-41bb-9a62-168fbd82efce.mp4

#### Tablet Android
https://user-images.githubusercontent.com/18648835/132239323-cb7ba5b8-2a4f-42bb-a6dd-c18500367f1f.mp4

- Migrate Tabs to Palette V3 - kizito

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-1696]: https://artsyproduct.atlassian.net/browse/CX-1696


ArtsyOpenSource:

- Cross-platform user-facing changes (Migrate Tabs to Palette V3 - kizito)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against b40dcf36be5430b8f6710793b0564b8c9b297072
</p>

src/lib/Scenes/Fair/Fair.tsx
```diff
@@ -212,15 +216,21 @@ export const Fair: React.FC<FairProps> = ({ fair }) => {
                   }
                   case "fairTabsAndFilter": {
                     const tabToShow = tabs ? tabs[activeTab] : null
+                    const TabVersionToUse = allowV3 ? NavigationalTabs : V2Tabs
```
olerichter00, line 219: Can move this logic to `/palette/elements/Tabs/index.tsx`? Then we can just import it via `import { Tabs } from "palette"`

src/lib/Scenes/Fair/Fair.tsx
```diff
@@ -212,15 +216,21 @@ export const Fair: React.FC<FairProps> = ({ fair }) => {
...
-                          setActiveTab={(index) => {
+                        <TabVersionToUse
+                          onTabPress={(_, index) => {
                             trackTappedNavigationTab(index as number)
                             setActiveTab(index)
                           }}
                           activeTab={activeTab}
                           tabs={tabs}
+                          setActiveTab={(i) => {
+                            // just to match v2tabs. Remove when retiring v2tabs
```
olerichter00, line 230: @pvinis introduced this pattern to identify code that can be removed when we clean up Palette V2:

```
/* TODO-PALETTE-V3 your comment goes here */
```

src/palette/PaletteFlag.tsx
```diff
@@ -16,6 +16,24 @@ export const usePaletteFlagStore = create<PaletteFlagState>((set) => ({
   toggleAllowV3: () => set((state) => ({ allowV3: !state.allowV3 })),
 }))
 
+// test util
+// tslint:disable-next-line:variable-name
```
olerichter00, line 20: I guess we can use a `/* TODO-PALETTE-V3 ` comment here as well.

src/palette/elements/Tabs/ContentTabs.tsx
```diff
@@ -0,0 +1,39 @@
+import { Tab, TabsProps } from "palette/elements/Tabs"
+import React, { useState } from "react"
+import { LayoutRectangle } from "react-native"
+import { TabBarContainer } from "./TabBarContainer"
+
+/**
+ * Renders a  scrollable list of tabs. Tabs are not evenly spaced across screen
+ */
+export const ContentTabs: React.FC<TabsProps> = ({ onTabPress, activeTab, tabs }) => {
+  const [tabLayouts, setTabLayouts] = useState<Array<LayoutRectangle | null>>(tabs.map(() => null))
```
olerichter00, line 10: Not sure if we have some guidelines on this, but what do you think of adding a blank line here as a separator? Maybe it's just me (or my personal preferences), but I have difficulties reading code when different context aren't separated

```suggestion
  const [tabLayouts, setTabLayouts] = useState<Array<LayoutRectangle | null>>(tabs.map(() => null))

```

src/palette/elements/Tabs/TabBarContainer.tsx
```diff
@@ -0,0 +1,139 @@
...
+  const color = useColor()
+  const screen = useScreenDimensions()
+
+  const allTabLayoutsArePresent = tabLayouts.every((l) => l)
+  const scrollViewRef = useRef<ScrollView>(null)
+
+  useEffect(() => {
+    if (allTabLayoutsArePresent) {
+      // attempt to center the active tab
+      let left = 20
```
olerichter00, line 37: Where does the `20` come from?

src/palette/elements/Tabs/TabBarContainer.tsx
```diff
@@ -0,0 +1,139 @@
...
+  // We resize this border using the `scaleX` transform property rather than the `width` property, to avoid running
+  // animations on the JS thread, so we need to set an initial, pre-transform span for the border.
+  const preTransformSpan = 100
+
+  const span = tabLayouts[activeTabIndex].width
+
+  let left = 0
+  for (let i = 0; i < activeTabIndex; i++) {
+    left += tabLayouts[i].width
+  }
```
olerichter00, line 106: What exactly is this code doing?

olerichter00 (commented):
Looks good! I added some comments and questions.

src/palette/elements/Tabs/ContentTabs.tsx
lordkiz: @olerichter00 Haha 😂. Ok I will add a new line.

src/palette/elements/Tabs/TabBarContainer.tsx
lordkiz: @olerichter00 This number can be anything. It can even be 0. It basically allows 20 as the value that the current selected tab can be adrift from the centre in order for the scrollview to recenter it. If set to 0 for example it will try to recenter the active tab even when it is adrift by half its width. i.e the lower the more eager 😁 .
Since we are keeping the old TabBar different from this new TabBars, I think setting this as 0 is better because we don't use any margins. Good catch. Will change to 0.

src/palette/elements/Tabs/TabBarContainer.tsx
lordkiz: Calculates by how much the underline bar will translate to on the horizontal axis when selected. It sums all the widths of the tabs ahead of the current Tab in order to get where to slide to.

src/palette/elements/Tabs/Tab.tsx
```diff
@@ -0,0 +1,43 @@
...
+            alignItems: "center",
+            justifyContent: "center",
+            paddingHorizontal: 15,
+            paddingVertical: 10,
+            ...style,
+          }}
+        >
+          <Text
+            fontSize={"16px"}
+            color={active ? color("black100") : pressed ? color("blue100") : color("black60")}
```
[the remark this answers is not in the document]
lordkiz: Per design, on tap this should be the effect. Let Jonathan do the design review. We can change this if the design team don't approve.

src/lib/Scenes/Fair/Fair.tsx
lordkiz: There are other Tabs besides NavigationalTabs. But I see your point. Will match v2Tabs to Navigational only.

src/palette/elements/Tabs/ContentTabs.tsx
olerichter00: Thanks :)

src/palette/elements/Tabs/TabBarContainer.tsx
olerichter00: Thanks for the explanation. Let's use a constant to save this value (see https://refactoring.guru/replace-magic-number-with-symbolic-constant)

src/lib/Scenes/Fair/Fair.tsx
olerichter00: Ok, we'll remove the v2 components anyway very soon. It probably doesn't matter if we change the logic here or in `index.tsx`.

src/palette/elements/Tabs/TabBarContainer.tsx
lordkiz: @olerichter00 There is no longer any need for a symbolic constant. A constant will be confusing if we are using 0. No Tab styling applies here.

PR #5454, opened 2021-09-16
fix(CX-1929): home page spacing fix

rquartararo:

This PR resolves [CX-1929]

### Description

This PR fixes the spacing issues between rails on the home page, specifically when a new user does not have all of the rails yet.  

Note: make a new account to test. 

- Spacing issues between home page rails is fixed - rquartararo

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

#### Screen recording

https://user-images.githubusercontent.com/50849237/133589242-1ddc6c02-0afc-4838-8151-e38f8d443372.mp4

 

[CX-1929]: https://artsyproduct.atlassian.net/browse/CX-1929

### DIFF
--- a/src/lib/Scenes/Home/Components/SalesRail.tsx
+++ b/src/lib/Scenes/Home/Components/SalesRail.tsx
@@ -46,12 +46,13 @@ const SalesRail: React.FC<Props & RailScrollProps> = (props) => {
     scrollToTop: () => listRef.current?.scrollToOffset({ offset: 0, animated: false }),
   }))
 
-  // Checks to see if the artworks rail is being rendered and hides the separator accordingly
-  useEffect(() => {
-    salesModule.results?.length ? onShow?.() : onHide?.()
-  }, [salesModule.results])
+  const hasSales = salesModule.results?.length
 
-  if (!salesModule.results?.length) {
+  useEffect(() => {
+    hasSales ? onShow?.() : onHide?.()
+  }, [hasSales])
+
+  if (!hasSales) {
     return null
   }
 
--- a/src/lib/Scenes/Home/Components/SalesRail.tsx
+++ b/src/lib/Scenes/Home/Components/SalesRail.tsx
@@ -15,7 +15,7 @@ import { extractNodes } from "lib/utils/extractNodes"
 import { compact } from "lodash"
 import { bullet, Flex, Sans, Text } from "palette"
 import { usePaletteFlagStore } from "palette/PaletteFlag"
-import React, { useImperativeHandle, useRef } from "react"
+import React, { useEffect, useImperativeHandle, useRef } from "react"
 import { FlatList, View } from "react-native"
 import { createFragmentContainer, graphql } from "react-relay"
 import { useTracking } from "react-tracking"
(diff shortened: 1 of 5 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (Spacing issues between home page rails is fixed - rquartararo)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against b5968d5453907eb23e1c97932904182d272458e8
</p>

rquartararo:
There is still an issue with a unique use case in which a user views only one artwork and that particular artwork does't generate the "Similar to Works You've Viewed" rail.

rquartararo:
I think this method may be a good short term solution, but I think we need to eventually reactor the home page to conditionally generate the "rowData" based on if the rails have data and are therefore going to render -- as opposed to having the hide component logic inside each rail after they are already passed into the flatlist. What do you think? @olerichter00 @lordkiz

src/lib/Scenes/Home/Components/SalesRail.tsx
```diff
@@ -39,11 +43,16 @@ const SalesRail: React.FC<Props & RailScrollProps> = (props) => {
     return `${subtitle} ${bullet} ${dateAt}`
   }
 
-  useImperativeHandle(props.scrollRef, () => ({
+  useImperativeHandle(scrollRef, () => ({
     scrollToTop: () => listRef.current?.scrollToOffset({ offset: 0, animated: false }),
   }))
 
-  if (!props.salesModule.results?.length) {
+  // Checks to see if the artworks rail is being rendered and hides the separator accordingly
```
olerichter00, line 50: Can we also do this check inside of the artworks rail and hide the separator?

lordkiz:
> I think this method may be a good short term solution, but I think we need to eventually reactor the home page to conditionally generate the "rowData" based on if the rails have data and are therefore going to render -- as opposed to having the hide component logic inside each rail after they are already passed into the flatlist. I assume with the new rails being added we will have even more issues with spacing. What do you think? @olerichter00 @lordkiz

I think we can even keep this logic if you don't use onHide/onShow as the prop name. There is no harm passing an explicit prop named hideSeparator for the purpose of hiding separators. I think onHide/onShow sounds like something I can use to chain multiple functions/effects reacting to a show or hide event.

olerichter00:
We probably need this logic in every rail that is personalized. Would be great to find a better solution eventually.

> I think we can even keep this logic if you don't use onHide/onShow as the prop name. There is no harm passing an explicit prop named hideSeparator for the purpose of hiding separators. I think onHide/onShow sounds like something I can use to chain multiple functions/effects reacting to a show or hide event.

The reason why I named it onHide/onShow and not hideSeparator is to decouple the components. The module shouldn't know anything about the separators (it could also be placed in a grid or somewhere else).

MounirDhahri (approved):
This LGTM. I do believe though that this might be no longer the best way to solve this. A better approach would probably be to use the same pattern as in [Sale.tsx](https://github.com/artsy/eigen/blob/1fdfc9f2c85a41cdbfc7dbf754378f048b23b3ef/src/lib/Scenes/Sale/Sale.tsx#L145-L145) and by leveraging `compact`, we only send the valid rails to the Flatlist and no longer worry about if it's there or not and remove the extra unneeded [coupling](https://proxy.c2.com/cgi/wiki?LawOfDemeterAndCoupling) as per  the [law of demeter](https://proxy.c2.com/cgi/wiki?LawOfDemeter)

PR #5493, opened 2021-09-22
feat(FX-3327, FX-3330): Keep selected pill as one entity, Reset the state of the search pills

dimatretyak:
The type of this PR is: **Refactor**, **Feature**

This PR resolves [FX-3327], [FX-3330]

### Description
* As a user when I 
  * delete my search term 
  * press `x` input button 
  * press `cancel` button
* I want to be able to start a search from scratch without any pill selected. 

### Selected pill shape
```javascript
{
  type: "elastic" | "algolia",
  displayName: string,
  name: string
}
```

#### iOS
https://user-images.githubusercontent.com/3513494/134516874-340a67b2-7b45-4363-970e-50b4b29e42c8.mp4

#### Android
https://user-images.githubusercontent.com/3513494/134525365-85ad5e1a-8c57-4c76-b118-bd0e73654fc5.mp4

- Reset the state of the pills when the search query is cleared/empty - dimatretyak

#### iOS user-facing changes

-

-

#### Dev changes

- Keep selected search pill as one entity - dimatretyak

</details>

[FX-3327]: https://artsyproduct.atlassian.net/browse/FX-3327

[FX-3330]: https://artsyproduct.atlassian.net/browse/FX-3330

### DIFF
--- a/src/lib/Components/Input/Input.tsx
+++ b/src/lib/Components/Input/Input.tsx
@@ -283,6 +283,7 @@ export const Input = React.forwardRef<TextInput, InputProps>(
                       localClear()
                     }}
                     hitSlop={{ bottom: 40, right: 40, left: 0, top: 40 }}
+                    accessibilityLabel="Clear input button"
                   >
                     <XCircleIcon fill="black30" />
                   </TouchableOpacity>
--- a/src/lib/Scenes/Search2/Search2.tsx
+++ b/src/lib/Scenes/Search2/Search2.tsx
@@ -41,6 +41,12 @@ interface SearchInputProps {
   onSubmitEditing: () => void
 }
 
+type SelectedPillType = "algolia" | "elastic"
+
+interface SelectedPill extends PillType {
+  type: SelectedPillType
+}
+
 const SEARCH_THROTTLE_INTERVAL = 500
 
 const SearchInput: React.FC<SearchInputProps> = ({ currentRefinement, placeholder, refine, onSubmitEditing }) => {
@@ -101,10 +107,8 @@ interface Search2Props {
 export const Search2: React.FC<Search2Props> = (props) => {
   const { system, relay } = props
   const [searchState, setSearchState] = useState<SearchState>({})
-  const [selectedAlgoliaIndex, setSelectedAlgoliaIndex] = useState("")
-  const [elasticSearchEntity, setElasticSearchEntity] = useState("")
+  const [selectedPill, setSelectedPill] = useState<SelectedPill | null>(null)
   const searchProviderValues = useSearchProviderValues(searchState?.query ?? "")
-  const [activePillDisplayName, setActivePillDisplayName] = useState("")
   const { searchClient } = useAlgoliaClient(system?.algolia?.appID!, system?.algolia?.apiKey!)
   const searchInsightsConfigured = useSearchInsightsConfig(system?.algolia?.appID, system?.algolia?.apiKey)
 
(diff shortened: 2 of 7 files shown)


ArtsyOpenSource:

- Dev changes (Keep selected search pill as one entity - dimatretyak)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 7ee329b68ead7fe7191a514c57a50fa949a6861f
</p>

src/lib/Scenes/Search2/Search2.tsx
```diff
@@ -39,32 +39,47 @@ interface SearchInputProps {
...
     <SearchBox
       ref={searchProviderValues.inputRef}
       enableCancelButton
       placeholder={placeholder}
-      onChangeText={handleChangeText}
+      onChangeText={(text) => {
+        trackEvent({
+          action_type: Schema.ActionNames.ARAnalyticsSearchStartedQuery,
+          query: text,
+        })
```
dimatretyak, line 74: is it ok that we send an analytics event every time text changes?

src/lib/Components/Input/Input.tsx
```diff
@@ -283,6 +283,7 @@ export const Input = React.forwardRef<TextInput, InputProps>(
                       localClear()
                     }}
                     hitSlop={{ bottom: 40, right: 40, left: 0, top: 40 }}
+                    accessibilityLabel="Clear input button"
```
gkartalis, line 286: 😍

src/lib/Scenes/Search2/Search2.tsx
gkartalis: I think it is since this is what we have been doing withe the previous implementation as well.

See [here](https://github.com/artsy/eigen/blob/master/src/lib/Scenes/Search/Search.tsx#L36-L43)

src/lib/Scenes/Search2/Search2.tsx
```diff
@@ -126,51 +150,49 @@ export const Search2: React.FC<Search2Props> = (props) => {
...
+    setSelectedPill(null)
+  }
+
   return (
     <SearchContext.Provider value={searchProviderValues}>
       <ArtsyKeyboardAvoidingView>
         <InstantSearch
           searchClient={searchClient}
-          indexName={selectedAlgoliaIndex}
+          indexName={selectedPill?.name ?? ""}
```
gkartalis, line 195: **suggestion/thought:** I was thinking here, the indexName theoretically should be only `algoliaIndex`, and by passing `selectedPill?.name` we are passing `ARTWORK` as name when Artwork Pill is selected. What about changing to something like?

```suggestion
          indexName={selectedPill?.type === "algolia" ? selectedPill.name : ""}
```

src/lib/Scenes/Search2/Search2.tsx
```diff
@@ -76,7 +91,9 @@ const SearchInput: React.FC<SearchInputProps> = ({ currentRefinement, placeholde
         trackEvent({
           action_type: Schema.ActionNames.ARAnalyticsSearchCleared,
         })
```
gkartalis, line 93: **suggestion(non-blocking):** This tracks that the search term was cleared, maybe we could move this inside `handleReset` so everytime we call handle reset we can track that the `query` was cleared? What do you think about that?

gkartalis (approved):
Approved! Just some minor suggestions, would like to hear your opinion on the comments.
Really liked that you simplified the pill state as well.

src/lib/Scenes/Search2/Search2.tsx
dimatretyak: You are absolutely right. Thank you for the changes!

src/lib/Scenes/Search2/Search2.tsx
dimatretyak: I think it makes sense

dimatretyak: Or would it be better to track the event inside the throttle function, as it is [done now](https://github.com/artsy/eigen/blob/master/src/lib/Scenes/Search2/Search2.tsx#L54-L57)?

gkartalis: I would say let's go with the one that you suggested above 👆 and not inside the throttle func

PR #5497, opened 2021-09-22
feat:[CX-1813] add trove to homepage

MrSltun:

This PR resolves [CX-1813]

### Description

This PR adds the Trove in the homepage into Eigen

The Trove in Eigen

### Follow Up
This update requires adding the Trove into the app hero units in the CMS, for more details please [click here 🔒](https://www.notion.so/artsy/Adding-a-new-Trove-to-the-App-37be64dab61040819de90b2fb288adde)

- Add Trove into Eigen - mrsltun

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-1813]: https://artsyproduct.atlassian.net/browse/CX-1813


ArtsyOpenSource:

- Cross-platform user-facing changes (Add Trove into Eigen - mrsltun)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 3698c476d2182946871ed6953595ccf9ecd7a374
</p>

src/lib/Components/Home/HeroUnit.tsx
```diff
@@ -0,0 +1,86 @@
...
+            locations={[0.4, 1]}
+            style={{
+              position: "absolute",
+              width,
+              height,
+            }}
+          />
+        )}
+        <Flex maxWidth={isPad() ? "65%" : undefined}>
+          <Sans size="8" color="white">
```
lordkiz, line 46: I think Text can now be used in place of Sans.

src/lib/Components/Home/HeroUnit.tsx
```diff
@@ -0,0 +1,86 @@
...
+        )}
+        <Flex maxWidth={isPad() ? "65%" : undefined}>
+          <Text variant="xl" color="white">
+            {unit.title}
+          </Text>
+          {unit.subtitle ? (
+            <Text variant="sm" color="white" mt={0.5}>
+              {unit.subtitle}
+            </Text>
+          ) : null}
```
olerichter00, line 53:
```suggestion
          {!!unit.subtitle && (
            <Text variant="sm" color="white" mt={0.5}>
              {unit.subtitle}
            </Text>
          )}
```

src/lib/Scenes/Home/Home.tsx
```diff
@@ -188,6 +191,8 @@ const Home = (props: Props) => {
                       scrollRef={scrollRefs.current[index]}
                     />
                   )
+                case "trove":
+                  return homePageBelow ? <TroveContainer trove={homePageBelow} /> : <></>
```
olerichter00, line 195: We'll need some logic to hide the separator above the 'Trove' module in case the module doesn't render anything. If we don't hide the separator we'll have too much space between the 'Articles' and the 'Viewing Room' module.

For other modules we use `onShow` and `onHide` props to conditionally show and hide the separator (e.g. [here](https://github.com/artsy/eigen/blob/e9761dbe660bca1fa731f70a56f8bb4d39741c97/src/lib/Scenes/Home/Home.tsx#L169-L170)).

src/lib/Components/Home/HeroUnit.tsx
```diff
@@ -0,0 +1,86 @@
+import { isPad } from "lib/utils/hardware"
+import { useScreenDimensions } from "lib/utils/useScreenDimensions"
+import { Flex, Text, useColor } from "palette"
+import React, { useState } from "react"
+import { Image, TouchableOpacity, View } from "react-native"
+import LinearGradient from "react-native-linear-gradient"
+
+interface Props {
+  unit: any
```
olerichter00, line 9: Let's try to avoid the use of `any` whenever possible. To get the correct type from Relay you can use:

```suggestion
  unit: NonNullable<NonNullable<HomeHero_homePage["heroUnits"]>[0]>
```

src/lib/Components/Home/HeroUnit.tsx
```diff
@@ -0,0 +1,86 @@
...
+  onPress: () => void
+  isTrove?: boolean
+}
+
+const useHeroDimensions = () => {
+  const { width, height: screenHeight } = useScreenDimensions()
+  const height = Math.floor(Math.min(480, screenHeight * 0.6))
```
olerichter00, line 16: **suggestion:** I'm wondering if it might be a good idea to add a comment to explain where these values (`480` and `0.6`) come from or/and use constants to avoid [magic numbers](https://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad#:~:text=In%20programming%2C%20a%20%22magic%20number,in%20more%20than%20one%20place.).

src/lib/Scenes/Home/Components/Trove.tsx
```diff
@@ -0,0 +1,46 @@
+import { Trove_trove } from "__generated__/Trove_trove.graphql"
+import { HeroUnit } from "lib/Components/Home/HeroUnit"
+import { navigate } from "lib/navigation/navigate"
+import React, { useEffect } from "react"
+import { createFragmentContainer, graphql } from "react-relay"
+
+interface TroveProps {
+  trove: Trove_trove
+  onHide?: () => void
+  onShow?: () => void
```
olerichter00, line 10: 👍

src/lib/Scenes/Home/Components/Trove.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  const handleOnPress = () => {
+    const path = troveUnit!.href!
+    navigate(path)
+  }
+
+  return <HeroUnit unit={troveUnit} onPress={handleOnPress} isTrove />
+}
+
+export const TroveContainer = createFragmentContainer(Trove, {
```
olerichter00, line 34:
```suggestion
export const TroveFragmentContainer = createFragmentContainer(Trove, {
```

olerichter00 (approved):
Amazing work, this looks very good 🌟 
I added a couple of suggestions, but feel free to merge the PR.

src/lib/Components/Home/HeroUnit.tsx
```diff
@@ -0,0 +1,93 @@
...
+export const HeroUnit: React.FC<Props> = ({ unit, onPress, isTrove = false }) => {
+  const color = useColor()
+  const [hasLoaded, setHasLoaded] = useState(false)
+
+  const { width, height } = useHeroDimensions()
+
+  const linkText = !isTrove && (unit as any).linkText
+
+  return (
+    <TouchableOpacity activeOpacity={0.9} onPress={onPress}>
```
MounirDhahri, line 34: **nitpick:** ideally we use our custom component called `Touchable` that support haptic feedback here

src/lib/Components/Home/HeroUnit.tsx
```diff
@@ -0,0 +1,93 @@
...
+  return { width, height }
+}
+
+export const HeroUnit: React.FC<Props> = ({ unit, onPress, isTrove = false }) => {
+  const color = useColor()
+  const [hasLoaded, setHasLoaded] = useState(false)
+
+  const { width, height } = useHeroDimensions()
+
+  const linkText = !isTrove && (unit as any).linkText
```
MounirDhahri, line 31: **issue:** It seems like you are already setting the unit to be the right type in line 13. I don't understand why you need to cast it as `any` here since we are trying to avoid using any

src/lib/Components/Home/HeroUnit.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  return (
+    <TouchableOpacity activeOpacity={0.9} onPress={onPress}>
+      <Flex height={height} justifyContent="flex-end" p="2" style={{ backgroundColor: color("black30") }}>
+        <Image
+          style={{ width, height, position: "absolute" }}
+          source={{ uri: unit.backgroundImageURL! }}
```
MounirDhahri, line 38: **issue(blocker):** I think we need to be checking here if `unit.backgroundImageURL` is already there before returning the `<Image />` since it's not mandatory in the form and remove the `!`

src/lib/Components/Home/HeroUnit.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const [hasLoaded, setHasLoaded] = useState(false)
+
+  const { width, height } = useHeroDimensions()
+
+  const linkText = !isTrove && (unit as any).linkText
+
+  return (
+    <TouchableOpacity activeOpacity={0.9} onPress={onPress}>
+      <Flex height={height} justifyContent="flex-end" p="2" style={{ backgroundColor: color("black30") }}>
+        <Image
```
MounirDhahri, line 36: **issue(non-blocker):** I think we better use `<OpaqueImageView />` here which supports caching, fade animation, and is also optimized to load the right image size

src/lib/Scenes/Home/Components/__tests__/Trove-tests.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { renderWithWrappers } from "lib/tests/renderWithWrappers"
+import React from "react"
+import { Image, TouchableOpacity } from "react-native"
+import { graphql, QueryRenderer } from "react-relay"
+import { createMockEnvironment, MockPayloadGenerator } from "relay-test-utils"
+import { TroveFragmentContainer } from "../Trove"
+
+jest.unmock("react-relay")
+
+describe("Trove", () => {
```
MounirDhahri, line 13: **praise:** Thanks for the nice tests! 👏

MounirDhahri (commented):
Nice PR! I left few minor comments and we should be clear to merge this after

src/lib/Components/Home/HeroUnit.tsx
olerichter00: Is there any way to mark exports from npm packages as deprecated within Eigen? It would be very helpful to see a deprecation warning pointing to an alternative in the editor.

src/lib/Components/Home/HeroUnit.tsx
MrSltun: the point to cast linkText to `any` is `linkText` only requested in HomeHero query, not Trove query, so when calling the `HeroUnit` component in Trove it asks for `linkText` since it is inside the `HomeHero_homePage` type

MounirDhahri: then I believe it would make sense to have something like `(unit as any)?.linkText`

MounirDhahri:
thank you for the changes 🙌

PR #5544, opened 2021-09-28
feat: Artist's Articles in My Collection

rajsam003:

This PR resolves [CX-1941]

### Description

Native screen showing all articles in My Collection instead of showing in the webView. Paired with @lordkiz 

Screenshot 2021-09-29 at 16 51 20

- Native screen showing all articles instead of showing in webView - rajsam003 

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-1941]: https://artsyproduct.atlassian.net/browse/CX-1941


src/lib/Scenes/Articles/ArticlesList.tsx
```diff
@@ -17,6 +16,7 @@ interface ArticlesListProps {
   refreshing: boolean
   handleLoadMore: () => void
   handleRefresh: () => void
+  title: any
```
olerichter00, line 19:
```suggestion
  title: string
```

src/lib/Scenes/ArtistArticles/ArtistArticles.tsx
```diff
@@ -0,0 +1,148 @@
...
+      title={artist.name + " articles"}
+      hasMore={relay.hasMore}
+      refreshing={refreshing}
+      handleLoadMore={handleLoadMore}
+      handleRefresh={handleRefresh}
+    />
+  )
+}
+
+export const ArticlesPlaceholder = () => {
```
olerichter00, line 57: I think we can use `ArticlePlaceholder` from [Articles](https://github.com/artsy/eigen/blob/451dbca386a08e5d853a1046a444b06101da56cb/src/lib/Scenes/Articles/Articles.tsx#L74) (and maybe even move the placeholder component to `ArticlesList.tsx`.

olerichter00: And we need to increase articles placeholder top padding. It doesn't really match the articles from the loaded screen

src/lib/Scenes/ArtistArticles/ArtistArticles.tsx
```diff
@@ -0,0 +1,148 @@
...
+  return (
+    <ArticlesList
+      articles={articles as any}
+      isLoading={relay.isLoading}
+      title={artist.name + " articles"}
+      hasMore={relay.hasMore}
+      refreshing={refreshing}
+      handleLoadMore={handleLoadMore}
+      handleRefresh={handleRefresh}
+    />
```
olerichter00, line 53: Not sure where this space at the bottom of the artist articles screen comes from, but it looks like it's too much...
simulator_screenshot_AAD11F9B-2634-4203-A0C7-AACC3123B552

olerichter00 (commented):
Great work! 🌟  Just left a few comments.

src/lib/utils/track/schema.ts
```diff
@@ -145,6 +145,7 @@ export enum OwnerEntityTypes {
   AllArtistSeries = "AllArtistSeries",
   Artist = "Artist",
   ArtistSeries = "ArtistSeries",
+  ArtistArticles = "ArtistArticles",
```
olerichter00, line 148: @rajsam003 Let's use [Cohesion](https://github.com/artsy/cohesion) for [tracking](https://github.com/artsy/eigen/blob/master/docs/analytics_and_tracking.md#L1)

src/lib/Scenes/Articles/ArticlesList.tsx
```diff
@@ -111,6 +113,38 @@ export const useNumColumns = () => {
   return orientation === "portrait" ? 2 : 3
 }
 
+export const ArticlesPlaceholder = () => {
+  const numColumns = 1
+
```
lordkiz, line 118:
```suggestion

// non hooks version of useNumColumns
const nonHookNumOfColumns = (): number => {
  const { width, height } = Dimensions.get("window")
  const isPortrait = height > width
  const isTablet = isPad()
  if ((isTablet && isPortrait) || (!isTablet && !isPortrait)) {
    return 2
  }
  if (isTablet && !isPortrait) {
    return 3
  }
  return 1
}

export const ArticlesPlaceholder = () => {
  const numColumns = nonHookNumOfColumns()

```

@rajsam003 in addition to @olerichter00 concerns about the ArticlesList placeholder, I have also noticed that it does not look good on tablets in both portrait and landscape mode. The placeholders do not match the list eventually displayed, because numColumns cannot be 1 on tablets.
<p align='center'>

</p>

olerichter00: Using the placeholder from `articles.tsx` will solve the problem. It uses `useNumColumns()` to calculate the correct number of columns.

lordkiz: @olerichter00 This placeholder is what is being used in Articles.tsx. It will also not look good on tablets.  I don't think useNumColumns will work as it is a hook.

olerichter00: I just checked and it calculates the correct number of columns on tablets. The placeholder is only a component so hooks should work as expected :)

lordkiz: Ohhh. @olerichter00 I get you now. You are referring to Articles.tsx in the current master. I couldn't find it in this branch, I can now see this PR removes it. 
Regarding if he can use hooks or not, I was referring to the fact that he cannot use hooks if he actually wants to use the new Placeholder in ArtistArticlesQueryRenderer's renderPlaceholder as is. i.e `renderPlaceholder: ArticlesPlaceholder`. I was assuming there are other places in the app that uses it that way that he doesn't want to touch hence the need to use a non hooks approach in order to continue with what he already has. Thanks @olerichter00 
In that case, @rajsam003, I will suggest either of the following:
1. Use the already existing ArticlesPlaceholder in Articles.tsx.
2. Or if you still want to keep things as they currently are in this branch, then you'd have to do the following:
`in ArticlesList.tsx`
```suggestion
export const ArticlesPlaceholder = () => {
  const numColumns = useNumColumns()
```

`in render prop of ArtistArticlesQueryRenderer inside ArtistArticles.tsx`

```
        Container: ArtistArticlesContainer,
        renderPlaceholder: () => <ArticlesPlaceholder />, // instead of renderPlaceholder: ArticlesPlaceholder
      })}
```

olerichter00 (approved):
🚀

lordkiz (approved):
Awesome work Sam. 🎉

src/lib/Scenes/ArtistArticles/ArtistArticles.tsx
```diff
@@ -0,0 +1,116 @@
...
+  {
+    getConnectionFromProps: ({ artist }) => artist.articlesConnection,
+    getVariables: (__, { count, cursor }, _fragmentVariables) => ({
+      ..._fragmentVariables,
+      count,
+      cursor,
+    }),
+    query: graphql`
+      query ArtistArticlesQuery($artistID: String!, $count: Int, $cursor: String) {
+        artist(id: $artistID) @principalField {
```
MounirDhahri, line 85: @rajsam003 is there a reason why you added the @principalField directive here?

src/lib/Scenes/ArtistArticles/__tests__/ArtistArticles-tests.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { renderWithWrappers } from "lib/tests/renderWithWrappers"
+import React from "react"
+import { graphql, QueryRenderer } from "react-relay"
+import { act } from "react-test-renderer"
+import { createMockEnvironment, MockPayloadGenerator } from "relay-test-utils"
+import { ArtistArticles, ArtistArticlesContainer } from "../ArtistArticles"
+
+jest.unmock("react-relay")
+
+describe("Artist Articles", () => {
```
MounirDhahri, line 11: **issue(blocker):** would you mind adding more tests for this component to make sure that the list of articles is getting rendered if articles exists and also to render no articles if no articles were found.

PR #5557, opened 2021-09-29
fix(FX-3278): keep the "show alerts" button disabled when the user selects the same custom price again

dimatretyak:

This PR resolves [FX-3278]

#### Before
https://user-images.githubusercontent.com/3513494/135315790-96be0ce3-1364-4e06-8744-0b85aa72857a.mp4

#### After
https://user-images.githubusercontent.com/3513494/135315655-c96a8544-50e9-428c-a4c3-13ee3993c451.mp4

-

#### iOS user-facing changes

-

-

#### Dev changes

- keep the "show alerts" button disabled when the user selects the custom price again - dimatretyak

</details>

[FX-3278]: https://artsyproduct.atlassian.net/browse/FX-3278

### DIFF
--- a/src/lib/Components/ArtworkFilter/ArtworkFilterStore.tsx
+++ b/src/lib/Components/ArtworkFilter/ArtworkFilterStore.tsx
@@ -29,6 +29,7 @@ export interface ArtworkFiltersModel {
   setFiltersCountAction: Action<ArtworkFiltersModel, FilterCounts>
   setFilterTypeAction: Action<ArtworkFiltersModel, FilterType>
   setInitialFilterStateAction: Action<ArtworkFiltersModel, FilterArray>
+  setSelectedFiltersAction: Action<ArtworkFiltersModel, FilterArray>
 }
 
 export type ArtworkFiltersState = State<ArtworkFiltersModel>
@@ -134,6 +135,11 @@ export const ArtworkFiltersModel: ArtworkFiltersModel = {
   setInitialFilterStateAction: action((state, payload) => {
     state.appliedFilters = payload
     state.previouslyAppliedFilters = payload
+    state.applyFilters = false
+  }),
+
+  setSelectedFiltersAction: action((state, payload) => {
+    state.selectedFilters = payload
     state.applyFilters = false
   }),
 }
--- a/src/lib/Components/ArtworkFilter/__tests__/ArtworkFiltersStore-tests.tsx
+++ b/src/lib/Components/ArtworkFilter/__tests__/ArtworkFiltersStore-tests.tsx
@@ -2205,3 +2205,137 @@ describe("SetFilterCounts", () => {
     })
   })
 })
+
+describe("SetSelectedFiltersAction", () => {
+  it("set the selected filters when the previously selected filters is empty", () => {
+    filterState = {
+      applyFilters: false,
+      appliedFilters: [],
+      previouslyAppliedFilters: [],
+      selectedFilters: [],
+      aggregations: [],
+      filterType: "artwork",
+      counts: {
+        total: null,
+        followedArtists: null,
+      },
+    }
+
+    const filterArtworksStore = getFilterArtworksStore(filterState)
+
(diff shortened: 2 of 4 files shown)


ArtsyOpenSource:

- Dev changes (keep the "show alerts" button disabled when the user selects the custom price again - dimatretyak)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against f873941d614d31358c98186394f7c200383ab2a5
</p>

src/lib/Components/ArtworkFilter/ArtworkFilterStore.tsx
```diff
@@ -72,32 +72,11 @@ export const ArtworkFiltersModel: ArtworkFiltersModel = {
...
-      if (appliedFilter.paramValue === paramValue) {
-        // Ignore this case when we are setting back the default sorting for saleArtworks
-        return (
-          defaultFilterOptions[paramName] === appliedFilter.paramValue &&
-          appliedFilter.paramName === FilterParamName.sort &&
-          state.filterType === "saleArtwork"
-        )
-      }
-
-      return true
```
dimatretyak, line 100: As far as I understand, there is no longer a need to do checks for `saleArtwork` type, since we no longer [have the same](https://github.com/artsy/eigen/blob/705197695e881657d264328f7072222e3fe184ca/src/lib/Components/ArtworkFilter/Filters/SortOptions.tsx#L81-L112) `paramValue`

src/lib/Components/ArtworkFilter/ArtworkFilterStore.tsx
```diff
@@ -136,6 +137,11 @@ export const ArtworkFiltersModel: ArtworkFiltersModel = {
     state.previouslyAppliedFilters = payload
     state.applyFilters = false
   }),
+
+  setSelectedFiltersAction: action((state, payload) => {
+    state.selectedFilters = payload
+    state.applyFilters = false
+  }),
```
MounirDhahri, line 144: **thought:** was there a reason why you created this new action instead of using `selectFiltersAction`. I think you can probably achieve the same output using it (you might need to tweak it a bit)

src/lib/Components/ArtworkFilter/__tests__/ArtworkFiltersStore-tests.tsx
```diff
@@ -2205,3 +2205,137 @@ describe("SetFilterCounts", () => {
     })
   })
 })
+
+describe("SetSelectedFiltersAction", () => {
```
MounirDhahri, line 2209: thanks for the thorough tests 🙌

MounirDhahri (approved):
LGTM 👏, I left a small question/suggestion

src/lib/Components/ArtworkFilter/ArtworkFilterStore.tsx
dimatretyak: According to the conditions of the task, we need to make the "show results" button enabled only when the user **has entered something in custom inputs**

##### Consider the situation
1. User previously specified custom values (for example, `min width = 5 cm`, `max width = 10 cm`)
2. User has selected one of the predefined values (for example, `Large (over 40in)`)
3. Since the user has selected a predefined value, we need to reset the custom width and height values (`width = "*-*"`, `height = "*-*"`)
4. User again selects the "Custom Size" option again, but doesn't enter anything into the inputs
5. Since we have `selectedFilters` stored `width = "*-*"`, `height = "*-*"`, by [this reason](https://github.com/artsy/eigen/blob/master/src/lib/Components/ArtworkFilter/ArtworkFilter.tsx#L141-L142) our "show results" button will be enabled (in the `before` section you can see a demo video with this problem)

`setSelectedFiltersAction` allows us to reset the selected size values if the user has re-selected the `Custom Size` option

MounirDhahri: thanks for the thorough explanation. Please feel free to merge

PR #5652, opened 2021-10-15
feat(DX): name test files `*.tests.ts[x]` and move them next to the component

pvinis:

This PR resolves []

### Description

The way I did this was with this script
```rb
require 'find'

stories = []
tests = []
testsunderscores = []
Find.find('.') do |path|
  next if path =~ /\.git/
  next if path =~ /\.jest/
  next if path =~ /\.metro/
  next if path =~ /Pods/
  next if path =~ /node_modules/
  next if path =~ /derived_data/
  # puts path

  stories << path if path =~ /-stories/
  tests << path if path =~ /-tests/
  testsunderscores << path if path =~ /__tests__/
end

stories.each do |path|
  File.rename(path, path.gsub(/(.*)-stories(.*)/, '\1.stories\2'))
end
tests.each do |path|
  next if path =~ /metaphysics-queries-freshness-tests/
  next if path =~ /build-for-tests/
  File.rename(path, path.gsub(/(.*)-tests(.*)/, '\1.tests\2'))
end
testsunderscores.each do |path|
  File.rename(path, path.gsub(/(.*)__tests__\/(.*)/, '\1\2'))
end
```
and then I fixed the import errors

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/lib/relay/middlewares/cacheMiddleware.ts
+++ b/src/lib/relay/middlewares/cacheMiddleware.ts
@@ -15,7 +15,6 @@ export const cacheMiddleware = () => {
       // @ts-expect-error STRICTNESS_MIGRATION --- 🚨 Unsafe legacy code 🚨 Please delete this and fix any type errors if you have time 🙏
       const dataFromCache = await cache.get(queryID, variables)
       if (dataFromCache) {
-        console.log("RELAY", operation.name, variables, "satisfied from cache")
         return JSON.parse(dataFromCache)
       }
     }


MounirDhahri:
It seems like we will be losing our git history due to renaming the files and I am not sure this is something we would want to do 😬
If there is no way we can rename these files without losing their history I think we should consider just moving them. What do you all think?

AlicanAkyuz:
what I understand from a lil research is that git doesn’t drop the history of renamed files (if the content is unchanged) but doesn’t show it by default. using git log with --follow and path seems to work. for example: `git log --follow /Users/alicanakyuz/Desktop/artsy/eigen/src/lib/Components/Artist/ArtistAbout/ArtistAboutShows.tests.tsx` shows the history for me on this branch.

i'm using gitLens in vs code and there is a setting for following the history of renamed files:
Screenshot 2021-10-18 at 09 35 49

then I have access to the history:
Screenshot 2021-10-18 at 09 21 46

so to me, it overall seems fine to rename the files.

PR #5658, opened 2021-10-18
feat: Add comparable works section to auction result screen

olerichter00:

This PR resolves [CX-942]

### Description

Adds comparable works section to auction result screen.

**Todo:**

- [x] Rework tracking
- [x] Add a feature flag (+ Echo flag)
- [x] Add `@optionalField` directive to `comparableAuctionResults` fragment

**Events:**

```
[Event tracked] {
  "action": "tappedAuctionResultGroup",
  "context_screen_owner_type": "auctionResult",
  "context_screen_owner_id": "5668260",
  "destination_screen_owner_type": "auctionResult",
  "destination_screen_owner_id": "362630",
  "horizontal_slide_position": 0,
  "type": "thumbnail"
}
```

### To the reviewers 👀

- Add comparable works section to auction result screen - ole

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-942]: https://artsyproduct.atlassian.net/browse/CX-942?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/Scenes/AuctionResult/AuctionResult.tests.tsx
+++ b/src/lib/Scenes/AuctionResult/AuctionResult.tests.tsx
@@ -1,7 +1,7 @@
 import { AuctionResultTestsQuery } from "__generated__/AuctionResultTestsQuery.graphql"
 import { AuctionResultsMidEstimate } from "lib/Components/AuctionResult/AuctionResultMidEstimate"
 import { FancyModalHeader } from "lib/Components/FancyModal/FancyModalHeader"
-import { AuctionResultListItem } from "lib/Components/Lists/AuctionResultListItem"
+import { AuctionResultFragmentContainer } from "lib/Components/Lists/AuctionResultListItem"
 import { extractText } from "lib/tests/extractText"
 import { renderWithWrappers } from "lib/tests/renderWithWrappers"
 import React from "react"
@@ -113,7 +113,7 @@ describe("AuctionResult", () => {
     })
   })
 
-  describe("Auction Result is not empty", () => {
+  describe.only("Auction Result is not empty", () => {
     it("show the mid-estimate", () => {
       const tree = renderWithWrappers(<TestRenderer />)
       mockEnvironment.mock.resolveMostRecentOperation((operation) =>
@@ -143,7 +143,7 @@ describe("AuctionResult", () => {
         })
       )
 
-      expect(tree.root.findAllByType(AuctionResultListItem)).toHaveLength(3)
+      expect(tree.root.findAllByType(AuctionResultFragmentContainer)).toHaveLength(3)
     })
   })
 
--- a/src/lib/Scenes/AuctionResult/AuctionResult.tests.tsx
+++ b/src/lib/Scenes/AuctionResult/AuctionResult.tests.tsx
@@ -113,7 +113,7 @@ describe("AuctionResult", () => {
     })
   })
 
-  describe.only("Auction Result is not empty", () => {
+  describe("Auction Result is not empty", () => {
     it("show the mid-estimate", () => {
       const tree = renderWithWrappers(<TestRenderer />)
       mockEnvironment.mock.resolveMostRecentOperation((operation) =>
--- a/src/lib/Scenes/AuctionResult/AuctionResult.tests.tsx
+++ b/src/lib/Scenes/AuctionResult/AuctionResult.tests.tsx
@@ -1,29 +1,84 @@
+import { AuctionResultTestsQuery } from "__generated__/AuctionResultTestsQuery.graphql"
 import { AuctionResultsMidEstimate } from "lib/Components/AuctionResult/AuctionResultMidEstimate"
 import { FancyModalHeader } from "lib/Components/FancyModal/FancyModalHeader"
+import { AuctionResultListItem } from "lib/Components/Lists/AuctionResultListItem"
 import { extractText } from "lib/tests/extractText"
-import { mockEnvironmentPayload } from "lib/tests/mockEnvironmentPayload"
 import { renderWithWrappers } from "lib/tests/renderWithWrappers"
 import React from "react"
 import { Image, ScrollView } from "react-native"
-import { createMockEnvironment } from "relay-test-utils"
-import { AuctionResultQueryRenderer } from "./AuctionResult"
+import { graphql, QueryRenderer } from "react-relay"
+import { act, ReactTestRenderer } from "react-test-renderer"
+import { createMockEnvironment, MockPayloadGenerator } from "relay-test-utils"
+import { AuctionResult } from "./AuctionResult"
 
 jest.unmock("react-relay")
 
 describe("AuctionResult", () => {
   let mockEnvironment: ReturnType<typeof createMockEnvironment>
-
-  const renderAuctionResult = (mockedData = {}) => {
-    const tree = renderWithWrappers(
-      <AuctionResultQueryRenderer
-        artistID="artist-id"
-        auctionResultInternalID="auction-result-id"
-        environment={mockEnvironment}
-      />
-    )
-    mockEnvironmentPayload(mockEnvironment, mockedData)
-    return tree
-  }
+  const TestRenderer = () => (
+    <QueryRenderer<AuctionResultTestsQuery>
+      environment={mockEnvironment}
+      query={graphql`
+        query AuctionResultTestsQuery($auctionResultInternalID: String!, $artistID: String!) @relay_test_operation {
+          auctionResult(id: $auctionResultInternalID) {
+            id
+            internalID
+            artistID
+            boughtIn
+            currency
+            categoryText
+            ...ComparableWorks_auctionResult
+            dateText
+            dimensions {
+              height
+              width
+            }
+            dimensionText
+            estimate {
+              display
+              high
+              low
+            }
+            images {
+              thumbnail {
+                url(version: "square140")
+                height
+                width
+                aspectRatio
+              }
+            }
+            location
+            mediumText
+            organization
+            performance {
+              mid
+            }
+            currency
+            priceRealized {
+              cents
+              centsUSD
+              display
+              displayUSD
+            }
+            saleDate
+            saleTitle
+            title
+          }
+          artist(id: $artistID) {
+            name
+            href
+          }
+        }
+      `}
+      variables={{ artistID: "artist-id", auctionResultInternalID: "auction-result-id" }}
+      render={({ props }) => {
+        if (props?.auctionResult && props?.artist) {
+          return <AuctionResult auctionResult={props.auctionResult} artist={props.artist} />
+        }
+        return null
+      }}
+    />
+  )
 
   beforeEach(() => {
     mockEnvironment = createMockEnvironment()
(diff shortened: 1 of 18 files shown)


src/lib/Scenes/AuctionResult/AuctionResult.tests.tsx
```diff
@@ -1,29 +1,84 @@
...
-      />
-    )
-    mockEnvironmentPayload(mockEnvironment, mockedData)
-    return tree
-  }
+  const TestRenderer = () => (
+    <QueryRenderer<AuctionResultTestsQuery>
+      environment={mockEnvironment}
+      query={graphql`
+        query AuctionResultTestsQuery($auctionResultInternalID: String!, $artistID: String!) @relay_test_operation {
```
olerichter00, line 22: **Idea:** We could add an `AuctionResultFragmentContainer` to `AuctionResult.tsx`. This would allow us to use the fragment in the test instead of writing the whole query.

src/lib/store/config/features.ts
```diff
@@ -201,6 +201,12 @@ export const features = defineFeatures({
     showInAdminMenu: true,
     echoFlagKey: "AREnableVisualProfileIconAndBio",
   },
+  AREnableAuctionResultComparableWorks: {
+    readyForRelease: false,
+    description: "Enable Comparable works",
+    showInAdminMenu: true,
+    // echoFlagKey: "AREnableAuctionResultComparableWorks", // TODO : Need to merge this https://github.com/artsy/echo/pull/173
```
olerichter00, line 208: 👀

olerichter00 (commented):
Looking good! 🎉 

It would be great to see screenshots of the changes :)

src/lib/Scenes/AuctionResult/ComparableWorks.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  return (
+    <Flex testID="comparableWorks">
+      <Text variant="md" my={2}>
+        Comparable Works
+      </Text>
+
+      <FlatList
+        data={auctionResultsByFollowedArtists}
+        keyExtractor={(_, index) => String(index)}
```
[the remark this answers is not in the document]
MounirDhahri: the **key** is the thing that React uses to identify DOM elements. when we push an item to the list or remove something in the middle (and this is probably after a pull to refresh)? If the key is the same as before React assumes that the DOM element represents the same component as before and this is wrong since the key for that element have shifted (++1 or --1)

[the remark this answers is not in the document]
MounirDhahri: Each item should therefore have a permanent and unique property, the id for example.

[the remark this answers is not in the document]
MounirDhahri: we do use the index as a key sometimes when the content is not going to change, and we are dealing with some data with no unique identifier, but I recommend holding off from it if possible to make your code more robust

src/lib/Scenes/AuctionResult/ComparableWorks.tsx
```diff
@@ -0,0 +1,92 @@
...
+        horizontal={false}
```
MounirDhahri, line 35:
```suggestion
```

src/lib/Scenes/AuctionResult/ComparableWorks.tsx
```diff
@@ -0,0 +1,92 @@
...
+        initialNumToRender={3}
+        ItemSeparatorComponent={() => (
+          <Flex py={2}>
```
MounirDhahri, line 38: **issue(non-blocker)**: I think you should have the padding as part of the artist item and not inside the flex here because this will result in 20 non-clickable pixels above and below the separator

src/lib/Scenes/AuctionResult/ComparableWorks.tsx
```diff
@@ -0,0 +1,92 @@
...
+        )}
+        renderItem={({ item, index }) => {
+          if (!item) {
+            return <></>
+          }
+
+          return (
+            <AuctionResultFragmentContainer
+              showArtistName
+              withHorizontalPadding={false}
```
MounirDhahri, line 50: **issue(blocker):** I think we should not have this prop, it adds unnecessary coupling to the auction result and is avoidable by wrapping the component in some padding or even better probably the Flatlist

src/lib/Scenes/AuctionResult/ComparableWorks.tsx
```diff
@@ -0,0 +1,92 @@
...
+const ComparableWorks: React.FC<ComparableWorks> = ({ auctionResult }) => {
+  const { trackEvent } = useTracking()
+
+  const auctionResultsByFollowedArtists = extractNodes(auctionResult.comparableAuctionResults)
+
+  const hasAuctionResults = auctionResultsByFollowedArtists?.length
+
+  if (!hasAuctionResults) {
+    return null
+  }
```
MounirDhahri, line 24: **issue(non-blocker):** I believe we should avoid this pattern as it breaks all the optimizations the Flalist came up with (when we load data or data is removed). We should instead use `ListEmptyComponent` (or probably nothing in conjunction with `ListHeaderComponent`)

src/lib/Scenes/AuctionResult/ComparableWorks.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  return (
+    <Flex testID="comparableWorks">
+      <Text variant="md" my={2}>
+        Comparable Works
+      </Text>
```
MounirDhahri, line 30: **issue(non-blocker)**: This should instead be inside the `Flatlist`'s `ListHeaderComponent`

src/lib/Scenes/AuctionResult/ComparableWorks.tsx
```diff
@@ -0,0 +1,92 @@
...
+  )
+}
+
+export const ComparableWorksFragmentContainer = createFragmentContainer(ComparableWorks, {
+  auctionResult: graphql`
+    fragment ComparableWorks_auctionResult on AuctionResult {
+      comparableAuctionResults(first: 3) @optionalField {
+        totalCount
+        edges {
+          cursor
```
MounirDhahri, line 70: **question:** do we intentionally want no pagination here? if so, then why are we querying for the totalCount and the cursor?

src/lib/Scenes/AuctionResult/AuctionResult.tsx
```diff
@@ -20,6 +21,7 @@ import { graphql, QueryRenderer } from "react-relay"
 import { useTracking } from "react-tracking"
 import { RelayModernEnvironment } from "relay-runtime/lib/store/RelayModernEnvironment"
 import { getImageDimensions } from "../Sale/Components/SaleArtworkListItem"
+import { ComparableWorksFragmentContainer as ComparableWorks } from "./ComparableWorks"
```
MounirDhahri, line 24: **nitpick:** no need to use an alias here as we are moving out of using aliases because they are confusing

MounirDhahri (commented):
Good to see this stale work moving and soon we will be having comparable auction results. I left some comments in the PR, let me know if anything wasn't clear

src/lib/Scenes/AuctionResult/ComparableWorks.tsx
```diff
@@ -0,0 +1,92 @@
...
+        }}
+      />
+    </Flex>
+  )
+}
+
+export const ComparableWorksFragmentContainer = createFragmentContainer(ComparableWorks, {
+  auctionResult: graphql`
+    fragment ComparableWorks_auctionResult on AuctionResult {
+      comparableAuctionResults(first: 3) @optionalField {
```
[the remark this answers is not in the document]
olerichter00: I guess the argument to use it is that the section is indeed optional and it would be better to not show it instead of showing an "Unable to load" screen in the case something goes wrong. If this query fails often in case there is a bug in one of the involved services we would see the error anyway in DataDog. 

But indeed this is a discussion with good arguments on both sides.

src/lib/Scenes/AuctionResult/ComparableWorks.tsx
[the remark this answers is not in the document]
rajsam003: Thankyou for explaining

src/lib/Scenes/AuctionResult/ComparableWorks.tsx
```diff
@@ -0,0 +1,92 @@
...
+        initialNumToRender={3}
+        ItemSeparatorComponent={() => (
+          <Flex py={2}>
+            <Separator borderColor="black10" />
+          </Flex>
+        )}
+        renderItem={({ item, index }) => {
+          if (!item) {
+            return <></>
+          }
```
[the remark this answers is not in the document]
rajsam003: Best idea 💡

src/lib/store/config/features.ts
```diff
@@ -201,6 +201,12 @@ export const features = defineFeatures({
     showInAdminMenu: true,
     echoFlagKey: "AREnableVisualProfileIconAndBio",
   },
+  AREnableAuctionResultComparableWorks: {
+    readyForRelease: true,
```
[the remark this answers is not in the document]
MounirDhahri: Yes, I agree but I am afraid if we do this often we start to always bring features that might not be working fully and it could bite us sometime

src/lib/Scenes/AuctionResult/ComparableWorks.tsx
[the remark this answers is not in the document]
MounirDhahri: Yes,  this makes sense. thanks for explaining. I will resolve this then 🙌

src/lib/store/config/features.ts
```diff
@@ -201,6 +201,12 @@ export const features = defineFeatures({
...
+    description: "Enable Comparable works",
```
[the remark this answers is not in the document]
rajsam003: The text gets overflowed on screen if its lengthy.

src/lib/store/config/features.ts
rajsam003: 😂

src/lib/Scenes/AuctionResult/ComparableWorks.tsx
rajsam003: I dont think we can do it without it. This FlatList is used in AuctionResult.tsx and there if I remove px={2}, everything gets messy

src/lib/store/config/features.ts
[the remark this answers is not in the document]
MounirDhahri: aha makes sense. I will remove resolve this then. thanks

src/lib/Scenes/AuctionResult/ComparableWorks.tsx
MounirDhahri: thanks for explaining privately, we can resolve this then

src/lib/store/config/features.ts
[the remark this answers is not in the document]
olerichter00: You're right, we should be careful.

MounirDhahri (dismissed):
nice, thanks for the changes 🙌

MounirDhahri (approved):

PR #5741, opened 2021-11-01
feat(dx): feature flag check script + cron

brainbicycle:

### Description

little future friday DX idea, based on our retro discussion add a reminder to mobile practice to check feature flags before release:

- Adds a js script for reading live and hidden feature flags from the project, outputs to a file flags.json
- Adds a fastlane script to call this script, read the file and send a slack message
- Adds a circle cron to try to run every Thursday
- Adds a wrapper bash script to only run every _other_ week

### Follow-ups:

- [x] some tests, specifically around regex
- [ ] would be nice to also have a stale feature flag check, e.g. a feature has been live for ~2 releases remove the flag

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- Add a biweekly reminder about feature flags - Brian

</details>


ArtsyOpenSource:

- Dev changes (Add a biweekly reminder about feature flags - Brian)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against cd26e051349d3eb9e14959ed124f359b43dfb10c
</p>

scripts/check-flags/check-flags.js
```diff
@@ -0,0 +1,96 @@
...
+    if (!result.groups) {
+      break
+    }
+    if (result.groups['flag']) {
+      const currentFlag = result.groups['flag']
+      allFlags.push(currentFlag)
+    }
+  }
+  return allFlags.filter(flag => !releasedFlags.includes(flag));
+}
```
pvinis, line 91: I see that released is not using the matchAll, and hidden does. Maybe we could make released and hidden functions use the same structure? Maybe they could do the same thing (match all) and then filter for true or false?

Does that make sense? Now they look kinda hard to read 🤔

src/lib/store/config/features.tests.ts
```diff
@@ -29,3 +29,30 @@ describe("features and devToggles", () => {
...
+  /**
+   * These are pretty silly tests:
+   * - enforcing a convention
+   * - a file location
+   * - something that is already enforced with typescript
+   * They are meant to protect this script from breaking:
+   * ./scripts/check-flag.js, if you want to update these conventions
+   * please also update the script!
+   */
+  it("should start with AR prefix", () => {
```
pvinis, line 43: nice tests 👌

scripts/check-flags/check-flags.tests.ts
```diff
@@ -0,0 +1,121 @@
...
+          readyForRelease: true,
+          echoFlagKey: "AROptionsNewFirstInquiry",
+        },
+      `
+      const releasedFlags = parseReleasedFlags(flagContent)
+      expect(releasedFlags).toEqual(["AROptionsArtistSeries", "AROptionsNewFirstInquiry"])
+    })
+  })
+  describe("hidden flags parsing", () => {
+    it("defaults to returning all flags", () => {
```
pvinis, line 81: If we change the parseHiddenFlags to not get the released as an arg, then we could also change this test to not show the `readyForRelease: true` flag 🤔

scripts/check-flags/check-flags.js
brainbicycle: I agree this would be better if hidden was just the opposite of released but I couldn't find a good way to express that via regex. the problem is the readyForRelease flags we need to match all flags that contain **readyForRelease** and then filter those that are true but the hidden flags are not just those that have **readyForRelease: false**, they also need to include those that don't have the **readyForRelease** key at all which was the reasoning for just getting all the flag keys and filtering those that we found as released earlier if that makes sense. That said there is probably a better way.

pvinis: ah gotcha. I might try too, I'm always curious with regex

.circleci/config.yml
```diff
@@ -260,6 +260,24 @@ jobs:
...
+
+    macos:
+      xcode: 13.0.0
+
+    steps:
+      - checkout
+      - generate-checksums
+      - install-node
+      - install-gems
+      - install-node-modules
```
MounirDhahri, line 275: thought: do we necessarily need all node modules?

MounirDhahri (approved):
really nice PR 👏

scripts/check-flags/check-flags.tests.ts
brainbicycle: yeah I agree writing this didn't feel great, hiddenFlags defaults to returning all flags unless you pass released flags is confusing, open to better ways

brainbicycle:
Alright fixed the export and I think this should be good to go, I might also try to take a crack still at making the hidden flags bit a bit more readable because I agree it is not great now

.circleci/config.yml
brainbicycle: good catch I don't think we do!

scripts/check-flags/check-flags.js
brainbicycle: 🤔 I might just have been being dumb, this is a required key on typescript side and we have the test now to check even if we change that on typescript side so we don't really have to handle key absence, updated much cleaner, thanks for suggesting ✨

MounirDhahri (approved):
🙌

PR #4471, opened 2021-02-23
chore: refactor filters modal to use easy-peasy

MounirDhahri:
The type of this PR is: **refactor**

### Description

**Goal from this refactoring:**
- Facilitate adding new filters
- Make the filtering API easier to digest
- Use easy-peasy instead of React context API
- Ability to easily persist filters in the future (design asked about this)

### Refactoring plan:

**Step1: Duplicate the existing logic but using easy-peasy**
- [x] Duplicate everything related to Artworks Filters
- [x] Implement store model using easy-peasy on ArtworkFiltersStore2
- [x] Make sure to make no logic changes (refactor one step a time)
- [x] Implement t ArtworkHelpers2 based on ArtworkFiltersStore2 and remove everything not needed
- [x] Make the necessary adjustments on the FilterModal2
- [x] Use new logic on one of the screens (I picket the sale screen)
- [x] Comment all options screens and pick an easy one to make sure nothing (I used ViewAsOptionsScreen)
- [x] Adjust Filter Modal tests and make sure nothing is broken
- [x] Migrate the remaining options screens to use the new store
- [x] Refactor the tests and make sure nothing is broken
- [x] Migrate the remaining screens to use the new store
___
**Auction filters**

https://user-images.githubusercontent.com/11945712/114587903-6476f380-9c86-11eb-88b1-d9489546b0b8.mov

___
**Artwork filters**

https://user-images.githubusercontent.com/11945712/114590016-883b3900-9c88-11eb-8e05-b1d0898ffafb.mp4

___
**Collection filters**
(the ways to buy filter has a bug non-related to this PR and is being fixed)

https://user-images.githubusercontent.com/11945712/114588806-5d041a00-9c87-11eb-8ec5-201c74b65b90.mov

___
**Fairs filters**

https://user-images.githubusercontent.com/11945712/114587910-680a7a80-9c86-11eb-96d5-aa9be4f900ad.mov

___
**Shows filters**

https://user-images.githubusercontent.com/11945712/114588063-8cfeed80-9c86-11eb-9351-4d3a32eb0178.mov

___
**Auction Results filters**

https://user-images.githubusercontent.com/11945712/114588116-9e47fa00-9c86-11eb-97c1-74ec4dcee7f2.mov

### PR Checklist (tick all before merging)

- [x] I have included screenshots or videos to illustrate my changes, or I have not changed anything that impacts the UI.
**__I tested all filters manually and everything seems to be working 🎉__**
[...]

PR #5742, opened 2021-11-01
feature(mycollection): add sort and filter in header

pvinis:

This PR resolves [CX-2136]

### Description

This PR basically gets rid of some old headers and separators, and replace with the newest one. Then it adds the ability for filters to work with a local setup.

Disclaimer: This code is just a patch on top of the filter store. I tried to not refactor anything to simplify. I tried to keep the same "api" for the sortOptions and filterOptions.

Happy to pair to explain further if needed.

### To the reviewers 👀

- sort and filter in mycollection (behind flag) - pavlos

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-2136]: https://artsyproduct.atlassian.net/browse/CX-2136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 3d6f06daa9f1d9eb33ec370d27419f2a244868d4
</p>

src/lib/Components/ArtworkGrids/InfiniteScrollArtworksGrid.tsx
```diff
@@ -248,7 +250,10 @@ class InfiniteScrollArtworksGrid extends React.Component<Props & PrivateProps, S
       sectionedArtworks.push([])
```
pvinis, line 250: check if it works with 100 artworks

pvinis:
we do have pagination in the end, so i think increasing the page size is the easiest solution right now. removing the pagination seems counter-productive. in order for the local processing to work we need to pull all (or as much as possible) data so that we have values to give to the filters.

So I ended up doing something like this https://github.com/artsy/eigen/pull/5742/files#diff-711500e0a24b1933f2661d17ecef6a7f95dc64efda954f07f2917d9116211ce0R321.

Weird? Yes. Best we can do for now? I guess.

src/lib/Components/ArtworkFilter/ArtworkFilterOptionsScreen.tsx
```diff
@@ -49,6 +49,10 @@ export interface FilterDisplayConfig {
   filterType: FilterScreen
   displayText: string
   ScreenComponent: keyof ArtworkFilterNavigationStack
+
+  // for `local` filtering
+  values?: any[]
+  localSortAndFilter?: (items: any[], value?: any) => any[]
```
brainbicycle, line 55: **thought (non-blocking):** 
- would be nice to get rid of anys here, I know we tried
- if some parts of this type are specific and required for local filters would be nice to make an either type where if you have a local filter screen you have to provide these props and if you have a regular filter screen you don't

src/lib/Components/ArtworkFilter/ArtworkFilterStore.tsx
```diff
@@ -1,6 +1,7 @@
 import { Action, action, createContextStore, State } from "easy-peasy"
 import { assignDeep } from "lib/store/persistence"
 import { filter, find, isEqual, unionBy } from "lodash"
+import { FilterDisplayConfig } from "."
```
brainbicycle, line 4: **nitpick (non-blocking):** hmm, I haven't seen an import like this, can we make file explicit?

src/lib/Components/ArtworkFilter/FilterModal.tests.tsx
```diff
@@ -20,7 +20,7 @@ import { GlobalStoreProvider } from "lib/store/GlobalStore"
 import { extractText } from "lib/tests/extractText"
 import { mockEnvironmentPayload } from "lib/tests/mockEnvironmentPayload"
 import { renderWithWrappers } from "lib/tests/renderWithWrappers"
-import { Sans, Theme } from "palette"
+import { bullet, Sans, Theme } from "palette"
```
brainbicycle, line 23: nice

src/lib/Components/ArtworkGrids/InfiniteScrollArtworksGrid.tsx
```diff
@@ -109,6 +109,8 @@ export interface Props {
 
   /** Whether to add the tapped artwork to recent searches */
   updateRecentSearchesOnTap?: boolean
+
+  localSortAndFilterArtworks?: (artworks: any[]) => any[]
```
brainbicycle, line 113: **non-blocking**: same comment on any, I might try this afternoon but shouldn't block PR

brainbicycle (approved):
Great work! I think it is actually quite a nice api for local filters even if our filter code is pretty complex

brainbicycle:
> we do have pagination in the end, so i think increasing the page size is the easiest solution right now. removing the pagination seems counter-productive. in order for the local processing to work we need to pull all (or as much as possible) data so that we have values to give to the filters.
> 
> So I ended up doing something like this https://github.com/artsy/eigen/pull/5742/files#diff-711500e0a24b1933f2661d17ecef6a7f95dc64efda954f07f2917d9116211ce0R321.
> 
> Weird? Yes. Best we can do for now? I guess.

Oof, I thought for sure we checked that, so on the UX side if a user has more than 100 artworks and tries to filter?

pvinis:
another test fixed.

src/lib/Scenes/Search/SearchArtworksGrid.tests.tsx
```diff
@@ -89,7 +89,7 @@ describe("SearchArtworksGrid", () => {
 
     mockEnvironmentPayload(environment)
 
-    expect(getByText("Filter")).toBeTruthy()
+    expect(getByText("Sort & Filter")).toBeTruthy()
```
pvinis, line 92: Honestly, I don't understand how this was passing before. It's not the first time our tests feel like they break randomly. I'd like to investigate. Maybe next KS or something.

pvinis, line 92: Honestly, I don't understand how this was passing before. It's not the first time our tests feel like they break randomly. I'd like to investigate. Maybe next KS or something.

src/lib/Components/ArtworkFilter/ArtworkFilterStore.tsx
pvinis: We have a few like it, more in palette. We can change it, but this is also ok. It usually goes like that when there is an `index` file. And unfortunately the filter store has an index file 🤷.

src/lib/Components/ArtworkFilter/ArtworkFilterOptionsScreen.tsx
pvinis: I agree, but since I'm off next week, if you didn't figure anything out last evening, should we just make a ticket pointing to this? Or something in `HACKS.md`?

pvinis:
For the 100 artworks, since we dont have people with that many yet, let's not block, and put a ticket on jira and schedule it soon/high prio, so we can take a look before we get there?

src/lib/Scenes/MyCollection/MyCollection.tsx
```diff
@@ -55,10 +63,94 @@ const MyCollection: React.FC<{
...
+  const filterOptions = ArtworksFiltersStore.useStoreState((s) => s.filterOptions)
+
+  useEffect(() => {
+    setFilterType("local")
+    setSortOptions([
+      {
+        paramName: FilterParamName.sort,
+        displayText: "Alphabetical Artist Name (A-Z)",
+        paramValue: "local-alpha-a-z",
+        // tslint:disable-next-line: no-shadowed-variable
```
olerichter00, line 87: 👀

src/lib/Components/ArtworkFilter/ArtworkFilterHelpers.ts
```diff
@@ -79,6 +79,7 @@ export const getSortDefaultValueByFilterType = (filterType: FilterType) => {
     auctionResult: "DATE_DESC",
     geneArtwork: "-partner_updated_at",
     tagArtwork: "-partner_updated_at",
+    local: "",
```
olerichter00, line 82: I'm not quite sure if I understand why `local` is another filter type. Shouldn't it be that any filter type could be `local` even if we just use `local` for "My Collection" artworks now?

olerichter00 (commented):
Looks good to me 👍

MounirDhahri (commented):
Nice one 👏 I left few small comments

src/lib/Scenes/MyCollection/MyCollection.tsx
pvinis: I've said before that shadow rule should go away.. it's not useful.

src/lib/Components/ArtworkFilter/ArtworkFilterHelpers.ts
pvinis: I'm more sure I understand what you mean 😬

src/lib/Components/ArtworkFilter/ArtworkFilterStore.tsx
```diff
@@ -21,21 +22,26 @@ export interface ArtworkFiltersModel {
...
-  selectFiltersAction: Action<ArtworkFiltersModel, FilterData>
-  resetFiltersAction: Action<ArtworkFiltersModel>
-  clearFiltersZeroStateAction: Action<ArtworkFiltersModel>
-  setAggregationsAction: Action<ArtworkFiltersModel, any>
-  setFiltersCountAction: Action<ArtworkFiltersModel, FilterCounts>
-  setFilterTypeAction: Action<ArtworkFiltersModel, FilterType>
-  setInitialFilterStateAction: Action<ArtworkFiltersModel, FilterArray>
-  setSelectedFiltersAction: Action<ArtworkFiltersModel, FilterArray>
+  sortOptions?: FilterData[]
+  filterOptions?: FilterDisplayConfig[]
```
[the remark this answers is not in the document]
pvinis: true. the naming on this whole file is kinda not great, so I made it simple. I hope at some point when we rework it, this name will become more accurate?

src/lib/Scenes/MyCollection/MyCollection.tsx
```diff
@@ -55,10 +63,94 @@ const MyCollection: React.FC<{
...
   const [isRefreshing, setIsRefreshing] = useState(false)
 
+  const setFilterType = ArtworksFiltersStore.useStoreActions((s) => s.setFilterTypeAction)
+  const setSortOptions = ArtworksFiltersStore.useStoreActions((s) => s.setSortOptions)
+  const setFilterOptions = ArtworksFiltersStore.useStoreActions((s) => s.setFilterOptions)
+  const filterOptions = ArtworksFiltersStore.useStoreState((s) => s.filterOptions)
+
+  useEffect(() => {
+    setFilterType("local")
+    setSortOptions([
```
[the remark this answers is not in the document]
pvinis: you mean together with the filteroptions?

src/lib/Scenes/MyCollection/MyCollection.tsx
```diff
@@ -55,10 +63,94 @@ const MyCollection: React.FC<{
...
+            artworks,
+            (a) => {
+              const date = DateTime.fromISO(a.date)
+              return date.isValid ? date.toMillis() : 0
+            },
+            "desc"
+          ),
+      },
+    ])
+    setFilterOptions([
```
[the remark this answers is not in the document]
pvinis: I don't agree again. that's the reason we got in this messy code. the filters should be fluid and easy to change. here is not the ideal place either, but it's better than keeping them in some other file, especially when the filtering step function contains logic that makes sense to be here.

src/lib/Components/ArtworkFilter/ArtworkFilterHelpers.ts
olerichter00: Well, why local is a `FilterType`. Couldn't any of the other filter types be local as well?

olerichter00: And would it make sense to call this type `myCollectionArtwork` instead of local? This would make more sense to me looking at this helper. But I'm not sure if this would work 🤔

pvinis: the reason bits not called myollection is that I'm trying to get us out of this pattern that made us have complicated filter code. the store doesn't need to know where we use it and how. it just needs to do it's work of making the filters work. the actual application of the filters can be elsewhere.

my idea with local is that for now we can use local to make filters that have this API of handling them in the component we use them, and later we could remove all the old code that predefines filters per usage, and make it all local. then we can remove the name local.

I hope this makes sense.

olerichter00: I'm not sure about the part of making all filters local. Filtering local only works for "My Collection" artworks because the number of artworks is limited. In other places, we would have to load a big amount of artworks before we could filter/sort them. This approach unfortunately wouldn't scale.

pvinis: right, I missed that part on the explanation. sorry for that.

when I say "do the rest like local" I don't mean do the filtering local for everything. I mean do the code like local. so keep the filters and the paramNames and values in the screen we use them rather than in the filterstore. the filterstore will still be the thing that does the actual handling.

so in artistartworks we can have a setFilterOptions that doesn't have a localfilter step, but has just the MP paramValue. so the code is kept in the place that makes sense, and the filterstore is the one that would put the MP string together to send to MP, so the actual filtering will happen in MP for that screen.

so what I meant by "like local" is only for the code. not the filtering. the filtering will support both local and MP.

src/lib/Components/ArtworkFilter/ArtworkFilterStore.tsx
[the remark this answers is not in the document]
MounirDhahri: I am fine with the file name, it's a store and it contains a store, it seems logical to me. I was only referring to `filterOptions` as it's only used for local filters now.

src/lib/Components/ArtworkFilter/ArtworkFilterHelpers.ts
olerichter00: Thanks a lot for the explanation. Local in this context makes sense to me 🙇

src/lib/Scenes/MyCollection/MyCollection.tsx
```diff
@@ -55,10 +63,94 @@ const MyCollection: React.FC<{
...
+        values: uniqBy(
+          artworks.map(
+            (a): FilterData => ({
+              displayText: a.medium ?? "N/A",
+              paramName: FilterParamName.additionalGeneIDs,
+              paramValue: a.medium ?? undefined,
+            })
+          ),
+          (m) => m.paramName
+        ),
```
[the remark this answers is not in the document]
MounirDhahri: This looks to me as a UI logic that is tied to the `AdditionalGeneIDsOptionsScreen` screen and I don't see why it needs to be inside MyCollection. Adding it here will actually make it more coupled to my collection and reduce its cohesion with the `AdditionalGeneIDsOptionsScreen` modules. 

I agree with the method above being here as we might need to implement different methods of filtering locally if we need to somewhere else (I hope we don't).

src/lib/Scenes/MyCollection/MyCollection.tsx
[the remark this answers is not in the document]
MounirDhahri: I don't feel the code is messy, it was just not designed to do local filtering which I believe makes sense. I definitely don't agree with some areas of how the filters work but introducing more complexity now is avoidable IMO and I think we should move those filtering screens to be in the same file as the other screens for consistency and then if we need to refactor the filters, we can move the logic to the screens but I don't think that's the right way to  do it as it will unify the User Interface Logic and the Business Logic.

src/lib/Scenes/MyCollection/MyCollection.tsx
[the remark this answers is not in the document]
MounirDhahri: my comment here is related to the actions that sets the sort options from my collection. What I meant is, if we set those sort options on the SortOptions screen itself, then we won't need to have an action and we will still keep our store small.

src/lib/Scenes/MyCollection/MyCollection.tsx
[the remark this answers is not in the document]
MounirDhahri: we did some work initially around refactoring the filters [here](https://github.com/artsy/eigen/pull/4471) and [here](https://github.com/artsy/eigen/pull/4629). You can check it out in case you would like to work on rewriting the filters at a future time

pvinis:
as I'm off this whole week, if we need this soon, everyone feel free to rename things etc. otherwise I can work on this next week.

src/lib/Components/ArtworkFilter/ArtworkFilterStore.tsx
brainbicycle: alright not a huge deal either way

PR #5778, opened 2021-11-10
feat: User sees initial Artworks of the Day widget GRO-663

jonallured:
This PR adds an iOS widget to Eigen - looks something like this:

<details>

IMG_0366

Simulator Screen Shot - iPhone 12 Pro - 2021-11-09 at 15 34 17

</details>

I did a discovery doc here that includes lots more discussion:

https://www.notion.so/artsy/Technical-Discovery-of-iOS-Widgets-bc43497399754899bd22347f14d34b53

But I do want to note a few things here:

* ~The second commit in this PR is a WIP that I don't understand - why does my Podfile always have churn when I'm just pulling from master??~
* Brand is going to be taking a look at this so I'm deferring any design-related feedback until I hear back from them - then go wild telling me that it's weird looking!! 😝 

Things that are left to do:

- [x] update name and description of widget for add screen
- [x] verify fallback artist/artwork are cleared
- [x] add test coverage with my pal @brainbicycle 
- [x] add a POST to volley with the widget type to track timeline computation

When it comes to the visual QA, known issues include:

- [x] medium version is broken because the logo and text are off-canvas
- [x] large version has spacing issues with logo and text
- [x] text can be illegible - how would we like to address this?

https://artsyproduct.atlassian.net/browse/GRO-663

/cc @artsy/grow-devs @artsy/collector-experience


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  ❌ **It seems like you added some non WebP images to Eigen, please convert them to WebPs using `source images/script.sh` script **
  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against eb08ddfb42b63478568d4741015924f9e94248a6
</p>

kathytafel:
For the visual QA/illegible text
Have you looked at it in both light and dark mode? Is it illegible in both modes? Why is it illegible? 
I'm imagining it's because of whatever's in the lower left of an image. 
I'm a huge fan of 'cheating'. Are we manually curating the feed that I see on observables? Is a human faster than a machine? Then only select images where the text will be legible as part of the curation process.
If I wanted to make an algorithm, maybe I would during a hackathon work with the ML engineers to extract that as a feature and train a set of photos on what is legible and not legible, and then IDK somehow make that image processing code part of the ingest stream. Might be overkill.

jonallured:
> I'm imagining it's because of whatever's in the lower left of an image.

Yep that's right. I had a really good example of how terrible it can look but I seem to have misplaced it - we can probably imagine though.

The way the feed of widget artworks is created is based on a gene called "Trove" that powers this page:

https://www.artsy.net/gene/trove

So that's my "cheat"! I believe it's the curatorial team and they manually pick those works so we could def update their guidance to include checking for works that look real real bad in the widget. The engineer in me wants to do some fancy thing like the visual effect you linked to but yeah just picking better images could work too!

admbtlr:
The standard workaround here is to add a little gradient strip that goes from like 0.3 (at the bottom) to 0 (about 100px or whatever from the bottom) opacity across the bottom of the image, just to slightly darken the area under the text. If you play around a bit with the opacity and the height of the strip, you can usually get the text to be legible without messing up the image too much. You would probably then need to reduce the padding around the text so that it moves down to the bottom, but that might not be a bad thing anyway, since you want to hide as little of the image as possible.

You can actually see a similar effect on the Carrot widget in your screenshot above.

jonallured:
Going to tinker with this blur idea for an hour or so but also what if we just remove the text altogether? Or remove it at the smallest size?

jonallured:
Oh and this is about the worst I've seen it be:

<details>

Simulator Screen Shot - iPhone 12 Pro - 2021-11-11 at 12 20 23

</details>

kathytafel:
I honestly think curation is the key. anything else takes away from the art. especially at small size. You might be onto something to delete the text at small size. If you put a banner under the text then that's all your eye would see.

ArtsyWidget/Info.plist
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>NSExtension</key>
+	<dict>
+		<key>NSExtensionPointIdentifier</key>
+		<string>com.apple.widgetkit-extension</string>
```
brainbicycle, line 8: 🎉

ArtsyWidget/Models/Artist.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+
+public struct Artist: Decodable {
+    public let name: String
+    
+    public init(name: String) {
+        self.name = name
+    }
```
brainbicycle, line 8:
```suggestion
```
don't actually think you need this init method, should get one automatically with the property!

ArtsyWidget/Models/Artwork.swift
```diff
@@ -0,0 +1,27 @@
...
+        let url = URL(string: imageUrl)!
+        return url
+    }
+    
+    public init(id: String, title: String, artist: Artist, images: [ArtworkImage]) {
+        self.id = id
+        self.title = title
+        self.artist = artist
+        self.images = images
+    }
```
brainbicycle, line 24:
```suggestion
```
similarly don't think you need this

ArtsyWidget/Models/ArtworkImage.swift
```diff
@@ -0,0 +1,14 @@
...
+    public static func fallback() -> ArtworkImage {
+        let fallbackUrl = "https://d32dm0rphc51dk.cloudfront.net/02XpcHIJyIsmdFj5u_2nOw/:version.jpg"
+        return ArtworkImage(imageUrl: fallbackUrl)
+    }
+    
+    public let imageUrl: String
+    
+    public init(imageUrl: String) {
+        self.imageUrl = imageUrl
+    }
```
brainbicycle, line 13:
```suggestion
```
same comment, but you should make sure it compiles without has be a while

ArtsyWidget/Widget/Entry.swift
```diff
@@ -0,0 +1,38 @@
...
+        self.date = date
+    }
+    
+    init(artistName: String, artworkImage: UIImage, artworkSlug: String, artworkTitle: String, date: Date) {
+        self.artistName = artistName
+        self.artworkImage = artworkImage
+        self.artworkSlug = artworkSlug
+        self.artworkTitle = artworkTitle
+        self.date = date
+    }
```
brainbicycle, line 25:
```suggestion
```

this one you should get automatically, custom one you need

Podfile.lock
```diff
@@ -825,49 +825,27 @@ DEPENDENCIES:
...
 SPEC REPOS:
-  https://github.com/artsy/Specs.git:
-    - Aerodramus
-    - "Artsy+UIColors"
-    - "Artsy+UIFonts"
-    - "Artsy+UILabels"
-    - Artsy-UIButtons
-    - Extraction
-  trunk:
+  https://cdn.cocoapods.org/:
```
brainbicycle, line 828: I think this churn can be avoided: https://www.willowtreeapps.com/craft/how-to-fix-unexpected-podfile-lock-diffs-after-pod-install

brainbicycle (dismissed):
Looks great! I think you can remove some of the inits that match the properties on structs but will want to confirm it still compiles. The source changing in the podfile.lock is something we keep running into, attached an article to the comment there but I think it just means you need to remove an old source from your pod repo settings. Let me know if I can help.

jonallured:
Ok one step forward one step back: I have added the Volley code so we can see usage over time but I did accidentally clobber those tests we added @brainbicycle! 😬  I've still got them in my local reflog so I'll dig them back out tomorrow and get this PR updated.

But more importantly I've heard back from Design and Brand and here's my plan:

* only provide the large full bleed version and remove the logo from this one, looks like this:

<details>

simulator_screenshot_7B7A3599-7CF4-4CEF-BD85-32A28714733E

</details>

* follow up quickly with the featured artworks type of widget, looks like this:

<details>


Screen Shot 2021-11-18 at 2 11 44 PM

</details>

* stretch goalllllll -> get an editorial widget landed too, looks like this:

<details>


Screen Shot 2021-11-18 at 4 34 51 PM

</details>

This one is tricky because I want to use the RSS feed from editorial but that means parsing XML. 😭 

Anyway, getting real close pals!!

ArtsyWidget/Shared/PrimaryText.swift
```diff
@@ -0,0 +1,14 @@
+import SwiftUI
+
+struct PrimaryText: View {
+    let name: String
+    var color: Color = .black
+    
+    var body: some View {
+        let font = Font.system(size: 12, weight: .medium)
```
pvinis, line 8: i wonder if we care/want to use our fonts here 🤔, probably not a big issue for v1 though.

ArtsyWidget/Shared/Schedule.swift
```diff
@@ -0,0 +1,43 @@
+import Foundation
+
+struct Schedule {
```
pvinis, line 3: magical 🪄

ArtsyWidget/Shared/PrimaryText.swift
jonallured: I spent about an hour one day trying to get a custom font to show up on a widget but it never worked. I'm not sure if it's unsupported or what but I bailed. 🤷

Podfile.lock
jonallured: Worked like a charm, thanks so much!!

ArtsyWidget/Shared/PrimaryText.swift
MounirDhahri: you can do that probably later. I think pavlos is working on having the fonts as part of the project and not the artsy-fonts pod and that will make using them hopefully straightforward for you later

MounirDhahri (approved):
LGTM 👏 I left 1 minor comment

ArtsyWidget/Models/Artwork.swift
```diff
@@ -0,0 +1,42 @@
+import Foundation
+import UIKit
+
+struct Artwork: Codable {
+    static func fallback() -> Artwork {
+        let artist = Artist(name: "Chloe Wise")
```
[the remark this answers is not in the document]
jonallured: Good call I was def not handling this case at all. 👍  I was hoping there'd be something in Gravity's default JSON document for this but maybe not? I can get the names like so:

```
$ http "https://api.artsy.net/api/v1/artwork/man-ray-exhibition-man-ray-1945-exhibition-catalogue-julien-levy-gallery-nyc-designed-by-duchamp-rare?access_token=$USER_TOKEN" | jq .artists[].name
"Man Ray"
"Marcel Duchamp"
```

But then I'm left with having to combine those strings together ugh why is nothing easy. 🙄

brainbicycle (approved):
🥇 🔥

ArtsyWidget/Models/Artwork.swift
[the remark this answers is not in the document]
MounirDhahri: nice solution 🙌

PR #5813, opened 2021-11-22
feature(tests): Easier relay tests

pvinis:

This PR resolves []

### Description

Relay tests are always a bit of a pain. This PR tried to make them slightly less of a pain.

In tests, things like this pop up:
```ts
mockEnvironmentPayload(mockEnvironment, {
  Me: () => ({
    lotStandings: [],
  }),
})
```
and they seem weird. The graphql query is:
```ts
{
  me: {
    lotStandings
  }
}
```

**So why do we make `me: {` into `Me: () => ({`?**

Well, because relay is weird, and because its a bit of a leaky abstruction and because its trying to do many things and unluckily the end result API is kinda funky.

So this PR fixes this. We still support the exact format. No tests break. In fact, we need to support this format for some very few special cases where we need some fancy mocking like https://github.com/artsy/eigen/blob/master/src/lib/Components/Artist/ArtistAbout/ArtistAbout.tests.tsx#L44-L50 where `Boolean` is a special graphql thing like `ID` and `String` here https://github.com/artsy/eigen/blob/master/src/lib/tests/mockEnvironmentPayload.ts#L45-L48.
Just to repeat, this is still supported, and needed for these special cases.

**But for all other cases, where we just do `Me: () => ({`, now we can simply do `me: {`!***

The above test can now be written as:
```ts
mockEnvironmentPayload(mockEnvironment, {
  me: {
    lotStandings: [],
  },
})
```

Yey.
 

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- making relay mocking more sane - pavlos

</details>

### DIFF
--- a/src/lib/Components/Artist/ArtistHeader.tests.tsx
+++ b/src/lib/Components/Artist/ArtistHeader.tests.tsx
@@ -41,9 +41,7 @@ describe("ArtistHeader", () => {
   it("renders properly", () => {
     const tree = renderWithWrappers(<TestRenderer />)
 
-    mockEnvironmentPayload(mockEnvironment, {
-      Artist: () => mockArtist,
-    })
+    mockEnvironmentPayload(mockEnvironment, mockArtist)
 
     expect(tree.root.findAllByType(Sans)[0].props.children).toMatch("Marcel Duchamp")
   })
@@ -51,21 +49,17 @@ describe("ArtistHeader", () => {
   it("displays follow button for artist", () => {
     const tree = renderWithWrappers(<TestRenderer />)
 
-    mockEnvironmentPayload(mockEnvironment, {
-      Artist: () => mockArtist,
-    })
+    mockEnvironmentPayload(mockEnvironment, mockArtist)
 
     expect(tree.root.findAllByType(Button)[0].props.children).toMatch("Follow")
   })
 
   it("does not show followers count when it is < 2", () => {
-    mockArtist.counts.follows = 1
+    mockArtist.artist.counts.follows = 1
 
     const tree = renderWithWrappers(<TestRenderer />)
 
-    mockEnvironmentPayload(mockEnvironment, {
-      Artist: () => mockArtist,
-    })
+    mockEnvironmentPayload(mockEnvironment, mockArtist)
 
     const text = extractText(tree.root)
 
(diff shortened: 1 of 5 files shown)


ArtsyOpenSource:

- Dev changes (making relay mocking more sane - pavlos)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 2847568f7152763f490140f71d0df4cdffd50386
</p>

araujobarret (approved):
Thanks for the effort clearing those tests 🚀 

I have more like an open question regarding the approach testing relay components, usually we vote for testing the fragments rather than the QueryRenderers, so why not use our structure from this [guy](https://github.com/artsy/eigen/blob/master/src/lib/tests/setupTestWrapper.tsx#L43) and update the tests you have to touch with `setupTestWrapper`? This would save some time of `creatMoEnvironment` that we already do in the wrapper.

MounirDhahri:
Although I agree that this is an easier syntax to save some text, I think it adds an extra level of abstraction to something that I believe requires no further abstraction. I personally, would feel more comfortable seeing an API that resembles the original documentation and using it directly regardless of needing to use a context or not instead of adding an additional layer of complexity that would make new joiners lost initially

araujobarret:
@pvinis that @MounirDhahri said is what we chatted about yesterday. The expected mock is a resolver type like `Me, Artwork`, the library is expecting a mocked resolver there not an object.
I also don't think it's a good idea to abstract this point 😢

damassi:
I agree with @MounirDhahri and @araujobarret on this one. When testing fragments, its easy enough to see the type:

```
fragment Foo on Artist {
  name
  bio
}
```
Which corresponds exactly to the type in the test:
```tsx
Artist: () => ({
  name: "Hello",
  bio: "World"
})
```

It's as straight forward as it can be, IMO. And since types are predominate in graphql, this keeps things closer to the original design.

alloy:
I think you may actually be able to do this with MockPayloadGenerator out-of-the-box. Recall that a root field is but just a field on the query type, e.g.:

```ts

mockEnvironmentPayload(mockEnvironment, {
  Query: () => ({
    me: {
      lotStandings: [],
    }
  }),
})
```

alloy:
(I also tend to agree that staying as close to the sparse docs is probably preferable.)

alloy:
A thing I’ve found helpful from time to time is to generate a default mock with MockPayloadGenerator and then annotate the result using the `@raw_response_type` typing to further modify it in a surgical, type-safe, and DX friendly manner.

pvinis:
I will close this PR. Seems like we are not in agreement, and I think this comment https://github.com/artsy/eigen/pull/5813#issuecomment-991317759 would be ok to use to simplify, by wrapping in `Query: () => ({` in our helper.

Thanks everyone for your feedback 🙌.

PR #5815, opened 2021-11-22
feat: User sees featured artworks widget GRO-673

jonallured:
This PR introduces a second type of widget - I'm calling it `FeaturedArtworksWidget`. I'm adding this type to the existing `FullBleedWidget` and you can see them together here:

<details>

2B2C00F0-AF5E-4D9B-9EF6-7E6428B2CFFD

</details>

## Rotating Featured Artworks

In order to make the Featured Artworks Widget feel lively and fun I have introduced a rotation strategy. The way it works is that I find the same four artworks from the Full Bleed Widget and load them into an array. I then rotate that array one at a time so each work ends up in the featured spot up top at some point during the day. Maybe this will help visualize:

1. [A, B, C, D] // line up at 8am
2. [B, C, D, A] // line up at noon
3. [C, D, A, B] // line up at 4pm
4. [D, A, B, C] // line up at 8pm

So at 8am Artwork A takes the top spot and B, C, D run along the bottom. Then at noon Artwork A moves down to make way for B, etc, etc. Perfect solution? Maybe not but I've been using this on device and I like it. Open to thoughts here and can certainly evolve as we learn!

Note this mostly applies to the large version and also the medium one. In that case we only run three so I modify this a bit. Then at small I only run one at a time.

## Artwork Fixtures

There are times when the widget must return a static view. Like when you are in the widget management interface and are choosing which type of widget you want to add to your home screen. That now looks like this:

<details>

707ED6E2-D33E-4FFC-A18A-3789A12B6B58

</details>

The artworks used to populate that view are static and checked into the repo. I have formalized this concept by creating an enum called Fixtures with four artworks:

* primaryArtwork
* secondaryArtwork
* tertiaryArtwork
* quaternaryArtwork

So these can be used anywhere in the widget code when we want something static. I intend to use this same strategy as I move into the editorial widget too.

https://artsyproduct.atlassian.net/browse/GRO-673

/cc @artsy/grow-devs @artsy/collector-experience


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  ❌ **It seems like you added some non WebP images to Eigen, please convert them to WebPs using `source images/script.sh` script **
  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against b089247f4ea2d7b59175264d8eb8b4ebddeae6e6
</p>

ArtsyWidget/Assets.xcassets/BlackArtsyLogo.imageset/Contents.json
```diff
@@ -0,0 +1,23 @@
+{
+  "images" : [
+    {
+      "filename" : "android-chrome-512x512.png",
```
pvinis, line 4: whats all this in our ios code? :p

ArtsyWidget/Models/ArtworkImage.swift
```diff
@@ -2,8 +2,10 @@ import Foundation
 
 struct ArtworkImage: Codable {
     static func fallback() -> ArtworkImage {
-        let fallbackUrl = "https://d32dm0rphc51dk.cloudfront.net/02XpcHIJyIsmdFj5u_2nOw/:version.jpg"
-        return ArtworkImage(imageUrl: fallbackUrl)
+        let artwork = Fixtures.primaryArtwork
```
pvinis, line 5: nice touch

ArtsyWidget/FeaturedArtworks/FeaturedArtworks+Timeline.swift
```diff
@@ -0,0 +1,31 @@
...
+                let entries = updateTimesToArtworks.map() { (date, artworks) in Entry(artworks: artworks, date: date) }
+                let timeline = WidgetKit.Timeline(entries: entries, policy: .after(schedule.nextUpdate))
+                completion(timeline)
+            }
+        }
+    }
+}
+
+struct Something {
+    static func rotateArray(array: [Artwork]) -> [[Artwork]] {
```
pvinis, line 19: can this not be just a function? Why do we need `Something`?

pvinis (dismissed):

pvinis:
just one question basically, then feel free to merge when you want 👍

ArtsyWidget/FeaturedArtworks/FeaturedArtworks+Timeline.swift
jonallured: Ugh good catch this was my mistake - I meant for this to coexist with the other timeline func but didn't do it. Updated!

PR #5819, opened 2021-11-23
feat: Add order information in conversation details

patrinoua:

[NX-3019]

### Description

https://user-images.githubusercontent.com/15792853/144826222-1f8bff66-b33f-4b6b-b122-186d7acd26bd.mov

The size looks intimidating, but since we had to add a bunch of fragments, most of the files and changes are from auto-generated graphql files. This PR adds offer/order information from an inquiry in the conversation details section.

### To the reviewers 👀

- Conversation details present offer/order information - araujobarret

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[NX-3019]: https://artsyproduct.atlassian.net/browse/NX-3019?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against b11d61a93f7ec3cdcd3ddb45fba8ca7b81e269e2
</p>

src/lib/Scenes/Inbox/Components/Conversations/ItemInfo.tsx
```diff
@@ -1,93 +1,30 @@
...
-    }
-  } else if (item.__typename === "Show") {
-    itemInfo = {
-      imageURL: item.image?.thumbnailUrl!,
-      line1: item?.name!,
-      line2: item?.partner?.name!,
-      line3: item.exhibitionPeriod!,
-    }
+export const ItemInfo: React.FC<ItemInfoProps> = ({ item }) => {
+  switch (item.__typename) {
```
araujobarret, line 12: This component was already doing so much, I refactored it splitting the 2 supported types

src/lib/Scenes/Inbox/Components/Conversations/Preview/Attachment/AttachmentList.tests.tsx
```diff
@@ -1,89 +1,68 @@
...
-      environment={env}
-      query={graphql`
-        query AttachmentListTestsQuery($conversationID: String!) @relay_test_operation {
-          me {
-            conversation(id: $conversationID) {
-              messagesConnection {
-                ...AttachmentList_messageConnection
-              }
-            }
+  const { renderWithRelay } = setupTestWrapperTL({
```
araujobarret, line 11: Refactored a bit to use RTL

src/lib/Scenes/Inbox/Screens/ConversationDetails.tests.tsx
```diff
@@ -1,143 +1,53 @@
...
 import "react-native"
-import { graphql, QueryRenderer } from "react-relay"
-import { act } from "react-test-renderer"
-import { createMockEnvironment } from "relay-test-utils"
+import { graphql } from "react-relay"
 import { ConversationDetailsFragmentContainer } from "./ConversationDetails"
 
 jest.unmock("react-relay")
 
 describe("ConversationDetailsFragmentContainer", () => {
```
araujobarret, line 10: Another one refactored to use RTL, most of important tests are covered by the respective fragment test

src/lib/tests/setupTestWrapper.tsx
```diff
@@ -13,6 +14,42 @@ interface SetupTestWrapper<T extends OperationType> {
   variables?: T["variables"]
 }
 
+export const setupTestWrapperTL = <T extends OperationType>({
```
araujobarret, line 17: @gkartalis mimicking Force I place this guy in order to test better the Relay Fragments

gkartalis: Thanks for that!

src/lib/tests/setupTestWrapper.tsx
```diff
@@ -13,6 +14,42 @@ interface SetupTestWrapper<T extends OperationType> {
...
+  Component,
+  query,
+  variables = {},
+}: SetupTestWrapper<T>) => {
+  const renderWithRelay = (mockResolvers: MockResolvers = {}) => {
+    const env = createMockEnvironment()
```
gkartalis, line 23: **suggestion:** Can we change that to the following? The reason for that is that [here](https://github.com/artsy/eigen/blob/08be4b60547f0ec7bf29a14abec56e3d5078eed8/src/setupJest.ts#L42) we already mock the default environment.
```suggestion
    type MockEnvironment = ReturnType<typeof createMockEnvironment>
    const mockEnvironment = defaultEnvironment as MockEnvironment

```

gkartalis (commented):
Left only a suggestion for using defaultEnvironment vs createMockEnvironment, also there are some tests failing but if you rebase master they should be fixed! 

Thanks for creating the wrapper!

src/lib/tests/setupTestWrapper.tsx
araujobarret: my only concern about using the `defaultEnvironment` is the difference with Force, there we're unmocking relay in each test file and used the approach that I pulled here

gkartalis: Sounds reasonable.

My only concern here is that in eigen before we started using the `defaultEnvironment`, when we had two consecutive requests (let's say we fetch the initial query and then we press a button and we trigger a mutation) for some reason the second request was never detected in the queue of relay requests. This resulted in `getMostRecentOperation` being empty when it should contain the mutation. 

If by this new approach this is solved then I am happy to ditch that in favour of this solution!

(I think the initial file that I encountered this behaviour was this test in [SavedAddresses.tests](https://github.com/artsy/eigen/blob/master/src/lib/Scenes/SavedAddresses/SavedAddresses.tests.tsx#L188-L237))

I can follow-up on that later so feel free to merge from my side (I've only reviewed the rntl specific code) and then I can check if `setupTestWrapperTL` works in the other cases as well!

araujobarret: I know what you mean, special when we need to test refetches we need to be able to observer the relay env.

I'm working in this question now in force, the PR is not open there it, but from the I`returned` altogether the RTL `view` and the `env` from the `renderWithRelay` fn, so the test has access to `env` and can mock the most recent operation again, if needed.

I can refactor the wrapper function to export the env from here, which would be better in my opinion because we will be forcing people to use the `env` exported, acting like single source of truth for the tests.

src/lib/Scenes/Inbox/Components/Conversations/OrderInformation.tsx
```diff
@@ -0,0 +1,100 @@
...
+export const OrderInformation: React.FC<OrderInformationProps> = ({ artwork, order }) => {
+  if (!order || !artwork) {
+    return null
+  }
+
+  return (
+    <>
+      <Flex flexDirection="column" p={2} key="support-section">
+        <Join separator={<Spacer my={0.5} />}>
+          <>
```
patrinoua, line 22: It seems like we only need <> around <Your offer> and <List Price> blocks but we can remove it from the rest of the places

src/lib/Scenes/Inbox/Components/Conversations/OrderInformation.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+            <Flex justifyContent="space-between" flexDirection="row">
+              <Text variant="xs" color="black60">
+                List price
+              </Text>
+              <Text variant="xs" color="black60">
+                {artwork.listPrice?.display}
+              </Text>
+            </Flex>
+          </>
```
patrinoua, line 44: let's keep these 👍

src/lib/Scenes/Inbox/Components/Conversations/OrderInformation.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+          <>
+            <Flex justifyContent="space-between" flexDirection="row">
+              <Text color="black60">Shipping</Text>
+              <Text color="black60">{order.shippingTotal ?? "—"}</Text>
+            </Flex>
+          </>
```
patrinoua, line 51: But remove these

src/lib/Scenes/Inbox/Components/Conversations/OrderInformation.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+          <>
+            <Flex justifyContent="space-between" flexDirection="row">
+              <Text color="black60">Tax</Text>
+              <Text color="black60">{order.taxTotal ?? "—"}</Text>
+            </Flex>
+          </>
```
patrinoua, line 58: these

src/lib/Scenes/Inbox/Components/Conversations/OrderInformation.tsx
```diff
@@ -0,0 +1,100 @@
...
+          <>
+            <Flex justifyContent="space-between" flexDirection="row">
+              <Text weight="medium">Total</Text>
+              <Text weight="medium">{order.buyerTotal ?? ""}</Text>
+            </Flex>
+          </>
+        </Join>
+      </Flex>
+      <Separator />
+    </>
```
patrinoua, line 69: and these 😊

src/lib/Scenes/Inbox/Components/Conversations/Preview/Attachment/AttachmentList.tests.tsx
patrinoua: What is RTL? 👀

patrinoua (commented):
I left some minor comments, love the refactoring and learned a lot about testing with relay!

jo-rs:
Small behaviour issue, it looks like it's currently displaying the list price for price range, even if it's marked as price range 🤔 
Screenshot 2021-12-13 at 17 59 35
Screenshot 2021-12-13 at 17 59 21

src/lib/Scenes/Inbox/Components/Conversations/PaymentMethod.tsx
```diff
@@ -0,0 +1,46 @@
...
+}
+
+export const PaymentMethod: React.FC<PaymentMethodProps> = ({ order }) => {
+  if (!order.creditCard) {
+    return null
+  }
+
+  const { lastDigits, expirationMonth, expirationYear } = order.creditCard
+  const monthString = expirationMonth.toString()
+  const month = monthString.length === 1 ? `0${expirationMonth}` : expirationMonth
```
jo-rs, line 18: **nitpick**
Cleaner alternative:
`('0' + expirationMonth).slice(-2)` 🙃

jo-rs: Or, I guess we're using `padStart` generally 🤔

src/lib/Scenes/Inbox/Components/Conversations/OrderInformation.tsx
jo-rs: Agreed ☝️ some of these fragments seem unnecessary.

src/lib/Scenes/Inbox/Components/Conversations/SellerReplyEstimate.tsx
```diff
@@ -0,0 +1,85 @@
...
+      }
+      break
+    case "APPROVED":
+      if (order.requestedFulfillment?.__typename !== "CommercePickup") {
+        message = `Thank you for your purchase. You will be notified when the work has shipped, typically within 5–7 business days.`
+      }
+      break
+    case "PROCESSING":
+      message = getDeliveryEstimateMessage(order?.lineItems?.edges?.[0]?.node?.selectedShippingQuote?.displayName ?? "")
+      break
```
jo-rs, line 46: We should probably move some of this logic to `metaphysics` since we have a similar implementation on `force`. I'll put up a ticket.

src/lib/Scenes/Inbox/Components/Conversations/Preview/Attachment/AttachmentList.tests.tsx
jo-rs: `React Testing Library` 😊

src/lib/Scenes/Inbox/Components/Conversations/Support.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export const Support = () => {
+  return (
+    <Flex flexDirection="column" p={2} key="support-section">
+      <Text variant="md" mb={1} weight="medium">
+        Support
+      </Text>
+      <Touchable
+        onPress={() => {
+          navigate("https://support.artsy.net/hc/en-us/sections/360008203054-Contact-a-gallery", { modal: true })
```
jo-rs, line 14: This doesn't seem to be working locally as a modal, and I'm not entirely sure if we generally use complete URLs vs. setting them as `webViewRoute`s 🤔 
Maybe worth checking with #practice-mobile to see what the accepted standard is,

jo-rs (commented):
It's a big boy, so I hopefully didn't miss anything 😕 
Looks excellent, @araujobarret and @patrinoua ! 💯 
Just a couple of nitpicks/questions and one minor issue to resolve.

src/lib/Scenes/Inbox/Components/Conversations/SellerReplyEstimate.tsx
araujobarret: Thanks and I super agree!

src/lib/Scenes/Inbox/Components/Conversations/OrderInformation.tsx
araujobarret: good catch this fragment is not necessary!

src/lib/Scenes/Inbox/Components/Conversations/OrderInformation.tsx
araujobarret: Actually, they're necessary, I could have explained before sorry 😄
The way `Join` component works with the `spacer` prop, it considers each 1st level children and put the specific `spacer` between them, so in this case, I forced to put the `spacer` component between the fragments.
in the example below I will give a better idea:

```javascript
  <>
    <Text>Some Text</Text>
     <Text>Some Text</Text>
  </>
  
  <Text>Some other Text</Text>
  
  <>
    <Text>Another Text</Text>
    <Text>Another Text</Text>
  </>
</Join>
```
As the rendered result we will have:
```
React.Fragment
  Text
  Text
Spacer
Text
Spacer
React.Fragment
  Text
  Text
```

There is also a way to ignore the nesting level of the children inside `Join` with the prop `flatten`, more example and information about the `Join` buddy [here](https://github.com/artsy/eigen/blob/main/src/palette/elements/Join/Join.tsx#L14)

src/lib/Scenes/Inbox/Components/Conversations/OrderInformation.tsx
araujobarret: removed on [d46673] (https://github.com/artsy/eigen/pull/5819/commits/d46673e49437d463454103f4e9f3e4b6878ce552) 👍

src/lib/Scenes/Inbox/Components/Conversations/OrderInformation.tsx
araujobarret: removed on [d46673](https://github.com/artsy/eigen/pull/5819/commits/d46673e49437d463454103f4e9f3e4b6878ce552) 👍

src/lib/Scenes/Inbox/Components/Conversations/OrderInformation.tsx
araujobarret: removed on [d46673](https://github.com/artsy/eigen/pull/5819/commits/d46673e49437d463454103f4e9f3e4b6878ce552) 👍

jo-rs (approved):

PR #5833, opened 2021-11-26
feat: Track screen reader status and preferred theme of users

gkartalis:

This PR resolves []

### Description

- introduces tracking for users that have screen readers enabled on their devices.
- replaces tracking theme preferences for iOS only to both iOS and android
- removes native code for fresh install and use custom hook for both platforms

|Fresh Install| traits screen reader enabled / preferred theme|
|---|---|
|Screenshot 2021-12-10 at 14 57 30|Screenshot 2021-12-10 at 15 05 03|

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- track screen reader status and theme status - gkartalis

</details>


ArtsyOpenSource:

- Dev changes (track screen reader status and theme status - gkartalis)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 98f0533fb84e9db92765485552024a5b231f3536
</p>

src/lib/utils/track/constants.ts
```diff
@@ -1,8 +1,3 @@
 export const AnalyticsConstants = {
   FreshInstall: "first user install",
```
gkartalis, line 2: notice that [here](https://github.com/artsy/eigen/pull/5833/files#diff-dd9db7bec93b7bd1368805ddd7e76e84917db6b0393de2b1661425ec25c40cefL72) we are using the FreshInstall event ONLY in android. Is there a reason for that? Do we want to also track iOS or I am missing some context?

pvinis: if i remember correctly, we had a different one for ios.

yep, here it is https://github.com/artsy/eigen/blob/b78d277f1b6d545d4703eac66bd77b1ddba003c2/Artsy/App/ARAnalyticsConstants.m#L12.

we could totally remove the native one on ios and use the ts one on both platforms.

src/lib/utils/track/constants.ts
```diff
@@ -1,8 +1,3 @@
...
-
-  UserInterfaceStyle: {
```
pvinis, line 4: 👋

src/lib/utils/usePreferredThemeTracking.tests.ts
```diff
@@ -0,0 +1,19 @@
+import { renderHook } from "@testing-library/react-hooks"
+import { postEventToProviders } from "./track/providers"
+import { usePreferredThemeTracking } from "./usePreferredThemeTracking"
+
+describe("usePreferredThemeTracking", () => {
+  it("Should track the preferred theme of the user", () => {
+    renderHook(() => usePreferredThemeTracking())
+
+    expect(postEventToProviders).toHaveBeenCalledTimes(1)
+    expect((postEventToProviders as jest.Mock).mock.calls[0]).toMatchInlineSnapshot(`
```
pvinis, line 10: 🙌

src/lib/utils/usePreferredThemeTracking.ts
```diff
@@ -0,0 +1,18 @@
+import { useEffect } from "react"
+import { ColorSchemeName, useColorScheme } from "react-native"
+import { postEventToProviders } from "./track/providers"
+
+// the purpose of this hook is to track the preferred theme of a user
+export const usePreferredThemeTracking = () => {
+  const colorScheme = useColorScheme()
+  useEffect(() => {
+    postEventToProviders(tracks.getUserPreferredTheme(colorScheme))
```
pvinis, line 9: we talked about this on our meeting, i still think having `SegmentTrackingProvider.identify?.(` makes more sense than just `postEventToProviders`, because that way this context of `user_preferred_theme` stays with the user.

see here with the old name:
Screenshot 2021-12-02 at 15 43 52

src/lib/utils/useScreenReaderTracking.ts
```diff
@@ -0,0 +1,24 @@
...
+export const useScreenReaderTracking = () => {
+  const [isScreenReaderEnabled, setIsScreenReaderEnabled] = useState(false)
+  useEffect(() => {
+    const getScreenReaderStatus = async () => {
+      const isEnabled = await AccessibilityInfo.isScreenReaderEnabled()
+      setIsScreenReaderEnabled(isEnabled)
+    }
+
+    getScreenReaderStatus()
+    postEventToProviders(tracks.trackScreenReaderStatus(isScreenReaderEnabled))
```
pvinis, line 15: same here

pvinis (changes requested):
I commented about the `identify` call. i think it makes more sense to go that way. waht do you think?

src/lib/utils/usePreferredThemeTracking.ts
MounirDhahri: I think we don't necessarily need it since using the anonymous id we will be able to trace it back to the user, unless we need the user id for something

src/lib/utils/usePreferredThemeTracking.ts
```diff
@@ -0,0 +1,18 @@
...
+  }, [])
+}
+
+export const tracks = {
+  getUserPreferredTheme: (preferredTheme: ColorSchemeName) => ({
+    name: "user_preferred_theme",
+    user_interface_style: preferredTheme,
```
MounirDhahri, line 16: **issue:** would you mind adding an interface for the event in cohesion

MounirDhahri: **though:** I also believe the platform can be useful here

src/lib/utils/useScreenReaderTracking.ts
```diff
@@ -0,0 +1,24 @@
...
+import { AccessibilityInfo } from "react-native"
+import { postEventToProviders } from "./track/providers"
+
+// the purpose of this hook is to track the screen reader status of the user's device
+export const useScreenReaderTracking = () => {
+  const [isScreenReaderEnabled, setIsScreenReaderEnabled] = useState(false)
+  useEffect(() => {
+    const getScreenReaderStatus = async () => {
+      const isEnabled = await AccessibilityInfo.isScreenReaderEnabled()
+      setIsScreenReaderEnabled(isEnabled)
```
MounirDhahri, line 11: **thought:** I am not sure why we need to use the state here. Can't we maybe do
```suggestion
      postEventToProviders(tracks.trackScreenReaderStatus(isEnabled))
```
and not use a state

MounirDhahri (dismissed):
Nice to see PRs coming from the practice 🕺 I left few minor comments and we should be clear to go.

src/lib/utils/useScreenReaderTracking.ts
gkartalis: Good point. The initial approach was without `useState`. The reason I used state here is because if I don't (and my code being like this):
```suggestion
  export const useScreenReaderTracking = () => {
    useEffect(() => {
      const getScreenReaderStatus = async () => {
        const isEnabled = await AccessibilityInfo.isScreenReaderEnabled()
        return isEnabled
      }

      getScreenReaderStatus()
      postEventToProviders(tracks.trackScreenReaderStatus(getScreenReaderStatus()))
    }, [])
  }
```
I get the following ts error `Argument of type 'Promise<boolean>' is not assignable to parameter of type 'boolean'.ts(2345)`

MounirDhahri: ah yeah. it's ts sense/non-sense  but I guess it's understandable in case the promise fails

src/lib/utils/usePreferredThemeTracking.ts
gkartalis: > **issue:** would you mind adding an interface for the event in cohesion

Will do that :) 
> **though:** I also believe the platform can be useful here

I can add that as well although the event in segment gives us the following data that also includes device info like id. manufacturer, model, name and type so not sure if we really need to add the platform �
Screenshot 2021-12-03 at 15 06 49
�

src/lib/utils/usePreferredThemeTracking.ts
gkartalis: Hmmm, we will be able to trace it back to the user **for sure** since we have all the data.
The thing that @pvinis says though is that if we track it with `SegmentTrackingProvider.identify?.` we will have it under `traits` property in every event of the app it will "follow" the user under `traits`. 
Not sure which of the two we prefer.

I am thinking that it might be better for consistency with the previous collected data to continue having the `user inteface style` under traits like we had it before and maybe do the same with the `screen_reader_enabled`?

What do you think?

pvinis:
I think its best if we add a custom thing in cohesion. Something that allows us to send like
```
{
  action: "custom-track",
  key: "screen-reader",
  value: "whatever the value is",
}
```
then we (as devs or teams) we can track things that are not needed in cohesion, and we can freely and easily add or remove these when we want, without needed extra changes in cohesion or bothering the data team.

src/lib/utils/useFreshInstallTracking.ts
```diff
@@ -0,0 +1,14 @@
...
+import { AnalyticsConstants } from "./track/constants"
+import { SegmentTrackingProvider } from "./track/SegmentTrackingProvider"
+
+export const useFreshInstallTracking = () => {
+  useEffect(() => {
+    const launchCount = getCurrentEmissionState().launchCount
+    if (launchCount > 1) {
+      return
+    }
+    SegmentTrackingProvider.postEvent({ name: AnalyticsConstants.FreshInstall })
```
pvinis, line 12: better use the `postEventToProviders` here, so it makes suse we send to anything we might use.

src/lib/utils/usePreferredThemeTracking.ts
MounirDhahri: I don't oppose this if we were using previous data, but I am not sure if we were using it before. Otherwise, yes that's a good point

src/lib/utils/usePreferredThemeTracking.ts
MounirDhahri: aha we have the `os` there already. I didn't know that, in other areas of the app we are tracking the platform, good to know it's avoidable. we usually have it as a separate value for the sake of simplicity, because who knows what the "os" name will be in a browser, will it be "web", "firefox", "ubuntu" etc... but it's not important here since this is not a browser 😄

src/lib/utils/useScreenReaderTracking.ts
pvinis: I agree with getting rid of the state.

maybe try to move the thing in a try-catch, so if you get a boolean you postevent, if you get an error you send nothing or send a "N/A" or something.

pvinis (approved):
I think we could get rid of the state, but I'll approve so I don't block this. we can change it later too.

PR #5835, opened 2021-11-29
feat: Move home screen separators into modules

olerichter00:

This PR resolves [CX-2012]

### Description

Moves home screen module separators into the modules/rails to avoid rendering two separators in the case a module doesn't show anything.

| Personalized | Not Personalized |
| --- | --- |
| Simulator Screen Shot - iPhone 8 - 2021-11-29 at 15 14 36 | Simulator Screen Shot - iPhone 8 - 2021-11-29 at 15 16 26 |

### To the reviewers 👀

- Fix "too much space between modules" issue on home screen - ole 

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-2012]: https://artsyproduct.atlassian.net/browse/CX-2012?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/Components/Home/ArtistRails/ArtistRail.tsx
+++ b/src/lib/Components/Home/ArtistRails/ArtistRail.tsx
@@ -33,6 +33,7 @@ interface Props extends ViewProps {
   subtitle?: string
   relay: RelayProp
   rail: ArtistRail_rail
+  mb?: number
 }
 
 const ArtistRail: React.FC<Props & RailScrollProps> = (props) => {
@@ -178,7 +179,7 @@ const ArtistRail: React.FC<Props & RailScrollProps> = (props) => {
   }
 
   return artists.length ? (
-    <View>
+    <Flex mb={props.mb}>
       <Flex pl="2" pr="2">
         <SectionTitle title={props.title} subtitle={props.subtitle} />
       </Flex>
@@ -215,7 +216,7 @@ const ArtistRail: React.FC<Props & RailScrollProps> = (props) => {
           )
         }}
       />
-    </View>
+    </Flex>
   ) : null
 }
 
--- a/src/lib/Components/LotsByArtistsYouFollowRail/LotsByFollowedArtistsRail.tests.tsx
+++ b/src/lib/Components/LotsByArtistsYouFollowRail/LotsByFollowedArtistsRail.tests.tsx
@@ -11,9 +11,6 @@ import { LotsByFollowedArtistsRailContainer, PAGE_SIZE } from "./LotsByFollowedA
 
 jest.unmock("react-relay")
 
-const onShowMock = jest.fn()
-const onHideMock = jest.fn()
-
 describe("LotsByFollowedArtistsRail", () => {
   let mockEnvironment: ReturnType<typeof createMockEnvironment>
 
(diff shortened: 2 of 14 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 0ccfec6d0069fbb18033e03e85c718706319a8d8
</p>

src/lib/Components/LotsByArtistsYouFollowRail/LotsByFollowedArtistsRail.tests.tsx
```diff
@@ -30,14 +27,7 @@ describe("LotsByFollowedArtistsRail", () => {
       variables={{}}
       render={({ props }) => {
         if (props?.me) {
-          return (
-            <LotsByFollowedArtistsRailContainer
-              title="Auctions"
-              me={props.me}
-              onShow={onShowMock}
-              onHide={onHideMock}
```
olerichter00, line 38: No need to use `onShow` and `onHide` anymore.

src/lib/Components/Home/ArtistRails/ArtistRail.tsx
```diff
@@ -33,6 +33,7 @@ interface Props extends ViewProps {
   subtitle?: string
   relay: RelayProp
   rail: ArtistRail_rail
+  mb?: number
```
olerichter00, line 36: Adding `mb` as an optional prop will prevent the rail to render the bottom margin in other places of the app. I'm not sure if we should give the prop another name because it just adds the margin if the rail renders any data. I went with `mb` because we use the same prop in **Palette**.

src/lib/Scenes/Home/Home.tsx
```diff
@@ -42,6 +42,8 @@ import { ShowsRailFragmentContainer } from "./Components/ShowsRail"
 import { TroveFragmentContainer } from "./Components/Trove"
 import { RailScrollRef } from "./Components/types"
 
+const MODULE_SEPARATOR_HEIGHT = 6
```
rajsam003, line 45: I think it should be 4 instead of 6

image

rajsam003: You can refer https://www.figma.com/file/ye05MSfEEZ265CIAUJ8erN/Home-Iterations?node-id=1637%3A8403

olerichter00: Thanks for looking into that. The space increased with Palette V3 from `4` to `6` (see Home V3 in Figma file).

src/lib/Components/Home/ArtistRails/ArtistRail.tsx
rajsam003: That's good. But I don't see any point in passing it as a prop. Coz all those containers are used only in the `Home.tsx` file. So my advice would be to add `mb=6` (or 4) to the wrapper of each container.

rajsam003:
Great work. Just added a few comments from my side.

src/lib/Components/Home/ArtistRails/ArtistRail.tsx
olerichter00: The problem is that we can't figure out if a module renders a rail or nothing in `Home`. If we would use a wrapper this wrapper would render a bottom margin even if the rail doesn't render and we would have the same problem: too much space between modules.

olerichter00: This is a good point. I've decided to add it as a prop in case we want to use any of these rails somewhere else. For me it makes more sense to pass the margin from `Home` instead of adding it directly to the rail/module.

PR #5851, opened 2021-12-02
Chore: Refactor Add My Collection Artwork form from Modal to Individual Pages

lordkiz:

This PR resolves [CX-2165]

### Description

- Refactored MyCollection Artwork form flow from using Modal to Using a normal Nav Screen

Other refactor work which included:
- Hack to pass functions as navigational props on iOS
- Removed unnecessary refs in Medium Picker
- Fix MaxListener exceeded warning when using so many Inputs in a long form

https://user-images.githubusercontent.com/18648835/144646574-ca7d9af4-9e7d-49ca-a11b-75aaf9c1e07a.mp4

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- Chore: Refactor My Collection Artwork Form flow - kizito

</details>

[CX-2165]: https://artsyproduct.atlassian.net/browse/CX-2165?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 238ed8fb82b2b815682651700c7cbe50b3faa423
</p>

src/lib/Scenes/MyCollection/Screens/ArtworkFormModal/MyCollectionArtworkFormModal.tsx
```diff
@@ -38,14 +38,19 @@ export type ArtworkFormMode = "add" | "edit"
 export type ArtworkFormModalScreen = {
```
brainbicycle, line 38: **thought:** maybe we should rename if no longer a modal?

src/lib/Scenes/MyCollection/Screens/ArtworkFormModal/MyCollectionArtworkFormModal.tsx
```diff
@@ -38,14 +38,19 @@ export type ArtworkFormMode = "add" | "edit"
...
+  }
+  AddPhotos: {
+    onHeaderBackButtonPress(): void
   }
-  AdditionalDetails: undefined
-  AddPhotos: undefined
 }
 
-type MyCollectionArtworkFormModalProps = { visible: boolean; onDismiss: () => void; onSuccess: () => void } & (
+type MyCollectionArtworkFormModalProps = { onSuccess: () => void } & (
```
brainbicycle, line 53: same comment as above

src/lib/Scenes/MyCollection/Screens/ArtworkFormModal/MyCollectionArtworkFormModal.tsx
```diff
@@ -184,32 +195,47 @@ export const MyCollectionArtworkFormModal: React.FC<MyCollectionArtworkFormModal
...
+            name="AdditionalDetails"
+            component={MyCollectionAdditionalDetailsForm}
+            initialParams={{ onHeaderBackButtonPress }}
+          />
+          <Stack.Screen
+            name="AddPhotos"
+            component={MyCollectionAddPhotos}
+            initialParams={{ onHeaderBackButtonPress }}
+          />
+        </Stack.Navigator>
```
brainbicycle, line 237: ✨

src/lib/Scenes/MyCollection/Screens/ArtworkFormModal/Screens/MyCollectionArtworkFormMain.tsx
```diff
@@ -43,7 +43,11 @@ export const MyCollectionArtworkFormMain: React.FC<StackScreenProps<ArtworkFormM
 
   return (
     <>
-      <FancyModalHeader leftButtonText="Cancel" onLeftButtonPress={() => route.params.onDismiss()}>
+      <FancyModalHeader
```
brainbicycle, line 46: similar, maybe if we are using this for more than just modals we should name it something like `NavigationHeader`?

src/lib/store/PropsStore.tests.ts
```diff
@@ -0,0 +1,59 @@
+import { propsStore } from "./PropsStore"
+
+describe("propsStore", () => {
+  describe(propsStore.setPendingProps, () => {
+    it("sets pendingProps", () => {
```
brainbicycle, line 5: nitpick: would be nice to be a little more descriptive for people who come to this module and aren't as familiar
```suggestion
    it("stores and retrieves props for a module", () => {
```

src/lib/store/PropsStore.tests.ts
```diff
@@ -0,0 +1,59 @@
...
+      const pendingProps = {
+        str: "a string",
+        afunc: () => null,
+      }
+      propsStore.setPendingProps("myModule", pendingProps)
+      expect(propsStore.getPropsForModule("myModule")).toEqual(pendingProps)
+    })
+    it("overwrites whatever pending props", () => {
```
brainbicycle, line 13:
```suggestion
    it("overwrites whatever props already exist for a module", () => {
```

src/lib/store/PropsStore.tests.ts
```diff
@@ -0,0 +1,59 @@
...
+      const expected = { str: first.str, deezFunc: second.deezFunc }
+      const moduleName = "Modularrr"
+      propsStore.setPendingProps(moduleName, first)
+      propsStore.mergeNewPropsForModule(moduleName, second)
+      expect(propsStore.getPropsForModule(moduleName)).toEqual(expected)
+    })
+  })
+
+  describe(propsStore.updateProps, () => {
+    it("callback with updated props", () => {
```
brainbicycle, line 47:
```suggestion
    it("calls callback with updated props", () => {
```

src/lib/store/PropsStore.tests.ts
```diff
@@ -0,0 +1,59 @@
...
+      const callback = jest.fn()
+      const first = { str: "a string" }
+      const second = { str: "a diff string", deezFunc: () => null }
+      const expected = { str: second.str, deezFunc: second.deezFunc }
+      const moduleName = "banananana"
```
brainbicycle, line 52: 🍌

brainbicycle (changes requested):
🙌 
Overall this looks great to me, some minor requests on naming specifically if things are no longer being used only for modals or aren't modals can we remove that from the name, and then some nitpicks around test descriptions but otherwise looks great! Thanks

src/lib/Scenes/MyCollection/Screens/ArtworkFormModal/MyCollectionArtworkFormModal.tsx
lordkiz: @brainbicycle I thought to but this will make the PR large and difficult to review. I will do it anyway.

src/lib/Scenes/MyCollection/Screens/Artwork/MyCollectionArtwork.tsx
```diff
@@ -42,22 +40,24 @@ export const MyCollectionArtwork: React.FC<MyCollectionArtworkProps> = ({ artwor
...
+              passProps: {
+                mode: "edit",
+                artwork,
+                onSuccess: () => {
+                  popToRoot()
+                },
+                onDelete: () => {
+                  setTimeout(() => {
+                    popToRoot()
+                  }, 50)
```
olerichter00, line 57: **(question):** Why do we need a timeout for `onDelete`?

src/lib/Scenes/MyCollection/Screens/ArtworkForm/MyCollectionArtworkForm.tsx
```diff
@@ -145,6 +152,10 @@ export const MyCollectionArtworkFormModal: React.FC<MyCollectionArtworkFormModal
             await myCollectionDeleteArtwork(props.artwork.internalID)
             refreshMyCollection()
             props.onDelete()
+            setTimeout(() => {
+              // reset form values incase user deleted after editing
+              GlobalStore.actions.myCollection.artwork.resetForm()
+            }, 500)
```
olerichter00, line 158: **(question)**: Is there any way to remove the form data inside an effect instead? I don't think using a timer with an arbitrary timeout is a good idea.

src/lib/Scenes/MyCollection/Screens/ArtworkFullDetails/MyCollectionArtworkFullDetails.tsx
```diff
@@ -2,55 +2,46 @@ import { editCollectedArtwork } from "@artsy/cohesion"
...
+            passProps: {
+              mode: "edit",
+              artwork: props.artwork,
+              onSuccess: () => {
+                popParentViewController()
+              },
+              onDelete: () => {
+                setTimeout(() => {
+                  popToRoot()
+                }, 50)
```
olerichter00, line 35: **(question):** Why do we need a timeout here?

src/lib/store/PropsStore.tests.ts
olerichter00: 🍌 🍌 🍌

rajsam003 (approved):
LGTM 👍🏽

src/palette/elements/Input/Input.tsx
```diff
@@ -111,7 +111,11 @@ export const Input = React.forwardRef<TextInput, InputProps>(
     const fontFamily = theme.fonts.sans.regular
 
     useEffect(() => {
-      inputEvents.addListener("clear", localClear)
+      if (!inputEvents.listeners("clear").length) {
+        // in a long form with so many Inputs, we will easily exceed maxLimit for
+        // listeners for clear. So add it only if not already added.
+        inputEvents.addListener("clear", localClear)
+      }
```
olerichter00, line 118: I don`t really understand how we exceed the `maxLimit` for listeners in the form. When I look at the form, I see less than 20 inputs. Are you sure that the reason for the warning is not due to an error?

olerichter00 (commented):
This is great work! 🙌  I've added a couple of questions and ideas.

src/lib/Scenes/MyCollection/MyCollection.tsx
```diff
@@ -147,8 +142,13 @@ const MyCollection: React.FC<{
               <Button
                 testID="add-artwork-button-zero-state"
                 onPress={() => {
-                  setShowModal(true)
                   trackEvent(tracks.addCollectedArtwork())
+                  navigate("my-collection/add-my-collection-artwork", {
```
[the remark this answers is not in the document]
pvinis: I like this idea

src/lib/Scenes/MyCollection/Screens/ArtworkForm/Screens/MyCollectionArtworkFormMain.tsx
```diff
@@ -43,7 +43,11 @@ export const MyCollectionArtworkFormMain: React.FC<StackScreenProps<ArtworkFormM
 
   return (
     <>
-      <FancyModalHeader leftButtonText="Cancel" onLeftButtonPress={() => route.params.onDismiss()}>
+      <FancyModalHeader
+        onLeftButtonPress={route.params.onHeaderBackButtonPress}
+        rightButtonText={isFormDirty() ? "Clear" : undefined}
+        onRightButtonPress={isFormDirty() || __TEST__ ? () => route.params.clearForm() : undefined}
```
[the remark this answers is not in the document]
pvinis: this has caused problems in the sticky header too. I agree we should avoid it.

src/lib/AppRegistry.tsx
```diff
@@ -194,9 +197,13 @@ const InnerPageWrapper: React.FC<PageWrapperProps> = ({ fullBleed, isMainView, V
 @track()
 class PageWrapper extends React.Component<PageWrapperProps> {
   render() {
+    const props = {
+      ...this.props,
+      viewProps: { ...this.props.viewProps, ...propsStore.getPropsForModule(this.props.moduleName) },
```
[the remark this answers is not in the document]
lordkiz: @olerichter00 I see. However, PropsStore only hold props that'd be passed around in react native. If I name a react native props viewProps, that would imply that these are props meant for only react native Views. Here each module is registered and treated as native views, so it makes sense to name whatever set of props specific to that native view viewProps. If you for example write `propsStore.getViewPropsForModule` outside of `AppRegistry.tsx`, it becomes really really confusing.

src/lib/Scenes/MyCollection/MyCollection.tsx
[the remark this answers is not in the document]
lordkiz: @olerichter00 Nice idea. Can we have multiple routes match the same module?

src/lib/Scenes/MyCollection/Screens/Artwork/MyCollectionArtwork.tsx
lordkiz: @olerichter00 I don't want to popToRoot while root (in this case MyCollection.tsx) is still being refreshed. You don't think it is important?

brainbicycle (dismissed):

src/lib/Scenes/MyCollection/MyCollection.tsx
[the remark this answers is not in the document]
olerichter00: That should work but why do we need multiple routes for the same module?

src/lib/Scenes/MyCollection/Screens/Artwork/MyCollectionArtwork.tsx
olerichter00: What happens if we remove the timeout?

src/lib/Scenes/MyCollection/Screens/ArtworkForm/MyCollectionArtworkForm.tsx
lordkiz: @olerichter00  Inside an effect? I don't understand. 
Also, reseting the form will trigger a reload if we have not popped this screen yet. Also we want to avoid reseting the form as we are popping the view. Resetting before popping the view will just cause an unnecessary rerender. Why is using a timeout bad in this case?

src/lib/Scenes/MyCollection/Screens/ArtworkForm/MyCollectionArtworkForm.tsx
```diff
@@ -184,38 +195,53 @@ export const MyCollectionArtworkFormModal: React.FC<MyCollectionArtworkFormModal
     }
 
     GlobalStore.actions.myCollection.artwork.resetForm()
-    props.onDismiss?.()
+  }
+
+  const onHeaderBackButtonPress = () => {
+    const currentRoute = navContainerRef.current?.getCurrentRoute()
+    if (!currentRoute?.name || currentRoute?.name === "ArtworkForm") {
```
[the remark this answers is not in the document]
lordkiz: @olerichter00 Nein. `MyCollectionArtworkForm` just holds the Navigator, the routes are from the stack screens. ArtworkForm is the first screen of this Navigator. Usually before any navigation, currentRoute will be nil, that's why I did `!currentRoute?.name ...`

src/lib/Scenes/MyCollection/MyCollection.tsx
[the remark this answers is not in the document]
lordkiz: @olerichter00 for `my-collection/artworks/edit`

src/lib/Scenes/Partner/Components/PartnerShows.tests.tsx
```diff
@@ -1,47 +1,59 @@
-import { PartnerShows_partner } from "__generated__/PartnerShows_partner.graphql"
-import { PartnerShowsTestsQueryRawResponse } from "__generated__/PartnerShowsTestsQuery.graphql"
-import { GlobalStoreProvider } from "lib/store/GlobalStore"
-import { renderRelayTree } from "lib/tests/renderRelayTree"
+import { PartnerShowsTestsQuery, PartnerShowsTestsQueryRawResponse } from "__generated__/PartnerShowsTestsQuery.graphql"
```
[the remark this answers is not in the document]
lordkiz: Haha. These are from George's test fix. I think my Artsy vscode settings reformatted them or something.

src/palette/elements/Input/Input.tsx
lordkiz: @olerichter00 EventEmitter will continue to add listeners for a particular event until it gets to 10, then it will start showing warnings even though it will continue to add these events listeners. In forms like add artwork form more than 10 'clear' event listeners will be added so we'd get warnings all over the place. We don't ever need more than 1 clear event at any time.

src/lib/Scenes/MyCollection/Screens/ArtworkForm/Screens/MyCollectionArtworkFormMain.tsx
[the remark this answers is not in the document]
lordkiz: @olerichter00 @pvinis thanks. I agree, it doesn't look nice. How do you test clearForm() in this case if it is dependent on isFormDirty() ?

src/lib/Scenes/MyCollection/Screens/Artwork/MyCollectionArtwork.tsx
lordkiz: @olerichter00 occassionally you might see the artwork that would be deleted when you arrive at the root, followed by a jumpy MyCollection page as it is refreshed. It doesn't look bad though, it animates gracefully. I guess I can remove these timeouts. But I'd rather not see the refresh happening in my opinion.

src/lib/Scenes/MyCollection/MyCollection.tsx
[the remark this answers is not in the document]
olerichter00: This would then probably be `my-collection/artworks/:artworkID/edit` and use the same form, yes 👍

src/lib/Scenes/MyCollection/Screens/ArtworkForm/MyCollectionArtworkForm.tsx
olerichter00: I was thinking about using the [useEffect cleanup function](https://blog.logrocket.com/understanding-react-useeffect-cleanup-function/).

src/lib/Scenes/MyCollection/Screens/ArtworkForm/Screens/MyCollectionArtworkFormMain.tsx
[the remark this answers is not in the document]
olerichter00: Can you make the form dirty to test it?

src/palette/elements/Input/Input.tsx
olerichter00: At which point do we add these events? Isn't this the problem we want to solve instead?

src/lib/Scenes/MyCollection/Screens/ArtworkForm/MyCollectionArtworkForm.tsx
lordkiz: Cleanup yeah. Why didn't I think of that? 👍

src/lib/Scenes/MyCollection/MyCollection.tsx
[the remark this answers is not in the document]
lordkiz: So we'd be passing an id that we'll never use? 🤔

src/palette/elements/Input/Input.tsx
lordkiz: This ☝️ is the point we add the event listener.

olerichter00: Hmm, but looking at the code the event listener is added once the Input is mounted and then it is removed again once the input is unmounted. How can the clear event be added multiple times for one `Input` component?

```
useEffect(() => {
      inputEvents.addListener("clear", localClear)
      return () => {
        inputEvents.removeListener("clear", localClear)
      }
    }, [])
```

src/lib/Scenes/MyCollection/Screens/ArtworkForm/MyCollectionArtworkForm.tsx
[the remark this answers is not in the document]
olerichter00: Ok, can you add a comment to explain this in the code (for people like me who don't understand it ;))?

src/lib/Scenes/MyCollection/MyCollection.tsx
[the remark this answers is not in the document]
olerichter00: If we edit an artwork we would need the ID of the artwork to know which to edit, right?

src/lib/Scenes/MyCollection/Screens/Artwork/MyCollectionArtwork.tsx
olerichter00: I would remove it. Adding these timeouts can always lead to unpredictable behaviour.

src/palette/elements/Input/Input.tsx
lordkiz: Each Input will add a `clear` event listener when it mounts. That's why in a long form with many Inputs you can end up adding so many. Listeners are not local to the component that added them. That's why you remove them when you unmount. Listeners can continue to exist even after the component that added them is destroyed.

olerichter00: > That's why you remove them when you unmount. Listeners can continue to exist even after the component that added them is destroyed.

Doesn't this mean that the listener will be removed when the component is unmounted? Maybe you can create an example that reproduces the issue and we can see how to fix it. Adding `if (!inputEvents.listeners("clear").length)` won't work because it brings back the bug that has been fixed in https://github.com/artsy/eigen/pull/5681 .

src/lib/Scenes/MyCollection/Screens/ArtworkForm/MyCollectionArtworkForm.tsx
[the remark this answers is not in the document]
lordkiz: Oh, this is obvious. You are just confusing the Navigator for screens. This page just holds the Navigator that holds the screens. React-navigation by default will use the first screen you defined in the Navigator as the initial screen which in this case is ArtworkForm. Before we navigate within the Navigator, currentRoute can be null (remember our navigation to this Navigator on ios is from Native pushView, while on android this Navigator is nested inside NavStack.tsx)

src/lib/Scenes/MyCollection/MyCollection.tsx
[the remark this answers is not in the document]
lordkiz: Not necessarily. We already have the artwork, and can just pass it to the component. No need to refetch it. But there is no harm in adhering to the standards of Restful routes even though we pass the artwork for the sake of convenience. It is much neater.

src/palette/elements/Input/Input.tsx
lordkiz: @olerichter00 It will not bring back that bug. The Input in Search Page will register a new event listener for clear if the event listener is not already registered. So no issues here.

lordkiz: > Doesn't this mean that the listener will be removed when the component is unmounted?
Yes, but a 100 Inputs will add 100 listeners for the same event

lordkiz: @olerichter00 Fixed the bug where unmounted components prevent clear from firing. This is apparently because the event listener was using an anonymous function as a callback. Changing it to a named function should allow any instance of Input to be able to receive clear event and clear without necessarily registering a new event listener.

olerichter00: I still get the same bug with your current changes. I created a [code sandbox](https://codesandbox.io/s/js-playground-forked-u2nk4?file=/src/index.js) with a simplified example to explain why we need to add an event listener for every input.

src/palette/elements/Input/Input.tsx
```diff
@@ -90,6 +91,7 @@ export const Input = React.forwardRef<TextInput, InputProps>(
       multiline,
       maxLength,
       showLimit,
+      shouldAddListenerForClear = true,
```
olerichter00, line 94: Why not default to `false` since we just need the clear listener in one place in the app (Search Screen)?

```suggestion
      addClearListener = false,
```

lordkiz: I didn't know we needed it only in one place. It definitely should default to false. 👍

olerichter00 (dismissed):
🚀

olerichter00 (approved):

olerichter00:
#squashongreen

PR #5904, opened 2021-12-15
fix(DX): some small but nice things

pvinis:

This PR resolves []

### Description

This started as an investigation to solve `RangeError: Maximum call stack exceeded` with @olerichter00 and @MounirDhahri. We failed, but at least I fixed the following on the way.

- moved `setupSentry` in `index.ios.js` so it doesnt get called again
- removed a deprecated `enableOutOfMemoryTracking` from sentry init
- added an `Open RN Dev Menu` menu item in the admin menu, so we can open the RN dev menu on a device!
- reworked the `useScreenReaderTracking` to not use state

https://user-images.githubusercontent.com/100233/146233912-19aa5f4d-4380-42f3-9348-a6ba8642a1e1.mov

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- moved `setupSentry` in `index.ios.js` so it doesnt get called again - pavlos
- removed a deprecated `enableOutOfMemoryTracking` from sentry init - pavlos
- added an `Open RN Dev Menu` menu item in the admin menu, so we can open the RN dev menu on a device! - pavlos
- reworked the `useScreenReaderTracking` to not use state - pavlos

</details>

### DIFF
--- a/index.ios.js
+++ b/index.ios.js
@@ -17,8 +17,6 @@ if (startStorybook) {
 } else {
   require("react-native-gesture-handler")
   require("react-native-screens").enableScreens()
-  require("./src/lib/ErrorReporting")
+  require("./src/lib/ErrorReporting").setupSentry()
   require("./src/lib/AppRegistry")
 }
-
-// you can safely delete this comment. its only here to test a build trigger.
--- a/src/lib/ErrorReporting.ts
+++ b/src/lib/ErrorReporting.ts
@@ -1,4 +1,4 @@
-import Sentry from "@sentry/react-native"
+import * as Sentry from "@sentry/react-native"
 import { useEffect } from "react"
 import { Platform } from "react-native"
 import Config from "react-native-config"
--- a/src/lib/ErrorReporting.ts
+++ b/src/lib/ErrorReporting.ts
@@ -1,11 +1,11 @@
-import * as Sentry from "@sentry/react-native"
+import Sentry from "@sentry/react-native"
 import { useEffect } from "react"
 import { Platform } from "react-native"
 import Config from "react-native-config"
 import { sentryReleaseName } from "../../app.json"
 import { GlobalStore, useFeatureFlag } from "./store/GlobalStore"
 
-function setupSentry(props: Partial<Sentry.ReactNativeOptions> = {}) {
+export const setupSentry = (props: Partial<Sentry.ReactNativeOptions> = {}) => {
   if (!__DEV__ && Config.SENTRY_DSN) {
     Sentry.init({
       dsn: Config.SENTRY_DSN,
@@ -19,8 +19,6 @@ function setupSentry(props: Partial<Sentry.ReactNativeOptions> = {}) {
     })
   }
 }
-
-setupSentry()
 
 export function useSentryConfig() {
   const environment = GlobalStore.useAppState((store) => store.config.environment.env)
--- a/src/lib/utils/useScreenReaderTracking.ts
+++ b/src/lib/utils/useScreenReaderTracking.ts
@@ -1,21 +1,24 @@
-import { useEffect } from "react"
+import { useEffect, useState } from "react"
 import { AccessibilityInfo } from "react-native"
 import { AnalyticsConstants } from "./track/constants"
 import { SegmentTrackingProvider } from "./track/SegmentTrackingProvider"
 
 // the purpose of this hook is to track the screen reader status of the user's device
 export const useScreenReaderTracking = () => {
+  const [isScreenReaderEnabled, setIsScreenReaderEnabled] = useState(false)
   useEffect(() => {
-    const doIt = async () => {
+    const getScreenReaderStatus = async () => {
       const isEnabled = await AccessibilityInfo.isScreenReaderEnabled()
-      SegmentTrackingProvider.identify?.(null, {
-        [AnalyticsConstants.ScreenReaderStatus.key]: (() => {
-          return isEnabled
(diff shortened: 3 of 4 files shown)


ArtsyOpenSource:

- Dev changes (moved `setupSentry` in `index.ios.js` so it doesnt get called again - pavlos,removed a deprecated `enableOutOfMemoryTracking` from sentry init - pavlos,added an `Open RN Dev Menu` menu item in the admin menu, so we can open the RN dev menu on a device! - pavlos,reworked the `useScreenReaderTracking` to not use state - pavlos)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against e84dae8ee58ef400c1e2b2f3cc2da07d716e1766
</p>

src/lib/utils/useScreenReaderTracking.ts
```diff
@@ -1,24 +1,21 @@
...
 import { AccessibilityInfo } from "react-native"
 import { AnalyticsConstants } from "./track/constants"
 import { SegmentTrackingProvider } from "./track/SegmentTrackingProvider"
 
 // the purpose of this hook is to track the screen reader status of the user's device
 export const useScreenReaderTracking = () => {
-  const [isScreenReaderEnabled, setIsScreenReaderEnabled] = useState(false)
   useEffect(() => {
-    const getScreenReaderStatus = async () => {
+    const doIt = async () => {
```
olerichter00, line 9: Not sure if renaming it to `doIt` would be any good when someone tries to debug this code. This could be quite confusing.

gkartalis:
Tried to test it locally but it crashes with this error:
simulator_screenshot_F2DB2F12-EAFB-41F9-8FD3-513F44138C3C

olerichter00 (changes requested):
I like the title a lot 🥇 

Just left a small comment about renaming `getScreenReaderStatus`.

src/lib/ErrorReporting.ts
```diff
@@ -1,11 +1,11 @@
-import * as Sentry from "@sentry/react-native"
+import Sentry from "@sentry/react-native"
```
gkartalis, line 1: When reverted this line to previous changes it stopped crashing

src/lib/utils/useScreenReaderTracking.ts
```diff
@@ -1,24 +1,21 @@
-import { useEffect, useState } from "react"
+import { useEffect } from "react"
 import { AccessibilityInfo } from "react-native"
 import { AnalyticsConstants } from "./track/constants"
 import { SegmentTrackingProvider } from "./track/SegmentTrackingProvider"
 
 // the purpose of this hook is to track the screen reader status of the user's device
 export const useScreenReaderTracking = () => {
-  const [isScreenReaderEnabled, setIsScreenReaderEnabled] = useState(false)
```
MounirDhahri, line 8: **praise:** nice

MounirDhahri (commented):
Thanks for the refactoring

pvinis:
all of you have noticed this PR being broken. ill try to address things or revert things. trying to make PRs after a debugging frustration leads to funky PRs 😆.

src/lib/utils/useScreenReaderTracking.ts
pvinis: i think ill revert this whole file and try again later. i renamed it to `doIt` because i made it so that the useEffect has only one function, an async one, and then just call it. Since thats the "right" way to do async useEffects, and i removed the need for the useState, i thought ill be able to do it this way, and im kinda used to using `doIt` for this specific case.

unfortunately, this didnt work. also, this change broke the test, which btw @gkartalis i think the test might be not actually testing anything, because a mock is missing. maybe we can pair and try to remove the useState and play with the test together?

src/lib/ErrorReporting.ts
pvinis:
```suggestion
import * as Sentry from "@sentry/react-native"
```

src/lib/utils/useScreenReaderTracking.ts
gkartalis: Happy to pair on that, with the previous implementation the test was working though 🤔 , which mock is missing?

olerichter00: Instead of naming it `doIt` we could also use an IIFE.

```
useEffect(() => {
    (async () => {
      // do it!
    })()
  }, [])
```

pvinis:
there we go. we should be good now.

MounirDhahri (approved):
🚀

index.ios.js
```diff
@@ -17,8 +17,6 @@ if (startStorybook) {
 } else {
   require("react-native-gesture-handler")
   require("react-native-screens").enableScreens()
-  require("./src/lib/ErrorReporting")
+  require("./src/lib/ErrorReporting").setupSentry()
```
MounirDhahri, line 20: **thought:** I just realized that we do this initialization in AppRegistry, maybe we don't even need this actually https://github.com/artsy/eigen/blob/1fe9becf36c0c6420be4491b8653dde616913680/src/lib/AppRegistry.tsx#L420

olerichter00 (approved):

src/lib/utils/useScreenReaderTracking.ts
pvinis: yea, thats another way, both feel too boilerplate-y. but yea, i think the IIFE suggestion is cleaner.

index.ios.js
MounirDhahri: I agree

PR #5913, opened 2021-12-16
chore: send better formatted errors to sentry

MounirDhahri:

### Description

- The exceptions that we receive in Sentry are either named `<unknown>` or `RRNLRequestError`. This makes it harder to understand them unless we open their particular page. The goal from this PR is to take advantage of the `formatGraphQLErrors()` API in `react-relay-network-modern` to make our dashboard easier to understand.

Group 14

Group 13

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- send better formatted errors to sentry - mounir

</details>

### DIFF
--- a/src/lib/ErrorReporting.ts
+++ b/src/lib/ErrorReporting.ts
@@ -4,11 +4,8 @@ import Config from "react-native-config"
 import { sentryReleaseName } from "../../app.json"
 import { GlobalStore, useFeatureFlag } from "./store/GlobalStore"
 
-export const setupSentry = (
-  props: Partial<Sentry.ReactNativeOptions> = {},
-  captureExceptionsInSentryOnDev = !__DEV__
-) => {
-  if (!captureExceptionsInSentryOnDev && Config.SENTRY_DSN) {
+export const setupSentry = (props: Partial<Sentry.ReactNativeOptions> = {}, captureExceptions = !__DEV__) => {
+  if (!captureExceptions && Config.SENTRY_DSN) {
     Sentry.init({
       dsn: Config.SENTRY_DSN,
       release: sentryReleaseName,
--- a/src/lib/ErrorReporting.ts
+++ b/src/lib/ErrorReporting.ts
@@ -2,10 +2,13 @@ import * as Sentry from "@sentry/react-native"
 import { useEffect } from "react"
 import Config from "react-native-config"
 import { sentryReleaseName } from "../../app.json"
-import { GlobalStore } from "./store/GlobalStore"
+import { GlobalStore, useFeatureFlag } from "./store/GlobalStore"
 
-export const setupSentry = (props: Partial<Sentry.ReactNativeOptions> = {}) => {
-  if (!__DEV__ && Config.SENTRY_DSN) {
+export const setupSentry = (
+  props: Partial<Sentry.ReactNativeOptions> = {},
+  captureExceptionsInSentryOnDev = !__DEV__
+) => {
+  if (!captureExceptionsInSentryOnDev && Config.SENTRY_DSN) {
     Sentry.init({
       dsn: Config.SENTRY_DSN,
       release: sentryReleaseName,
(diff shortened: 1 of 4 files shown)


ArtsyOpenSource:

- Dev changes (send better formatted errors to sentry - mounir)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against cba017f3b6186dc564d97233b04a2d577e92b4ca
</p>

src/lib/utils/AdminMenu.tsx
```diff
@@ -389,3 +404,33 @@ const EnvironmentOptions: React.FC<{ onClose: () => void }> = ({ onClose }) => {
     </>
   )
 }
+
+export const FeatureFlagMenuItem: React.FC<{
```
MounirDhahri, line 408: This is probably not needed and we can use maybe with some effort adjust the `MenuItem`, again (we already adjusted it a few times to support new behaviors), to support multiple lines but I am afraid one day we will break it somewhere else in the app (it's used in 6 screens) and we won't notice it.

src/lib/relay/middlewares/errorMiddleware.ts
```diff
@@ -24,8 +23,14 @@ const throwError = (req: GraphQLRequest, res: RelayNetworkLayerResponse) => {
...
-    console.log(createRequestError(req, res))
-    Sentry.captureException(resJson.errors && resJson.errors[0]?.message)
+
+    const sentryFormattedError = createRequestError(req, res)
+
+    // All errors returned by createRequestError are ttled "RRNLRequestError", this makes it hard to identify which
+    // issues we should pay attention to in the issues list until we open the issue page separately
+    // We want to fix that by changing that title into a properly formatted error
+    sentryFormattedError.name = formatGraphQLErrors(req, res.errors!)
+    Sentry.captureException(sentryFormattedError)
```
admbtlr, line 33: But we're still going to have the problem that we get two Sentry issues for each error, aren't we? Is it possible to add some of the date in `SentryFormattedError` to the `createRequestError()`, or vice versa, and just throw one? Or is it maybe possible to have Sentry ignore the thrown `createRequestError()`?

admbtlr (changes requested):

src/lib/utils/AdminMenu.tsx
```diff
@@ -129,21 +129,38 @@ export const AdminMenu: React.FC<{ onClose(): void }> = ({ onClose = dismissModa
...
+          }}
+        />
+        <FeatureFlagMenuItem
+          title="Clear Relay Cache"
+          onPress={() => {
+            RelayCache.clearAll()
+          }}
+        />
+        <FeatureFlagMenuItem
+          title="Log out"
```
rajsam003, line 158:
```suggestion
          title="Logout"
```

rajsam003: Actually I am not sure. Haha

src/lib/utils/AdminMenu.tsx
rajsam003: **question:** How does it relate to the requirement of the ticket, which is about sentry error format?

MounirDhahri: it was hard to get feature flags that have a long description to be visible on the screen and we needed to always cut them, few people reported that (including you I guess). so I wanted to fix that in order to show a clear description.

src/lib/utils/AdminMenu.tsx
MounirDhahri: I wanted to have a verb here instead of a noun to match the other title for the other feature flags. Enable, show...

MounirDhahri: I referred to this https://www.quora.com/What-is-the-difference-between-log-out-and-logout

src/lib/relay/middlewares/errorMiddleware.ts
MounirDhahri: I actually checked that inside the clean `Eigen` sentry project (not `eigen-legacy`) and I noticed that the thrown error below does not show up in Sentry.

> Is it possible to add some of the date in `sentryFormattedError`
Yes, that's possible, I think the date is usually available by default but yes I can add it for easier debugging to avoid scrolling to the top of the page every time

src/lib/ErrorReporting.ts
```diff
@@ -2,10 +2,13 @@ import * as Sentry from "@sentry/react-native"
...
 import Config from "react-native-config"
 import { sentryReleaseName } from "../../app.json"
-import { GlobalStore } from "./store/GlobalStore"
+import { GlobalStore, useFeatureFlag } from "./store/GlobalStore"
 
-export const setupSentry = (props: Partial<Sentry.ReactNativeOptions> = {}) => {
-  if (!__DEV__ && Config.SENTRY_DSN) {
+export const setupSentry = (
+  props: Partial<Sentry.ReactNativeOptions> = {},
+  captureExceptionsInSentryOnDev = !__DEV__
```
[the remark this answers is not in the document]
MounirDhahri: Probably there is no need to add `InSentry` but the reason behind adding `onDev` here is to make sure whoever reads this later knows that this is only related to dev since we are always capturing errors in non-dev builds.
How about we go for `captureExceptionsInDev` ?

PR #5928, opened 2021-12-20
chore: Adding tooltip for Select components

rajsam003:

This PR resolves [CX-2231]

### Description

image

### To the reviewers 👀

- Tooltip for select components on storybooks

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-2231]: https://artsyproduct.atlassian.net/browse/CX-2231?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ArtsyOpenSource:

- Cross-platform user-facing changes (Tooltip for select components on storybooks)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 7eaac22bdbc4c143d71c982b146bfd7f86a2ba50
</p>

olerichter00 (commented):
Is this PR ready for review or still a draft?

rajsam003:
@olerichter00 Its ready now

src/palette/elements/Select/Select.stories.tsx
```diff
@@ -25,6 +25,14 @@ storiesOf("Select", module)
     <List>
       <Select title="Title" showTitleLabel={false} options={options} value="option-1" onSelectValue={() => null} />
       <Select title="Title" options={options} value={null} onSelectValue={() => null} />
+      <Select
+        title="Title"
```
olerichter00, line 29:
```suggestion
        title="With Tooltip"
```

src/palette/elements/Select/Select.tsx
```diff
@@ -108,23 +114,36 @@ const SelectButton: React.FC<{
...
+              {subTitle}
+            </Text>
+          ) : (
+            <Spacer mb={0.5} />
+          )}
+        </Flex>
+        {tooltip ? (
+          <Flex justifyContent="flex-end" marginLeft="auto">
+            <Text variant="xs" color="black60" mb={0.5} onPress={onPressTooltip}>
+              What is this?
```
olerichter00, line 142: Maybe we should pass the text that is shown as a prop instead of a boolean as `tooltip`.

src/palette/elements/Select/Select.tsx
```diff
@@ -108,23 +114,36 @@ const SelectButton: React.FC<{
...
-      )}
+          {subTitle ? (
+            <Text variant="xs" color="black60" mb={0.5}>
+              {subTitle}
+            </Text>
+          ) : (
+            <Spacer mb={0.5} />
+          )}
+        </Flex>
+        {tooltip ? (
```
olerichter00, line 139:
```suggestion
        {!!tooltip && (
```

olerichter00 (changes requested):
This looks pretty good so far! 

I would expect to pass the text ("What is this?") and the content of the tooltip as props to the component.

src/palette/elements/Select/Select.stories.tsx
```diff
@@ -25,6 +25,14 @@ storiesOf("Select", module)
...
+        options={options}
+        value={null}
+        onSelectValue={() => null}
+        tooltip
+        onPressTooltip={() => null}
```
pvinis, line 34: I think this story can get more useful with some actual tooltip. Maybe add something here? Then we can see how this thing is supposed to work.

src/palette/elements/Select/Select.tsx
```diff
@@ -108,23 +114,36 @@ const SelectButton: React.FC<{
   subTitle?: string
   placeholder?: string
   hasError?: boolean
+  tooltip?: boolean
   testID?: string
   onPress(): void
-}> = ({ value, placeholder, onPress, title, showTitleLabel, subTitle, hasError, testID }) => {
+  onPressTooltip?(): void
```
pvinis, line 120: What do you think about `onTooltipPress` or even `onHelpPress`?

Tooltip feels like the wrong name for this whole thing, since we don't actually show a tooltip, but we will just navigate to a help page.

Also, usually the names go `on*Press`, like `onLongPress`.

pvinis:
I tried with a subtitle as well, and it gets funky. Could you ask Jonathan about this case? Maybe the "whats this" stays on the same line as the title?
Screenshot 2021-12-20 at 18 53 25

src/palette/elements/Select/Select.tsx
pvinis, line 142: I wonder if having this text here makes sense, or if its better to replace the `tooltip` prop with something like `tooltipText` or `helpText` so we can customize.

Maybe ask Jonathan about this, and if he sees this changing often, lets to it with the title, if he sees this being the same for a while, then lets keep it as is.

pvinis (changes requested):
Had some ideas and some comments

olerichter00:
I guess this is a `feature` :)

src/palette/elements/Select/Select.stories.tsx
rajsam003: Yes, it is just for navigation. Lemme try adding a modal.

rajsam003:
> I tried with a subtitle as well, and it gets funky. Could you ask Jonathan about this case? Maybe the "whats this" stays on the same line as the title? Screenshot 2021-12-20 at 18 53 25

This makes sense. But in any case, it will overlap. I mean even if the title text is long. 
Let's see what Jonathan says.

olerichter00:
> > I tried with a subtitle as well, and it gets funky. Could you ask Jonathan about this case? Maybe the "whats this" stays on the same line as the title? Screenshot 2021-12-20 at 18 53 25
> 
> This makes sense. But in any case, it will overlap. I mean even if the title text is long. Let's see what Jonathan says.

You should probably give the `Flex` holding the title and the subtitle give the prop `flex: 1` to prevent the overlap. We then need to decide if restrict the number of lines for the title and the subtitle:

**Without line restriction:**
Screenshot 2021-12-28 at 11 18 48

**With line restriction:**
Screenshot 2021-12-28 at 11 19 14

rajsam003:
> > > I tried with a subtitle as well, and it gets funky. Could you ask Jonathan about this case? Maybe the "whats this" stays on the same line as the title? Screenshot 2021-12-20 at 18 53 25
> > 
> > 
> > This makes sense. But in any case, it will overlap. I mean even if the title text is long. Let's see what Jonathan says.
> 
> You should probably give the `Flex` holding the title and the subtitle the prop `flex: 1` to prevent the overlap. Not sure though whether to restrict the number of lines for the title and the subtitle:
> 
> **Without line restriction:** Screenshot 2021-12-28 at 11 18 48
> 
> **With line restriction:** Screenshot 2021-12-28 at 11 19 14

So according to Jonathan, we will want to test this in situ to check on any min-width issues. If the design is done well, probably we should not run into anything quite as sever (or narrow) as the [examples](https://www.figma.com/file/m6gDpKHEWDbYJyrwsVZDBr/Artsy-3.0-Design-System?node-id=5299%3A3124)

But I have added `ml={1}` to the tooltip flexbox to make it look good.
image

olerichter00:
It's still overlapping if the title is too long and I think we should address this.

simulator_screenshot_0DDF0EDF-811E-4A03-ACC8-79E123B5579A

src/palette/elements/Select/Select.tsx
```diff
@@ -108,23 +114,47 @@ const SelectButton: React.FC<{
...
+  onTooltipPress,
+}) => {
   const color = useColor()
   const textStyle = useTextStyleForPalette("sm")
 
   return (
     <Flex>
-      {showTitleLabel ? <Text variant="xs">{title?.toUpperCase()}</Text> : null}
+      <Flex flexDirection="row">
+        <Flex flex={1}>
```
rajsam003, line 139: This can help in adjusting the overlapping of texts. But it displays as overlapped only on storybook as storybook needs some adjustment with width.

PR #5954, opened 2021-12-28
feat: Add prefetching to bottom navigation links

olerichter00:

This PR resolves [CX-1781]

### Description

 Add prefetching to bottom navigation links.

### To the reviewers 👀

-  Add prefetching to bottom navigation links - ole

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-1781]: https://artsyproduct.atlassian.net/browse/CX-1781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/Scenes/Home/Home.tsx
+++ b/src/lib/Scenes/Home/Home.tsx
@@ -28,6 +28,7 @@ import {
   RandomWidthPlaceholderText,
   useMemoizedRandom,
 } from "lib/utils/placeholders"
+import { usePrefetch } from "lib/utils/queryPrefetching"
 import { ProvideScreenTracking, Schema } from "lib/utils/track"
 import { useTreatment } from "lib/utils/useExperiments"
 import { compact, times } from "lodash"
@@ -70,6 +71,14 @@ interface Props extends ViewProps {
 }
 
 const Home = (props: Props) => {
+  const prefetchUrl = usePrefetch()
+
+  useEffect(() => {
+    prefetchUrl("search")
+    prefetchUrl("my-profile")
+    prefetchUrl("inbox")
+  }, [])
+
   const {
     homePageAbove,
     homePageBelow,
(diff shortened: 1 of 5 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (Add prefetching to bottom navigation links - ole)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 976d30e7344a361e9a0b1a23f4a236b218c5ffe8
</p>

src/lib/Scenes/Home/Home.tsx
```diff
@@ -70,6 +71,14 @@ interface Props extends ViewProps {
 }
 
 const Home = (props: Props) => {
+  const prefetchUrl = usePrefetch()
+
+  useEffect(() => {
+    prefetchUrl("search")
+    prefetchUrl("my-profile")
+    prefetchUrl("inbox")
```
olerichter00, line 79: I moved the prefetching from the `BottomTabs` component to the home component because the effect wasn't always called. This works out because the home screen is the first screen when the user opens the app but I'm open to suggestions for a better place.

brainbicycle (approved):
Looks good to me!

brainbicycle:
**question:** not related to this PR but I see we are both using a treatment and a feature flag and was wondering what the reasoning was, it seems like we can turn off completely the treatment so wouldn't need the flag?

olerichter00:
> **question:** not related to this PR but I see we are both using a treatment and a feature flag and was wondering what the reasoning was, it seems like we can turn off completely the treatment so wouldn't need the flag?

You are right and we can disable prefetching via SplitIO. Nevertheless, I find the feature flag useful sometimes to turn off prefetching when debugging anything related to loading states.

pvinis:
> Nevertheless, I find the feature flag useful sometimes to turn off prefetching when debugging anything related to loading states.


sounds like a good devtoggle instead of flag 🤔. but yea, what you say makes sense.

olerichter00:
> It would be great to have a toggle in the Admin Settings to turn on/off every treatment. Similar to the feature flags toggles.

It would be great to have a toggle in the Admin Settings to turn on/off every treatment. Similar to the feature flags toggles.

brainbicycle:
> > It would be great to have a toggle in the Admin Settings to turn on/off every treatment. Similar to the feature flags toggles.
> 
> It would be great to have a toggle in the Admin Settings to turn on/off every treatment. Similar to the feature flags toggles.

Yeah makes sense hadn't considered the dev menu, maybe we could even make the treatment hook add this automatically to dev menu for treatments, thanks for explaining!

PR #5958, opened 2021-12-29
Bugfix(cx-2091): fix broken sharing to instagram stories

lordkiz:

This PR resolves [CX-2091]

### Description
- Fixes broken sharing to Instagram stories on android
- Replace sharing small images (because these images are blurry) with viewshot image

<p>

</p>

### To the reviewers 👀

- Fix sharing to Instagram Stories - kizito

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-2091]: https://artsyproduct.atlassian.net/browse/CX-2091?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/Scenes/Artwork/Components/ArtworkHeader.tsx
+++ b/src/lib/Scenes/Artwork/Components/ArtworkHeader.tsx
@@ -9,7 +9,7 @@ import { useCanOpenURL } from "lib/utils/useCanOpenURL"
 import { useScreenDimensions } from "lib/utils/useScreenDimensions"
 import { Box, Flex, InstagramAppIcon, LinkIcon, MoreIcon, ShareIcon, Spacer, WhatsAppAppIcon } from "palette"
 import React, { useRef, useState } from "react"
-import { Button, Modal, Platform } from "react-native"
+import { Button, Modal } from "react-native"
 import { ScrollView } from "react-native-gesture-handler"
 import Share from "react-native-share"
 import ViewShot from "react-native-view-shot"
@@ -36,11 +36,8 @@ export const ArtworkHeader: React.FC<ArtworkHeaderProps> = (props) => {
   const [shareSheetVisible, setShareSheetVisible] = useState(false)
   const toast = useToast()
 
-  const whatsappUrlScheme = Platform.OS === "android" ? "whatsapp://send?phone=+491898" : "whatsapp://test"
-  const showWhatsAppItem = useCanOpenURL(whatsappUrlScheme)
-  const instagramStoryUrlScheme =
-    Platform.OS === "android" ? "instagram://user?username=instagram" : "instagram-stories://test"
-  const showInstagramStoriesItem = useCanOpenURL(instagramStoryUrlScheme)
+  const showWhatsAppItem = useCanOpenURL("whatsapp://send?phone=+491898")
+  const showInstagramStoriesItem = useCanOpenURL("instagram://user?username=instagram")
 
   const currentImage = (artwork.images ?? [])[currentImageIndex]
   const currentImageUrl = (currentImage?.url ?? "").replace(":version", "large")
--- a/src/lib/Scenes/Artwork/Components/ArtworkHeader.tsx
+++ b/src/lib/Scenes/Artwork/Components/ArtworkHeader.tsx
@@ -15,7 +15,6 @@ import Share from "react-native-share"
 import ViewShot from "react-native-view-shot"
 import { createFragmentContainer, graphql } from "react-relay"
 import { useTracking } from "react-tracking"
-import RNFetchBlob from "rn-fetch-blob"
 import { ArtworkActionsFragmentContainer as ArtworkActions, shareContent } from "./ArtworkActions"
 import { ArtworkTombstoneFragmentContainer as ArtworkTombstone } from "./ArtworkTombstone"
 import { ImageCarouselFragmentContainer } from "./ImageCarousel/ImageCarousel"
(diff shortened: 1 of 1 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (Fix sharing to Instagram Stories - kizito)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 5b081d518fca6dc56f8ce86e31709619aff805fc
</p>

pvinis:
I will take a look at this tomorrow if thats ok. I need to run this.

pvinis:
would you mind explaining a bit how this PR fixes the android issue?

lordkiz:
> would you mind explaining a bit how this PR fixes the android issue?

@pvinis 
The issue on android is that we are not using the `shareArtworkOnInstagramStory` function that you had written because `const showInstagramStoriesItem = useCanOpenURL("instagram-stories://test")` wont work for android as this only checks the iOS urlScheme. So it defaults to using the general `shareArtwork` function which gets and attempts to share a plain image that does not meet the image spec as specified here: https://developers.facebook.com/docs/instagram/sharing-to-stories#shareable-content . All I did was replace the image in the `shareArtwork` function with the one we generate from ViewShot.

pvinis (changes requested):
Sorry it took a while to review.

This is not what we want here. The `shareArtwork` function is only called when we have the custom share flag disabled. It's the old share, and we don't need to change anything there.

To make sure you're testing the right thing, make sure you have the custom sharesheet flag enabled, and that you have instagram installed so that the `useCanOpenURL` returns true.

Only then you will see the actual issue, where the instagram app opens to share, but it's just a black screen.

lordkiz:
> Sorry it took a while to review.
> 
> This is not what we want here. The `shareArtwork` function is only called when we have the custom share flag disabled. It's the old share, and we don't need to change anything there.
> 
> To make sure you're testing the right thing, make sure you have the custom sharesheet flag enabled, and that you have instagram installed so that the `useCanOpenURL` returns true.
> 
> Only then you will see the actual issue, where the instagram app opens to share, but it's just a black screen.

@pvinis I see. With this approach we will still have the issue if the user opts to use the native android sharesheet. But I will look into it.

pvinis:
I guess we should enable the custom one by now, its been there for a while. So I'd say if you fix the android issue in this PR, we could just enable the flag by default, and soon after remove the old code 👍. How does that sound?

lordkiz:
> I guess we should enable the custom one by now, its been there for a while. So I'd say if you fix the android issue in this PR, we could just enable the flag by default, and soon after remove the old code 👍. How does that sound?

@pvinis that can work. How about we use the ViewShot in both the custom share and the default share? I used the default share for sharing to my drive etc and the image was really bad.

src/lib/Scenes/Artwork/Components/ArtworkHeader.tsx
```diff
@@ -36,12 +35,15 @@ export const ArtworkHeader: React.FC<ArtworkHeaderProps> = (props) => {
   const shotRef = useRef<ViewShot>(null)
   const [shareSheetVisible, setShareSheetVisible] = useState(false)
   const toast = useToast()
-  const showWhatsAppItem = useCanOpenURL("whatsapp://test")
-  const showInstagramStoriesItem = useCanOpenURL("instagram-stories://test")
+
+  const whatsappUrlScheme = Platform.OS === "android" ? "whatsapp://send?phone=+491898" : "whatsapp://test"
```
pvinis, line 39: did you try this and it didn't work on android?

if you tried and it didn't work, first of all, weird 🤔. does android need more things than just the protocol part before `://`? if it does, then let's keep that, for both android and ios, since ios only cares for the protocol. if it doesn't, lets keep again one of them, to keep things simple.

same for instagram.

pvinis (changes requested):
looks good, one thing is that we should just use either `whatsapp://test` or `whatsapp://send?phone=+491898`, no need to complicate this code with two if we can make it work with just one of them.

one more question, so does this change make instagram story share work? i dont see anything that would make any difference for the black screen 🤔. were we not rendering maybe.. could we do a quick call/pairing to make sure we are having the right behavior?

lordkiz:
> one more question, so does this change make instagram story share work? i dont see anything that would make any difference for the black screen 🤔. were we not rendering maybe.. could we do a quick call/pairing to make sure we are having the right behavior?

@pvinis Absolutely. This fixes the Instagram story share issue on android. The issue was because the previous image does not adhere to required spec.

src/lib/Scenes/Artwork/Components/ArtworkHeader.tsx
lordkiz: Updated the PR. Yes, they can share the same. Thanks @pvinis

pvinis (approved):
the code looks good. I haven't run it on a device so I'll trust you have tried that to make sure it works and posts a story.

you wanna remove the flag and use the new share and delete the old code in this PR or a separate one?

lordkiz:
> you wanna remove the flag and use the new share and delete the old code in this PR or a separate one?

The custom share sheet is enabled by default. By `remove old code`, do you mean remove the default share? If so, we can't because the user can opt to use the default share through the custom share sheet.

pvinis:
ok lets keep it as is and we can remove the flag later. ill assign you and sam, so if you have tried on android and it shares a story, then awesome, merge when you want.

PR #5989, opened 2022-01-06
fix: (CX-2253) fix skip bug in my collection artwork forms

lordkiz:

This PR resolves [CX-2253]

### Description

Fixes bug where previously selected my collection artwork form details is not cleared if a user goes back and clicks Skip.

https://user-images.githubusercontent.com/18648835/148369006-f9933e12-2d57-4399-a566-08b79c7b14a6.mp4

### To the reviewers 👀

- fix skip bug in my collection artwork forms - kizito

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-2253]: https://artsyproduct.atlassian.net/browse/CX-2253?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/Scenes/MyCollection/Screens/ArtworkForm/Screens/MyCollectionArtworkFormArtwork.tsx
+++ b/src/lib/Scenes/MyCollection/Screens/ArtworkForm/Screens/MyCollectionArtworkFormArtwork.tsx
@@ -71,12 +71,22 @@ export const MyCollectionArtworkFormArtwork: React.FC<StackScreenProps<ArtworkFo
 
   const navigateToNext = () => navigation.navigate("ArtworkFormMain", { ...route.params })
 
+  const skip = () => {
+    setLoading(true)
+    GlobalStore.actions.myCollection.artwork.ResetFormButKeepArtist()
+    // setTimeout because we must wait for all actions to resolve before loading the next set of forms
+    setTimeout(() => {
+      setLoading(false)
+      navigateToNext()
+    }, 1500)
+  }
+
   return (
     <>
       <FancyModalHeader
         onLeftButtonPress={route.params.onHeaderBackButtonPress}
         rightButtonText="Skip"
-        onRightButtonPress={navigateToNext}
+        onRightButtonPress={skip}
         hideBottomDivider
       >
         Select an Artwork
@@ -85,7 +95,7 @@ export const MyCollectionArtworkFormArtwork: React.FC<StackScreenProps<ArtworkFo
         <ScreenMargin>
           {!!formik.values.artistSearchResult && <ArtistSearchResult result={formik.values.artistSearchResult} />}
           <Spacer mb={2} />
-          <ArtworkAutosuggest onResultPress={updateFormValues} onSkipPress={navigateToNext} />
+          <ArtworkAutosuggest onResultPress={updateFormValues} onSkipPress={skip} />
         </ScreenMargin>
       </ScrollView>
       <LoadingModal isVisible={loading} />
--- a/src/lib/Scenes/MyCollection/Screens/ArtworkForm/Screens/MyCollectionArtworkFormArtwork.tsx
+++ b/src/lib/Scenes/MyCollection/Screens/ArtworkForm/Screens/MyCollectionArtworkFormArtwork.tsx
@@ -72,13 +72,10 @@ export const MyCollectionArtworkFormArtwork: React.FC<StackScreenProps<ArtworkFo
   const navigateToNext = () => navigation.navigate("ArtworkFormMain", { ...route.params })
 
   const skip = () => {
-    setLoading(true)
-    GlobalStore.actions.myCollection.artwork.ResetFormButKeepArtist()
-    // setTimeout because we must wait for all actions to resolve before loading the next set of forms
-    setTimeout(() => {
-      setLoading(false)
+    requestAnimationFrame(() => {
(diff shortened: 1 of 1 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (fix skip bug in my collection artwork forms - kizito)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against ae9b57f3df606562ab76f829f114bd764e975cc7
</p>

src/lib/Scenes/MyCollection/Screens/ArtworkForm/Screens/MyCollectionArtworkFormArtwork.tsx
```diff
@@ -71,12 +71,22 @@ export const MyCollectionArtworkFormArtwork: React.FC<StackScreenProps<ArtworkFo
...
   const navigateToNext = () => navigation.navigate("ArtworkFormMain", { ...route.params })
 
+  const skip = () => {
+    setLoading(true)
+    GlobalStore.actions.myCollection.artwork.ResetFormButKeepArtist()
+    // setTimeout because we must wait for all actions to resolve before loading the next set of forms
+    setTimeout(() => {
+      setLoading(false)
+      navigateToNext()
+    }, 1500)
```
olerichter00, line 81: I don't think this is a pattern we should use because this slows down the UI. Maybe `requestAnimationFrame` could be used here to achieve the same ([example](https://github.com/artsy/eigen/blob/ae9b57f3df606562ab76f829f114bd764e975cc7/src/lib/Scenes/MyCollection/Screens/ArtworkForm/Screens/MyCollectionArtworkFormArtwork.tsx#L65)

olerichter00 (commented):
I'm not 100% sure if this is the behaviour we want (and I'm aware that not clearing the form isn't optimal either). I started a conversation on the Jira ticket about my concern.

olerichter00:
> Hi there! 👋
> 
> We're trialing semantic commit formatting which has not been detected in your PR title.
> 
> Refer to [this RFC](https://github.com/artsy/README/issues/327#issuecomment-698842527) and [Conventional Commits](https://www.conventionalcommits.org) for PR/commit formatting guidelines.
> 
> Generated by 🚫 [dangerJS](https://danger.systems/js) against [ae9b57f](https://github.com/artsy/eigen/commit/ae9b57f3df606562ab76f829f114bd764e975cc7)

@lordkiz 👀

src/lib/Scenes/MyCollection/Screens/ArtworkForm/Screens/MyCollectionArtworkFormArtwork.tsx
lordkiz: Yesssss. Thanks @olerichter00 . I didn't think the issue could have stemmed from frames not being flushed. I thought I just had a slow store update. Thanks 🙌

olerichter00: I've experienced similar issues when updating the store in a callback. Hope `requestAnimationFrame` works.

olerichter00 (approved):
LGTM 👍

MounirDhahri:
Thanks for solving this Kizito.

However, I am wondering if we should do this inside the artwork details screen instead when the user presses on the back arrow/button. We can show them an alert saying are you sure you would like to go back, this will reset your form... and when they press yes, we revert the form values.

lordkiz:
> Thanks for solving this Kizito.
> 
> However, I am wondering if we should do this inside the artwork details screen instead when the user presses on the back arrow/button. We can show them an alert saying are you sure you would like to go back, this will reset your form... and when they press yes, we revert the form values.

@MounirDhahri I am not sure I understand. After the user presses back, for them to get back on the artwork detail form they will either have to select a new artwork or skip. If they select a new artwork, its properties will overwrite the previous artwork anyway. If they skip the form will be reset.

MounirDhahri:
> If they select a new artwork, its properties will overwrite the previous artwork anyway. If they skip the form will be reset.

Yes, that's true, I do however believe the logic to reset the details should live in the details screen. The main advantage of such an implementation, in my opinion, would be the Separation of concerns (SoC): if we ever implement another middle step where the user can fill in the form details, refactor the auto-suggest artwork screen, or reuse this form details screen somewhere else, we won't need to make sure that we reset the form values once the user presses on skip.

_Of course my comment here is a nitpick and definitely not a blocker_

lordkiz:
@MounirDhahri Ohhh. I understand now. You mean move the reset logic into the Back logic. That was my initial approach, but I quickly realised that I would have to also worry about android's hardware back button and iOS's back swipe. It would entail I add a 'beforeRemove' listener to navigation therefore rendering whatever logic is in `onHeaderBackButtonPress` not directly applicable to MyCollectionArtworkMain screen. While I don't see any problem with this approach, I felt it would be confusing to devs if they see multiple back logic for different screens in the same stack. 
But I see your point @MounirDhahri . Let me restore the branch and implement this your way and re-request your review so we can consider the merits/demerits.

lordkiz:
@MounirDhahri find this approach here: https://github.com/artsy/eigen/pull/6005

PR #6023, opened 2022-01-12
feat: support 2fa

brainbicycle:

This PR resolves [MOPLAT-2]

### Description

- Adds a basic screen for inputing 2fa codes
- Updates our sign in action to return more detailed statuses

<details><summary>Flow</summary>

https://user-images.githubusercontent.com/49686530/149218672-d0b73c1f-d312-4bbd-87a4-292c0fd49694.mp4

</details>

<details><summary>Screenshots</summary>

standard-otp
otp-on-demand
otp-error

</details>

## TODO

- [x] Review screen with design
- [x] Update generic error 
- [x] Figure out how to handle old versions that don't support 2fa
- [x] Support on demand otp requests
- [x] Write some tests

### To the reviewers 👀

- Support 2fa codes in sign in - Brian

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[MOPLAT-2]: https://artsyproduct.atlassian.net/browse/MOPLAT-2?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 7659227aa46d88dbe966b5a7bbd81410e3c58aee
</p>

ArtsyOpenSource:

- Cross-platform user-facing changes (Support 2fa codes in sign in - Brian)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against c79b5944da4a2dbdc7add7f2587ca8d29b2776f5
</p>

src/lib/store/AuthModel.ts
```diff
@@ -295,10 +297,23 @@ export const getAuthModel = (): AuthModel => ({
...
+    const resultJSON = await result.json()
+    if (resultJSON?.error === "otp_missing") {
+      return "otp_missing"
+    }
+
+    if (resultJSON?.error === "on_demand_otp_missing") {
+      return "on_demand_otp_missing"
+    }
+
+    if (resultJSON?.error_description === "invalid two-factor authentication code") {
```
joeyAghion, line 312: Out of curiosity, is there an `error` code matching this case, similar to the handling above? (I see an `otp_invalid` error code in Gravity, but not sure where it ends up.) If so, that seems neater than matching on the message language.

joeyAghion (dismissed):
CC-ing @mc-jones (who worked on back-end and web) in case you have additional feedback.

src/lib/store/AuthModel.ts
brainbicycle: hmm let me check, I think I copied this from force but if there is an error code it is definitely better

brainbicycle: the error I get back is `invalid_grant` seems like I can safely use that instead

joeyAghion: It [looks like](https://github.com/artsy/gravity/search?q=invalid_grant) `invalid_grant` is returned for a variety of possible errors. Maybe this is why Force checks the error description instead? @mc-jones can advise.

If that's the case, I think your existing implementation is fine rather than potentially showing a 2FA-specific message in a non-2FA scenario.

mc-jones: Yeah, believe @joeyAghion is correct. Looks like the original approach was to return `:invalid_grant` for any type of oAuth failure state (e.g. [missing email and missing password](https://github.com/artsy/gravity/blob/221436bd6a4a8f36263ca17fd093f63546942ddc/app/controllers/oauth_controller.rb#L149-L159)) and then inspect the description client side. In recent work, the error code has been slightly more specific, but the description is still the authority. For example, an error with `otp_missing` can be either a [missing two-factor authentication code](https://github.com/artsy/gravity/blob/221436bd6a4a8f36263ca17fd093f63546942ddc/app/controllers/oauth_controller.rb#L166) description or [missing on on-demand authentication code](https://github.com/artsy/gravity/blob/221436bd6a4a8f36263ca17fd093f63546942ddc/app/controllers/oauth_controller.rb#L187) description. 

This is different than how logins that go through `sessions_controller` are handled which, as you saw in `Force`, tends to rely on the error code for conditional logic.

Took me a bit to work through the differences a bit in my additions, and I definitely think we can do some work to be more consistent within each auth implementation.

mc-jones: Also, happy to pair if you run into issues and that's easier.

src/lib/Scenes/Onboarding/OnboardingLoginWithOTP.tsx
```diff
@@ -0,0 +1,142 @@
...
+    validationSchema: otpSchema,
+  })
+
+  return (
+    <FormikProvider value={formik}>
+      <OnboardingLoginWithOTPForm
+        navigation={navigation}
+        route={route}
+        email={email}
+        password={password}
```
MounirDhahri, line 137: **issue:** I think you don't need to send the `email` and `password` here since they're not being used inside `OnboardingLoginWithOTPForm`

src/lib/Scenes/Onboarding/OnboardingLogin.tsx
```diff
@@ -56,6 +59,12 @@ export const OnboardingLoginWithEmailForm: React.FC<OnboardingLoginProps> = ({ n
     return clearTimeout(timeout)
   }, [])
 
+  useEffect(() => {
+    if (otpMode) {
+      navigation.navigate("OnboardingLoginWithOTP", { email: values.email, password: values.password, otpMode })
```
MounirDhahri, line 64: **thought:** I think we better move this `navigate` inside the `onSubmit` at the formik form.

MounirDhahri (commented):
Thanks for leading this effort Brian. I left 2 minor comments, the rest looks fine for me. 
Btw, would you mind if you hide your changes behind a feature flag?

src/lib/Scenes/Onboarding/OnboardingLoginWithOTP.tsx
```diff
@@ -0,0 +1,142 @@
...
+    validateOnBlur: true,
+    initialValues,
+    initialErrors: {},
+    onSubmit: async ({ otp }, { setErrors, validateForm }) => {
+      validateForm()
+      const res = await GlobalStore.actions.auth.signIn({
+        oauthProvider: "email",
+        email,
+        password,
+        otp,
```
MounirDhahri, line 119: **nitpick:** it might be nice to trim the `otp` here or in the form above

src/lib/store/AuthModel.ts
brainbicycle: alright I updated them all to use the message for consistency, let me know if anything looks off!

src/lib/Scenes/Onboarding/OnboardingLoginWithOTP.tsx
brainbicycle: heh yep removed

src/lib/Scenes/Onboarding/OnboardingLogin.tsx
brainbicycle: 😅 why didn't I do that in the first place, thanks

MounirDhahri (approved):
Nice, thank you Brian 🎉

brainbicycle:
> Thanks for leading this effort Brian. I left 2 minor comments, the rest looks fine for me.
> Btw, would you mind if you hide your changes behind a feature flag?

Sorry I missed this comment, I left out the feature flag for this one because the flag in gravity on our application essentially acts as a feature flag for this. When that flag is not set we shouldn't get the error back that results in this screen.

MounirDhahri:
Great, thanks for answering back 🙏

PR #6060, opened 2022-01-19
feat: Optimize artwork image quality for widget GRO-679

jonallured:
When I first started working on the code for our iOS widget I made a choice to use the artwork image URL that comes from the payload. I started off using the `square` variant and then moved to the `larger` variant when I needed to respect the aspect ratio.

I knew that this was not optimal but I wasn't totally sure what I should be doing instead. After a pairing session with @dzucconi I came to understand that what I really wanted was to use the crop endpoint on Gemini to fetch an image that was the size I was actually going to render. I didn't quite know how I was going to do this but I did understand in theory what I should do.

Time passed.

I got it into my head this week that I should resolve this. So I started by picking an example artwork to establish a baseline.

```
$ http https://api.artsy.net/api/v1/artwork/JOHwLQB2iZlApcHOnLKpzg X-Access-Token:$TOKEN | jq ".images[0].image_url, .images[0].image_versions"
"https://d32dm0rphc51dk.cloudfront.net/JOHwLQB2iZlApcHOnLKpzg/:version.jpg"
[
  "large_rectangle",
  "medium_rectangle",
  "large",
  "tall",
  "small",
  "square",
  "larger",
  "medium",
  "normalized"
]
```

This gives us an image url template and a list of valid formats that we can substitute in for the version. Like I said at the top what I've been naively doing up to this point is subbing in `larger` and calling that good enough:

```
$ http https://d32dm0rphc51dk.cloudfront.net/JOHwLQB2iZlApcHOnLKpzg/larger.jpg > larger.jpg
$ du -sh larger.jpg
100K    larger.jpg
$ identify larger.jpg
larger.jpg JPEG 924x1024 924x1024+0+0 8-bit sRGB 102092B 0.000u 0:00.001
```

So that's our baseline - this image is 924x1024, in jpeg format and clocks in at 100K, can we do better?

## Well what size are we rendering it at?

The first thing we need to figure out is what size this image will be when rendered on our iOS widget. This is more complex than you might think because we have four variants to consider:

* full bleed
* featured artwork small
* featured artwork medium
* featured artwork large

<details>

<summary>Screenshot to remind you which one is which</summary>

146584373-af309c1a-d44a-4eb3-8179-a883a4ae5098

</details>

The size of the artwork image depends on which of these widgets the user has installed. Further complicating things are those last two variants. For them the container for the artwork image depends on whether it's running as primary or secondary. I bailed on this and instead decided to use the full size of the widget. Artwork images are never wider than the width of the widget so let's just do the dumbest thing possible and use that.

## What about pixel density?

Yep, that's in play too! We have to remember that iOS devices render at various pixel densities. Check out this from the HIG:

https://developer.apple.com/design/human-interface-guidelines/ios/icons-and-images/image-size-and-resolution/

So we have to know the size of the widget and the scaling factor for the device in order to make an appropriate crop call to Gemini.

## Where does this even happen?

The WidgetKit contract is that our extension code is called via a `getTimeline` method. We're meant to return a Timeline that includes Entry records that then get passed to a rendering function. The `getTimeline` call is where we parse the JSON that fuels the widget and discover the image urls. Those images are downloaded and included in the payload that goes to the rendering code. This is where we have our chance to alter the image urls to provide optimized ones.

And the great news is that when that method is called it's passed a `TimelineProviderContext` ([docs](https://developer.apple.com/documentation/widgetkit/timelineprovidercontext)) object that includes a method called `displaySize` that returns the width/height of the widget that the user has installed. Then we also have access to the density like this: `UIScreen.main.scale`. So we now have what we need to calculate the total width/height of the widget that we'll be rendering and the maximum size our artwork image needs to be:

```swift

let scale = UIScreen.main.scale
let size = context.displaySize

let width = size.width * scale
let height = size.height * scale
```

I ran this and found it to work great, here are some examples:

Screen Shot 2022-01-19 at 3 09 07 PM
Screen Shot 2022-01-19 at 3 09 46 PM

## What about webp?

You're so right we should totally tack onto this little experiment moving to webp rather than using jpegs. That's done with an optional Gemini flag called `convert_to` and we can pass `webp` there.

## Will this be worth it?

I wanted to know whether this would be worth it so I wrote a little Ruby script to pull down a JSON file for the widget and compare the various sizes of the artwork images. It computes the larger and normalized urls and then crops to 1,000 pixels and 500 pixels to simulate a 3x large and small widget. The cropped images are set to 50% quality and use webp:

```
$ ruby image_tinkering/profile_sizes.rb
image_tinkering/downloads/JOHwLQB2iZlApcHOnLKpzg
https://d7hftxdivxxvm.cloudfront.net/?convert_to=webp&height=1000&quality=50&resize_to=width&token=JOHwLQB2iZlApcHOnLKpzg&width=1000
fullBleedCropped.webp WEBP 1000x1108 1000x1108+0+0 8-bit sRGB 9606B 0.010u 0:00.012
larger.jpg JPEG 924x1024 924x1024+0+0 8-bit sRGB 102092B 0.000u 0:00.000
normalized.jpg JPEG 4458x4941 4458x4941+0+0 8-bit sRGB 2.75173MiB 0.000u 0:00.000
smallFeaturedCropped.webp WEBP 500x554 500x554+0+0 8-bit sRGB 3174B 0.010u 0:00.002

image_tinkering/downloads/M9SFRyRuUUwqnnJeeEAafA
https://d7hftxdivxxvm.cloudfront.net/?convert_to=webp&height=1000&quality=50&resize_to=width&token=M9SFRyRuUUwqnnJeeEAafA&width=1000
fullBleedCropped.webp WEBP 1000x1000 1000x1000+0+0 8-bit sRGB 35668B 0.020u 0:00.014
larger.jpg JPEG 1024x1024 1024x1024+0+0 8-bit sRGB 69234B 0.000u 0:00.000
normalized.jpg JPEG 3456x3456 3456x3456+0+0 8-bit sRGB 536761B 0.000u 0:00.000
smallFeaturedCropped.webp WEBP 500x500 500x500+0+0 8-bit sRGB 11240B 0.000u 0:00.003

image_tinkering/downloads/XX47G1ihglUqWAYCAO90hA
https://d7hftxdivxxvm.cloudfront.net/?convert_to=webp&height=1000&quality=50&resize_to=width&token=XX47G1ihglUqWAYCAO90hA&width=1000
fullBleedCropped.webp WEBP 1000x1373 1000x1373+0+0 8-bit sRGB 138842B 0.030u 0:00.027
larger.jpg JPEG 746x1024 746x1024+0+0 8-bit sRGB 760230B 0.000u 0:00.000
normalized.jpg JPEG 1720x2362 1720x2362+0+0 8-bit sRGB 3.38882MiB 0.000u 0:00.000
smallFeaturedCropped.webp WEBP 500x687 500x687+0+0 8-bit sRGB 50252B 0.000u 0:00.006

image_tinkering/downloads/sWGEYo1xK0VCyXGxpCeZAQ
https://d7hftxdivxxvm.cloudfront.net/?convert_to=webp&height=1000&quality=50&resize_to=width&token=sWGEYo1xK0VCyXGxpCeZAQ&width=1000
fullBleedCropped.webp WEBP 1000x1413 1000x1413+0+0 8-bit sRGB 158924B 0.030u 0:00.030
larger.jpg JPEG 725x1024 725x1024+0+0 8-bit sRGB 352990B 0.000u 0:00.000
normalized.jpg JPEG 4808x6795 4808x6795+0+0 8-bit sRGB 6.3011MiB 0.000u 0:00.000
smallFeaturedCropped.webp WEBP 500x707 500x707+0+0 8-bit sRGB 80094B 0.010u 0:00.008
```

Slightly easier to read like this:

```
$ du -sh image_tinkering/downloads/**/*.*
 12K    image_tinkering/downloads/JOHwLQB2iZlApcHOnLKpzg/fullBleedCropped.webp
100K    image_tinkering/downloads/JOHwLQB2iZlApcHOnLKpzg/larger.jpg
2.8M    image_tinkering/downloads/JOHwLQB2iZlApcHOnLKpzg/normalized.jpg
4.0K    image_tinkering/downloads/JOHwLQB2iZlApcHOnLKpzg/smallFeaturedCropped.webp

 36K    image_tinkering/downloads/M9SFRyRuUUwqnnJeeEAafA/fullBleedCropped.webp
 68K    image_tinkering/downloads/M9SFRyRuUUwqnnJeeEAafA/larger.jpg
528K    image_tinkering/downloads/M9SFRyRuUUwqnnJeeEAafA/normalized.jpg
 12K    image_tinkering/downloads/M9SFRyRuUUwqnnJeeEAafA/smallFeaturedCropped.webp

136K    image_tinkering/downloads/XX47G1ihglUqWAYCAO90hA/fullBleedCropped.webp
744K    image_tinkering/downloads/XX47G1ihglUqWAYCAO90hA/larger.jpg
3.4M    image_tinkering/downloads/XX47G1ihglUqWAYCAO90hA/normalized.jpg
 52K    image_tinkering/downloads/XX47G1ihglUqWAYCAO90hA/smallFeaturedCropped.webp

156K    image_tinkering/downloads/sWGEYo1xK0VCyXGxpCeZAQ/fullBleedCropped.webp
348K    image_tinkering/downloads/sWGEYo1xK0VCyXGxpCeZAQ/larger.jpg
6.3M    image_tinkering/downloads/sWGEYo1xK0VCyXGxpCeZAQ/normalized.jpg
 80K    image_tinkering/downloads/sWGEYo1xK0VCyXGxpCeZAQ/smallFeaturedCropped.webp
```

So for each of these four artworks you can see that the normalized version is huge, the larger version is much smaller but the webp cropped versions are damn tiny! Did we do better than our baseline? Hell yeah for the smallest widget we went from 100K to 4K and even at the largest size full bleed we still went from 100K to 12K - that's huge! I'll also note that the naive larger version I was using was actually being pixelated at 3 times scale even though the normalized version was more than big enough to accommodate. 😎 

## Wrap it up Jon

Ok I'm almost done - so this is what this PR does:

* use the context and the main UIScreen to compute the size of the total widget drawing area
* pull out the gemini token from the image data and ignore those image urls
* construct a gemini crop url with these inputs set to 50% quality and using webp

From there all the remaining code is unchanged - this is just about supplying the rendering code with a smaller about of binary image data. There's even more room to optimize here because again I'm actually using the entire drawing area when for some of these widgets the artworks are drawn smaller. I bailed on that when I saw how tiny these webp images were! 😍 

Final note: I have added and the removed the Ruby script I used to profile these image optimizations so I would like this PR to be rebased and NOT squashed so that it survives merging.

https://artsyproduct.atlassian.net/browse/GRO-679

/cc @artsy/grow-devs


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against c026e7b2693267b72869f4c476102356c8aa59f5
</p>

ArtsyWidget/Shared/ImageUrl.swift
```diff
@@ -0,0 +1,30 @@
...
+
+struct ImageUrl {
+    static func build(token: String, context: TimelineProviderContext) -> URL {
+        let geminiProxy = "https://d7hftxdivxxvm.cloudfront.net/"
+                
+        let scale = UIScreen.main.scale
+        let size = context.displaySize
+        
+        let width = "\(size.width * scale)"
+        let height = "\(size.height * scale)"
```
dzucconi, line 13: 👍

dzucconi:
I actually didn't know we could use a `token` param. Which version does this resize from, ultimately? (In all cases we want to be using the `normalized` version as the source from which we do the resizing.)

jonallured:
@dzucconi yeah! I find that using the token makes for a much more sane looking URL. 😝 

The way the Gemini crop endpoint works is that it ultimately finds an entry and then uses the `original_file` method to get at what I think we call normalized:

https://github.com/artsy/gemini/blob/7c327d67f92bd17121cde7f5b4fee2c5bb129b81/app/util/entry_image_processor.rb#L13
https://github.com/artsy/gemini/blob/main/app/models/entry.rb#L62

dzucconi:
Interesting! I wonder if there's a difference in initial performance? `normalized` != `original`, which would be much larger/possibly overkill.

pvinis (approved):
an amazing PR, but what else can we expect?
@jonallured being @jonallured, ladies and gentlemen.

PR #6072, opened 2022-01-21
feat: pass go back props

dimatretyak:

This PR resolves []

### Description
In some situations, it would be useful to pass props for `goBack` method. For example, in this way we can specify which screen the user returned from and perform the logic we need.

This PR adds this feature, and also solves the problem with executing unnecessary network requests for Saved Search Alert Flow when `goBack` event for `navigationEvents` is called (now when this event is called, requests are executed to get a list of all alerts + refetch information for a specific alert)

#### Before
In this example, the `SavedSearchesListQuery` query should not be executed, since this data is not used for the current screen

https://user-images.githubusercontent.com/3513494/150771252-39b08e2e-f443-4d87-8d60-d8b39d7b1c5e.mp4

#### After
https://user-images.githubusercontent.com/3513494/150771272-665a2394-96da-4771-8e09-85eb274e93e4.mp4

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- Add ability to pass props for goBack method - dimatretyak

</details>

### DIFF
--- a/src/lib/Scenes/ArtworkAttributionClassFAQ/ArtworkAttributionClassFAQ.tsx
+++ b/src/lib/Scenes/ArtworkAttributionClassFAQ/ArtworkAttributionClassFAQ.tsx
@@ -42,7 +42,7 @@ export const ArtworkAttributionClassFAQ: React.FC<Props> = ({ artworkAttribution
               works.
             </Text>
 
-            <Button onPress={goBack} block>
+            <Button onPress={() => goBack()} block>
               OK
             </Button>
           </Join>
--- a/src/lib/navigation/navigate.ts
+++ b/src/lib/navigation/navigate.ts
@@ -19,14 +19,14 @@ export interface ViewDescriptor extends ViewOptions {
   props: object
 }
 
-export interface BackProps {
-  fromScreen?: string
+export interface GoBackProps {
+  previousScreen?: string
 }
 
 export interface NavigateOptions {
   modal?: boolean
   passProps?: {
-    backProps?: BackProps
+    backProps?: GoBackProps
     [key: string]: any
   }
   replace?: boolean
@@ -167,7 +167,7 @@ export function dismissModal() {
   }
 }
 
-export function goBack(backProps?: BackProps) {
+export function goBack(backProps?: GoBackProps) {
   LegacyNativeModules.ARScreenPresenterModule.goBack(unsafe__getSelectedTab())
   navigationEvents.emit("goBack", backProps)
 }
--- a/src/lib/navigation/navigate.ts
+++ b/src/lib/navigation/navigate.ts
@@ -19,9 +19,16 @@ export interface ViewDescriptor extends ViewOptions {
   props: object
 }
 
+export interface BackProps {
+  fromScreen?: string
+}
+
 export interface NavigateOptions {
   modal?: boolean
-  passProps?: object
+  passProps?: {
+    backProps?: BackProps
+    [key: string]: any
+  }
   replace?: boolean
   // Only when onlyShowInTabName specified
   popToRootTabView?: boolean
@@ -160,9 +167,9 @@ export function dismissModal() {
   }
 }
 
-export function goBack() {
+export function goBack(backProps?: BackProps) {
   LegacyNativeModules.ARScreenPresenterModule.goBack(unsafe__getSelectedTab())
-  navigationEvents.emit("goBack")
+  navigationEvents.emit("goBack", backProps)
 }
 
 export function popParentViewController() {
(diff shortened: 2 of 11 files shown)


ArtsyOpenSource:

- Dev changes (Pass back props - dimatretyak)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 6ca3d888c56f4345b76a0a4835bb2ed95dba913f
</p>

src/lib/navigation/navigate.ts
```diff
@@ -19,9 +19,16 @@ export interface ViewDescriptor extends ViewOptions {
...
+  fromScreen?: string
+}
+
 export interface NavigateOptions {
   modal?: boolean
-  passProps?: object
+  passProps?: {
+    backProps?: BackProps
+    [key: string]: any
+  }
```
dimatretyak, line 31: maybe it's better to pass `backProps` not to `passProps`?

Something like this
```javascript
export interface NavigateOptions {
  modal?: boolean
  passProps?: object
  backProps?: BackProps
  replace?: boolean
  // Only when onlyShowInTabName specified
  popToRootTabView?: boolean
  showInTabName?: BottomTabType
}
```

gkartalis: 🤔  I think this is a better approach, although I don't think I really like the backProps naming, don't find it really explicit about what it is doing. Maybe using `fromScreen` directly is a better name?

src/lib/navigation/navigate.ts
```diff
@@ -160,9 +167,9 @@ export function dismissModal() {
   }
 }
 
-export function goBack() {
+export function goBack(backProps?: BackProps) {
```
gkartalis, line 170: something like this? 🤔 
```suggestion
export function goBack(fromScreen?: string) {
```

gkartalis (commented):
Love the solution, left a comment about the naming though, let me know what you think

src/lib/navigation/navigate.ts
gkartalis: or maybe something like that:
```javascript
 passProps?: {
    previousScreen?: string
    [key: string]: any
  }
```

src/lib/navigation/navigate.ts
dimatretyak: Yes, this option also makes sense. Just using the object may come in handy in the future, when we want to pass not only `fromScreen` prop (for example, some useful information from the screen or form data 🤔)

src/lib/Scenes/ArtworkAttributionClassFAQ/ArtworkAttributionClassFAQ.tsx
```diff
@@ -42,7 +42,7 @@ export const ArtworkAttributionClassFAQ: React.FC<Props> = ({ artworkAttribution
               works.
             </Text>
 
-            <Button onPress={goBack} block>
+            <Button onPress={() => goBack()} block>
```
MounirDhahri, line 45: **thought:** was this and the similar changes below required?

src/lib/navigation/navigate.ts
```diff
@@ -19,9 +19,16 @@ export interface ViewDescriptor extends ViewOptions {
   props: object
 }
 
+export interface BackProps {
+  fromScreen?: string
+}
```
MounirDhahri, line 24: I agree with George's comment below about the naming
```suggestion
export interface GoBackProps {
  previousScreen?: string
}
```

MounirDhahri (dismissed):
Nice PR, I left one minor renaming comment. It's not blocking though.

src/lib/navigation/navigate.ts
dimatretyak: Yes, otherwise there will be such a type incompatibility error
image

src/lib/Scenes/ArtworkAttributionClassFAQ/ArtworkAttributionClassFAQ.tsx
dimatretyak: Yes, otherwise there will be such a type incompatibility error
image

src/lib/navigation/navigate.ts
dimatretyak: Got it. I'll change it

src/lib/Scenes/ArtworkAttributionClassFAQ/ArtworkAttributionClassFAQ.tsx
MounirDhahri: aha I got you now. Thanks

PR #6099, opened 2022-01-25
style:hide loggers

patrinoua:

This PR makes logging ACTIONS and other things conditionally

**TODO:** 
You'll need to update the keys in [`emission.d.ts`](https://github.com/artsy/eigen/blob/869d35e0d83d4afae2cb62ebeab924f420944b0f/typings/emission.d.ts#L58-L72) and [`setupJest.ts`](https://github.com/artsy/eigen/blob/4654bacbcdc8624fb2799e9f86ad7717c5ab604b/src/setupJest.ts#L319-L331).

**TODO:** 
Remove Require cycle warnings 

- [ ] I have added a feature flag, or my changes don't require a feature flag. ([How do I add one?](https://github.com/artsy/eigen/blob/main/docs/developing_a_feature.md))
- [ ] I have added an app state migration, or my changes do not require one. ([What are migrations?](https://github.com/artsy/eigen/blob/main/docs/adding_state_migrations.md))

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/docs/adding_new_keys.md
+++ b/docs/adding_new_keys.md
@@ -2,6 +2,13 @@
 
 Keys are stored in `.env.shared` file, accessed through [`react-native-config`](https://github.com/luggit/react-native-config). You need to rebuild in Xcode for changes to the file to take effect.
 
+.env.shared is for artsy people to be able to work and compile.
+Is in .gitignore, and is downloaded by developers using the yarn setup:artsy script.
+It is also the main file that the app gets all the real env vars, keys etc.
+
+.env.example is for open source people to be able to work and compile and see what env vars need/use.
+It is committed in git, and we try to keep the exact layout copied over from .env.shared, but without any actual keys, we replace them with "-" or similar.
+
 On the React Native side:
 
 ```ts
--- a/docs/adding_new_keys.md
+++ b/docs/adding_new_keys.md
@@ -1,6 +1,6 @@
 # Adding a New Key
 
-Keys are stored in a `.env` file, accessed through [`react-native-config`](https://github.com/luggit/react-native-config). You need to rebuild in Xcode for changes to the file to take effect.
+Keys are stored in `.env.shared` file, accessed through [`react-native-config`](https://github.com/luggit/react-native-config). You need to rebuild in Xcode for changes to the file to take effect.
 
 On the React Native side:
 
@@ -29,3 +29,6 @@ BuildConfig.KEY_NAME;
 ```
 
 You'll need to update the keys in [`emission.d.ts`](https://github.com/artsy/eigen/blob/869d35e0d83d4afae2cb62ebeab924f420944b0f/typings/emission.d.ts#L58-L72) and [`setupJest.ts`](https://github.com/artsy/eigen/blob/4654bacbcdc8624fb2799e9f86ad7717c5ab604b/src/setupJest.ts#L319-L331).
+
+<!-- Shall we keep either .env.shared or .env.example?
+Updating the .env.shared seems to be enough. -->
--- a/src/lib/utils/hideLoggers.js
+++ b/src/lib/utils/hideLoggers.js
@@ -1,0 +1,9 @@
+export const ACTION_LOGGERS_ACTIVE = true
+export const RELAY_LOGGERS_ACTIVE = true
+export const DATADOG_LOGGERS_ACTIVE = true
+export const EVENT_TRACKED_LOGGERS_ACTIVE = true
+export const PREFETCHING_LOGGERS_ACTIVE = true
+export const NOTIFICATION_LOGGERS_ACTIVE = true
+export const OPERATION_LOGGERS_ACTIVE = true
+export const ROUTE_LOGGERS_ACTIVE = true
+export const RUNNING_REQUEST_LOGGERS_ACTIVE = true
--- a/src/lib/utils/queryPrefetching.ts
+++ b/src/lib/utils/queryPrefetching.ts
@@ -4,7 +4,6 @@ import { defaultEnvironment } from "lib/relay/createEnvironment"
 import { GlobalStore, useFeatureFlag } from "lib/store/GlobalStore"
 import { RateLimiter } from "limiter"
 import { useEffect } from "react"
-import Config from "react-native-config"
 import {
   createOperationDescriptor,
   Environment,
@@ -14,6 +13,7 @@ import {
   Variables,
 } from "relay-runtime"
 import RelayModernEnvironment from "relay-runtime/lib/store/RelayModernEnvironment"
+import { PREFETCHING_LOGGERS_ACTIVE } from './hideLoggers'
 import { useTreatment } from "./useExperiments"
 
 const DEFAULT_QUERIES_PER_INTERVAL = 60
@@ -85,7 +85,7 @@ const prefetchUrl = async (environment: Environment, url: string, variables: Var
   }
 
   const options = { ...result.params, ...variables }
-  if (Config.ACTION_LOGGERS_ACTIVE === "TRUE") {
+  if (PREFETCHING_LOGGERS_ACTIVE) {
     console.log(`Prefetching "${url}"`)
   }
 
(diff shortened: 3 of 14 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 2d0375c1104cb25f560482ac7eb55ad3b34f2f26
</p>

docs/adding_new_keys.md
```diff
@@ -29,3 +29,6 @@ BuildConfig.KEY_NAME;
 ```
 
 You'll need to update the keys in [`emission.d.ts`](https://github.com/artsy/eigen/blob/869d35e0d83d4afae2cb62ebeab924f420944b0f/typings/emission.d.ts#L58-L72) and [`setupJest.ts`](https://github.com/artsy/eigen/blob/4654bacbcdc8624fb2799e9f86ad7717c5ab604b/src/setupJest.ts#L319-L331).
+
+<!-- Shall we keep either .env.shared or .env.example?
```
pvinis, line 33: we want both.

`.env.shared` is in `.gitignore`, and is downloaded by developers using the `yarn setup:artsy` script. it is also the main file that the app gets all the real env vars, keys etc.

`.env.example` is committed in git, and if you notice, we try to keep the exact layout copied over from `.env.shared`, but without any actual keys, we replace them with `-` or similar.

basically `.env.shared` is for artsy people to be able to work and compile, and `.env.example` is for open source people to be able to work and compile and see what env vars need/use.

docs/adding_new_keys.md
```diff
@@ -1,6 +1,6 @@
 # Adding a New Key
 
-Keys are stored in a `.env` file, accessed through [`react-native-config`](https://github.com/luggit/react-native-config). You need to rebuild in Xcode for changes to the file to take effect.
+Keys are stored in `.env.shared` file, accessed through [`react-native-config`](https://github.com/luggit/react-native-config). You need to rebuild in Xcode for changes to the file to take effect.
```
pvinis, line 3: 👍

src/lib/utils/queryPrefetching.ts
```diff
@@ -84,8 +85,9 @@ const prefetchUrl = async (environment: Environment, url: string, variables: Var
   }
 
   const options = { ...result.params, ...variables }
-
-  console.log(`Prefetching "${url}"`)
+  if (Config.ACTION_LOGGERS_ACTIVE === "TRUE") {
```
pvinis, line 88: i dont think this is a good way to do this. let me explain.

as a developer, writing `console.log` is quick and easy to add and remove from the code.
adding 3 lines, to do the same, and an import, is definitely not something i would want to do.

i think this idea is really good, to have a better log situation. what i would suggest, and we can pair if you want too, is to use our own `log` function, or just wrap the `console` so that `log` and `warn` etc are our own functions. this way we can add this env var check within the logger functions.

this way i will still just write one call, either `console.log` or something like `log`, and it will do the checks needed under the hood.

how does that sound?

pvinis (changes requested):
i think we can implement this great idea in a much more friendly way for DX.

docs/adding_new_keys.md
MounirDhahri: **issue:** I thought we have patched react-native-config to use the `.env.shared` in Dev https://github.com/artsy/eigen/blob/11dda2712a2b6cb122e32b6db3a7e7e76af0ec5b/patches/react-native-config+1.3.3.patch#L28

docs/adding_new_keys.md
MounirDhahri: You can take a look at our shared configuration docs [here](https://github.com/artsy/README/blob/main/playbooks/development-environments.md#shared-configuration). we're still missing the `.env` but the mobile platform team will probably eventually add it for consistency with other products

src/lib/utils/queryPrefetching.ts
MounirDhahri: I agree with Pavlos comment here, I also think that we should not have these coming from the env vars as this will mean we will need to build the project again everytime, I think we can achieve the same thing using globals like in here https://github.com/artsy/eigen/blob/9007c1cb4346b9eac90e1e78abf4c91262ec911e/index.ios.js#L2

MounirDhahri (changes requested):
Thanks for pushing this through, I look forward to being able to remove those console logs

patrinoua:
@pvinis @MounirDhahri thank you both for reviewing and the suggestions! 
I went for an env variable because someone mentioned it at some point and it sounded good, however of course we can try to do it in a better way and at least now we have all the places where the console logs are coming from in one place.
I still didn't manage to locate a couple and turn off the warnings so maybe that's also something to check together.

docs/adding_new_keys.md
MounirDhahri: yeah that would be great

patrinoua:
@pvinis @MounirDhahri after your comments & suggestions I took another approach, removed the .env variables and added flags for our loggers.

I have changed everything so you can give me some feedback and we can merge this either now or later after metaflags file is also public, I just thought it will be nice to have some of it ready for whenever we want to use it.

Thanks again! 🙏

src/lib/utils/hideLoggers.js
```diff
@@ -0,0 +1,9 @@
+export const ACTION_LOGGERS_ACTIVE = true
```
pvinis, line 1: I have two issues here:
- i think we should have these `false` by default. if devs want to turn these off, they can, but I think having these off by default might be confusing?
- I still think we should use the metaflags (which is coming soon from moplat) instead of this, for the above reason. if devs have the change locally, then they will have a git diff thats not empty at all times, and they need to remember to not commit this. having something like the metaflags will be ignored, so locally each dev can do anything they like.

I want to investigate if these do show up in the crash logs. I am pretty sure the console.logs go directly to the system log for each app, and these can be exported. I would leave all these there for production. If that is the case, I would say we only look at these flags when `__DEV__ === true`, and not in other cases.

So in general, even though I looooove this idea of better loggers, since ours right now is a bit of a mess, I think we should take a look in mobile practice, ideally after the metaflags are merged. Working with flipper in the meantime might be the best solution :(

pvinis (changes requested):

src/lib/utils/hideLoggers.js
patrinoua: Thanks Pavlos, currently the flags are set to **true**, so the console logs would appear properly.
Someone could save them to false if they want to hide them.
However we can of course use the metaflags, I thought maybe this will make it easier to detect all the files that need to be touched when we do that. 
I'll continue with local filtering in the mean time. thank you :)

MounirDhahri: I agree with Angela, I think we should still set the flags to true since we are using them right now for Sentry and soon hopefully we can get rid of all of them and replace them with just `addBreadcrum` from sentry

pvinis:
we will get back to this right after https://github.com/artsy/eigen/pull/6169, so we can use metaflags.json for the console flags.

pvinis:
we will get back to this right after https://github.com/artsy/eigen/pull/6169, so we can use metaflags.json for the console flags.

src/lib/utils/hideLoggers.js
```diff
@@ -0,0 +1,9 @@
...
+export const RELAY_LOGGERS_ACTIVE = true
+export const DATADOG_LOGGERS_ACTIVE = true
+export const EVENT_TRACKED_LOGGERS_ACTIVE = true
+export const PREFETCHING_LOGGERS_ACTIVE = true
+export const NOTIFICATION_LOGGERS_ACTIVE = true
+export const OPERATION_LOGGERS_ACTIVE = true
+export const ROUTE_LOGGERS_ACTIVE = true
+export const RUNNING_REQUEST_LOGGERS_ACTIVE = true
```
damassi, line 9: #superminor - could we alphabetize this list for readability?

PR #6107, opened 2022-01-26
chore: rewrite my collection artwork using relay hooks

MounirDhahri:

### Description

- This PR implements my collection artwork using relay hooks instead of the existing HOCs approach. This comes up as a follow-up on the https://github.com/artsy/eigen/issues/6080

**Why is this PR a draft?**
We still have to try using `useFragment`,  `useRefetchableFragment `, `usePaginationFragment` and `useMutation` at our next KS meeting

-

#### iOS user-facing changes

-

-

#### Dev changes

- Rewrite my collection artwork using relay hooks - cx team

</details>

### DIFF
--- a/src/lib/Scenes/MyCollection/Screens/Artwork/MyCollectionArtwork.tsx
+++ b/src/lib/Scenes/MyCollection/Screens/Artwork/MyCollectionArtwork.tsx
@@ -13,18 +13,16 @@ import {
 import { Divider } from "lib/Components/Bidding/Components/Divider"
 import { FancyModalHeader } from "lib/Components/FancyModal/FancyModalHeader"
 import { navigate, popToRoot } from "lib/navigation/navigate"
-import { defaultEnvironment } from "lib/relay/createEnvironment"
 import { ScreenMargin } from "lib/Scenes/MyCollection/Components/ScreenMargin"
 import { MyCollectionArtworkInsightsFragmentContainer } from "lib/Scenes/MyCollection/Screens/Artwork/Components/ArtworkInsights/MyCollectionArtworkInsights"
 import { GlobalStore } from "lib/store/GlobalStore"
 import { PlaceholderBox, PlaceholderText } from "lib/utils/placeholders"
-import { renderWithPlaceholder } from "lib/utils/renderWithPlaceholder"
 import { ProvideScreenTrackingWithCohesionSchema } from "lib/utils/track"
 import { screen } from "lib/utils/track/helpers"
 import { Button, Flex, Join, Spacer, Text } from "palette"
-import React from "react"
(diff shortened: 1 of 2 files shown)


ArtsyOpenSource:

- Dev changes (Rewrite my collection artwork using relay hooks - cx team)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 0cba105890b96c9e7de97272035cc54123fcf829
</p>

src/lib/Scenes/MyCollection/Screens/Artwork/MyCollectionArtwork.tsx
```diff
@@ -149,47 +147,58 @@ export const ArtworkMetaProps = graphql`
...
+}) => {
+  const data = useLazyLoadQuery<MyCollectionArtworkQuery>(
+    MyCollectionArtworkScreenQuery,
+    {
+      artworkSlug,
+      artistInternalID,
+      medium,
+    },
+    {
+      fetchPolicy: "store-and-network",
```
pvinis, line 178: based on this and reading https://relay.dev/docs/guided-tour/reusing-cached-data/fetch-policies/, i have two questions:

- why `and` and not `or`? i get the difference in functionality, but i dont get why we prefer that functionality.
- if we decide on `and`, would we want to make this our default everywhere?

Ill ping @olerichter00 here since he has some more context too.

olerichter00: If the artwork is already in the cache we don't have to show any loading state and because of the additional fetch we will have up to date data as soon as the request returns. It's basically the [stale-while-revalidate caching pattern](https://www.google.com/search?q=stale-while-revalidate+caching&oq=stale-while-revalidate+caching&aqs=chrome..69i57j0i22i30l2.2855j0j7&sourceid=chrome&ie=UTF-8).

olerichter00: We could experiment and see if it makes sense for us to use this pattern. It would be great if we could find a way to [retail queries](https://relay.dev/docs/guided-tour/accessing-data-without-react/retaining-queries/) so they don't get garbage collected or save the queries somehow (e.g. with [https://github.com/morrys/react-relay-offline](https://github.com/morrys/react-relay-offline)).

pvinis: gotcha. both sound like good ideas. thanks for the explanation.

MounirDhahri:
Closing PR for inactivity - please feel free to reopen it if needed 👌

PR #6117, opened 2022-01-27
fix: Broken back navigation

MounirDhahri:
**This is a long description for very few lines of code, but it was fun to debug 😄**

This PR resolves [CX-2285]

### Issue description
**Reproduction**
1. Open the app
2. Select an artwork (**Artwork X**)
3. Scroll down to the end of the page to see related artworks
4. Select an artwork (**Artwork Y**)
5. Press back

**Expected result**
Get back to **Artwork X** screen 🍾

**Actual result**
User lands on **Artwork Y** screen 😲
(video below)

### How I debugged the issue
**What I noticed**
When I press the back button, the screen is already loaded with the **Artwork Y** Content without any loading at all. Usually, there should be some loading even if its props are cached in the relay store.

**My theory (might be right or might be wrong and might be both right and wrong at the same time 😵‍💫)**
The screen could be actually getting rendered on the background 🤔

**How can I confirm my theory?**
Obviously, I will write some console.logs 😄. I will add one at the top of the artwork screen render block and see if **by navigating to one artwork screen to an other, we are going to get the same number of console.logs** (in a perfect world, it should be just one `console.log` and there will be no re-render but yeah, let's be honest, this is not the case)

**Output**
```
/* I navigate to artwork X */
render artwork screen X// console.log("render artwork screen X")
render artwork screen X// console.log("render artwork screen X")
/* I navigate to artwork Y */
render artwork screen Y// console.log("render artwork screen Y")
render artwork screen Y// console.log("render artwork screen Y")
render artwork screen X// console.log("render artwork screen X")
render artwork screen X// console.log("render artwork screen X")
```

EasygoingGrayAntipodesgreenparakeet-max-1mb

We got **double** `console.log` when navigating to **Artwork Y** screen. by navigating to **Artwork Z** we even got more!
```
/* I navigate to artwork Z */
render artwork screen Z// console.log("render artwork screen Z")
render artwork screen Z// console.log("render artwork screen Z")
render artwork screen Y// console.log("render artwork screen Y")
render artwork screen Y// console.log("render artwork screen Y")
render artwork screen X// console.log("render artwork screen X")
render artwork screen X// console.log("render artwork screen X")
```

Screenshot 2022-01-27 at 17 41 37

**Yeah** kind of... there is one way to be sure, and that's our `PageWrapper` inside `Appregistry.tsx` since that's our highest HOC. Again, how can I confirm it?
consolelog-consolelog-everywhere

Screenshot 2022-01-27 at 17 44 51

And here I noticed that this is not optimized and the props better would better live inside a reference as so
```
  pageProps: PageWrapperProps

  constructor(props: PageWrapperProps) {
    super(props)
    this.pageProps = {
      ...this.props,
      viewProps: {
        ...this.props.viewProps,
        ...propsStore.getPropsForModule(this.props.moduleName),
      },
    }
  }
...
```

### How I should have debugged the issue
Probably [Dichotomic search](https://en.wikipedia.org/wiki/Dichotomic_search) using:
- [Git-bisect](https://git-scm.com/docs/git-bisect)
or
- simpler and better IMO in this case: just installing old builds from Testflight until I find the first broken build (it was **7.1.3 2021.12.10.6**), then checking when it got broken in git history

**Before**

https://user-images.githubusercontent.com/11945712/151406848-ff184fa3-5c48-4ed3-83f6-2a378a978672.mov

**After**

https://user-images.githubusercontent.com/11945712/151406887-7195dc81-c69b-414a-9de7-267bf8cd1439.mov

### To the reviewers 👀

-

#### iOS user-facing changes

- fix back navigation on iOS - mounir

-

#### Dev changes

-

</details>

[CX-2285]: https://artsyproduct.atlassian.net/browse/CX-2285?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/AppRegistry.tsx
+++ b/src/lib/AppRegistry.tsx
@@ -227,17 +227,23 @@ const InnerPageWrapper: React.FC<PageWrapperProps> = ({
 // provide the tracking context so pages can use `useTracking` all the time
 @track()
 class PageWrapper extends React.Component<PageWrapperProps> {
-  render() {
-    const props = {
+  pageProps: PageWrapperProps
+
+  constructor(props: PageWrapperProps) {
+    super(props)
+    this.pageProps = {
       ...this.props,
       viewProps: {
         ...this.props.viewProps,
         ...propsStore.getPropsForModule(this.props.moduleName),
       },
     }
+  }
+
+  render() {
     return (
       <AppProviders>
-        <InnerPageWrapper {...props} />
+        <InnerPageWrapper {...this.pageProps} />
       </AppProviders>
     )
   }


ArtsyOpenSource:

- iOS user-facing changes (fix back navigation on iOS - mounir)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 55dbe9b60820d8986c95a9ee519c2754ca7a7f1c
</p>

dimatretyak:
👏 Special thanks for the detailed description and memes

rajsam003:
Love this memes description 😂🙌

src/lib/AppRegistry.tsx
```diff
@@ -227,17 +227,23 @@ const InnerPageWrapper: React.FC<PageWrapperProps> = ({
 // provide the tracking context so pages can use `useTracking` all the time
 @track()
 class PageWrapper extends React.Component<PageWrapperProps> {
```
rajsam003, line 229: **Non-blocker:** Just a thought, why don't we consider changing this component into a functional component?

olerichter00 (approved):
🤩

src/lib/AppRegistry.tsx
MounirDhahri: I actually first started with that and had to update react-tracking since hooks were not supported by the version we have. I will have another PR open in a bit to handle that. The only way with the version we have is using the `@track` decorator

Screenshot 2022-01-28 at 12 33 16

Screenshot 2022-01-28 at 12 33 20

rajsam003 (approved):
Thanks for the change 👍🏽

MounirDhahri:
Here is the follow-up PR @rajsam003  https://github.com/artsy/eigen/pull/6119/files

pvinis:
Awesome to see this fixed, nice job mounir.

I have to say im not 100% clear on how the code fixes this, but from what i understand, basically we keep the props out of render instead of making them on every render. kinda like state. and im also guessing that somehow the props would change on render, but now that they are outside render and not changing, they work.

i might be very wrong, or i might be right. please help me understand 😅, whenever you got a min, thank you.

MounirDhahri:
Yes, sure. Glad we spoke about it at the QA session today

PR #6119, opened 2022-01-28
chore: update react tracking

MounirDhahri:

This PR updates **react-tracking** to the latest version

### Description

Update react tracking from v7 to v9. The latest version has full hooks support 🍾

cc @pvinis 

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- Update react-tracking to the latest version - mounir

</details>


package.json
```diff
@@ -220,7 +220,7 @@
     "@types/react-native-push-notification": "^7.3.3",
     "@types/react-relay": "11.0.3",
     "@types/react-test-renderer": "^16.9.3",
-    "@types/react-tracking": "7.0.1",
+    "@types/react-tracking": "^8.1.2",
```
MounirDhahri, line 223: `v9.1.0` types are still not available. There are almost no differences in the types between v8 and v9 except for the debug mode types. I did have to upgrade to v8 though which supports `Track` inside `useTracking`

ArtsyOpenSource:

- Dev changes (Update react-tracking to the latest version - mounir)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 277b57523054b93cff44f8c02ffb6ea087a51ef8
</p>

package.json
```diff
@@ -167,7 +167,7 @@
     "react-native-webview": "^11.3.1",
     "react-relay": "12.0.0",
     "react-spring": "8.0.23",
-    "react-tracking": "7.2.1",
+    "react-tracking": "^9.1.0",
```
pvinis, line 170:
```suggestion
    "react-tracking": "9.1.0",
```

we are really trying to conform with this idea of pinning exact versions, so i would suggest we keep the `^` out.

package.json
pvinis, line 223:
```suggestion
    "@types/react-tracking": "8.1.2",
```
same

package.json
pvinis: also, i have to say..


UPGRADES WOOP WOOP 💜 🥳 🎉

src/lib/Scenes/MyCollection/Screens/Artwork/Components/MyCollectionArtworkHeader.tsx
```diff
@@ -71,7 +71,7 @@ export const MyCollectionArtworkHeader: React.FC<MyCollectionArtworkHeaderProps>
           images={imagesToDisplay as any}
           cardHeight={dimensions.height / 2.5}
           paginationIndicatorType="scrollBar"
-          onImagePressed={trackEvent(tracks.tappedCollectedArtworkImages(internalID, slug))}
+          onImagePressed={() => trackEvent(tracks.tappedCollectedArtworkImages(internalID, slug))}
```
pvinis, line 74: nice catch

package.json
MounirDhahri: done

package.json
MounirDhahri: done

pvinis:
did gh just lose my comments? omg

src/lib/AppRegistry.tsx
```diff
@@ -224,29 +224,20 @@ const InnerPageWrapper: React.FC<PageWrapperProps> = ({
   )
 }
 
-// provide the tracking context so pages can use `useTracking` all the time
-@track()
```
pvinis, line 228: here is what i talked about above. i like that we turn this to a func, but we still need this `@track`.

- if react-tracking has another way to add this, lets use that.
- if they dont, should we make a class comp that is just a wrapper? something like
```
@track()
class PageWrapperWrapper extends Component {
  render() {
    return <PageWrapper {...this.props} /> 
  }
}
```

src/lib/AppProviders.tsx
```diff
@@ -11,28 +12,34 @@ import { defaultEnvironment } from "./relay/createEnvironment"
...
-              </ActionSheetProvider>
-            </RetryErrorBoundary>
-          </Theme>
-        </RelayEnvironmentProvider>
-      </GlobalStoreProvider>
-    </ProvideScreenDimensions>
-  </SafeAreaProvider>
-)
+export const AppProviders = ({ children }: { children: ReactNode }) => {
+  const { Track } = useTracking()
```
pvinis, line 16: i dont think this will work? in order for us to use this hook here, we need to have a tracking context in one of the parents. the context was provided by `@track` unfortunately, the one in the PageWrapper.

pvinis (changes requested):

MounirDhahri:
I can see them but they're resolved for some reason 🤷‍♂️, this happened today's morning with me and @patrinoua and it was resolving comments with no reason 😵‍💫 something is wrong with github

pvinis:
ah no no, the others i resolved, after you posted `done`. i had two more that were lost, i rewrote them, the ones about `@track`.

src/lib/AppRegistry.tsx
MounirDhahri: Thanks for the comment, as per v8, the `@track` decorator can be replaced by wrapping the app with <Track />. that was the main reason of my upgrade after this comment here [PR](https://github.com/artsy/eigen/pull/6117#discussion_r794198292)

src/lib/AppProviders.tsx
MounirDhahri: It actually works this way, I followed the react-tracking docs. I also tried it on my simulator. 
Without the decorator and the track I get this error
Screenshot 2022-01-28 at 12 33 20
When using `<Track>` it's solved

MounirDhahri:
ah oki, I just replied to your comments

pvinis:
interesting. ok if you tested this and it works, cool. thanks for the comments.

src/lib/Scenes/MyCollection/Screens/Artwork/Components/MyCollectionArtworkHeader.tests.tsx
```diff
@@ -79,8 +79,7 @@ describe("MyCollectionArtworkHeader", () => {
       }),
     })
     wrapper.root.findAllByType(ImageWithLoadingState)[0].props.onPress()
-    // expect 2 calls: 1 for our custom tracking, 2 for tracking deep zoom taps
-    expect(mockTrackEvent).toHaveBeenCalledTimes(2)
+    expect(mockTrackEvent).toHaveBeenCalledTimes(1)
```
MounirDhahri, line 82: this was happening because of 
`onImagePressed={trackEvent(tracks.tappedCollectedArtworkImages(internalID, slug))}`
I am not sure how I missed it

pvinis (approved):

src/lib/Scenes/MyCollection/Screens/Artwork/Components/MyCollectionArtworkHeader.tests.tsx
MounirDhahri, line 82: this was happening because of 
`onImagePressed={trackEvent(tracks.tappedCollectedArtworkImages(internalID, slug))}`
I am not sure how we missed it

rajsam003 (commented):
I am not much familiar with `react-tracking` but the code looks good to me

PR #6153, opened 2022-02-03
fix: Deactivate placeholder layout animation for android

olerichter00:

🔴 Depends on https://github.com/artsy/echo/pull/198

### Description

This PR 
- deactivates the placeholder layout animation for Android devices (because it causes issues), 
- adds an Echo feature flag, 
- reactors the placeholder animation hook,
- and sets the layout duration to 100ms.

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/lib/utils/placeholders.tsx
+++ b/src/lib/utils/placeholders.tsx
@@ -1,4 +1,5 @@
 import { GenericGridPlaceholder } from "lib/Components/ArtworkGrids/GenericGrid"
+import { useFeatureFlag } from "lib/store/GlobalStore"
 import { Flex, useColor } from "palette"
 import React, { useContext, useEffect, useMemo, useRef } from "react"
 import { LayoutAnimation, Platform, View, ViewStyle } from "react-native"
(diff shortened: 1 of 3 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 5e46407e820ac9b89af8217c44987240d8ae4bb6
</p>

src/lib/utils/placeholders.tsx
```diff
@@ -36,8 +38,28 @@ function useCurrentTime() {
   return clock
 }
 
+const useLayoutAnimation = () => {
+  const enablePlaceholderLayoutAnimation = useFeatureFlag("AREnablePlaceholderLayoutAnimation")
+
+  // Deactivating the layout animation for Android because it causes problems.
+  if (Platform.OS === "android" || !enablePlaceholderLayoutAnimation) {
```
MounirDhahri, line 45: **issue:** I believe this should be instead moved inside the `useEffect` to avoid the early return breaking the hooks rules. Otherwise we might be getting an error "number of hooks should be the same...." or something like that

olerichter00: This shouldn't happen unless an iOS device suddenly becomes an Android device, but you're right, it's not very clean. I moved the check to avoid the issue 👍

MounirDhahri (approved):
🟢🟢🟢

olerichter00:
#squashongreen

src/lib/utils/placeholders.tsx
MounirDhahri: ah no 😄 this won't happen I understand, the only way it would occur I think is by enabling/disabling the feature flag during the placeholder animation. Still, I wanted to mention this to you in case you were not aware of it https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

PR #6172, opened 2022-02-07
feat(flags): adding Unleash

pvinis:

This PR resolves [MOPLAT-126]

### Description

oh god this PR! i would appreciate some reviews here.

Adding Unleash, and renaming Split stuff for now, before we remove later.

@abhitip and audrey maybe can take a quick look at the tracking of this.

- [x] ~I have added a feature flag, or my changes don't require a feature flag. ([How do I add one?](https://github.com/artsy/eigen/blob/main/docs/developing_a_feature.md))~ My changes *are* the feature flag :p

### To the reviewers 👀

- [x] I would like at least one of the reviewers to run this PR on the simulator or device, so they can see how unleash works with the dashboard.

- adding Unleash for experiments and flags - pavlos

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[MOPLAT-126]: https://artsyproduct.atlassian.net/browse/MOPLAT-126?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.editorconfig
```diff
@@ -1,3 +1,5 @@
+insert_final_newline = true
```
pvinis, line 1: noticed many files dont end it a newline, so im adding this here.

.env.example
```diff
@@ -1,18 +1,41 @@
+ARTSY_API_CLIENT_KEY=-
```
pvinis, line 1: finally update the .env.example file to the layout of .env.shared.

Podfile
```diff
@@ -86,7 +86,7 @@ target 'Artsy' do
   pod 'NPKeyboardLayoutGuide'
   pod 'UICKeyChainStore'
   pod 'MARKRangeSlider'
-  pod 'EDColor'
+  pod 'EDColor', '1.0.1'
```
pvinis, line 89: same as with package.json file, lets add some specific versions. makes our lock file more.. reproducible.

scripts/update-env-for-artsy
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+
+
+aws s3 cp .env.shared s3://artsy-citadel/dev/.env.eigen
```
pvinis, line 5: yey new easy script to update the .env.shared. dont forget to
- get the latest first
- make your changes
- update using this script
- update 1password (and circle too if needed)

src/lib/AppProviders.tsx
```diff
@@ -10,36 +11,39 @@ import { RetryErrorBoundary } from "./Components/RetryErrorBoundary"
...
-                      </_FancyModalPageWrapper>
-                    </PopoverMessageProvider>
-                  </ActionSheetProvider>
-                </RetryErrorBoundary>
-              </Theme>
-            </RelayEnvironmentProvider>
-          </GlobalStoreProvider>
-        </ProvideScreenDimensions>
-      </SafeAreaProvider>
+      <FlagProvider config={unleashConfig}>
```
pvinis, line 22: thats the addition here.

src/lib/Components/StickyHeaderPage.tsx
```diff
@@ -1,5 +1,5 @@
 import { useAnimatedValue } from "lib/Components/StickyTabPage/reanimatedHelpers"
-import { useUpdadeShouldHideBackButton } from "lib/utils/hideBackButtonOnScroll"
+import { useUpdateShouldHideBackButton } from "lib/utils/hideBackButtonOnScroll"
```
pvinis, line 2: just a typo

src/lib/store/AuthModel.ts
```diff
@@ -148,9 +148,7 @@ export interface AuthModel {
   >
   signOut: Thunk<this>
 
-  notifyTracking: Thunk<this, { userId: string | null }>
```
pvinis, line 151: all this is replaced by `useIdentifyUser`

src/lib/utils/useDebugging.ts
```diff
@@ -0,0 +1,9 @@
+import AsyncStorage from "@react-native-community/async-storage"
+import { useEffect } from "react"
+import RNAsyncStorageFlipper from "rn-async-storage-flipper"
+
+export const useDebugging = () => {
+  useEffect(() => {
+    RNAsyncStorageFlipper(AsyncStorage)
```
pvinis, line 7: oooooo faaaaaancy! async storage in our flipper yall

src/lib/utils/useIdentifyUser.ts
```diff
@@ -0,0 +1,16 @@
+import { GlobalStore } from "lib/store/GlobalStore"
+import { useEffect } from "react"
+import { SegmentTrackingProvider } from "./track/SegmentTrackingProvider"
+import { useUpdateExperimentsContext } from "./useExperiments"
+
+export const useIdentifyUser = () => {
```
pvinis, line 6: look at this nice little hook, that replaces some of the redux state! when the userID changes in the GlobalStore, we dont need thunks etc. with this hook it will just run the segment identify call, plus it will add the userId to the experiments (using unleash) context.

typings/react-native-config.d.ts
```diff
@@ -2,18 +2,21 @@ import type { NativeConfig } from "react-native-config"
 
 declare module "react-native-config" {
   interface NativeConfig {
-    ARTSY_API_CLIENT_SECRET: string
     ARTSY_API_CLIENT_KEY: string
+    ARTSY_API_CLIENT_SECRET: string
```
pvinis, line 6: some sorting and cleaning, basically mirroring the .env.example

src/setupJest.ts
```diff
@@ -539,7 +539,6 @@ jest.mock("react-native-config", () => ({
   SENTRY_DSN: "sentry_dsn",
   GOOGLE_MAPS_API_KEY: "google_maps_api_key",
   MAPBOX_API_CLIENT_KEY: "mapbox_api_client_key",
-  SAILTHRU_KEY: "sailthru_key",
```
pvinis, line 542: 👋

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 79693cf39f81c38b2e81736689d33deed2c9087a
</p>

src/app/AndroidApp.tsx
```diff
@@ -54,7 +56,7 @@ const Main: React.FC = () => {
   useWebViewCookies()
   useDeepLinks()
   useInitialNotification()
-  useExperiments()
+  useSplitExperiments()
```
pvinis, line 59: renaming while we still have it here.

src/app/utils/experiments/helpers.ts
```diff
@@ -0,0 +1,16 @@
...
+  const client = getUnleashClient()
+  return client.updateContext(newContext)
+}
+
+// this might get called multiple times in a row because of ios having many providers but one client.
+// we will make sure to throttle it.
+export const forceFetchToggles = throttle(unthrottledForceFetch, 2000, { trailing: false })
+function unthrottledForceFetch(unleashEnv: "production" | "staging") {
+  const client = getUnleashClient(unleashEnv)
+  client.start()
```
pvinis, line 15: this is where we force the refetch.

src/app/utils/experiments/helpers.ts
```diff
@@ -0,0 +1,16 @@
...
+import { getUnleashClient } from "./unleashClient"
+
+export function updateExperimentsContext(newContext: IMutableContext) {
+  const client = getUnleashClient()
+  return client.updateContext(newContext)
+}
+
+// this might get called multiple times in a row because of ios having many providers but one client.
+// we will make sure to throttle it.
+export const forceFetchToggles = throttle(unthrottledForceFetch, 2000, { trailing: false })
```
pvinis, line 12: throttling is needed, because on "appForeground", on ios, all providers of all active screens will fire this, so we need to make sure we only do this once. throttle helps with this.

src/app/utils/experiments/UnleashProvider.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+export function UnleashProvider({ children }: { children?: ReactNode }) {
+  const [lastUpdate, setLastUpdate] = useState<UnleashContext["lastUpdate"]>(null)
+  const unleashEnv = useDevToggle("DTUseProductionUnleash") || !__DEV__ ? "production" : "staging"
+
+  useEffect(() => {
+    const client = getUnleashClient(unleashEnv)
+
+    // tslint:disable-next-line: no-empty
+    client.on("initialized", () => {})
```
pvinis, line 21: these are kinda placeholders so we know what we can do. the api is kinda silly on unleash.

src/app/utils/experiments/UnleashProvider.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+    // tslint:disable-next-line: no-empty
+    client.on("ready", () => {})
+
+    client.on("update", () => {
+      setLastUpdate(new Date())
```
pvinis, line 27: update is the important one. we only change the `lastUpdate` and based on that we can do our other updates/checks for flags.

src/app/store/config/features.ts
```diff
@@ -251,6 +251,9 @@ export const devToggles = defineDevToggles({
   DTCaptureExceptionsInSentryOnDev: {
     description: "Enable capturing exceptions in Sentry on DEV",
   },
+  DTUseProductionUnleash: {
```
pvinis, line 254: useful when we want to check stuff while developing that are deployed in production. should not be a regular need, but good to have.

normally, while developing eigen will connect to the dev env of unleash, and a regular build will go to prod

src/app/Scenes/Search/Search.tsx
```diff
@@ -112,6 +117,16 @@ export const Search: React.FC = () => {
...
 
+  const smudgeValue = useExperimentVariant("test-search-smudge")
+  nonCohesionTracks.experimentVariant(
+    "test-search-smudge",
+    smudgeValue.enabled,
+    smudgeValue.variant,
+    smudgeValue.payload
+  )
+  const smudge2Value = useExperimentFlag("test-eigen-smudge2")
+  nonCohesionTracks.experimentFlag("test-eigen-smudge2", smudge2Value)
```
pvinis, line 128: i left these in, but made them appear only on `__DEV__`, so we can remove them at any point, and we have an example for now.

also, the tracking happens here.

src/app/utils/experiments/reporter.ts
```diff
@@ -0,0 +1,86 @@
+import { ContextModule, OwnerType } from "@artsy/cohesion"
+import { postEventToProviders } from "../track/providers"
+
+const reportedExperimentFlags: Record<string, boolean> = {}
+export function maybeReportExperimentFlag({
```
pvinis, line 5: its called `maybe*` because it will not always report to segment. it will only report the first time we see an experiment name, and any time there is a change in the `enabled` value of that experiment, and all that is per session.

so next time we open the app, it will report again one time per name per changed value.

src/app/utils/experiments/reporter.ts
```diff
@@ -0,0 +1,86 @@
...
+    context_page_owner_type?: string
+  }) => ({
+    action: "experiment_viewed",
+    nonInteraction: 1,
+    ...rest,
+    experiment_id: name,
+    experiment_name: name,
+    variation_id: enabled ? variant : false,
+    variation_name: enabled ? variant : false,
+    // payload,
```
pvinis, line 84: this is not determined how or if we will send it, so i left it here until we know.

src/app/utils/experiments/hooks.ts
```diff
@@ -0,0 +1,37 @@
+import { useContext, useEffect, useState } from "react"
+import { getUnleashClient } from "./unleashClient"
+import { UnleashContext } from "./UnleashProvider"
+
+export function useExperimentFlag(name: string) {
```
pvinis, line 5: pretty simple hooks for flag and variant over here.

src/app/utils/experiments/unleashClient.ts
```diff
@@ -0,0 +1,66 @@
...
+import { EXPERIMENT_NAME, experiments } from "../experiments"
+
+/**
+ * This will return an unleash client
+ * If `env` is undefined, it will use whatever enviroment we are currently using,
+ * production by default.
+ * If a specific env is asked, then it will either use the current one if it's right,
+ * or create a new one in the requested env.
+ */
+export function getUnleashClient(env?: "production" | "staging") {
```
pvinis, line 13: setup happens here. we make the client once, and only get it after that. basically a singleton. we only recreate it if the env changes, and thats what this `envBeingUsed` does at the bottom.

src/app/utils/experiments/unleashClient.ts
```diff
@@ -0,0 +1,66 @@
...
+    bootstrap: (Object.keys(experiments) as EXPERIMENT_NAME[]).map((key) => ({
+      name: key,
+      enabled: experiments[key].fallbackEnabled,
+      variant: {
+        enabled: experiments[key].fallbackVariant ?? false,
+        name: experiments[key].fallbackVariant ?? "disabled",
+        payload:
+          experiments[key].fallbackPayload === undefined
+            ? undefined
+            : { type: "string", value: String(experiments[key].fallbackPayload) },
```
pvinis, line 59: unleash has the weirdest and worst types and requirements lol. its sooo inconsistent in the docs and even in their client libs. smh

src/app/utils/useDebugging.ts
```diff
@@ -0,0 +1,9 @@
+import AsyncStorage from "@react-native-community/async-storage"
+import { useEffect } from "react"
+import RNAsyncStorageFlipper from "rn-async-storage-flipper"
+
+export function useDebugging() {
+  useEffect(() => {
+    RNAsyncStorageFlipper(AsyncStorage)
```
pvinis, line 7: a nice little hook to put all our debugging initialization, for now just the asyncstorage flipper integration.

src/app/utils/nullAndUndef.ts
```diff
@@ -0,0 +1,7 @@
+export function nullToUndef<T>(value: T | null): T | undefined {
```
pvinis, line 1: these are tiny typescript helpers. many times ive needed that, and not only me, i remember @patrinoua most recently and others in the past needing something like that.

its used like this:
```
let userId: string | undef
const someOtherUserIdWithOtherTypes: string | null
// when doing the following, we get an error
// userId = someOtherUserIdWithOtherTypes
// but with the the helper we can do the following with no error
userId = nullToUndef(someOtherUserIdWithOtherTypes)
```
lovely

src/app/utils/useIdentifyUser.ts
```diff
@@ -13,5 +15,6 @@ export function useIdentifyUser() {
   useEffect(() => {
     Sentry.setUser({ id: userId ?? "none" })
     SegmentTrackingProvider.identify?.(userId, { is_temporary_user: userId === null ? 1 : 0 })
+    updateExperimentsContext({ userId: nullToUndef(userId) })
```
pvinis, line 18: heres one use of this. the unleash context expects `string | undefined` but we like `string | null`, so we use the helpers and the types and the values of all these are happy and handled.

docs/add_an_experiment.md
```diff
@@ -1,62 +1,77 @@
...
-1. Log into your Split.io using your personal Artsy account
-   `https://app.split.io/login`
-2. Choose the **Splits** tab from the left menu
-3. Tap on **Create split** and fill in required details as described in screenshot
-   ![create-split](./screenshots/create-split.png)
-4. Tap on **Add rules** and define your treatments.
-   A _treatment_ refers to possible values for your experiment. The values are of `string` type.
-   ![split-rules](./screenshots/split-rules.png)
-5. Tap on **Save** and confirm while using **Pre-production** as a Status.
+1. Log into Unleash using your personal Artsy account, then find the admin account in 1Password.
```
dblandin, line 7: I think there are two auth phases, right? Would something like the following be a bit clearer?

> To login to our Unleash instance, use your Artsy Google account to proceed past the Cloudflare Access prompt and then use the shared admin credentials in 1Password to login to the dashboard.

It looks like the open-source version of Unleash supports [multiple users](https://unleash.artsy.net/admin/users) - would we consider provisioning individual accounts? Maybe for more accurate event/audit logs?

Looks like Single Sign-On is only available in the Enterprise version of Unleash but maybe something to consider if the trial is successful.

package.json
```diff
@@ -186,6 +187,7 @@
     "styled-system": "5.1.5",
     "supercluster": "6.0.2",
     "tipsi-stripe": "https://github.com/artsy/tipsi-stripe.git#fix-infinite-loop",
+    "unleash-proxy-client": "1.11.0",
```
dblandin, line 190: **question** - Curious if there's some documentation around the decision to use the [unleash-proxy-client](https://github.com/Unleash/unleash-proxy-client-js) package directly vs the higher-level [proxy-client-react](https://github.com/Unleash/proxy-client-react) package. Did we try that out and find that it wasn't configurable enough?

Looks like it defines some similar providers and hooks that we're including ourselves in this PR.

src/app/utils/experiments/unleashClient.ts
```diff
@@ -0,0 +1,66 @@
...
+    url:
+      envBeingUsed === "production"
+        ? Config.UNLEASH_PROXY_URL_PRODUCTION
+        : Config.UNLEASH_PROXY_URL_STAGING,
+    clientKey:
+      envBeingUsed === "production"
+        ? Config.UNLEASH_PROXY_CLIENT_KEY_PRODUCTION
+        : Config.UNLEASH_PROXY_CLIENT_KEY_STAGING,
+    appName: "eigen",
+    refreshInterval: 0, // don't refresh. we will manually refresh on appStart and goingForeground.
```
dblandin, line 41: **suggestion**: There's also a `disableRefresh` configuration variable which we can use instead ([docs](https://github.com/Unleash/unleash-proxy-client-js#available-options)).

dblandin (commented):
Exciting to see this take form!

scripts/update-env-for-artsy
brainbicycle: totally not blocking but an idea for future, with our current set up it would be pretty easy to overwrite the .env file in s3 with bad values. it would be nice if this script had a confirmation step, something along the lines of:

- fetch the latest .env file in s3 
- compare it to local env file
- present the delta to the user
- ask user to confirm they want to overwrite these values

docs/add_an_experiment.md
pvinis: thats a better phrase, ill fix this.

for multiple users, i guess we could? @joeyAghion is probably the one to ask, and then we can change the docs too.

package.json
pvinis: we did try it out, and for ios it would not work. their react lib makes a lot of assumptions that eigen on ios is not, and it was causing issues with syncing and fetching a lot etc.

Do you think it makes sense to add a HACKS entry for this? I was thinking of making a blogpost, a release of our own proxy-client-react with fewer assumptions would be cool, and maybe after the appshell work we could try their one, but i wouldn't prefer it to ours, since ours is superior.

src/app/utils/experiments/unleashClient.ts
pvinis: i kinda prefer the `refreshInterval` for this, making it easy to change it to any number if we want in the future, rather than having to find the right prop again. seems like `refreshInterval` has the functionality of both controlling if it will fetch, and when, so i think its fine to keep as it. does that make sense?

scripts/update-env-for-artsy
pvinis: thats a good idea! i actually move this file away from this PR, but i guess this is still left here.

the way i thought ill do a quick "pay attention" is that i named the script with a bang here https://github.com/artsy/eigen/blob/main/package.json#L49, kinda ruby style "dont just call thing, only call this with purpose".

but i really like the idea of a confirmation. ill make a ticket, so we dont forget, but its separate from this PR here.

docs/add_an_experiment.md
```diff
@@ -1,62 +1,76 @@
...
 ## Using an experiment
 
-In order to use an experiment, we have two custom hooks that we created that support querying for a single experiment (`useTreatment`).
+In order to use an experiment, we have two custom hooks that we created that support querying for a flag (`useExperimentFlag`) or an experiment (`useExperimentVariant`). The first one returns a boolean, the second returns a small object for variants and their payloads etc.
 
-### Querying for a single treatment
+### Querying for a single flag
 
-You can access the treatment value in a functional react component using `useTreatment`.
+You can access the flag value in a functional react component using `useExperiementFlag`.
```
brainbicycle, line 42:
```suggestion
You can access the flag value in a functional react component using `useExperimentFlag`.
```
minor typo

docs/add_an_experiment.md
```diff
@@ -1,62 +1,76 @@
...
 
 ```diff
-+ const treatment = useTreatment("HomeScreenWorksForYouVsWorksByArtistsYouFollow")
++ const ourNewFlagEnabled = useExperiementFlag("our-new-flag")
```
brainbicycle, line 45:
```suggestion
+ const ourNewFlagEnabled = useExperimentFlag("our-new-flag")
```

docs/add_an_experiment.md
```diff
@@ -1,62 +1,76 @@
...
 ```
 
-## Removing/Killing an Experiment
+### Querying for a single experiment
 
-Once an experiment is done, usually we have a winner treatment. In order to roll out that treatment for everyone targeted by it, we will need to set it as a default treatment before "killing" it.
+You can access the variant value in a functional react component using `useExperimentVariant`.
+
+```diff
++ const ourNewExperiment = useExperiementFlag("our-new-experiment")
```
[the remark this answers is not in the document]
pvinis: oh yea, good catch

src/app/Scenes/Search/Search.tsx
brainbicycle: maybe we can put behind a dev toggle like `DTShowExampleExperiment` that way devs aren't confused when they come across it?

src/app/utils/experiments/unleashClient.ts
```diff
@@ -0,0 +1,66 @@
+import AsyncStorage from "@react-native-community/async-storage"
+import { Config } from "react-native-config"
+import { UnleashClient } from "unleash-proxy-client"
+import { EXPERIMENT_NAME, experiments } from "../experiments"
+
+/**
+ * This will return an unleash client
+ * If `env` is undefined, it will use whatever enviroment we are currently using,
```
brainbicycle, line 8:
```suggestion
 * If `env` is undefined, it will use whatever environment we are currently using,
```
minor type :)

src/app/utils/nullAndUndef.ts
brainbicycle: ooh I have needed this a lot

scripts/update-env-for-artsy
brainbicycle: yeah definitely separate, thanks for the clean up in the first place!

brainbicycle (dismissed):
WOO! looks great to me, some minor typos and a thought about putting the example experiment behind a dev toggle, but nothing blocking looks great! Thanks for making the API simple!

src/app/utils/experiments/unleashClient.ts
pvinis: oh god so many! i wonder how vscode did catch them. or maybe i didnt see it.. thanks brian.

src/app/Scenes/Search/Search.tsx
pvinis: hm. sure? i dont mind either way. im ok removing them, keeping as is, or hidding behind a flag. i wouldnt make it a devtoggle, maybe just a regular flag that can go away soon?

docs/add_an_experiment.md
joeyAghion: We should use individual user authorizations rather than shared accounts wherever possible, including here. (Cloudflare "access" plus the shared account was enough to get started but that's all.)

We can improve this later, or if you want to anticipate that improvement, you could instruct folks to "ask someone with access to [add you as a user](https://unleash.artsy.net/admin/create-user)."

package.json
dblandin: If the goal is an improved React library for Unleash, I feel like we should at least try to contribute these improvements/fixes upstream. I'd be hesitant to open-source an alternative without at least trying that.

What were some of the issues with the iOS integration? Why is our implementation better? Can we contribute fixes to the issues or at use [patch-package](https://github.com/ds300/patch-package) to get around them in the short-term?

If we do end up going the patch-package route, a HACKS entry makes sense. If we stick with our own hooks / provider / etc, it would be nice to document the specific reasons why the existing package didn't work for us. Either way, thinking of improving the existing open-source library seems preferable to reimplementing something similar with less support from Unleash and less usage in the community.

src/app/Scenes/Search/Search.tsx
brainbicycle: Reason I am thinking is I think having UI that shows in DEV that isn't in a prod is a bid jarring, ideally what I am looking at while developing is as close as possible to what is in prod. Like if I am a dev and I know nothing about unleash and I come across this colored squares in search my first thought is "What is this? is this a bug?"  a regular flag that can go away soon works as well

src/app/utils/experiments/unleashClient.ts
dblandin: Seems reasonable! One compromise might be to include both but comment out refreshInterval:

```typescript
disableRefresh: true, // explicitly refreshed on appStart and goingForeground
// refreshInterval: 15
```

package.json
pvinis: that makes sense. i wish i had more time to add tests that fail and then fix them and make a PR to them. ill put a reminder for next friday and maybe ill do that. 🤔

src/app/utils/experiments/unleashClient.ts
pvinis: i think i prefer the no comment way for the extra thing. but thank you for suggesting it! a very respectful decline 🙇

src/app/utils/experiments/unleashClient.ts
```diff
@@ -0,0 +1,66 @@
...
+}
+
+let unleashClient: UnleashClient | null = null
+
+const storageName = (name: string) => `unleash-values:${name}`
+
+const createUnleashClient = () => {
+  console.debug("Unleash creating client", envBeingUsed)
+
+  return new UnleashClient({
```
MounirDhahri, line 31: @pvinis would it make sense to probably wrap this with a try catch and send errors to sentry in case something goes wrong with the unleash SDK initialization

dblandin (commented):
**meta**: Some general review feedback - it was bit tricky to focus on the Unleash-specific changes here. It would've been a clearer to review had these other changes been introduced in separate PRs:

- Add rn-async-storage-flipper / Improve debugging / jest setup adjustment
- Format files with prettier

pvinis:
I know.. sorry for that. It was a chaotic PR and I could have done a better job separating. Most things came up from the trials and investigations with Unleash. I appreciate the feedback though, I'll definitely try to do these more cleanly 💜

src/app/utils/experiments/reporter.ts
```diff
@@ -0,0 +1,86 @@
...
+  if (reportedExperimentVariants[name] === combinedValue) {
+    return
+  }
+
+  reportedExperimentVariants[name] = combinedValue
+  postEventToProviders(tracks.experimentVariant({ name, enabled, variant, payload, ...rest }))
+}
+
+// this is not on cohesion yet, so the schema might change
+const tracks = {
```
MounirDhahri, line 36: thanks for the comment

MounirDhahri (approved):
Great work Pavlos 👏👏👏   I left 2 minor comments but they're definitely non blocking

src/app/utils/experiments/unleashClient.ts
```diff
@@ -0,0 +1,66 @@
...
+        const data = await AsyncStorage.getItem(storageName(name))
+        return data ? JSON.parse(data) : undefined
+      },
+    },
+    bootstrapOverride: false,
+    bootstrap: (Object.keys(experiments) as EXPERIMENT_NAME[]).map((key) => ({
+      name: key,
+      enabled: experiments[key].fallbackEnabled,
+      variant: {
+        enabled: experiments[key].fallbackVariant ?? false,
```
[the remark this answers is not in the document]
pvinis: i didnt go for this because i wanted it to be more obvious what the "default" expectation from unleash is. they dont expect `undefined`, they expect `false` specifically, so i thought ill do that? 🤔 but maybe with `!!` is also good?

src/app/utils/experiments/unleashClient.ts
pvinis: oo good idea. in my trials (with a lot of config trials), it only errored out when the client key or the url where missing or did not include an `https`. but i will add this. that makes sense.

pvinis: ok scratch that. im trying it now and i see that actually the config init is totally ok with whatever values. `start()` is also ok. when it errors out is when it tried to fetch, where it will give `TypeError: Network request failed`, which I think we catch already with Sentry without any extra setup, right? So I think we are good right here.

brainbicycle (approved):

src/app/utils/experiments/reporter.ts
```diff
@@ -0,0 +1,86 @@
...
+  const combinedValue = `${enabled}-${variant}-${payload}`
+  if (reportedExperimentVariants[name] === combinedValue) {
+    return
+  }
+
+  reportedExperimentVariants[name] = combinedValue
+  postEventToProviders(tracks.experimentVariant({ name, enabled, variant, payload, ...rest }))
+}
+
+// this is not on cohesion yet, so the schema might change
```
mc-jones, line 35: Will cohesion know the source of the event? I imagine I will be using the same schema for Force (but let me know if this is an incorrect assumption, as I haven't worked with our analytics yet) and we need to be able to differentiate.

brainbicycle:
Since I don't see any blocking comments I am going to go ahead and merge but please raise anything you think needs addressing and we can do follow-ups!

src/app/utils/experiments/unleashClient.ts
MounirDhahri: If the error we catch is `TypeError: Network request failed` I assume it's fine to ignore this. I assumed though that they will be returning a more useful error for us to act on that we can explicitly capture using Sentry within its scope

src/app/utils/experiments/unleashClient.ts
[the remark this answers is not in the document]
MounirDhahri: sounds good to me

src/app/utils/experiments/reporter.ts
pvinis: i think we will use the same schema, so it might be worth working together to add it to cohesion. wanna pair on this when you are at that point with force? 🤔 

it should know the source. `rest` is all these `context_page_owner_id` things

PR #6175, opened 2022-02-07
fix: broken artist screen animation

MounirDhahri:

This PR resolves issue with the Artist Screen failing to load and we get the error below
<details>
<summary>Screenshot</summary>

Screenshot_1644237860

</details>

https://user-images.githubusercontent.com/11945712/152802627-2d7f4094-228e-41d4-a290-e77c50b44f7f.mov

### Description

Reanimated failed on Android to animate a value from 0 to a duration set to 0. It's still not yet clear to me why this is the case only on Android, but I think this is not something I should be spending time investigating for now since we can simply remove the animation because we are not animating the header anymore

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- refactor disabling the animation approach - mounir

</details>

### DIFF
--- a/src/lib/utils/useAutoCollapsingMeasuredView.tsx
+++ b/src/lib/utils/useAutoCollapsingMeasuredView.tsx
@@ -1,12 +1,11 @@
-import React, { useRef, useState } from "react"
+import React, { useState } from "react"
 import { View } from "react-native"
-import Animated, { Easing } from "react-native-reanimated"
+import Animated from "react-native-reanimated"
 
 export function useAutoCollapsingMeasuredView(content: React.ReactChild | null) {
   const [nativeHeight, setNativeHeight] = useState<Animated.Value<number>>(
     __TEST__ ? new Animated.Value(100) : new Animated.Value(-1)
   )
-  const animation = useRef<Animated.BackwardCompatibleWrapper | null>(null)
 
   return {
     nativeHeight,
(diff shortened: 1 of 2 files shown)


ArtsyOpenSource:

- Dev changes (refactor disabling the animation approach - mounir)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 379f8d3518c8f33e676a801d94eefdfff9ffed3a
</p>

olerichter00 (changes requested):
Thanks for the fix :) 

The problem is that this fix is not collapsing the header when scrolling up (you can reproduce it on the artist screen). 

Not sure what the most elegant solution would be, but we could just set the duration to `1` to fix the issue 🤔

src/lib/utils/useAutoCollapsingMeasuredView.tsx
```diff
@@ -27,18 +26,7 @@ export function useAutoCollapsingMeasuredView(content: React.ReactChild | null)
...
-              }
-              Animated.timing(nativeHeight, {
-                toValue: e.nativeEvent.layout.height,
-                duration: 0,
-                easing: Easing.linear,
-              }).start()
-            } else {
-              setNativeHeight(new Animated.Value(e.nativeEvent.layout.height))
-            }
+            setNativeHeight(new Animated.Value(e.nativeEvent.layout.height))
```
olerichter00, line 29: This seems to work: 

```suggestion
            nativeHeight.setValue(e.nativeEvent.layout.height)
```

olerichter00 (changes requested):

olerichter00:
#squashongreen

PR #6176, opened 2022-02-07
[feat]: AlicanAkyuz/Submission Flow Photo Upload

AlicanAkyuz:

This PR resolves [SWA-86]

### Description
This PR rebuilds the photo-upload step of the artwork submission flow. 

#### Video IOS and Convection
https://user-images.githubusercontent.com/42584148/153413225-5024b2bb-e5bb-4003-bffc-1d7d12fc8dc6.mov

### To the reviewers 👀

#### Dev changes
Since this feature is behind a feature flag, the changes apply to devs only. 

</details>

[SWA-86]: https://artsyproduct.atlassian.net/browse/SWA-86?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 5711f59d5362e889cd688f4453eb6b553ff0ac5b
</p>

src/lib/Scenes/Consignments/Screens/SubmitArtworkOverview/ArtworkDetails/Components/ArtistAutosuggest.tsx
```diff
@@ -52,7 +52,7 @@ export const ArtistAutosuggest: React.FC = () => {
       />
 
       {!!focused && !isArtistSelected && artist?.length > 2 && (
-        <Box height={198}>
+        <Box height={200}>
```
AlicanAkyuz, line 55: **note**: this is unrelated to the main task, just a small improvement to height after collapse animation is fixed.

src/lib/Scenes/Consignments/Screens/SubmitArtworkOverview/ArtworkDetails/Components/ArtistAutosuggestResults.tsx
```diff
@@ -264,7 +264,7 @@ const ArtistAutosuggestResultsPlaceholder: React.FC = () => {
     <Flex
       p={2}
       mb={1}
-      mt={1}
+      mt={0.5}
```
AlicanAkyuz, line 267: **note**: another small unrelated improvement to height after collapse improvement.

src/lib/Scenes/Consignments/Screens/SubmitArtworkOverview/ArtworkDetails/Components/LocationAutocomplete.tsx
```diff
@@ -125,7 +125,7 @@ export const LocationPredictions = ({
   }
 
   return (
-    <Box height={175}>
+    <Box height={175} mt={0.5}>
```
AlicanAkyuz, line 128: **note**: same here

src/lib/store/GlobalStoreModel.ts
```diff
@@ -2,7 +2,7 @@ import { Action, action, createStore, State, thunkOn, ThunkOn } from "easy-peasy
 import { LegacyNativeModules } from "lib/NativeModules/LegacyNativeModules"
 import { BottomTabsModel, getBottomTabsModel } from "lib/Scenes/BottomTabs/BottomTabsModel"
 import {
-  getSubmissionSubmissionModel,
+  getSubmissionModel,
```
AlicanAkyuz, line 5: **note**: a small typo fix unrelated to the main task.

src/lib/Scenes/Consignments/Screens/SubmitArtworkOverview/Mutations/removeAssetFromConsignmentSubmissionMutation.ts
```diff
@@ -0,0 +1,36 @@
...
+  removeAssetFromConsignmentSubmissionMutation,
+} from "__generated__/removeAssetFromConsignmentSubmissionMutation.graphql"
+import { defaultEnvironment } from "lib/relay/createEnvironment"
+import { commitMutation, graphql } from "relay-runtime"
+
+export const removeAssetFromSubmission = (input: RemoveAssetFromConsignmentSubmissionInput) => {
+  return new Promise<string>((resolve, reject) => {
+    commitMutation<removeAssetFromConsignmentSubmissionMutation>(defaultEnvironment, {
+      mutation: graphql`
+        mutation removeAssetFromConsignmentSubmissionMutation(
```
patrinoua, line 12: Just a wording related question here: Can an asset be something except a photo? If not, I'd rather have it _removePhotoFromConsignmentSubmissionMutation_ photo for consistency reasons.

iskounen: These are called "assets" in Convection, and I think that they are called "assets" instead of "photos" because there may have been plans to allow users/admins to upload documents such as certificates of authenticity.

Screen Shot 2022-02-11 at 7 11 52 AM

However, it doesn't seem like we ever pursued that so we can still only upload images. In this case, even if they upload a certificate of authenticity (or any other document) it would be a photo, so I am not against changing this.

src/lib/Scenes/Consignments/Screens/SubmitArtworkOverview/UploadPhotos/PhotoRow.tests.tsx
```diff
@@ -0,0 +1,113 @@
...
+import { createMockEnvironment } from "relay-test-utils"
+import { PhotoRow } from "./PhotoRow"
+import { Photo } from "./validation"
+
+jest.unmock("react-relay")
+
+const mockEnvironment = defaultEnvironment as ReturnType<typeof createMockEnvironment>
+const mockHandlePhotoDelete = jest.fn()
+
+describe("PhotoRow", () => {
```
iskounen, line 16: **praise:** These tests are clear and helped me understand the component well ❤️

src/lib/Scenes/Consignments/Screens/SubmitArtworkOverview/UploadPhotos/validation.ts
```diff
@@ -0,0 +1,49 @@
...
+        height: Yup.string(),
+        isDefault: Yup.string(),
+        imageURL: Yup.string(),
+        internalID: Yup.string(),
+        path: Yup.string(),
+        width: Yup.number(),
+        imageVersions: Yup.string(),
+        loading: Yup.boolean(),
+        error: Yup.boolean(),
+        errorMsg: Yup.string(),
```
iskounen, line 44: **question:** Why do we need to validate the error message (or size display value) on a photo?

I am confused, because my understanding of Yup validation is that you use it to validate form values before sending them to the backend. Things like the error message and size display are purely for the user's benefit, so I don't understand why we would validate them.

src/lib/Scenes/Consignments/Screens/SubmitArtworkOverview/UploadPhotos/utils/calculatePhotoSize.ts
```diff
@@ -0,0 +1,18 @@
...
+import { Photo } from "../validation"
+
+export const calculatePhotoSize = (photo: Photo): Photo => {
+  if (!photo.size) {
+    photo.error = true
+    photo.sizeDisplayValue = "Size not found"
+    return photo
+  }
+
+  if (photo.size > 30000000) {
```
iskounen, line 11: **issue:** On web, the criteria for this error was that the **sum** of all photos couldn't exceed 30 MB. Is that logic different for the app?

src/lib/Scenes/Consignments/Screens/SubmitArtworkOverview/UploadPhotos/validation.ts
```diff
@@ -0,0 +1,49 @@
...
+  height?: number
+  isDefault?: boolean
+  imageURL?: string
+  internalID?: string
+  path?: string
+  width?: number
+  imageVersions?: string[]
+  loading?: boolean
+  error?: boolean
+  errorMsg?: string
```
iskounen, line 19: **nitpick:** Is it a convention in Eigen to shorten "message" to "Msg"? If not, it would be nice to spell it out since all the other props have full names.

iskounen (commented):
I have one issue with the logic to determine if a user has reached the 30 MB limit, but I want to double-check that with you in case we decided to change it for the app.

Otherwise I have minor comments. Thank you for breaking out logic into utils - that makes it a lot easier for me to review React code!

src/lib/Scenes/Consignments/Screens/SubmitArtworkOverview/Mutations/removeAssetFromConsignmentSubmissionMutation.ts
AlicanAkyuz: note also that if we changed the name of this mutation, we'd also need to alter it in MP schema, sync MP schema here,   and repeat the same for all of its consumers (in Force for sure, I'm not sure if it's consumed elsewhere).

I agree that this can be considered later outside of the scope here.

src/lib/Scenes/Consignments/Screens/SubmitArtworkOverview/UploadPhotos/utils/calculatePhotoSize.ts
AlicanAkyuz: thank you for catching this. @kiryl-zubarau also warned me – I'm about to change the logic so that 30MB limit is applied to the total of photos.

src/lib/Scenes/Consignments/Screens/SubmitArtworkOverview/UploadPhotos/validation.ts
AlicanAkyuz: makes sense, I'll update the naming.

src/lib/Scenes/Consignments/Screens/SubmitArtworkOverview/UploadPhotos/validation.ts
AlicanAkyuz: I will remove the unnecessary fields from Yup shape. 

**Context**: Yup will validate only the specified types, and fields that are marked with `required()`. In our case, it is only the `id` and `geminiToken` that we validate via Yup before enabling the Save & Continue button. I wanted to have Yup object's shape to reflect all fields of Photo type. but you are right in that we don't need most of those fields in Yup and I guess having a Photo interface above it is clear enough.

iskounen: Ah! So without `required()` the Yup schema will act like a type check. That makes total sense, and I was forgetting about that use case when I was reviewing your code. Thanks for the lesson!

src/lib/Scenes/Consignments/Screens/SubmitArtworkOverview/UploadPhotos/utils/calculatePhotoSize.tests.ts
```diff
@@ -0,0 +1,15 @@
+import { calculateSinglePhotoSize } from "./calculatePhotoSize"
+
+describe("calculatePhotoSize", () => {
```
iskounen, line 3: **suggestion (non-blocking):** It would also be helpful to test `isSizeLimitExceeded`

src/lib/Scenes/Consignments/Screens/SubmitArtworkOverview/UploadPhotos/UploadPhotos.tsx
```diff
@@ -1,12 +1,58 @@
...
+      >
+        {({ values, isValid }) => {
+          const isAnyPhotoLoading = values.photos.some((photo: Photo) => photo.loading)
+
+          return (
+            <>
+              <UploadPhotosForm />
+              <Spacer mt={2} />
+              <CTAButton
+                disabled={!isValid || isAnyPhotoLoading || isSizeLimitExceeded(values.photos)}
```
iskounen, line 46: **suggestion (non-blocking):** If you want to keep all validation logic in one place, I think there is a way to call `isSizeLimitExceeded` from the Yup schema. (I am marking this non-blocking because it takes a while to figure out the syntax.)

iskounen (approved):
I have two non-blocking suggestions, but I think this is ready to go!

src/lib/Scenes/Consignments/Screens/SubmitArtworkOverview/UploadPhotos/UploadPhotos.tsx
AlicanAkyuz: thank you! this is good to know, I was unaware of this capability. I'm looking into it in case it can be done swiftly.

patrinoua:
The PR looks good. 

Here are a few potential improvements that we could discuss with the rest of the team before implementing.

Photo Placeholder format: 
- Show upload status
- Display error message if the photo upload is unsuccessful 
- Add an X symbol on top right so that user can stop the upload in case their changed their mind (eg by accident uploaded the wrong file, upload takes too long)

- Sometimes a Photo upload takes a longer time: Maybe add a spinner "This takes longer than expected, please hold on..." 
- When you close the app does the submission stop? If so we should ask them not to close the app while their data is being processed.
- Display an error/info message if the user tries to upload an unsupported format
- What happens if the user tries to upload one photo and then directly another one? If this is not supported, Add Photo could be disabled.
- Upload takes a long time (at least for me)
- Add info message: "Please wait while we upload your photo"
- Maybe we can display the size of the photo that's being uploaded (so that the user knows it's a big file for example and can expect it to last longer) 

Other comments:
- On Artwork Details screen, maybe "Rarity" could be a dropdown instead of a modal
- Location autosuggest looks a little weird (bigger letters)

patrinoua:
Personally I'd be fine with merging this PR and opening another PR with the suggested enhancements if we decide it's something we want to implement. 

Also since this is not a trivial ticket I would also consider ask for a review from Mounir, Pavlos or Brian.

PR #6181, opened 2022-02-08
feat [SWA-205] CollapsibleView component

patrinoua:

[SWA-205]

A replacement of Collapse component, using a View instead of Spring.
 
This resolves the bug of Collapse component not appearing on Android and sometimes on iOS due to height calculation issues.

- [x] Has animations

To quickly test the behaviour you can go to ArtworkDetails component and
1) comment out line 55 (disabled under CTAButton)
```
   <CTAButton
              // disabled={!isValid}
              onPress={() => handleArtworkDetailsSubmit(values)}
              testID="Submission_ArtworkDetails_Button"
            >
              Save & Continue
            </CTAButton>
```

2) replace handleArtworkDetailsSubmit (line 17) with 

```
  const handleArtworkDetailsSubmit = async (values: ArtworkDetailsFormModel) => {
    handlePress()
    // try {
    //   const id = await createOrUpdateSubmission(values, submissionId)
    //   if (id) {
    //     GlobalStore.actions.artworkSubmission.submission.setSubmissionId(id)
    //     GlobalStore.actions.artworkSubmission.submission.setArtworkDetailsForm(values)
    //     handlePress()
    //   }
    // } catch (error) {
    //   captureMessage(JSON.stringify(error))
    //   setSubmissionError(true)
    // }
  }
```
 
ANDROID
https://user-images.githubusercontent.com/36475005/152972476-007431b6-c1e7-4341-8f38-f02e5e49b4f4.mov

IOS

https://user-images.githubusercontent.com/36475005/152972692-ae90677a-7a64-454b-955e-babdeebfc109.mov

This PR resolves []

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[SWA-205]: https://artsyproduct.atlassian.net/browse/SWA-205?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/Scenes/Consignments/Screens/SubmitArtworkOverview/ArtworkDetails/ArtworkDetailsForm.tsx
+++ b/src/lib/Scenes/Consignments/Screens/SubmitArtworkOverview/ArtworkDetails/ArtworkDetailsForm.tsx
@@ -10,17 +10,17 @@ import { rarityOptions } from "./utils/rarityOptions"
 import { limitedEditionValue } from "./utils/rarityOptions"
 import { ArtworkDetailsFormModel, Location } from "./validation"
 
+const StandardSpace = () => <Spacer mt={4} />
+
 export const ArtworkDetailsForm: React.FC = () => {
   const { values, setFieldValue } = useFormikContext<ArtworkDetailsFormModel>()
   const [isRarityInfoModalVisible, setIsRarityInfoModalVisible] = useState(false)
   const [isProvenanceInfoModalVisible, setIsProvenanceInfoModalVisible] = useState(false)
 
-  const StandarSpace = () => <Spacer mt={4} />
-
   return (
     <>
       <ArtistAutosuggest />
-      <StandarSpace />
+      <StandardSpace />
       <Input
         title="Title"
         placeholder="Add Title or Write 'Unknown'"
@@ -28,7 +28,7 @@ export const ArtworkDetailsForm: React.FC = () => {
         value={values.title}
         onChangeText={(e) => setFieldValue("title", e)}
       />
-      <StandarSpace />
+      <StandardSpace />
       <Input
         title="Year"
         placeholder="YYYY"
@@ -36,7 +36,7 @@ export const ArtworkDetailsForm: React.FC = () => {
         value={values.year}
         onChangeText={(e) => setFieldValue("year", e)}
       />
-      <StandarSpace />
+      <StandardSpace />
       <Input
         title="Materials"
         placeholder="Oil on Canvas, Mixed Media, Lithograph.."
@@ -44,7 +44,7 @@ export const ArtworkDetailsForm: React.FC = () => {
         value={values.medium}
         onChangeText={(e) => setFieldValue("medium", e)}
       />
-      <StandarSpace />
+      <StandardSpace />
       <Select
         onSelectValue={(e) => setFieldValue("attributionClass", e)}
         value={values.attributionClass}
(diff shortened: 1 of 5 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against e83f1d7a15b239d59c5eea4b1c8b4f268725b9cb
</p>

AlicanAkyuz (commented):
I think this is a nice working solution for our needs now 🚀  I have 2 suggestions below. after they are considered, let's ship this! 

**Suggestion 1**: since CollapseView is a one-liner and there's currently no collapse animation in it, I'd suggest putting `{isOpen ? <View>{children}</View> : null}` directly inside `CollapsibleMenuItem` instead of creating a new component for this. what do you think? 

**Suggestion 2**: since this PR resolves the height issue, now we need to remove the static `height: 1340` from ArtworkDetails component, and adjust the spacing between inputs as per designs in `ArtworkDetailsForm` component. I think it's a good idea to get them done in this PR. if you disagree, let me know so that I create a ticket for this.

AlicanAkyuz:
@patrinoua thank you for removing the height 🙏  are you also able to arrange the spacing between inputs in `ArtworkDetailsForm` component? or else I can tackle it quickly.

src/lib/Scenes/Consignments/Screens/SubmitArtworkOverview/ArtworkDetails/ArtworkDetailsForm.tsx
```diff
@@ -15,30 +15,36 @@ export const ArtworkDetailsForm: React.FC = () => {
   const [isRarityInfoModalVisible, setIsRarityInfoModalVisible] = useState(false)
   const [isProvenanceInfoModalVisible, setIsProvenanceInfoModalVisible] = useState(false)
 
+  const StandarSpace = () => <Spacer mt={4} />
```
AlicanAkyuz, line 18: Two things to note about this implementation: 

1- there is a small typo in the word _Standard_

2- if you want to have `<Spacer mt={4} />` centralised like this, we don't have any reason to set it to a function and have that function live inside `ArtworkDetailsForm` component. we probably don't want to:

- make several unnecessary function calls from within the return statement
- re-create this function every time the component re-renders 

my suggestion is to directly use `<Spacer mt={4} />` inside the return statement. if you prefer to stick with this centralised approach, my request is to set it to a constant and have that constant outside of `ArtworkDetailsForm` component.

src/palette/elements/Collapse/CollapseView.tsx
```diff
@@ -0,0 +1,11 @@
...
+import { View } from "react-native"
+
+interface Props {
+  isOpen: boolean
+  children: ReactNode
+}
+
+export const CollapseView: React.FC<Props> = ({ isOpen, children }) => {
+  return <>{isOpen ? <View>{children}</View> : null}</>
+}
```
AlicanAkyuz, line 11: **request**: if you find keeping this one-liner as a separate component more appropriate, I'd suggest refactoring it a little bit. 

1- you can drop `children` from the `Props` interface, as it'll always be available as a node. 
2- we can remove the fragment from the return statement, since it has only one child (View). 

it'd look like something like this: 
```
import React from "react"
import { View } from "react-native"

export const CollapseView: React.FC<{ isOpen: boolean }> = ({ isOpen, children }) =>
  isOpen ? <View>{children}</View> : null

```

AlicanAkyuz (approved):

PR #6243, opened 2022-02-16
Docs: Update preferred_practices.md

patrinoua:
The type of this PR is: **docs**

This PR updates the phrasing and removes some old parts of the documentation. 

preferred_practices.md is there to give us a broad overview of the history and how we prefere to do things.

@olerichter00 could you maybe contribute on the **relay hooks** part?

@gkartalis could you contribute on the **testing** part? 

We don't need to fix everything under this PR, but this could be a start to make sure we keep our current practices up to date.
 

### To the reviewers 👀

### Changelog updates

- Rename modal title - patrinoua

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/docs/preferred_practices.md
+++ b/docs/preferred_practices.md
@@ -31,7 +31,7 @@ The Artsy app was initially written in Objective-C and Swift and React Native wa
 
 - [Why Artsy uses React Native](http://artsy.github.io/blog/2016/08/15/React-Native-at-Artsy/)
 - [All React Native posts on Artsy's Engineering Blog](http://artsy.github.io/blog/categories/reactnative/)
-- Some great React Native components:
+- Sample React Native components:
   - [Partner](https://github.com/artsy/eigen/blob/main/src/app/Scenes/Partner/Partner.tsx) is a simple top-level component.
   - [PartnerShows](https://github.com/artsy/eigen/blob/main/src/app/Scenes/Partner/Components/PartnerShows.tsx) is a fragment container that uses FlatList to paginate through Relay data.
   - [Search](https://github.com/artsy/eigen/blob/main/src/app/Scenes/Search/Search.tsx) is a functional component that loads data in response to user input.
--- a/docs/preferred_practices.md
+++ b/docs/preferred_practices.md
@@ -16,6 +16,7 @@ Last update: February 2022
 - [Testing](#testing)
 - [Navigation](#Navigation)
 - [Analytics and tracking](#analytics-and-tracking)
+- [Formik](#formik)
 
 ### Examples & Hacks
 
@@ -148,3 +149,7 @@ See our documentation on adding a route for more details: [Adding a new route](h
 ### Analytics and tracking
 
 In React-native, we use react-tracking as a wrapper for the tracking events we send to Segment. You can read more about the implementation [here](./analytics_and_tracking.md).
+
+### Formik
+
+We use Formik for handling forms. You can see an example that's also using form validation [here](https://github.com/artsy/eigen/blob/main/src/app/Scenes/Consignments/Screens/SubmitArtworkOverview/ContactInformation/ContactInformation.tsx)
--- a/docs/preferred_practices.md
+++ b/docs/preferred_practices.md
@@ -139,18 +139,6 @@ Refactoring old containers to use hooks is encouraged.
 - We have native unit tests too. See [`getting_started.md`](./getting_started.md)
 - We don't like snapshot tests; they produce too much churn for too little value. It's okay to test that a component doesn't throw when rendered, but use [`extractText`](https://github.com/artsy/eigen/blob/4c7c9be69ab1c2095f4d2fed11a040b1bde6eba8/src/app/tests/extractText.ts) (or similar) to test the actual component tree.
 
-Here are some great examples of what tests and test coverage should look like.
-
-- [Tests for Gene component](https://github.com/artsy/emission/blob/751d24306a2d6ace58b21491e25b37f345c7a206/src/app/Containers/__tests__/Gene-tests.tsx)
-- [Tests for Consignments submission flow](https://github.com/artsy/emission/blob/751d24306a2d6ace58b21491e25b37f345c7a206/src/app/Components/Consignments/Screens/__tests__/Confirmation-tests.tsx)
-- [Tests for Consignments photo-selection component interactions](https://github.com/artsy/emission/blob/751d24306a2d6ace58b21491e25b37f345c7a206/src/app/Components/Consignments/Screens/__tests__/SelectFromPhotoLibrary-tests.tsx).
-- Consignments Overview is a really complex component, so tests are broken into four test files:
-  - [General component tests](https://github.com/artsy/emission/blob/751d24306a2d6ace58b21491e25b37f345c7a206/src/app/Components/Consignments/Screens/__tests__/Overview-tests.tsx)
-  - [Analytics tests](https://github.com/artsy/emission/blob/751d24306a2d6ace58b21491e25b37f345c7a206/src/app/Components/Consignments/Screens/__tests__/Overview-analytics-tests.tsx)
-  - [Local storage tests](https://github.com/artsy/emission/blob/751d24306a2d6ace58b21491e25b37f345c7a206/src/app/Components/Consignments/Screens/__tests__/Overview-local-storage-tests.tsx)
-  - [Image uploading tests](https://github.com/artsy/emission/blob/751d24306a2d6ace58b21491e25b37f345c7a206/src/app/Components/Consignments/Screens/__tests__/Overview-uploading-tests.tsx)
-  - [`CollectionsRail` tests](https://github.com/artsy/eigen/blob/39644610eb2a5609d992f434a7b37b46e0953ff4/src/app/Scenes/Home/Components/__tests__/CollectionsRail-tests.tsx) demonstrate `relay-test-utils`.
-
 ### Navigation
 
 _Use `react-navigation` for navigating between screens, navigate function and new routes when necessary_
--- a/docs/preferred_practices.md
+++ b/docs/preferred_practices.md
@@ -140,8 +140,10 @@ But we also use `test-renderer` and `enzyme` (in order of preference), that we'd
 
 We use `react-navigation` for navigating between screens, navigate function and new routes.
 
+In iOS we use the Native Switchboard.
+
 See our documentation on adding a route for more details: [Adding a new route](https://github.com/artsy/eigen/blob/main/docs/adding_a_new_route.md).
 
 ### Analytics and tracking
 
-Implementing analytics and tracking [here](./analytics_and_tracking.md).
+In React-native, we use react-tracking as a wrapper for the tracking events we send to Segment. You can read more about the implementation [here](./analytics_and_tracking.md).
--- a/docs/preferred_practices.md
+++ b/docs/preferred_practices.md
@@ -3,6 +3,8 @@
 This is a living document, expected to be updated regularly, with a broad overview of the history and how we prefere to do things on eigen.
 Here you can find links to the tools we use, examples, pull requests with interesting discussions & blog posts.
 
+Last update: February 2022
+
 ## Contents
 
 - [Examples and Hacks](#examples-and-hacks)
@@ -17,15 +19,15 @@ Here you can find links to the tools we use, examples, pull requests with intere
 
 ### Examples & Hacks
 
-Check out our list of [examples](#../../../EXAMPLES.md) and [hacks](#../../../HACKS.md)
+Check out our lists of [examples](#../../../EXAMPLES.md) and [hacks](#../../../HACKS.md).
 
 ### History
 
-The app was initially written in Objective-C and Swift and React Native was added in 2016. Some parts of the app are also written with Kotlin.
+The Artsy app was initially written in Objective-C and Swift and React Native was added in 2016. Some parts of the app are also written with Kotlin.
 
-- **React Native** By default we use React Native for new features.
-- **Objective-C**, **Java**, can be used for bridging code to react native (this is referring to native modules that need to talk to javascript, more info in the react native docs here: https://reactnative.dev/docs/native-modules-ios)
-- **Swift**, **Kotlin** for native functionality that can't be done in React Native (such as: an iOS Widget or a Push Notification Extension).
+- **React Native** is our preferred option for developing new features.
+- **Objective-C** and **Java** can be used for bridging code to react native (this is referring to native modules that need to talk to javascript, more info in the react native docs here: https://reactnative.dev/docs/native-modules-ios)
+- **Swift** and **Kotlin** are used for native functionality that can't be done in React Native (such as: an iOS Widget or a Push Notification Extension).
 
 - [Why Artsy uses React Native](http://artsy.github.io/blog/2016/08/15/React-Native-at-Artsy/)
 - [All React Native posts on Artsy's Engineering Blog](http://artsy.github.io/blog/categories/reactnative/)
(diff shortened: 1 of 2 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (Rename modal title - patrinoua)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against bbc4a2903d5b592ea6b3f534ac3dbf23789d2c2c
</p>

docs/preferred_practices.md
```diff
@@ -26,39 +21,26 @@ Links should point to specific commits, and not a branch (in case the branch or
...
-- Initializing the React Native runtime.
+### History
 
-Everything else is written in React Native.
-
-### Use React Native for new feature development
-
-New features should be built in React Native. The React Native runtime currently requires an existing user ID and access token to be loaded, and sign up/in is still handled in Native code.
+The app was initially written in Objective-C and Swift and React Native was added in 2016. Some parts of the app are also written with Kotlin.
+New features should be built in React Native.
```
admbtlr, line 27: Something like `Swift and Kotlin should only be used for "glue" code, or native functionality that is impossible to implement with React Native. Objective-C should be avoided.`

docs/preferred_practices.md
```diff
@@ -120,30 +103,32 @@ However, if we have a `Button` folder which exports only one button component, w
...
-- [Artsy JavaScriptures seminar on Relay](https://github.com/artsy/javascriptures/tree/master/4_intro-to-relay)
+- [Artsy's Relay Workshop](https://github.com/artsy/relay-workshop)
 - Collections
+  // should we update these links?
   - [A top-level Relay component](https://github.com/artsy/eigen/blob/39644610eb2a5609d992f434a7b37b46e0953ff4/src/lib/Scenes/Collection/Collection.tsx)
   - [A fragment container](https://github.com/artsy/eigen/blob/39644610eb2a5609d992f434a7b37b46e0953ff4/src/lib/Scenes/Collection/Components/FeaturedArtists.tsx)
+    We recently also started using relay-hooks and will soon link documents to that.
 
-### Prefer Relay containers (Higher Order Components) over Hooks
+### Prefer Relay containers (Higher Order Components) over Hooks // !! is this still valid?
```
admbtlr, line 118: We should prefer relay hooks for all new implementations. Refactoring old containers to use hooks is encouraged.

docs/preferred_practices.md
```diff
@@ -120,30 +103,32 @@ However, if we have a `Button` folder which exports only one button component, w
...
 
 We have a preference for Relay containers due to `relay-hooks` hooks not being compatible with Relay containers which represent the majority of our components using Relay.
```
admbtlr, line 120: Pretty sure this can be removed - right @olerichter00 ?

docs/preferred_practices.md
```diff
@@ -120,30 +103,32 @@ However, if we have a `Button` folder which exports only one button component, w
...
 
 - [Relay Container approach](https://github.com/artsy/eigen/blob/21fbf9e24eaa281f3e16609da5d38a9fb62a5449/src/lib/Scenes/MyAccount/MyAccount.tsx#L70)
```
admbtlr, line 122: and this should be removed and replaced with a canonical relay hooks-based component

admbtlr:
Thanks @patrinoua , this is a great start!

docs/preferred_practices.md
brainbicycle: minor delta from RFC: https://github.com/artsy/README/pull/426 Objective-C can still be used for the thin bridging code, Swift for major feature work.

patrinoua: @brainbicycle Thanks! 
Since you are more familiar with this topic, could you perhaps help me with some more context and the right phrasing here?

How about something like this:

- **React Native** By default we use React Native for new features. 
- **Kotlin**  should only be used for "glue" code, or native functionality that is impossible to implement with React Native. 
- **Objective-C** can be used for the thin bridging code (personal note: i am not sure what that means) and 
- **Swift** for major feature work that can't be done in React Native (such as: ___).

brainbicycle: Sure! I think it should be something like this:

- **React Native** By default we use React Native for new features.
- **Objective-C**, Java can be used for bridging code to react native (this is referring to native modules that need to talk to javascript, more info in the react native docs here: https://reactnative.dev/docs/native-modules-ios)
- **Swift, Kotlin** for native functionality that can't be done in React Native (such as: an iOS Widget or a Push Notification Extension).

brainbicycle: let me know if you have any questions on that

brainbicycle:
Thanks for taking initiative on this! I can volunteer to update the navigation section which is way out of date at this point.

docs/preferred_practices.md
```diff
@@ -120,30 +103,32 @@ However, if we have a `Button` folder which exports only one button component, w
...
 
 ### styled-system / styled-components
 
+// !! is this still valid?
```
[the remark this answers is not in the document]
pvinis: i feel like the following two list items could go away and be replaced by something like `We use [palette](src/palette) which uses styled-system under the hood`.

olerichter00:
@patrinoua Thanks a lot, this is great work! 🌟 

I'll create another PR and revise sections/docs about Relay according to this [RFC](https://github.com/artsy/eigen/issues/6080) about using Relay Hooks in favor of the container-based APIs.

olerichter00:
I've updated the sections about Relay in https://github.com/artsy/eigen/pull/6287

docs/preferred_practices.md
```diff
@@ -1,64 +1,54 @@
...
 
-- Sign up/in flow ("onboarding").
-- Live Auctions Integration (LAI) view controller and networking.
-- The Auction view controller.
-- The SwitchBoard (see "SwitchBoard" section below) to navigate between view controllers.
-- The top-level tab bar, and each tab's navigation controller.
-- Deep-link and notification handling (via SwitchBoard).
-- Analytics for Native UI.
-- Initializing the React Native runtime.
+Check out our list of [examples](#../../../EXAMPLES.md) and [hacks](#../../../HACKS.md)
```
pvinis, line 27:
```suggestion
Check out our list of [examples](/EXAMPLES.md) and [hacks](/HACKS.md)
```
I think this should make it link nicely

pvinis (dismissed):

docs/preferred_practices.md
```diff
@@ -120,64 +87,56 @@ However, if we have a `Button` folder which exports only one button component, w
...
-- [Emission switchboard to call out to Eigen](https://github.com/artsy/emission/blob/751d24306a2d6ace58b21491e25b37f345c7a206/Pod/Classes/Core/ARSwitchBoardModule.m)
-- [Callbacks between JS and native code are set up here](https://github.com/artsy/emission/blob/24c0fdaf91aa79654a33fd6e476405047819be5b/Pod/Classes/TemporaryAPI/ARTemporaryAPIModule.m).
+- We write native unit tests when we work with native code
+- We don’t use snapshot tests; they produce too much churn for too little value. It’s okay to test that a component doesn’t throw when rendered, but use [`extractText`](https://github.com/artsy/eigen/blob/4c7c9be69ab1c2095f4d2fed11a040b1bde6eba8/src/lib/tests/extractText.ts) (or similar) to test the actual component tree.
 
-### Analytics
+### Navigation
 
-#### Follow the tracking docs and examples
+We use `react-navigation` for navigating between screens, navigate function and new routes.
```
brainbicycle, line 136:
```suggestion
We use `react-navigation` for navigating between screens. For adding a screen that corresponds to a page on artsy.net use the `navigate(<route-name>)` function and add a new route.
```

brainbicycle (changes requested):
minor wording change request on navigation, otherwise looks great to me!

docs/preferred_practices.md
MounirDhahri: @patrinoua we don't actually use react-navigation for navigation in iOS. we use instead the Native Switchboard there. I would probably mention that and then do as you did by referring users to the add a new route docs

docs/preferred_practices.md
```diff
@@ -120,30 +109,32 @@ However, if we have a `Button` folder which exports only one button component, w
 
 `Note:` Updating capitalisation on folders can cause issues in git and locally so please refrain from renaming existing folders until we come up with a strategy about this. (TODO)
 
-### Use Relay for Network Requests
+### Relay
```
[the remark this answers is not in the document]
patrinoua: Thanks Brian, I added "We use relay for network requests." to make it even more clear, while keeping it friendly 🤣, is it ok like this?

pvinis:
Do you wanna rename the file in this PR too?

docs/preferred_practices.md
```diff
@@ -120,64 +87,56 @@ However, if we have a `Button` folder which exports only one button component, w
...
+We use `react-navigation` for navigating between screens, navigate function and new routes.
 
-See our docs on implementing analytics [here](./analytics_and_tracking.md)
+See our documentation on adding a route for more details: [Adding a new route](https://github.com/artsy/eigen/blob/main/docs/adding_a_new_route.md).
 
-### Miscellaneous
+### Analytics and tracking
 
-- [Making network requests outside of Relay](https://github.com/artsy/emission/blob/019a106517b31cebfb1c5293891215cc7ebf7a4d/src/app/Components/Consignments/Screens/Overview.tsx#L135-L150)
+Implementing analytics and tracking [here](./analytics_and_tracking.md).
```
MounirDhahri, line 147: In React-native, we use `react-tracking` as a wrapper for the tracking events we send to Segment. You can read more about the implementation [here](./analytics_and_tracking.md).

MounirDhahri (changes requested):

docs/preferred_practices.md
```diff
@@ -1,80 +1,47 @@
...
+Here you can find links to the tools we use, examples, pull requests with interesting discussions & blog posts.
 
-This document is a map. Not of Eigen at a specific time, but a map of how we got here and where we want to go next. This is a living document, expected to be updated regularly, of links to:
-
-- Example code.
-- Pull requests with interesting discussions.
-- Conversations on Slack.
-- Blog posts.
-
-Links should point to specific commits, and not a branch (in case the branch or file is deleted, these links should always work). But it's possible that a file is outdated, that our understanding has moved on since it was linked to; in that case, please update this document.
```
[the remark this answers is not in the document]
patrinoua: I thought I'd make it a bit shorter and less "instructive", as anyway we have quite some reviewing before we add something here.
So I removed the instruction about the edge case of "do not add a link to a branch, but to a commit".

Some parts of the history are also mentioned later.

In general I thought the voice before sounded like "we are struggling a bit" and tried to make it less of that, while still conveying the same information. 

Would you like to bring back some specific thing that is missing or revert all of it?

docs/preferred_practices.md
patrinoua: @brainbicycle @MounirDhahri can you please have a look at this and come up with the right wording for this section and create a PR that we can add here?

docs/preferred_practices.md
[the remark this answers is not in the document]
MounirDhahri: I personally find some of those tips useful but wanted to check with you to understand your point of view. I am fine with removing them definitely

docs/preferred_practices.md
```diff
@@ -1,80 +1,47 @@
...
+Check out our lists of [examples](#../../../EXAMPLES.md) and [hacks](#../../../HACKS.md).
 
-- Sign up/in flow ("onboarding").
-- Live Auctions Integration (LAI) view controller and networking.
-- The Auction view controller.
-- The SwitchBoard (see "SwitchBoard" section below) to navigate between view controllers.
-- The top-level tab bar, and each tab's navigation controller.
-- Deep-link and notification handling (via SwitchBoard).
-- Analytics for Native UI.
-- Initializing the React Native runtime.
```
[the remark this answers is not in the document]
MounirDhahri: I think here we are describing the parts of the app that are still being handled in native instead of react-native on iOS and this is the only place where they're being described

MounirDhahri (dismissed):
Thanks for leading this effort Angela and updating those really old docs. I only have one comment left unresolved about mentioning the parts that are still being handled in native, but it's not blocking.
Thanks again

patrinoua:
@MounirDhahri , @brainbicycle Thanks once again for your feedback and help on meticulously updating our docs. 🙏 
Apart from the benefit this will have for eigen and our newly joining developers it also helps me get a better understanding of how we get things done.

With my latest commit
- I've updated the parts that have to do with navigation and added a section on Android and iOS parts.
- I've added a Miscellaneous section and as a subsection "Parts of the app that are still being handled in native code (Objective-C and Swift) instead of react-native on iOS" with the list of old components still being used.

--- 

One thing I'm not 100% on board with and would like to discuss a bit further is adding links to commits:

I consider this to be an edge case because I'm thinking that if we point to a file as a best practice for example, and then the file gets a more recent update, the more recent update is also going to have an even better practice.

While updating these docs, I've added a couple of links to files but not PRs/branches/commits. 

There are a couple of links that point to commits instead of files left under testing. And I can't say I'd feel 100% good about using those without asking, as I'd wonder if they are up to date, since some of them are from 2020 and a lot has changed since then.

So I think that adding a commit to a file that is on the main branch is helping us from keeping things that are outdated. 
Of course if the file gets removed then ideally we also eventually remove it from our docs. 
But I guess this is a far less the case, than committing more things to a relatively active part of the app / component.

This is why I think the standard practice should be adding links to files on main and not just a specific commit when it comes to files.

In general I am not in favour of pointing to PRs/branches but is there a good case where we do that? Do we really need it as part of our preferred_practices ?

I'd love to also see your thoughts about that 🙏

MounirDhahri:
> One thing I'm not 100% on board with and would like to discuss a bit further is adding links to commits:
> 
> I consider this to be an edge case because I'm thinking that if we point to a file as a best practice for example, and then the file gets a more recent update, the more recent update is also going to have an even better practice.

The issue with linking files is that they get bigger, get more complicated and someone else can for example update the file and might mistakenly use a non-recommended practice. Files location can change as well, for example after the latest `src/app` rename, all our links to files would have broke if we were using links to files.  
The downside of linking to commits IMO is that the file that we need to link to inside a commit, needs to be as good as we can get it since it won't get updates later.
In both cases, as is the case always for docs, no matter how well it's written, since we're making updates to our codebase, we will need always to update our docs.

patrinoua:
> > One thing I'm not 100% on board with and would like to discuss a bit further is adding links to commits:
> > I consider this to be an edge case because I'm thinking that if we point to a file as a best practice for example, and then the file gets a more recent update, the more recent update is also going to have an even better practice.
> 
> The issue with linking files is that they get bigger, get more complicated and someone else can for example update the file and might mistakenly use a non-recommended practice. Files location can change as well, for example after the latest `src/app` rename, all our links to files would have broke if we were using links to files. The downside of linking to commits IMO is that the file that we need to link to inside a commit, needs to be as good as we can get it since it won't get updates later. In both cases, as is the case always for docs, no matter how well it's written, since we're making updates to our codebase, we will need always to update our docs.

@MounirDhahri if we have an agreed way of doing things the non recommended practice should get caught on review. 
This is the mechanism for maintaining a good code base and why we need to have agreed best practices...

brainbicycle:
> > > One thing I'm not 100% on board with and would like to discuss a bit further is adding links to commits:
> > > I consider this to be an edge case because I'm thinking that if we point to a file as a best practice for example, and then the file gets a more recent update, the more recent update is also going to have an even better practice.
> > 
> > 
> > The issue with linking files is that they get bigger, get more complicated and someone else can for example update the file and might mistakenly use a non-recommended practice. Files location can change as well, for example after the latest `src/app` rename, all our links to files would have broke if we were using links to files. The downside of linking to commits IMO is that the file that we need to link to inside a commit, needs to be as good as we can get it since it won't get updates later. In both cases, as is the case always for docs, no matter how well it's written, since we're making updates to our codebase, we will need always to update our docs.
> 
> @MounirDhahri if we have an agreed way of doing things the non recommended practice should get caught on review. This is the mechanism for maintaining a good code base and why we need to have agreed best practices...

I don't really agree with this, files move and change all the time in a code base and linking to a commit ensures we are linking to exactly the practice we are trying to showcase at a point in time. Say we add a non-commit link to an example test file for testing relay responses but the associated component gets deleted or renamed a few months down the line, now your link is dead. I don't really think we can reasonably expect people to be checking these links every time new code gets added or removed from the codebase. Similarly maybe the data fetching got moved to another component so now the test doesn't show the preferred practice we were trying to showcase at all anymore but might still adhere to best practices otherwise.

docs/preferred_practices.md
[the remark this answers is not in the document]
brainbicycle: @admbtlr added a section with this so I think this can be marked resolved!

patrinoua:
@brainbicycle @MounirDhahri thanks for elaborating more on the commit to link vs file topic. 🙏
I see your point and already on Friday added a commit bringing that instruction back, but probably forgot to push it! I've added today though. [link](https://github.com/artsy/eigen/pull/6243/commits/d54a75039e7c47a6c4c01b7127da2392b512e75c) 
Shall we have another look and if everything is ok merge?

MounirDhahri (approved):
Great, this sounds good to me. This is a big step in refreshing Eigen docs. Thanks @patrinoua for leading this big effort

brainbicycle (approved):
Woo! Thanks a ton for doing this and pushing through all our comments! This is much better!

patrinoua:
🎉 🥳 🎉 We meeerged! 🎉 🥳 🎉

Next I'll make a small PR renaming preferred_practices to best_practices as was discussed in #mobile-practice.

Thank you all so much for contributing to this!

PR #6259, opened 2022-02-18
feat: add release-ios-app and check-pending-release scripts

MounirDhahri:
**[FUTURE FRIDAY]**

### Description

This PR does the following:
- Adds a script to release the ios app from CI. 
    - We actually used this to release **7.2.2** 🎉
- Adds a cron-job that checks if the app has been approved at the store every 4 hours during work days and sends a slack notification to the team
    - It would be nice to iterate on this to make it faster and to stop it from notifying the #practice-mobile if it already notified it, but yeah first good, then great then perfect 🤷‍♂️
- Update the docs to reflect the new changes

Screenshot 2022-02-18 at 15 48 52

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- add release ios app script - mounir, nikita
- add cron-job to check if the app has been approved at the store - mounir

</details>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -502,6 +502,17 @@ jobs:
           command: node scripts/changelog/commitChangelog.js
       - store_artifacts:
           path: .manifests
+  release-ios-app:
+    environment:
+      BUNDLE_PATH: vendor/bundle # path to install gems and use for caching
+    macos:
+      xcode: 13.2.1
+    steps:
+      - checkout
+      - install-gems
+      - run:
+          name: Release app version
+          command: ./scripts/release-ios-app
 workflows:
   version: 2
   nightly:
@@ -544,6 +555,14 @@ workflows:
             branches:
               only:
                 - main
+  release-ios-app-workflow:
+    jobs:
+      - release-ios-app:
+          filters:
+            branches:
+              only:
+                - release_ios_app
+                - add-release-ios-app-script
   promote:
     jobs:
       - promote-beta-to-app-store:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -289,6 +289,21 @@ jobs:
           name: Check flags
           command: ./scripts/biweekly-flag-check.sh
 
+  check-if-app-has-been-approved-for-release:
+    environment:
+      BUNDLE_PATH: vendor/bundle # path to install gems and use for caching
+
+    macos:
+      xcode: 13.2.1
+
+    steps:
+      - checkout
+      - generate-checksums
+      - install-gems
+      - run:
+          name: Check flags
+          command: ./scripts/check-if-app-has-been-approved-for-release
+
   check-metaphysics-freshness:
     executor:
       name: node/default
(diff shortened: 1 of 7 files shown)


brainbicycle (approved):
nice one! lgtm! Could we update the release docs to specify this as the preferred way to release iOS?

MounirDhahri:
I am trying to find a way to automate the step below as well, then I will update the docs and hopefully, we won't need to invite people to apple store connect anymore because everything will be automated
Screenshot 2022-02-18 at 15 22 48

fastlane/Fastfile
```diff
@@ -259,6 +259,7 @@ lane :promote_beta_ios do
   puts "Let's deliver beta #{next_app_store_version} (#{build_number}) with build number #{build_number}."
   deliver(
     api_key: api_key,
+    app_version: app_json['version'],
```
MounirDhahri, line 262: It seems like skip_app_version_update is not sufficient (for now) to create a new version in Apple store connect, and we need to explicitly set the version with Deliver in order for Fastlane to create it. I already submitted a build so I can't test this for now but will test it next build and confirm

MounirDhahri: https://github.com/fastlane/fastlane/issues/18571

brainbicycle: I was actually thinking this would be a separate step from the promotion script, like once the app is approved as part of the next script we could also create the new version? what do you think?

brainbicycle (approved):
Nice one 🤩 ! For creating the new version I was thinking this could be a separate `deliver` call as part of the steps we take after app is approved and released, and maybe as part of the `./scripts/next` script. Since we are doing prep for next release there anyway makes sense to me to have the next version creation be part of that. That said this is already a big improvement so happy to merge as is and follow up on that!

brainbicycle:
> Thanks for the quick review @briansw
> Do you think running `deliver` on its own as a command would create the version in the store? I was under the impression it requires a build. Let me give it a try

there is a `skip_binary_upload` option so I think you can not supply an ipa and it will just create a new version with metadata? I could be wrong though

.circleci/config.yml
```diff
@@ -526,16 +552,17 @@ workflows:
...
-  #         filters:
-  #           branches:
-  #             only:
-  #               - main
-  #   jobs:
-  #     - check-metaphysics-freshness
+  check-if-app-has-been-approved-for-release-workflow:
+    triggers:
+      - schedule:
+          cron: "0 4 * * 1-5"
```
pvinis, line 558:
```suggestion
          cron: "0 4 * * 1-5" # 04:00 every Monday-Friday
```

MounirDhahri:
Loool 😄 sorry other Brian

> there is a skip_binary_upload option so I think you can not supply an ipa and it will just create a new version with metadata? I could be wrong though.

yeah, that makes sense. I will try it out then. But I am going to sign off now because I am meeting a friend in half an hour. But if you or @pvinis would like to give it a try before then, please feel free

pvinis:
> Thanks for the quick review @briansw

lol at that @ :p

pvinis (approved):
omg YES! such a good one!

.circleci/config.yml
MounirDhahri: actually my goal here was to set it to lunch every 4 hours, let me know if it's wrong. I used https://crontab.guru/#0_6_*_*

brainbicycle (approved):

MounirDhahri:
Niiice thanks 🙌

PR #6266, opened 2022-02-18
feat(flag): just adding a flag for now, for dark mode

pvinis:

This PR resolves []

### Description

Every now and then I want to work on dark mode and try things, change things. I always get nowhere, because there is no support in the app at all!
I think if i have a flag, then i can start doing stuff here and there, even try it out with jonathan etc.

anyway, this PR is just adding a flag, and then some small trials:
- a setting screen to sync with the system, or force light or dark mode
- a `v5` theme, and a `v5dark` theme, juuuust in case this takes so long that palette moves to v4 🤣 
- the tab bar supports dark mode 🤔 

Again, everything is behind a flag, and even the calculated state takes the flag in consideration.

Screenshot 2022-02-18 at 22 33 32
Screenshot 2022-02-18 at 22 33 47

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- dark mode flag - pavlos

</details>


ArtsyOpenSource:

- Dev changes (dark mode flag - pavlos)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 81c90c5a3d82afd2378f68f55cf5d75d2fa79c74
</p>

src/lib/Scenes/MyProfile/DarkModeSettings.tsx
```diff
@@ -0,0 +1,37 @@
...
+import React from "react"
+
+export function DarkModeSettings() {
+  const syncWithSystem = GlobalStore.useAppState((state) => state.settings.darkModeSyncWithSystem)
+  const forceMode = GlobalStore.useAppState((state) => state.settings.darkModeForceMode)
+
+  return (
+    <PageWithSimpleHeader title="Dark Mode Settings">
+      <Text variant="lg" textAlign="center">
+        Choose your destiny
```
MounirDhahri, line 14: ⚔️

src/lib/Scenes/MyProfile/SettingsModel.ts
```diff
@@ -0,0 +1,33 @@
+import { action, Action, computed, Computed } from "easy-peasy"
+import { GlobalStoreModel } from "lib/store/GlobalStoreModel"
+import { Appearance } from "react-native"
+
+export interface SettingsModel {
+  // dark mode
+  darkMode: Computed<this, "light" | "dark", GlobalStoreModel>
```
MounirDhahri, line 7: **thought:** I suggest we maybe rename this to `colorScheme` to match the react-native Appearance API

src/lib/Scenes/MyProfile/SettingsModel.ts
```diff
@@ -0,0 +1,33 @@
...
+  darkModeSyncWithSystem: boolean
```
MounirDhahri, line 8: **question:** does this aim to override system default color scheme?

src/lib/store/GlobalStoreModel.ts
```diff
@@ -40,6 +41,7 @@ interface GlobalStoreStateModel {
   config: ConfigModel
   auth: AuthModel
   toast: ToastModel
+  settings: SettingsModel
```
MounirDhahri, line 44: **thought:** would it make sense to move this to the Config model maybe?

MounirDhahri (approved):
This is great Pavlos, Thanks for keeping on pushing on dark mode. I just left few suggestions but none are blocking and are personal preferences
I suggest you probably create a notion doc where we can slowly migrate all screens one by one during future Fridays or mobile office hours and maybe soon we can release this

src/lib/Scenes/MyProfile/SettingsModel.ts
pvinis: good idea. yea, its kind of a rough first PR, just to get things rolling. but ill rename.

src/lib/store/GlobalStoreModel.ts
pvinis: i did think of that, and the reason i didn't go for it, is that it seemed that Config was more "artsy devs configure the app", like flags etc, and Settings is more "users configure their app".

do you think it would be more clear in the config? i honestly dont mind it either way, just tried to keep these two ideas separate, but maybe its overdoing it?

src/lib/Scenes/MyProfile/SettingsModel.ts
pvinis: so the `darkModeSyncWithSystem` means that then this is true, it will do whatever the system says. system is light -> app is light. system is dark -> app is dark.
(again, i think your idea to rename makes sense. ill definitely rename.)

then the one under this one, `darkModeForceMode` is the one that will override the system.


syncWithSystem: true
forceMode: any
result: whatever the system says (system is light -> app is light, system is dark -> app is dark)

syncWithSystem: false
forceMode: light or dark
result: forceMode is light -> app is light (ignore system), forceMode is dark -> app is dark (ignore system)


then the `darkMode` is a computed that gives teh resulting colorscheme, that comes from the logic above.

does this make sense?

src/lib/store/GlobalStoreModel.ts
MounirDhahri: yes, I also don't feel strong about any of the approaches. we also have a user preferences model, I was going to suggest it there but I guess it should be device specific and not user specific as you did.

MounirDhahri: I think maybe the config model can be probably soon renamed to devicePreferences and this will make it easier for us to tell where to place new models (userPreferences or devicePreferences)

pvinis: maybe we could rename all these actually 🤔. some names that suggest these things. like DevicePrefs, UserPrefs, and.. DevPrefs or like ArtsyPrefs?

MounirDhahri: great idea, go for it. it will make life easier

pvinis: done.

src/lib/Scenes/MyProfile/SettingsModel.ts
pvinis: i renamed all these to use hopefully cleaner and clearer names.

MounirDhahri: awesome ,thanks 🙌

src/lib/store/GlobalStoreModel.ts
MounirDhahri: Let me know when I can review this again

pvinis: adding some migrations, and ill ping

src/lib/store/migration.tests.ts
```diff
@@ -511,3 +511,37 @@ describe("App version Versions.AddArtworkViewOption", () => {
...
+
+    const migratedState = migrate({
+      state: previousState,
+      toVersion: migrationToTest,
+    }) as any
+
+    expect(migratedState.devicePrefs.usingSystemColorScheme).toEqual(false)
+    expect(migratedState.devicePrefs.forcedColorScheme).toEqual("light")
+
+    expect(migratedState.config).toEqual(undefined)
```
pvinis, line 539: some extra tests cause this felt scary and wanted to be sure.

pvinis:
this should be good to go 👍

src/lib/store/GlobalStoreModel.ts
pvinis: ready

PR #6268, opened 2022-02-19
feat(startup): sentry small cleanup

pvinis:

This PR resolves []

### Description

I tested on ios.
I would appreciate someone trying out if android builds and shows the app normally for this. I cant seem to log in on my android emu.

The main change here is that I split the hook and the setup parts in two files, so when they get imported they are cleaner with cycles.
- also moved the `bootstrap` env to the index files for a more explicit setup
- added the `useIdentifyUser`, this comes from some work in the unleash PR that will come soon. Having one place to do all the user identification seems helpful.
- seems like android didnt have the sentry setup for the app bootstrapping, added that.
- made the flag into a devtoggle, seems more fitting there.

### To the reviewers 👀

- 

#### iOS user-facing changes

-

-

#### Dev changes

- small sentry cleanup - pavlos

</details>


src/lib/store/config/features.ts
```diff
@@ -267,6 +262,9 @@ export const devToggles = defineDevToggles({
   DTShowInstagramShot: {
     description: "Show Instagram viewshot",
   },
+  DTCaptureExceptionsInSentryOnDev: {
```
MounirDhahri, line 265: **praise:** 👏

src/lib/utils/useIdentifyUser.ts
```diff
@@ -0,0 +1,15 @@
...
+import { SegmentTrackingProvider } from "./track/SegmentTrackingProvider"
+
+export const useIdentifyUser = () => {
+  const userId = GlobalStore.useAppState((store) => store.auth.userID)
+
+  useEffect(() => {
+    console.log("userId", userId)
+
+    Sentry.setUser({ id: userId ?? "none" })
+    SegmentTrackingProvider.identify?.(userId, { is_temporary_user: userId === null ? 1 : 0 })
```
MounirDhahri, line 13: I thought we were doing that already somewhere else, was it not working properly?

MounirDhahri (commented):
Nice PR! I left one small question

src/lib/utils/useIdentifyUser.ts
pvinis: it was, but sentry had its useEffect, segment its own, and when i was adding unleash i thought lets merge, since its all the same thing. so i extracted it, and on the unleash PR ill just add one line here and boom done.
Does that make sense?

src/lib/utils/useIdentifyUser.ts
```diff
@@ -0,0 +1,15 @@
+import * as Sentry from "@sentry/react-native"
+import { GlobalStore } from "lib/store/GlobalStore"
+import { useEffect } from "react"
+import { SegmentTrackingProvider } from "./track/SegmentTrackingProvider"
+
+export const useIdentifyUser = () => {
+  const userId = GlobalStore.useAppState((store) => store.auth.userID)
+
+  useEffect(() => {
+    console.log("userId", userId)
```
olerichter00, line 10: 👀

src/lib/utils/useIdentifyUser.ts
```diff
@@ -0,0 +1,15 @@
+import * as Sentry from "@sentry/react-native"
+import { GlobalStore } from "lib/store/GlobalStore"
+import { useEffect } from "react"
+import { SegmentTrackingProvider } from "./track/SegmentTrackingProvider"
+
+export const useIdentifyUser = () => {
```
olerichter00, line 6: What do you think about adding a comment to document the function?

```suggestion
/**
 * bla bla blub
 */
export const useIdentifyUser = () => {
```

src/lib/utils/useIdentifyUser.ts
pvinis: I thought it's kinda nice to log this so we see when that happens 🤔. seems like an important log to happen. to you think I should remove it?

src/lib/utils/useIdentifyUser.ts
pvinis: oh yea good idea. I'll add it.

src/lib/utils/useIdentifyUser.ts
MounirDhahri: yes, that makes sense but what I meant is that so far, I think we've been calling it inside AuthModel after rehydration and I am not sure if we need it here
https://github.com/artsy/eigen/blob/601f770edd75412b24ff758ecc2535b257b84bb9/src/lib/store/AuthModel.ts#L781

src/lib/utils/useIdentifyUser.ts
olerichter00: Not sure, but we can leave it here. Just wanted to make sure it's intentional 👍

src/lib/utils/useIdentifyUser.ts
pvinis: ah gotcha. let me doublecheck. I am sure I removed it from there, but maybe it was in the unleash PR. I can edit this one and fix this. Good catch!

src/lib/utils/useIdentifyUser.ts
pvinis: yea it was intentional. thanks for checking :)

src/lib/utils/useIdentifyUser.ts
pvinis: updated this PR. much cleaner now, with all the user id notifications out of our state. and easier to add new services in one place rather than different ones every time.

src/lib/utils/useIdentifyUser.ts
araujobarret: aren't we creating some noise on Sentry for all users in this case?

araujobarret (approved):
A lot of decoupling I liked it, gj!

src/lib/utils/useIdentifyUser.ts
MounirDhahri: awesome, thanks Pavlos

MounirDhahri: Would it make sense then to make sure if the state is already hydrated? I mean something like:
```
useEffect(() => {
  if (isHydrated) {
     ....
  }
}, [userId, isHydrated])
```

MounirDhahri: This is just to avoid calling sentry once before the data hydrates

src/lib/utils/useIdentifyUser.ts
pvinis: hm. i guess we are. but i would first argue that our logs are way way to noisy (especially relay), and one extra line after login is fine. but i can see this going away too.

angela has been working on a log cleanup PR in #6099, that we will look at in the next days, so that might be a good place to do some thinking about what should log for devs by default and what we can turn on and off locally, as well as what should we log to sentry, for making it easier for us to read/debug a sentry error 🤔.

olerichter00: I'd also tend to keep the logs to a minimum.

pvinis: what if i make this a `console.debug`? so its not there normally? otherwise i can completely remove this one 👍

src/lib/utils/useIdentifyUser.ts
pvinis: that is actually how this works. once it hydrates, we will get the userid and add it to sentry with `setUser` so that its sent as context in future sentry calls. so the userId change in the useEffect should be enough. adding the `isHydrated` will not give us any more info. in your code here, the only case i see that would be different is if the userid changes while we are hydrated, which would not call setUser in that case. and we definitely want that, because once we hydrate, if the user logs out and then logs in (maybe other user?) then we lose that. so having just the user id an nothing else gives us all we need and nothing more. thats what i think, unless im missing something?

src/lib/utils/useIdentifyUser.ts
araujobarret: There is a discussion in platform team regarding this kind of extra logging, in December we had exceeded the quotas there and we disabled sentry in MP, I see no value in this log, so maybe better removing

src/lib/utils/useIdentifyUser.ts
MounirDhahri: oh if that's intentional then yes, looks good to me. my goal was to not trigger  `Sentry.setUser` and `SegmentTrackingProvider.identify` until we're sure that the app has rehydrated. My goal is to avoid having every session starting with user id as `none`, which is fine for sentry, but not sure if that's fine for Segment

src/lib/utils/useIdentifyUser.ts
olerichter00: I'd remove it unless there is a good reason to log it. But if you'd like to log the user id here to use it, go for it.

olerichter00 (approved):

src/lib/utils/useIdentifyUser.ts
pvinis: good point for sure. i think for sentry its totally fine to have `none`, and for segment it is also ok to have `null`, since once we have the user id and set it, it will be working with the events from before that, thanks to this anonymous id or whatever they call it. i think we are good with this one, but ill keep an eye 👍

src/lib/utils/useIdentifyUser.ts
pvinis: removed :)

pvinis:
i think we are good to go 👍

src/lib/utils/useIdentifyUser.ts
MounirDhahri: great, thanks. I will resolve this then

MounirDhahri (approved):
🚀

PR #6278, opened 2022-02-22
[fix]: 30mb err msg in Submission photo upload step

AlicanAkyuz:

This PR resolves [[SWA-246](https://artsyproduct.atlassian.net/browse/SWA-246)]

### Description
This PR fixes a bug in the photo upload step of the new submission flow. The bug is that when user uploads photos of which total size exceeds 30MB, we are supposed to display an error message to the user. The fix:
- displays the error message on the photo that causes the limit to be exceeded 
- stops the upload of other further photos, if any 
- disables the continue and add photos buttons, when limit exceed
- and clears the error message, when a photo is deleted and the total size is no longer above the limit

#### Videos 
https://user-images.githubusercontent.com/42584148/155089173-cb2bc4e7-0254-475d-b80e-6b0f3e8108e4.mp4

### To the reviewers 👀

#### Dev changes
The changes apply to the devs only, since this whole feature is behind a feature flag for now. 

</details>

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1,5 +1,5 @@
 {
-  "version": "1.1.0",
+  "version": "1.2.0",
   "plugins_used": [
     {
       "name": "ArtifactoryDetector"
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1,5 +1,5 @@
 {
-  "version": "1.2.0",
+  "version": "1.1.0",
   "plugins_used": [
     {
       "name": "ArtifactoryDetector"
--- a/src/lib/Scenes/Consignments/Screens/SubmitArtworkOverview/UploadPhotos/UploadPhotosForm.tsx
+++ b/src/lib/Scenes/Consignments/Screens/SubmitArtworkOverview/UploadPhotos/UploadPhotosForm.tsx
@@ -34,6 +34,7 @@ export const UploadPhotosForm: React.FC<{ isAnyPhotoLoading?: boolean }> = ({
         if (uploadedPhoto?.id) {
           const sizedPhoto = calculateSinglePhotoSize(uploadedPhoto)
           const isTotalSizeLimitExceeded = isSizeLimitExceeded([...values.photos, sizedPhoto])
+          // when total size limit exceeded, set photo's err state and stop the upload loop
           if (isTotalSizeLimitExceeded) {
             sizedPhoto.error = true
             sizedPhoto.errorMessage =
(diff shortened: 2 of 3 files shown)


ArtsyOpenSource:

- Dev changes (The changes apply to the devs only, since this whole feature is behind a feature flag for now.)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against a336606fd19692005a58456e4d9bcca8f50d6b42
</p>

.secrets.baseline
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "1.1.0",
+  "version": "1.2.0",
```
iskounen, line 2: **question (non-blocking):** What does it mean when this version number gets incremented?

src/lib/Scenes/Consignments/Screens/SubmitArtworkOverview/UploadPhotos/UploadPhotosForm.tsx
```diff
@@ -27,12 +28,20 @@ export const UploadPhotosForm: React.FC<{ isAnyPhotoLoading?: boolean }> = ({
...
         // upload & size the photo, and add it to processed photos
         const uploadedPhoto = await addPhotoToConsignment(photo, submission.submissionId)
         if (uploadedPhoto?.id) {
           const sizedPhoto = calculateSinglePhotoSize(uploadedPhoto)
+          const isTotalSizeLimitExceeded = isSizeLimitExceeded([...values.photos, sizedPhoto])
+          // when total size limit exceeded, set photo's err state and stop the upload loop
+          if (isTotalSizeLimitExceeded) {
+            sizedPhoto.error = true
+            sizedPhoto.errorMessage =
+              "File exceeds the total size limit. Please delete photos or upload smaller file sizes"
```
iskounen, line 41: **nitpick:** We should add a period (full-stop) at the end of this message. (I'm sorry for missing that in the Jira ticket.)

iskounen (approved):
These changes look good 🎉 

I have not yet tested on a simulator, but I am curious to see what happens in the following cases:

- When you upload more than 1 photo that causes the total to exceed 30 MB.
- When you have already exceeded 30 MB and upload another photo.

.secrets.baseline
iskounen: Never mind. I saw the thread in Slack about this :)

AlicanAkyuz:
@iskounen 
1- When you upload more than 1 photo that causes the total to exceed 30 MB -> the process stops at the photo that causes the excess. The error shown for that photo, uploading of other photos is cancelled. 

2- When you have already exceeded 30 MB and upload another photo -> when 30mb limit is exceeded, Add Photos buttton is disabled. User cannot add another photo unless they delete a photo and open up space.

iskounen:
@AlicanAkyuz I am seeing (2) but I am not seeing (1) happening. Here is my test case:

1. Upload 20.1 MB of images
2. Wait for processing to complete
3. Upload another 20.1 MB of images

The result is that the Save & Continue button and the Add Photo buttons are disabled, but I do not see an error message.

multi-file-overload

AlicanAkyuz:
@iskounen thank you for catching this! 🙏  It happens when, in the second patch of photos, one of the photos that's not the 1st one causes the excess. This is because `values.photos` are updated outside of the for loop, so I had to bring `processedPhotos` into the calculation like below. 

```
 const isTotalSizeLimitExceeded = isSizeLimitExceeded([
            ...values.photos,
            ...processedPhotos,
            sizedPhoto,
])
```

AlicanAkyuz:
Adding a video after the last commit for convenience. 

https://user-images.githubusercontent.com/42584148/155155274-d7d15f27-e601-48ea-ace5-5989190f4bcd.mp4

patrinoua:
Thanks @AlicanAkyuz this looks good! 
The only thing I'm a bit worried about (@iskounen for your reference) is that the upload takes too long, we don't have a message so the user knows they should wait and might think it's stuck! Maybe we could add a small thing there?

PR #6279, opened 2022-02-22
fix: Add root Suspense component

olerichter00:

### Description

Because we mix Relay query renderers with Relay hooks, it can happen that components using `useFragment` suspend without being wrapped in a `Suspense` component. This PR adds a root `Suspense` component to avoid them from throwing an error. 

**Example:**

```
<MyQueryRenderer>
  ...
    <NestedComponentUsingRelayHooks />
  ...
</MyQueryRenderer>
```

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- Add root Suspense component - ole

</details>

### DIFF
--- a/src/lib/AppProviders.tsx
+++ b/src/lib/AppProviders.tsx
@@ -41,7 +41,9 @@ const RelayDefaultEnvProvider = (props: { children?: ReactNode }) => (
   <RelayEnvironmentProvider environment={defaultEnvironment} {...props} />
 )
 
-const SuspenseProvider = () => <Suspense fallback={<Spinner />} />
+const SuspenseProvider = (props: { children?: ReactNode }) => (
+  <Suspense fallback={<Spinner />} {...props} />
+)
 
 // react-track has no provider, we make one using the decorator and a class wrapper
 const TrackingProvider = (props: { children?: ReactNode }) => <PureWrapper {...props} />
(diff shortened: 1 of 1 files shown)


pvinis (approved):
very nice

src/lib/AppProviders.tsx
```diff
@@ -25,6 +25,7 @@ export const AppProviders = ({ children }: { children?: ReactNode }) =>
       RelayDefaultEnvProvider,
       ThemeProvider, // uses: GlobalStoreProvider
       RetryErrorBoundary,
+      SuspenseProvider,
```
MounirDhahri, line 28: **thought:** would it make sense to maybe move this a bit more down in the hierarchy to be able to use it inside the FancyModal context?

olerichter00: @pvinis `SuspenseProvider` should wrap `_FancyModalPageWrapper` if it is above it, right?

MounirDhahri: I thought of this in the opposite way where we wrap the `SuspenseProvider` with  `_FancyModalPageWrapper` to stay withing the fancy modal context but it's probably not something we do need for now but we better keep it in mind

MounirDhahri: this is probably not needed anyway because we have suspense wrappers on the component level as well 😄

olerichter00: 👍

PR #6308, opened 2022-02-23
feature(DX): add an fps counter and some docs for devtoggles

pvinis:
The type of this PR is: **Feature/Documentation**

This PR resolves []

### Description

When pairing with @araujobarret a few days ago, we needed an fps counter in eigen. unfortunately it doesnt work on ios, so i added one with a small library and a little view. pretty nice little dev toggle.

I also added an example for how to add dev toggles, since I was there.

This is how it looks if you enable it:
Screenshot 2022-02-23 at 20 17 06
Screenshot 2022-02-23 at 20 30 13

And if you tap it, it will go almost invisible, so its not in the way if you need to see something behind it.
Screenshot 2022-02-23 at 20 17 14

(Its actually purple, the other screenshots are before i put the right color in!)

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- add an fps counter dev toggle - pavlos

</details>

### DIFF
--- a/src/app/Components/FPSCounter.tsx
+++ b/src/app/Components/FPSCounter.tsx
@@ -1,0 +1,61 @@
+import React, { useState } from "react"
+import { useFps } from "react-fps"
+import { Text, TouchableOpacity, View, ViewStyle } from "react-native"
+
+export function FPSCounter({ style }: { style?: ViewStyle }) {
+  const { currentFps, avgFps, fps, maxFps } = useFps(20)
+  const [opacity, setOpacity] = useState(1)
+  const textHeight = 20
+  const graphHeight = 40
+  const borderWidth = 1
+  const viewHeight = textHeight + graphHeight + 2 * borderWidth
+
+  return (
+    <View
+      style={{
+        position: "absolute",
+        height: viewHeight,
+        width: 86,
+        left: 4,
+        bottom: 4,
+        backgroundColor: "white",
+        borderColor: "black",
+        borderWidth,
+        paddingHorizontal: 2,
+        opacity,
+        ...style,
+      }}
+    >
+      <TouchableOpacity
+        onPress={() => {
+          setOpacity(opacity === 1 ? 0.15 : 1)
+        }}
+      >
+        <View style={{ height: textHeight }}>
+          <Text style={{ fontSize: 10 }}>
+            FPS {currentFps} <Text style={{ fontSize: 9 }}>(avg {Number(avgFps).toFixed(1)})</Text>
+          </Text>
+        </View>
+        <View
+          style={{
+            flexDirection: "row",
+            alignItems: "flex-end",
+            height: graphHeight,
+          }}
+        >
+          {fps.map((val, i) => (
+            <View
+              key={`${i}`}
+              style={{
+                backgroundColor: "#6E1EFF",
+                height: (graphHeight * val) / maxFps,
+                width: 4,
+                marginBottom: 2,
+              }}
+            />
+          ))}
+        </View>
+      </TouchableOpacity>
+    </View>
+  )
+}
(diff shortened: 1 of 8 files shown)


src/app/Components/FPSCounter.tsx
```diff
@@ -0,0 +1,61 @@
+import React, { useState } from "react"
+import { useFps } from "react-fps"
+import { Text, TouchableOpacity, View, ViewStyle } from "react-native"
```
pvinis, line 3: using RN stuff and not palette here, because we could render it anywhere, even outside `Theme` 🤔

src/app/Components/FPSCounter.tsx
```diff
@@ -0,0 +1,61 @@
...
+            flexDirection: "row",
+            alignItems: "flex-end",
+            height: graphHeight,
+          }}
+        >
+          {fps.map((val, i) => (
+            <View
+              key={`${i}`}
+              style={{
+                backgroundColor: "#6E1EFF",
```
pvinis, line 50: artsy purple boi

src/app/store/config/features.ts
```diff
@@ -237,19 +237,22 @@ export const devToggles = defineDevToggles({
     },
   },
   DTShowAnalyticsVisualiser: {
-    description: "Show analytics visualiser",
```
pvinis, line 240: I shortened the descriptions here, to remove the unnessesary "Show" and "Enable" etc.

ArtsyOpenSource:

- Dev changes (add an fps counter dev toggle - pavlos)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against e3aaa439d77fcdaa40b55560dd1fab18a6087521
</p>

MounirDhahri (approved):
I love this! I needed this many times and I will definitely use it

rajsam003:
That is so cool. I used to have something similar for computer games. It's fairly new to apps for me.
Also could you please mention what the capped fps value should be in order to have a good performance so that because of the code we wrote, it will be easier for us to see that the value exceeds the normal limit?

src/app/Scenes/Onboarding/Onboarding.tsx
```diff
@@ -126,6 +128,7 @@ export const Onboarding = () => {
           {!!showNetworkUnavailableModal && <NetworkAwareProvider />}
         </ArtsyKeyboardAvoidingView>
       </ArtsyKeyboardAvoidingViewContext.Provider>
+      {!!fpsCounter && <FPSCounter style={{ bottom: Platform.OS === "ios" ? 40 : undefined }} />}
```
araujobarret, line 131: why do we need to declare it here too? Isn't together with BottomTabs enough?

araujobarret (approved):
Good job! 💪 
I still like the bare bone that comes with RN but in our setup is hard to bring it back to life 💀  
image

araujobarret:
> That is so cool. I used to have something similar for computer games. It's fairly new to apps for me.
> Also could you please mention what the capped fps value should be in order to have a good performance so that because of the code we wrote, it will be easier for us to see that the value exceeds the normal limit?

There is no rule @rajsam003 , sometimes you can't avoid certain bottlenecks, this is a tool to help you when you're building something and you want to make sure you're not messing up with the UX having so many frames dropped causing a laggy effect to the user.
if you wanna more info about frames in RN take a look: https://reactnative.dev/docs/performance#what-you-need-to-know-about-frames

src/app/Scenes/Onboarding/Onboarding.tsx
pvinis: not with the current setup :(

The navigation is a little all over the place, so the onboarding is a little special for now. we want to bring things together with the app shell work, but until then we need to add it there too 🤷

pvinis:
I would be very happy to remove this when the regular one works again :D

MounirDhahri:
@pvinis I think the regular performance monitor works for Android only now

pvinis:
Yep. I think its because of the weird appshell, so i hope it will work for ios after the appshell work too, but until then, nice to have this on ios, and doesnt hurt to have a second one for now on android too.

MounirDhahri:
looking forward to testing this out, i feel we've been dropping lots of frames recently and it's nice to have this to validate it

brainbicycle (approved):

PR #6364, opened 2022-03-03
[fix]: swa 277/numeric keyboard for inputs in submission flow

AlicanAkyuz:

This PR resolves [SWA-277]

### Description
This PR turns 4 inputs (for year, width, height, depth) in the new submission flow form into numeric inputs. 

### To the reviewers 👀

#### Dev changes
Changes applicable to devs only as the feature behind a flag. 

</details>

[SWA-277]: https://artsyproduct.atlassian.net/browse/SWA-277?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/Consignments/Screens/SubmitArtworkOverview/ArtworkDetails/ArtworkDetailsForm.tsx
+++ b/src/app/Scenes/Consignments/Screens/SubmitArtworkOverview/ArtworkDetails/ArtworkDetailsForm.tsx
@@ -37,6 +37,7 @@ export const ArtworkDetailsForm: React.FC = () => {
         testID="Submission_TitleInput"
         value={values.title}
         onChangeText={(e) => setFieldValue("title", e)}
+        accessibilityLabel="Title"
       />
       <StandardSpace />
       <Input
@@ -46,6 +47,7 @@ export const ArtworkDetailsForm: React.FC = () => {
         testID="Submission_YearInput"
         value={values.year}
         onChangeText={(e) => setFieldValue("year", e)}
+        accessibilityLabel="Year"
       />
       <StandardSpace />
       <Input
@@ -54,6 +56,7 @@ export const ArtworkDetailsForm: React.FC = () => {
         testID="Submission_MaterialsInput"
         value={values.medium}
         onChangeText={(e) => setFieldValue("medium", e)}
+        accessibilityLabel="Materials"
       />
       <StandardSpace />
       <Select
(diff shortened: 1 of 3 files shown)


ArtsyOpenSource:

- Dev changes (Changes applicable to devs only as the feature behind a flag.)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against c8f1e80e08bb07c72fada1613331d5b60e9a76f9
</p>

src/app/Scenes/Consignments/Screens/SubmitArtworkOverview/ArtworkDetails/ArtworkDetailsForm.tsx
```diff
@@ -121,6 +122,7 @@ export const ArtworkDetailsForm: React.FC = () => {
         <Box width="31%" mr={1}>
           <Input
             title="Height"
+            keyboardType="number-pad"
```
kiryl-zubarau, line 125: Is it possible to change keyboardType? I mean if the user wants to type '1 1/2' will it be possible?

kiryl-zubarau:
I think it also makes sense to turn into numeric inputs also EDITION NUMBER and EDITION SIZE fields

src/app/Scenes/Consignments/Screens/SubmitArtworkOverview/ArtworkDetails/ArtworkDetailsForm.tsx
patrinoua: @iskounen what do you think about this?

AlicanAkyuz: I've just pushed a commit aligning the keyboardTypes between this form and the form in myCollection. in this case: 

- `year` is _number-pad_ (user can only type numbers) 
- `editionNumber, editionSize, height, width, depth` are _decimal_pad_ (user can only type numbers and `,`)

let me know if anyone thinks it should be different. cc @iskounen

src/app/Scenes/Consignments/Screens/SubmitArtworkOverview/ArtworkDetails/ArtworkDetailsForm.tsx
```diff
@@ -37,6 +37,7 @@ export const ArtworkDetailsForm: React.FC = () => {
         testID="Submission_TitleInput"
         value={values.title}
         onChangeText={(e) => setFieldValue("title", e)}
+        accessibilityLabel="Title"
```
patrinoua, line 40: ❤️

iskounen: ❤️

src/app/Scenes/Consignments/Screens/SubmitArtworkOverview/ArtworkDetails/ArtworkDetailsForm.tsx
iskounen: Thanks for matching the keyboard type used in My Collection. I think that is the correct solution for this iteration.

I agree with @kiryl-zubarau that the web flow supports characters for imperial units (ie. `"`, `'`, `/`), and _ideally_ we would support those on the app as well. If not being able to enter that becomes a problem for users, then we can solve it for both My Collection as well as Sell with Artsy together.

(It's not as convenient or customary, but users have the option of entering `1.5` instead of `1" 1/2` at the moment.)

PR #6365, opened 2022-03-03
Patrinoua/swa 274/keyboard covers input field

patrinoua:
The type of this PR is: **bug**

This PR resolves [SWA-274]
Keyboard covers half of phone number field
as well as a small fix for "what's this?" text colour 

Screenshot 2022-03-03 at 13 57 54

156367355-6e43b8ad-a7c2-4870-9ca3-66e7503f7cb8

### To the reviewers 👀

### Changelog updates

- Rename modal title - patrinoua

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[SWA-274]: https://artsyproduct.atlassian.net/browse/SWA-274?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Components/ArtsyKeyboardAvoidingView.stories.tsx
+++ b/src/app/Components/ArtsyKeyboardAvoidingView.stories.tsx
@@ -1,0 +1,35 @@
+import { storiesOf } from "@storybook/react-native"
+import { Button, Flex, Input, Spacer, Text } from "palette"
+import React from "react"
+import { ScrollView } from "react-native"
+import { ArtsyKeyboardAvoidingView } from "./ArtsyKeyboardAvoidingView"
+
+storiesOf("ArtsyKeyboardAvoidingView", module).add("ArtsyKeyboardAvoidingView", () => (
+  <ArtsyKeyboardAvoidingView>
+    <Flex>
+      <ScrollView>
+        <Text>"Hello"</Text>
+        <Text>"Hello"</Text>
+        <Text>"Hello"</Text>
+        <Text>"Hello"</Text>
+        <Text>"Hello"</Text>
+        <Text>"Hello"</Text>
+        <Text>"Hello"</Text>
+        <Text>"Hello"</Text>
+        <Text>"Hello"</Text>
+        <Text>"Hello"</Text>
+        <Input title="Email" />
+        <Input title="Hello" />
+        <Input title="Hello" />
+        <Input title="Hello" />
+        <Input title="Hello" />
+        <Input title="Hello" />
+        <Input title="Hello" />
+        <Input title="Hello" />
+        <Input title="Hello" />
+        <Spacer mt={4} />
+        <Button>"Submit"</Button>
+      </ScrollView>
+    </Flex>
+  </ArtsyKeyboardAvoidingView>
+))
--- a/src/app/Scenes/Consignments/Screens/SubmitArtworkOverview/ContactInformation/ContactInformation.tsx
+++ b/src/app/Scenes/Consignments/Screens/SubmitArtworkOverview/ContactInformation/ContactInformation.tsx
@@ -8,7 +8,6 @@ import { GlobalStore } from "app/store/GlobalStore"
 import { Formik } from "formik"
 import { CTAButton, Flex, Input, Spacer, Text } from "palette"
 import React, { useState } from "react"
-import { ScrollView } from "react-native"
 import { createFragmentContainer, graphql, QueryRenderer } from "react-relay"
 import { useTracking } from "react-tracking"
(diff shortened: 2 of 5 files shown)


AlicanAkyuz (commented):
**Question**: this looks good on a simulator, but I wonder: would it scroll below, so that the phone input is above the keyboard, when I click on the phone input on a real device? 🤔 

Simulator Screen Shot - iPod touch (7th generation) - 2022-03-03 at 14 38 38

patrinoua:
@kiryl-zubarau @pvinis what do you think about this solution, is there maybe a better way to do this?

src/app/Scenes/Consignments/Screens/SubmitArtworkOverview/ContactInformation/ContactInformation.tsx
```diff
@@ -66,43 +67,49 @@ export const ContactInformation: React.FC<{
...
+              onBlur={handleBlur("userPhone")}
+            />
+            <Spacer mt={6} />
+            <CTAButton onPress={handleSubmit} disabled={!isValid}>
+              Submit Artwork
+            </CTAButton>
+            <Spacer mt={6} />
+            <Spacer mt={6} />
+            <Spacer mt={6} />
+            <Spacer mt={6} />
```
iskounen, line 111: **thought:** It looks like you are solving this problem by elongating the content and wrapping it in a scrollview, which doesn't seem like a "best practice." (However, I can't find another form in the app that handles this problem well.)

I searched the app for another form that may suffer from this issue, and it seems like the password form in the settings screen also has a field (confirm new password) that is hidden by the keyboard.

patrinoua: I added some space so that the user can scroll up when the keyboard is active.
I am not sure if there is a better way to do this, but of course I'm happy to change it if we come up with something better.
@pvinis @kiryl-zubarau @MounirDhahri for your reference

kiryl-zubarau: In react web it may be solved using steps like that:
1. You have to understand is the input on the screen or not (may be solved by using getBoundingClientRect api)
2. Scroll to the input if it isn't on the screen (scrollIntoView)
I think react native should have something like that, but I'm not sure.

pvinis:
this is definitely a funky workaround and it would not work great on different sizes of phones etc. I would suggest looking into the `keyboardaware*` components we have, or if you need more help we can find a time to pair on 🤔.

dariakoko:
Looks like we've got the same issue on "Edit Profile" screen
I've just tried wrapping ScrollView with `ArtsyKeyboardAvoidingView` - it does nothing and `KeyboardAvoidingView` - it makes the screen scroll to the text input

Try like this:
```
<KeyboardAvoidingView behavior="position" enabled>
<ScrollView>
.....
</ScrollView>
</KeyboardAvoidingView>
```

patrinoua:
Thank you all for the suggestions, I implemented ArtsyKeyboardAvoidingView and it seems to work now

src/app/Scenes/Consignments/Screens/SubmitArtworkOverview/ContactInformation/ContactInformation.tsx
```diff
@@ -77,14 +77,17 @@ export const ContactInformation: React.FC<{
             value={values.userName}
             onBlur={handleBlur("userName")}
             error={touched.userName ? errors.userName : undefined}
+            accessibilityLabel="Name"
```
pvinis, line 80: nice

src/app/Components/ArtsyKeyboardAvoidingView.stories.tsx
```diff
@@ -0,0 +1,35 @@
+import { storiesOf } from "@storybook/react-native"
+import { Button, Flex, Input, Spacer, Text } from "palette"
+import React from "react"
+import { ScrollView } from "react-native"
+import { ArtsyKeyboardAvoidingView } from "./ArtsyKeyboardAvoidingView"
+
+storiesOf("ArtsyKeyboardAvoidingView", module).add("ArtsyKeyboardAvoidingView", () => (
```
pvinis, line 7: awesome

pvinis (commented):
code looks good and clean

kiryl-zubarau (approved):

AlicanAkyuz (approved):

PR #6374, opened 2022-03-04
feat: Update currency and metrics

rajsam003:

This PR resolves [CX-2320]

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- Update currency and length unit - sam

</details>

[CX-2320]: https://artsyproduct.atlassian.net/browse/CX-2320?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ArtsyOpenSource:

- Dev changes (Update currency and length unit - sam)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 234a9813f0455c7948c699bf61e5cc6b44acda64
</p>

src/app/Scenes/MyCollection/Screens/ArtworkForm/MyCollectionArtworkForm.tsx
```diff
@@ -91,6 +102,10 @@ export const MyCollectionArtworkForm: React.FC<MyCollectionArtworkFormProps> = (
   const handleSubmit = async (values: ArtworkFormValues) => {
     setLoading(true)
     try {
+      await updateMyUserProfile({
```
rajsam003, line 105: Instead of calling the API in the model. I called it here. With this, we dont need any migration.

src/app/Scenes/MyCollection/Screens/ArtworkForm/MyCollectionArtworkForm.tsx
```diff
@@ -104,7 +119,24 @@ export const MyCollectionArtworkForm: React.FC<MyCollectionArtworkFormProps> = (
...
 
+  const loadLengthUnitAndCurrency = async () => {
+    try {
+      const me = await fetchMe()
+
+      const metricUnitFromGravity = me?.lengthUnitPreference
+      const currencyFromGravity = me?.currencyPreference
+
+      if (metricUnitFromGravity && currencyFromGravity) {
+        GlobalStore.actions.userPrefs.setMetric(metricUnitFromGravity.toLowerCase() as Metric)
```
rajsam003, line 130: I used `lowercase()` because in the API we have uppercase and in Eigen we have lowercase values. `["cm", "in"]`

src/app/Scenes/MyCollection/Screens/ArtworkForm/MyCollectionArtworkForm.tsx
```diff
@@ -206,7 +238,12 @@ export const MyCollectionArtworkForm: React.FC<MyCollectionArtworkFormProps> = (
             <Stack.Screen
               name="ArtworkFormArtwork"
               component={MyCollectionArtworkFormArtwork}
-              initialParams={{ onDelete, clearForm, mode: props.mode, onHeaderBackButtonPress }}
+              initialParams={{
```
rajsam003, line 241: This is not my change, this is prettier's change.

src/app/Scenes/MyCollection/Screens/ArtworkForm/MyCollectionArtworkForm.tsx
rajsam003: If we want to change we have to check in the whole project. We might need a separate ticket for that. 
Also, the naming is different. In the API it is `lengthUnitPreference` and in Eigen it is `Metric`

src/app/Scenes/MyCollection/Screens/ArtworkForm/MyCollectionArtworkForm.tsx
olerichter00: Wouldn't it make more sense to move the logic to [UserPrefsModel](https://github.com/artsy/eigen/blob/ac35bedebfa65d75c373f0f6030208be43816285/src/app/Scenes/Search/UserPrefsModel.tsx#L27)?

We already use this pattern [here](https://github.com/artsy/eigen/blob/ac35bedebfa65d75c373f0f6030208be43816285/src/app/Scenes/BottomTabs/BottomTabsModel.ts#L37)

src/app/Scenes/MyCollection/Screens/ArtworkForm/MyCollectionArtworkForm.tsx
```diff
@@ -104,7 +119,24 @@ export const MyCollectionArtworkForm: React.FC<MyCollectionArtworkFormProps> = (
     }
   }
 
+  const loadLengthUnitAndCurrency = async () => {
```
olerichter00, line 122: What do you think about hydrating the store with the user's remote preferences when the app starts? The preferences will potentially be used in different places in the app and hydrating the store for every screen is probably not the best idea.

We do something similar with [Echo](https://github.com/artsy/eigen/blob/ac35bedebfa65d75c373f0f6030208be43816285/src/app/store/config/EchoModel.tests.tsx#L45).

src/app/Scenes/MyCollection/Screens/ArtworkForm/MyCollectionArtworkForm.tsx
olerichter00: 👍

olerichter00 (changes requested):
This looks great so far! I would just move some of the code.

Will there be a follow-up PR to update preferences in Gravity if the user changes them on the device?

olerichter00:
@rajsam003 I'm sure this is a **feature** and not a **chore**. https://gist.github.com/joshbuchea/6f47e86d2510bce28f8e7f42ae84c716

rajsam003:
> @rajsam003 I'm sure this is a **feature** and not a **chore**. https://gist.github.com/joshbuchea/6f47e86d2510bce28f8e7f42ae84c716

We are not introducing any new feature. So I thought it should be a chore. 🤔

olerichter00:
> We are not introducing any new feature. So I thought it should be a chore. 🤔

Well, the new feature is that the app considers the artsy-wide metric preferences of the user.

olerichter00:
Even if we haven't implemented any logic to update the preferences yet, this change will already set the metric preferences to country-specific default values.

src/app/Scenes/MyCollection/Screens/ArtworkForm/MyCollectionArtworkForm.tsx
rajsam003: Thanks for this idea. I just checked [the docs](https://easy-peasy.vercel.app/docs/api/persist.html#rehydrating-your-store). I will try this.

src/app/Scenes/MyCollection/Screens/ArtworkForm/MyCollectionArtworkForm.tsx
rajsam003: Do you mean calling [this query](https://github.com/artsy/eigen/blob/234a9813f0455c7948c699bf61e5cc6b44acda64/src/app/Scenes/MyCollection/Screens/ArtworkForm/MyCollectionArtworkForm.tsx#L335)
?

rajsam003: Because this line of code is actually updating gravity based on user selection and while saving the artwork.

olerichter00 (commented):
Thanks a lot for the update 🙇  

Just left a comment with a question.

olerichter00 (changes requested):
Very nice!

Now we just have to make sure to set the metrics when the user logs in or signs up and to prevent setting the metrics in `didRehydrate` when the user is not logged in. 

The current solution will set both metric prefs to `undefined` if the store rehydrates when the user is not logged in.

rajsam003:
> 

Thanks @olerichter00 
I just checked this case. When the user is not logged and the store rehydrates,  the preferences gets default values from the modal i.e. `USD` for currency and `cm` for the metric unit. 

Currently, there is [no default value for metirc](https://github.com/artsy/eigen/blob/b1b053f02cd1b13918c10e75d5e08a8ce0c32338/src/app/Scenes/Search/UserPrefsModel.tsx#L22). I am pushing the change in my next commit. 

This can overcome the situation of `undefined` values. Also, I dont think there is a way to use the app without login or signup. 🤔

olerichter00:
> Thanks @olerichter00 I just checked this case. When the user is not logged in and the store rehydrates, the preferences gets default values from the modal i.e. `USD` for currency and `cm` for the metric unit.

My idea is that we want to have the default values from Gravity as soon as the user logs in or signs up because these default values are country specific.
> 
> Currently, there is [no default value for metirc](https://github.com/artsy/eigen/blob/b1b053f02cd1b13918c10e75d5e08a8ce0c32338/src/app/Scenes/Search/UserPrefsModel.tsx#L22). I am pushing the change in my next commit.

This is ok, I guess.

> 
> This can overcome the situation of `undefined` values. Also, I dont think there is a way to use the app without login or signup. 🤔

If you're starting the app while being logged I see a warning that the currency is not supported because the query returns `null` in this case.

src/app/Scenes/Search/UserPrefsModel.tsx
```diff
@@ -43,19 +49,44 @@ export const getUserPrefsModel = (): UserPrefsModel => ({
...
     } else {
       console.warn("Metric/Dimension Unit not supported")
     }
+  }),
+  fetchRemoteUserPrefs: thunk(async (actions) => {
+    const me = await fetchMe()
 
-    // TODO: update unit in gravity
+    actions.setMetric(me?.lengthUnitPreference.toLowerCase() as Metric)
+    actions.setCurrency(me?.currencyPreference as Currency)
```
olerichter00, line 64: We don't want to set the values if the query fails or returns null.

```suggestion
    const me = await fetchMe()

    if(!me) return

    actions.setMetric(me.lengthUnitPreference.toLowerCase() as Metric)
    actions.setCurrency(me.currencyPreference as Currency)
```

rajsam003:
> My idea is that we want to have the default values from Gravity as soon as the user logs in or signs up because these default values are country specific.

That makes sense. Gravity should send the country-specific default values. Is it already implemented in Gravity or do we need a ticket?

rajsam003:
> If you're starting the app while being logged I see a warning that the currency is not supported because the query returns null in this case.

So we need a ticket to work on Gravity?

olerichter00:
> That makes sense. Gravity should send the country-specific default values. Is it already implemented in Gravity or do we need a ticket?

Yes, it is already implemented.

olerichter00:
> > If you're starting the app while being logged I see a warning that the currency is not supported because the query returns null in this case.
> 
> So we need a ticket to work on Gravity?

We need to fetch the values from Gravity as well as soon as the user logs in or signs up.

olerichter00 (changes requested):
Does the metric preferences get fetched when the user signs up/in? I can't find the code.

src/app/Scenes/Search/UserPrefsModel.tsx
```diff
@@ -15,7 +19,7 @@ const DEFAULT_CURRENCY =
   (userCurrencies.find((userCurrency) =>
     (currencies as unknown as string[]).includes(userCurrency)
   ) as Currency) ?? "USD"
-const DEFAULT_METRIC = ""
+const DEFAULT_METRIC = "cm"
```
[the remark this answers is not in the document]
rajsam003: Should it be `in`?
Because on UI `cm` is the first radio button. So it makes sense to make it default. Just like how `USD` is the first option in Currency select.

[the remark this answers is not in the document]
olerichter00: Good question. In the backend `in` is the default because most of our customers live in the US.

src/app/store/AuthModel.ts
```diff
@@ -379,6 +380,12 @@ export const getAuthModel = (): AuthModel => ({
 
       onSignIn?.()
 
+      // Setting up user prefs from gravity after successsfull login.
+      const me = await fetchMe()
+
+      store.getStoreActions().userPrefs.setMetric(me?.lengthUnitPreference.toLowerCase() as Metric)
+      store.getStoreActions().userPrefs.setCurrency(me?.currencyPreference as Currency)
+
```
rajsam003, line 388: @olerichter00 What do you think about this? It is setting the preferences after successful login. 
Didnt add the logic for` if (!me) {return}` because I dont want to return it if `me` is undefined.

src/app/store/AuthModel.ts
```diff
@@ -455,6 +462,12 @@ export const getAuthModel = (): AuthModel => ({
           assertNever(oauthProvider)
       }
 
+      // Setting up user prefs from gravity after successsfull registration.
+      const me = await fetchMe()
+
+      store.getStoreActions().userPrefs.setMetric(me?.lengthUnitPreference.toLowerCase() as Metric)
+      store.getStoreActions().userPrefs.setCurrency(me?.currencyPreference as Currency)
+
```
rajsam003, line 470: Sam as above

src/app/Scenes/Search/UserPrefsModel.tsx
```diff
@@ -43,19 +50,55 @@ export const getUserPrefsModel = (): UserPrefsModel => ({
...
+
+export const setUserPrefsFromGravity = async () => {
+  const me = await fetchMe()
+
+  if (!me) {
+    return
+  }
+  GlobalStore.actions.userPrefs.setMetric(me?.lengthUnitPreference.toLowerCase() as Metric)
+  GlobalStore.actions.userPrefs.setCurrency(me?.currencyPreference as Currency)
+}
```
rajsam003, line 104: Made this more generic

src/app/store/AuthModel.ts
```diff
@@ -379,6 +380,9 @@ export const getAuthModel = (): AuthModel => ({
 
       onSignIn?.()
 
+      // Setting up user prefs from gravity after successsfull login.
+      setUserPrefsFromGravity()
```
olerichter00, line 384: Can we just call `fetchRemoteUserPrefs` from the user preference model here?

src/app/Scenes/Search/UserPrefsModel.tsx
```diff
@@ -43,19 +50,55 @@ export const getUserPrefsModel = (): UserPrefsModel => ({
...
+        }
+      }
+    `,
+    {}
+  ).toPromise()
+
+  return result?.me
+}
+
+export const setUserPrefsFromGravity = async () => {
```
olerichter00, line 96: Maybe we could give this a different name. Eigen should be agnostic about which service stores the preferences.

olerichter00 (changes requested):

src/app/Scenes/Search/UserPrefsModel.tsx
rajsam003: Then I will revert it to the previous change. No need of this function.

src/app/store/migration.ts
```diff
@@ -175,6 +176,9 @@ export const artsyAppMigrations: Migrations = {
     delete state.config
     delete state.userPreferences
   },
+  [Versions.AddUserPrefsMetricsUnit]: (state) => {
+    state.userPrefs.metric = "in"
```
olerichter00, line 180: Wouldn't this migration override all existing metric preferences?

src/app/store/AuthModel.ts
```diff
@@ -379,6 +379,9 @@ export const getAuthModel = (): AuthModel => ({
 
       onSignIn?.()
 
+      // Setting up user prefs from gravity after successsfull login.
+      GlobalStore.actions.userPrefs.fetchRemoteUserPrefs()
```
olerichter00, line 383: Nice 👍

olerichter00 (changes requested):

src/app/store/migration.ts
rajsam003: Yes, but we get the user preference set by the user from Gravity once log in. That will value will be overridden everywhere.

rajsam003: I think it will override even without login.

olerichter00: 👍

olerichter00 (approved):
🙌

PR #6431, opened 2022-03-15
create a ProgressBar

patrinoua:

Create ProgressBar component to be used under [SWA-267]

- Design followed: https://www.figma.com/file/mcQv2yvaXiacFKZbC74BE0/SWA-Submissions-Flow?node-id=2465%3A16433

- Ideally we could also remove the local implementation of ProgressBar in src/app/Components/ArtsyReactWebView.tsx and update it to use this one. This would also offer us a consistent User experience as the design is slightly different (The bar with unloaded content is white, while in this implementation gray). 

Screenshot 2022-03-15 at 13 35 16

This PR resolves [SWA-2]

### To the reviewers 👀

### Changelog updates

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[SWA-267]: https://artsyproduct.atlassian.net/browse/SWA-267?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SWA-2]: https://artsyproduct.atlassian.net/browse/SWA-2?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/palette/elements/ProgressBar/ProgressBar.tests.tsx
```diff
@@ -0,0 +1,11 @@
+import { renderWithWrappersTL } from "app/tests/renderWithWrappers"
+import React from "react"
+import { ProgressBar } from "./ProgressBar"
+
+describe("ProgressBar", () => {
+  it("renders a ProgressBar", () => {
```
pvinis, line 6: lets remove this test? i see we have a bunch of these "renders" tests, but i feel they are useless. lets remove or add some tests that test something specific on the component? what do you and others think here?

pvinis:
i cant find the designs in the link 🙈

src/palette/elements/ProgressBar/ProgressBar.tests.tsx
patrinoua: I wanted to make this so that it doesn't break, initially, and we make sure it's being rendered.
I'm not sure how to add tests for checking the progression, any idea?

patrinoua: But I'm also happy to remove the test file since it's a simple component

patrinoua:
> i cant find the designs in the link 🙈
Screenshot 2022-03-16 at 10 42 18

pvinis (approved):

src/palette/elements/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,21 @@
+import { Flex } from "palette"
+import React from "react"
+
+export const ProgressBar: React.FC<{ progress: number | null }> = ({ progress }) => {
+  if (progress === null) {
+    return null
+  }
```
MounirDhahri, line 7: **suggestion:** I would instead of returning nothing when the progress is null, make this check out of the component and make sure always that progress is a number

src/palette/elements/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+  const progressPercentage = Math.max(progress * 100, 2)
+
+  return (
+    <Flex width="100%" backgroundColor="black30" my={1}>
```
MounirDhahri, line 12: **nitpick:**I recommend to always use `color(black30)` instead of `black30` to make the transition to a newer design system easier later

MounirDhahri (commented):
Niice, I left one minor comment about having progress as null. We can then merge this I believe

src/palette/elements/ProgressBar/ProgressBar.tsx
pvinis: I think even without color it's still fine, since we use color inside, so we could do changes there. same as with v3 :)

MounirDhahri: ah thanks, I was using color for `useColor` recently to use the hook explicitly and be able to migrate colors one by one if we need to in the future. It looks like I was getting it wrong. thanks for explaining

MounirDhahri (dismissed):
🙌

src/palette/elements/ProgressBar/ProgressBar.tsx
pvinis: both ways work. just the one is simpler, but it depends on the case :). Use whatever fits best for you for each case.

PR #6461, opened 2022-03-18
fix: virtualization issues inside my collection

MounirDhahri:

This PR resolves [CX-2452] the following issue inside my collection artwork and my collection add artist pages
Simulator Screen Shot - iPhone 13 Pro - 2022-03-17 at 14 17 49

### Description
- This is occurring because we are nesting scrollview and this is an anti pattern that the react-native team strongly advice against using. The reason behind that is that it breaks virtualization and it makes it practically impossible for react native to tell on which level to apply the virtualization and therefore makes scrollview behave in a non deterministic manner. The solution is always to make sure that we don't nest any scrollviews and that all scroll events bubble into only one scrollview.

Both screens are working after the update
Simulator Screen Shot - iPhone 13 Pro - 2022-03-18 at 17 53 36

Simulator Screen Shot - iPhone 13 Pro - 2022-03-18 at 17 53 31

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- fix virtualization issues inside my collection screens - mounir

</details>

[CX-2452]: https://artsyproduct.atlassian.net/browse/CX-2452?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -1146,7 +1146,7 @@ SPEC CHECKSUMS:
   CocoaAsyncSocket: 065fd1e645c7abab64f7a6a2007a48038fdc6a99
   CocoaLumberjack: b7e05132ff94f6ae4dfa9d5bce9141893a21d9da
   DHCShakeNotifier: 64048427ecaa763f2472d0032f58bf7a10074eee
-  DoubleConversion: 831926d9b8bf8166fd87886c4abab286c2422662
+  DoubleConversion: cde416483dac037923206447da6e1454df403714
   EDColor: 91d127cd67d7c1d3862846fb11ef4b3851151bfa
   Emission: ce472ad11d5dcd1e332938722aef6e97945e4efb
   Expecta: 3b6bd90a64b9a1dcb0b70aa0e10a7f8f631667d5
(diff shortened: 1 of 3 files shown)


Podfile.lock
```diff
@@ -1146,7 +1146,7 @@ SPEC CHECKSUMS:
   CocoaAsyncSocket: 065fd1e645c7abab64f7a6a2007a48038fdc6a99
   CocoaLumberjack: b7e05132ff94f6ae4dfa9d5bce9141893a21d9da
   DHCShakeNotifier: 64048427ecaa763f2472d0032f58bf7a10074eee
-  DoubleConversion: 831926d9b8bf8166fd87886c4abab286c2422662
```
olerichter00, line 1149: 👀 do you know where they come from? We updated the `Podfile.lock` in [this](https://github.com/artsy/eigen/pull/6446) PR.

olerichter00 (approved):
Thanks for the fix 🙇 

Can you create a ticket for it to not forget the manual QA step?

Podfile.lock
MounirDhahri: This is happening to me actually also on main, I think it's because we forgot to run pod install on main maybe

PR #6472, opened 2022-03-22
fix(SWA-311): visualised lists bug

patrinoua:

This PR resolves [SWA-311]

A VirtualizedList should not be used inside a ScrollView so we removed the Flatlist to get rid of that error

Used an implementation similar to Ole's https://github.com/artsy/eigen/pull/6466/files

This is still not working perfect when erasing text and adding again, which should probably be resolved under a different ticket.

https://user-images.githubusercontent.com/36475005/160102592-aa0aff56-1f23-4b05-9c25-24899e7d4f96.mov

#### Artist Auto Suggest
- [x] Move artist auto suggest to app/components 
- [x] Use Flex instead of a FlatList

#### Location Auto Suggest 
- [x] Move Location Auto Suggest  to app/components 
resolved in different PR by Ole https://github.com/artsy/eigen/pull/6466/files

### To the reviewers 👀

#### iOS user-facing changes
-
-
#### Dev changes
- resolves "Visualised lists" error 

</details>

[SWA-267]: https://artsyproduct.atlassian.net/browse/SWA-267?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[SWA-311]: https://artsyproduct.atlassian.net/browse/SWA-311?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/Components/ArtistAutosuggest/ArtistAutosuggestResults.tsx
```diff
@@ -21,32 +19,18 @@ export type ArtistAutosuggestResult = NonNullable<
...
-      userHasStartedScrolling.current = true
-      loadMore()
-    }
-  }, [])
-
-  const onEndReached = useCallback(() => {
-    if (userHasStartedScrolling.current) {
-      loadMore()
-    }
-  }, [])
```
MounirDhahri, line 49: **thought:** do we never use this?

MounirDhahri:
Thanks for quickly working on this Angela 🙏

src/app/Components/ArtistAutosuggest/ArtistAutosuggestResults.tsx
patrinoua: These were going into the Flatlist and since we've now removed the Flatlist we probably don't need them.
I've also just updated the query to directly fetch 64 instead of 34 artists to eliminate the need for loadMore()

patrinoua:
Also I noticed that ArtistAutosuggest is implemented twice. 

Once in src/app/Components/ArtistAutosuggest/ArtistAutosuggest.tsx (originally implemented under the new submission flow but extracted in this PR)

and once in src/app/Scenes/Search/AutosuggestResults.tsx where we are still using a FlatList so we are probably getting the same error somewhere else as well.

I am not sure this PR is where we should fix this.

However, I think we should seriously consider keeping an up to date toolkit with standard implementations of things in a central location and not implementing components every time from scratch in the file where we use them, so we can easily implement changes and bug fixes.

pvinis:
feel free to dedup that artistsuggestion component. but this can be a separate work for office hours :)

src/app/Components/ArtistAutosuggest/ArtistAutosuggestResults.tsx
pvinis: would you mind adding a video on the PR description of how this works?

patrinoua: done

patrinoua:
@MounirDhahri @pvinis can we merge this?

src/app/Components/ArtistAutosuggest/ArtistAutosuggestResults.tsx
```diff
@@ -88,51 +62,52 @@ const ArtistAutosuggestResultsFlatList: React.FC<{
...
-            }
-          : null
-      }
-      renderItem={({ item, index }) => {
-        return (
-          <Flex key={index} mb={1}>
-            <ArtistAutosuggestRow highlight={query} result={item} onResultPress={onResultPress} />
+  if (nodes.length > 0) {
+    return (
+      <ScrollView
```
MounirDhahri, line 67: By using the scroll view here I feel like we are not solving the issue of nested scrollviews but instead only going around it. We will though loose the virtualization benefits by using the scrollview directly and we won't be able to use this in other places since it's now only using a fixed number of artists

MounirDhahri (changes requested):
Thanks for revisiting this Angela and making the refactorings. I do think though that the most straightforward way here would be for you to recreate the component only for the consignments flow. It would be a simple view that loads for example a maximum of 5 artists and we won't need to have any logic around the scrollviews/flatlist

pvinis:
in the video, i see a warning for "same key children", have you handled that?

patrinoua:
@MounirDhahri @pvinis Thanks for the comments & suggestions! 🙏 
I've updated the PR description and the implementation to be similar to Ole's https://github.com/artsy/eigen/pull/6466/files let me know if it's ok or I should update something. 
I've also noticed that there is some weird behaviour when erasing but this should probably be resolved under a different ticket.

AlicanAkyuz:
@patrinoua is this the weird behaviour you mentioned about erasing 👇 

https://user-images.githubusercontent.com/42584148/160105509-37db9e51-ba15-4174-88fb-ed758510cf1d.mp4

patrinoua:
@AlicanAkyuz It seems like this problem also existed and still exists on the main branch before I make any changes.
Can you maybe also check, and if it's so we can make a separate ticket and address this separately? 

https://user-images.githubusercontent.com/36475005/160386896-696cf48f-358f-4baf-81f1-2152ac0b5318.mov

AlicanAkyuz:
@patrinoua yes it looks like it's on main :( 

let's aim to fix this before the next release as it's BIG imho. let me know if you have time to tackle it, otherwise I'll take care of it.

patrinoua:
@MounirDhahri can we merge this?

src/app/Components/ArtistAutosuggest/ArtistAutosuggestResults.tsx
```diff
@@ -78,65 +53,63 @@ const ArtistAutosuggestResultsFlatList: React.FC<{
...
+          flex: 1,
+          padding: space(2),
+          borderStyle: "solid",
+          borderColor: "#707070",
+          borderWidth: 1,
+          marginTop: 3,
+          height: 175,
+          overflow: "hidden",
+        }}
+      >
```
AlicanAkyuz, line 76: with these change, it is not possible for me to scroll down the suggested artists – as soon as I tap on the suggestion box in order to scroll down, the artist is selected.

AlicanAkyuz (commented):
I am not entirely familiar with the agreement about what should go into the components folder and be available to the app instead of specific flows. 

My personal opinion is that these location and artist  auto-suggest components should be inside the submission flow instead of /components, since they are quite specific to the submission flow by design and behaviour. However, if there's an agreement that they should go into components folder, I'd be curious to see them used by other sections of the app.

src/app/Components/ArtistAutosuggest/ArtistAutosuggestResults.tsx
patrinoua: Hmm it works for me, can you maybe upload a video and let's maybe have a look together

MounirDhahri:
Sorry for the late reply here, this sprint has been crazy 😄 
This looks good to me, since @AlicanAkyuz as well did the same PR inspired by your changes and fixed the focusing issue - I will merge [his PR](https://github.com/artsy/eigen/pull/6510/files) 
Thanks a lot for fixing the Angela and addressing all comments

patrinoua:
Replaced by https://github.com/artsy/eigen/pull/6510

PR #6476, opened 2022-03-22
feat: react 17 FC types

pvinis:

This PR resolves []

After #6449 and a bunch of comments there and a friend (hey @bitttttten), I realized that `@types/react` [exports](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L66) `React` types as globals, which makes our types simpler. We don't even need to `import { FC } from "react"` to keep our `FC`s! We can just change them to `React.FC` and then magically have the types globally, which is so nice.

In this PR I change all our `FC`s to `React.FC`s, same for `ReactNode`.

- [ ] I have added a feature flag, or my changes don't require a feature flag. ([How do I add one?](https://github.com/artsy/eigen/blob/main/docs/developing_a_feature.md))
- [ ] I have added an app state migration, or my changes do not require one. ([What are migrations?](https://github.com/artsy/eigen/blob/main/docs/adding_state_migrations.md))

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>


bitttttten:
👋

You can reference all types from the namespace React without needing to import them or importing "React". So React.HTMLProps, React.ReactNode, React.ComponentProps, and so on. For everything else, so the things used at runtime like Fragment, useMemo, etc. I would import.

damassi (dismissed):
I actually didn't know this! But i guess it makes sense since i've been adding these types a lot   and we already have React 17 in force, and we never import it. Lol, just never noticed i guess, but that's cool👌

damassi:
> Maybe also React.Component is clearer here? I have no preference here, so maybe we keep it consistent? My only reason for going with just Component was to avoid the React import, and just import what we actually need/use. 

No pref! if `React.Component` works globally without import might be a bit nicer. (Haven't looked at class components in a long while and had to use one the other day and totally forgot how to type props / state, etc)

pvinis:
unfortunately it requires an import 😬. so it's either import React or import just Component 🤷‍♂️

pvinis:
nice. yea I kinda like just Component too.

oh yea, I definitely will make an RFC 😁. I still think FC is a worse option, but no imports at least make it clean haha. It has not changed my mind about anything else, I very much want to trial and evaluate not using FC.

PR #6488, opened 2022-03-23
fix(CX-2433): Fix Newly added MyCollection Artwork's artist not showing in Filters

lordkiz:

This PR resolves [2433]

### Description

- When an artwork is added to MyCollection, the artist does not show in the MyCollection Filters. This is because the filters were initialized with previous artworks and are not updated. This PR fixes that.

https://user-images.githubusercontent.com/18648835/159748677-b1b703ed-ad5a-4d3b-91e3-903f481ed6cb.mp4

### To the reviewers 👀

- Fix Newly added MyCollection Artwork's artist not showing in Filters - kizito

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Scenes/MyCollection/MyCollection.tsx
+++ b/src/app/Scenes/MyCollection/MyCollection.tsx
@@ -72,7 +72,7 @@ const MyCollection: React.FC<{
 
   const artworks = extractNodes(me?.myCollectionConnection)
 
-  useLocalArtworkFilter(artworks)
+  const { reInitializeLocalArtworkFilter } = useLocalArtworkFilter(artworks)
 
   const [isRefreshing, setIsRefreshing] = useState(false)
   useEffect(() => {
@@ -170,6 +170,10 @@ const MyCollection: React.FC<{
       setJSX(null)
     }
   }, [artworks.length, filtersCount, isSearchBarVisible])
+
+  useEffect(() => {
+    reInitializeLocalArtworkFilter(artworks)
+  }, [artworks.length])
 
   return (
     <ProvideScreenTrackingWithCohesionSchema
--- a/src/app/Scenes/MyCollection/MyCollection.tsx
+++ b/src/app/Scenes/MyCollection/MyCollection.tsx
@@ -173,7 +173,7 @@ const MyCollection: React.FC<{
 
   useEffect(() => {
     reInitializeLocalArtworkFilter(artworks)
-  }, [artworks.length])
+  }, [artworks])
 
   return (
     <ProvideScreenTrackingWithCohesionSchema
--- a/src/app/Scenes/MyCollection/utils/localArtworkSortAndFilter.ts
+++ b/src/app/Scenes/MyCollection/utils/localArtworkSortAndFilter.ts
@@ -14,7 +14,7 @@ import { DateTime } from "luxon"
 import { useEffect } from "react"
 import { MyCollectionArtworkEdge } from "../MyCollection"
 
-export const useLocalArtworkFilter = (artworks: any[]) => {
+export const useLocalArtworkFilter = (artworksList?: any[] | null) => {
   const setFilterType = ArtworksFiltersStore.useStoreActions((s) => s.setFilterTypeAction)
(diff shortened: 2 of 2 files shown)


MounirDhahri (dismissed):
nice 🙌

src/app/Scenes/MyCollection/MyCollection.tsx
```diff
@@ -171,6 +171,10 @@ const MyCollection: React.FC<{
     }
   }, [artworks.length, filtersCount, isSearchBarVisible])
 
+  useEffect(() => {
+    reInitializeLocalArtworkFilter(artworks)
+  }, [artworks.length])
```
olerichter00, line 176: Just checking for the number of artworks wouldn't update the filters in a couple of cases. `artworks.length` would stay the same if an artwork gets updated.

```suggestion
  }, [artworks])
```

src/app/Scenes/MyCollection/utils/localArtworkSortAndFilter.ts
```diff
@@ -250,7 +250,14 @@ export const useLocalArtworkFilter = (artworks: any[]) => {
         },
       ])
     )
+  }
+  useEffect(() => {
+    if (artworksList) {
+      initLocalArtworkFilter(artworksList)
+    }
   }, [])
```
olerichter00, line 258: I don't think we need to return `reInitializeLocalArtworkFilter`. We could just move the effect from `MyCollection` to `useLocalArtworkFilter`. 

```suggestion
  useEffect(() => {
    if (artworksList) {
      initLocalArtworkFilter(artworksList)
    }
  }, [artworks])
```

olerichter00 (changes requested):
Looks great! I've added a suggestion to simplify the filter reinitialization.

src/app/Scenes/MyCollection/MyCollection.tsx
lordkiz: True that! Thanks.

src/app/Scenes/MyCollection/utils/localArtworkSortAndFilter.ts
lordkiz: Moving the effect to useLocalArtworkFilter will make debugging / reproducing bugs related to Local Filters really really difficult.

dariakoko (approved):
Looks good 👍

src/app/Scenes/MyCollection/utils/localArtworkSortAndFilter.ts
olerichter00: I'm not sure if I can follow. Why does it make debugging harder? I only see the benefit from having all logic related to the filters inside of the hook instead of spreading it between the hook and the component.

lordkiz: @olerichter00 Well, I should have said "...really really difficult `for me`". 😂 
1. If we move that effect, the initialization will still be done once and we'd be back to square 1. Because we will need an effect that depends on artwork for the component to rerender for any reinitialization to occur.
2. I'm always for letting components use methods/hooks however they like and whenever they like. I'd rather a component initialize with artworks (or not) or initialize following an arbitrary event or even initialize with extra objects we may not be displaying. Helps me narrow down what is happening when and where.

lordkiz: Oh @olerichter00 I see your point now. Deleting my previous comment....

lordkiz: Well, I think having components able to initialise localArtworkFilter when and how they want would always help narrow down what's happening where, how and when. I'm always against abstracting so much away from components.  Happy to change it if you insist.

olerichter00: I can't think of another case where the filters need to be updated. They completely depend on the list artworks.

olerichter00 (approved):

PR #6495, opened 2022-03-24
feat(palette): adding some header and screen component things

pvinis:
This PR resolves [MOPLAT-121]

### Description

I added some Screen structure components, some docs and examples, some stories (in the `fakestories` file, because storybook doesnt allow for outside-safe-area rendering, so i used fakestories to make that work in the app), and three screens that now use this. Here are some examples, but it will be much clearer to see this in the fakestories file.

I have added some colors to make it easier to see the components.

`Regular` (pink `Screen.Body`, useful for regular screens, with safe areas).
Screenshot 2022-04-06 at 12 13 18

`Regular with Full Background` (pink `Screen.Body`, blue `FullBackground`, useful for screens like the onboarding welcome, with a full screen image).
Screenshot 2022-04-06 at 12 14 18

`Regular with Header` (pink body, orange `Screen.Header`, this will in the future be able to move the titles on the header).
Screenshot 2022-04-06 at 12 16 29

`Regular with Floating Header` (pink body, notice the floating back button, it has support for a right button too, like a share button).
Screenshot 2022-04-06 at 12 17 34

All the above are available with scroll too.

`Scroll with Bottom View` (pink body, white `Screen.BottomView`, with support for keyboard avoiding).
Screenshot 2022-04-06 at 12 18 38
Screenshot 2022-04-06 at 12 18 50

Here are the three screens I changed to use these new components. Left is new (with `Screen`), right is old (with `Flex` etc).
You will notice they are mostly the same, and if there are tiny changes, they are fixes (like for example the text at the bottom of the welcome screen being in the safe area now, and the buttons on the same screen, now having the correct horizonal margins!).

These components make the screen code nicer, because we do less of the Flex structuring within the code, and we do that more in the start of the code. We can just use spacers and that should give us the results we expect, more consistently in our screens.

Screenshot 2022-03-24 at 19 05 07
Screenshot 2022-03-23 at 16 49 45
Screenshot 2022-04-05 at 19 28 50

- [ ] Feature flag around my changes.
- [ ] App state migration.
- [x] Screenshots/videos of my changes.
- [ ] Tested on **iOS** and **Android**.
- [x] Tests/stories for my changes.

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[MOPLAT-121]: https://artsyproduct.atlassian.net/browse/MOPLAT-121?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.secrets.baseline
```diff
@@ -490,15 +490,6 @@
         "line_number": 20
       }
     ],
-    "src/app/Scenes/Onboarding/OnboardingLogin.tsx": [
```
pvinis, line 493: removed a secret

src/app/Scenes/Onboarding/Onboarding.tsx
```diff
@@ -23,6 +22,7 @@ import { OnboardingLoginWithOTP, OTPMode } from "./OnboardingLoginWithOTP"
 import { OnboardingPersonalization } from "./OnboardingPersonalization/OnboardingPersonalization"
 import { AppleToken, GoogleOrFacebookToken, OnboardingSocialLink } from "./OnboardingSocialLink"
 import { OnboardingWelcome } from "./OnboardingWelcome"
+// import { OnboardingWelcome } from "palette/organisms/screenStructure/Screen.fakestories"
```
pvinis, line 25: this is here for easy testing and developing of `Screen` and friends.

src/app/Scenes/Onboarding/OnboardingLogin.tsx
```diff
@@ -29,157 +24,13 @@ export const loginSchema = Yup.object().shape({
...
-          </Button>
-        </Flex>
-      </ArtsyKeyboardAvoidingView>
-    </Flex>
-  )
-}
-
 const initialValues: OnboardingLoginValuesSchema = { email: "", password: "" }
 
-export const OnboardingLoginWithEmail: React.FC<OnboardingLoginProps> = ({ navigation, route }) => {
```
pvinis, line 181: no need for a wrapper just for formik 🙌

src/app/Scenes/Onboarding/OnboardingLogin.tsx
```diff
@@ -29,157 +24,13 @@ export const loginSchema = Yup.object().shape({
...
-    }
-
-    // When the user presses back on the back button immediately
-    // after opening the screen, we need to clear the timeout to avoid
-    // setting params on an unmounted screen
-    return clearTimeout(timeout)
-  }, [])
-
-  return (
-    <Flex flex={1} backgroundColor="white" flexGrow={1} paddingBottom={10}>
```
pvinis, line 74: try and look at this file with the whitespace diff thing enabled on gh. it gives a clearer diff. you can see we no longer need to bother with keyboardavoiding, with scrollviews with paddings and safe areas, it becomes a tiny bit easier, and easier to reproduce in another screen.

src/palette/atoms/BackButton/BackButton.tsx
```diff
@@ -0,0 +1,41 @@
+import { Flex } from "palette"
+import { ChevronIcon, CloseIcon } from "palette/svgs"
+import { TouchableOpacity } from "react-native"
+
+interface BackButtonProps {
+  onPress?: () => void
+  showX?: boolean
+}
+
+export const BackButton: React.FC<BackButtonProps> = ({ onPress, showX = false }) => {
```
pvinis, line 10: yes, now we have another BackButton component in our code. I did not want to replace the one we have everywhere, it can be a ticket or it will naturally happen as we start using `Screen` from palette more.

src/app/Scenes/Onboarding/OnboardingWelcome.tsx
```diff
@@ -78,87 +78,94 @@ export const OnboardingWelcome: React.FC<OnboardingWelcomeProps> = ({ navigation
   }, [navigation])
 
   return (
-    <Flex flex={1} style={{ paddingBottom: 20 }}>
```
pvinis, line 81: feels good

src/palette/atoms/Spacer/Spacer.tsx
```diff
@@ -1,14 +1,14 @@
-import React from "react"
 import { HeightProps, SpaceProps, WidthProps } from "styled-system"
 import { Box } from "../../elements/Box"
 
-export interface SpacerProps extends SpaceProps, WidthProps, HeightProps {}
+export interface SpacerProps extends SpaceProps, WidthProps, HeightProps {
+  x?: SpaceProps["ml"]
```
pvinis, line 5: finally making the spacer more clear.
`<Spacer x={2} />` > `<Spacer ml={2} />`

src/palette/organisms/screenStructure/Screen.stories.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { Screen } from "./Screen"
+
+storiesOf("palette/organisms/Screen", module)
+  .addDecorator(withTheme)
+  .add("Regular", () => (
+    <Screen>
+      <Screen.Body>
+        <Text>Storybook doesn't support outside-of-safe-area rendering.</Text>
+        <Text>
+          To test our `Screen`, use `yarn start` and navigate to `Screen.fakestories.tsx` in your
```
pvinis, line 13: i added this just as a way to guide devs to how they can develop `Screen` and friends.

src/palette/organisms/screenStructure/Screen.fakestories.tsx
```diff
@@ -0,0 +1,183 @@
+import { Button, Flex, Input, Screen, Text } from "palette"
+
+/**
```
pvinis, line 3: important :)

src/palette/organisms/screenStructure/Screen.fakestories.tsx
```diff
@@ -0,0 +1,183 @@
...
+    <Screen.Body fullwidth>
+      <Screen.BodyPadding flex={1} backgroundColor="orange">
+        <Text>Hello</Text>
+      </Screen.BodyPadding>
+    </Screen.Body>
+  </Screen>
+)
+
+const RegularHeader = () => (
+  <Screen>
```
pvinis, line 65: ugh, this code is so CLEAN!

src/palette/organisms/screenStructure/Screen.tsx
```diff
@@ -0,0 +1,248 @@
...
+import {
+  getChildByType,
+  getChildrenByType,
+  getChildrenByTypeDeep,
+  removeChildrenByType,
+} from "react-nanny"
+import { EmitterSubscription, Keyboard, ScrollView } from "react-native"
+import LinearGradient from "react-native-linear-gradient"
+import { useSafeAreaInsets } from "react-native-safe-area-context"
+
```
pvinis, line 15: ok this file is where the magic happens, ill explain a couple of things.

src/palette/organisms/screenStructure/Screen.tsx
```diff
@@ -0,0 +1,248 @@
...
+interface ScreenContextState {
```
pvinis, line 16: context. this is for communicating between `Screen` and friends.
for now we have `handleTopSafeArea` and `bottomViewHeight`. this is per screen, and its so that headers and bottom views can instruct the top level screen component for things, so that the rendering can happen correctly.

src/palette/organisms/screenStructure/Screen.tsx
```diff
@@ -0,0 +1,248 @@
...
+  handleTopSafeArea: boolean
+  bottomViewHeight: number
+}
+
+interface ScreenContextValue {
```
pvinis, line 21: and this is how the context is read and written.

src/palette/organisms/screenStructure/Screen.tsx
```diff
@@ -0,0 +1,248 @@
...
+        options,
+        setOptions: (opts) => setStateOptions((prevOpts) => ({ ...prevOpts, ...opts })),
+      }}
+    >
+      <ScreenRoot>{children}</ScreenRoot>
+    </ScreenContext.Provider>
+  )
+}
+
+const ScreenRoot = ({ children }: { children?: React.ReactNode }) => {
```
pvinis, line 52: here is the main part. this comp will do the regular screen stuff, and it will find any headers, background, and other things, and render them.

src/palette/organisms/screenStructure/Screen.tsx
```diff
@@ -0,0 +1,248 @@
...
+
+      {header}
+      {bodyChildren}
+
+      {headerFloating /* floating, so keep close to the bottom */}
+    </Flex>
+  )
+}
+
+const useUpdateScreenContext = ({ header }: { header: "none" | "regular" | "floating" }) => {
```
pvinis, line 70: here is an example of the context update. if the header is regular, then the screen should not care about top safe area, because we are guarded by the header. but if there is no header of if the header is a floating one, the screen should handle the top safe area, as there is nothing guarding us.

src/palette/organisms/screenStructure/Screen.tsx
```diff
@@ -0,0 +1,248 @@
...
+const SCREEN_HORIZONTAL_PADDING: SpacingUnit = 2
+
+interface BodyProps extends Pick<FlexProps, "backgroundColor"> {
+  children?: React.ReactNode
+  scroll?: boolean
+  nosafe?: boolean
+  fullwidth?: boolean
+}
+
+const Body = ({
```
pvinis, line 134: here is the second main part. this body comp will render the children, any bottom views, will handle safe areas, make sure all screens have our nice padding on the sides etc.

src/palette/organisms/screenStructure/Screen.tsx
```diff
@@ -0,0 +1,248 @@
...
+            </ScrollView>
+            {bottomView}
+          </ArtsyKeyboardAvoidingView>
+        </Wrap>
+      </Flex>
+    </>
+  )
+}
+
+const Background: React.FC = ({ children }) => (
```
pvinis, line 176: this is for making it possible to draw on the background, even outside safe areas.

src/palette/organisms/screenStructure/Screen.tsx
```diff
@@ -0,0 +1,248 @@
...
+  <Flex position="absolute" top={0} bottom={0} left={0} right={0}>
+    {children}
+  </Flex>
+)
+
+const BottomView: React.FC = ({ children }) => {
```
pvinis, line 182: bottom view is a nice view for having these bottom buttons that sit on top of normal content, and also take care to avoid the keyboard. it works, and even though the animation or implementation is not perfect, it is definitely a nice easy way to use these. we can improve on this later.

src/palette/organisms/screenStructure/Screen.tsx
```diff
@@ -0,0 +1,248 @@
...
+  )
+}
+
+/**
+ * Only use with `<Screen.Body fullwidth>`.
+ * It's basically an easy way to get the right padding when you also need the fullwidth body.
+ * One use case might be if you need to put an image background or something in the body,
+ * but you also need some content with the right padding.
+ */
+const BodyPadding: React.FC<FlexProps> = (props) => (
```
pvinis, line 237: a simple way to have padding without hardcoding it on screens, if we need the padding but we also want to have a fullwidth item. think of Trove in the home screen.

src/app/Scenes/Onboarding/OnboardingLogin.tsx
```diff
@@ -201,15 +52,132 @@ export const OnboardingLoginWithEmail: React.FC<OnboardingLoginProps> = ({ navig
...
+      navigation.setParams({ withFadeAnimation: false })
+    }, 1000)
+    if (route.params?.email) {
+      f.handleChange("email")(route.params.email)
+    }
+
+    // When the user presses back on the back button immediately
+    // after opening the screen, we need to clear the timeout to avoid
+    // setting params on an unmounted screen
+    return clearTimeout(timeout)
```
araujobarret, line 79: From what I remember you need to return some function signature for useEffect cleanup, something like
```jsx
return () => clearTimeout(timeout)
// or
return function myCustomClear() {
  clearTimeout(timeout)
}
```

src/app/Scenes/Onboarding/OnboardingSocialPick.tsx
```diff
@@ -89,193 +85,165 @@ export const OnboardingSocialPick: React.FC<OnboardingSocialPickProps> = ({ mode
...
 
-  const isiOS = Platform.OS === "ios"
   return (
-    <Flex justifyContent="center" flex={1} backgroundColor="white">
-      <BackButton onPress={() => navigation.goBack()} />
-      <Flex px={1.5}>
-        <Join separator={<Spacer height={60} />}>
-          <Text variant="xxl">{mode === "login" ? "Log in" : "Sign Up"}</Text>
+    <Screen>
+      <Screen.Header onBack={() => navigation.goBack()} />
```
araujobarret, line 111: For me, this pattern is a bit inconsistent 😢 with the way we create smaller composable components at Artsy, idk if we decided to move into this pattern or not, but for me, it's very straightforward to compose like:
```jsx
<Screen>
  <Header>
  </Header>
  <Body>
  </Body>
</Screen>
```
Imagine someone coming importing a component and now he needs to examine the component if has smaller blocks or not, if not he needs to grab it from another place, this is my main concern for DX but definitely not a blocker 😬

araujobarret (commented):
Good stuff looks fine but I probably have missed something 😅 

In this kind of PRs, I would avoid touching so many other lines that were not needed, which makes it hard to review but also could lead to 🐛

pvinis:
I think the only extra would be the spacer x and y 😬.
everything else feels necessary?

src/app/Scenes/Onboarding/Onboarding.tsx
pvinis: oof that was my first instinct too. I saw the code and it felt like I needed to touch more than just one file, and the rendering looked a bit messy so I didn't want to try that. I can try it to verify 😬

src/app/Scenes/Onboarding/OnboardingLogin.tsx
pvinis: oh totally. I only copied that useeffect that was there already without looking what it does.

src/app/Scenes/Onboarding/OnboardingSocialPick.tsx
pvinis: are you talking about Header vs Screen.Header?
if yes, I started with just Header but because they all communicate and depend on each other, I thought it will be better to keep them under one. 🤔 I took that pattern from libs that use many components to compose a thing. also, if it's all under one, it's easy to find. if it's all separate, how would a dev guess that there is a thing called FullBackground that they can use for drawing in the safe area 😬.

it's kinda funky but I think for this case it works. what do you think?

also if you mean something else, I totally missed it 😅

araujobarret:
> I think the only extra would be the spacer x and y 😬. everything else feels necessary?

I guess just the prettier changes maybe screamed too much and caused that feeling, no worries about it

pvinis:
oh yea.. I don't know how prettier didn't do that for anyone already, but 🤷‍♂️. I get it though.

src/app/Scenes/Onboarding/Onboarding.tsx
brainbicycle: I agree this is a bit weird, is the purpose just to workaround the above issue with rendering outside safe area? does it render okay but slightly different or is there some other failure?

brainbicycle (dismissed):
The screen component looks great 🙌 ! Very clean! One change I'd request: I agree that the Screen fakestories thing is a bit weird though, tbh I would prefer to make the Screen a little harder to test than to have the commented out code in the codebase. I may not fully understand the problem though.

pvinis:
I will give this a try tomorrow 👍

src/app/Scenes/Onboarding/OnboardingSocialPick.tsx
araujobarret: Yeah, Screen.subcomponent thing 😂 
It's a pattern used in the community, but since we don't use it anywhere that's why I said could cause confusion, I'd rather open RFC and have some React decision around that aspect than implement it right away, sounds bureaucratic but helps us to be in the same page when touching in react components 😬 

> it's kinda funky but I think for this case it works. what do you think?
indeed, it works 🚢 

One example I think, let's say we introduce a new `ScreenHipsterBerliner` element, but we want to use the same header, then I would have:
```jsx
  <ScreenHipsterBerliner>
    <Screen.Header>
    </Screen.Header>
  </ScreenHipsterBerliner>
```
that would be confusing where to import from instead of importing just the subcomponent alone, would mess up when organizing the way to composite components.

Maybe I'm just picky in that aspect but we always have good outcomes from our discussions 😄 

PS: sorry non-hipster berliners

patrinoua:
It would be nice to go together through those changes in one of the mobile meetings! Would you be up for a small presentation on the next mobile-practice meeting?

src/app/Scenes/Onboarding/OnboardingSocialPick.tsx
pvinis: I like the rfc idea just for visibility even. or maybe a topic in mobile practice. I will do that!

for the hipster example, I don't think we will or should use the subcomponents anywhere outside the base which is Screen. that is the pattern. one doesn't use Bla.Sub outside Bla. a crash or some bad behavior should be expected in that case.
and Screen is supposed to be our top level single thing that wraps all screens. kinda like that html tag is, and you have body and head etc in it. there is no htmlhipsterberliner nor will there be one 🤔. I hope that makes sense.

src/palette/organisms/screenStructure/Screen.tsx
```diff
@@ -0,0 +1,248 @@
...
+interface ScreenContextValue {
+  options: ScreenContextState
+  setOptions: (opts: Partial<ScreenContextState>) => void
+}
+
+const ScreenContext = createContext<ScreenContextValue>(null!)
+function useScreenContext() {
+  const context = useContext(ScreenContext)
+  if (!context) {
+    throw new Error(`useScreenContext must be used within a Screen`)
```
pvinis, line 30:
```suggestion
    throw new Error("useScreenContext must be used within a Screen")
```

src/palette/organisms/screenStructure/Screen.fakestories.tsx
```diff
@@ -0,0 +1,183 @@
...
+      <Text>Hello</Text>
+      <Text>Hello</Text>
+      <Text>Hello</Text>
+      <Text>Hello</Text>
+      <Text>Hello</Text>
+      <Text>Hello</Text>
+      <Text>Hello</Text>
+      <Text>Hello</Text>
+      <Text>Hello</Text>
+      <Text>Hello</Text>
```
olerichter00, line 157:
```suggestion
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>konnichiwa</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>hola</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>asalaam alaikum</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>bonjour</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>nǐn hǎo</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
```

src/palette/organisms/screenStructure/Screen.fakestories.tsx
```diff
@@ -0,0 +1,183 @@
+import { Button, Flex, Input, Screen, Text } from "palette"
```
olerichter00, line 1: Can we give this file a more descriptive name (for example `Screen.fullscreenstories.tsx`) and add why we need this file in the comment below? Looking at it I would have a hard time understanding it.

src/app/Scenes/Onboarding/OnboardingSocialPick.tsx
araujobarret: it does, was kinda sharing one concern that pops out from time to time when we design those components, I guess it's fine to live with it 👍

src/palette/organisms/screenStructure/Screen.fakestories.tsx
pvinis: i have #6597 that fixes that actually, that we can merge on this PR here and then to main 👍

pvinis:
#6597 is ready to be merged on top of this PR here, then to main.

src/palette/organisms/screenStructure/Screen.fakestories.tsx
pvinis: oh no haha i merged and it created a conflict. ill revert, but i love it :D

src/palette/organisms/screenStructure/Screen.fakestories.tsx
```diff
@@ -0,0 +1,183 @@
...
+      <Text>Hello</Text>
+      <Text>Hello</Text>
+      <Text>Hello</Text>
+      <Text>Hello</Text>
+      <Text>Hello</Text>
+      <Text>nǐn hǎo</Text>
+      <Text>Hello</Text>
+      <Text>Hello</Text>
+      <Text>Hello</Text>
+      <Text>Hello</Text>
```
pvinis, line 155:
```suggestion
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>hola</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
      <Text>Hello</Text>
```

src/palette/organisms/screenStructure/Screen.tsx
```diff
@@ -0,0 +1,248 @@
...
+interface BodyProps extends Pick<FlexProps, "backgroundColor"> {
+  children?: React.ReactNode
+  scroll?: boolean
+  nosafe?: boolean
+  fullwidth?: boolean
+}
+
+const Body = ({
+  scroll = false,
+  nosafe = false,
```
MounirDhahri, line 136: **suggestion:**
```suggestion
  insideSafeArea = false, // inSafeContext 
```

src/palette/organisms/screenStructure/Screen.tsx
```diff
@@ -0,0 +1,248 @@
...
+}: BodyProps) => {
+  const childrenExceptBottomView = removeChildrenByType(children, Screen.BottomView)
+  const bottomView = getChildrenByType(children, Screen.BottomView)
+  const { options } = useScreenContext()
+  const insets = useSafeAreaInsets()
+  const withTopSafeArea = options.handleTopSafeArea && !nosafe
+  const withBottomSafeArea = !nosafe
+
+  return (
+    <>
```
MounirDhahri, line 149:
```suggestion
```

src/palette/organisms/screenStructure/Screen.tsx
```diff
@@ -0,0 +1,248 @@
...
+      <Flex
+        flex={1}
+        mt={withTopSafeArea ? insets.top : undefined}
+        mb={withBottomSafeArea ? insets.bottom : undefined}
+        {...restFlexProps}
+      >
+        <Wrap if={scroll}>
+          <ArtsyKeyboardAvoidingView>
+            <ScrollView
+              keyboardShouldPersistTaps="handled"
```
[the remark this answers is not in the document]
MounirDhahri: no need to update anything here, I am just thinking out loud

src/palette/organisms/screenStructure/Screen.tsx
```diff
@@ -0,0 +1,248 @@
...
+            </ScrollView>
+            {bottomView}
+          </ArtsyKeyboardAvoidingView>
+        </Wrap>
+      </Flex>
+    </>
+  )
+}
+
+const Background: React.FC = ({ children }) => (
```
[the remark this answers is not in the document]
MounirDhahri: I am not sure though this is the right place for this component since it seems to me not part of the screen and instead a reusable component that we can place in `src/components` or as its own component inside palette

src/palette/organisms/screenStructure/Screen.tsx
```diff
@@ -0,0 +1,248 @@
...
+  )
+}
+
+/**
+ * Only use with `<Screen.Body fullwidth>`.
+ * It's basically an easy way to get the right padding when you also need the fullwidth body.
+ * One use case might be if you need to put an image background or something in the body,
+ * but you also need some content with the right padding.
+ */
+const BodyPadding: React.FC<FlexProps> = (props) => (
```
MounirDhahri, line 237: **thought:** I would have preferred to have this as a prop with body. like `<Body withPadding>`. The goal from that is to get people to think about extending body instead of creating a new component.

MounirDhahri: probably would even make sense to have the prop as `noPadding` since we have padding in basically all screens

MounirDhahri (commented):
Amazing work @pvinis! this is something we've been missing forever. I left some comments, most are personal preferences, please let me know what you think

src/palette/atoms/Spacer/Spacer.tsx
```diff
@@ -1,14 +1,14 @@
...
+}
 
-/**
- * A component used to inject space where it's needed
- */
-export const Spacer: React.FC<SpacerProps & { id?: string }> = (props) => {
-  return <Box {...props} />
+/** A component used to inject space where it's needed */
+export const Spacer: React.FC<SpacerProps & { id?: string }> = ({ x, y, ...props }) => {
+  return <Box {...props} ml={x ?? props.ml} mt={y ?? props.mt} />
```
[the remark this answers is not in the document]
pvinis: hm good catch

src/palette/organisms/screenStructure/Screen.tsx
```diff
@@ -0,0 +1,248 @@
...
+      <ScreenRoot>{children}</ScreenRoot>
+    </ScreenContext.Provider>
+  )
+}
+
+const ScreenRoot = ({ children }: { children?: React.ReactNode }) => {
+  const header = getChildByType(children, Screen.Header)
+  const headerFloating = getChildByType(children, Screen.FloatingHeader)
+  const background = getChildByType(children, Screen.Background)
+  const bodyChildren = getChildrenByTypeDeep(children, Screen.Body)
```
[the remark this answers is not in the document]
pvinis: this is actually getting the body children, so the name is correct. it's different than the things above.

src/palette/organisms/screenStructure/Screen.tsx
```diff
@@ -0,0 +1,248 @@
...
+  const insets = useSafeAreaInsets()
+  const withTopSafeArea = options.handleTopSafeArea && !nosafe
+  const withBottomSafeArea = !nosafe
+
+  return (
+    <>
+      <Flex
+        flex={1}
+        mt={withTopSafeArea ? insets.top : undefined}
+        mb={withBottomSafeArea ? insets.bottom : undefined}
```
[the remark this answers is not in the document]
pvinis: we still have not done the work I started that would make this safe on palette. if we end up with a weird case like that and if we haven't done the work, then we at least can fix it here by adding px

src/palette/organisms/screenStructure/Screen.tsx
pvinis: that was exactly the idea. if someone uses regular padding then Body is enough.
if they use Body with no padding, but they need padding inside for some item, we use BodyPadding rather than a random Flex.

imagine Home. it can be Body with no padding, so children will be like
<BodyPadding>
<CarrouselA/> //has padding
<CarrouselB/>
</BodyPadding>
<Trove/> // this is full width
<BodyPadding>
etc..

pvinis:
thank you for the comments!

src/palette/organisms/screenStructure/Screen.tsx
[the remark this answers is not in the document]
MounirDhahri: ah cool thanks

src/palette/organisms/screenStructure/Screen.tsx
MounirDhahri: noSafeArea seems clearer to me

src/palette/organisms/screenStructure/Screen.tsx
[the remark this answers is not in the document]
MounirDhahri: I am not sure I agree with fixing this when it arises, I will leave it to you to decide

src/palette/organisms/screenStructure/Screen.tsx
MounirDhahri: I get you now, thanks for explaining

brainbicycle (approved):
Thank you 🙏 ! Awesome work

brainbicycle:
some of the tests are failing in the modified components, not sure why 🤔

pvinis:
i can take a look at the tests and merge when green

pvinis:
i fixed a bunch, still a couple of files left, with very old relay tests. i will look into it tomorrow.

src/app/tests/renderWithWrappers.tsx
```diff
@@ -2,29 +2,29 @@ import { render } from "@testing-library/react-native"
...
-            <PopoverMessageProvider>
-              <ProvideScreenDimensions>{children}</ProvideScreenDimensions>
-            </PopoverMessageProvider>
-          </ToastProvider>
-        </Theme>
-      </GlobalStoreProvider>
-    </TrackProvider>
+const Wrappers: React.FC = ({ children }) =>
+  combineProviders(
+    [
```
patrinoua, line 17: ❤️ ❤️ ❤️

src/palette/atoms/BackButton/BackButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+      {showX ? (
+        <CloseIcon fill="black100" width={26} height={26} />
+      ) : (
+        <ChevronIcon direction="left" />
+      )}
+    </TouchableOpacity>
+  )
+}
+
+export const BackButtonWithBackground: React.FC<BackButtonProps> = ({ onPress, showX = false }) => {
```
patrinoua, line 22: I'm wondering if we could export one BackButton as we do with our Buttons in general and have a variant prop which also has "WithBackground" for example

src/palette/elements/Text/Text.stories.tsx
```diff
@@ -1,100 +1,104 @@
...
+  return (
+    <Flex flex={1}>
+      <Text>System font</Text>
+      <List>
+        <Text style={style}>regular TEXT.</Text>
+        <Text style={style}>ALL CAPS text.</Text>
+      </List>
+    </Flex>
+  )
+}
```
patrinoua, line 104: 👏🏻

src/palette/organisms/screenStructure/Screen.stories.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+export const NoHeader: ScreenStory = () => (
+  <Screen>
+    <Screen.Body>
+      <Text>Hello</Text>
+    </Screen.Body>
+  </Screen>
+)
+
+export const FloatingHeader: ScreenStory = () => (
```
patrinoua, line 76: Same comment here as with the button. Could we maybe have one Header and one ScrollScreen and pass variants instead of exporting different components?

src/palette/organisms/screenStructure/Screen.tsx
patrinoua: ✚1 insideSafeArea seems like a very descriptive prop to me

patrinoua:
Sorry for the late review and thanks for this massive effort, it looks great!

src/palette/atoms/BackButton/BackButton.tsx
pvinis: we could. I was not sure about what the best "API" is for the back button, so I kept it simple and we see how we end up using it and change it then. how does that sound?

src/palette/organisms/screenStructure/Screen.stories.tsx
pvinis: header feels like many separate ones is best. the reason is that they will behave differently, with titles scrolling up etc. so they might look similar now but they will be muuuuch easier to make them all work when they are separate components and now a huge if-spaghetti.

PR #6498, opened 2022-03-25
feat(FX-3739): Size Filter Metric

gkartalis:

This PR resolves [FX-3739]

### Description

- adds metric radio buttons on size option filter screen
- adds useLocalizedUnit hook to retrieve user preferred metric
- updated tests
- refactored saved search pill logic to useSavedSearchPill hook

When viewing pills the user will always see their user preferred metric.

https://user-images.githubusercontent.com/21178754/160404643-3445a324-5709-4734-9746-1a891b72f227.mp4

### To the reviewers 👀

- adds metric change option in size filter - gkartalis

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[FX-3739]: https://artsyproduct.atlassian.net/browse/FX-3739?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Components/ArtworkFilter/ArtworkFilterStore.tsx
+++ b/src/app/Components/ArtworkFilter/ArtworkFilterStore.tsx
@@ -1,3 +1,4 @@
+import { Metric } from "app/Scenes/Search/UserPrefsModel"
 import { assignDeep } from "app/store/persistence"
 import { Action, action, createContextStore, State } from "easy-peasy"
 import { filter, find, isEqual, unionBy } from "lodash"
@@ -24,6 +25,7 @@ export interface ArtworkFiltersModel {
   counts: FilterCounts
   sortOptions?: FilterData[]
   filterOptions?: FilterDisplayConfig[]
+  sizeMetric: Metric
 
   applyFiltersAction: Action<this>
   selectFiltersAction: Action<this, FilterData>
@@ -32,6 +34,7 @@ export interface ArtworkFiltersModel {
   setAggregationsAction: Action<this, any>
   setFiltersCountAction: Action<this, FilterCounts>
   setFilterTypeAction: Action<this, FilterType>
+  setSizeMetric: Action<this, Metric>
   setInitialFilterStateAction: Action<this, FilterArray>
   setSelectedFiltersAction: Action<this, FilterArray>
   setSortOptions: Action<this, this["sortOptions"]>
@@ -55,6 +58,7 @@ export const getArtworkFiltersModel = (): ArtworkFiltersModel => ({
     total: null,
     followedArtists: null,
   },
+  sizeMetric: "cm",
 
   /**
    * Store actions
@@ -126,6 +130,10 @@ export const getArtworkFiltersModel = (): ArtworkFiltersModel => ({
   setAggregationsAction: action((state, payload) => {
     state.aggregations = payload
     state.applyFilters = false
+  }),
+
+  setSizeMetric: action((state, payload) => {
+    state.sizeMetric = payload
   }),
 
   setFiltersCountAction: action((state, payload) => {
--- a/src/app/Components/ArtworkFilter/ArtworkFilterStore.tsx
+++ b/src/app/Components/ArtworkFilter/ArtworkFilterStore.tsx
@@ -13,6 +13,7 @@ import {
   getSortDefaultValueByFilterType,
   getUnitedSelectedAndAppliedFilters,
 } from "./ArtworkFilterHelpers"
+import { LOCALIZED_UNIT } from "./Filters/helpers"
 import { FilterDisplayConfig } from "./types"
 
 export interface ArtworkFiltersModel {
@@ -58,7 +59,7 @@ export const getArtworkFiltersModel = (): ArtworkFiltersModel => ({
     total: null,
     followedArtists: null,
   },
-  sizeMetric: "cm",
+  sizeMetric: LOCALIZED_UNIT,
 
   /**
    * Store actions
@@ -134,6 +135,7 @@ export const getArtworkFiltersModel = (): ArtworkFiltersModel => ({
 
   setSizeMetric: action((state, payload) => {
     state.sizeMetric = payload
+    state.applyFilters = false
   }),
 
   setFiltersCountAction: action((state, payload) => {
--- a/src/app/Components/ArtworkFilter/Filters/SizesOptionsScreen.tsx
+++ b/src/app/Components/ArtworkFilter/Filters/SizesOptionsScreen.tsx
@@ -1,6 +1,5 @@
 import { StackScreenProps } from "@react-navigation/stack"
 import { Metric } from "app/Scenes/Search/UserPrefsModel"
-import { useLocalizedUnit } from "app/utils/useLocalizedUnit"
 import { Box, Flex, RadioButton, Spacer, Text } from "palette"
 import React, { useEffect, useState } from "react"
 import { ArtworkFilterNavigationStack } from ".."
@@ -160,9 +159,9 @@ export const SizesOptionsScreen: React.FC<SizesOptionsScreenProps> = ({ navigati
     (state) => state.selectFiltersAction
   )
 
-  const { localizedUnit } = useLocalizedUnit()
+  const setSelectedMetric = ArtworksFiltersStore.useStoreActions((state) => state.setSizeMetric)
+  const selectedMetric = ArtworksFiltersStore.useStoreState((state) => state.sizeMetric)
 
-  const [selectedMetric, setSelectedMetric] = useState(localizedUnit)
   const sizesOptions = getSizeOptions(selectedMetric)
 
   const {
@@ -214,7 +213,7 @@ export const SizesOptionsScreen: React.FC<SizesOptionsScreenProps> = ({ navigati
     if (customSizeSelected) {
       handleCustomInputsChange(customValues)
     }
-  }, [localizedUnit, selectedMetric])
+  }, [selectedMetric])
 
   const selectCustomFiltersAction = (values: CustomSize) => {
     CUSTOM_SIZE_OPTION_KEYS.forEach((paramName) => {
--- a/src/app/Components/ArtworkFilter/Filters/SizesOptionsScreen.tsx
+++ b/src/app/Components/ArtworkFilter/Filters/SizesOptionsScreen.tsx
@@ -1,11 +1,13 @@
 import { StackScreenProps } from "@react-navigation/stack"
-import { Box, Spacer } from "palette"
-import React, { useState } from "react"
+import { Metric } from "app/Scenes/Search/UserPrefsModel"
+import { useLocalizedUnit } from "app/utils/useLocalizedUnit"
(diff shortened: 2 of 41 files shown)


src/app/Components/ArtworkFilter/Filters/SizesOptionsScreen.tsx
```diff
@@ -102,25 +110,49 @@ const CustomSizeInputsContainer: React.FC<CustomSizeInputsContainerProps> = ({
...
+          onPress={() => handleMetricChange("cm")}
+        />
+        <Text marginRight="3">cm</Text>
+        <RadioButton
+          accessibilityLabel="in"
+          accessibilityState={{ checked: isIn }}
+          selected={isIn}
+          onPress={() => handleMetricChange("in")}
+        />
+        <Text>in</Text>
```
dimatretyak, line 139: **suggestion**: how do you look at such a small refactoring?

```javascript
const metrics: Metric[] = ["cm", "in"]

<Flex flexDirection="row">
  {metrics.map(currentMetric => {
    const isSelected = metric === currentMetric
    
    return (
      <>
        <RadioButton
          key={currentMetric}
          accessibilityLabel={currentMetric}
          accessibilityState={{ checked: isSelected }}
          selected={isSelected}
          onPress={() => handleMetricChange(currentMetric)}
        />
        <Text marginRight="3">cm</Text>
      </>
    )
  })}
</Flex>
```

src/app/Components/ArtworkFilter/SavedSearch/convertersToFilterParams.ts
```diff
@@ -75,6 +76,7 @@ export const convertSizeToFilterParams = (
 
   if (Array.isArray(sizesValues)) {
     const sizeOptions = sizesValues.map((sizeValue) => {
+      const SIZES_OPTIONS = IS_USA ? USA_SIZE_OPTIONS : EUROPE_SIZE_OPTIONS
```
dimatretyak, line 79: **thought**: you can import `SIZES_OPTIONS` from `src/app/Components/ArtworkFilter/Filters/SizesOptionsScreen.tsx`

dimatretyak:
It will also be necessary to update the logic [here](https://github.com/artsy/eigen/blob/8aaa4b5766676b4981e64d4b1174e5fa67a43a23/src/app/Components/ArtworkFilter/SavedSearch/convertersToFilterParams.ts#L43-L44) (now the user will always be shown the values in `in` regardless of his chosen measurement system)

src/app/Components/ArtworkFilter/SavedSearch/convertersToFilterParams.ts
```diff
@@ -43,7 +44,7 @@ export const convertCustomSizeToFilterParamByName = (paramName: FilterParamName,
   const widthMaxLocalized = localizeDimension(max, "in")
```
dimatretyak, line 44: **suggestion**: I think we may not use `localizeDimension` (we get the value in `in` and return it `in`)

gkartalis: So you are suggesting to change that to something like:
```javascript
export const convertCustomSizeToFilterParamByName = (paramName: FilterParamName, range: string) => {
  const { min, max } = parseRange(range)

  return {
    displayText: `${min}-${max}`,
    paramValue: `${min}-${max}`,
    paramName,
  }
}
```

dimatretyak: yep

src/app/Scenes/SavedSearchAlert/pillExtractors.tests.ts
```diff
@@ -50,19 +50,23 @@ describe("extractPillFromAggregation", () => {
 
 describe("extractSizeLabel", () => {
   it("returns correcly label when full range is specified", () => {
-    expect(extractSizeLabel("w", "5-10")).toBe("w: 5-10 in")
+    expect(extractSizeLabel({ prefix: "w", value: "5-10", unit: "in" })).toBe("w: 5-10 in")
+    expect(extractSizeLabel({ prefix: "w", value: "5-10", unit: "cm" })).toBe("w: 12.7-25.4 cm")
```
dimatretyak, line 54: 👏 👏 👏

dimatretyak:
previously selected metric value is automatically overwritten (even if filters have been applied) if the metric has been updated globally. As a result, the custom values in the size filter and saved search pills may be displayed incorrectly

Not quite sure if this is a problem and more like an edge-case, but I decided to describe it

https://user-images.githubusercontent.com/3513494/160455940-0cc252e5-9c43-4529-ad50-bafbc2edd4d1.mp4

dimatretyak:
Saved search pills is displayed in the unit that the user specified globally, and not in the size filter (if we need to do this as part of a separate task, then just ignore this message)

https://user-images.githubusercontent.com/3513494/160457137-76c2e604-bc43-4027-8a00-48d7538cc788.mp4

nickskalkin (approved):
looks good to me! 👍

dimatretyak (dismissed):

src/app/utils/useLocalizedUnit.ts
```diff
@@ -0,0 +1,11 @@
+import { LOCALIZED_UNIT } from "app/Components/ArtworkFilter/Filters/helpers"
+import { GlobalStore } from "app/store/GlobalStore"
+
+export const useLocalizedUnit = () => {
+  const userPreferredMetric =
+    GlobalStore.useAppState((state) => state.userPrefs.metric) || LOCALIZED_UNIT
```
dblandin, line 6: **question / non-blocking**: Is this synced with the preferred metric stored in Gravity and accessible via Metaphysics? I know that's a new addition so fine if that's a follow-up - just wanted to clarify for myself.

dblandin (approved):
LGTM! Left one question about the user's persisted metric preference.

src/app/utils/useLocalizedUnit.ts
gkartalis: Yes! We set this [on auth](https://github.com/artsy/eigen/blob/9ed7146f348185bd649c8fb502c909195934e960/src/app/store/AuthModel.ts#L375) and it is accessible afterwards via this line `GlobalStore.useAppState((state) => state.userPrefs.metric)` throughout the app

src/app/Components/ArtworkFilter/Filters/helpers.ts
```diff
@@ -22,7 +22,7 @@ export const localizeDimension = (value: Numeric, unit: Metric) => {
     return inToCm(value)
   }
 
-  return value
+  return round(value)
```
gkartalis, line 25: @dimatretyak added this to have a rounded (precision of 2) value in both cases

gkartalis: the `inToCm` func does this by default but I added this for the inches too

dblandin: **question**: Does this only apply to the filter pill display value or would it also affect the value set into an input field?

gkartalis: It will affect the input fields only in one case: when user selects a metric for filtering with custom size that **is NOT** the preferred one, leaves the Size screen.

When the user will return in this screen the metric will be changed to the default one and the values of the inputs will be converted to this one

dimatretyak: Checked the changes, looks good. 

### Before
https://user-images.githubusercontent.com/3513494/162008560-2fdc414e-7bc7-4126-99b9-3015bf7c4ae0.mp4

### After
https://user-images.githubusercontent.com/3513494/162008687-2e6187fb-cd5d-47d2-8d1e-2ca18031ba29.mp4

dimatretyak (approved):

gkartalis:
We discussed this with @dblandin and decided that it would be better for the user to have an experience like this:

 scenario
		 - Collector's global metric preference is "cm"
		 - Collector visits artist screen > artworks grid
		 - Collector opens filter menu
		 - Collector opens size filter
		 - Collector sees metric toggle set to "cm"
		 - Collector filters by inches 50 min / 100 max width
		 - Collector clicks "Show results"
		 - Collector reopens filter menu > size filter
		 - Size filter still reflects options in inches 50 min / 100 max width
		 
For this reason I did the changes in [this commit](https://github.com/artsy/eigen/pull/6498/commits/3bf9165d88569741b20639018cebe4367f075c66).

The next commit just adds initial state to tests.

src/app/Components/ArtworkFilter/Filters/SizesOptionsScreen.tsx
```diff
@@ -160,9 +159,9 @@ export const SizesOptionsScreen: React.FC<SizesOptionsScreenProps> = ({ navigati
     (state) => state.selectFiltersAction
   )
 
-  const { localizedUnit } = useLocalizedUnit()
+  const setSelectedMetric = ArtworksFiltersStore.useStoreActions((state) => state.setSizeMetric)
+  const selectedMetric = ArtworksFiltersStore.useStoreState((state) => state.sizeMetric)
```
dimatretyak, line 163: 👏 👏

src/app/Components/ArtworkFilter/ArtworkFilterStore.tsx
```diff
@@ -55,6 +58,7 @@ export const getArtworkFiltersModel = (): ArtworkFiltersModel => ({
     total: null,
     followedArtists: null,
   },
+  sizeMetric: "cm",
```
dimatretyak, line 61: **question**: should we use `LOCALIZED_UNIT` here?

src/app/Components/ArtworkFilter/ArtworkFilterStore.tsx
```diff
@@ -128,6 +132,10 @@ export const getArtworkFiltersModel = (): ArtworkFiltersModel => ({
     state.applyFilters = false
   }),
 
+  setSizeMetric: action((state, payload) => {
+    state.sizeMetric = payload
+  }),
```
dimatretyak, line 137: **question**: should we set `applyFilters` to `false`, as it happens in other actions ([#1](https://github.com/artsy/eigen/blob/faa8c5b13eef0633adb55ca66d89cf7cc7fd6436/src/app/Components/ArtworkFilter/ArtworkFilterStore.tsx#L115), [#2](https://github.com/artsy/eigen/blob/faa8c5b13eef0633adb55ca66d89cf7cc7fd6436/src/app/Components/ArtworkFilter/ArtworkFilterStore.tsx#L119), [#3](https://github.com/artsy/eigen/blob/faa8c5b13eef0633adb55ca66d89cf7cc7fd6436/src/app/Components/ArtworkFilter/ArtworkFilterStore.tsx#L132), etc...)?

dimatretyak (approved):

dimatretyak (approved):

PR #6547, opened 2022-04-04
docs: suggestion for update PR template

pvinis:
This PR resolves []

### Description

After some chat with the moplat team, and seeing a lot of past PRs, I'm trying to simplify this template.
I saw a lot of PRs with either all of them checked, or none checked. The idea here is to maybe have a "true representation". What does this PR have? Does it have a feature flag or not? Does it have tests or not? etc. We dont expect all PRs to have migrations for example, but we expect that if a PR has a flag, the checkbox would be checked. Filling this in should be quick and easy and clear. (Maybe in the future we will have some script doing this for us. Not yet though.)

I have simplified the checkboxes, and removed some. Maybe we can remove more?

I have not changed the changelog part of this template, until we work on the changelog scripts, but I think we can definitely simplify.

- [ ] Feature flag around my changes.
- [ ] App state migration ⚠️ (If you touched any file with a `*Model` or `*Store` in it, you might need a migration, ask moplat if unsure)!
- [ ] Screenshots/videos of my changes.
- [ ] Tested on **iOS** and **Android**.
- [ ] Tests/stories for my changes.

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/docs/pull_request_template.md
+++ b/docs/pull_request_template.md
@@ -2,21 +2,7 @@
      `type(thing): what changed`
 like:
      `feat(profile): allow editing of profile photo`.
-Here are some types:
-     hotfix: If an urgent fix is required
-     feat:   A new feature
-     fix:    A bug fix
-     docs:   Documentation only changes
-     perf:   A change that improves performance
-     test:   Adding missing tests or correcting existing tests
-     format: Changes that do not affect the meaning of the code (white-space, formatting, linting, etc)
-     revert: Revert something
-     update/upgrade: Update dependencies or pods or other related things
-     refactor: A change that keeps functionality mostly the same
-     build:    Changes that affect the build system or external dependencies (example scopes: pods, fastlane, yarn, etc)
-     uxui:     Changes that affect how something looks or how the user experiences it
-     cleanup:  When the main change is removing code, old or unused
-     other/chore: Catch-all, and once we have had a bunch of those, we can possibly make a new category
+Find some examples in the "conventional commits" summary.
 -->
 
 <!-- Jira ticket in square brackets like [PROJECT-XXXX] -->
--- a/docs/pull_request_template.md
+++ b/docs/pull_request_template.md
@@ -1,31 +1,43 @@
-The type of this PR is: **TYPE**
+<!-- Use a PR title in the form of 
+     `type(thing): what changed`
+like:
+     `feat(profile): allow editing of profile photo`.
+Here are some types:
+     hotfix: If an urgent fix is required
+     feat:   A new feature
+     fix:    A bug fix
+     docs:   Documentation only changes
+     perf:   A change that improves performance
+     test:   Adding missing tests or correcting existing tests
+     format: Changes that do not affect the meaning of the code (white-space, formatting, linting, etc)
+     revert: Revert something
+     update/upgrade: Update dependencies or pods or other related things
+     refactor: A change that keeps functionality mostly the same
+     build:    Changes that affect the build system or external dependencies (example scopes: pods, fastlane, yarn, etc)
+     uxui:     Changes that affect how something looks or how the user experiences it
+     cleanup:  When the main change is removing code, old or unused
+     other/chore: Catch-all, and once we have had a bunch of those, we can possibly make a new category
+-->
 
-<!-- Bugfix/Feature/Enhancement/Documentation -->
-
-<!-- If applicable, write the Jira ticket number in square brackets e.g. [PROJECT-XXXX]
-     The Jira integration will turn it into a clickable link for you. -->
-
+<!-- Jira ticket in square brackets like [PROJECT-XXXX] -->
 This PR resolves []
 
 ### Description
 
-<!-- Implementation description -->
+<!-- Some info, Implementation, Before/After shots, follow-up work, etc -->
 
 ### PR Checklist (tick all before merging)
 
-<!-- 💡 This checklist is experimental. CX warmly welcomes any feedback about the list or how it impacts your workflow -->
-
-- [ ] I have included screenshots or videos to illustrate my changes, or I have not changed anything that impacts the UI.
-- [ ] I have tested my changes on **iOS** and **Android**.
-- [ ] I have added tests/stories for my changes, or my changes don't require testing/stories, or I have included a link to a separate Jira ticket covering the tests.
-- [ ] I have added a feature flag, or my changes don't require a feature flag. ([How do I add one?](https://github.com/artsy/eigen/blob/main/docs/developing_a_feature.md))
-- [ ] I have documented any follow-up work that this PR will require, or it does not require any.
-- [ ] I have added an app state migration, or my changes do not require one. ([What are migrations?](https://github.com/artsy/eigen/blob/main/docs/adding_state_migrations.md))
-- [ ] I have added a changelog entry below or my changes do not require one.
+<!-- 💡 This checklist is experimental. MOPLAT warmly welcomes any feedback about the list or how it impacts your workflow. -->
+- [ ] Feature flag around my changes.
+- [ ] App state migration.
+- [ ] Screenshots/videos of my changes.
+- [ ] Tested on **iOS** and **Android**.
+- [ ] Tests/stories for my changes.
 
 ### To the reviewers 👀
 
-- [ ] I would like at least one of the reviewers to run this PR on the simulator or device.
+- [ ] I would like **at least one** of the reviewers to run this PR on the simulator or device.
 
 <details><summary>Changelog updates</summary>
 


docs/pull_request_template.md
```diff
@@ -1,31 +1,43 @@
...
+     test:   Adding missing tests or correcting existing tests
+     format: Changes that do not affect the meaning of the code (white-space, formatting, linting, etc)
+     revert: Revert something
+     update/upgrade: Update dependencies or pods or other related things
+     refactor: A change that keeps functionality mostly the same
+     build:    Changes that affect the build system or external dependencies (example scopes: pods, fastlane, yarn, etc)
+     uxui:     Changes that affect how something looks or how the user experiences it
+     cleanup:  When the main change is removing code, old or unused
+     other/chore: Catch-all, and once we have had a bunch of those, we can possibly make a new category
+-->
```
gkartalis, line 20: 😍

docs/pull_request_template.md
```diff
@@ -1,31 +1,43 @@
...
 
-<!-- 💡 This checklist is experimental. CX warmly welcomes any feedback about the list or how it impacts your workflow -->
-
-- [ ] I have included screenshots or videos to illustrate my changes, or I have not changed anything that impacts the UI.
-- [ ] I have tested my changes on **iOS** and **Android**.
-- [ ] I have added tests/stories for my changes, or my changes don't require testing/stories, or I have included a link to a separate Jira ticket covering the tests.
-- [ ] I have added a feature flag, or my changes don't require a feature flag. ([How do I add one?](https://github.com/artsy/eigen/blob/main/docs/developing_a_feature.md))
-- [ ] I have documented any follow-up work that this PR will require, or it does not require any.
-- [ ] I have added an app state migration, or my changes do not require one. ([What are migrations?](https://github.com/artsy/eigen/blob/main/docs/adding_state_migrations.md))
-- [ ] I have added a changelog entry below or my changes do not require one.
```
gkartalis, line 24: this sometimes reminded me that I need to add the changelog entry 🤔 although I think it is okay to go

gkartalis (approved):
thanks @pvinis

docs/pull_request_template.md
admbtlr: I don't like this, predictably. Why should Eigen be different to everything else at Artsy? What's wrong with standards?

pvinis: its true. i guess i had both reminders "try the new list" and after last retro "clean up PR template", so i thought ill combine.

we could try this new list still on eigen, or we can clean up the pr and try the list later?

docs/pull_request_template.md
pvinis: i thought since its right underneath, it would be ok? possibly the changelog script will change and could grab things from the title of the PR or the description, so i thought we can remove it here and handle it in the scripts in a nice way 🤔

docs/pull_request_template.md
admbtlr: What would "trying it" look like? What problem would you be trying to solve? What metric would you use to decide whether the trial had been successful or not?

My guess is that the trial would just go on forever because there was no real purpose to it, everyone would get confused because Eigen is not like the other repos, and we end up in a situation where everyone just does what they want.

pvinis: i will change these and make a more detailed PR with what we would be trying, how to evaluate etc.

patrinoua:
Much cleaner, thanks! 🙏

docs/pull_request_template.md
```diff
@@ -1,31 +1,29 @@
...
+<!-- 💡 This checklist is experimental. MOPLAT warmly welcomes any feedback about the list or how it impacts your workflow. -->
+- [ ] Feature flag around my changes.
```
MounirDhahri, line 18: **suggestion:** I prefer the old format of saying "I have my changes hidden behind a feature flag"

docs/pull_request_template.md
```diff
@@ -1,31 +1,29 @@
...
+- [ ] App state migration.
```
MounirDhahri, line 19: **suggestion:** same here: "I added a state app migration ([What are migrations?](https://github.com/artsy/eigen/blob/main/docs/adding_state_migrations.md))"

docs/pull_request_template.md
```diff
@@ -1,31 +1,29 @@
...
+- [ ] Screenshots/videos of my changes.
+- [ ] Tested on **iOS** and **Android**.
+- [ ] Tests/stories for my changes.
```
MounirDhahri, line 22: same comment here

MounirDhahri (commented):
Nice, these changes make sense to me

pvinis:
@MounirDhahri i could change it back. The reason I made these much smaller is to help devs read them and check or leave unchecked. I feel like many devs dont read when its 2 lines of text, but like 5 words are ok. I do agree that the phrasing was better before though.

🤔 what do you think?

MounirDhahri:
Even if it's a bit longer than 2 words, I think people will read it if we announce it in the #practice-mobile channel. My main thinking here is to let open-source devs who read our PRs to understand easily our check list. Nonetheless they're also clear now, it's just a personal preference

docs/pull_request_template.md
MounirDhahri: 14608107_1180665285312703_1558693314_n

MounirDhahri: Great idea @brainbicycle

MounirDhahri: cc @patrinoua

docs/pull_request_template.md
patrinoua: @MounirDhahri @brainbicycle thank you for having a closer look into this. 🙏

Since this is something that a lot of people read multiple times a day, and given the fact that the old format was quite lengthly (I personally admit I didn't read it every time), I think what @pvinis tried to do and I added to it, is to make it as minimal as possible, by eliminating things that are easily implied, and keep it as short as possible.

In "I added a state migration to my changes if necessary" the parts "I added", "to my changes" and "if necessary"
are easily implied so we have minimised those to be simple reminders of the things you need to make sure to check.

We did not approach the PR template as a documentation so did not include links such as [What are migrations?](https://github.com/artsy/eigen/blob/main/docs/adding_state_migrations.md) but I really like the idea of providing a link to our docs on the bottom of the PR template. 

It could be something like this

"Need help with some of the above? Have a look on our [docs](https://github.com/artsy/eigen/blob/main/docs/README.md)! " 

This way anyone can find all the information about feature flags (what are they and how we use them, migrations etc.)

What do you think?

patrinoua: Just to make sure we are all on the same page, we have also merged this in the mean time https://github.com/artsy/eigen/pull/6567

patrinoua: And I've just added this https://github.com/artsy/eigen/pull/6576 which adds the links to the docs.

docs/pull_request_template.md
patrinoua: Thanks @MounirDhahri , I've answered in more detail on a thread above, however to also comment here my reasoning against @brainbicycle suggestion is :
- [x] There are too many words instead of a brief reminder text that's easy to read.
- [x] It's ordered by importance not by what occurs more often in development. (eg I don't always need to add a feature flag, and while there can be 20 prs for a feature I only need to add a FF once.)

I've also created [this](https://github.com/artsy/eigen/pull/6576) which provides

- [x] Brief reminder text that's easy to read.
- [x] Order by what occurs more often in development.
- [x] **Provide Link to the documentation**

Looking forward to your thoughts 🙏

docs/pull_request_template.md
MounirDhahri: thanks 🙏

docs/pull_request_template.md
MounirDhahri: @patrinoua @pvinis it's fine for me if we do go with the short version. I do though believe that we are making things simpler than possible only for the sake of keeping the template short - I think that all engineers actually already know the checklist by heart now since it's in every PR and people must have already read it the first time they created a PR in Eigen. We read docs all day, it's part of our job, and providing full context as an open source company shows empathy and helps us give a good impression to people interesting in joining us and checking our PRs. This is of course a **biased personal preference** but I am 100% glad to stick with whatever the team thinks

brainbicycle: I agree with ordering based on what is common but what I was more suggesting is that the meaning of the checkboxes should be clear to an outsider and a new engineer and the new short versions in my opinion aren't. I don't really understand what you mean about `easily implied words` if I come in as a new engineer and see a checkbox `- [ ] feature flag` it means much less to me than `- [ ] I have added a feature flag`. Adding a link to the readme I think is a good add but that leaves new engineers to go through our documentation rather than just adding a couple words in the template? It seems like we are favoring brevity over context which I don't agree with.

brainbicycle: you could leave off the `my changes` part I will give you that 😆

MounirDhahri: > `- [ ] feature flag` it means much less to me than `- [ ] I have added a feature flag`

this is actually what we want to not have. the checkbox is actually there to mention that `the changes are hidden behind a feature flag` and not about adding a feature flag. we can be explicit here and keep the changes as @pvinis had them maybe `my changes around a feature flag`

PR #6558, opened 2022-04-05
fix: update refetch logic for Sale screen

dimatretyak:
Similar PR: #6534 

### Description
At the moment, refetch logic doesn't work correctly and for this reason the user cannot get up-to-date info about auction on Sale screen using Pull-to-refresh pattern. This PR includes fixes and improvements for it. 

#### refetch logic consists of:
* getting up-to-date info about auction ([above query](https://github.com/artsy/eigen/blob/1ab3429053e97c0ca20d72bdfa2536baf74f19a2/src/app/Scenes/Sale/Sale.tsx#L384-L387))
* getting up-to-date info about sale artworks ([below query](https://github.com/artsy/eigen/blob/1ab3429053e97c0ca20d72bdfa2536baf74f19a2/src/app/Scenes/Sale/Sale.tsx#L388-L405)), taking into account the selected and applied filters

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- Update refetch logic for Sale screen - dimatretyak

</details>

### DIFF
--- a/src/app/Components/ArtworkFilter/useArtworkFilters.ts
+++ b/src/app/Components/ArtworkFilter/useArtworkFilters.ts
@@ -1,5 +1,6 @@
 import { PAGE_SIZE } from "app/Components/constants"
-import { useEffect, useMemo } from "react"
+import { EventEmitter } from "events"
+import { useEffect, useMemo, useRef } from "react"
 import { RelayPaginationProp, Variables } from "react-relay"
 import {
   aggregationForFilter,
@@ -9,6 +10,13 @@ import {
   prepareFilterArtworksParamsForInput,
 } from "./ArtworkFilterHelpers"
 import { ArtworksFiltersStore, selectedOptionsUnion } from "./ArtworkFilterStore"
+
+const ArtworksFiltersEvents = new EventEmitter()
+const REFRESH_KEY = "refresh"
+
+export function refreshArtworks() {
+  ArtworksFiltersEvents.emit(REFRESH_KEY)
+}
 
 interface UseArtworkFiltersOptions {
   relay?: RelayPaginationProp
@@ -67,6 +75,9 @@ export const useArtworkFilters = ({
     }
   }
 
+  const refetchRef = useRef(refetch)
+  refetchRef.current = refetch
+
   useEffect(() => {
     if (applyFilters) {
       refetch()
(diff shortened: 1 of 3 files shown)


src/app/Components/ArtworkFilter/useArtworkFilters.ts
```diff
@@ -64,11 +72,33 @@ export const useArtworkFilters = ({
...
+    const callback = () => {
+      refetchRef.current()
+    }
+
+    ArtworksFiltersEvents.addListener(REFRESH_KEY, callback)
+
+    return () => {
+      ArtworksFiltersEvents.removeListener(REFRESH_KEY, callback)
+    }
+  }, [])
```
dimatretyak, line 101: There was also another solution to pass refetch method to parent component (see example below). But it seems to me that the solution using `EventEmitter` is more readable and easy to understand. What do you think? If you have other ideas and suggestions, I am happy to listen to them

```javascript
const ParentComponent = () => {
  const [refetch, setRefetch] = useState<() => void | null>(null)

  useArtworkFilters({
    relay,
    aggregations,
    ...
    setRefetch, // new field
  })

  ...
}

...

const useArtworkFilters = ({
  relay,
  aggregations,
  setRefetch,
  ...
}) => {
  ...

  const refetch = useCallback(() => {
    // some logic
  }, [appliedFilters, refetchVariables, ...])

  useEffect(() => {
    setRefetch(refetch)
  }, [refetch])

  return ...
}
```

MounirDhahri (commented):
Thanks for this nice PR Dima, I left a small comment please let me know if there is something that wasn't clear

src/app/Components/ArtworkFilter/useArtworkFilters.ts
dimatretyak: > By adding an event listener to the useArtworksFilter we could be ending up with different screens listening to the same key in case one user has more than one sale screen open

good catch!

dimatretyak: > even better I would probably go with the approach that you described above which looks pretty solid for me

Slightly updated the logic of this approach. We need to pass [refetchRef](https://github.com/artsy/eigen/blob/bccc9db70dbb182dcdceb4b84390593e82d52b71/src/app/Scenes/Sale/Components/SaleLotsList.tsx#L106) instead of `setRefetch` and keeping `callback` function in state (see code and comments below). 

```javascript
const ParentComponent = () => {
  const [refetch, setRefetch] = useState<() => void | null>(null)

  useArtworkFilters({
    relay,
    aggregations,
    ...
    setRefetch: (callback) => {
      /**
       * if we just pass callback to setRefetch, callback function will be called immediately
       *
       * This happens because setState supports the function updater to get a new state based on the old one.
       * (e.g, for our case setRefetch(prevState => {}))
       *
       * It seems to me that this behavior is not entirely obvious and would confuse other people
       */
      setRefetch(() => callback)

      // or we would have to store callback function in ref
      // so I decided it would be better to pass refetchRef right away
      refetchRef.current = callback
    }
  })


  ...
}
```

MounirDhahri: really nice approach 👏

MounirDhahri (approved):
Nice PR 🎉 LGTM

dimatretyak:
> Nice PR 🎉 LGTM

thank you @MounirDhahri

PR #6568, opened 2022-04-06
feat: Explanatory banner for empty Collector Profile 

rajsam003:
This PR resolves [CX-2473]

### Description

The PR added an explanatory banner at the top of the Collector Profile screen. It will appear for all the users, new and existing, as long as they have an empty profile and haven’t added any information.

image

- [x] Feature flag around my changes.
- [ ] App state migration.
- [x] Screenshots/videos of my changes.
- [x] Tested on **iOS** and **Android**.
- [x] Tests/stories for my changes.

### To the reviewers 👀

- Explanatory banner for empty Collector Profile - sam

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-2473]: https://artsyproduct.atlassian.net/browse/CX-2473?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/MyCollection/MyCollection.tsx
+++ b/src/app/Scenes/MyCollection/MyCollection.tsx
@@ -104,7 +104,7 @@ const MyCollection: React.FC<{
 
   const hasBeenShownBanner = async () => {
     const hasSeen = await AsyncStorage.getItem(HAS_SEEN_MY_COLLECTION_NEW_WORKS_BANNER)
-    const shouldShowConsignments = showSessionVisualClue("ArtworkSubmissionBanner")
+    const shouldShowConsignments = showSessionVisualClue("ArtworkSubmissionMessage")
     return {
       hasSeenBanner: hasSeen === "true",
       shouldShowConsignments: shouldShowConsignments === true,
@@ -164,7 +164,7 @@ const MyCollection: React.FC<{
                 title="Artwork added to My Collection"
                 text="The artwork you submitted for sale has been automatically added."
                 showCloseButton
-                onClose={() => removeClue("ArtworkSubmissionBanner")}
+                onClose={() => removeClue("ArtworkSubmissionMessage")}
               />
             )}
           </Flex>
--- a/src/app/Scenes/MyProfile/MyProfileHeaderMyCollectionAndSavedWorks.tsx
+++ b/src/app/Scenes/MyProfile/MyProfileHeaderMyCollectionAndSavedWorks.tsx
@@ -7,12 +7,13 @@ import { FancyModalHeader } from "app/Components/FancyModal/FancyModalHeader"
 import { StickyTabPage } from "app/Components/StickyTabPage/StickyTabPage"
 import { navigate } from "app/navigation/navigate"
 import { defaultEnvironment } from "app/relay/createEnvironment"
-import { useFeatureFlag } from "app/store/GlobalStore"
+import { setVisualClueAsSeen, useFeatureFlag, useVisualClue } from "app/store/GlobalStore"
 import { renderWithPlaceholder } from "app/utils/renderWithPlaceholder"
 import { ProvideScreenTrackingWithCohesionSchema } from "app/utils/track"
 import { screen } from "app/utils/track/helpers"
 import {
   Avatar,
+  Banner,
   Box,
   BriefcaseIcon,
   Button,
@@ -24,7 +25,7 @@ import {
   Text,
   useColor,
 } from "palette"
-import React, { useContext } from "react"
+import React, { useContext, useEffect, useState } from "react"
 import { createRefetchContainer, QueryRenderer } from "react-relay"
 import { graphql } from "relay-runtime"
 import { FavoriteArtworksQueryRenderer } from "../Favorites/FavoriteArtworks"
@@ -71,10 +72,29 @@ export const MyProfileHeader: React.FC<{ me: MyProfileHeaderMyCollectionAndSaved
   const navigation = useNavigation()
 
   const showCollectorProfile = useFeatureFlag("AREnableCollectorProfile")
+  const showCollectorProfileExplanatoryBannerFeatureFlag = useFeatureFlag(
+    "ARShowCollectorProfileExplanatoryBanner"
+  )
+  const { showVisualClue } = useVisualClue()
+
+  const [showCollectorProfileExplanatoryBanner, setShowCollectorProfileExplanatoryBanner] =
+    useState(showVisualClue("CollectorProfileExplanatoryBanner"))
 
   const { localImage } = useContext(MyProfileContext)
 
   const userProfileImagePath = localImage || me?.icon?.url
+
+  const collectorProfileIsNotComplete =
+    !me.bio || !me.icon || !me.profession || !me.otherRelevantPositions
+
+  useEffect(() => {
+    setVisualClueAsSeen("CollectorProfileExplanatoryBanner")
+  }, [])
+
+  const showBanner =
+    showCollectorProfileExplanatoryBannerFeatureFlag &&
+    collectorProfileIsNotComplete &&
+    showCollectorProfileExplanatoryBanner
 
   return (
     <>
@@ -85,6 +105,18 @@ export const MyProfileHeader: React.FC<{ me: MyProfileHeaderMyCollectionAndSaved
           navigate("/my-profile/settings")
         }}
       />
+      {showBanner && (
+        <Banner
+          title="Why complete your Colletor Profile?"
+          text="A complete profile helps you build a relationship with sellers. Select “Edit Profile” to see which details are shared when you contact sellers."
+          showCloseButton
+          containerStyle={{ mb: 2, backgroundColor: color("black5") }}
+          titleStyle={{ variant: "xs", color: color("black100") }}
+          bodyTextStyle={{ variant: "xs", color: color("black60") }}
+          onClose={() => setShowCollectorProfileExplanatoryBanner(false)}
+        />
+      )}
+
       <Flex flexDirection="row" alignItems="center" px={2}>
         <Box
           height="99"
--- a/src/app/Scenes/MyProfile/MyProfileHeaderMyCollectionAndSavedWorks.tsx
+++ b/src/app/Scenes/MyProfile/MyProfileHeaderMyCollectionAndSavedWorks.tsx
@@ -13,13 +13,13 @@ import { ProvideScreenTrackingWithCohesionSchema } from "app/utils/track"
 import { screen } from "app/utils/track/helpers"
 import {
   Avatar,
-  Banner,
   Box,
   BriefcaseIcon,
   Button,
   Flex,
   Join,
   MapPinIcon,
+  Message,
   MuseumIcon,
   Spacer,
   Text,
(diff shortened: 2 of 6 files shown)


src/app/store/config/visualClues.ts
```diff
@@ -21,4 +21,7 @@ export const visualClues = defineVisualClues({
   ArtworkSubmissionBanner: {
     description: "The banner shown after artwork submission with SWA flow",
   },
+  CollectorProfileExplanatoryBanner: {
+    description: "The banner shown if the collector profile is empty",
+  },
```
dariakoko, line 26: I think this is redundant

dariakoko (dismissed):
Looks good to me!

I left one comment, not blocking, just a cleanup

src/app/Scenes/MyProfile/MyProfileHeaderMyCollectionAndSavedWorks.tsx
```diff
@@ -63,19 +65,35 @@ export const MyProfileHeaderMyCollectionAndSavedWorks: React.FC<{
...
+    useState(false)
 
   const showCollectorProfile = useFeatureFlag("AREnableCollectorProfile")
 
   const { localImage } = useContext(MyProfileContext)
 
   const userProfileImagePath = localImage || me?.icon?.url
 
+  const hasSeenTheBanner = async () =>
+    await AsyncStorage.getItem(HAS_SEEN_COLLECTOR_PROFILE_EXPLANATORY_BANNER)
```
olerichter00, line 86: You can use [useVisualClue](https://github.com/artsy/eigen/blob/3b5868e769c0ab83d02092b82b50bc0738196cb8/src/app/store/GlobalStore.tsx#L170) instead.

src/app/Scenes/MyProfile/MyProfileHeaderMyCollectionAndSavedWorks.tsx
```diff
@@ -63,19 +65,35 @@ export const MyProfileHeaderMyCollectionAndSavedWorks: React.FC<{
...
+
+  const collectorProfileIsEmpty =
+    !me.bio && !me.icon && !me.profession && !me.otherRelevantPositions
```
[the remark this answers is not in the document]
olerichter00: Is this intended? I would expect to show the banner always or if at least one of the fields is empty.

src/app/Scenes/MyProfile/MyProfileHeaderMyCollectionAndSavedWorks.tsx
```diff
@@ -85,6 +103,20 @@ export const MyProfileHeader: React.FC<{ me: MyProfileHeaderMyCollectionAndSaved
           navigate("/my-profile/settings")
         }}
       />
+      {showCollectorProfile && showCollectorProfileExplanatoryBanner && (
```
olerichter00, line 106: Can we introduce a new feature flag just for this feature?

src/app/Scenes/MyProfile/MyProfileHeaderMyCollectionAndSavedWorks.tsx
```diff
@@ -85,6 +103,20 @@ export const MyProfileHeader: React.FC<{ me: MyProfileHeaderMyCollectionAndSaved
...
+        <Banner
+          title="Why complete your Colletor Profile?"
+          text="A complete profile helps you build a relationship with sellers. Select “Edit Profile” to see which details are shared when you contact sellers."
+          showCloseButton
+          containerStyle={{ mb: 2, backgroundColor: color("black5") }}
+          titleStyle={{ variant: "xs", color: color("black100") }}
+          bodyTextStyle={{ variant: "xs", color: color("black60") }}
```
olerichter00, line 113: Can we get rid of this? I would expect banners to look the same everywhere in the app.

src/app/Scenes/MyCollection/MyCollection.tests.tsx
```diff
@@ -91,9 +91,9 @@ describe("MyCollection", () => {
     })
 
     it("shows zerostate", () => {
-      expect(extractText(tree.root)).toContain("Your art collection in your pocket.")
+      expect(extractText(tree.root)).toContain("Primed and ready for artworks.")
       expect(extractText(tree.root)).toContain(
-        "Keep track of your collection all in one place and get market insights"
+        "Add works from your collection to access price and market insights."
```
olerichter00, line 96: Are these changes intended? 👀

src/app/Components/States/ZeroState.tsx
```diff
@@ -30,6 +37,7 @@ export const ZeroState = (props: ZeroStateProps) => {
               size="3"
               maxWidth={props.title ? "100%" : "80%"}
               lineHeight="20"
+              textAlign={props.alignItems ? "center" : "left"}
```
olerichter00, line 40: Are these changes intended? 👀

olerichter00 (changes requested):
Looks good so far! I've added a couple of comments and questions.

src/app/Scenes/MyProfile/MyProfileHeaderMyCollectionAndSavedWorks.tsx
rajsam003: It doesn't work for me ☹️. 
Do you have an example?

src/app/Scenes/MyProfile/MyProfileHeaderMyCollectionAndSavedWorks.tsx
rajsam003: Sure ok

src/app/Components/States/ZeroState.tsx
rajsam003: https://www.figma.com/file/l3plkt7gMSg4p1Or0E8SLC/My-Collection-2022?node-id=719%3A7259

src/app/Scenes/MyProfile/MyProfileHeaderMyCollectionAndSavedWorks.tsx
rajsam003: https://www.figma.com/file/l3plkt7gMSg4p1Or0E8SLC/My-Collection-2022?node-id=719%3A7259

rajsam003:
@olerichter00 Most of your comment is about the design. Let's bring this to the channel and see what Pierluca says. 
But anyway thanks for the review :)

src/app/Scenes/MyProfile/MyProfileHeaderMyCollectionAndSavedWorks.tsx
olerichter00: You could search for `useVisualClue` in the code base for examples.

src/app/Scenes/MyProfile/MyProfileHeaderMyCollectionAndSavedWorks.tsx
olerichter00: Is there already a conversation about it? I can only find discussions about the copy in the Figma file. 

If you like I can start the discussion about it in Slack :)

src/app/Scenes/MyProfile/MyProfileHeaderMyCollectionAndSavedWorks.tsx
olerichter00: This solution should work: https://gist.github.com/olerichter00/fa08c32849d873253181a7147847a110

src/app/Scenes/MyProfile/MyProfileHeaderMyCollectionAndSavedWorks.tsx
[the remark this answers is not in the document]
olerichter00: Ideally, we would set the banner as seen at the end of the session or mark the banner in the session and then set it as seen at the beginning of the next session. This current solution is probably working as expected though.

But, even that wouldn't guarantee that the user has noticed the banner in the case the user opens the Profile screen and doesn't stay long enough to check the banner. What do you think? Should we raise that concern?

src/app/Scenes/MyProfile/MyProfileHeaderMyCollectionAndSavedWorks.tsx
olerichter00: Ideally, we would set the banner as seen at the end of the session or mark the banner in the session and then set it as seen at the beginning of the next session. This current solution is probably working as expected though.

But, even that wouldn't guarantee that the user has noticed the banner in the case the user opens the Profile screen and doesn't stay long enough to check the banner. What do you think? Should we raise that concern?

src/app/store/config/visualClues.ts
dariakoko: Since Ole is suggesting to use visualClues, we should keep these changes

dariakoko:
ping

rajsam003:
Based on Pierluca's comment. I have updated the code
https://artsy.slack.com/archives/C01B2P6LJUU/p1649431070486749?thread_ts=1649260836.908909&cid=C01B2P6LJUU

src/app/Scenes/MyProfile/MyProfileHeaderMyCollectionAndSavedWorks.tsx
```diff
@@ -71,11 +72,30 @@ export const MyProfileHeader: React.FC<{ me: MyProfileHeaderMyCollectionAndSaved
...
+  const { showVisualClue } = useVisualClue()
+
+  const [showCollectorProfileExplanatoryBanner, setShowCollectorProfileExplanatoryBanner] =
+    useState(showVisualClue("CollectorProfileExplanatoryBanner"))
 
   const { localImage } = useContext(MyProfileContext)
 
   const userProfileImagePath = localImage || me?.icon?.url
 
+  const collectorProfileIsEmpty =
```
olerichter00, line 87:
```suggestion
  const collectorProfileIsNotComplete =
```

src/app/Scenes/MyProfile/MyProfileHeaderMyCollectionAndSavedWorks.tsx
olerichter00: Can we do this in another way? I don't think we want to have banners styled when there are used. Instead I'd expect to pass a `variant` prop or have different versions of `Banner`.

olerichter00 (commented):
Looks good. Just two small comments :)

rajsam003:
> Can we do this in another way? I don't think we want to have banners styled when there are used. Instead, I'd expect to pass a variant prop or have different versions of Banner.

@olerichter00 I agree but I think we can style the banner however we want based on the design. Thats why we have `containerStyle`, `titleStyle` and `bodyTextStyle` as props. 
And I dont support creating a new banner when we have one already. 
Shall we bring this to the channel and see how others feel about it?

olerichter00:
> > Can we do this in another way? I don't think we want to have banners styled when there are used. Instead, I'd expect to pass a variant prop or have different versions of Banner.
> 
> @olerichter00 I agree but I think we can style the banner however we want based on the design. Thats why we have `containerStyle`, `titleStyle` and `bodyTextStyle` as props. And I dont support creating a new banner when we have one already having props passed. Shall we bring this to the channel and see how others feel about it?

Yes, let's bring this discussion to Slack 👍 

Ideally, we should always prevent styling components from `Palette` when possible to have a consistent look throughout the app.

olerichter00:
> And I dont support creating a new banner when we have one already having props passed.

I mean to either use variants and pass it as a prop or create wrapper components like `DarkBanner` and `LightBanner`.

olerichter00:
@rajsam003 I took over the ticket and will continue the work in this PR.

rajsam003:
> @rajsam003 I took over the ticket and will continue the work in this PR.

Thank you @olerichter00

src/app/Scenes/MyCollection/MyCollection.tsx
```diff
@@ -104,7 +104,7 @@ const MyCollection: React.FC<{
 
   const hasBeenShownBanner = async () => {
     const hasSeen = await AsyncStorage.getItem(HAS_SEEN_MY_COLLECTION_NEW_WORKS_BANNER)
-    const shouldShowConsignments = showSessionVisualClue("ArtworkSubmissionBanner")
+    const shouldShowConsignments = showSessionVisualClue("ArtworkSubmissionMessage")
```
olerichter00, line 107: **(note):** I've renamed `banner` to `message` because of https://github.com/artsy/eigen/pull/6637

src/app/Scenes/MyProfile/MyProfileHeaderMyCollectionAndSavedWorks.tsx
```diff
@@ -69,13 +70,31 @@ export const MyProfileHeader: React.FC<{ me: MyProfileHeaderMyCollectionAndSaved
...
 
   const userProfileImagePath = localImage || me?.icon?.url
 
+  const [showCompleteCollectorProfileMessage, setShowCompleteCollectorProfileMessage] = useState(
+    showVisualClue("CompleteCollectorProfileMessage")
+  )
+
+  useEffect(() => {
+    setVisualClueAsSeen("CompleteCollectorProfileMessage")
+  }, [])
```
olerichter00, line 88: **(note):** This sets the visual clue as seen as soon as the user opens the screen. Because we store `showCompleteCollectorProfileMessage` in a local state, the banner will be shown until the component unmounts (which should be the whole session).

olerichter00 (approved):

MounirDhahri (approved):
🙌

MounirDhahri:
Good use of the visual clues

dariakoko:
Hey @olerichter00 
Shouldn't the body text color be grey?
image

PR #6577, opened 2022-04-08
feat(NX-3118): Display edition set selection modal only when the artwork is non-unique

lean8086:
This PR resolves [NX-3118]

### Description

- Click on Make Offer in Inquiry button should open a modal window to confirm the artwork ONLY when there is more than one edition set.
- Unique artworks should take the user directly to the checkout flow
- Artworks with only one edition set should take the user directly to the checkout flow

cc @artsy/negotiate-devs 

- [ ] Screenshots and videos 
- [x] Added Tests and Stories
- [ ] App state migration
- [ ] Feature flag

### To the reviewers 👀

- Display edition set selection modal only when the artwork is non-unique - LeandroL

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[NX-3118]: https://artsyproduct.atlassian.net/browse/NX-3118?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/Inbox/Components/Conversations/InquiryMakeOfferButton.tsx
+++ b/src/app/Scenes/Inbox/Components/Conversations/InquiryMakeOfferButton.tsx
@@ -16,6 +16,7 @@ export interface InquiryMakeOfferButtonProps {
   disabled?: boolean
   conversationID: string
   replaceModalView?: boolean
+  onPress?: () => void
 }
 
 export interface State {
@@ -53,7 +54,7 @@ export class InquiryMakeOfferButton extends React.Component<InquiryMakeOfferButt
   }
 
   handleCreateInquiryOfferOrder() {
-    const { relay, artwork, editionSetID, conversationID, replaceModalView } = this.props
+    const { relay, artwork, editionSetID, conversationID, replaceModalView = true } = this.props
     const { isCommittingCreateOfferOrderMutation } = this.state
     const { internalID } = artwork
 
@@ -126,19 +127,25 @@ export class InquiryMakeOfferButton extends React.Component<InquiryMakeOfferButt
 
   render() {
     const { isCommittingCreateOfferOrderMutation } = this.state
+    const { onPress, disabled, variant, buttonText } = this.props
 
     return (
       <Button
-        onPress={() => this.handleCreateInquiryOfferOrder()}
+        onPress={() => {
+          if (onPress) {
+            onPress()
+          }
+          this.handleCreateInquiryOfferOrder()
+        }}
         loading={isCommittingCreateOfferOrderMutation}
         size="large"
-        disabled={this.props.disabled}
+        disabled={disabled}
         block
         width={100}
-        variant={this.props.variant}
+        variant={variant}
         haptic
       >
-        {this.props.buttonText ? this.props.buttonText : "Make an Offer"}
+        {buttonText ? buttonText : "Make an Offer"}
       </Button>
     )
   }
--- a/src/app/Scenes/Inbox/Components/Conversations/InquiryMakeOfferButton.tsx
+++ b/src/app/Scenes/Inbox/Components/Conversations/InquiryMakeOfferButton.tsx
@@ -3,7 +3,7 @@ import { InquiryMakeOfferButtonOrderMutation } from "__generated__/InquiryMakeOf
 import { navigate } from "app/navigation/navigate"
 import { Button, ButtonProps } from "palette"
 import React from "react"
-import { Alert } from "react-native"
+import { Alert, GestureResponderEvent } from "react-native"
 import { commitMutation, createFragmentContainer, graphql, RelayProp } from "react-relay"
 
 export interface InquiryMakeOfferButtonProps {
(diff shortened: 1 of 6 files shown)


src/app/Scenes/Inbox/Components/Conversations/InquiryMakeOfferButton.tsx
```diff
@@ -105,7 +106,7 @@ export class InquiryMakeOfferButton extends React.Component<InquiryMakeOfferButt
               } else if (orderOrError.__typename === "CommerceOrderWithMutationSuccess") {
                 navigate(`/orders/${orderOrError.order.internalID}`, {
                   modal: true,
-                  replace: true,
```
araujobarret, line 108: are we sure about this change? those who are using this component without passing `replaceModalView` prop will have this replace changed to `false` instead of `true`

src/app/Scenes/Inbox/Components/Conversations/OpenInquiryModalButton.tsx
```diff
@@ -39,21 +43,30 @@ export const OpenInquiryModalButton: React.FC<OpenInquiryModalButtonProps> = ({
...
+            }}
+            size="large"
+            variant="fillDark"
+            block
+            width={100}
+          >
+            Make an Offer
+          </Button>
+        ) : (
+          <InquiryMakeOfferButtonFragmentContainer
```
araujobarret, line 63: what about tracking here? before we tracked both cases, unique and edition sets, with this change here we don't track anymore unique artworks dispatching the event

araujobarret (changes requested):

src/app/Scenes/Inbox/Components/Conversations/OpenInquiryModalButton.tsx
lean8086: Good catch! I completely missed that one.

src/app/Scenes/Inbox/Components/Conversations/InquiryMakeOfferButton.tsx
lean8086: Agreed! I'll define it as `true` as default.

src/app/Scenes/Inbox/Components/Conversations/OpenInquiryModalButton.tsx
lean8086: I added tests to make sure we always track `tappedMakeOffer` 👍 
Let me know what you think about [sending an `onPress` property](https://github.com/artsy/eigen/pull/6577/files#diff-b20d881f9613cdd5d80dfb1d14cceb12684176faa444db82c17136aacfc3ce85R68) to track the event. I'm curious to see what's the cleanest way @araujobarret

src/app/Scenes/Inbox/Components/Conversations/InquiryMakeOfferButton.tsx
```diff
@@ -125,19 +127,25 @@ export class InquiryMakeOfferButton extends React.Component<InquiryMakeOfferButt
...
   render() {
     const { isCommittingCreateOfferOrderMutation } = this.state
+    const { onPress, disabled, variant, buttonText } = this.props
 
     return (
       <Button
-        onPress={() => this.handleCreateInquiryOfferOrder()}
+        onPress={() => {
+          if (onPress) {
+            onPress()
```
lean8086, line 136: This is outside of `this.handleCreateInquiryOfferOrder` because that one is called again if the mutation fails (`onMutationError`) and the user taps Retry

araujobarret (dismissed):
Looks good! 

Left a comment something nice in regards that component design and alignment with `ButtonProps` but nothing critical 👍

src/app/Scenes/Inbox/Components/Conversations/InquiryMakeOfferButton.tsx
araujobarret: sorry wrong reply box, should be in the last one you tagged me 😭

src/app/Scenes/Inbox/Components/Conversations/OpenInquiryModalButton.tsx
araujobarret: that's great, thanks!
I commented on the other chat thread regarding some minor improvement there 👍

src/app/Scenes/Inbox/Components/Conversations/InquiryMakeOfferButton.tsx
lean8086: Perfect! Just to be sure about the first point: By using the right children prop, you mean like in [Force's implementation](https://github.com/artsy/force/blob/main/src/v2/Apps/Conversation/Components/ConfirmArtworkButton.tsx#L87), right?

araujobarret: Sorry, my connection was broken yesterday, and wasn't able to come back here.

Yes, exactly 👍

lean8086: No problem at all! I opened a [follow-up (draft) PR](https://github.com/artsy/eigen/pull/6606/files#diff-c164780d59564bbe0fa13eff2a35f53b3d919c2a88ed72bf8d32b1c7e6e4ead8R145) with some questions re: using children.

PR #6592, opened 2022-04-11
feat: Change copy of my collection when no artwork

rajsam003:
This PR resolves [CX-2480]

### Description

This PR changes the copy in my collection when there are no artworks listed

|Before|After|
|---|---|
|image|image|

- [x] Screenshots and videos 
- [x] Added Tests and Stories
- [x] App state migration
- [x] Feature flag

### To the reviewers 👀

- Change copy in my collection when there are no artworks - sam

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-2480]: https://artsyproduct.atlassian.net/browse/CX-2480?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/Components/States/ZeroState.tsx
```diff
@@ -30,6 +37,7 @@ export const ZeroState = (props: ZeroStateProps) => {
               size="3"
               maxWidth={props.title ? "100%" : "80%"}
               lineHeight="20"
+              textAlign={props.alignItems ? "center" : "left"}
```
olerichter00, line 40: **(issue):** This would set `textAlign` to `"center"` whenever the prop `alignItems` is passed. I don't think we want that.

src/app/Components/States/ZeroState.tsx
```diff
@@ -1,18 +1,25 @@
...
 
 import { Flex, Sans, useColor } from "palette"
+import { AlignItemsValue } from "../Bidding/Elements/types"
 
 interface ZeroStateProps {
   title?: string
   subtitle?: string
   separators?: boolean
   callToAction?: JSX.Element
+  alignItems?: AlignItemsValue
```
olerichter00, line 11: This prop seems to do a lot more than it should. It sets the padding for the container and `textAlign` which is quite confusing.

src/app/Components/States/ZeroState.tsx
```diff
@@ -1,18 +1,25 @@
...
 }
 
 export const ZeroState = (props: ZeroStateProps) => {
   const color = useColor()
   return (
-    <Flex py="6" px="2" justifyContent="center">
+    <Flex
+      py="6"
+      px={props.alignItems ? "0" : "2px"}
+      justifyContent="center"
```
olerichter00, line 20: Why is this needed? When removed it looks the same for me.

src/app/Components/States/ZeroState.tsx
dariakoko, line 40: Why not doing something like this?
textAlign={props.alignItems ? **_props.alignItems_** : "left"}

dariakoko: You'd have an issue with types here? 
you use prop alignItems for textAlign

olerichter00 (changes requested):
Thanks for the work 🙇 

I don't think we should create two different flavors of `ZeroState`. Could you please check if this is intended?


| Saved Works | My Collection |
| --- | --- |
| simulator_screenshot_263B3A43-C0C7-4E77-8790-92EFBA037AC6 |simulator_screenshot_1342FA5D-F034-4B39-A29F-8C020787DF3F |

src/app/Components/States/ZeroState.tsx
dariakoko: You can add a new boolean prop and call it "centered" for example
You could use it for both textAlign and alignItems

textAlign={props.centered ? "center" : "left"}
alignItems={props.centered ? "center" : "stretch"}
("stretch" is default value for alignItems)

dariakoko: const { title, subtitle, callToAction } = props
👀  the code will look cleaner

dariakoko (changes requested):
Thanks for the work
I left some comments

olerichter00 (changes requested):
Looks great! 🌟 

Are these changes to [images/lock.webp](https://github.com/artsy/eigen/pull/6592/files#diff-d893ebda77f1af59e03b1e74b25f82d7e7e7ebf5e6dce0e61d0d584779cfbdcc) intentional?

rajsam003:
> Looks great! 🌟
> 
> Are these changes to [images/lock.webp](https://github.com/artsy/eigen/pull/6592/files#diff-d893ebda77f1af59e03b1e74b25f82d7e7e7ebf5e6dce0e61d0d584779cfbdcc) intentional?

Yes. As per the design. It seems like we need it.

olerichter00 (approved):

PR #6598, opened 2022-04-11
chore: disable build-test-js on beta builds

MounirDhahri:
This PR resolves []

### Description
Whenever we are trying to build a beta - we run the `build-test-js` workflow and we don't really need it. The reason behind that is this workflow already runs on every PR and runs on main as well. This will save us about 15 minutes for andoid and for android as well

- [ ] Tested on **iOS** and **Android**
- [ ] Screenshots and videos 
- [ ] Added Tests and Stories
- [ ] App state migration
- [ ] Feature flag

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- disable build-test-js on beta builds - mounir

</details>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -610,6 +610,8 @@ workflows:
                 - app_store_submission
                 - play_store_submission
                 - update-changelog
+                - beta-android
+                - beta-ios
       - horizon/block:
           context: horizon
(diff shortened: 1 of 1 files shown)


pvinis (approved):
I guess that makes sense 👍

PR #6603, opened 2022-04-12
feat(SWA-270): Make SubmitArtwork the default flow for Artwork Submission 🙌🏻

patrinoua:
https://user-images.githubusercontent.com/17421923/163423274-ae513e15-f51c-4f4e-a3bb-b2f72640f3a4.mov

This PR resolves [SWA-270] : The aim of this PR is to Clean up the old Artwork Submission flow and keep the new updated SellWithArtsy, SubmitArtwork Flow.

✅  Clean up the old Artwork Submission flow
✅  Remove Feature Flag
✅  Following the product decision to completely retire the word "Consignment", this PR renames everything "Consignment" related (also ConsignmentsHome) to "SellWithArtsy" (or SellWithArtsyHome)
✅  Move components like **ArtistAutosuggest** and **LocationAutocomplete** to app/Components instead of local definition
✅  Move **SubmitArtworkOverview** to app/scenes and remove **Consignment** 
✅  Rename **SubmitArtworkOverview** to **SubmitArtwork**

### TODO: PR Checklist

- [x] Screenshots and videos 

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- retire the old submission flow

</details>

[SWA-270]: https://artsyproduct.atlassian.net/browse/SWA-270?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/best_practices.md
```diff
@@ -13,7 +13,7 @@ _Please note: Links should point to specific commits, and not a branch (in case
 - [History](#history)
 - [File Structure Organization](#file-structure-organization)
 - [Frontend](#frontend)
-  - [Storybook](#storybook)  
+  - [Storybook](#storybook)
```
patrinoua, line 16: not sure why these whitespace changes are appearing, I didn't really touch that file 👀  🤷🏻‍♀️

src/app/Scenes/SubmitArtwork/ArtworkDetails/ArtworkDetails.tests.tsx
```diff
@@ -9,20 +9,20 @@ import { RelayEnvironmentProvider } from "react-relay"
 import { act } from "react-test-renderer"
 import { useTracking } from "react-tracking"
 import { createMockEnvironment } from "relay-test-utils"
-import { createConsignSubmission, updateConsignSubmission } from "../Mutations"
+import { createConsignSubmission, updateConsignSubmission } from "../utils/mutations"
```
patrinoua, line 12: **m**utations instead of **M**utations 🎉

src/app/Containers/Inquiry.tsx
```diff
@@ -83,6 +84,42 @@ const ResponseRateLine = styled.View`
...
+  showSeparator?: boolean
+}
+
+export const BottomAlignedButton: React.FC<BottomAlignedProps> = ({
+  buttonText,
+  onPress,
+  children,
+  disabled,
+  showSeparator = true,
+}) => (
```
damassi, line 101: Since code always eventually gets returned to, its best to [avoid implicit returns on multiline components](https://github.com/artsy/force/blob/main/docs/best_practices.md#avoid-implicit-returns-on-react-components). It might feel a little cleaner when writing it, but code is never complete, and eventually someone will need to add a `debugger` statement in here, or a variable, or a console.log, etc.

src/app/Components/ArtistAutosuggest/ArtistAutosuggest.tests.tsx
```diff
@@ -6,7 +6,7 @@ import React from "react"
 import { RelayEnvironmentProvider } from "react-relay"
 import { act } from "react-test-renderer"
 import { createMockEnvironment } from "relay-test-utils"
-import { ArtworkDetails } from "../ArtworkDetails"
+import { ArtworkDetails } from "../../Scenes/SubmitArtwork/ArtworkDetails/ArtworkDetails"
```
MounirDhahri, line 9: **nitpick:** in situations where you find yourself importing a component from a different scene, it's better to extract that component into a shared component

MounirDhahri: same note applies to the two file below as well  `ArtistAutoSuggest.tsx`

MounirDhahri: same note for `ArtistAutosuggestResults.tsx` and the other files

src/app/Containers/Inquiry.tsx
```diff
@@ -83,6 +84,42 @@ const ResponseRateLine = styled.View`
...
+  <ArtsyKeyboardAvoidingView>
```
MounirDhahri, line 102: **thought:** I don't think we should be having the keyboard avoidance logic inside the button, maybe it's better to have this inside the wrapping screen

MounirDhahri: ah nvm it's been like this for a while apparently, but would have been great if it were not there

MounirDhahri (commented):
I am not sure if the PR is reviewable yet or not since I can see some open questions, the PR checklist is missing, and no screenshots/video for the changes.

Would you also mind if you rename the PR title to follow Adam's [RFC](https://github.com/artsy/README/issues/327)

Nonetheless, I did enjoy reviewing the PRand left some small comments, none are blocking though, and just about files location.

src/app/Containers/Inquiry.tsx
patrinoua: Thanks! In this case I just moved the definition of this function here without changing here so that I can get rid of the file, since it was defined in a place far away and we only used it here. I didn't touch it further and if I was to do that I would also add a story and a test and move this under palette or components but I'm guessing refactoring this is outside the scope of this ticket :)

patrinoua: But could we somehow have a #refactor flag?

src/app/Components/ArtistAutosuggest/ArtistAutosuggest.tests.tsx
patrinoua: Thanks! The PR is indeed not 100% there yet but maybe 95%?

In this case I just moved the definition of these components on a general place without refactoring, since they were locally defined, but now they are more easily accessible. I didn't touch it further as I'm guessing refactoring this is outside the scope of this ticket :) 

But can we maybe add a #refactor flag and perhaps do some refactoring in the scope of the refinathon?

src/app/Containers/Inquiry.tsx
patrinoua: Thanks! In this case I just moved the definition of this function here without changing here so that I can get rid of the file, since it was defined in a place far away and we only used it here. I didn't touch it further and if I was to do that I would also add a story and a test and move this under palette or components but I'm guessing refactoring this is outside the scope of this ticket :) 

But can we maybe add a #refactor flag and perhaps do some refactoring in the scope of the refinathon?

src/app/Components/ArtistAutosuggest/ArtistAutosuggest.tests.tsx
MounirDhahri: sounds like a nice idea, I think we need a board for refactoring - not sure if it should be with the product team jira board or somewhere else in github or notion but we definitely need somewhere where we can track those components to be refactored

src/app/Containers/Inquiry.tsx
MounirDhahri: sounds good to me

src/app/Components/ArtistAutosuggest/ArtistAutosuggest.tests.tsx
MounirDhahri: the only thing I am afraid off is actually us leaving things to the refinathon and not refactoring components when we see an opportunity

patrinoua: In general I completely agree that it's nice to do refactorings when we see an opportunity. However, given the size of this PR and the fact that the scope of it is just removing the old flow, I think it would be better not to create more chaos and leave further refactoring outside this. 

In the specific case I also think that the definition of ArtworkDetails belongs where it's currently defined. But feel free to have a closer look and let me know and I'll gladly refactor it if necessary :-)

MounirDhahri: sounds good to me. no need to do any further refactorings 🙏

src/app/AppRegistry.tsx
```diff
@@ -43,9 +43,6 @@ import { CitySavedListQueryRenderer } from "./Scenes/City/CitySavedList"
 import { CitySectionListQueryRenderer } from "./Scenes/City/CitySectionList"
 import { CollectionQueryRenderer } from "./Scenes/Collection/Collection"
 import { CollectionFullFeaturedArtistListQueryRenderer } from "./Scenes/Collection/Components/FullFeaturedArtistList"
-import { Consignments } from "./Scenes/Consignments"
-import { ConsignmentsHomeScreenQuery } from "./Scenes/Consignments/ConsignmentsHome/ConsignmentsHome"
-import { ConsignmentsSubmissionForm } from "./Scenes/Consignments/ConsignmentsHome/ConsignmentsSubmissionForm"
```
patrinoua, line 48: Following the product decision to completely retire the word "Consignment", this PR renames everything "Consignment" related to "SellWithArtsy"

MounirDhahri:
I think you forgot to update the PR checklist to use the one we added the past few days as well as adding a changelog entry

MounirDhahri (approved):
Sounds good to me 🙌
Did you also remove the feature flag at some point?

PR #6606, opened 2022-04-13
chore: render children as button content instead of a specific prop

lean8086:
### Description

Replace property `buttonText` with `children` on InquiryMakeOfferButton.
Keep the default value of `Make an Offer`
See [implementation on Force](https://github.com/artsy/force/blob/main/src/v2/Apps/Conversation/Components/ConfirmArtworkButton.tsx#L87).

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- Replace property `buttonText` with `children` on InquiryMakeOfferButton - LeandroL

</details>

[docs]: /docs/README.md

### DIFF
--- a/src/app/Scenes/Inbox/Components/Conversations/InquiryMakeOfferButton.tsx
+++ b/src/app/Scenes/Inbox/Components/Conversations/InquiryMakeOfferButton.tsx
@@ -139,7 +139,7 @@ export class InquiryMakeOfferButton extends React.Component<InquiryMakeOfferButt
         variant={variant}
         haptic
       >
-        {children || "Make an Offer"}
+        {children}
       </Button>
     )
   }
--- a/src/app/Scenes/Inbox/Components/Conversations/InquiryMakeOfferButton.tsx
+++ b/src/app/Scenes/Inbox/Components/Conversations/InquiryMakeOfferButton.tsx
@@ -3,20 +3,16 @@ import { InquiryMakeOfferButtonOrderMutation } from "__generated__/InquiryMakeOf
 import { navigate } from "app/navigation/navigate"
 import { Button, ButtonProps } from "palette"
 import React from "react"
-import { Alert, GestureResponderEvent } from "react-native"
+import { Alert } from "react-native"
 import { commitMutation, createFragmentContainer, graphql, RelayProp } from "react-relay"
 
-export interface InquiryMakeOfferButtonProps {
+export interface InquiryMakeOfferButtonProps extends ButtonProps {
   artwork: InquiryMakeOfferButton_artwork
   relay: RelayProp
   // EditionSetID is passed down from the edition selected by the user
   editionSetID: string | null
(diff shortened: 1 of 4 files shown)


src/app/Scenes/Inbox/Components/Conversations/InquiryMakeOfferButton.tsx
```diff
@@ -143,7 +142,7 @@ export class InquiryMakeOfferButton extends React.Component<InquiryMakeOfferButt
         variant={variant}
         haptic
       >
-        {buttonText ? buttonText : "Make an Offer"}
+        {children || "Make an Offer"}
```
lean8086, line 145: I'm keeping the same default value from before, although we are using [the other option "Confirm" as default in Force](https://github.com/artsy/force/blob/main/src/v2/Apps/Conversation/Components/ConfirmArtworkButton.tsx#L87).

src/app/Scenes/Inbox/Components/Conversations/OpenInquiryModalButton.tsx
```diff
@@ -67,7 +67,9 @@ export const OpenInquiryModalButton: React.FC<OpenInquiryModalButtonProps> = ({
             conversationID={conversationID}
             onPress={() => trackEvent(tracks.trackTappedMakeOffer(conversationID))}
             replaceModalView={false}
-          />
+          >
+            Make an Offer
```
lean8086, line 71: Although this is the default option, I preferred being explicit with the content here (see the entire file to check readability)

src/app/Scenes/Inbox/Components/Conversations/InquiryMakeOfferButton.tsx
```diff
@@ -12,7 +12,6 @@ export interface InquiryMakeOfferButtonProps {
   // EditionSetID is passed down from the edition selected by the user
```
araujobarret, line 12: this interface could extend the `ButtonProps` and then we don't need to re-declare some props like variant, disabled, onPress, we would be respecting and drilling atomic elements props in the way we do with other components.

araujobarret: was a requested change too from the previous comment in the other PR ☝️

lean8086: Woops 👍

src/app/Scenes/Inbox/Components/Conversations/OpenInquiryModalButton.tsx
lean8086: nvm, children is not optional on `ButtonProps` so it needs to be explicitly defined.

araujobarret (approved):
Looks good, thanks for cleaning it up!

PR #6636, opened 2022-04-19
(BID-294): Add timers sale artwork grid

annacarey:
This PR resolves [BX-294](https://artsyproduct.atlassian.net/browse/BID-294)

### Description

This PR adds the countdown timer to the lot grid on the sale page for sales with cascading end times turned on:

Lots are in the middle of closing:
Screen Shot 2022-04-19 at 11 39 36 AM

Lot is less than the cascade end interval (1 min) from closing:
Screen Shot 2022-04-19 at 11 38 31 AM

Lots are minutes from closing:
Screen Shot 2022-04-19 at 11 37 12 AM

Lots are hours from closing:
Screen Shot 2022-04-19 at 11 26 51 AM

When the sale is open but lots are more than a day from closing (formatted lot end time):
Screen Shot 2022-04-19 at 11 04 48 AM

When the sale is not yet open (no label):
Screen Shot 2022-04-19 at 10 39 32 AM

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[docs]: /docs/README.md

### DIFF
--- a/src/app/Components/ArtworkGrids/ArtworkGridItem.tsx
+++ b/src/app/Components/ArtworkGrids/ArtworkGridItem.tsx
@@ -11,14 +11,13 @@ import {
   PlaceholderRaggedText,
   RandomNumberGenerator,
 } from "app/utils/placeholders"
-import { Duration } from "moment"
 import { Box, Flex, Sans, Spacer, Text, TextProps, Touchable } from "palette"
 import React, { useRef } from "react"
 import { View } from "react-native"
 import { createFragmentContainer, graphql } from "react-relay"
 import { useTracking } from "react-tracking"
 import { DurationProvider } from "../Countdown"
-import { getTimerInfo } from "../Countdown/Ticker"
+import { LotCloseInfo } from "./LotCloseInfo"
 
 export interface ArtworkProps {
   artwork: ArtworkGridItem_artwork
@@ -173,7 +172,7 @@ export const Artwork: React.FC<ArtworkProps> = ({
               <Text variant="xs" numberOfLines={1} caps {...lotLabelTextStyle}>
                 Lot {artwork.saleArtwork.lotLabel}
               </Text>
-              {!!artwork.sale && (
+              {!!artwork.sale && !!artwork.sale.cascadingEndTimeIntervalMinutes && (
                 <DurationProvider startAt={artwork.saleArtwork.endAt!}>
                   <LotCloseInfo
                     duration={null}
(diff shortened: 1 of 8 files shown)


src/app/store/config/features.ts
```diff
@@ -112,6 +112,12 @@ export const features = defineFeatures({
     showInAdminMenu: true,
     echoFlagKey: "ARDisableCascadingEndTimerLotPage",
   },
+  ARDisableCascadingEndTimerSalePageGrid: {
```
olerichter00, line 115: What do you think about renaming the flag to `AREnableCascadingEndTimerSalePageGrid`? Usually, this is the schema we use for feature flags.

src/app/Components/ArtworkGrids/ArtworkGridItem.tsx
```diff
@@ -166,9 +170,22 @@ export const Artwork: React.FC<ArtworkProps> = ({
...
-            <Text variant="xs" numberOfLines={1} caps {...lotLabelTextStyle}>
-              Lot {artwork.saleArtwork.lotLabel}
-            </Text>
+            <>
+              <Text variant="xs" numberOfLines={1} caps {...lotLabelTextStyle}>
+                Lot {artwork.saleArtwork.lotLabel}
+              </Text>
+              {!!artwork.sale &&
+                !!artwork.sale.cascadingEndTimeIntervalMinutes &&
+                !!cascadingEndTimeFeatureEnabled && (
```
olerichter00, line 179:
```suggestion
              {!!artwork.sale?.cascadingEndTimeIntervalMinutes &&
                !!cascadingEndTimeFeatureEnabled && (
```

olerichter00 (approved):

src/app/store/config/features.ts
annacarey: I talked to pavlos and he told me I should do disable for these as it makes it easier to deprecate them. In this case, these flags are only meant to be a kill-switch should something go wrong when added to the release. So they can be removed once we see the feature as stable. I'm totally cool with either, but that was my understanding. @brainbicycle did I discuss this with you too?

src/app/Components/ArtworkGrids/ArtworkGridItem.tsx
annacarey: Great call!

src/app/store/config/features.ts
brainbicycle: I don't think it is a big deal to do this once but I would personally prefer we keep to the standard for future stuff, I think with a mixture things get confusing.

olerichter00: I just remember that we have [this](https://github.com/artsy/eigen/blob/main/fastlane/Fastfile#L554) script that posts all hidden features for an upcoming release on Slack. If I understand it correctly it relies on feature flags enabling features.

Screenshot 2022-04-20 at 14 12 04

brainbicycle: heh you are right it does I forgot about that

brainbicycle: the other problem with features being enabled by default is that it compromises the other purpose of feature flags which is to allow for in progress work to get into the main branch without risk of it getting shipped to customers. With features enabled by default in progress work my accidentally get shipped.

PR #6676, opened 2022-04-28
fix(CX-2543): Fix performance with my collection search

lordkiz:
This PR resolves [CX-2543]

#### Previous Issues this PR fixes:
- MyCollection screen hangs because we try to store yScrollOffset
- Header animation into view is a bit janky
- When moving from FilteredStateZero to another state, SearchBar jumps across the screen from the top to the bottom
- Switching from Grid to List view, search bar jumps from top to bottom
- SearchBar is added to StickyHeader thereby reducing available scrollable area

#### What was done in this PR:
- Remove need for scrolloffset, instead using staticHeader height to scroll
- Remove hideStaticHeader and showStaticHeader. Instead get a handle on the inner flatlist and perform scroll on it. This removes the jumping from bottom to top.
- Pass the search bar as a header component instead, thereby not reducing the scrollable area.
- contentOffset to initially hide the header.
- When switching from Grid to List or to ZerofilteredArtworks state: SearchBar should remain at its previous position and not jump across screens, SearchBar should remain focused if was previously focused on the previous View, and do not hide the searchbar if it has been interacted with on a previous View

#### Possible Remaining issues:
- iOS: None
- Android: User will have to have scrolled atleast an offset of greater than zero to reveal hidden search bar. 

https://user-images.githubusercontent.com/18648835/165947777-543d108d-2f98-456a-adfc-493501ff5baf.mp4

https://user-images.githubusercontent.com/18648835/165947614-a15daf58-798f-482a-ba99-3dd440553e20.mp4

### PR Checklist

### To the reviewers 👀

- Fix performance issues associated with MyCollectionSearch - kizito

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[docs]: /docs/README.md

[CX-2543]: https://artsyproduct.atlassian.net/browse/CX-2543?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Components/ArtworkGrids/InfiniteScrollArtworksGrid.tsx
+++ b/src/app/Components/ArtworkGrids/InfiniteScrollArtworksGrid.tsx
@@ -420,6 +420,7 @@ class InfiniteScrollArtworksGrid extends React.Component<Props & PrivateProps, S
             if (autoFetch) {
               this.handleFetchNextPageOnScroll(ev)
             }
+            // [Android ContentOffset Bug]: See Hacks.MD
             if (Platform.OS === "android" && ev.nativeEvent.contentOffset.y - 0 === 0) {
               LayoutAnimation.configureNext({
                 ...LayoutAnimation.Presets.linear,
@@ -428,9 +429,11 @@ class InfiniteScrollArtworksGrid extends React.Component<Props & PrivateProps, S
               this.setState({ marginTop: 0 })
             }
           }}
+          // [Android ContentOffset Bug]: See Hacks.MD for why we are using marginTop here to hide header
           contentContainerStyle={
             Platform.OS === "android" && hideHeaderInitially ? { marginTop: -marginTop } : undefined
           }
+          // [Android ContentOffset Bug]: contentOffset not working on android. This will only apply to iOS. See Hacks.MD
           contentOffset={hideHeaderInitially ? { x: 0, y: headerHeight } : undefined}
           scrollEventThrottle={50}
           onLayout={this.onLayout}
--- a/src/app/Components/ArtworkGrids/InfiniteScrollArtworksGrid.tsx
+++ b/src/app/Components/ArtworkGrids/InfiniteScrollArtworksGrid.tsx
@@ -18,6 +18,7 @@ import React from "react"
 import {
   ActivityIndicator,
   Dimensions,
+  LayoutAnimation,
   LayoutChangeEvent,
   Platform,
   ScrollView,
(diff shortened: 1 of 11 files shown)


src/app/Components/ArtworkGrids/InfiniteScrollArtworksGrid.tsx
```diff
@@ -363,7 +379,13 @@ class InfiniteScrollArtworksGrid extends React.Component<Props & PrivateProps, S
...
     }
 
-    return React.isValidElement(HeaderComponent) ? HeaderComponent : <HeaderComponent />
+    return React.isValidElement(HeaderComponent) ? (
+      <View onLayout={this.onHeaderLayout}>{HeaderComponent}</View>
+    ) : (
+      <View onLayout={this.onHeaderLayout}>
+        <HeaderComponent />
+      </View>
+    )
```
olerichter00, line 388: Can we just check for the type of the header component and change this to:

```suggestion
    return (
     <View onLayout={this.onHeaderLayout}>
        <HeaderComponent />
      </View>
    )
```

src/app/Scenes/MyCollection/Components/MyCollectionArtworkList.tsx
```diff
@@ -41,9 +64,60 @@ export const MyCollectionArtworkList: React.FC<{
...
+      return null
+    }
+
+    return React.isValidElement(HeaderComponent) ? (
+      <View onLayout={onHeaderLayout}>{HeaderComponent}</View>
+    ) : (
+      <View onLayout={onHeaderLayout}>
+        <HeaderComponent />
+      </View>
+    )
```
olerichter00, line 85: Same comment as above.

olerichter00 (approved):
This is great work! 🌟  I've tested it on the simulator and it works very well. I'll approve but please let someone else review as well before merging.

src/app/Scenes/MyCollection/Components/MyCollectionArtworkList.tsx
lordkiz: @olerichter00 I think it reduces ambiguity. Reason being that this allows for devs to pass () => </> or Element as is. Typically this Header should be in a ListHeaderComponent where you would pass element that passes `React.isValidElement`, but it isn't because of how the grid is composed. I think the original authors wanted devs to have the freedom to use it as they would typically use a FlatList's ListHeaderComponent and I think that makes a lot of sense. I only wrapped a view around it to measure the height.

src/app/Components/ArtworkGrids/InfiniteScrollArtworksGrid.tsx
olerichter00: Ok

MounirDhahri (approved):
Niice thanks a lot Kizito!

PR #6690, opened 2022-04-29
feat(FX-3928): upload file using GraphQL query/mutation

dimatretyak:
This PR resolves [FX-3928]

### Useful links
* [Spike](https://www.notion.so/artsy/Spike-Developer-can-pass-an-image-as-an-argument-to-a-query-in-Metaphysics-29536efe82fe4743836b0bbe29ccaab4)
* [Extra context about upload middleware](https://www.notion.so/artsy/Spike-Developer-can-pass-an-image-as-an-argument-to-a-query-mutation-29536efe82fe4743836b0bbe29ccaab4#b1ea88ccc9694fc3a1f37495cf55a03a)
* [Technical Plan](https://www.notion.so/artsy/Reverse-Image-Search-for-Artworks-aka-Shazam-for-Art-33c9ece7293641b697788bad921c58bd)

### Description
A developer can enable a feature flag, tap a new icon, take an image, and execute a new Metaphysics query field which accepts file data.

### PR Checklist

- [x] I hid my changes behind a feature flag

### iOS
https://user-images.githubusercontent.com/3513494/166492278-8be01c70-c61d-4679-8fab-3b4c7658b721.mp4

### Android
https://user-images.githubusercontent.com/3513494/166487582-ca59c551-b498-42a7-9523-c42e1abf6c59.mp4

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- upload file using GraphQL query - dimatretyak

</details>

[docs]: /docs/README.md

[FX-3928]: https://artsyproduct.atlassian.net/browse/FX-3928?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


typings/extract-files.d.ts
```diff
@@ -0,0 +1,22 @@
+// Type definitions for extract-files 12.0.0
```
dimatretyak, line 1: [@types/extract-files](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b24e5f670af4b29aa131dedc1db578d08d40c028/types/extract-files/index.d.ts) is deprecated (declared for the 8.0.0 version, but the current version is 12.0.0)

metro.config.js
```diff
@@ -15,4 +15,8 @@ module.exports = {
       },
     }),
   },
+
+  resolver: {
+    sourceExts: ['jsx', 'js', 'ts', 'tsx', 'mjs'],
+  },
```
dimatretyak, line 21: `extract-files` uses `mjs` file format

package.json
```diff
@@ -126,6 +126,7 @@
     "autosuggest-highlight": "3.1.1",
     "easy-peasy": "3.3.0",
     "events": "3.3.0",
+    "extract-files": "12.0.0",
```
olerichter00, line 129: Would you mind creating an RFC about the introduction of `extract-files`? I'm not sure if we have this documented, but we try to create an [RFC](https://github.com/artsy/eigen/issues?q=is%3Aissue+rfc) for each library we introduce.

src/app/Scenes/Fair/Fair.tsx
```diff
@@ -55,6 +65,7 @@ export const Fair: React.FC<FairProps> = ({ fair }) => {
 
   const flatListRef = useRef<FlatList>(null)
   const [isFilterArtworksModalVisible, setFilterArtworkModalVisible] = useState(false)
+  const isImageSearchEnabled = useExperimentFlag("eigen-image-search")
```
olerichter00, line 68: Could we achieve the same with a [feature flag](https://github.com/artsy/eigen/blob/5c7ec254bec1af7219c081a3912e2d32e74697d8/src/app/store/config/features.ts#L36). The recommendation is to use Echo feature flags if possible ([Slack Thread about it](https://artsy.slack.com/archives/C02BAQ5K7/p1650447651295539)).

olerichter00 (commented):
Great work 🌟

src/app/relay/middlewares/uploadMiddleware.ts
```diff
@@ -0,0 +1,57 @@
...
+ * Note: this request will fail if you are using React Native debugger with "Network Inspect" feature enabled
+ * Note: we must definitely specify `name` for file, otherwise Metaphysics will assume that the file is NOT being passed
+ *
+ * Upload middleware from `react-relay-network-modern` with ReactNativeFile support
+ * Original: https://github.com/relay-tools/react-relay-network-modern/blob/master/src/middlewares/upload.js
+ *
+ * If we try to pass Blob/File with specified `name` field to `fetch()`, we will get an error on Android.
+ * For this reason, support for these data formats is extracted from original `upload` middleware
+ * TODO: Add support for Blob/File when problem will be fixed for Android
+ */
```
dblandin, line 15: **question**: Does this address a bug within the react-relay-network-modern upload middleware? If so, what do you think of using [patch-package](https://github.com/ds300/patch-) to patch the bug and hopefully we can get this fixed upstream as well?

It's a bit tricky to see where the _fix_ is within this new file. I assume it's the addition of `isReactNativeFile` within the `extract-files` call. A benefit of using patch-package, is that we have a clear diff (even if the diff isn't in TypeScript) with the specific changes.

package.json
dblandin: It looks like we're already pulling in `extract-files 9.0.0` which is a dependency of `react-relay-network-modern`. Is that version sufficient?

src/app/Scenes/Fair/Fair.tsx
dblandin: **clarification**: To re-iterate (and hopefully confirm) the gist of that Slack thread:

- We use [Echo](https://github.com/artsy/echo) to manage most remotely configurable feature flags in Eigen
- We have patterns to enable local overrides (via the admin menu) of Echo-backed flags
- Overriding Unleash-backed feature flags in-app is not yet implemented
- Until we're able to easily override Unleash-backed feature flags in-app, we'll prefer either entirely local flags or Echo-backed remote flags.

Does that sound about right?

dblandin (commented):
Looking good!

To summarize some comments:
- Agree with @olerichter00 that we should probably switch to a local feature flag (not yet Echo enabled)
- Consider using patch-package to address the upload middleware bug
- Might be able to reuse the existing extract-files v9 dependency

src/app/Scenes/Fair/Fair.tsx
olerichter00: Thanks for the summary! That sounds about right.

I would add only one argument for using local/Echo feature flags for **all** features and unleash experiments for A/B testing and percentage rollouts: consistency.

package.json
olerichter00: @MounirDhahri maybe you have some more thoughts about RFCs for dependencies to add here.

src/app/relay/createEnvironment.ts
```diff
@@ -35,6 +36,7 @@ export function createEnvironment(
       metaphysicsExtensionsLoggerMiddleware(),
       simpleLoggerMiddleware(),
       __DEV__ ? relayErrorMiddleware() : null,
+      uploadMiddleware(),
```
MounirDhahri, line 39: **thought:** I think we might be better moving this before the `errrorMiddleware`

src/app/relay/middlewares/uploadMiddleware.ts
```diff
@@ -0,0 +1,57 @@
+import { isReactNativeFile, ReactNativeFile } from "app/utils/ReactNativeFile"
+import extractFiles from "extract-files/extractFiles.mjs"
+import { Middleware, RelayNetworkLayerRequestBatch } from "react-relay-network-modern/node8"
+
+/**
+ * Note: this request will fail if you are using React Native debugger with "Network Inspect" feature enabled
+ * Note: we must definitely specify `name` for file, otherwise Metaphysics will assume that the file is NOT being passed
+ *
```
MounirDhahri, line 8: **praise:** thanks for the note, I was looking for this 😄

src/app/relay/middlewares/uploadMiddleware.ts
```diff
@@ -0,0 +1,57 @@
...
+ * Upload middleware from `react-relay-network-modern` with ReactNativeFile support
+ * Original: https://github.com/relay-tools/react-relay-network-modern/blob/master/src/middlewares/upload.js
+ *
+ * If we try to pass Blob/File with specified `name` field to `fetch()`, we will get an error on Android.
+ * For this reason, support for these data formats is extracted from original `upload` middleware
+ * TODO: Add support for Blob/File when problem will be fixed for Android
```
MounirDhahri, line 14: **thought:** this could be added to our `Hacks.md` as a resolution

src/app/utils/ReactNativeFile.ts
```diff
@@ -0,0 +1,21 @@
+export interface ReactNativeFileOptions {
+  uri: string
+  type?: string
+  name?: string
+}
+
+export class ReactNativeFile {
+  uri: string
+  type?: string
+  name?: string
```
MounirDhahri, line 10: **suggestion:** you mentioned above that the name of the file is needed to avoid issues on Android, I believe we won't be anytime soon having feature disparity with iOS. How about we make the field here required?

typings/extract-files.d.ts
```diff
@@ -0,0 +1,22 @@
...
+// Project: https://github.com/jaydenseric/extract-files#readme
+// Definitions by: Edward Sammut Alessi <https://github.com/Slessi>
+//                 Alex K <https://github.com/lynxtaa>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
```
MounirDhahri, line 5: **praise:** thanks, very thoughtful

MounirDhahri (commented):
LGTM, I left some small thoughts/suggestions but that's it

src/app/relay/createEnvironment.ts
dimatretyak: Done

typings/extract-files.d.ts
dimatretyak: thanks!

dimatretyak:
> It looks like we're already pulling in extract-files 9.0.0 which is a dependency of react-relay-network-modern. Is that version sufficient?

@dblandin yes, `9.0.0` version will be enough, but I decided to use `12.0.0` for the following reasons:
- if `9.0.0` version is used, then we will receive notifications from `dependabot` that it is necessary to update `extract-files` dependency to `12.0.0` version
- `9.0.0` version includes `ReactNativeFile` file, in `12.0.0` version this file was removed. Therefore, when upgrading to a fresh version, you would have to add this file anyway. I thought it would be better to use the latest version of `extract-files` right away than to update the old version in the future.

dblandin:
> if 9.0.0 version is used, then we will receive notifications from dependabot that it is necessary to update extract-files dependency to 12.0.0 version

I think dependabot would only prompt you to upgrade if `extract-files` is listed in the package.json as a direct dependency. If instead we patched the bug with `patch-package`, I don't think we would need to add `extract-files` as a dependency and wouldn't receive those automated PRs.

Even so, prompts for major version upgrades from dependabot are not merged automatically. We can close a PR to "ignore" that prompt if so desired.

I don't feel too strongly against maintaining our own forked upload middleware but if the diff is small, it might be simpler for our future selves to maintain a patch of the package rather than introducing new files and a new dependency.

dimatretyak:
> I think dependabot would only prompt you to upgrade if `extract-files` is listed in the package.json as a direct dependency. If instead we patched the bug with `patch-package`, I don't think we would need to add `extract-files` as a dependency and wouldn't receive those automated PRs.

Aah, I got what you mean 👍 

Will other developers understand that `extract-files` is used by [react-relay-network-modern](https://github.com/relay-tools/react-relay-network-modern/blob/master/package.json#L36) and therefore we can import and use this dependency in the project? Won't this add confusion to developers, since this dependency won't be in package.json?

@MounirDhahri @olerichter00 what do you think about it?

dblandin:
It looks like `extract-files` is already imported in the package's implementation of the upload middleware:

```
$ cat node_modules/react-relay-network-modern/lib/middlewares/upload.js
```

If we were to use patch-package, and extract-files is still only used within react-relay-network-modern, I don't think it would add too much confusion. patch-package would also warn us if we ever tried to upgrade react-relay-network-modern and the patch became incompatible. We would end up with a new file in this [patches/](https://github.com/artsy/eigen/tree/main/patches) folder.

dimatretyak:
> It looks like `extract-files` is already imported in the package's implementation of the upload middleware:

yep, but we need to import `ReactNativeFile` and use it also [here](https://github.com/artsy/eigen/blob/b868ca87c308063b39d183cf14e8420855528936/src/app/Scenes/Fair/Fair.tsx#L187-L191)

dblandin:
> We need to import ReactNativeFile and use it also [here](https://github.com/artsy/eigen/blob/b868ca87c308063b39d183cf14e8420855528936/src/app/Scenes/Fair/Fair.tsx#L187-L191) + we will need to install @types/extract-files to fix this

That makes sense - thanks for pointing that out 👍🏼 

Does installing only the types package fix that DX issue or do we need to hoist `extract-files` as well?

dimatretyak:
> Does installing only the types package fix that DX issue or do we need to hoist `extract-files` as well?

We can install only `@types/extract-files`. Here is a [commit](https://github.com/artsy/eigen/commit/78ab757aad3973d608803d654541cc17c42c6d6f) with all the changes if we decide to use patch

MounirDhahri:
> > I think dependabot would only prompt you to upgrade if `extract-files` is listed in the package.json as a direct dependency. If instead we patched the bug with `patch-package`, I don't think we would need to add `extract-files` as a dependency and wouldn't receive those automated PRs.
> 
> @dblandin Aah, I got what you mean 👍
> 
> Will other developers understand that `extract-files` is used by [react-relay-network-modern](https://github.com/relay-tools/react-relay-network-modern/blob/master/package.json#L36) and therefore we can import and use this dependency in the project? Won't this add confusion to developers, since this dependency won't be in package.json?
> 
> @MounirDhahri @olerichter00 what do you think about it?

I don't mind the explicit import again for `extract-files` for sure.

dblandin:
> Here is a [commit](https://github.com/artsy/eigen/commit/78ab757aad3973d608803d654541cc17c42c6d6f) with all the changes if we decide to use patch and extract-files from react-relay-network-modern

That looks pretty sweet to me!! Only change I might make is to omit the patch-package changes to the `upload.js.flow` file. I think only the compiled `.js` file will be imported.

dblandin:
And to re-iterate, patch-package is a great tool to patch packages when needed and if any of these patches would be useful upstream we should try to PR as a follow-up.

src/app/relay/middlewares/uploadMiddleware.ts
dimatretyak: Done

src/app/utils/ReactNativeFile.ts
dimatretyak: Done

src/app/Scenes/Fair/Fair.tsx
```diff
@@ -55,6 +65,7 @@ export const Fair: React.FC<FairProps> = ({ fair }) => {
 
   const flatListRef = useRef<FlatList>(null)
   const [isFilterArtworksModalVisible, setFilterArtworkModalVisible] = useState(false)
+  const isImageSearchEnabled = useFeatureFlag("AREnableImageSearch")
```
dblandin, line 68: **question**: Should we add an entry in [src/app/store/config/features.ts](https://github.com/artsy/eigen/blob/main/src/app/store/config/features.ts) so that it shows up in the admin menu?

patches/react-relay-network-modern+6.2.1.patch
```diff
@@ -0,0 +1,24 @@
...
++    });
+ 
+     if (files.size) {
+       const formData = new FormData();
+@@ -37,7 +39,7 @@ function uploadMiddleware() {
+       formData.append('map', JSON.stringify(pathMap));
+       i = 0;
+       files.forEach((paths, file) => {
+-        formData.append(++i, file, file.name);
++        formData.append(++i, file);
```
dblandin, line 21: **praise**: Looks great 👍🏼

src/app/Scenes/Fair/Fair.tsx
```diff
@@ -164,6 +176,44 @@ export const Fair: React.FC<FairProps> = ({ fair }) => {
 
   const hideBackButtonOnScroll = useHideBackButtonOnScroll()
 
+  const getFileNameByPath = (imagePath: string) => {
+    return imagePath.split("/").pop()!
+  }
+
+  const handleSeachByImage = async () => {
+    try {
+      const images = await showPhotoActionSheet(showActionSheetWithOptions, true, false)
```
dblandin, line 185: **praise**: This will be very useful to test with 👍🏼

src/app/relay/middlewares/uploadMiddleware.tests.ts
```diff
@@ -0,0 +1,151 @@
+import { ReactNativeFile } from "extract-files"
```
dblandin, line 1: **question**: Should we remove this file? If we have a spec that specifically exercises the change in our patch, we could leave it in, but otherwise I don't think we need to test something within a package.

dblandin (approved):
Looks great! Left a few comments but nothing major ✨

src/app/Scenes/Fair/Fair.tsx
dimatretyak: Aah, got you. [showInAdminMenu](https://github.com/artsy/eigen/blob/main/src/app/store/config/features.ts#L124) field already exists in `AREnableImageSearch`

dblandin: Oh! I forgot we already had that feature flag to reuse here. Thanks for clarifying 👍🏼

dblandin (dismissed):
💯

src/app/Scenes/Fair/Fair.tsx
dimatretyak: you're welcome

olerichter00 (approved):

dimatretyak:
### Quick QA
#### Android - API 26 (Pixel 2)
In progress...

#### Android - API 28 (Pixel 2)
https://user-images.githubusercontent.com/3513494/166702845-0d9ea655-edc8-4c09-887b-4bded747a7e2.mp4

#### Android - API 29 (Pixel 2)
https://user-images.githubusercontent.com/3513494/166703875-5fb5bee0-2963-4a7f-b11f-2289f9146237.mp4

#### Android - API 30 (Pixel 2)
https://user-images.githubusercontent.com/3513494/166704891-d8dc3bcb-937d-44f0-82f6-99f6d2c0acf1.mp4

PR #6698, opened 2022-05-04
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/beta-ios-manual.yml
+++ b/.github/workflows/beta-ios-manual.yml
@@ -7,6 +7,8 @@ permissions:
 
 jobs:
   build:
+      permissions:
+      contents: none
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/beta-ios-manual.yml
+++ b/.github/workflows/beta-ios-manual.yml
@@ -2,6 +2,9 @@ name: Trigger beta iOS release
 
 on: workflow_dispatch
 
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


.github/workflows/beta-ios-manual.yml
```diff
@@ -2,6 +2,9 @@ name: Trigger beta iOS release
 
 on: workflow_dispatch
 
+permissions:
+  contents: read
+
 jobs:
   build:
```
MounirDhahri, line 9: **thought:** I think you missed setting the permissions inside the job here as you did for other actions
```suggestion
  build:
      permissions:
      contents: none
```

MounirDhahri (commented):
Thanks for your contribution @naveensrinivasan I left a small comment and we should be clear to merge this afterward

naveensrinivasan:
> Thanks for your contribution @naveensrinivasan I left a small comment and we should be clear to merge this afterward

@MounirDhahri Done. Thanks.  PTAL.

MounirDhahri:
LGTM 🙌 Thanks 🎉

PR #6709, opened 2022-05-06
chore: move sentry upload artifact step to after submitting a build

MounirDhahri:
This PR resolves an issue where builds do not get shipped if something goes wrong with sentry

**PS:** It's still important though that the Sentry upload artifacts succeeds because that's how we can monitor builds later.

### PR Checklist

### To the reviewers 👀

#nochangelog

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -67,8 +67,6 @@ lane :ship_beta_ios do
 
   build_ios_app(configuration: 'Store', silent: true)
 
-  upload_sentry_artifacts(sentry_release_name: sentry_release_name, dist_version: bundle_version, platform: "ios")
-
   # Send to the app store
   beta_app_review_info = {
     contact_email: ENV['BETA_CONTACT_EMAIL'],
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -85,6 +83,9 @@ lane :ship_beta_ios do
         itc_provider: 'ArtsyInc',
         distribute_external: true,
         groups: ['Artsy People'])
+
+  upload_sentry_artifacts(sentry_release_name: sentry_release_name, dist_version: bundle_version, platform: "ios")
+
```
gkartalis, line 88: **question:** The pilot function I guess sends the app to the app store right? 

If yes I think that it will still result in a failure, but won't prevent us for "deploying" the beta which is what we wanted 😍

Same for the android one but the `pilot` is `supply` right?

gkartalis (approved):

fastlane/Fastfile
MounirDhahri: yes exactly, you are right

PR #6716, opened 2022-05-09
Enhancement(CX-2556): Standardize Toast Duration and Make RequestPriceEstimate Toast duration long

lordkiz:
This PR resolves [CX-2556]

### Description
- This PR standardizes the duration used by Toasts, providing `enum ToastDuration { Short = 2500, Long = 5000 }`
- Toast duration after Submission of Request Price Estimate changed to Long duration

### PR Checklist

### To the reviewers 👀

- Standardize Toast Duration and Make RequestPriceEstimate Toast duration long - kizito

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[docs]: /docs/README.md

[CX-2556]: https://artsyproduct.atlassian.net/browse/CX-2556?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Components/Toast/types.ts
+++ b/src/app/Components/Toast/types.ts
@@ -1,6 +1,11 @@
 import { ActionSheetProps } from "@expo/react-native-action-sheet"
 import { Color, IconProps } from "palette"
 
+export enum ToastDuration {
+  LONG = 5000,
+  SHORT = 2500,
+}
+
 export type ToastPlacement = "middle" | "top" | "bottom"
 
 export interface ToastOnPressHelpers {
@@ -18,6 +23,7 @@ export interface ToastDetails {
   onPress?: (helpers: ToastOnPressHelpers) => void
   Icon?: React.FC<IconProps>
   backgroundColor?: Color
+  duration?: ToastDuration | number
 }
 
-export type ToastOptions = Pick<ToastDetails, "onPress" | "Icon" | "backgroundColor">
+export type ToastOptions = Pick<ToastDetails, "onPress" | "Icon" | "backgroundColor" | "duration">
--- a/src/app/Components/Toast/types.ts
+++ b/src/app/Components/Toast/types.ts
@@ -1,7 +1,7 @@
 import { ActionSheetProps } from "@expo/react-native-action-sheet"
 import { Color, IconProps } from "palette"
 
-export type ToastDuration = "LONG" | "SHORT"
+export type ToastDuration = "long" | "short"
 
 export type ToastPlacement = "middle" | "top" | "bottom"
 
--- a/src/app/Components/Toast/types.ts
+++ b/src/app/Components/Toast/types.ts
@@ -1,10 +1,7 @@
 import { ActionSheetProps } from "@expo/react-native-action-sheet"
 import { Color, IconProps } from "palette"
 
-export enum ToastDuration {
-  LONG = 5000,
-  SHORT = 2500,
-}
+export type ToastDuration = "LONG" | "SHORT"
 
 export type ToastPlacement = "middle" | "top" | "bottom"
 
@@ -23,7 +20,7 @@ export interface ToastDetails {
   onPress?: (helpers: ToastOnPressHelpers) => void
   Icon?: React.FC<IconProps>
   backgroundColor?: Color
-  duration?: ToastDuration | number
+  duration?: ToastDuration
 }
 
 export type ToastOptions = Pick<ToastDetails, "onPress" | "Icon" | "backgroundColor" | "duration">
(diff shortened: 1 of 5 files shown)


dariakoko (dismissed):
LGTM! Thank you for the detailed tests 👏 👏 👏

src/app/Components/Toast/types.ts
```diff
@@ -18,6 +23,7 @@ export interface ToastDetails {
   onPress?: (helpers: ToastOnPressHelpers) => void
   Icon?: React.FC<IconProps>
   backgroundColor?: Color
+  duration?: ToastDuration | number
```
olerichter00, line 26: (non-blocking): I don't think we need to allow custom durations. Can we change this prop to:

```suggestion
  duration?: "short" | "long"
```

? Like this we wouldn't need to import an enum everywhere we use the component.

lordkiz: Thanks @olerichter00. I thought about that, and decided to allow custom because short and long are strictly ToastAndroid behaviour. If this would not be confusing to devs, I'm very happy to change it.

lordkiz: Done

src/app/Components/Toast/types.ts
```diff
@@ -1,6 +1,8 @@
 import { ActionSheetProps } from "@expo/react-native-action-sheet"
 import { Color, IconProps } from "palette"
 
+export type ToastDuration = "LONG" | "SHORT"
```
olerichter00, line 4:
```suggestion
export type ToastDuration = "long" | "short"
```

Let's keep it consistent with other props.

lordkiz: @olerichter00 What other props? Capitalizing duration denotes that it is a constant, and even if we are passing a string, anyone can read this and immediately understand that the string must have been mapped to a constant somewhere. In this case, where we are not using enums, uppercase strings conveys the same message. I think this is a standard naming convention in most languages (Or is it different in JS?)
Eg: [Java](https://www.oracle.com/java/technologies/javase/codeconventions-namingconventions.html#:~:text=The%20names%20of%20variables%20declared,%2C%20for%20ease%20of%20debugging.)
Also similar convention used here in React Native: [ToastAndroid](https://reactnative.dev/docs/toastandroid)

olerichter00: > Also similar convention used here in React Native: [ToastAndroid](https://reactnative.dev/docs/toastandroid)
Here the key is in capitalcase. The docs don't say anything about the content of it.

> Eg: [Java](https://www.oracle.com/java/technologies/javase/codeconventions-namingconventions.html#:~:text=The%20names%20of%20variables%20declared,%2C%20for%20ease%20of%20debugging.)
We're using TypeScript ;)

>What other props?
For example [here](https://github.com/artsy/eigen/blob/681d9a06058c6020161ca5427300150327788cd8/src/app/Components/Toast/types.ts#L4) and [here](https://github.com/artsy/eigen/blob/681d9a06058c6020161ca5427300150327788cd8/src/palette/Theme.tsx#L24) and pretty much everywhere in the codebase.

lordkiz: > For example [here](https://github.com/artsy/eigen/blob/681d9a06058c6020161ca5427300150327788cd8/src/app/Components/Toast/types.ts#L4) and [here](https://github.com/artsy/eigen/blob/681d9a06058c6020161ca5427300150327788cd8/src/palette/Theme.tsx#L24) and pretty much everywhere in the codebase.

😄 I see. @olerichter00 in the examples you have listed here, these props are used as is - i.e as the strings that they are or they map to another string, so it makes sense to use them as plain strings and they do not map to a fixed number constant. But by convention, when you map a string(key) to a final number constant you definitely should uppercase the name so it is obvious to everyone reading it, else to me there is no way I will ever think that "long" maps to a number.  It is not just a java thing. 
For eg: https://github.com/facebook/react-native/blob/main/Libraries/Components/ToastAndroid/NativeToastAndroid.js#L15

olerichter00: You can also use "long" as a key, I don't see any issue with that and would prefer to keep the API of the components consistent.

@MounirDhahri what do you think?

lordkiz: Will change it 👍 . Consistency is more important here I guess. A mixture of upper and lowercased props all over the place doesn't look clean. @olerichter00

olerichter00: Thanks a lot 🙇

olerichter00 (approved):
🚀

PR #6721, opened 2022-05-10
feat(NX-3133): Add Purchase button to conversations

araujobarret:
This PR resolves [NX-3133]

### Description

Adds Purchase button (hidden under feature flag) to conversations when it's `isAcquireable`. Covers unique and edition sets, since we re-used some boilerplate code from MOI it's already covering Edition Sets. Foreword in regards to some code duplication, the way we define the routes in Eigen is kinda funky, so abstracting to reuse the same modal component for example would be tricky and require some not type-safe prop drilling from the AppRegistry, so that's one of the main reasons.

Screenshot 2022-05-10 at 19 13 00

Screenshot 2022-05-10 at 18 45 23

cc @artsy/negotiate-devs 

### PR Checklist

### To the reviewers 👀

- add purchase button to conversations - araujobarret

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[docs]: /docs/README.md

[NX-3133]: https://artsyproduct.atlassian.net/browse/NX-3133?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -87,6 +87,7 @@
                 <data android:pathPrefix="/inquiry" />
                 <data android:pathPrefix="/map" />
                 <data android:pathPrefix="/make-offer" />
+                <data android:pathPrefix="/purchase" />
                 <data android:pathPrefix="/my-account" />
                 <data android:pathPrefix="/my-profile" />
                 <data android:pathPrefix="/my-profile/saved-addresses" />
--- a/src/app/Scenes/Artwork/Artwork.tsx
+++ b/src/app/Scenes/Artwork/Artwork.tsx
@@ -240,7 +240,11 @@ export const Artwork: React.FC<ArtworkProps> = ({
       return sections
     }
 
-    if (enableConversationalBuyNow && artworkAboveTheFold?.isAcquireable) {
+    if (
+      enableConversationalBuyNow &&
+      (artworkAboveTheFold?.isAcquireable ||
+        (!artworkAboveTheFold?.isInquireable && artworkAboveTheFold?.isOfferable))
+    ) {
       sections.push({
         key: "contactGallery",
         element: <Questions artwork={artworkBelowTheFold} />,
--- a/src/app/Scenes/Inbox/Components/Conversations/ConversationCTA.tsx
+++ b/src/app/Scenes/Inbox/Components/Conversations/ConversationCTA.tsx
@@ -1,4 +1,5 @@
 import { ConversationCTA_conversation } from "__generated__/ConversationCTA_conversation.graphql"
+import { useFeatureFlag } from "app/store/GlobalStore"
 import { extractNodes } from "app/utils/extractNodes"
 import React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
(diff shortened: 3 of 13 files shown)


android/app/src/main/AndroidManifest.xml
```diff
@@ -87,6 +87,7 @@
                 <data android:pathPrefix="/inquiry" />
                 <data android:pathPrefix="/map" />
                 <data android:pathPrefix="/make-offer" />
+                <data android:pathPrefix="/purchase" />
```
araujobarret, line 90: not sure if we really need to enable deep linking to this route, but added since we added for the others

src/app/Scenes/Artwork/Artwork.tsx
```diff
@@ -240,7 +240,11 @@ export const Artwork: React.FC<ArtworkProps> = ({
       return sections
     }
 
-    if (enableConversationalBuyNow && artworkAboveTheFold?.isAcquireable) {
+    if (
```
araujobarret, line 243: small fix because wasn't displaying the new cta for MO only artworks

src/app/Scenes/Inbox/Components/Conversations/ConversationCTA.tsx
```diff
@@ -13,68 +14,73 @@ interface Props {
...
+  const enableConversationalBuyNow = useFeatureFlag("AREnableConversationalBuyNow")
 
   if (liveArtwork?.__typename !== "Artwork") {
     return null
   }
 
   const isOfferableFromInquiry = liveArtwork?.isOfferableFromInquiry
-
-  let CTA: JSX.Element | null = null
+  const isOfferableConversationalBuyNow = liveArtwork?.isOfferable && enableConversationalBuyNow
```
araujobarret, line 24: we have 2 checks now, one for `isAcquireable` and the other for `isOfferable`, both glued with feature flag

src/app/Scenes/Inbox/Components/Conversations/PurchaseModal.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+export const PurchaseModalQueryRenderer: React.FC<{
+  artworkID: string
+  conversationID: string
+}> = ({ artworkID, conversationID }) => {
+  return (
+    <ProvideScreenTrackingWithCohesionSchema
+      info={{
+        action: ActionType.screen,
+        context_screen_owner_type: OwnerType.conversation,
```
araujobarret, line 127: @daytavares I guess this property here is not the correct as we discussed, we can have a follow up PR just to change it, for MOI we were using `OwnerType.conversationMakeOfferConfirmArtwork` on it, but not sure which one we should use here

jo-rs: Should I ticket this follow-up?

jo-rs (approved):
It generally looks good, and there's good test coverage 👍  
It's a lot, so hopefully, I didn't miss anything. I'll push it through since we want to test this, and things are behind a flag, but I'll do further testing/checking in the morning.

src/app/Scenes/Inbox/Components/Conversations/PurchaseModal.tsx
daytavares: @jo-rs better 😊

src/app/Scenes/Inbox/Components/Conversations/MakeOfferModal.tsx
```diff
@@ -98,6 +98,7 @@ export const MakeOfferModalFragmentContainer = createFragmentContainer(MakeOffer
       editionSets {
         internalID
         editionOf
+        isAcquireable
```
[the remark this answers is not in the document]
araujobarret: there's a violation in the way we use relay components here, we're prop drilling those props to the Edition Set component controller when it should be a Relay Fragment, and here we would have something like
```jsx
   ...editonSets_artwork
```
gonna create a follow-up PR, this PR is already too big, good catch!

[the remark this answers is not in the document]
araujobarret: addressed in #6730

PR #6729, opened 2022-05-11
feat(FX-3928): upload middleware for Relay

dimatretyak:
This PR resolves [FX-3928]

Related PR #6690 

### Problem
https://user-images.githubusercontent.com/3513494/167919013-00bd6a92-523c-4d2a-81a1-d281b01819dd.MP4

### Description
In beta build, when we try to upload a file, we get an error (it works for **development** environment).

This is due to the fact that `req.operation.text` is `null` in [upload middleware](https://github.com/relay-tools/react-relay-network-modern/blob/master/src/middlewares/upload.js#L15) from `react-relay-network-modern` for production environment. After a little investigation, it was discovered that we set the value for `req.operation.text` only for development environment in [persistedQueryMiddleware](https://github.com/artsy/eigen/blob/e49c702e264a186561b33711ab0d1f0dc658d68e/src/app/relay/middlewares/metaphysicsMiddleware.ts#L114) (because for production environment we send `documentID` and then by this id we extract query body on [Metaphysics](https://github.com/artsy/metaphysics/blob/main/src/lib/fetchPersistedQuery.ts) side).

But this solution will not work for upload flow and for this reason we need to [extract the query](https://github.com/artsy/eigen/blob/a3383c784329dcc1349414846734e70649bd3e57/src/app/relay/middlewares/uploadMiddleware.ts#L32-L36) by id from `complete.queryMap.json` in `upload` middleware.

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- upload middleware for Relay - dimatretyak

</details>

[docs]: /docs/README.md

[FX-3928]: https://artsyproduct.atlassian.net/browse/FX-3928?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/relay/middlewares/uploadMiddleware.ts
```diff
@@ -0,0 +1,61 @@
...
+ * TODO: Add support for Blob/File when problem will be fixed for Android
+ */
+export const uploadMiddleware = (): Middleware => {
+  return (next) => async (req) => {
+    if (req instanceof RelayNetworkLayerRequestBatch) {
+      throw new Error("RelayRequestBatch is not supported")
+    }
+
+    const operations = {
+      query: require("../../../../data/complete.queryMap.json")[req.id],
```
dimatretyak, line 26: is there any other way to get query from `req` here?

src/app/relay/middlewares/uploadMiddleware.ts
```diff
@@ -0,0 +1,61 @@
...
+    const { variables, id } = req
+    const result = extractFiles<ReactNativeFile>({ variables }, "", isReactNativeFile)
+    const { clone: extractedVariables, files } = result
+
+    if (files.size) {
+      if (!queryMap) {
+        queryMap = require("../../../../data/complete.queryMap.json")
+      }
+
+      const query = queryMap[id]
```
dimatretyak, line 36: Is such a solution suitable or is there any other way to get query from `req` here?

cc: @olerichter00

olerichter00: Have you tried `req.operation.query`? You can use `GraphQLRequest` as a type for `req` (like [here](https://github.com/artsy/eigen/blob/main/src/app/relay/middlewares/cacheMiddleware.ts#L8)).

dimatretyak: > Have you tried req.operation.query? 

Yep, I tried. Unfortunately, it's `undefined` 😢

demo

olerichter00: And if you look for `req.operation.query` instead of `req.query` or inspect `req` to see how it looks like?

dimatretyak: Ah, sorry, my bad. I looked, I just wrote debug label incorrectly

```
console.log("[debug] req.query", req.operation.query)
```

dimatretyak: > inspect req to see how it looks like

Inspected `req`, but unfortunately there is nothing there either

olerichter00: Oh true, these requests don't send just the query ID and not the query itself. Looking at the `queryMap` makes sense then 👍

olerichter00 (approved):
Looking good 👐

src/app/relay/middlewares/uploadMiddleware.tests.ts
```diff
@@ -0,0 +1,152 @@
+import { ReactNativeFile } from "extract-files"
+import { RelayNetworkLayerResponse } from "react-relay-network-modern/node8"
+import { GraphQLRequest } from "./types"
+import { uploadMiddleware } from "./uploadMiddleware"
+
+describe("uploadMiddleware", () => {
```
MounirDhahri, line 6: **praise:** thanks for the thourough tests

src/app/relay/middlewares/uploadMiddleware.ts
MounirDhahri, line 36: great to see you using the querymap here!

PR #6656, opened 2022-04-21
feat(BX-211): Add details sale info page

annacarey:
This PR resolves [BX-211](https://artsyproduct.atlassian.net/browse/BID-211)

### Description

*Important Note:*
Do not review this PR yet. I've decided to break out the absolute vs. relative sale info to try to get it into the next release. The relative time info part is more complex and isn't quite working properly yet. ~~Here's the PR with the first portion of the work here: https://github.com/artsy/eigen/pull/6657~~ <-- this is merged!

Add new sale details for cascading end time feature that includes the "Lots close at x-minute intervals" label, the absolute time info, and the relative time info.

<table>
  <tr>
    <td>Sale starts today:</td>
     <td>Sale starts in 1-2 days:</td>
     <td>Sale starts the next day but less than 24 hours from now:</td>
  </tr>
  <tr>
    <td valign="top"></td>
    <td valign="top"></td>
    <td valign="top"></td>
  </tr>
 </table>

<table>
  <tr>
    <td>Sale is open and ends in 1-2 days:</td>
     <td>Sale is open and ends in < 1 hour:</td>
     <td>Lots are closing</td>
  </tr>
  <tr>
    <td valign="top"></td>
    <td valign="top"></td>
    <td valign="top"></td>
  </tr>
 </table>

Sale starts today:
Screen Shot 2022-04-21 at 4 34 00 PM

Sale starts in 1-2 days:
Screen Shot 2022-04-21 at 4 35 03 PM

Sale starts in 2+ days:
Screen Shot 2022-04-21 at 4 35 50 PM

Sale is open and ends in 2+ days:
Screen Shot 2022-04-21 at 4 37 20 PM

Sale is open and ends in 1-2 days:
Screen Shot 2022-04-21 at 4 42 01 PM

Sale is open and ends in <1 day:
Screen Shot 2022-04-21 at 4 44 01 PM

Sale is open and ends in < 1 hour:
Screen Shot 2022-04-21 at 4 46 09 PM

Sale is open and lots are closing:
Screen Shot 2022-04-21 at 4 48 33 PM

### PR Checklist

- [x] I tested my changes on **iOS** / **Android**.
- [x] I added screenshots or videos to illustrate my changes.
- [x] I added Tests and Stories for my changes.
- [ ] I added an [app state migration].
- [x] I hid my changes behind a [feature flag].

### To the reviewers 👀

- [ ] I would like **at least one** of the reviewers to run this PR on the simulator or device.

<details><summary>Changelog updates</summary>

[...]

PR #6751, opened 2022-05-16
fix(BID-356): Lot Page in Eigen not updating status

lordkiz:
This PR resolves [BID-356]

### Description
Prefetching Auctions + associated `environment.retain(operation)` allows for stale records in auction related records. It is  essential that these records are kept fresh at all times. This PR removes prefetching for these sales.

### PR Checklist

### To the reviewers 👀

- Fix Lot Page in Eigen not updating status - kizito

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[BID-356]: https://artsyproduct.atlassian.net/browse/BID-356?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/AppRegistry.tsx
+++ b/src/app/AppRegistry.tsx
@@ -97,10 +97,10 @@ import { PartnerQueryRenderer } from "./Scenes/Partner"
 import { PartnerLocationsQueryRenderer } from "./Scenes/Partner/Screens/PartnerLocations"
 import { PrivacyRequest } from "./Scenes/PrivacyRequest"
 import { ReverseSearchImageResults } from "./Scenes/ReverseSearchImageResults/ReverseSearchImageResults"
-import { SaleQueryRenderer, SaleScreenQuery } from "./Scenes/Sale"
+import { SaleQueryRenderer } from "./Scenes/Sale"
 import { SaleFAQ } from "./Scenes/SaleFAQ/SaleFAQ"
 import { SaleInfoQueryRenderer } from "./Scenes/SaleInfo"
-import { SalesQueryRenderer, SalesScreenQuery } from "./Scenes/Sales"
+import { SalesQueryRenderer } from "./Scenes/Sales"
 import { SavedAddressesQueryRenderer } from "./Scenes/SavedAddresses/SavedAddresses"
 import { SavedAddressesFormQueryRenderer } from "./Scenes/SavedAddresses/SavedAddressesForm"
 import { EditSavedSearchAlertQueryRenderer } from "./Scenes/SavedSearchAlert/EditSavedSearchAlert"
@@ -328,8 +328,8 @@ export const modules = defineModules({
   ArtworkMedium: reactModule(ArtworkMediumQueryRenderer),
   ArtworkAttributionClassFAQ: reactModule(ArtworkAttributionClassFAQQueryRenderer),
   ArtworkSubmissionStatusFAQ: reactModule(ArtworkSubmissionStatusFAQ),
-  Auction: reactModule(SaleQueryRenderer, { fullBleed: true }, [SaleScreenQuery]),
-  Auctions: reactModule(SalesQueryRenderer, {}, [SalesScreenQuery]),
+  Auction: reactModule(SaleQueryRenderer, { fullBleed: true }),
+  Auctions: reactModule(SalesQueryRenderer),
   AuctionInfo: reactModule(SaleInfoQueryRenderer),
   AuctionFAQ: reactModule(SaleFAQ),
   AuctionResult: reactModule(AuctionResultQueryRenderer),


src/app/AppRegistry.tsx
```diff
@@ -328,8 +328,8 @@ export const modules = defineModules({
   ArtworkMedium: reactModule(ArtworkMediumQueryRenderer),
   ArtworkAttributionClassFAQ: reactModule(ArtworkAttributionClassFAQQueryRenderer),
   ArtworkSubmissionStatusFAQ: reactModule(ArtworkSubmissionStatusFAQ),
-  Auction: reactModule(SaleQueryRenderer, { fullBleed: true }, [SaleScreenQuery]),
-  Auctions: reactModule(SalesQueryRenderer, {}, [SalesScreenQuery]),
+  Auction: reactModule(SaleQueryRenderer, { fullBleed: true }),
+  Auctions: reactModule(SalesQueryRenderer),
```
olerichter00, line 332: Removing screen queries from the app registry prevents prefetching, but also results in warnings in the console because prefetching simply fails. If you want to disable prefetching for these screens you should look for places where the URL is prefetched.

olerichter00 (changes requested):
I don't think this would fix the issue. Even though the queries are prefetched, entering the screen will trigger another request which should update the data in the store. 

You might want to add `force: true` to [SalesQueryRenderer](https://github.com/artsy/eigen/blob/5fbad7d8bb5420e5cf5730c32eb9e974bf89079c/src/app/Scenes/Sales/index.tsx#L126)'s cache config to make sure it's not using stale data.

lordkiz:
> I don't think this would fix the issue. Even though the queries are prefetched, entering the screen will trigger another request which should update the data in the store.
> 
> You might want to add `force: true` to [SalesQueryRenderer](https://github.com/artsy/eigen/blob/5fbad7d8bb5420e5cf5730c32eb9e974bf89079c/src/app/Scenes/Sales/index.tsx#L126)'s cache config to make sure it's not using stale data.

Thanks @olerichter00 . In another PR https://github.com/artsy/eigen/pull/6656/files#diff-3aabde61aed1fad2f6add85d35ce07c298ebe6ced62fbe4fa06df98c2a5585e7R92 , I have modified this to useLazyLoadQuery which has `force: true` as default. The persistence of stale records was greatly reduced.

src/app/AppRegistry.tsx
lordkiz: @olerichter00 I can see that the queries are optional. Passing empty array would also work but it is dirty. I do not understand what you mean by `where the URL is prefetched`. My understanding is that `environment.retain` would persist the data we get from these queries.

olerichter00:
Shall we close this one then?

lordkiz:
> Shall we close this one then?
@olerichter00 I won't say that. Individual Auction pages do not have stale records when I removed the prefetch query. As screens that would heavily depend on realtime data in coming PRs, prefetching on them would have undesired effects. Since the queries are optional, I removed them and I don't see any warnings. Maybe i am missing something.

olerichter00:
> prefetching on them would have undesired effects. 

As I explained above, it wouldn't have any undesired effect because the query gets fired again as soon as the user enters the screen.

lordkiz:
> As I explained above, it wouldn't have any undesired effect because the query gets fired again as soon as the user enters the screen.

@olerichter00 Thanks for explaining. Before closing this, I want to be certain: as long as I have `force: true`, I do not need to worry about `environment.retain(data)` because the data will be updated no matter what? I am finding it difficult to ascertain this on the UI because it is not straightforward to observe the effects of relay caching.

olerichter00:
Maybe this doc about [Relay Fetch Policies](https://relay.dev/docs/guided-tour/reusing-cached-data/fetch-policies/) might help to understand how it works. `store-and-network` is the fetch policy you want to use here in combination with the cache config `force: true`. 

The current caching setup is quite complicated (Relay store + additional network caching middleware). But these two configs will guarantee you fresh data.

lordkiz:
> Maybe this doc about [Relay Fetch Policies](https://relay.dev/docs/guided-tour/reusing-cached-data/fetch-policies/) might help to understand how it works. `store-and-network` is the fetch policy you want to use here in combination with the cache config `force: true`.
> 
> The current caching setup is quite complicated (Relay store + additional network caching middleware). But these two configs will guarantee you fresh data.
> 
> Even without query prefetching, you could run into a situation where the user sees the data from the cache when coming back to a screen.

Fair enough. Thanks @olerichter00

lordkiz:
Closing in favour of used `force: true` config in this PR:  https://github.com/artsy/eigen/pull/6656/files#diff-3aabde61aed1fad2f6add85d35ce07c298ebe6ced62fbe4fa06df98c2a5585e7R92

lordkiz:
@olerichter00 Reopening this PR. The issue is not improved. It improves with removal of prefetching for these screens.

olerichter00:
@lordkiz ok, then let's deactivate prefetching for these two screens for now. Could you please let me know how to reproduce the bug and which screen hasn't been improved when changing the cache config? I'd like to investigate further to make sure this doesn't affect other screens.

src/app/AppRegistry.tsx
olerichter00: We use [prefetchUrl()](https://github.com/artsy/eigen/blob/6020f328d87260b33efe6da9c0bf773d23c721e7/src/app/utils/queryPrefetching.ts#L57) to prefetch queries for given URLs. This helper uses the routing algorithm to find the correct query to prefetch for a given URL. To prevent prefetching the lot/auction/sale screen, you'd probably want to remove [these](https://github.com/artsy/eigen/blob/6020f328d87260b33efe6da9c0bf773d23c721e7/src/app/Scenes/Home/Components/SalesRail.tsx#L79) two lines.

lordkiz: @olerichter00 This won't fix the general issue as this would prefetch the individual sale on that rail only. Queries passed as part of Module props in AppRegistry are injected into the Component and at app start these queries are prefetched. We can safely remove the console.errors when `!module.Queries`, right?

lordkiz:
> @lordkiz ok, then let's deactivate prefetching for these two screens for now. Could you please let me know how to reproduce the bug and which screen hasn't been improved when changing the cache config? I'd like to investigate further to make sure this doesn't affect other screens.

Sure. For eg: Modifying an auction you have opened after the app is already running. Values remain stale even after reopening the auction page. You might need to reload the app to see the updated state.

src/app/AppRegistry.tsx
olerichter00: [These](https://github.com/artsy/eigen/blob/6020f328d87260b33efe6da9c0bf773d23c721e7/src/app/Scenes/Home/Home.tsx#L78) are the only queries that are prefetched when the app starts. I don't know of any other mechanism that would prefetch queries at app start.

olerichter00:
> Sure. For eg: Modifying an auction you have opened after the app is already running. Values remain stale even after reopening the auction page. You might need to reload the app to see the updated state.

Thanks! Just one detail, how exactly are you reopening the auction page?

lordkiz:
> Thanks! Just one detail, how exactly are you reopening the auction page? I couldn't reproduce the bug like this, but maybe I just did it differently.

@olerichter00 By just going back to the auction page

olerichter00:
> @olerichter00 By just going back to the auction page

But back from where? There is for example a difference if click the back button and then the lot again in the rail or if you use the bottom navigation. The first will re-render the screen (and reload the data) and the second won't. Maybe you could send a video or give clear steps on how to reproduce it?

lordkiz:
> But back from where? There is for example a difference if you click the back button and then the lot again in the rail or if you use the bottom navigation. The first will re-render the screen (and reload the data) and the second won't. Maybe you could send a video or give clear steps on how to reproduce it?

@olerichter00 
-From Homepage hit View All to land on auctions page.
 -Open an auction
-Open an artwork in the auction
-Go to auction-staging and edit the auction
-Go back all the way to Home.
-Click View all again
-Inspect the auctions: data is stale
- Also if you edit the auction while still on Homepage without ever going to the Auctions page at all. When you open the Auctions page and select an auction and inspect edited auction, you'd see the data is stale because it was fetched before you opened it.

olerichter00:
I've adjusted the relay config for both screens (Sales/Auctions & Artwork) to the following in https://github.com/artsy/eigen/pull/6767 which fixed the issue for me. This will also make sure that even if a user revisits one of the screens, the data will be reloaded.

```
{
  fetchPolicy: "store-and-network",
  networkCacheConfig: { force: true },
}
```

PR #6768, opened 2022-05-18
feat(CX-2565): add my collection insights empty state

MounirDhahri:
This PR resolves [CX-2565]

### Description
- add my collection insights empty state
- Refactor tests that were not properly made

Simulator Screen Shot - iPhone 13 - 2022-05-18 at 18 03 22

### PR Checklist

### To the reviewers 👀

- Add my collection empty state - mounir

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-2565]: https://artsyproduct.atlassian.net/browse/CX-2565?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Artsy.xcodeproj/project.pbxproj
```diff
@@ -3288,6 +3288,8 @@
 				"${PODS_ROOT}/Artsy+UIFonts/Pod/Assets/Unica77LL-Regular.otf",
 				"${PODS_ROOT}/Artsy+UILabels/Pod/Assets/Chevron_Black.png",
 				"${PODS_ROOT}/Artsy+UILabels/Pod/Assets/Chevron_Black@2x.png",
+				"${PODS_ROOT}/../emission/Pod/Assets/Emission.js",
+				"${PODS_ROOT}/../emission/Pod/Assets/assets",
```
MounirDhahri, line 3292: @pvinis I am getting those added after I run `pod install` and build the project. Am I doing something wrong here or is there an issue with my setup?

MounirDhahri, line 3292: @pvinis @brainbicycle  I am getting those added after I run pod install and build the project. Am I doing something wrong here or is there an issue with my setup?

dariakoko (approved):
Looks good!
And thanks for refactoring my code 😅

dariakoko (dismissed):

src/app/Scenes/MyCollection/Screens/Insights/MyCollectionInsights.tsx
```diff
@@ -1,47 +1,96 @@
...
+        paddingTop: space("2"),
+      }}
+    >
+      {myCollectionArtworksCount > 0 ? renderContent() : <MyCollectionInsightsEmptyState />}
     </StickyTabPageScrollView>
   )
 }
 
+const MarketSignalsSectionHeader = () => {
+  const [shouldShowarketSignalsModal, setShouldShowarketSignalsModal] = useState(false)
```
olerichter00, line 53:
```suggestion
  const [shouldShowMarketSignalsModal, setShouldShowMarketSignalsModal] = useState(false)
```

olerichter00 (dismissed):
Looks good! I've just added a few small comments.

src/app/Scenes/MyCollection/Screens/Insights/MyCollectionInsights.tsx
```diff
@@ -1,47 +1,96 @@
...
+      style={{
+        flex: 1,
+      }}
+      contentContainerStyle={{
+        flexGrow: 1,
+        justifyContent: "center",
+        paddingTop: space("2"),
+      }}
+    >
+      {myCollectionArtworksCount > 0 ? renderContent() : <MyCollectionInsightsEmptyState />}
```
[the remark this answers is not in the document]
MounirDhahri: I am not aware of the conversation, is there a main reason behind that? the main reason why I prefer going with methods inside of functional components is the amount of flexibility I get with it when I want to. Let me explain:

In the case where `renderContent` is a method that would require a lot of computing, I would be able to wrap it immediately with a `useCallback` and add the required optimizations to its dependencies array to make sure it only updates when necessary. The same could be achieved in the case of a component but wrapping it with a memo is not a pattern I see often in projects and that's why I didn't use it. 

I am glad to change this as well for consistency reasons if the entire strongly prefers the other approach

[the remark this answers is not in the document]
MounirDhahri: > Wouldn't React do this optimization automatically if you use <Content {...whateverProps} />
True, React would do this optimization when the component is defined outside of the component and not inside it. Was this what you meant already? having the component separated already 😄? if so then yeah, I am glad to extract it outside of the component. 

> and my understanding is that useCallback should be used within the component doing the heavy computation (in this case MyCollectionInsights should not contain logic that depends on what is inside Content).
True as well. I do tend though to only use `useCallback` when I notice a performance issue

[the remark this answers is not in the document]
olerichter00: > True, React would do this optimization when the component is defined outside of the component and not inside it. Was this what you meant already? having the component separated already 😄? if so then yeah, I am glad to extract it outside of the component.

Oh yeah, good point. If you define the component inside of the parent component it gets regenerated on every render. I was assuming the component was defined outside.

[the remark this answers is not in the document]
MounirDhahri: cool then, that makes sense. I don't always do that because a lot of the time I end up doing some prop drilling so I got used to only extract the component outside when I am close to sure it's not going to need much props from its parent component later. In the case of this component, I will have to be sending  `auctionResults` from `data.me.auctionResultsConnection` and we will be sending later the other props from the other sections and that's why I left it inside the component.

Artsy.xcodeproj/project.pbxproj
pvinis: these are safe to ignore and not commit. once we remove emission it should be ok. could you remove these from the PR? 🙏

src/app/Components/StickyTabPage/testHelpers.tsx
```diff
@@ -0,0 +1,13 @@
+import React from "react"
+import { StickyTabPage } from "./StickyTabPage"
+
+export const withStickyTabPage = (WrappedComponent: React.FC<any>) => (
```
pvinis, line 4: heh. interesting HOC 👍

pvinis (approved):

olerichter00 (approved):

src/app/Components/StickyTabPage/testHelpers.tsx
MounirDhahri: we have this in my screens with the `stickytabpage` in order to access the `StickyTabPage` context and I thought I can just add it as a helper maybe

Artsy.xcodeproj/project.pbxproj
MounirDhahri: yeah of course

PR #6784, opened 2022-05-20
chore: Clear relay cache when app is updated

lordkiz:
This PR resolves []

### Description
- Context: https://artsy.slack.com/archives/C01B2P6LJUU/p1653042937308069
- Clears the Relay cache when a user moves from one version to another, thus keeping cache fresh across versions.
- 

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- Clear relay cache when app is updated - kizito

</details>

### DIFF
--- a/src/app/relay/usePurgeCacheOnAppUpdate.ts
+++ b/src/app/relay/usePurgeCacheOnAppUpdate.ts
@@ -1,0 +1,19 @@
+import AsyncStorage from "@react-native-async-storage/async-storage"
+import { RelayCache } from "app/relay/RelayCache"
+import { useEffect } from "react"
+import { getVersion } from "react-native-device-info"
+
+const APP_CURRENT_VERSION_KEY = "EIGEN_APP_CURRENT_VERSION_KEY"
+
+/** Clears the relay cache when a user runs their app for the first time after updating the app */
+export const usePurgeCacheOnAppUpdate = () => {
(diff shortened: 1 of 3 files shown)


src/app/relay/usePurgeCacheOnAppUpdate.ts
```diff
@@ -0,0 +1,19 @@
+import AsyncStorage from "@react-native-async-storage/async-storage"
+import { RelayCache } from "app/relay/RelayCache"
+import { useEffect } from "react"
+import { getVersion } from "react-native-device-info"
+
+const APP_CURRENT_VERSION_KEY = "EIGEN_APP_CURRENT_VERSION_KEY"
+
+/** Clears the relay cache when a user runs their app for the first time after updating the app */
+export const usePurgeCacheOnAppUpdate = () => {
```
pvinis, line 9: that makes sense. I wonder if we could instead, move that clear all call to the migration.ts file, as a step that always happens before a migration?

is it after every migration we want this or actually any app update in general?

lordkiz: @pvinis We want this after an app update, not migration.

pvinis (approved):

olerichter00 (commented):
Looks good 👏

PR #6815, opened 2022-05-25
fix(FX-3988): partners section for auctions should not be pressable

gkartalis:
This PR resolves [FX-3988]

### Description

Context: 

Auction partners usually don't have a page, and we have a property under Partner in graphql called `isLinkable` that returns falsy value if the partner is not linkable.

This pr utilizes this to show just a text with the auction partner name when this is the case.

- ⚠️ Changes are behind `AREnableCreateArtworkAlert` feature flag. ⚠️
- partner should not be pressable if `isLinkable` is falsy
- most auction partners are not linkable

Came up during CreateArtworkAlert QA Session.

#### Screenshots

|auction partner|auction partner|gallery partner|
|---|---|---|
|Simulator Screen Shot - iPhone 13 - 2022-05-25 at 10 40 19|Simulator Screen Shot - iPhone 13 - 2022-05-25 at 10 40 56|simulator_screenshot_8569805D-CDC6-4844-9A0F-BCDDE9E44101|

### PR Checklist

### To the reviewers 👀

- partners section for auctions should not be pressable - gkartalis

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[FX-3988]: https://artsyproduct.atlassian.net/browse/FX-3988?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/Scenes/Artwork/Artwork.tsx
```diff
@@ -252,11 +252,13 @@ export const Artwork: React.FC<ArtworkProps> = ({
       })
     }
 
-    if (!!partner?.href && !!partner?.name && enableCreateArtworkAlert) {
+    if (enableCreateArtworkAlert && !!partner?.href && !!partner?.name) {
```
dimatretyak, line 255: maybe we shouldn't check for `href` if the partner is not clickable? and in this case, is it better to put all the code in a separate component?

gkartalis: Good point, was thinking to pull this out to a component as well, was waiting first to see if some tests will fail and will also add some extra test cases. But not 100% sure for the `href` 🤔 

Seems to me that even the auction partners have a href, although it doesn't redirect us anywhere, but happy to refactor it 🤘

gkartalis: @dimatretyak decided not to move the section it's own component since it is a small one and I think that moving it to a component will add extra work to go there and find out what it does 🤔

PR #6907, opened 2022-06-09
fix: links should open a webview when not assigned to a native screen

pvinis:
This PR resolves []

### Description

trying to open https://www.artsy.net/art-appraisals was going to a 404 in the app. after this, it will open a webview.

we also need to put that link to the blocklist so it doesn't open the app, but its a good default to not go to 404 when we have forgotten to block the next one.

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Scenes/VanityURL/VanityURLEntity.tsx
+++ b/src/app/Scenes/VanityURL/VanityURLEntity.tsx
@@ -1,6 +1,5 @@
 import { VanityURLEntity_fairOrPartner$data } from "__generated__/VanityURLEntity_fairOrPartner.graphql"
 import { VanityURLEntityQuery } from "__generated__/VanityURLEntityQuery.graphql"
-import { ArtsyWebView } from "app/Components/ArtsyWebView"
 import { HeaderTabsGridPlaceholder } from "app/Components/HeaderTabGridPlaceholder"
 import { defaultEnvironment } from "app/relay/createEnvironment"
 import { renderWithPlaceholder } from "app/utils/renderWithPlaceholder"
@@ -61,10 +60,6 @@ interface RendererProps {
 }
 
 export const VanityURLEntityRenderer: React.FC<RendererProps> = ({ entity, slugType, slug }) => {
-  if (slugType === undefined) {
-    return <ArtsyWebView url={`/${slug}`} />
-  }
-
   if (slugType === "fairID") {
     return <FairQueryRenderer fairID={slug} />
   } else {
@@ -81,6 +76,7 @@ export const VanityURLEntityRenderer: React.FC<RendererProps> = ({ entity, slugT
         `}
         variables={{ id: slug }}
         render={renderWithPlaceholder({
+          showNotFoundView: false,
           renderFallback: () => <VanityURLPossibleRedirect slug={slug} />,
           renderPlaceholder: () => {
             switch (entity) {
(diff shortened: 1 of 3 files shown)


src/app/Components/ArtsyReactWebView.tsx
```diff
@@ -1,361 +0,0 @@
-import { OwnerType } from "@artsy/cohesion"
```
pvinis, line 1: i moved this file to `ArtsyWebView.*`

src/app/Scenes/SavedSearchAlert/screens/EmailPreferencesScreen.tsx
```diff
@@ -16,7 +16,7 @@ export const EmailPreferencesScreen: React.FC<Props> = (props) => {
   return (
     <Box flex={1}>
       <FancyModalHeader hideBottomDivider onLeftButtonPress={handleLeftButtonPress} />
-      <ArtsyReactWebView url="/unsubscribe" />
```
pvinis, line 19: `ArtsyReactWebView` was the webview, and `ArtsyWebView` was the webview with a fancymodal and a header.

Now `ArtsyWebView` is the webview, and `ArtsyWebViewPage` is the one with the header.

src/app/Scenes/VanityURL/VanityURLEntity.tsx
```diff
@@ -54,6 +55,10 @@ interface RendererProps {
 }
 
 export const VanityURLEntityRenderer: React.FC<RendererProps> = ({ entity, slugType, slug }) => {
+  if (slugType === undefined) {
```
pvinis, line 58: main change here

src/app/Components/ArtsyWebView.tsx
```diff
@@ -1,6 +1,361 @@
...
+import { Platform } from "react-native"
+import Share from "react-native-share"
+import WebView, { WebViewProps } from "react-native-webview"
+import { useTracking } from "react-tracking"
+import { useScreenDimensions } from "shared/hooks"
+import { ArtsyKeyboardAvoidingView } from "shared/utils"
+import { parse as parseURL } from "url"
+import { FancyModalHeader } from "./FancyModal/FancyModalHeader"
 
-export const ArtsyWebView: React.FC<{ url: string }> = ({ url }) => {
```
pvinis, line 4: this file was just a wrapper. i moved the whole thing here.

src/app/Components/ArtsyWebView.tests.tsx
```diff
@@ -1,12 +1,306 @@
...
+}))
 
 describe("ArtsyWebView", () => {
   it("shows react-native-webview", () => {
     const tree = renderWithWrappers(<ArtsyWebView url="random-url" />)
     expect(tree.root.findAllByType(WebView).length).toEqual(1)
   })
 })
+
+describe("ArtsyWebViewPage", () => {
```
pvinis, line 42: keeping the one test we had, plus adding all the ones from the wrapped component.

brainbicycle (approved):
Thanks for the fix! 
nitpick (non-blocking): I think the cleanup code could have been a separate PR?

pvinis:
>  nitpick (non-blocking): I think the cleanup code could have been a separate PR?
its true.. shall it? i should I guess

pvinis:
done

brainbicycle:
I made a little modification to the behavior here, I think this broke when we introduced the custom not found view for 404 errors. What should happen in the case of a bare url like this: we try to render VanityURL component, this first tries to see if there is a profile associated with the url (partner or gallery), if there isn't (404) we render the VanityURLRedirect component which will see if there is a force redirect for that url, if there is a redirect we render whatever component that redirect points to, if there isn't we open the url in a webview. Since the error that is supposed to show the VanityURLRedirect is a 404 this was not happening and instead we were showing the custom 404 view. I added a param to opt out of this behavior just for this component, I think for most components it is what we want this is just a bit of a special case.

olerichter00 (approved):
Thanks for the fix 🙇

brainbicycle:
here are the two problem links working:

https://user-images.githubusercontent.com/49686530/173826342-f6f1dbfd-26ab-48ff-92de-03305d005096.mov

PR #6932, opened 2022-06-15
fix: fix onboarding links with new app shell

brainbicycle:
This PR resolves [MOPLAT-301]

### Description

Fixes navigating to onboarding links in new app shell on iOS.
Previously we had custom logic on iOS that used our native navigation that no longer exists, now uses same logic as Android.
Not sure if it is something that broke recently but the back action in these webviews from onboarding was not working since it has its own nav stack. 

- Add a little wrapper webview to get white bg header and handle back button with onboarding nav
- Replace individual webviews in nav stack with this custom webview and pass route when calling

<details><summary>Screenshots</summary>

Uploading low-qual-onboarding-links-no-audio.mov…

</details>

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

- Fix links in onboarding with new app shell - Brian 

-

#### Dev changes

-

</details>

[MOPLAT-301]: https://artsyproduct.atlassian.net/browse/MOPLAT-301?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/Scenes/Onboarding/OnboardingWebView.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export interface OnboardingWebViewProps
+  extends StackScreenProps<OnboardingNavigationStack, "OnboardingWebView"> {}
+
+export type OnboardingWebViewRoute = "/privacy" | "/terms" | "/conditions-of-sale"
+
+export const OnboardingWebView: React.FC<OnboardingWebViewProps> = ({ route, navigation }) => {
+  return (
+    <Flex flex={1} style={{ backgroundColor: "white" }}>
+      <ArtsyWebViewPage url={route.params.url} backAction={() => navigation.goBack()} />
```
pvinis, line 14: is just `ArtsyWebViewPage` not doing the right thing? I wonder if it makes sense to put that `Flex` inside the ArtsyWebViewPage if its a better way to display it 🤔. or maybe separate like here is ok? I dont have a suggestion besides the question..

What do you think?

pvinis (approved):
besides the question, this seems right. ill assign back to you, and merge when you want!

src/app/Scenes/Onboarding/OnboardingWebView.tsx
brainbicycle: I am actually not sure, the background appeared gray without this, maybe because there is a transparent background in onboarding? I can try moving that flex to the webview page though, think we want the white background in all cases anyway?

src/app/Scenes/Onboarding/Onboarding.tsx
```diff
@@ -95,13 +94,8 @@ export const OnboardingWelcomeScreens = () => {
           component={OnboardingCreateAccountWithEmail}
         />
         <StackNavigator.Screen name="OnboardingSocialLink" component={OnboardingSocialLink} />
-        <StackNavigator.Screen
-          name="ForgotPassword"
-          component={ForgotPassword}
-          options={{ headerShown: false }}
```
brainbicycle, line 101: this is default behavior specified above, so don't think we need to pass this option

PR #6982, opened 2022-06-24
fix(CX-2172): refresh control rendering too high in MyC, MyC Insights, Saved

dariakoko:
This PR resolves [CX-2172]

### Description

The issue occurred, because the refresh control is rendered higher than the `ListHeaderComponent` inside the `<AnimatedFlatList />`. `ListHeaderComponent` has a height set to `staticHeaderHeight`, pushing the refresh control up.

To fix this I've moved `<RefreshControl />` inside `<StickyTabPageFlatList />`  and added `progressViewOffset` property to it. This property should have the same value as the ListHeaderComponent's height (staticHeaderHeight). However, `staticHeaderHeight` here has the type `Animated.Node<number>`, but for the refresh control, I needed `number`. I didn't find a way to cast `staticHeaderHeight` to the type I need (_is it possible?_), soooo instead I'm getting the height of the header from the `<MyProfileHeaderMyCollectionAndSavedWorks />` component and passing it to the `<StickyTabPageFlatList />`.

|iOS | Android |
|---|---|
|https://user-images.githubusercontent.com/36167539/175568883-fbf56a6e-5408-4e2e-a5d7-58e603d82b3d.mp4 | android refresh control |

### PR Checklist

### To the reviewers 👀

- fix refresh control in My Collection - daria

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-2172]: https://artsyproduct.atlassian.net/browse/CX-2172?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dariakoko:
Note: This probably makes [the issue](https://artsyproduct.atlassian.net/browse/CX-2631) we've discussed at the Refinement meeting even more complicated.

src/app/Components/StickyTabPage/StickyTabPage.tsx
```diff
@@ -35,6 +35,7 @@ interface StickyTabPageProps {
   // disableBackButtonUpdate allows the original BackButton visibility state. Useful when using StickyTabPage
   // as a root view where you don't want BackButton to ever be visible.
   disableBackButtonUpdate?: boolean
+  refreshControllPadding?: number
```
lordkiz, line 38: This might be better named `refreshControlProgressViewOffset`.

src/app/Components/StickyTabPage/StickyTabPageFlatList.tsx
```diff
@@ -93,11 +96,19 @@ export const StickyTabPageFlatList: React.FC<StickyTabFlatListProps> = (props) =
...
+  const { data, style, refresh, isRefreshing, ...otherProps } = props
 
   return (
     <Animated.View style={{ flex: 1, paddingTop: totalStickyHeaderHeight }}>
       <AnimatedFlatList
+        refreshControl={
+          <RefreshControl
+            progressViewOffset={refreshControllPadding ?? 0}
+            refreshing={!!isRefreshing}
+            onRefresh={refresh || null || undefined}
```
lordkiz, line 108:
```suggestion
            onRefresh={refresh}
```
Because onRefresh should be `() => void | undefined`

dariakoko: You're right, thanks!

src/app/Components/StickyTabPage/StickyTabPageFlatList.tsx
```diff
@@ -93,11 +96,19 @@ export const StickyTabPageFlatList: React.FC<StickyTabFlatListProps> = (props) =
   // to avoid jumping content in situations where certain items have fixed height.
   // This doesn't make a lot of sense but sometimes you just go with what works ¯\_(ツ)_/¯
   const [headerDidMount, setHeaderDidMount] = useState(__TEST__)
-  const { data, style, ...otherProps } = props
+  const { data, style, refresh, isRefreshing, ...otherProps } = props
 
   return (
     <Animated.View style={{ flex: 1, paddingTop: totalStickyHeaderHeight }}>
       <AnimatedFlatList
+        refreshControl={
```
lordkiz, line 104: Instead of explicitly adding refreshControl here, why not let the components that needs refreshControl pass the refreshControl prop as part of Flatlist props (which will be passed as ...otherProps)? The reason is that this customizes StickyTabPageFlatList too strictly and assumes we want refreshControl every time we use it. Therefore, any time a user does a drag down gesture on this component, a refresh control is shown even if we are not refreshing or the data does not need refreshing.

lordkiz (commented):
Great job @dariakoko . Left a few comments.

MrSltun (commented):
Thank you Daria! 🌟  
I just left a few comments :)

src/app/Components/StickyTabPage/StickyTabPageFlatList.tsx
MrSltun: I prefer the way you suggested, but my only concern is the refreshing feature would appear even when the `refresh` is not being used, maybe we could try it and see if it works

dariakoko: if I use ```...**refresh &&** (
            <RefreshControl...``` it wont

lordkiz: @dariakoko That's a clever alternative. But I am concerned that we will end up in a situation where there are so many ways to use refresh control in StickyTabPage. It might be confusing in the long term and harder to refactor.
Something like this might work:
`BigHeaderRefreshControl.tsx`
```
export const BigHeaderRefreshControl = () => (
  <RefreshControl
     progressViewOffset={refreshControlProgressViewOffset ?? 0}
     refreshing={!!isRefreshing}
     onRefresh={refresh}
      style={{ zIndex: 1 }}
   />
)
```
Then pass BigHeaderRefreshControl as the refreshControl prop everywhere the refreshControl is too high up due to Big Headers. This way places we don't need refresh can omit refreshControl and you don't have to refactor every component that already passes the refreshControl prop to StickyTabPage.
What do you think @dariakoko ?

dariakoko: Sounds good, thank you
I've done what you suggested

src/app/Components/StickyTabPage/BigHeaderRefreshControl.tsx
```diff
@@ -0,0 +1,18 @@
+import { RefreshControl } from "react-native"
+import { useStickyTabPageContext } from "./StickyTabPageContext"
+
```
lordkiz, line 3:
```suggestion

/** RefreshControl for StickyTabPage when the static header is large. Allows for the refreshControl to be visible **/
```

src/app/Components/StickyTabPage/BigHeaderRefreshControl.tsx
```diff
@@ -0,0 +1,19 @@
...
+/** RefreshControl for StickyTabPage when the static header is large. Allows for the refreshControl to be visible **/
+export const BigHeaderRefreshControl: React.FC<{
+  refreshing: boolean
+  onRefresh: (() => void) | undefined
+}> = ({ refreshing, onRefresh }) => {
+  const { refreshControlProgressViewOffset } = useStickyTabPageContext()
```
lordkiz, line 9: You don't need this. You already have access to `StaticHeaderHeight`.
```js
const { staticHeaderHeight } = useStickyTabPageContext()
// convert the animatedValue to number
const refreshControlProgressViewOffset = useNativeValue(staticHeaderHeight, 0)
```

src/app/Components/StickyTabPage/BigHeaderRefreshControl.tsx
```diff
@@ -0,0 +1,19 @@
+import { RefreshControl } from "react-native"
+import { useStickyTabPageContext } from "./StickyTabPageContext"
+
+/** RefreshControl for StickyTabPage when the static header is large. Allows for the refreshControl to be visible **/
+export const BigHeaderRefreshControl: React.FC<{
```
lordkiz, line 5: Also maybe we should name this component something that communicates that it should be used in StickTabPage.
Maybe something like `StickTabPageRefreshControl`. And if you already use staticHeaderHeight to determine the offset, it can be safe to use everywhere you use StickyTabPage.

dariakoko: Sounds better
Done

src/app/Components/StickyTabPage/BigHeaderRefreshControl.tsx
lordkiz: @dariakoko What I mean is that the `staticHeaderHeight` you are measuring here: https://github.com/artsy/eigen/pull/6982/files#diff-02e8598434b7dbd07be230aa63e4fb58f93ed3ea72e960dee67d0e224e82020fR69 is already automatically measured in the StickyTabPage. You can get it from the context and convert it to a number with `useNativeValue`

dariakoko: Finally, it should be okay now :)
Thanks for pointing out `useNativeValue`! I didn't realise we can easily change types
I've removed the new prop I've introduced and replaced refresh controls for the tabs on the Settings -> Follow page

src/app/Components/StickyTabPage/StickTabPageRefreshControl.tsx
```diff
@@ -0,0 +1,28 @@
...
+  refreshing: boolean
+  onRefresh: (() => void) | undefined
+}> = ({ refreshing, onRefresh }) => {
+  const { staticHeaderHeight } = useStickyTabPageContext()
+
+  const setprogressViewOffset = () => {
+    if (staticHeaderHeight) {
+      return useNativeValue(staticHeaderHeight, 0)
+    }
+  }
```
lordkiz, line 18: @dariakoko good changes. However, this breaks the rules of hooks: https://reactjs.org/docs/hooks-rules.html#:~:text=Don't%20call%20Hooks%20inside,each%20time%20a%20component%20renders. and also can be buggy.
This is fine:
```ts
  const { staticHeaderHeight } = useStickyTabPageContext()

  const progressViewOffset = staticHeaderHeight ? useNativeValue(staticHeaderHeight, 0) : 0

  return (
    <RefreshControl
      progressViewOffset={progressViewOffset}
      refreshing={refreshing}
      onRefresh={onRefresh}
      style={{ zIndex: 1 }}
    />
  )
```

lordkiz (approved):
Looks great! Thanks for fixing.

dariakoko:
Thanks for all your comments Kizito @lordkiz!

dariakoko:
@brainbicycle hello! 
It turned out to be a pretty easy fix. Do you mind merging it?
Pinging you, cause you've reported this issue a while ago

PR #6983, opened 2022-06-25
refactor(tests): prevent button text rendered twice in testing environment

gkartalis:
This PR resolves []

### Description

In the button component we use `MeasuredView` in order to prevent the button from changing size when the text changes (eg. save -> saved).

This results in the testing environment to have to select the button element with `getAllByText("btn text")[0]` due to button text being rendered twice in the testing tree snapshot even though in the real environment this is not displayed to the user.

This solution will prevent that.

TODO:
- [x] Also replaced all cases where we used `getAllByText("btn text")[0]` -> `getByText("btn text")`

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- prevent button text from rendering twice in testing environment - gkartalis

</details>

### DIFF
--- a/src/palette/elements/Button/Button.tsx
+++ b/src/palette/elements/Button/Button.tsx
@@ -182,11 +182,18 @@ export const Button: React.FC<ButtonProps> = ({
                       <Spacer mr={0.5} />
                     </>
                   ) : null}
-                  <MeasuredView setMeasuredState={setLongestTextMeasurements}>
-                    <Text color="red" style={textStyle}>
-                      {longestText ? longestText : children}
-                    </Text>
-                  </MeasuredView>
+                  {/* This makes sure that in testing environment the button text is
+                      not rendered twice, in normal environment this is not visible.
+                      This will result in us being able to use getByText over
+                      getAllByText()[0] to select the buttons in the test environment.
+                  */}
+                  {!__TEST__ && (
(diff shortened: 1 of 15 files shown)


MounirDhahri (approved):

MounirDhahri:
Niiice

src/palette/elements/Button/Button.tsx
```diff
@@ -182,11 +182,18 @@ export const Button: React.FC<ButtonProps> = ({
...
-                    <Text color="red" style={textStyle}>
-                      {longestText ? longestText : children}
-                    </Text>
-                  </MeasuredView>
+                  {/* This makes sure that in testing environment the button text is
+                      not rendered twice, in normal environment this is not visible.
+                      This will result in us being able to use getByText over
+                      getAllByText()[0] to select the buttons in the test environment.
+                  */}
+                  {!__TEST__ && (
```
pvinis, line 190: 🤔
i am not convinced, but im willing to try it out.
the main reason im not convinced is that we basically will be rendering things differently for tests now. the diff will be the size of the button, so we should be ok, since we dont rely on pixel positioning, so let's try it. but still, one extra way we do things differently in tests.

pvinis (approved):

PR #6986, opened 2022-06-27
RFC: Stop usage of `index.ts` files

olerichter00:
### Description

We should try to avoid the use of `index.ts` files to prevent noise in the file structure and circular dependencies and make it easier to navigate between files.

Some cases where we still use `index.ts` files for shorter and easier imports are (because Palette can be treated as a library):

- `src/palette`

------------- 
Here are a few articles about (not) using `index.ts` files.

- https://medium.com/@alonmiz1234/why-you-should-avoid-index-js-3321a9902120
- https://josselinbuils.me/blog/stop-using-index-files


gkartalis (commented):
I generally like this idea, and this might also resolve some already existing circular dependencies in eigen! 

We have a ticket for resolving some circular deps in [refineathon board for eigen](https://www.notion.so/artsy/Remove-Require-Cycles-a297dd7576a346389dc98e23ec03769c).

Will we start with a plan like first stop adding them and then open some prs to remove the old ones? 

What do you have in mind in order to adopt this?

olerichter00:
Exactly, I plan to remove them within the next weeks and add a section about not using `index.ts` files in the best_practices doc.

PR #6988, opened 2022-06-28
fix: secrets detection

pvinis:
This PR resolves []

### Description

Context: https://artsy.slack.com/archives/C02BAQ5K7/p1656407038938159

In short, we were using our own brew version of detect-secrets because the docker image we use had an older version. We don't need that anymore.

We now use the fresh python version of detect-secrets, and in another PR we will replace the docker image, so we don't have issues in the future.

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/Brewfile
+++ b/Brewfile
@@ -1,2 +1,1 @@
-brew "artsy/formulas/detect_secrets" # using brew version. read more here: https://github.com/artsy/homebrew-formulas/pull/13
 brew "jq"
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,1 @@
+detect-secrets >= 1.2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -220,9 +220,7 @@ jobs:
     working_directory: /usr/src/app
     steps:
       - checkout
-      - run: ./scripts/secrets-check-all
-      - run: git status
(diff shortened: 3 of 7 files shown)


Brewfile
```diff
@@ -1,2 +1 @@
-brew "artsy/formulas/detect_secrets" # using brew version. read more here: https://github.com/artsy/homebrew-formulas/pull/13
```
pvinis, line 1: cya

requirements.txt
```diff
@@ -0,0 +1 @@
+detect-secrets >= 1.2
```
pvinis, line 1: new version

artsyjian (commented):
lgtm

brainbicycle:
this looks good to me, thanks for the fix!
I think long term if we can I would prefer to use the brew version just to avoid having python + pip as another requirement for running Eigen since we already have quite a few requirements to run but curious what others think.

brainbicycle (approved):

brainbicycle:
it is a fair point but when brew installs python I don't have to think about it for the most part is my experience, and some of the extra control might come at a cost, e.g issues with different versions of python. Either way an issue for another day.

PR #7024, opened 2022-07-01
fix: allow building eigen for oss contributors

brainbicycle:
This PR resolves [MOPLAT-6]

### Description

Resolves a few issues preventing building and running for OSS contributors.

- Adds an OSS flag
- Adds instructions for getting a Mapbox token
- Don't initialize firebase app for open source contributors
- Don't initialize unleash for open source contributors
- Don't initialize sentry for open source contributors

Still to do: 
- [x] Add back the google files + add some phony files to allow compilation for OSS
- [x] Add the old open source client id and secret to .env.example
- [x] Clean up some of the handling, don't love the unleash handling and scattered if checks
- [ ] Handle failing relay requests once logged in (friendly error message?)
- [ ] Document known issues with running this way 

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- Allow compiling and running app for open source contributors - Brian 

</details>

[MOPLAT-6]: https://artsyproduct.atlassian.net/browse/MOPLAT-6?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ios/Artsy/ArtsyUIFonts/UIFont+ArtsyFonts.m
```diff
@@ -39,7 +35,10 @@ + (void)loadFontWithFileName:(NSString *)fontFileName extension:(NSString *)exte
         }
 
         CFRelease(font);
+    } else {
+        @throw [NSException exceptionWithName:NSInternalInconsistencyException reason:@"Font failed to load if you are an open source contributor set OSS=true in your ENV file." userInfo:@{}];
```
olerichter00, line 39: Does `env=true` work as well or should we adjust the error message? 🤔 


```suggestion
        @throw [NSException exceptionWithName:NSInternalInconsistencyException reason:@"Font failed to load if you are an open source contributor set OSS=True in your ENV file." userInfo:@{}];
```

src/app/system/errorReporting/sentrySetup.ts
```diff
@@ -13,19 +13,28 @@ export const eigenSentryReleaseName = () => {
...
+  }
+
+  if (!sentryDSN) {
     console.error("Sentry DSN not set!!")
     return
   }
 
   Sentry.init({
-    dsn: Config.SENTRY_DSN,
+    dsn: !ossUser ? sentryDSN : undefined,
```
brainbicycle, line 37: can just pass the dsn here we will return before hitting if they are an oss user

gkartalis:
just adding [this issue](https://github.com/artsy/eigen/issues/8411) that was opened yesterday cause its related

ios/Artsy/ArtsyUIFonts/UIFont+ArtsyFonts.m
```diff
@@ -1,16 +1,12 @@
 // NOTE: This file should be a copy and paste of
 //       https://github.com/artsy/Artsy-OSSUIFonts/blob/master/Pod/Classes/UIFont+ArtsyFonts.m
-//       but with useClosedFonts = YES;
 
 #import "UIFont+ArtsyFonts.h"
 #import <CoreText/CoreText.h>
+#import <react-native-config/ReactNativeConfig.h>
```
brainbicycle, line 6: don't need this
```suggestion
```

scripts/setup-oss-fonts
```diff
@@ -0,0 +1,18 @@
...
+mkdir -p ./fonts
+
+for i in "${fonts[@]}"
+do
+   aws s3 cp "s3://artsy-webfonts-production/$i.otf" "./fonts/$i.otf"
+   cp "./fonts/$i.otf" "$ANDROID_FONT_DIR"
+   cp "./fonts/$i.otf" "$IOS_FONT_DIR"
+done
+
+rm -rf ./fonts
```
damassi, line 18: 👍

src/app/App.tsx
```diff
@@ -80,6 +81,9 @@ const Main = () => {
   useRageShakeDevMenu()
   useDebugging()
   useEffect(() => {
+    if (Config.OSS === "True") {
```
damassi, line 84: maybe just `true`? That's more common from env files

src/app/utils/experiments/unleashClient.ts
```diff
@@ -12,6 +12,10 @@ import { UnleashClient } from "unleash-proxy-client"
  * or create a new one in the requested env.
  */
 export function getUnleashClient(env?: "production" | "staging", userId?: string | null) {
+  if (Config.OSS === "True") {
+    return null
```
damassi, line 16: I wonder how we can avoid all of the extra `?` that are now required in the code. Our unleash usage is going to increase  pretty dramatically through time, especially once we retire echo. 

Non-blocking, but worth thinking about! Maybe we can just turn it off via the api? Like, initialize the client, but then immediately call [stop](https://github.com/Unleash/unleash-proxy-client-js/blob/main/src/index.ts#L317)? 

Or catch errors with logbox? Not too sure.

brainbicycle: the problem is if you try to initialize the client without keys or with bad keys it crashes I believe. I can try again but at least the unleash stuff is abstracted behind the hook so really that is the only place the ? should affect

damassi: Yah its not so bad, just wondering if there might be some easy thing we can do like maybe return a stub api and cast it with an unleast type:

```tsx
if (oss) {
  return {
    foo: x => x,
    bar: x => x, 
  } as unknown as UnleashClient 
}
```

 But that's probably no prettier. In any case, whatever y'all think is best 👍

brainbicycle: oh that is an interesting thought, I can try

src/app/utils/experiments/unleashClient.ts
```diff
@@ -4,14 +4,31 @@ import { nullToUndef } from "app/utils/nullAndUndef"
...
 
+// We want to return a phony unleash client for oss builds
+// Some typescript magic so we only have to specify public fields
+type PublicFields<T> = {
+  [K in keyof T]: T[K] extends { [P in keyof T[K]]: T[K][P] } ? K : never
+}[keyof T]
+
+type PublicType<T> = Pick<T, PublicFields<T>>
+
+export type PublicUnleashClient = PublicType<UnleashClient>
```
brainbicycle, line 15: chatGPT helped me write this one :)

src/app/utils/experiments/unleashClient.ts
brainbicycle: this seems like it will work, take a look and let me know what you think 🙏

MounirDhahri:
👏👏👏

SamRozen:
🔥 Nice one! Thanks @brainbicycle for taking this on.

PR #7062, opened 2022-07-14
fix: Use first artwork image in widget GRO-1135

jonallured:
Turns out that using the first image in an array is insufficient! 😝 In the case of this artwork:

https://www.artsy.net/artwork/lynne-drexler-untitled-46

The first image in the array is the signature by the artist, not the shot of the work! Instead, you're better off using the `position` property on the `images` array. This PR does that. Pretty pictures:

### before

Simulator Screen Shot - iPhone 13 - 2022-07-14 at 14 42 13

### after

Simulator Screen Shot - iPhone 13 - 2022-07-14 at 14 43 02

https://artsyproduct.atlassian.net/browse/GRO-1135

/cc @artsy/grow-devs

### DIFF
--- a/ios/ArtsyWidget/Fixtures.swift
+++ b/ios/ArtsyWidget/Fixtures.swift
@@ -90,7 +90,7 @@ extension Fixtures {
     
     static var primaryArtwork: Artwork {
         let artist = Artist(name: "Alex Katz")
-        let artworkImage = ArtworkImage(geminiToken: "pd7rW3I1mXhW0vbAJDVm3Q")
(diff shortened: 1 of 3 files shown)


pvinis:
is it this position here https://github.com/artsy/gravity/blob/000b3af443c8b770a0066875497efaf19dfdc2d5/app/models/domain/image.rb#L24 that you are getting?

pvinis:
if it is, which i think it is, then i added this here quickly. https://github.com/artsy/metaphysics/pull/4234

pvinis (approved):

brainbicycle (approved):
Nice!

PR #7081, opened 2022-07-22
feat(NX-3183): adjust order and components of Artwork screen

araujobarret:
This PR resolves [NX-3183]

### Description

Changes the new Contact Gallery component in the Artwork, also incorporating the Partner link to it.

Before

https://user-images.githubusercontent.com/15792853/180398604-05d76d8e-a15b-4193-8b05-99d2de93a457.mov

After

https://user-images.githubusercontent.com/15792853/180398658-4248375b-6167-4649-8b91-ab5ae19aaa5d.mov

cc @artsy/negotiate-devs 

### PR Checklist

### To the reviewers 👀

- changes contact gallery component order and design - araujobarret

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[NX-3183]: https://artsyproduct.atlassian.net/browse/NX-3183?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/Artwork/Artwork.tsx
+++ b/src/app/Scenes/Artwork/Artwork.tsx
@@ -7,7 +7,7 @@ import { ArtworkBelowTheFoldQuery } from "__generated__/ArtworkBelowTheFoldQuery
 import { ArtworkMarkAsRecentlyViewedQuery } from "__generated__/ArtworkMarkAsRecentlyViewedQuery.graphql"
 import { AuctionTimerState, currentTimerState } from "app/Components/Bidding/Components/Timer"
 import { RetryErrorBoundaryLegacy } from "app/Components/RetryErrorBoundary"
-import { navigateToPartner, navigationEvents } from "app/navigation/navigate"
+import { navigationEvents } from "app/navigation/navigate"
 import { defaultEnvironment } from "app/relay/createEnvironment"
 import { ArtistSeriesMoreSeriesFragmentContainer as ArtistSeriesMoreSeries } from "app/Scenes/ArtistSeries/ArtistSeriesMoreSeries"
 import { useFeatureFlag } from "app/store/GlobalStore"
@@ -17,7 +17,7 @@ import { QAInfoPanel } from "app/utils/QAInfo"
 import { ProvideScreenTracking, Schema } from "app/utils/track"
 import { AuctionWebsocketContextProvider } from "app/Websockets/auctions/AuctionSocketContext"
 import { isEmpty } from "lodash"
-import { Box, LinkText, Separator, Text } from "palette"
+import { Box, Separator } from "palette"
 import { useEffect, useLayoutEffect, useRef, useState } from "react"
 import { FlatList, RefreshControl } from "react-native"
 import { commitMutation, createRefetchContainer, graphql, RelayRefetchProp } from "react-relay"
@@ -43,6 +43,7 @@ import {
   populatedGrids,
 } from "./Components/OtherWorks/OtherWorks"
 import { PartnerCardFragmentContainer as PartnerCard } from "./Components/PartnerCard"
+import { PartnerLink } from "./Components/PartnerLink"
 import { Questions } from "./Components/Questions"
 
 interface ArtworkProps {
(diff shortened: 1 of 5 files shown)


src/app/Scenes/Artwork/Components/Questions.tsx
```diff
@@ -11,20 +12,25 @@ export const Questions: React.FC<QuestionsProps> = (props) => {
...
-      <Text>Questions about this piece?</Text>
+    <Flex flexDirection="row" justifyContent="space-between" alignItems="center">
+      <Flex>
+        <Text variant="xs">Questions about this piece?</Text>
+        <PartnerLink artwork={artworkData} />
+      </Flex>
+
       <InquiryButtonsFragmentContainer
         artwork={artworkData}
         variant="outlineGray"
```
lean8086, line 23: The border color of this button has to match the one from Palette. The Figma files were outdated and we [changed Force to have the right color](https://github.com/artsy/force/pull/10080) back then.

lean8086: Though, I see this is not part of this PR and probably had to be done back then when the change was made in Force 🙈

araujobarret: Good catch 👍 
changed on https://github.com/artsy/eigen/pull/7081/commits/0050c384bb777d417d17b499cbcf68626f19e413

src/app/Scenes/Artwork/Components/PartnerLink.tsx
```diff
@@ -0,0 +1,43 @@
...
+export const PartnerLink: React.FC<PartnerLinkProps> = ({ artwork }) => {
+  const artworkData = useFragment(artworkFragment, artwork)
+
+  if (!artworkData?.partner) {
+    return null
+  }
+
+  const { name, href, isLinkable } = artworkData.partner
+
+  if (isLinkable && href) {
```
gkartalis, line 19: **suggestion:** What do you think of changing this to 👇 just to be more explicit? `isLinkable` is already `boolean` but seeing the href like that makes me feel 😳! Up to you though
```suggestion
  if (isLinkable && !!href) {
```

src/app/Scenes/Artwork/Artwork.tsx
```diff
@@ -256,27 +257,31 @@ export const Artwork: React.FC<ArtworkProps> = ({
...
         verticalMargin: 2,
       })
     }
 
+    if (
+      enableConversationalBuyNow &&
+      artworkBelowTheFold &&
+      (artworkAboveTheFold?.isAcquireable ||
+        (!artworkAboveTheFold?.isInquireable && artworkAboveTheFold?.isOfferable))
+    ) {
```
gkartalis, line 278: 👏

src/app/Scenes/Artwork/Artwork.tsx
```diff
@@ -256,27 +257,31 @@ export const Artwork: React.FC<ArtworkProps> = ({
...
     }
 
-    if (enableCreateArtworkAlert && !!partnerAbove?.name) {
-      const { isLinkable, name, href } = partnerAbove
+    if (
+      enableCreateArtworkAlert &&
+      !enableConversationalBuyNow &&
+      !!partnerAbove?.name &&
+      artworkAboveTheFold
+    ) {
```
gkartalis, line 265: 👏

src/app/Scenes/Artwork/Components/Questions.tests.tsx
```diff
@@ -28,17 +29,20 @@ describe("Questions", () => {
   beforeEach(() => (mockEnvironment = createMockEnvironment()))
 
   it("renders", async () => {
-    const { queryByText } = renderWithWrappers(
+    const { getByText } = renderWithWrappers(
```
gkartalis, line 32: **question:** Is there any reason why you changed that from `query` to `get`? 🤔 👀

gkartalis (dismissed):

src/app/Scenes/Artwork/Components/Questions.tests.tsx
araujobarret: was giving us a false positive in that case, the elements were not there because we didn't use `flushPromiseQueue`, so when we use it with the assertion `toBeDefined` gives us a false positive because it doesn't break like getBy. More info on https://testing-library.com/docs/queries/about/

src/app/Scenes/Artwork/Components/PartnerLink.tsx
araujobarret: done in https://github.com/artsy/eigen/pull/7081/commits/ee3f3ab9755829b8485e47fe1f6014e18329fbb1

We use this pattern in force at some parts, if the field can be null/string but I'm not sure for strings if it makes safer, I know it's a good approach when we deal with numbers and the value is `0`, so it's not null but will be false inside the if statement 🙃 
So, no strong attachment of not having it, thanks for pointing it out!

PR #7083, opened 2022-07-22
refactor(tests): remove relay unmocking, simplify relay tests ways

pvinis:
This PR resolves []

### Description

HUGE THANKS TO @gkartalis AND @MrSltun FOR PAIRING WITH ME A BUUUUUNCH ON THESE PRS! 💜 

Work done on the base PR:
- remove relay mock
- remove relay unmock calls
- simplify relay env getter (for react comps)
- simplify relay mock env getter (for tests)
- better render and relay resolution helpers
- some tests refactored
- update/write docs

⚠️ WHEN THIS IS GREEN, WE CAN MERGE. IT WILL NOT BE GREEN UNTIL WE CHANGE ALL THE TESTS TO USE THE NEW WAY. WE WILL MAKE A NEW PR(S) ON TOP OF THIS, TO COOPERATE WITH MORE PEOPLE ON FIXING ALL OF THEM. ⚠️

## PR PLAN:

- this PR (base) #7083
- the test refactor PR (where all the tests are fixed), to be merge onto this PR #7084
- the rest of PRs (where few tests are fixed a time), to be merged onto the refactor PR (links here)

So we merge bottom-up. once all are merged, this PR will go green, and all artsy kingdom will rejoice.

### QA Test Case(s)

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>


docs/testing.md
```diff
@@ -8,44 +8,4 @@ We have 3 ways of testing relay components:
 
 1. Components using both HoC & relay hooks !
 
-Further documentation on testing
```
pvinis, line 11: removing deprecated docs

__mocks__/react-relay.js
```diff
@@ -1,14 +0,0 @@
-// Comes from https://github.com/facebook/relay/issues/161
-// Should be used in tests as a replacement for `react-storybooks-relay-container`
-//
-var Relay = jest.requireActual("react-relay")
-
-module.exports = {
```
pvinis, line 6: OMG CYA NEVER AGAIN! 💜

__mocks__/svgMock.js
```diff
@@ -1,2 +0,0 @@
-module.exports = "SvgMock";
```
pvinis, line 1: this also was not used, so 🪦

patches/@types+relay-test-utils+6.0.5.patch
```diff
@@ -5,9 +5,25 @@ index 3dfa8f1..3295579 100755
...
      alias?: string | undefined;
+diff --git a/node_modules/@types/relay-test-utils/lib/RelayModernMockEnvironment.d.ts b/node_modules/@types/relay-test-utils/lib/RelayModernMockEnvironment.d.ts
+index c14ad95..906b2a0 100755
+--- a/node_modules/@types/relay-test-utils/lib/RelayModernMockEnvironment.d.ts
++++ b/node_modules/@types/relay-test-utils/lib/RelayModernMockEnvironment.d.ts
+@@ -37,9 +37,11 @@ export interface MockFunctions {
+     findOperation: (findFn: (operation: OperationDescriptor) => boolean) => OperationDescriptor;
+     queuePendingOperation: (query: GraphQLTaggedNode, variables: Variables) => void;
+     getMostRecentOperation: () => OperationDescriptor;
++    /** @deprecated Use our `resolveMostRecentRelayOperation` or `resolveMostRecentRelayOperationPayload` helpers instead. */
```
pvinis, line 22: adding some deprecation, only to suggest our own helpers.

src/app/AppProviders.tsx
```diff
@@ -42,7 +42,7 @@ export const AppProviders = ({ children }: { children?: React.ReactNode }) =>
 
 // relay needs the default environment
 const RelayDefaultEnvProvider = (props: { children?: React.ReactNode }) => (
-  <RelayEnvironmentProvider environment={defaultEnvironment} {...props} />
+  <RelayEnvironmentProvider environment={getRelayEnvironment()} {...props} />
```
pvinis, line 45: welcome to the first main change. no more `defaultEnvironment`, we now have a getter function.

The reason is that we can mock this way easier, and we don't end up with duplicate environments.

src/app/Scenes/Artist/ArtistSavedSearch.tests.tsx
```diff
@@ -1,91 +1,52 @@
...
-    console.warn = jest.fn()
-  })
-
-  afterEach(() => {
-    environment = createMockEnvironment()
-    console.error = originalError
-    console.warn = originalWarn
-  })
-
-  function mockMostRecentOperation(name: ArtistQueries, mockResolvers: MockResolvers = {}) {
```
pvinis, line 33: 👋

src/app/Scenes/Artist/ArtistSavedSearch.tests.tsx
```diff
@@ -1,91 +1,52 @@
...
+    renderWithWrappers(<TestRenderer searchCriteriaID="something" />)
 
-    rejectMostRecentRelayOperation(environment, new Error())
-    mockMostRecentOperation("ArtistAboveTheFoldQuery", MockArtistAboveTheFoldQuery)
+    rejectMostRecentRelayOperation(new Error())
+    resolveMostRecentRelayOperation(MockArtistAboveTheFoldQuery)
 
-    expect(getByText("Sorry, an error occured")).toBeTruthy()
-    expect(getByText("Failed to get saved search criteria")).toBeTruthy()
+    expect(screen.getByText("Sorry, an error occured")).toBeTruthy()
```
pvinis, line 39: since we touched that file, we used the new way of testing-lib too.

src/app/Scenes/Home/Home.tests.tsx
```diff
@@ -38,39 +47,17 @@ describe(HomeQueryRenderer, () => {
...
         },
       },
-    })
-
-    await flushPromiseQueue()
-
-    return tree
-  }
-
-  it("renders home screen module flat list", async () => {
```
pvinis, line 57: i dont know why the diff is so weird here, this test still exists in that file, a bit further up.

src/app/Scenes/Search/Search.tests.tsx
```diff
@@ -57,27 +54,10 @@ jest.mock("app/utils/experiments/hooks", () => ({
 }))
 
 describe("Search Screen", () => {
-  let mockEnvironment: ReturnType<typeof createMockEnvironment>
```
pvinis, line 60: this test was a big one. it involved a bunch of relay things, now all green. 😌

src/app/Scenes/VanityURL/VanityURLEntity.tests.tsx
```diff
@@ -25,31 +28,29 @@ const TestRenderer: React.FC<{
 )
 
 describe("VanityURLEntity", () => {
-  let env: ReturnType<typeof createMockEnvironment>
-
-  beforeEach(() => {
```
pvinis, line 30: omg goodbye. never again.

src/app/relay/createEnvironment.ts
```diff
@@ -1,56 +0,0 @@
-import {
```
pvinis, line 1: for this file, check the `defaultEnvironement` file below.

EXAMPLES.md
```diff
@@ -14,24 +14,60 @@ Links:
...
 - [SizesOptionsScreen.tests.tsx](src/app/Components/ArtworkFilter/Filters/SizesOptionsScreen.tests.tsx)
+- [Search.tests.tsx](src/app/Scenes/Search/Search.tests.tsx)
 
 ## I want to write a test with relay.
 
-Similar to above. For `relay` testing we use `resolveMostRecentRelayOperation` to mock the most recent operation.
-Look at https://github.com/artsy/relay-workshop for a great tutorial of how we use relay and test with it.
+This is an example of a good relay test with hooks.
+- Notice we use `async`, in preparation to the await calls later.
+- First thing, we use `renderWithRelayWrappers` to render the component.
```
araujobarret, line 30: I'm still questioning why we moved away from `setupTestWrapperTL` that already has built-in `resolveMostRecentRelayOperation`, I've seen a lot of developers using it for components we only care to mock the GraphQL resolvers once. 

If the reason is to control the mocked environment created inside `resolveMostRecentRelayOperation` we could enhance that function by allowing passing a mocked relay environment and then switching to manual control over it instead of doing the first `resolveMostRecentRelayOperation`.

src/app/relay/defaultEnvironment.ts
```diff
@@ -0,0 +1,62 @@
...
+    checkAuthenticationMiddleware(), // KEEP AS CLOSE TO THE BOTTOM OF THIS ARRAY AS POSSIBLE. It needs to run as early as possible in the middlewares.
+  ],
+  {
+    // `noThrow` is currently marked as "experimental" and may be deprecated in the future.
+    // See: https://github.com/relay-tools/react-relay-network-modern#advanced-options-2nd-argument-after-middlewares
+    noThrow: true,
+  }
+)
+const store = new Store(new RecordSource())
+const defaultEnvironment = new Environment({ network, store })
```
pvinis, line 45: basically, same as before, but without exporting.

We export the getter just below. (also a mock getter for types' sake.) also also, another environment just for the bottom tabs, its a bit of a special case.

The getter makes things easier, as you will see on the setupjest file too.

src/app/tests/createMockNetworkLayer/createMockNetworkLayer.tests.tsx
```diff
@@ -2,8 +2,6 @@ import { graphql } from "react-relay"
 import { Environment, fetchQuery, GraphQLTaggedNode, RecordSource, Store } from "relay-runtime"
 import { createMockNetworkLayer2 } from "./index"
 
-jest.unmock("react-relay")
```
pvinis, line 5: 🔫🔫🔫🔫🔫

src/app/tests/flushPromiseQueue.ts
```diff
@@ -1,5 +1,7 @@
 /**
- * test utility to wait for all promises to be resolved
+ * test utility to wait for all promises to be resolved.
+ *
+ * @deprecated use `waitForSuspenseToBeRemoved` or `waitForElementToBeRemoved` instead.
  */
 export const flushPromiseQueue = () =>
```
pvinis, line 6: imagine a world, where this function does not exist. isn't that world, greener? brighter? more beautiful? 🌞💐🟢

we want to deprecate this at some point, so until then, we add a note to start using the `waitFor*` helpers instead.

src/app/tests/rejectMostRecentRelayOperation.ts
```diff
@@ -1,12 +0,0 @@
-import { act } from "@testing-library/react-native"
-import { OperationDescriptor } from "relay-runtime"
-import { createMockEnvironment } from "relay-test-utils"
-
-export function rejectMostRecentRelayOperation(
```
pvinis, line 5: moved to another file

src/app/tests/renderWithWrappers.tsx
```diff
@@ -102,11 +103,9 @@ export const renderWithWrappers = (component: ReactElement) => {
   }
 }
 
-export const renderWithHookWrappersTL = (component: ReactElement, environment: Environment) => {
-  const jsx = (
-    <RelayEnvironmentProvider environment={environment}>
-      <Suspense fallback="Loading...">{component}</Suspense>
+export const renderWithRelayWrappers = (component: ReactElement) =>
```
pvinis, line 106: quick rename, to match the other funcs.

src/app/tests/renderWithWrappers.tsx
```diff
@@ -102,11 +103,9 @@ export const renderWithWrappers = (component: ReactElement) => {
...
+  renderWithWrappers(
+    <RelayEnvironmentProvider environment={getRelayEnvironment()}>
+      <Suspense fallback={<Text>TEST-SUSPENSE-LOADING</Text>}>{component}</Suspense>
```
pvinis, line 109: adding a text here, for waiting on in tests.

src/app/utils/Context.tsx
```diff
@@ -1,33 +0,0 @@
...
-import { defaultEnvironment } from "../relay/createEnvironment"
-
-interface BaseContextProps {
-  relayEnvironment?: Environment
-}
-
-/**
- * Catch-all for additional context values passed in during initialization.
- */
-export interface ContextProps extends BaseContextProps {
```
pvinis, line 12: what even is that file? goodbye.

src/setupJest.ts
```diff
@@ -519,13 +520,22 @@ jest.mock("app/utils/track/providers", () => ({
   postEventToProviders: jest.fn(),
 }))
 
-jest.mock("app/relay/createEnvironment", () => ({
-  defaultEnvironment: require("relay-test-utils").createMockEnvironment(),
-  createEnvironment: require("relay-test-utils").createMockEnvironment,
-  reset(this: { defaultEnvironment: any }) {
-    this.defaultEnvironment = require("relay-test-utils").createMockEnvironment()
-  },
-}))
```
pvinis, line 528: look at this messy code, hard to read, hard to make it work.

src/setupJest.ts
```diff
@@ -519,13 +520,22 @@ jest.mock("app/utils/track/providers", () => ({
...
+    getMockRelayEnvironment: mockedFunction,
+  }
+})
+const resetMockEnvironment = () => {
+  mockEnvironment = createMockEnvironment()
+}
+// resetting relay mock env before each test is needed, for a clean env to resolve on.
+beforeEach(() => {
+  resetMockEnvironment()
+})
```
pvinis, line 538: look at this code, still a bit messy, way easier to read and make it work.


basically, we create a mock env, then we use that mock env to return it in the `defaultEnv` file, (where we import the getter from in regular code), then we add a reset func (outside the mock, cleaner), and then we call this on `beforeEach`, instead of doing this on every single file.

:feelsrelieved:

src/app/relay/defaultEnvironment.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export const getRelayEnvironment = () => defaultEnvironment
+
+// We could get rid of this, if we could type `getRelayEnvironment`
+// to be a func that returns `Environment` for regular code and
+// to be a func that returns `RelayMockEnvironment` for test code.
+export const getMockRelayEnvironment = getRelayEnvironment as () => RelayMockEnvironment
```
olerichter00, line 52: I wonder if we should move this code to a file with test helpers in order to separate functions that are only relevant for testing 🤔

pvinis: yea.. we had the same thought. ideally we could just change the return type of the regular getter only for tests. but we didn't find a way. so we moved that in a new file. but then we didn't like it was in a far away place, separated, so we moved it here. 🤔

olerichter00: Let's leave it here then, it's only one line of code anyway.

docs/testing.md
```diff
@@ -8,44 +8,4 @@ We have 3 ways of testing relay components:
...
-We have initialisations of mocks such as
-
-```
- jest.mock("app/relay/createEnvironment", () => {
-   return {
-     defaultEnvironment: require("relay-test-utils").createMockEnvironment(),
-   }
- })
-```
+For examples on each of these, look at [EXAMPLES.md](/EXAMPLES.md)
```
patrinoua, line 11: Testing is a topic that is big and complicated enough, so it is nice to have all the information about what we do, how we do it, best practices and examples maintained into one place. 

My suggestion would be keeping the examples on this dedicated file and pointing to the examples section on testing.md from Examples.md.

EXAMPLES.md
damassi: Haven't read through the full PR yet but @araujobarret's comment resonates!

EXAMPLES.md
```diff
@@ -14,24 +14,60 @@ Links:
...
+      system: {
+        algolia: {
+          appID: "",
+          apiKey: "",
+          indices: [{ name: "Artist_staging", displayName: "Artists", key: "artist" }],
+        },
+      },
+    }),
+  })
+  await waitForSuspenseToBeRemoved("search-placeholder") // And now we wait..
```
damassi, line 57: It seems like all three of these actions can be combined into one function call -- a la `setupTestWrapper`. The waiting aspect seems like it could be automated away as well, or at least hidden behind a `waitForID` prop that is passed into the `setupTestWrapper` function. Having to write each of these setup actions for all of our tests is going to get tiresome over time and some clever dev will come through and wrap them behind a function eventually anyway.

EXAMPLES.md
pvinis: i did have this as a first trial, and it ended up being "modified" in more than a few places. and by "modified" i mean need for either an assertion before the resolution, or multiple resolutions, or other things, so i decided to keep it in the smallest parts 🤔

EXAMPLES.md
```diff
@@ -14,24 +14,60 @@ Links:
...
+
+  expect(screen.queryByText("City Guide")).toBeFalsy() // Verifing things.
+})
+```
+
+Look at https://github.com/artsy/relay-workshop for a more tutorials of how we use relay and test with it.
+
+> :warning: Note: Code with relay hooks, needs `renderWithRelayWrappers`, code with old relay components (`QueryRenderer` etc) needs `renderWithWrappers`. All new code should be using relay hooks. When converting old relay code to relay hooks, make sure to swap `renderWithWrappers` for `renderWithRelayWrappers`.
```
damassi, line 65:
```suggestion
> :warning: Note: Code with relay hooks needs `renderWithRelayWrappers`; code with old relay components (`QueryRenderer` etc) needs `renderWithWrappers`. All new code should be using relay hooks. When converting old relay code to relay hooks, make sure to swap `renderWithWrappers` for `renderWithRelayWrappers`.
```

docs/testing.md
pvinis: thats what we did 🤔. at some point we should fill this file here with more info, but until then, we thought pointing to EXAMPLES.md is the best thing we can do.

patches/@types+relay-test-utils+6.0.5.patch
damassi: Rather than using patch package, would it be possible to extend the relay type in a relay.d.ts file, which would automatically invoke the deprecation when used? (I'm fishing here for a way to get around this, but in any case this patch-package change is pretty harmless)

EXAMPLES.md
damassi: Maybe we can optimize for the most common use-case (most devs will just need to use the `setupTestWrapper` function)? For the few places where this needs to be modified we can use the extended version.

src/app/tests/createMockNetworkLayer/createMockNetworkLayer.tests.tsx
damassi: I'm still not following how this can be removed - mind explaining all the parts around how this works so that I can adapt to force?

patches/@types+relay-test-utils+6.0.5.patch
pvinis: hm, i could try. im not sure how well relay.d.ts plays with the original types coming from the package. to be honest i would prefer a homebrew d.ts to go away and have a patch, rather than keeping both around if they don't play well together.

ill look into it.

EXAMPLES.md
gkartalis: I don't have anything against setupTestWrapper, we can try to refactor it (after these first setup prs) I guess.  

Some "issues" with the `setupTestWrapper` that I've encountered in the past(noting them down here to remember them if we decide to optimize it):

- uses `QueryRenderer` and the newer components that we introduce in `eigen` use relay hooks
- In some cases (Home, Search) we are preloading the queries and although it might work with resolveMostRecentRelayOperation, relay docs have other suggestions for testing [preloaded queries](https://relay.dev/docs/guides/testing-relay-with-preloaded-queries).
 
Also for @araujobarret's point 
>If the reason is to control the mocked environment created inside the setupTestWrapperTL we could enhance that function by allowing to pass a mocked relay environment and then switching to manual control over it instead of doing the first resolveMostRecentRelayOperation.

The good thing is that now with `getRelayEnvironment()` we can get direct access to the environment to do whatever we want with it (use it to resolve subsequent mutations or queries in the test file)

pvinis: i guess that makes sense. my fear is that it might lead to cases where a test doesn't pass, and there is no real place to put a `screen.debug()`, since this will be one line, but ill try it again. @gkartalis lets try this next week 👍

src/app/tests/createMockNetworkLayer/createMockNetworkLayer.tests.tsx
pvinis: of course.

so basically, here is what `__mocks__/react-relay.js` had inside:
```
// Comes from https://github.com/facebook/relay/issues/161
// Should be used in tests as a replacement for `react-storybooks-relay-container`
//
var Relay = jest.requireActual("react-relay")

module.exports = {
  createFragmentContainer: component => component,
  createPaginationContainer: component => component,
  createRefetchContainer: component => component,
  graphql: Relay.graphql,
  commitMutation: (environment, { onCompleted, onError }) => {
    onCompleted && onCompleted()
  },
}
```

so it was basically making these functions an identity.. createFragmentContainer was just rendering the component and that's it. I never understood why we need this. If we don't use relay in a test file, we don't need this mock, and if we use relay then we were unmocking it. so I'm pretty sure it is just an ancient thing from this really old hack for this really old https://github.com/facebook/relay/issues/161 issue. We didn't have a HACKS.md file back then to and this and check later, so it stayed. It was just one of those hacks that ended up messing with basically all our test files that included relay.

If while we fix the tests we run into one that this is actually needed, we will def bring it up, but I'm pretty certain if that happens, it will be super easy to rewrite that test without using that weird mock.

So yea, I get why we needed the hack, even thought I don't know 100% what internal relay code it was fixing. I get how we might needed to unmock since this was breaking regular relay tests. My guess is that we should have mocked this only where we needed, rather than mock globally and unmock everywhere!!! And now we are removing this and with it we are making the tests run well with resolve and async etc. Not sure how that unmocking of the unmock was messing with relay timings and we needed to flush promises etc, but everything seems to work better without that extra complication.

For force, I would be curious what would happen if you removed the mock, removed all the unmock lines, and ran the tests. 🤔

damassi: ty @pvinis! I shall give it a try today and let y'all know 👍

EXAMPLES.md
araujobarret: Makes total sense to me, thanks for the explanation, I would be happy to jump in too 👍 
If we can reduce the relay test boilerplate without creating a strict structure, specially in sync what we do in Force/Eigen/AnotherClient would be Toll!

patches/@types+relay-test-utils+6.0.5.patch
pvinis: from @mdole, make it clear that this is our own "deprecation", not really an actual deprecation.

pvinis:
closing this in favor of https://github.com/artsy/eigen/pull/8033.

PR #7100, opened 2022-07-25
chore: CI uses detect secrets v1.3.0

ovasdi:
### Change

* This updates CI to use `artsy/detect-secrets:1.3.0`
* Regenerate the `.secrets.baseline` under the new version

### Motivation

Last Friday a new version of detect-secrets was [released](https://github.com/Yelp/detect-secrets/releases/tag/v1.3.0) and it trickled down into various projects that use 3rd party image, not Eigen 👏 (Thank you @pvinis). This caused issues in some projects when running in CI as the new version seems to be detecting additional secrets, not detected by v1.2.0.

After rebuilding the `baseline` [in Force](https://github.com/artsy/force/pull/10597) under the new version, I was going to announce in #dev asking developers to upgrade their local version of the tool however, I then realized that Eigen is pinned to v1.2.0. This updates Eigen CI to use the latest version of the tool so that we can ask developers to upgrade and not run into version conflicts when working in Eigen and other projects.

The new image has been built and pushed to [docker-hub](https://hub.docker.com/layers/detect-secrets/artsy/detect-secrets/1.3.0/images/sha256-60eccdb9dd700881b40998015044ba583375881f3701db40609b3239f99cdbce?context=explore).

```
➜  docker run artsy/detect-secrets:1.3.0 detect-secrets-hook --version
1.3.0
```

### DIFF
--- a/scripts/secrets-generate-baseline
+++ b/scripts/secrets-generate-baseline
@@ -11,4 +11,5 @@ detect-secrets scan \
   --exclude-files "\.jar$"    `# ignore jar files`   \
   --exclude-files "__generated__"      `# ignore relay generated files` \
   --exclude-files "ios\/.*\.xcscheme$" `# ignore Xcode scheme files`    \
+  --exclude-secrets '[a-fA-F0-9]{24}' `# ignore mongo ID like UUIDs` \
   > .secrets.baseline
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -216,7 +216,7 @@ commands:
 jobs:
   detect-secrets:
     docker:
-      - image: artsy/detect-secrets:1.2.0
+      - image: artsy/detect-secrets:1.3.0
     working_directory: /usr/src/app
     steps:
       - checkout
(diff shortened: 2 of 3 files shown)


scripts/secrets-generate-baseline
```diff
@@ -11,4 +11,5 @@ detect-secrets scan \
   --exclude-files "\.jar$"    `# ignore jar files`   \
   --exclude-files "__generated__"      `# ignore relay generated files` \
   --exclude-files "ios\/.*\.xcscheme$" `# ignore Xcode scheme files`    \
+  --exclude-secrets '[a-fA-F0-9]{24}' `# ignore mongo ID like UUIDs` \
```
pvinis, line 14: why is this needed?

ovasdi: I noticed that a number of secrets recorder was mongoID like strings. Adding that regex cuts down the `baseline` by ~600 lines. We do this in other projects as well, for example [pulse](https://github.com/artsy/pulse/blob/main/.secrets.baseline#L113).

mc-jones (approved):

pvinis (approved):

PR #7109, opened 2022-07-26
docs: Notes on images

patrinoua:
This PR adds more context on how we use images on the documentation.

### QA Test Case(s)

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/docs/images.md
+++ b/docs/images.md
@@ -1,0 +1,30 @@
+# Images
+
+Do you need to use an image?
+
+At artsy we have several ways of retrieving an image.
+
+Images are stored in metaphysics.
+
+In general, our current preferred practice is to use the normalized image and resize it as necessary using on-the-fly resizing.
+
+Here is an example query:
+
+```
+{
+  show(id: "62dbddc6b70486000d2332bb") {
+    images(default: true) {
+      url(version: "normalized")
+      cropped(height: 500, width: 500, version: "normalized") {
+        url
+      }
+
+      resized(height: 500, version: "normalized") {
+        url
+      }
+    }
+  }
+}
+```
+
+This doc [here](https://www.notion.so/artsy/Handling-Images-On-The-Web-In-Force-5c6832bf4b3e431cb4830061ce057399) is web specific but provides some more context on how we handle images.
--- a/docs/README.md
+++ b/docs/README.md
@@ -21,6 +21,7 @@
 - [State Migrations](adding_state_migrations.md)
 - [Analytics and Tracking](analytics_and_tracking.md)
 
+- [Images handling](images.md)
 - [Storybook](storybook.md)
 - [Converting an image to webp](convert_image.md)
 


docs/images.md
```diff
@@ -0,0 +1,30 @@
+# Images
+
+Do you need to use an image?
+
+At artsy we have several ways of retrieving an image.
+
+Images are stored in metaphysics.
```
pvinis, line 7:
```suggestion
Image urls are retrieved from metaphysics.
```

docs/images.md
```diff
@@ -0,0 +1,30 @@
...
+      }
+
+      resized(height: 500, version: "normalized") {
+        url
+      }
+    }
+  }
+}
+```
+
```
pvinis, line 29: maybe add an example for usage like `<Image source={{uri:show.image.url}} />`?

docs/images.md
```diff
@@ -0,0 +1,30 @@
...
+Here is an example query:
+
+```
+{
+  show(id: "62dbddc6b70486000d2332bb") {
+    images(default: true) {
+      url(version: "normalized")
+      cropped(height: 500, width: 500, version: "normalized") {
+        url
```
dblandin, line 19: **suggestion**: We should consider changing this to `srcSet` since that accounts for both 1x and 2x sizes. If we need to consider other scales, like 3x, we can propose an update in Metaphysics that support that ([GitHub](https://github.com/artsy/metaphysics/blob/7ded62cbafae927ea646e81e492fddaf3891ca72/src/schema/v2/image/resized.ts#L106)).

Are there some docs we can add about how to use `srcSet` within an Image component available in Eigen that automatically chooses the right src based on the component size and the device screen resolution

patrinoua: Since we want to mainly use 2x maybe this could be returned by default?

dblandin: `srcSet` returns both 1x and 2x as a set and lets the client figure out which is the right to use so it already includes 2x by default.

patrinoua: After some investigation yesterday, there seems to be no easy way to pass the srcSet to the <Image /> component, this is why we have currently opted for this implementation. However I'm happy to hear if you or someone else is aware of a way to pass this so we could use it!

patrinoua: This is the related PR https://github.com/artsy/energy/pull/117

docs/images.md
```diff
@@ -0,0 +1,89 @@
...
+      resized(height: 500, version: "normalized") {
+        url
+      }
+    }
+  }
+}
+
+```
+
+This doc [here](https://www.notion.so/artsy/Handling-Images-On-The-Web-In-Force-5c6832bf4b3e431cb4830061ce057399) is web specific but provides some more context on how we handle images.
```
mdole, line 82: Suggest moving this to the top; it's a bit buried down here and not super related to the things around it

docs/images.md
```diff
@@ -0,0 +1,89 @@
+# Images
+
+Do you need to use an image?
+
+At artsy we have several ways of retrieving an image.
+
+Image urls are retrieved from metaphysics.
+
+In general, our current preferred practice is to use the normalized image and resize it as necessary using on-the-fly resizing.
```
mdole, line 9: Suggestion: maybe worth mentioning "images should always be at least 2x the size you anticipate displaying them", perhaps linking out to an article on what that means and why it's important?

mdole (commented):
Woohoo! Thanks for improving the docs @patrinoua! Our future selves thank you 😁 

Two quick suggestions, let me know what you think!

docs/images.md
```diff
@@ -6,7 +6,11 @@ At artsy we have several ways of retrieving an image.
 
 Image urls are retrieved from metaphysics.
 
-In general, our current preferred practice is to use the normalized image and resize it as necessary using on-the-fly resizing.
+[This doc](https://www.notion.so/artsy/Handling-Images-On-The-Web-In-Force-5c6832bf4b3e431cb4830061ce057399), although web specific, provides some more context on how we handle images.
+
+In general, our current **preferred practice** is to use the normalized image and resize it as necessary using on-the-fly resizing.
+
+Images should always be at least 2x of the size we would anticipate to display them, since because of higher pixel densities as for example the latest iPhones have, high-resolution displays demand images with more pixels. You can read more about it [here](https://developer.apple.com/design/human-interface-guidelines/foundations/images/).
```
mdole, line 13: 👏 thank you!

mdole (approved):

docs/images.md
patrinoua: Thank's a lot for the suggestions! 😊

PR #7126, opened 2022-07-29
Docs: update pr template

patrinoua:
We updated the PR template to include test descriptions as well as a section that we can directly provide to Applause for testing.

Coauthored by @brainbicycle 

@artsy/engineering for your reference

### QA Test Case(s)

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/docs/pull_request_template.md
+++ b/docs/pull_request_template.md
@@ -13,20 +13,12 @@ This PR resolves []
 
 ### QA Testing
 
-Please provide information on how to test this PR.
-These tests will be run in recent changes QA, they do not have to be extensive, just a high level feature sanity check, you should do your own extensive QA with your team.
+<!-- Please provide information on how to test this PR.
+These tests will be run in recent changes QA, they do not have to be extensive, just a high level feature sanity check, you should do your own extensive QA with your team. -->
 
 - Acceptance Criteria:
+
 - Setup Instructions/Link:
-
-<!-- Example:
-
-- Feature: Search
-- Test Case: Save a search
-- Environment: Staging iOS
-- Acceptance Criteria: The user should be able to ..
-- Setup Instructions/Link: Start from opening the app and navigating to an artist...
--->
 
 ### PR Checklist
 
--- a/docs/pull_request_template.md
+++ b/docs/pull_request_template.md
@@ -20,7 +20,7 @@ These tests will be run in recent changes QA, they do not have to be extensive, 
 
 -
 
-#### Setup Instructions/Link
+#### Setup Instructions
 
 -
 
--- a/docs/pull_request_template.md
+++ b/docs/pull_request_template.md
@@ -27,7 +27,7 @@ These tests will be run in recent changes QA, they do not have to be extensive, 
 - [ ] I added Tests and Stories for my changes.
 - [ ] I added an [app state migration].
 - [ ] I hid my changes behind a [feature flag].
-- [ ] I have prefixed things that need to be tested during a release QA with **[NEEDS EXTERNAL QA]**
+- [ ] I have prefixed things that need to be tested during a release QA with **[NEEDS EXTERNAL QA]** on the changelog.
 
 ### To the reviewers 👀
 
--- a/docs/pull_request_template.md
+++ b/docs/pull_request_template.md
@@ -16,9 +16,13 @@ This PR resolves []
 <!-- Please provide information on how to test this PR.
 These tests will be run in recent changes QA, they do not have to be extensive, just a high level feature sanity check, you should do your own extensive QA with your team. -->
 
-- Acceptance Criteria:
+#### Acceptance Criteria
 
-- Setup Instructions/Link:
+-
+
+#### Setup Instructions/Link
+
+-
 
 ### PR Checklist
 
@@ -27,7 +31,7 @@ These tests will be run in recent changes QA, they do not have to be extensive, 
 - [ ] I added Tests and Stories for my changes.
 - [ ] I added an [app state migration].
 - [ ] I hid my changes behind a [feature flag].
-- [ ] I have prefixed things that need to be tested during a release QA with **[NEEDS EXTERNAL QA]** on the changelog.
+- [ ] I have prefixed changes that need to be tested during a release QA with **[NEEDS EXTERNAL QA]** on the changelog.
 
 ### To the reviewers 👀
 
--- a/docs/pull_request_template.md
+++ b/docs/pull_request_template.md
@@ -27,7 +27,7 @@ These tests will be run in recent changes QA, they do not have to be extensive, 
 - [ ] I added Tests and Stories for my changes.
 - [ ] I added an [app state migration].
 - [ ] I hid my changes behind a [feature flag].
-- [ ] We want to have this tested by Applause.
+- [ ] I have prefixed things that need to be tested during a release QA with **[NEEDS EXTERNAL QA]**
 
 ### To the reviewers 👀
 
@@ -59,10 +59,6 @@ These tests will be run in recent changes QA, they do not have to be extensive, 
 #### Dev changes
 
 -
-
-#### To be tested by Applause
-
-- If your change needs to be tested by Applause please paste here the name of the change as well as how to test it that you already filled up above.
 
 <!-- end_changelog_updates -->
 
--- a/docs/pull_request_template.md
+++ b/docs/pull_request_template.md
@@ -6,6 +6,7 @@
 -->
 
 ### Description
+
 <!-- Info, implementation, how to get there, before & after screenshots & videos, follow-up work, etc -->
 
 This PR resolves []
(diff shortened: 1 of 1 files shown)


docs/pull_request_template.md
```diff
@@ -1,28 +1,33 @@
...
-This PR resolves []
-
 ### Description
-
 <!-- Info, implementation, how to get there, before & after screenshots & videos, follow-up work, etc -->
 
-### QA Test Case(s)
+This PR resolves []
+
+### QA Testing
```
MounirDhahri, line 13: Such a section could make the job of whoever is the release captain easier but I am afraid it will cause everyone else to spend lots of extra time adding details here. I was under the impression such details should be included already in the Jira ticket. Maybe we can update the recent changes notion doc to include the original Jira ticket

docs/pull_request_template.md
```diff
@@ -1,28 +1,33 @@
...
 
-<!-- Does this PR need QA testing? (hint: it probably does). If so add details here. See example below. These tests will be run in recent changes QA, they do not have to be extensive, just a high level feature sanity check, you should do your own extensive QA with your team. -->
+Please provide information on how to test this PR.
+These tests will be run in recent changes QA, they do not have to be extensive, just a high level feature sanity check, you should do your own extensive QA with your team.
 
-| Test Case | Feature | Environment | Acceptance Criteria | Setup Instructions/Link |
-| --------- | :-----: | ----------: | ------------------: | ----------------------: |
-|           |         |             |                     |                         |
+- Feature:
```
MounirDhahri, line 18:
```suggestion
- Context:
```

docs/pull_request_template.md
```diff
@@ -1,28 +1,33 @@
...
+- Test Case:
+- Environment:
```
MounirDhahri, line 20: I feel this is not necessary since the environment will be almost always staging

docs/pull_request_template.md
```diff
@@ -1,28 +1,33 @@
...
 <!-- Info, implementation, how to get there, before & after screenshots & videos, follow-up work, etc -->
 
-### QA Test Case(s)
+This PR resolves []
+
+### QA Testing
 
-<!-- Does this PR need QA testing? (hint: it probably does). If so add details here. See example below. These tests will be run in recent changes QA, they do not have to be extensive, just a high level feature sanity check, you should do your own extensive QA with your team. -->
+Please provide information on how to test this PR.
+These tests will be run in recent changes QA, they do not have to be extensive, just a high level feature sanity check, you should do your own extensive QA with your team.
```
MounirDhahri, line 16: **thought:**
Was this uncommented intentionally?

docs/pull_request_template.md
```diff
@@ -32,6 +37,7 @@ This PR resolves []
 - [ ] I added Tests and Stories for my changes.
 - [ ] I added an [app state migration].
 - [ ] I hid my changes behind a [feature flag].
+- [ ] We want to have this tested by Applause.
```
MounirDhahri, line 40:
```suggestion
- [ ] We want to have this tested by our external QA vendor. 
```
Should we maybe not mention Applause name here? maybe it's fine

docs/pull_request_template.md
```diff
@@ -49,21 +55,22 @@ This PR resolves []
...
-
 -
 
 #### Dev changes
-
 -
 
+#### To be tested by Applause
+
+- If your change needs to be tested by Applause please paste here the name of the change as well as how to test it that you already filled up above.
```
MounirDhahri, line 71: I would suggest instead simply prefixing the previous items with something like `[NEEDS EXTERNAL QA]` or so to avoid adjusting our parser. I mean instead of having something like
```
### Cross-platform user-facing changes
- Updated login - mounir
```
we can do
```
### Cross-platform user-facing changes
- [NEEDS EXTERNAL QA] Updated login - mounir
```

docs/pull_request_template.md
olerichter00: Good point! I'd like to keep it as a comment as well to reduce the manual work of removing/commenting these lines.

docs/pull_request_template.md
olerichter00: Will this section be parsed automatically like the changelog?

patrinoua: > Will this section be parsed automatically like the changelog?

That would be great!

docs/pull_request_template.md
patrinoua: I have now removed this completely since the description of the pr provides that context. It would be great if we can parse this automatically in the changelog, together with the instructions / acceptance criteria.

docs/pull_request_template.md
patrinoua: Agreed, I removed this

docs/pull_request_template.md
patrinoua: I like that idea!

patrinoua:
@MounirDhahri and @olerichter00  thanks for your comments. I’ve updated the format here, it would be great if we can set up automatic parsing for the changes that need external QA, where we include the sections of "Acceptance Criteria" and "Setup Instructions/Link" . What do you think?

olerichter00:
@MounirDhahri do you think it would be easy to adjust the changelog script to include the sections?

MounirDhahri:
Yes, this should be possible by extending the list of available sections as we do in here https://github.com/artsy/eigen/blob/1fdfba74f74e859ef676de6870fbbf1139892a53/scripts/changelog/generateChangelogSectionTemplate.js#L3 and extending the PR body here https://github.com/artsy/eigen/blob/1fdfba74f74e859ef676de6870fbbf1139892a53/scripts/changelog/generatePlatformChangelog.js#L176

MounirDhahri (approved):

MounirDhahri:
@patrinoua what does `Setup Instructions/Link` refer to?

docs/pull_request_template.md
patrinoua: I don't see why not 🤷🏻‍♀️

patrinoua:
> @patrinoua what does Setup Instructions/Link refer to?
@MounirDhahri it was like that and I just copied it but I've now changed it to be just "Setup Instructions", which is what you need to do in order to run this PR.

patrinoua:
@MounirDhahri @olerichter00 the more I think about this, the more it seems to me like we may be repeating information.

Regarding the information we want to be providing for the ticket and the QA, I am guessing that the **Acceptance criteria** should be the same as the **Description** field, so maybe we could rename keep one field and rename it to **"Description, Acceptance Criteria, Setup instructions"**

Then we can keep the flag of [NEEDS EXTERNAL QA] and wherever there is that flag, we also add a link to the **"Description, Acceptance Criteria, Setup instructions"** part. 

What to do you think?

MounirDhahri:
> 
> @patrinoua what does Setup Instructions/Link refer to?
> 
> @MounirDhahri it was like that and I just copied it but I've now changed it to be just "Setup Instructions", which is what you need to do in order to run this PR.

thanks for explaining but now I am also lost with what Setup instructions mean here. Maybe we can remove it?

> Regarding the information we want to be providing for the ticket and the QA, I am guessing that the Acceptance criteria should be the same as the Description field, so maybe we could rename keep one field and rename it to "Description, Acceptance Criteria, Setup instructions"

I don't think we should nest it in the same field personally. IMO, if we should have an acceptance criteria field, we can have it separately otherwise I feel like we don't even need it since the acceptance criteria should be inside the Jira ticket.

docs/pull_request_template.md
pvinis: thats the idea. this table could automatically be added in the notion QA scripts. Also, it is something that should be "formalized" here, because jira tickets tend to be more loose in their wording, and when people from other teams see something, they are not 100% what to test. this table should make it simpler for everyone, and yes it might be a small extra work on the author, but this is for big feature changes, not small changes that can be QA'd within the team that developed them.

pvinis: with all that said, i definitely will insist this should remain as a table, that is replicated from notion. we should not have more things like more headers etc.

pvinis (changes requested):
i dont think we should remove the table.

patrinoua:
@pvinis I agree with collecting information for QA, but I'd like to examine if our current solution is effective by answering 1) is something that people use and 2) could it be improved? 

I think that the current implementation using a table is hard to fill out, so people may not be using it. 
It looks good when finished but while editing it's not that easy to use. So I'd be in favor of a format that's easier to fill out.
I'm interested to know if formatting is important to you for some reason and why, or if we could simplify this. 

The information, as discussed above, for some things would always be the same (eg staging) so perhaps we can remove it but we can of course encourage people to provide more information if they think it is needed, for example if some time it happens that the environment is not staging.

Test Case:
Acceptance Criteria:
Setup Instructions:

⬆️ The above is easier to fill out than the below ⬇️

Test Case | Feature | Environment | Acceptance Criteria | Setup Instructions/Link
-- | -- | -- | -- | --


And then I'd argue that 

Feature: Should come from the PR title

pvinis (approved):

PR #6612, opened 2022-04-14
fix(FX-3897): fix Sort & Filter choices reset when user scrolls down on Auction page

dimatretyak:
This reverts commit 3d84370322197d0ef10ed628a3cef9d21a50957e.

This PR resolves [FX-3897]

### PR Checklist

- [x] I tested my changes on **iOS** / **Android**.
- [ ] I added screenshots or videos to illustrate my changes.
- [ ] I added Tests and Stories for my changes.
- [ ] I added an [app state migration].
- [ ] I hid my changes behind a [feature flag].

### To the reviewers 👀

- [ ] I would like **at least one** of the reviewers to run this PR on the simulator or device.

<details><summary>Changelog updates</summary>

#### Cross-platform user-facing changes

-

#### iOS user-facing changes

-

#### Android user-facing changes

-

#### Dev changes

- Fix Sort & Filter choices reset when user scrolls down on Auction page - dimatretyak

</details>

Need help with something? Have a look on our [docs], or get in touch with us.

[app state migration]: /docs/adding_state_migrations.md
[feature flag]: /docs/developing_a_feature.md
[docs]: /docs/README.md

PR #7133, opened 2022-08-01
fix: cache problems for sale screen

dimatretyak:
This PR resolves []

Slack thread to get more context: https://artsy.slack.com/archives/C9SATFLUU/p1659353897000399?thread_ts=1659342586.520909&cid=C9SATFLUU

### Description
If the lot has moved to the “bidding closed” state and the info about this lot is in the cache (for example, the user previously opened this sale screen), then this lot will not be displayed at the end of the list

I think it’s related to [this problem](https://github.com/artsy/eigen/pull/6612#issue-1204382195) with `loadMore` (when only data from the cache is used and the fresh data received by the request is ignored) and some problem with cursor paginations/cache.

For example, in the first attached video, you can observe that **lot 7** and **lot 8** are not displayed at the end of the list until we manually update the data (but even after manually updating the data, go back from the screen and navigate to the screen again, **lot 7** and **lot 8** will not be displayed)

https://user-images.githubusercontent.com/3513494/182160164-7d736eeb-6f1f-4790-a15c-309152816eae.mp4

In the second video you can observe that **lot 10**-**lot 13** is not displayed

https://user-images.githubusercontent.com/3513494/182160287-cb9a769b-49f7-41e3-bf95-4bf3814d022e.mp4

### Ways to solve the problem:
* refactor the sales screen so that relay hooks are used instead of `createFragmentContainer`, `createPaginationContainer`, `QueryRenderer`
* add some sale queries to the allowed list so that the cache is always ignored for these queries 👈 Current PR

### QA Test Case(s)

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- cache problems for sale screen - dimatretyak

</details>

### DIFF
--- a/src/app/Scenes/Sale/Components/NewSaleLotsList.tsx
+++ b/src/app/Scenes/Sale/Components/NewSaleLotsList.tsx
@@ -181,7 +181,7 @@ export const NewSaleLotsListContainer = createPaginationContainer(
           first: $count
           after: $cursor
           input: $input
-          aggregations: [TOTAL]
+          aggregations: [TOTAL, FOLLOWED_ARTISTS]
         ) @connection(key: "NewSaleLotsList__artworksConnection") {
           counts {
             total
@@ -209,7 +209,7 @@ export const NewSaleLotsListContainer = createPaginationContainer(
       }
     },
     query: graphql`
-      query NewSaleLotsListQuery(
+      query NewSaleLotsListRefetchQuery(
         $saleID: ID!
         $count: Int
         $cursor: String
--- a/src/app/Scenes/Sale/Sale.tsx
+++ b/src/app/Scenes/Sale/Sale.tsx
@@ -9,7 +9,6 @@ import {
   ArtworkFilterNavigator,
   FilterModalMode,
 } from "app/Components/ArtworkFilter"
-import { DEFAULT_NEW_SALE_ARTWORKS_PARAMS } from "app/Components/ArtworkFilter/ArtworkFilterHelpers"
 import { ArtworkFiltersStoreProvider } from "app/Components/ArtworkFilter/ArtworkFilterStore"
 import { DEFAULT_NEW_SALE_ARTWORK_SORT } from "app/Components/ArtworkFilter/Filters/SortOptions"
 import { LoadFailureView } from "app/Components/LoadFailureView"
@@ -484,7 +483,7 @@ export const SaleQueryRenderer: React.FC<{
                       // @ts-ignore
                       input: {
                         sort: DEFAULT_NEW_SALE_ARTWORK_SORT.paramValue,
-                        priceRange: DEFAULT_NEW_SALE_ARTWORKS_PARAMS.priceRange,
+                        priceRange: "",
                       },
                     },
                   }
--- a/src/app/Scenes/Sale/Sale.tsx
+++ b/src/app/Scenes/Sale/Sale.tsx
@@ -9,6 +9,7 @@ import {
   ArtworkFilterNavigator,
   FilterModalMode,
 } from "app/Components/ArtworkFilter"
+import { DEFAULT_NEW_SALE_ARTWORKS_PARAMS } from "app/Components/ArtworkFilter/ArtworkFilterHelpers"
 import { ArtworkFiltersStoreProvider } from "app/Components/ArtworkFilter/ArtworkFilterStore"
 import { DEFAULT_NEW_SALE_ARTWORK_SORT } from "app/Components/ArtworkFilter/Filters/SortOptions"
 import { LoadFailureView } from "app/Components/LoadFailureView"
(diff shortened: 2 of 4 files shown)


src/app/Scenes/Sale/Components/NewSaleLotsList.tsx
```diff
@@ -181,7 +181,7 @@ export const NewSaleLotsListContainer = createPaginationContainer(
           first: $count
           after: $cursor
           input: $input
-          aggregations: [TOTAL]
+          aggregations: [TOTAL, FOLLOWED_ARTISTS]
```
dimatretyak, line 184: hack to solve [this problem](https://artsy.slack.com/archives/C9SATFLUU/p1659354970498689?thread_ts=1659342586.520909&cid=C9SATFLUU)

src/app/Scenes/Sale/Sale.tsx
```diff
@@ -483,6 +484,7 @@ export const SaleQueryRenderer: React.FC<{
                       // @ts-ignore
                       input: {
                         sort: DEFAULT_NEW_SALE_ARTWORK_SORT.paramValue,
+                        priceRange: DEFAULT_NEW_SALE_ARTWORKS_PARAMS.priceRange,
```
dimatretyak, line 487: for the initial and refetch query we need to send the same variables so that there are no problems with the fact that different data is displayed due to the cache

dimatretyak:
what do you think about these changes? By default all changes will be hidden under the feature flag, but maybe this flag will be marked as ready for release for 8.0.3

src/app/Scenes/Sale/Sale.tsx
```diff
@@ -483,6 +483,7 @@ export const SaleQueryRenderer: React.FC<{
                       // @ts-ignore
                       input: {
                         sort: DEFAULT_NEW_SALE_ARTWORK_SORT.paramValue,
+                        priceRange: "",
```
dimatretyak, line 486: for the initial and refetch query we need to send the same variables so that there are no problems with the fact that different data is returned for query due to the cache on Metaphysic/Gravity side

dimatretyak:
do you think we should release these changes in **8.0.3** (the changes are hidden under the feature flag, we will carefully QA it and make changes available to everyone if we don't find anything serious) or is it better to postpone them for the next version?

src/app/relay/middlewares/cacheMiddleware.ts
```diff
@@ -3,15 +3,22 @@ import { MiddlewareNextFn } from "react-relay-network-modern/node8"
 import { GraphQLRequest } from "./types"
 
 const IGNORE_CACHE_CLEAR_MUTATION_ALLOWLIST = ["ArtworkMarkAsRecentlyViewedQuery"]
+/**
+ * TODO:
+ * Replace NewSaleLotsListRefetchQuery with SaleLotsListQuery when AREnableArtworksConnectionForAuction is released
+ * or remove it when relay hooks will be used for Sale screen and cache problem for loadMore will be fixed
+ */
+const IGNORE_CACHE_QUERY_ALLOWLIST = ["NewSaleLotsListRefetchQuery"]
```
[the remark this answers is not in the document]
olerichter00: Nice solution! I still don't fully understand why setting `cacheConfig.force` to `true` doesn't work 🤔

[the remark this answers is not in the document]
dimatretyak: It works, but only for standard filters. If the user has applied some filters, then [this bug](https://github.com/artsy/eigen/pull/6612#issue-1204382195) occurs during pagination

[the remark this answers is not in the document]
dimatretyak: > Nice solution!

thanks 🙏

src/app/relay/middlewares/cacheMiddleware.ts
```diff
@@ -3,15 +3,22 @@ import { MiddlewareNextFn } from "react-relay-network-modern/node8"
 import { GraphQLRequest } from "./types"
 
 const IGNORE_CACHE_CLEAR_MUTATION_ALLOWLIST = ["ArtworkMarkAsRecentlyViewedQuery"]
+/**
+ * TODO:
+ * Replace NewSaleLotsListRefetchQuery with SaleLotsListQuery when AREnableArtworksConnectionForAuction is released
+ * or remove it when relay hooks will be used for Sale screen and cache problem for loadMore will be fixed
```
MounirDhahri, line 9: **nitpick:** a link to a Jira ticket here would be nice to make sure this is tracked in a team's board

MounirDhahri (approved):
Thanks for fixing this and sorry for the late review Dima. We could have released this but it's fine, we can release it next time.

src/app/relay/middlewares/cacheMiddleware.ts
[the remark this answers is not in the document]
olerichter00: Thanks for the explanation 🙏

src/app/relay/middlewares/cacheMiddleware.ts
dimatretyak: Added link to [Jira ticket](https://artsyproduct.atlassian.net/browse/FX-4133)

MounirDhahri (approved):

gkartalis (approved):
🔥 🔥 🔥

PR #7204, opened 2022-08-12
chore: hide median price screen under different ff

MounirDhahri:
### Description

This PR hides median price screen under different ff.

cc @dariakoko 

#nochangelog


dariakoko:
👍

PR #7212, opened 2022-08-15
chore: enable part 3 of insights

MounirDhahri:
### Description

Enable part 3 of insights

#nochangelog


MounirDhahri:
cc @dariakoko

PR #7254, opened 2022-08-23
fix: graph content pushed down when the text is too long

MounirDhahri:
### Description

This PR resolves [CX-2812]

When the content is too tall, the screen goes down. this won't happen after this fix since the text has fixed height now

**Before**

https://user-images.githubusercontent.com/11945712/186169349-3e55f7d1-d244-493c-8c16-9532a7aae509.mov

**After** Even if the text is long, the graph will keep position
Group 4

#### Acceptance Criteria

-

#### Setup Instructions

-

### PR Checklist

### To the reviewers 👀

- fix graph content pushed down when the text is too long - mounir

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-2812]: https://artsyproduct.atlassian.net/browse/CX-2812?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/palette/elements/LineGraph/LineGraphHeader.tsx
+++ b/src/palette/elements/LineGraph/LineGraphHeader.tsx
@@ -28,9 +28,13 @@ export const LineGraphHeader: React.FC<LineGraphHeaderProps> = ({
         </Flex>
       )}
       {!!text && (
-        <Text variant="xs" color="black60">
-          {text}
-        </Text>
+        // We need to have a fixed height here to make sure in case the text is too long,
+        // it doesn't push the graph content down
+        <Flex height={40}>
+          <Text variant="xs" color="black60">
+            {text}
+          </Text>
+        </Flex>
       )}
     </Flex>
   )
--- a/src/palette/elements/LineGraph/LineGraphHeader.tsx
+++ b/src/palette/elements/LineGraph/LineGraphHeader.tsx
@@ -1,5 +1,4 @@
 import { Flex, Text } from "palette"
-import { Dimensions } from "react-native"
 import { ColoredDot, DEFAULT_DOT_COLOR } from "./ColoredDot"
 import { LineChartData } from "./types"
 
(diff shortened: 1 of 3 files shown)


olerichter00 (dismissed):
Thanks for the fix! 🌟 

Have you tested the fix with font scaling? A fixed height might be not enough and we might want to multiply it by the scaling factor.

MounirDhahri:
@olerichter00 I haven't honestly, but If we expect to support font scaling there, I think we will also need to adjust other parts of the screen and I am leaning towards doing this separately

olerichter00:
Either way works for me. We can do it now or create a separate ticket.

Personally, I think that with any feature, we should make sure that the design supports font scaling to make the app accessible for all users.

MounirDhahri:
Good idea, I created a ticket for it and added it to the CX Up nex https://artsyproduct.atlassian.net/browse/CX-2825

olerichter00:
The rest of the screen looks ok with font scaling. Just the bottom margin is missing ([Ticket](https://artsyproduct.atlassian.net/browse/CX-2824)).

src/palette/elements/LineGraph/LineGraphHeader.tsx
```diff
@@ -28,9 +28,13 @@ export const LineGraphHeader: React.FC<LineGraphHeaderProps> = ({
         </Flex>
       )}
       {!!text && (
-        <Text variant="xs" color="black60">
-          {text}
-        </Text>
+        // We need to have a fixed height here to make sure in case the text is too long,
+        // it doesn't push the graph content down
+        <Flex height={40}>
```
olerichter00, line 33:
```suggestion
        <Flex height={PixelRatio.getFontScale() * 40}>
```

olerichter00 (commented):
Well, let's fix it now. Picking up the ticket later will be more work.

PR #7270, opened 2022-08-24
feat(GRO-1235): create a custom onboarding results grid

laurabeth:
### Description

This PR resolves [GRO-1235]

This PR creates a custom grid to circumvent some of the current limitations w/ our navigation stack, while also allowing users to save artworks during the onboarding flow.

https://user-images.githubusercontent.com/14044896/186535353-e94b1f9c-d004-4fb2-ba9c-fb4dc97363dd.mp4

#### Acceptance Criteria

- Displays artworks in a fashion identical to the in-app artwork grid
- Does not link to artist or artwork pages
- Has a functional “save” button for artworks

#### Setup Instructions

-

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[GRO-1235]: https://artsyproduct.atlassian.net/browse/GRO-1235?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Components/ArtworkGrids/utils/index.ts
+++ b/src/app/Components/ArtworkGrids/utils/index.ts
@@ -1,0 +1,2 @@
+export * from "./layout"
+export * from "./sections"
--- a/src/app/Components/ArtworkGrids/utils/layout.ts
+++ b/src/app/Components/ArtworkGrids/utils/layout.ts
@@ -1,0 +1,15 @@
+const BREAKPOINT = 700
+
+export function calculateLayoutValues(width: number, sectionMargin: number) {
+  const isPad = width > BREAKPOINT
+
+  const sectionCount = isPad ? 3 : 2
+  const sectionMargins = sectionMargin ?? 0 * (sectionCount - 1)
+  const artworkPadding = 20
+  const sectionDimension = (width - sectionMargins * sectionCount - artworkPadding) / sectionCount
+
+  return {
+    sectionCount,
+    sectionDimension,
+  }
+}
--- a/src/app/Scenes/Onboarding/OnboardingV2/Components/OnboardingResultsGrid.tsx
+++ b/src/app/Scenes/Onboarding/OnboardingV2/Components/OnboardingResultsGrid.tsx
@@ -1,0 +1,126 @@
+import { OnboardingResultsGridQuery } from "__generated__/OnboardingResultsGridQuery.graphql"
+import { extractNodes } from "app/utils/extractNodes"
+import { Box, Flex, OpaqueImageView, Text, TextProps } from "palette"
+import { FC, useState } from "react"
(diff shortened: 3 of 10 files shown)


src/app/Components/ArtworkGrids/utils/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./layout"
+export * from "./sections"
```
damassi, line 2: (non-blocking): Since the number of functions we're exporting here is only 3 spread across two files, I'd suggest skipping the index file and just importing the files directly from where they're used. Less ambiguous data in the system (`* from`) is always better. It also helps confuse the typescript auto-importer less, as sometimes it will auto import from the index path, sometimes directly from the file, etc. 

Force is a perfect case study in the messes that index files, `export * from` and the TS auto importer creates. Check out the diff here https://github.com/artsy/force/pull/10606 to see what I mean. No `export * from`, no problem = less information in the system.

src/app/Components/ArtworkGrids/utils/layout.ts
```diff
@@ -0,0 +1,15 @@
+const BREAKPOINT = 700
+
+export function calculateLayoutValues(width: number, sectionMargin: number) {
```
damassi, line 3: Since `layout.ts` only contains one function, we should name the file after the said exported function and keep things `1:1` (another example of less information in the system)

damassi (approved):
Nice work on this 🎉  lots of little wins. Only two minor non-blocking comments; take em or leave em 👍

src/app/Scenes/Onboarding/OnboardingV2/Components/OnboardingResultsGrid.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+export type GridItem = NonNullable<
+  NonNullable<OnboardingResultsGrid_connection$data["edges"]>[number]
+>["node"]
+
+interface OnboardingResultsGridProps {
+  connection: OnboardingResultsGrid_connection$key
+}
+
+export const OnboardingResultsGrid: FC<OnboardingResultsGridProps> = ({ connection }) => {
```
damassi, line 26: Is this grid more generalizable (since it fixes a few problems more generally)? maybe we can pull it out and put it in `Components/ArtworkGrids` and then import it into here for the relay bits.

gkartalis (approved):
I would say let's merge this one to be on time for the release train, and I will open a pr addressing all comments from @damassi! 

I pushed two commits, one adjusting the right padding of the grid and also removing the one test that was breaking (due to us removing the ability to navigate to the artwork page).

PR #7284, opened 2022-08-26
feat: allow releasing specific versions of apps

brainbicycle:
### Description

This PR resolves [MOPLAT-199]

Changes our promote scripts to present the last ~20 beta builds and ask the user to specify and confirm which build should be promoted. 

Here is an example of build selection on iOS, if you select 'y' it will continue to go through the promotion process with the selected build:

```
$ yarn setup:releases # get env vars necessary to release
$ ./scripts/promote-beta-to-submission-ios 
# ...... a lot of output, then eventually:
[15:25:13]: --------------------------
[15:25:13]: --- Last 20 iOS builds ---
[15:25:13]: --------------------------
[15:25:13]: Which build would you like to release?: 
1. 2022.8.26.14
2. 2022.8.26.6
3. 2022.8.25.16
4. 2022.8.25.6
5. 2022.8.24.11
6. 2022.8.24.6
7. 2022.8.23.15
8. 2022.8.23.8
9. 2022.8.23.6
10. 2022.8.23.5
11. 2022.8.22.12
12. 2022.8.18.6
13. 2022.8.17.6
14. 2022.8.16.14
15. 2022.8.16.6
16. 2022.8.15.18
17. 2022.8.15.10
18. 2022.8.13.6
19. 2022.8.12.14
20. 2022.8.12.13
?  7
[15:25:19]: Are you sure you would like to release '2022.8.23.15'? (y/n)
n
+---------------+------------------+
|           Lane Context           |
+---------------+------------------+
| PLATFORM_NAME |                  |
| LANE_NAME     | promote_beta_ios |
+---------------+------------------+
[15:25:22]: Stopping the train!
```

#### Acceptance Criteria

- Next release we should release this way, if it works it works, if any changes are needed or it doesn't work reach out to mobile platform. 

#### Setup Instructions

-

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-  Adds ability to release specific versions of the app - Brian

</details>

[MOPLAT-199]: https://artsyproduct.atlassian.net/browse/MOPLAT-199?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


fastlane/Fastfile
```diff
@@ -183,6 +192,7 @@ lane :ship_beta_android_play_store do
     skip_upload_metadata: true,
     skip_upload_images: true,
     skip_upload_screenshots: true,
+    version_codes_to_retain: past_version_codes
```
brainbicycle, line 195: supply by default overwrites old android builds so they cannot be selected for promotion unless we explicitly say to retain them

fastlane/Fastfile
```diff
@@ -5,6 +5,9 @@ require 'spaceship'
 require 'fileutils'
 require 'json'
 
+require 'dotenv'
+Dotenv.load('../.env.shared')
```
MounirDhahri, line 9: just to make sure, .env.shared in ci is the same as .env.ci right?

MounirDhahri: asking in case we deploy with wrong keys

fastlane/Fastfile
```diff
@@ -212,21 +222,32 @@ lane :ship_beta_android_firebase do
...
+
+  UI.header "Last 20 Android builds"
+
+  selected_version_code = UI.select("Which build would you like to release?: ", version_codes)
+
+  if UI.confirm("Are you sure you would like to release '#{selected_version_code}'?")
+    UI.success "Continuing the release!"
+  else
+    UI.user_error!("Stopping the train!")
+  end
```
MounirDhahri, line 238: **praise:** i like this confirmation question

MounirDhahri (commented):
LGTM, i only left a question about env vars but otherwise all look good to me. Thanks for adding this Brian

fastlane/Fastfile
brainbicycle: good catch ! we do want to use .env.ci here, updated

MounirDhahri (approved):

patrinoua:
Wow, thanks for doing that @brainbicycle ! 🎉 🙏 
A couple of minor suggestions:
- Offer the last 10 builds instead of 20 (realistically speaking 20 sounds too much?)
- provide an example of what the script expects like that:  ``` [15:25:13]: Which build would you like to release? (eg. 7) : ``` 
although I would  personally be more a fun of removing the numbers and just providing a list of build numbers like so:
```
2022.8.26.14
2022.8.26.6
2022.8.25.16
2022.8.25.6
2022.8.24.11
```
with text : ``` [15:25:13]: Which build would you like to release? (eg. 2022.8.26.14) : ``` 

Thanks again, this will definitely ease up our process and avoid mistakes in the future!

MounirDhahri:
If you don't mind, I would like to merge this in order to test it. cc @patrinoua

MounirDhahri:
failed when prompting for the app store API key :(

brainbicycle:
> failed when prompting for the app store API key :(

sorry I should have mentioned, someone overwrote the key in our env file to an old value so I updated it, if you run `yarn:setup-artsy` you should get the latest

brainbicycle:
> Wow, thanks for doing that @brainbicycle ! 🎉 🙏 A couple of minor suggestions:
> 
> * Offer the last 10 builds instead of 20 (realistically speaking 20 sounds too much?)
> * provide an example of what the script expects like that:
> * `[15:25:13]: Which build would you like to release? (eg. 7) :`
>   although I would  personally be more a fan of removing the numbers and just providing a list of build numbers like so:
> 
> ```
> 2022.8.26.14
> 2022.8.26.6
> 2022.8.25.16
> 2022.8.25.6
> 2022.8.24.11
> ```
> 
> with text : `[15:25:13]: Which build would you like to release? (eg. 2022.8.26.14)`
> 
> Thanks again, this will definitely ease up our process and avoid mistakes in the future!

thanks! I think limiting to 10 builds makes sense.

on the prompt, I am using fastlanes default selection prompt which uses the single digit numbers:
`selected_version_code = UI.select("Which build would you like to release?: ", version_codes)`,

I could potentially do what you suggest and write the prompt myself but I kind of like only having to type 1 char and think it is not really worth the extra code to write it ourselves? 

I can definitely update the prompt to have an example input though as you suggest.

MounirDhahri:
@brainbicycle I will try it out the next time for sure then 🙌 thanks for mentioning the envs issue

patrinoua:
@brainbicycle can we please add the example?

PR #7314, opened 2022-09-07
fix(CX-2899): fix highlight scroll to display more

lordkiz:
### Description
- Adds scrollview to allow users scroll down to reveal more
- Fixes uncategorized National Pavillion highlight kind

https://user-images.githubusercontent.com/18648835/188833778-7411881a-38b0-469b-af3b-842b388e4228.mp4

This PR resolves [CX-2899]

#### Acceptance Criteria

-

#### Setup Instructions

-

### PR Checklist

### To the reviewers 👀

- fix highlight scroll to display more - kizito

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Scenes/MyCollection/Screens/Insights/Components/CareerHighlightBottomSheetItem.tsx
+++ b/src/app/Scenes/MyCollection/Screens/Insights/Components/CareerHighlightBottomSheetItem.tsx
@@ -1,5 +1,16 @@
+import { BottomSheetScrollView } from "@gorhom/bottom-sheet"
 import { uniq } from "lodash"
-import { Box, FairIcon, Flex, GroupIcon, PublicationIcon, SoloIcon, Text } from "palette"
+import {
+  ArtworkIcon,
+  Box,
+  FairIcon,
+  Flex,
+  GroupIcon,
+  PublicationIcon,
+  SoloIcon,
+  Spacer,
+  Text,
+} from "palette"
 import { useMemo } from "react"
 import { useScreenDimensions } from "shared/hooks"
 import { CareerHighlightKindValueType } from "../CareerHighlightBottomSheet"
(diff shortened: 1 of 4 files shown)


src/app/Scenes/MyCollection/Screens/Insights/Components/CareerHighlightBottomSheetItem.tsx
```diff
@@ -55,6 +72,18 @@ const SectionedHighlight: React.FC<{ header: string; body: string[] }> = ({ head
           bodyCount > 1 ? "Included in multiple major biennials" : "Included in a major biennial"
         IconComponent = <FairIcon />
         break
+      case "National Pavillion":
```
MounirDhahri, line 75: **thought:** It would have been great if we were using these values as enums instead of strings

MounirDhahri (approved):
Niiice - LGTM, feel free to merge once Sultan approves as well

MounirDhahri:
One thought came to my mind - because of the current implementation, we are unable to support new career highlights I believe. I think this could have probably been avoided if we moved the display logic into metaphysics. It's fine as it is now because this will require quite a bit of work, it's something to keep in mind though for the future

src/app/Scenes/MyCollection/Screens/Insights/Components/CareerHighlightBottomSheetItem.tsx
lordkiz: They are types which we end up using as key type for a map. 
I can't remember why I had opted for a dictionary instead of enums. I think it was so I can easily map headers. I cant remember for sure, but will check.

https://github.com/artsy/eigen/pull/7314/files#diff-19e072b8adfbcfd4aedca53c77fc3dc7a30aeec02615b2649961e26b8e90af9dL19

lordkiz:
> One thought came to my mind - because of the current implementation, we are unable to support new career highlights I believe. I think this could have probably been avoided if we moved the display logic into metaphysics. It's fine as it is now because this will require quite a bit of work, it's something to keep in mind though for the future

Yesss. You are absolutely correct. In the long term that should be what we have to do.

src/app/Scenes/MyCollection/Screens/Insights/Components/CareerHighlightBottomSheetItem.tsx
MounirDhahri: awesome, thanks

MounirDhahri:
Please feel free to merge this if you need to do further testing, i will assign it back to you

PR #7358, opened 2022-09-14
feat: Increase max number of works per artist for new for you GRO-1297

jonallured:
This is a sibling PR to https://github.com/artsy/force/pull/10983 and updates Eigen in the same way for the same reasons.

https://artsyproduct.atlassian.net/browse/GRO-1297

/cc @artsy/grow-devs @shaq31415926 @isakelaris

### DIFF
--- a/src/app/Scenes/Home/Components/NewWorksForYouRail.tsx
+++ b/src/app/Scenes/Home/Components/NewWorksForYouRail.tsx
@@ -74,7 +74,7 @@ export const NewWorksForYouRail: React.FC<NewWorksForYouRailProps & RailScrollPr
 
 const artworksFragment = graphql`
   fragment NewWorksForYouRail_artworkConnection on Viewer {
-    artworksForUser(includeBackfill: true, first: 20) {
+    artworksForUser(maxWorksPerArtist: 3, includeBackfill: true, first: 20) {
       edges {
         node {
           title
--- a/src/app/Scenes/NewWorksForYou/NewWorksForYou.tsx
+++ b/src/app/Scenes/NewWorksForYou/NewWorksForYou.tsx
@@ -12,7 +12,7 @@ import { Box, SimpleMessage, Spacer } from "palette"
 import { createPaginationContainer, graphql, QueryRenderer, RelayPaginationProp } from "react-relay"
 
 const SCREEN_TITLE = "New Works for You"
-const PAGE_SIZE = 20
+const PAGE_SIZE = 100
 
 interface NewWorksForYouProps {
   relay: RelayPaginationProp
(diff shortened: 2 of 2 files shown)


src/app/Scenes/Home/Components/NewWorksForYouRail.tsx
```diff
@@ -74,7 +74,7 @@ export const NewWorksForYouRail: React.FC<NewWorksForYouRailProps & RailScrollPr
 
 const artworksFragment = graphql`
   fragment NewWorksForYouRail_artworkConnection on Viewer {
-    artworksForUser(includeBackfill: true, first: 20) {
+    artworksForUser(maxWorksPerArtist: 3, includeBackfill: true, first: 20) {
```
shaq31415926, line 77: Just for my clarification, what does this first: 20 do?

jonallured: There's a set of arguments in GraphQL that deal with pagination and this is one of them. Used on it's own it's simply saying "give me the first 20" but you can get fancier with cursors and page through a connection of items. More here in the docs!

https://graphql.org/learn/pagination/

shaq31415926: aah ok, thanks for the explanation. Just for my reference, why do we limit to the first 20?

shaq31415926: Could we increase this please? As we show more artworks per artist, we are potentially capping relevant artworks for the user?

jonallured: We just need to pick a number! Feel free to suggest a new page size and I can update it here!

shaq31415926: I am not sure why we are capping it. I would say 100?

jonallured: If a user does not have that many then backfill kicks in and then if there still isn't enough then it's just a limit and it'll show less. Does that make sense?

shaq31415926: Lets go with 100 then?!

jonallured: Done!

jonallured: Oh damn I'm totally wrong - this is for the rail on the home screen - 20 is very reasonable here! 😬 Going to go back to 20.

shaq31415926: aah ok, that makes sense haha.

src/app/Scenes/NewWorksForYou/NewWorksForYou.tsx
```diff
@@ -12,7 +12,7 @@ import { Box, SimpleMessage, Spacer } from "palette"
 import { createPaginationContainer, graphql, QueryRenderer, RelayPaginationProp } from "react-relay"
 
 const SCREEN_TITLE = "New Works for You"
-const PAGE_SIZE = 20
+const PAGE_SIZE = 100
```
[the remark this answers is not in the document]
jonallured: Good catch - I've updated the count in the other place so this should work now. I'll verify in a beta once this is merged!

pam- (approved):

PR #7454, opened 2022-10-06
fix(NX-3253): add purchase expired and accepted states in conversation

tam-kis:
### Description

Added `Purchase Accepted` and `Purchase Expired` states in conversations.
Refactored tests because `renderWithWrappersLEGACY` is deprecated.

This PR resolves [NX-3253](https://artsyproduct.atlassian.net/browse/NX-3253)

Simulator Screen Shot - iPhone 13 - 2022-09-27 at 14 56 10

### QA Testing
Check if `Purchase Accepted` and `Purchase Expired` are showing up after a purchase is made in Conversations.

#### Acceptance Criteria

-

#### Setup Instructions

-

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Scenes/Inbox/Components/Conversations/OrderUpdate.tests.tsx
+++ b/src/app/Scenes/Inbox/Components/Conversations/OrderUpdate.tests.tsx
@@ -1,8 +1,6 @@
 import { OrderUpdateTestsQuery } from "__generated__/OrderUpdateTestsQuery.graphql"
 import { navigate } from "app/navigation/navigate"
-import { extractText } from "app/tests/extractText"
-import { renderWithWrappersLEGACY } from "app/tests/renderWithWrappers"
-import { AlertCircleFillIcon, LinkText, MoneyFillIcon } from "palette"
+import { renderWithWrappers } from "app/tests/renderWithWrappers"
 import "react-native"
 import { QueryRenderer } from "react-relay"
 import { act } from "react-test-renderer"
@@ -77,7 +75,7 @@ const getWrapper = (event = {}) => {
       ],
     },
   }
-  const tree = renderWithWrappersLEGACY(<TestRenderer />)
+  const tree = renderWithWrappers(<TestRenderer />)
   const finalResolvers = { Conversation: () => mockConversation }
   act(() => {
     env.mock.resolveMostRecentOperation((operation) =>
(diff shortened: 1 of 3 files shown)


src/app/Scenes/Inbox/Components/Conversations/OrderUpdate.tsx
```diff
@@ -74,9 +75,9 @@ export const OrderUpdate: React.FC<OrderUpdateProps> = ({ event, conversationId
     } else if (orderUpdateState === "offer_rejected") {
       color = "red100"
       message = `Offer Declined`
-    } else if (orderUpdateState === "offer_lapsed") {
+    } else if (state === "CANCELED" && reasonLapsed) {
```
lean8086, line 78: Question (maybe for @lidimayra): should this condition be on the [order status helper](https://github.com/artsy/exchange/blob/main/lib/order_status_helper.rb) instead of force+eigen?

(btw, I made the [same comment on Force's PR](https://github.com/artsy/force/pull/10370#discussion_r907090335) but it was already merged)

lidimayra: Yess, that's a very good point!

I currently don't like the idea that the client side has to handle this kind of logic. It forces us to maintain this flow in different projects and it ends up leading to bugs like the one that's being fixed here (force was updated and eigen wasn't).

However, moving in that direction would require some decisions that should be defined and planned beforehand, and it would also increase the effort involved in this fix (which is focused on displaying the missing states to the user).

I remember @araujobarret registered [this ticket](https://artsyproduct.atlassian.net/browse/NX-3254) because there were already some thoughts regarding the way Force is currently handling it. That one was more focused on Force specifically, but I wonder if we should also move this discussion (server-side and client-side responsibilities) to that ticket and try to pull it in the next refinement session. It seems the kind of thing that we would also like to have TX team inputs as well (as taking these kind of decisions inside NX-only already raised some issues [in the past](https://artsy.slack.com/archives/C9YNS4X32/p1658951883431909))

lean8086: Thanks for the insights! I see how these decisions can impact other teams such as TX. Happy to discuss this as part of that follow-up ticket 👍

src/app/Scenes/Inbox/Components/Conversations/OrderUpdate.tests.tsx
```diff
@@ -93,132 +92,115 @@ it("renders without throwing an error", () => {
...
+    expect(getByText("Offer Declined")).toBeTruthy()
   })
 
   it("displays the offer expired event", () => {
-    const tree = getWrapper({
+    const { getByText } = getWrapper({
       __typename: "CommerceOrderStateChangedEvent",
       orderUpdateState: "offer_lapsed",
+      state: "CANCELED",
+      stateReason: ["_lapsed"],
```
lean8086, line 117: Not a blocker: not sure if I'm misinterpreting this, but I'd use a real possible value here.

src/app/Scenes/Inbox/Components/Conversations/OrderUpdate.tests.tsx
```diff
@@ -93,132 +92,115 @@ it("renders without throwing an error", () => {
...
 
-    expect(extractText(tree.root)).toMatch("You purchased this artwork")
-    expect(extractText(tree.root)).toMatch("See details")
-    tree.root.findByType(MoneyFillIcon)
+  it("shows an expired buy order", () => {
+    const { getByText } = getWrapper({
+      __typename: "CommerceOrderStateChangedEvent",
+      orderUpdateState: "buy_lapsed",
+      state: "CANCELED",
+      stateReason: ["_lapsed"],
```
lean8086, line 194: Same as my previous comment (unless I'm misinterpreting it!)

lean8086 (approved):
Looks good!! I left a non-blocking comment.

PR #7458, opened 2022-10-07
fix(CX-2965): Fix BottomSheet randomly showing up on ArtistSelect

lordkiz:
### Description
The BottomSheet randomly shows up on android when you select an artist to view their MedianSalePriceChart. This appears to be a longstanding react native BottomSheet issue on android. 
This PR changes our display logic and loads the BottomSheet only when a highlight has been selected instead of loading it before hand and making it visible when highlight has been selected.

Other approaches may include:
1. changing `android:windowSoftInputMode="adjustResize"` for MainActivity (which hosts the RN screens) in AndroidManifest to `android:windowSoftInputMode="adjustPan"` . But this approach might have unintended effect on how we handle KeyboardAvoiding view on android.
2. Set a timeout that forceCloses BottomSheet when we select a new artist - There will be a very jarring flicker and we cannot be 100% certain of how long to wait.

This PR resolves [CX-2965]

#### BEFORE

https://user-images.githubusercontent.com/18648835/194565055-cc94d5d4-40b0-4cbd-b59e-48149b2ccf35.mp4

-
-
-
#### AFTER

https://user-images.githubusercontent.com/18648835/194565087-78855dab-3061-4679-b1d4-fb25eb5160fb.mp4

#### Acceptance Criteria

-

#### Setup Instructions

-

### PR Checklist

### To the reviewers 👀

- Fix BottomSheet randomly showing up on ArtistSelect - kizito

#### iOS user-facing changes

-

- 

#### Dev changes

-

</details>

[CX-2965]: https://artsyproduct.atlassian.net/browse/CX-2965?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/MyCollection/Screens/Insights/CareerHighlightBottomSheet.tsx
+++ b/src/app/Scenes/MyCollection/Screens/Insights/CareerHighlightBottomSheet.tsx
@@ -1,6 +1,7 @@
 import BottomSheet, { BottomSheetBackdrop } from "@gorhom/bottom-sheet"
 import { CareerHighlightBottomSheet_query$key } from "__generated__/CareerHighlightBottomSheet_query.graphql"
 import { LegacyBackButtonContext } from "app/navigation/NavStack"
+import { delay } from "app/utils/delay"
 import { isPad } from "app/utils/hardware"
 import { compact } from "lodash"
 import { Flex } from "palette"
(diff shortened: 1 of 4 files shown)


src/app/Scenes/MyCollection/Screens/Insights/MedianSalePriceAtAuction.tsx
```diff
@@ -38,6 +38,19 @@ const MedianSalePriceAtAuctionScreen: React.FC<MedianSalePriceAtAuctionProps> =
...
+
+  useEffect(() => {
+    if (!isFirstMount.current && newArtistID) {
+      refetch(newArtistID)
+      return
+    }
+    if (isFirstMount.current) {
+      isFirstMount.current = false
+    }
+  }, [newArtistID])
```
lordkiz, line 53: @MrSltun I had to move the refetch logic on new Artist select to this effect because of 2 reasons:
1. Before the modal is fully dismissed, we trigger the refetch and so we fallback on the Suspense Skeleton fallback even before the modal is fully dismissed. By the time we try to update the component with the new data, we will get the `Can't perform a React state update on an unmounted component.` warning. This fixes that. But please suggest any better approach.
2. We do not dismiss the modal fast enough as it sometimes waits for refetch to complete and also if you look closely you will find the Skeleton fallback component being dismissed like a modal when we get new data.

src/app/Scenes/MyCollection/Screens/Insights/CareerHighlightBottomSheet.tsx
```diff
@@ -141,30 +123,39 @@ export const CareerHighlightBottomSheet: React.FC<CareerHighlightBottomSheetProp
...
-          <CareerHighlightBottomSheetItem year={item.year} highlights={item.highlights} />
-        )}
-        keyExtractor={(item) => item.year.toString()}
-        horizontal
-        pagingEnabled
-        showsHorizontalScrollIndicator={false}
-      />
-    </BottomSheet>
-  )
+  if (selectedXAxisHighlight !== null) {
```
olerichter00, line 126: **suggestion (non-blocking):** Here we could use the [early return pattern](https://medium.com/swlh/return-early-pattern-3d18a41bba8) to make it easier to read.

src/app/Scenes/MyCollection/Screens/Insights/CareerHighlightBottomSheet.tsx
```diff
@@ -141,30 +123,39 @@ export const CareerHighlightBottomSheet: React.FC<CareerHighlightBottomSheetProp
...
+          horizontal
+          pagingEnabled
+          showsHorizontalScrollIndicator={false}
+          onScrollToIndexFailed={({ index }) => {
+            const delay = new Promise((resolve) => setTimeout(resolve, 100))
+            delay.then(() => {
+              flatlistRef.current?.scrollToIndex({ index, animated: false })
+            })
+          }}
+        />
```
olerichter00, line 154: **question:** What exactly is this code doing and what is it needed for?

olerichter00: **suggestion:** We could introduce a `wait` helper function similar to the one we [use in Force](https://github.com/artsy/force/blob/main/src/Utils/wait.ts#L1)

olerichter00 (approved):
Thanks a lot for the fix 🌟 

I've only added a few questions and suggestions.

src/app/Scenes/MyCollection/Screens/Insights/CareerHighlightBottomSheet.tsx
lordkiz: Will refactor. 👍

src/app/Scenes/MyCollection/Screens/Insights/CareerHighlightBottomSheet.tsx
lordkiz: > **suggestion:** We could introduce a `wait` helper function similar to the one we [use in Force](https://github.com/artsy/force/blob/main/src/Utils/wait.ts#L1)

Looks like we already have a delay helper function. I didn't know we do.

olerichter00: Thanks for the explanation! 

`requestAnimationFrame` could potentially work here as well. Using `requestAnimationFrame` we could avoid the timeout to get a more deterministic behaviour. Maybe we can give it a try. Anyway if it gets too complicated, the timeout should work fine here as well 👍

lordkiz: Thanks @olerichter00  I just checked. And the flatlist throws an error demanding that we use getItemLayout or onScrollIndexFailed. This is correct because if the index we are trying to scrollto is greater than `zero` then that index will greater than the highest available index that the VirtualizedList is aware of, since it initialises that with 0: https://github.com/facebook/react-native/blob/43636267011a97490ed7495b08e500c5d0d54872/Libraries/Lists/VirtualizedList.js#L901

Also, this `_highestMeasuredFrameIndex` is calculated as each item internally calls `_onCellLayout` making it more of  a layout related thing, rather than a View's frame information as it animates over time into or out of view as with requestAnimationFrame.

So if we attempt to scroll to an index > 0 before the calculations are done, then we will have an error. This block is very explanatory: https://github.com/facebook/react-native/blob/43636267011a97490ed7495b08e500c5d0d54872/Libraries/Lists/VirtualizedList.js#L204

olerichter00: Alright, let's go with the timeout, I'll mark this conversation as resolved 💯

olerichter00 (approved):

PR #7495, opened 2022-10-18
feat(FX-4280): add shipping and taxes section on artwork page

dimatretyak:
This PR resolves [FX-4280] 

#### Description
* if the work has [any availability](https://github.com/artsy/gravity/blob/22bb3752463dccb316c4baa5e99258057f2ba4c8/app/models/domain/availability.rb) other than "for sale" we should **completely hide** the Shipping & Taxes section
* Otherwise, we'll always have some information to show in that section.
  * Shipping
    * If we have a `shippingOrigin`, show it ([might be null](https://github.com/artsy/metaphysics/blob/a4e42e52adc88775f6ece6d7402dbbe192051f6a/src/schema/v2/artwork/index.ts#L1130-L1137))
    * If we have `shippingInfo`, show it ([doesn't seem to return null](https://github.com/artsy/metaphysics/blob/a4e42e52adc88775f6ece6d7402dbbe192051f6a/src/schema/v2/artwork/index.ts#L1053))
  * Taxes
    * Show `priceIncludesTaxDisplay` ([currently nullable](https://github.com/artsy/metaphysics/blob/a4e42e52adc88775f6ece6d7402dbbe192051f6a/src/schema/v2/artwork/index.ts#L968-L973)

### Demo
https://user-images.githubusercontent.com/3513494/198628703-bf726227-1cb0-4a09-88e2-5f6aa0530e70.mp4

### PR Checklist

### To the reviewers 👀

- add shipping and taxes section on artwork page - dimatretyak

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[FX-4280]: https://artsyproduct.atlassian.net/browse/FX-4280?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/Scenes/Artwork/Artwork.tsx
```diff
@@ -96,6 +97,10 @@ export const Artwork: React.FC<ArtworkProps> = ({
     }
   }
 
+  const shouldRenderShippingAndTaxes = () => {
+    return !isInAuction && artworkBelowTheFold?.isEligibleForOnPlatformTransaction
```
dimatretyak, line 101: **question**: should we move this to the Metaphysics side? Something like this

```javascript
isEligibleForShippingAndTaxes: {
  type: new GraphQLNonNull(GraphQLBoolean),
  description: "Artwork is eligible for the Shipping and taxes",
  resolve: (artwork) => {
    const isInAuction = checkIsInAuction(artwork)
    const isEligibleForTransaction = isEligibleForOnPlatformTransaction(artwork)
    
    return !isInAuction && isEligibleForTransaction
  },
}
```

```graphql
{
  artwork(id: "artworkId") {
    isEligibleForShippingAndTaxes
  }
}
```

dimatretyak: **question**: should we check that the artwork is in an auction, and if yes, hide the "Shipping and taxes" section?

dblandin, line 101: **suggestion**: Can we simplify this conditional to not explicitly exclude artworks at auction? I think the MP-backed conditional should be sufficient as I don't think any artworks at auction would _also_ be enabled in any of our on-platform purchase options.

src/app/Scenes/Artwork/Components/CommercialInformation.tsx
```diff
@@ -205,12 +204,7 @@ export const CommercialInformation: React.FC<CommercialInformationProps> = ({
       newSaleMessage = "For sale"
     }
 
-    return (
-      <>
-        <SaleAvailability saleMessage={newSaleMessage ? newSaleMessage : saleMessage} />
-        {!artworkIsInClosedAuction && <CommercialPartnerInformation artwork={artwork} />}
-      </>
-    )
+    return <SaleAvailability saleMessage={newSaleMessage ? newSaleMessage : saleMessage} />
```
dblandin, line 207: **note / non-blocking**: We should see if we can push some of these sale message changes up to Metaphysics as well.

src/app/Scenes/Artwork/Artwork.tsx
dblandin: I think we might be able to skip the !isInAuction check unless we can point to some specific examples of artworks at auction which are _also_ somehow enabled for on-platform purchase options.

src/app/Scenes/Artwork/Components/CommercialInformation.tsx
dimatretyak: I think it's a good suggestion

src/app/store/config/features.ts
```diff
@@ -146,11 +146,6 @@ export const features = defineFeatures({
     description: "Enable placeholder layout animation",
     echoFlagKey: "AREnablePlaceholderLayoutAnimation",
   },
-  AREnableAvalaraPhase2: {
```
gkartalis, line 149: 😬 that was unused?

dimatretyak: this feature flag was not enabled by default, and in the new design "Taxes may apply at checkout" label is rendered by default. So, i think this flag is no longer relevant for this reason 🤷‍♂️

src/app/Scenes/Artwork/Artwork.tsx
dimatretyak: @dblandin do you think we need to check the `isSold` state as it is done on [the Force side](https://github.com/artsy/force/blob/main/src/Apps/Artwork/Components/ArtworkSidebar2/ArtworkSidebar2.tsx#L127)?

dimatretyak: in this case, the "isEligibleForOnPlatformTransaction" field will be enough and we should **NOT** add "isEligibleForShippingAndTaxes?

src/app/Scenes/Artwork/Components/CommercialInformation.tsx
dimatretyak: https://artsyproduct.atlassian.net/browse/FX-4389

gkartalis (dismissed):

src/app/Scenes/Artwork/Artwork.tsx
```diff
@@ -308,6 +309,13 @@ export const Artwork: React.FC<ArtworkProps> = ({
       })
     }
 
+    if (!!artworkBelowTheFold?.isEligibleForOnPlatformTransaction) {
```
dblandin, line 312: We might want to add `|| isInquireable` to check as well. While we are encouraging more activity on-platform, not all of our patterns are in regions which support on-platform transactions and they still have the ability to provide shipping information, even if the artwork is for sale and ineligible.

dimatretyak: **question**: maybe it's worth adding a new `isEligibleForShippingAndTaxes` field for the Metaphysi, so as not to duplicate checks for shipping and taxes section between Force/Eigen?

dimatretyak: > We might want to add || isInquireable to check as well

Should we also check that the artwork is not in auction in this case?
Artsy - Metaphysics (Staging) – Sandbox 2022-10-24 15-09-02

dblandin: The tax details seem to be included in the [shippingInfo](https://github.com/artsy/metaphysics/blob/7384cb1c1c8473806da64b09d34a17e1684ea9fb/src/schema/v2/artwork/index.ts#L1053) query field. We could add some new logic to that query field to return null when we want to hide this section. Is it reasonable to conditionally render this entire section based on whether the `shippingInfo` query field returns data or not?

dimatretyak: > Is it reasonable to conditionally render this entire section based on whether the shippingInfo query field returns data or not

in theory, yes. But it seems to me it's better to add the `isEligibleForShippingAndTaxes` field, so as not to break the logic somewhere, which is based on `shippingInfo`

dblandin: Yeah, you're probably right. Let's plan on spinning up that field in Metaphysics. We can bikeshed the name in that PR :)

src/app/Scenes/Artwork/Artwork.tsx
dimatretyak: [Here](https://artsy.slack.com/archives/C9SATFLUU/p1666875689139339) are the conditions when the "shipping and taxes" section will be displayed

dimatretyak:
Reminder: [double check changes](https://github.com/artsy/metaphysics/pull/4478#pullrequestreview-1159814169) on Metaphysics side

gkartalis (dismissed):

PR #7496, opened 2022-10-18
feat: Add Braze content cards to home feed GRO-1323

jonallured:
After some tinkering and getting feedback, I've landed here for getting Braze Content Cards in the app:

Screen Shot 2022-10-18 at 2 44 26 PM

They are coming in below the New Works For You rail and match this Figma file:

https://www.figma.com/file/k465puwy6YQOcEESEkppni/Smaller-Tasks-2022?node-id=317%3A2415

The first commit here removes the Trove item from the homefeed because that will morph into one of these content cards instead.

Next, my approach to adding the content cards was to mostly follow the ideas in this integration doc from them:

https://www.braze.com/docs/developer_guide/platform_integration_guides/react_native/content_cards/

And in terms of the Home component looks like this:

* add an item to the `modules` list for content cards
* add a `ContentCards` component to the switch statement on Home

This ContentCards component starts life as `null` and only pops in after an async call into the Braze sdk. In a `useEffect` hook I setup an event listener and then trigger it with a call to `ReactAppboy.requestContentCardsRefresh()`. Once this is resolved then I can update the state and cause a render with the cards array populated. If a user is not eligible for any cards, the homefeed will continue to NOT have this rail. This populating of the rail with content is done in the Braze dashboard and is outside the scope of this PR.

Ok so we have the cards array - next up we use a FlatList to horizontally render them and move onto a few more responsibilities:

* we have to register an impression as the user swipes through the cards
* we have to register a "click" when the user taps on a card

Those are just sdk function calls inside of handlers.

Last up was dealing with the indicator that there are more cards to look at. This is the bit underneath with the dots. I asked @pvinis and @brainbicycle about this and found that we have something in Palette for this:

https://palette.artsy.net/elements/progressdots/

But this is not in Eigen. Instead the closest we were able to find was the indicator in the artwork screen:

https://github.com/artsy/eigen/blob/main/src/app/Scenes/Artwork/Components/ImageCarousel/ImageCarouselPaginationIndicator.tsx

So I pondered adapting that code but in the end I just copy/pasted it over to this area and altered it to suite my needs. I'd love to improve this! Maybe I could extract a parent component so that we can use this elsewhere in Eigen? I'd also love to do the variant with the lines rather than the dots so that I could more closely match the Figma. But my feeling is that I can do all that type of work after landing this PR - what do you think?

https://artsyproduct.atlassian.net/browse/GRO-1323

/cc @artsy/grow-devs @kathytafel

### DIFF
--- a/src/app/Scenes/Home/Components/ContentCards.tsx
+++ b/src/app/Scenes/Home/Components/ContentCards.tsx
@@ -1,0 +1,110 @@
+import { Box, Button, Flex, Spacer, Text, Touchable } from "palette"
+import ReactAppboy from "react-native-appboy-sdk"
+import OpaqueImageView from "app/Components/OpaqueImageView/OpaqueImageView"
+import { navigate } from "app/navigation/navigate"
+import { useScreenDimensions } from "shared/hooks"
+import { FlatList } from "react-native"
+import React, { useCallback, useEffect, useState } from "react"
+import { PaginationDots } from "./PaginationDots"
+
+interface CardProps {
+  item: ReactAppboy.CaptionedContentCard
+}
+
+const ContentCard: React.FC<CardProps> = ({ item }) => {
+  const { width: screenWidth } = useScreenDimensions()
+  const height = 250
+  const imageWidth = 125
+  const handlePress = () => {
+    ReactAppboy.logContentCardClicked(item.id)
+    item.url && navigate(item.url)
+  }
+
+  return (
+    <Touchable key={item.id} onPress={handlePress}>
+      <Flex bg="black100" flexDirection="row" height={height} width={screenWidth}>
+        <OpaqueImageView
+          height={height}
+          imageURL={item.image}
+          resizeMode="cover"
+          width={imageWidth}
+        />
+        <Box p={2} width={screenWidth - imageWidth}>
+          <Text color="white100" mb={1} numberOfLines={2} variant="lg-display">
+            {item.title}
+          </Text>
+          <Text color="white100" mb={2} numberOfLines={3}>
+            {item.cardDescription}
+          </Text>
+          <Button size="small" variant="outlineLight" onPress={handlePress}>
+            {item.domain}
+          </Button>
+        </Box>
+      </Flex>
+    </Touchable>
+  )
+}
+
+export const ContentCards: React.FC = () => {
+  const [cards, setCards] = useState([] as ReactAppboy.CaptionedContentCard[])
+  const [currentCardIndex, setCurrentCardIndex] = useState(0)
+  const [viewedCards, setViewedCards] = useState([] as ReactAppboy.CaptionedContentCard[])
+
+  useEffect(() => {
+    const eventName = ReactAppboy.Events.CONTENT_CARDS_UPDATED
+    const callback = async () => {
+      const updatedCards = await ReactAppboy.getContentCards()
+      const sortedCards = updatedCards.sort((lhs, rhs) =>
+        lhs.extras.position > rhs.extras.position ? 1 : -1
+      )
+      setCards(sortedCards as ReactAppboy.CaptionedContentCard[])
+    }
+
+    const listener = ReactAppboy.addListener(eventName, callback)
+    ReactAppboy.requestContentCardsRefresh()
+
+    return () => {
+      listener.remove()
+    }
+  }, [])
+
+  const handleViewableItemsChanged = useCallback(
+    (viewable) => {
+      const viewableCards = viewable.viewableItems.map(
+        (viewableItem: any) => viewableItem.item
+      ) as ReactAppboy.CaptionedContentCard[]
+      const lastShown = viewableCards[viewableCards.length - 1]
+      const newCardIndex = cards.findIndex((card) => card.id === lastShown.id)
+      setCurrentCardIndex(newCardIndex)
+      const filteredCards = viewableCards.filter((card) => !viewedCards.includes(card))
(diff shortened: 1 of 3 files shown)


src/app/Scenes/Home/Components/PaginationDots.tsx
```diff
@@ -0,0 +1,43 @@
...
+  active: boolean
+}
+
+const PaginationDot: React.FC<PaginationDotProps> = (props) => {
+  const { active } = props
+  const opacity = active ? 1 : 0.1
+  const diameter = 5
+
+  return (
+    <Animated.View
```
MounirDhahri, line 14: **Q:** I am wondering why you are using an Animated View here if opacity is not animated.

MounirDhahri: You can use the `useAnimatedValue` hook to define the opacity and set it when the `active` value changes like we do in here https://github.com/artsy/eigen/blob/main/src/app/Scenes/Artwork/Components/ImageCarousel/FullScreen/ImageCarouselFullScreen.tsx#L51

src/app/Scenes/Home/Components/ContentCards.tsx
```diff
@@ -0,0 +1,110 @@
...
+            {item.domain}
+          </Button>
+        </Box>
+      </Flex>
+    </Touchable>
+  )
+}
+
+export const ContentCards: React.FC = () => {
+  const [cards, setCards] = useState([] as ReactAppboy.CaptionedContentCard[])
```
MounirDhahri, line 49: **nitpick:** `useState` supports generics now so something like this `useState<ReactAppboy.CaptionedContentCard[]>()` could also be used. It's actually the pattern we use in other screens

src/app/Scenes/Home/Components/ContentCards.tsx
```diff
@@ -0,0 +1,110 @@
...
+import { PaginationDots } from "./PaginationDots"
+
+interface CardProps {
+  item: ReactAppboy.CaptionedContentCard
+}
+
+const ContentCard: React.FC<CardProps> = ({ item }) => {
+  const { width: screenWidth } = useScreenDimensions()
+  const height = 250
+  const imageWidth = 125
```
MounirDhahri, line 17: **nitpick:** since we are not using these anywhere else, i would suggest to move them outside of the component

src/app/Scenes/Home/Components/ContentCards.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+    const listener = ReactAppboy.addListener(eventName, callback)
+    ReactAppboy.requestContentCardsRefresh()
+
+    return () => {
+      listener.remove()
+    }
+  }, [])
+
+  const handleViewableItemsChanged = useCallback(
```
MounirDhahri, line 71: **praise:** 👏

src/app/Scenes/Home/Home.tsx
```diff
@@ -316,9 +323,6 @@ const Home = (props: Props) => {
                     mb={MODULE_SEPARATOR_HEIGHT}
                   />
                 )
-
-              case "trove":
-                return <TroveFragmentContainer trove={item.data} mb={MODULE_SEPARATOR_HEIGHT} />
```
MounirDhahri, line 321: Goodbye Trove 👋

MounirDhahri: @jonallured I remember @mdole speaking of some changes to the Trove and I am wondering if you two spoke together about this removal. Unless the Trove is now coming from Braze

MounirDhahri (commented):
LGTM, I left some small comments, please let me know if something isn't clear

src/app/Scenes/Home/Home.tsx
jonallured: Yeah we've been in contact about this and all is good here!

src/app/Scenes/Home/Components/PaginationDots.tsx
jonallured: Good call, updated with [d5e3ce7](https://github.com/artsy/eigen/pull/7496/commits/d5e3ce793e411066ee284f60774351440f17ca19) per your code suggestion, thanks so much!!

src/app/Scenes/Home/Components/PaginationDots.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { Animated } from "react-native"
+
+interface PaginationDotProps {
+  active: boolean
+}
+
+const PaginationDot: React.FC<PaginationDotProps> = (props) => {
+  const { active } = props
+  const diameter = 5
+  const opacity = useRef(new Animated.Value(active ? 1 : 0.1)).current
```
pvinis, line 12: i do not want to block, but I would be curious to see if this code would be easier in this style https://github.com/artsy/eigen/blob/fa86ad4777752a196c76c254d1de20a465ee632c/src/app/Scenes/Onboarding/OnboardingWelcome.tsx#L35-L39.

again, not blocking 👍

pvinis (approved):
good stuff!

assigning back to you, merge when you want 👍

MounirDhahri (approved):

src/app/Scenes/Home/Components/PaginationDots.tsx
jonallured: Ok going to merge but I can tinker with this in a follow up PR! 👍

PR #7497, opened 2022-10-18
fix: 3rd party auth issues

brainbicycle:
This PR resolves [MOPLAT-464] 

### Description

Fixes some bugs in sign in and sign up:

- Catches uncaught exceptions in Facebook and Google sign in and presents an error
- Logs out Facebook users on general logout to avoid an edge case preventing sign in
- Doesn't fall through after resolving or rejecting promises to fix mysterious hanging issue
- Bonus: adds an error message when trying to login with google in firebase betas to make it less confusing to devs

Also: there is a little debug tool to log auth actions that was reverted since it makes the auth code messy but it could be useful if people need to test this in the future : https://github.com/artsy/eigen/pull/7497/commits/3ade8c071d4a858a1ef5e28c724e82dba9b7c005

### PR Checklist

### To the reviewers 👀

- Fix bugs in third party authentication - Brian 

#### iOS user-facing changes

-

-

#### Dev changes

- Add an error message when trying to sign in with google on firebase beta - Brian 

</details>

[MOPLAT-464]: https://artsyproduct.atlassian.net/browse/MOPLAT-464?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/store/AuthModel.ts
```diff
@@ -484,107 +486,142 @@ export const getAuthModel = (): AuthModel => ({
...
+        console.log("FBSDK stop 2")
 
-        if (!result || !result.email) {
+        if (declinedPermissions?.includes("email")) {
           reject(
-            new AuthError(
-              "There is no email associated with your Facebook account. Please log in using your email and password instead."
-            )
+            new AuthError("Please allow the use of email to continue.", "Email Permission Declined")
           )
```
brainbicycle, line 501:
```suggestion
          )
          return
```
add return

src/app/store/AuthModel.ts
```diff
@@ -484,107 +486,142 @@ export const getAuthModel = (): AuthModel => ({
...
+        }
+
+        console.log("FBSDK stop 3")
+
```
brainbicycle, line 505:
```suggestion
```

src/app/store/AuthModel.ts
```diff
@@ -484,107 +486,142 @@ export const getAuthModel = (): AuthModel => ({
...
 
-        if (options.signInOrUp === "signUp") {
-          const resultGravitySignUp = await actions.signUp({
-            email: result.email as string,
-            name: result.name as string,
-            accessToken: accessToken.accessToken,
-            oauthProvider: "facebook",
-            agreedToReceiveEmails: options.agreedToReceiveEmails,
-          })
+        console.log("FBSDK stop 4")
```
brainbicycle, line 512:
```suggestion
```

src/app/store/AuthModel.ts
```diff
@@ -484,107 +486,142 @@ export const getAuthModel = (): AuthModel => ({
...
-        reject(
-          new AuthError("Please allow the use of email to continue.", "Email Permission Declined")
-        )
-      }
-      const accessToken = !isCancelled && (await AccessToken.getCurrentAccessToken())
-      if (!accessToken) {
-        return
-      }
+      try {
+        console.log("FBSDK stop 1")
```
brainbicycle, line 490:
```suggestion
```

src/app/store/AuthModel.ts
```diff
@@ -484,107 +486,135 @@ export const getAuthModel = (): AuthModel => ({
...
-        return
-      }
-
-      const responseFacebookInfoCallback = async (error: any | null, result: any | null) => {
-        if (error) {
-          reject(new AuthError("Error fetching facebook data", error))
-          return
-        }
+      try {
+        const { declinedPermissions, isCancelled } = await LoginManager.logInWithPermissions([
```
brainbicycle, line 490: the `loginWithPermissions` call can throw and this leads to login failing silently, wrap in a try catch instead

src/app/store/AuthModel.ts
```diff
@@ -484,107 +486,135 @@ export const getAuthModel = (): AuthModel => ({
...
-            body: {
-              oauth_provider: "facebook",
-              oauth_token: accessToken.accessToken,
-              client_id: clientKey,
-              client_secret: clientSecret,
-              grant_type: "oauth_token",
-              scope: "offline_access",
-            },
-          })
+              return
```
brainbicycle, line 541: ran into a strange issue where authFacebook was being called in a loop and hanging, I believe this was because we were falling through after promises but I am not 100%

src/app/store/AuthModel.ts
```diff
@@ -768,6 +798,7 @@ export const getAuthModel = (): AuthModel => ({
         ? await LegacyNativeModules.ArtsyNativeModule.clearUserData()
         : Promise.resolve(),
       await signOutGoogle(),
+      LoginManager.logOut(),
```
brainbicycle, line 801: if a use logged in with one account and logs out before and then logged in with a different account it would fail because the token for the other user was still hanging around, clear it out on log out. Hopefully an edge case but it also happens a lot if you are testing so we should handle it.

src/app/store/AuthModel.ts
```diff
@@ -484,53 +486,171 @@ export const getAuthModel = (): AuthModel => ({
...
+        new GraphRequestManager().addRequest(infoRequest).start()
+      } catch (e) {
+        if (e instanceof Error) {
+          if (e.message === "User logged in as different Facebook user.") {
+            // odd and hopefully shouldn't happen often
+            // if the user has a valid session with another account
+            // and tries to log in with a new account they will hit this error
+            // log them out and try again
+            LoginManager.logOut()
+            GlobalStore.actions.auth.authFacebook(options)
```
brainbicycle, line 609: since we are now logging out users on logout in theory this should not happen but we can still handle it in case

src/app/store/AuthModel.ts
```diff
@@ -554,104 +674,42 @@ export const getAuthModel = (): AuthModel => ({
...
+          if (e.message === "DEVELOPER_ERROR") {
+            reject(
               new AuthError(
-                resultGravitySignUp.message,
-                resultGravitySignUp.error,
-                resultGravitySignUp.meta
+                "Google auth does not work in firebase beta, try again in a playstore beta",
+                e.message
               )
             )
```
brainbicycle, line 705: google auth is not working in firebase betas and right now it is just failing silently, let's show devs an error message otherwise it is unclear what is happening

src/app/store/AuthModel.ts
pvinis: i would love to move these promises to async/await code, with throws for rejections, maybe that could help simplify this code.

pvinis (approved):

src/app/store/AuthModel.ts
brainbicycle: yeah I would be up to give it a shot! I will leave it out of this PR for fear of breaking things but maybe after the release?

PR #7517, opened 2022-10-20
fix: broken tap areas in widget GRO-1348

jonallured:
Something I had not considered is what the widget does when you tap on areas that are not wrapped with a link. Turns out the app will crash if I don't tell it what to do!

So with this PR all I'm doing is wrapping the "Latest Articles" text and the artsy logo in links directing the user to the home feed.

https://artsyproduct.atlassian.net/browse/GRO-1348

/cc @artsy/grow-devs


pvinis:
works 👍 

https://user-images.githubusercontent.com/100233/197051177-9562efd9-9ffd-4e59-b1a8-f4db6e4ae799.mov

jonallured:
Big thanks go to @pvinis for assisting me in verifying this fix!! 🙌

pvinis (approved):

PR #7523, opened 2022-10-21
feat: remove design system v2 things

pvinis:
This PR resolves [] 

### Description

Everything is replaced with the actual value that it has had for a while now. We used to replace `0.3` with `5` pixels for a while. Now we removed the actual `0.3` for the correct v3 value which is `0.5`, which also translates to `5` pixels.

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Components/AnimatableHeader/AnimatableHeader.tsx
+++ b/src/app/Components/AnimatableHeader/AnimatableHeader.tsx
@@ -25,14 +25,14 @@ export const AnimatableHeader = (props: AnimatableHeaderProps) => {
     <Animated.View
       style={{
         flexDirection: "row",
-        paddingHorizontal: space(2),
+        paddingHorizontal: space("2"),
         alignItems: "center",
         height: headerHeight,
         backgroundColor: "white",
       }}
     >
       <TouchableOpacity
-        hitSlop={{ top: space(1), bottom: space(1), left: space(1), right: space(1) }}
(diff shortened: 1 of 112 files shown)


olerichter00 (approved):
🧹 🌟

gkartalis (dismissed):
nice cleanup, will this affect the styling of some sections though? 😬

pvinis:
we should definitely check, but there should be no styling change. the reason is that we already have all v2 made into v3 on the fly, so now that is gone and replaced with the actual values of v3. I guess I could have missed one or two, though. if that happened, then yea maybe some spacer will have the wrong size, if I missed a 0.3 to 0.5 conversation 🤔.

the reason I opened this now is that it's after the release, so we will have time. also, if there is something I missed, it will not be anything big like the text size, it will only be a spacing one, which is pretty easy to fix 😬.

how does that sound?

pvinis:
i think ill generate a beta and merge this on Tuesday, when I'm back.

dariakoko (dismissed):
nice cleanup 💪

pvinis:
this is ready. ill make a quick beta to test things out on phone.

gkartalis:
@pvinis Yell if you need an extra pair of 👁️ for an extra check

pvinis:
just tested the beta, seems great.

PR #7526, opened 2022-10-21
feat: add artsy guarantee section in artwork screen

gkartalis:
This PR resolves [FX-4281] 

### Description

Adds artsy guarantee section and tests, learn more redirect user to webview `/buyer-guarantee`

#### Screenshots

Simulator Screen Shot - iPhone 13 Pro - 2022-10-21 at 10 44 38

### PR Checklist

### To the reviewers 👀

- add artsy guarantee section in artwork screen - gkartalis

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[FX-4281]: https://artsyproduct.atlassian.net/browse/FX-4281?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/Scenes/Artwork/Artwork.tsx
```diff
@@ -307,6 +308,13 @@ export const Artwork: React.FC<ArtworkProps> = ({
       })
     }
 
+    if (artworkAboveTheFold?.isSold !== true && artworkAboveTheFold?.isEligibleForArtsyGuarantee) {
```
dimatretyak, line 311: **question**: do we need to check the `isSold` state, checking only `isEligibleForArtsyGuarantee` will not be enough?

gkartalis: We still need to check for the sold state as we do in [force](https://github.com/artsy/force/blob/main/src/Apps/Artwork/Components/ArtworkSidebar2/ArtworkSidebar2.tsx#L84-L85). 

The reason for that is that it is client specific logic, metaphysics doesn't check for [this](https://github.com/artsy/metaphysics/pull/4460/files#diff-54b84e0c164089fc202e3439d8ded89ab4925442fb613f4e36a9c22acdb66cb8R671-R681)

dimatretyak: hmm, should we check the `isSold` for shipping and taxes section in this case?

dimatretyak: if I understand correctly `isOfferable`, `isAcquireable`, `isOfferableFromInquiry` will be set to `false` when `isSold` is `true`. It seems to me in this case `isEligibleForArtsyGuarantee` will be enough

https://user-images.githubusercontent.com/3513494/197174657-824a5f20-4b15-491c-9cb0-5a45b0850ecb.mp4

dblandin: You might be right. Looks like all of these purchase option booleans flip to false if the work has sold ([source](https://github.com/artsy/gravity/blob/ffd6b8cb3e17d0258c732d88a4a47d71956e3bbe/app/models/util/for_sale.rb#L282-L306) 🔒).

gkartalis: interesting, happy to make this change then in both force/eigen

dblandin (dismissed):
LGTM! Agree with Dima that we might not need the additional isSold check after all.

dimatretyak (approved):

PR #7528, opened 2022-10-21
fix: Back button on MyC artwork screen

olerichter00:
This PR resolves [CX-3079] 

### Description

To prevent landing on the Request Price Estimate confirmation screen when pressing the back button on the MyC artwork screen, this PR changes the logic of the back button to always navigate the user to the My Collection overview.

💁 Not sure if there is a better fix for this bug. Please let me know if you have any ideas! 

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-3079]: https://artsyproduct.atlassian.net/browse/CX-3079?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/MyCollection/Screens/Artwork/MyCollectionArtwork.tsx
+++ b/src/app/Scenes/MyCollection/Screens/Artwork/MyCollectionArtwork.tsx
@@ -51,6 +51,11 @@ const MyCollectionArtwork: React.FC<MyCollectionArtworkScreenProps> = ({
     )
   }
 
+  const handleGoBack = () => {
+    // Navigating back to the MyCollection tab on the profile screen
+    navigate("/my-profile")
+  }
+
   const handleEdit = useCallback(() => {
     trackEvent(tracks.editCollectedArtwork(data.artwork!.internalID, data.artwork!.slug))
     GlobalStore.actions.myCollection.artwork.startEditingArtwork(data.artwork as any)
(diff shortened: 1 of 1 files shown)


MounirDhahri:
Thanks for this fix @olerichter00 . I am thinking now that having this as a screen now is not really the best way to go as I believe this should be modal or modal screen, that we can use `dismissModal` to close and to avoid handling any back logic

lordkiz:
With react-navigation, We can access the stacked routes on Request Price Estimate screen and pop to the desired route when overriding the back behavior. The profile in this stacked routes will be the root of the stack. So should be pretty straightforward. And I will suggest we should instead use the available prevent back flow as documented in react navigation to handle this overriding.
We can quickly do this on Monday morning @olerichter00 🙏🏽

olerichter00:
@lordkiz, that sounds way better than the solution in this PR. I'll close this PR in favor of your solution 🙏

MounirDhahri:
Not sure if that's possible yet since we are not really handling things with react-navigation in iOS yet so far so it's not really possible to do that yet there unless we bring the same logic to native code.

olerichter00:
We could also merge this PR now to make sure to have the new behaviour in the next release and try to rewrite the flow in a better way. What do you think @MounirDhahri and @lordkiz ?

lordkiz:
@MounirDhahri with the new app shell, we can handle these with react-navigation on iOS. For eg in this PR: https://github.com/artsy/eigen/pull/7532
@olerichter00 sure. If we decide to merge this however, instead of `navigate('profile')` we should `popToRoot`. Reason is that though since profile is a root tab, internally navigation will pop to it. But this behaviour can change and we'll end up with bugs if someone changes the default configuration.

olerichter00:
Great, I'll keep this one closed then in favor of https://github.com/artsy/eigen/pull/7532

PR #7561, opened 2022-10-31
feat(spike):  toggle bottom navbar in artwork screen  #doNotMerge 

gkartalis:
This PR resolves [] 

### Description

Do not merge, this is exploring how we could toggle the bottom navbar on the artwork screen

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>


src/app/BottomTabsToggleProvider.tsx
```diff
@@ -0,0 +1,49 @@
...
+    case "hide": {
+      return { visible: false }
+    }
+  }
+}
+
+function ToggleBottomTabProvider({ children }: ToggleBottomTabProviderProps) {
+  const [state, dispatch] = React.useReducer(toggleBottomTabReducer, { visible: false })
+  // NOTE: you *might* need to memoize this value
+  // Learn more in http://kcd.im/optimize-context
```
dimatretyak, line 32: yep, agree with it

src/app/BottomTabsToggleProvider.tsx
```diff
@@ -0,0 +1,49 @@
...
+      return { visible: true }
+    }
+    case "hide": {
+      return { visible: false }
+    }
+  }
+}
+
+function ToggleBottomTabProvider({ children }: ToggleBottomTabProviderProps) {
+  const [state, dispatch] = React.useReducer(toggleBottomTabReducer, { visible: false })
```
dimatretyak, line 30: **suggestion**: maybe just `useState` will be enough? Something like this

```javascript
const { useMemo } = require("react")
const { useState } = require("react")

...

function ToggleBottomTabProvider({ children }: ToggleBottomTabProviderProps) {
  const [visible, setVisible] = useState(true)

  const showBottomTab = () => {
    setVisible(true) 
  }

  const hideBottomTab = () => {
    setVisible(false)
  }

  const value = useMemo(() => ({
    showBottomTab,
    hideBottomTab,
    visible,
  }), [visible])

  return (
    <ToggleBottomTabStateContext.Provider value={value}>
      {children}
    </ToggleBottomTabStateContext.Provider>
  )
}
```

dimatretyak, line 30: **suggestion**: I think it would be better to set `visible` to `true` by default.

dimatretyak (approved):
Great spike and changes 👌

src/app/App.tsx
```diff
@@ -14,6 +14,7 @@ import RNBootSplash from "react-native-bootsplash"
 import { Settings } from "react-native-fbsdk-next"
 import RNShake from "react-native-shake"
 import { AppProviders } from "./AppProviders"
+import { ToggleBottomTabProvider } from "./BottomTabsToggleProvider"
```
dimatretyak, line 17: **suggestion**: I think we can also use the existing [BottomTabsModel](https://github.com/artsy/eigen/blob/5948d8cff7bb8707daf0955aba01971481c40963/src/app/Scenes/BottomTabs/BottomTabsModel.ts#L14)

src/app/Scenes/Artwork/Artwork.tsx
```diff
@@ -87,6 +90,16 @@ export const Artwork: React.FC<ArtworkProps> = ({
...
 
+  useEffect(() => {
+    if (isInAuction) {
+      dispatch({ type: "hide" })
+    }
+
+    return () => {
+      dispatch({ type: "show" })
+    }
+  }, [isInAuction])
```
dimatretyak, line 101: **question**: as far as I understand, with this implementation, there may be a problem with the fact that the bottom tabs will be hidden if you switch to another tab using `navigate` (for example, by clicking on some link on artwork screen), since the artwork screen will not be removed from stack navigations?

src/app/App.tsx
dimatretyak: Sandbox branch: https://github.com/artsy/eigen/tree/dimatretyak/spike/FX-4305

gkartalis: nice! was thinking to close this (since it was exploratory eitherway) and we can continue with the one above ☝️

PR #7574, opened 2022-11-02
fix: Improve support for dynamic text in content cards GRO-1403

jonallured:
Turns out it's pretty easy to get at the font scale for users that have increased it! Docs here:

https://reactnative.dev/docs/pixelratio#getfontscale

And so @pvinis had the idea to just use this scale to grow the content card height and that does in fact ensure that the button shows up.

<details>
<summary>screenshots</summary>

Screen Shot 2022-11-02 at 9 16 51 AM
Screen Shot 2022-11-02 at 9 17 12 AM

</details>

https://artsyproduct.atlassian.net/browse/GRO-1403

/cc @artsy/grow-devs

### DIFF
--- a/src/app/Scenes/Home/Components/ContentCards.tsx
+++ b/src/app/Scenes/Home/Components/ContentCards.tsx
@@ -2,7 +2,7 @@ import OpaqueImageView from "app/Components/OpaqueImageView/OpaqueImageView"
 import { navigate } from "app/navigation/navigate"
 import { Box, Button, Flex, Spacer, Text, Touchable } from "palette"
 import React, { useEffect, useRef, useState } from "react"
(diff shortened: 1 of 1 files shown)


pvinis:
We thought we could switch the layout on high font scales, so that the image is on top and text at the bottom, so that there is more horizontal space for the text to do its thing. 🤔 though, like @jonallured said, it's a whole thing again to go through design, which is annoying to block devs, but 🤷.

pvinis (dismissed):

jonallured:
Ok paired with @kathytafel and made a couple more improvements by clamping the image width and increasing the number of description lines - looks like this:

Screen Shot 2022-11-02 at 10 48 24 AM

PR #7551, opened 2022-10-28
refactor(FX-4392): auction artwork screen

gkartalis:
This PR resolves [FX-4392] 

### Description

- [x] refactor artworkTombstone component from class -> functional
- [x] update styles of request condition report cta
- [x] move lot number to the top of the artwork page

#### Screenshots

|Request condition report button|Lot on top of Artwork page|
|---|---|
|Simulator Screen Shot - iPhone 14 Pro - 2022-10-31 at 15 54 43|Simulator Screen Shot - iPhone 14 Pro - 2022-10-31 at 15 52 23|

### PR Checklist

- [x] I tested my changes on **iOS** / **Android**.
- [x] I added screenshots or videos to illustrate my changes.
- [ ] I added Tests and Stories for my changes.
- [ ] I added an [app state migration].
- [ ] I hid my changes behind a [feature flag].
- [ ] I have prefixed changes that need to be tested during a release QA with **[NEEDS EXTERNAL QA]** on the changelog.

### To the reviewers 👀

- [ ] I would like **at least one** of the reviewers to run this PR on the simulator or device.

<details><summary>Changelog updates</summary>

#### Cross-platform user-facing changes

-  auction artwork screen minor style changes - gkartalis

#### iOS user-facing changes

-

#### Android user-facing changes

-

#### Dev changes

-

</details>

Need help with something? Have a look at our [docs], or get in touch with us.

[app state migration]: ../blob/main/docs/adding_state_migrations.md
[feature flag]: ../blob/main/docs/developing_a_feature.md
[docs]: ../blob/main/docs/README.md

[FX-4392]: https://artsyproduct.atlassian.net/browse/FX-4392?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #7577, opened 2022-11-02
feat(FX-4419): Auctions progressbar/timer move to header

gkartalis:
This PR resolves [FX-4419] 

### Description

Moves timer label and progress bar to the top of the artwork page.
Feature is behind a feature flag

|Screenshots|||
|---|---|---|
|Simulator Screen Shot - iPhone 14 - 2022-12-01 at 16 54 41|Simulator Screen Shot - iPhone 14 - 2022-12-01 at 16 54 08|Simulator Screen Shot - iPhone 14 - 2022-12-01 at 16 57 23|

### PR Checklist

### To the reviewers 👀

- Auctions progressbar/timer move to header (behind a ff) - gkartalis

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[FX-4419]: https://artsyproduct.atlassian.net/browse/FX-4419?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Components/Bidding/Components/ArtworkAuctionProgressBar.tsx
+++ b/src/app/Components/Bidding/Components/ArtworkAuctionProgressBar.tsx
@@ -7,6 +7,7 @@ export interface ArtworkAuctionProgressBarProps {
   extendedBiddingIntervalMinutes: number
   biddingEndAt?: string | null
   hasBeenExtended: boolean
+  height?: number
 }
 
 export const ArtworkAuctionProgressBar: React.FC<ArtworkAuctionProgressBarProps> = ({
@@ -15,6 +16,7 @@ export const ArtworkAuctionProgressBar: React.FC<ArtworkAuctionProgressBarProps>
   extendedBiddingIntervalMinutes,
   biddingEndAt,
   hasBeenExtended,
+  height,
 }) => {
   if (!biddingEndAt) {
     return null
@@ -43,6 +45,10 @@ export const ArtworkAuctionProgressBar: React.FC<ArtworkAuctionProgressBarProps>
   const renderProgressBar = isWithinExtendedBiddingPeriod || hasBeenExtended
 
   return (
-    <>{renderProgressBar && <ProgressBar trackColor="red100" progress={percentComplete * 100} />}</>
+    <>
+      {renderProgressBar && (
+        <ProgressBar height={height} trackColor="red100" progress={percentComplete * 100} />
+      )}
+    </>
   )
 }
--- a/src/app/Components/Bidding/Components/ArtworkAuctionProgressBar.tsx
+++ b/src/app/Components/Bidding/Components/ArtworkAuctionProgressBar.tsx
@@ -42,13 +42,9 @@ export const ArtworkAuctionProgressBar: React.FC<ArtworkAuctionProgressBarProps>
   const percentComplete =
     (parsedSecondsUntilEnd + parsedMinutesUntilEnd * 60) / (extendedBiddingDuration * 60)
 
-  const renderProgressBar = isWithinExtendedBiddingPeriod || hasBeenExtended
+  if (!(isWithinExtendedBiddingPeriod || hasBeenExtended)) {
+    return null
+  }
 
-  return (
-    <>
-      {renderProgressBar && (
-        <ProgressBar height={height} trackColor="red100" progress={percentComplete * 100} />
-      )}
-    </>
-  )
+  return <ProgressBar height={height} trackColor="red100" progress={percentComplete * 100} />
 }
--- a/src/app/Components/Bidding/Components/Timer.tsx
+++ b/src/app/Components/Bidding/Components/Timer.tsx
@@ -1,7 +1,6 @@
 import { StateManager as CountdownStateManager } from "app/Components/Countdown"
 import { CountdownTimerProps } from "app/Components/Countdown/CountdownTimer"
 import { ModernTicker, SimpleTicker } from "app/Components/Countdown/Ticker"
-import { useFeatureFlag } from "app/store/GlobalStore"
 import { DateTime } from "luxon"
 import moment from "moment-timezone"
 import { Flex, Spacer, Text } from "palette"
@@ -161,12 +160,6 @@ export const Countdown: React.FC<CountdownProps> = ({
   extendedBiddingPeriodMinutes,
   biddingEndAt,
 }) => {
-  const enableArtworkRedesign = useFeatureFlag("ARArtworkRedesingPhase2")
-
-  if (!!enableArtworkRedesign) {
-    return null
-  }
-
   return (
     <Flex alignItems="center" accessibilityLabel="Countdown">
       {cascadingEndTimeIntervalMinutes ? (
@@ -186,7 +179,7 @@ export const Countdown: React.FC<CountdownProps> = ({
       <Text variant="xs" weight="medium" color="black60">
         {label}
       </Text>
-      {!!(extendedBiddingPeriodMinutes && !enableArtworkRedesign) && (
+      {!!extendedBiddingPeriodMinutes && (
         <>
           <Spacer mt={1} />
           <Text variant="xs" color="black60" textAlign="center">
--- a/src/app/Components/Bidding/Components/Timer.tsx
+++ b/src/app/Components/Bidding/Components/Timer.tsx
@@ -163,6 +163,24 @@ export const Countdown: React.FC<CountdownProps> = ({
 }) => {
   const enableArtworkRedesign = useFeatureFlag("ARArtworkRedesingPhase2")
 
+  if (!!enableArtworkRedesign) {
+    return (
+      <Flex alignItems="center" accessibilityLabel="Countdown">
+        <Text variant="xs" weight="medium" color="black60">
+          {label}
+        </Text>
+        {!!extendedBiddingPeriodMinutes && (
+          <>
+            <Spacer mt={1} />
+            <Text variant="xs" color="black60" textAlign="center">
+              *Closure times may be extended to accommodate last-minute bids
+            </Text>
+          </>
+        )}
+      </Flex>
+    )
+  }
+
   return (
     <Flex alignItems="center" accessibilityLabel="Countdown">
       {cascadingEndTimeIntervalMinutes ? (
--- a/src/app/Components/Bidding/Components/Timer.tsx
+++ b/src/app/Components/Bidding/Components/Timer.tsx
@@ -164,21 +164,8 @@ export const Countdown: React.FC<CountdownProps> = ({
   const enableArtworkRedesign = useFeatureFlag("ARArtworkRedesingPhase2")
 
   if (!!enableArtworkRedesign) {
-    return (
-      <Flex alignItems="center" accessibilityLabel="Countdown">
-        <Text variant="xs" weight="medium" color="black60">
-          {label}
-        </Text>
-        {!!extendedBiddingPeriodMinutes && (
-          <>
-            <Spacer mt={1} />
-            <Text variant="xs" color="black60" textAlign="center">
-              *Closure times may be extended to accommodate last-minute bids
-            </Text>
-          </>
-        )}
-      </Flex>
-    )
+    // move this to the timer? up to the thing and return null ? ? ?
+    return null
   }
 
   return (
--- a/src/app/Components/Bidding/Components/Timer.tsx
+++ b/src/app/Components/Bidding/Components/Timer.tsx
@@ -164,7 +164,6 @@ export const Countdown: React.FC<CountdownProps> = ({
   const enableArtworkRedesign = useFeatureFlag("ARArtworkRedesingPhase2")
 
   if (!!enableArtworkRedesign) {
-    // move this to the timer? up to the thing and return null ? ? ?
     return null
   }
 
--- a/src/app/Scenes/Artwork/Components/ArtworkLotTimer.tsx
+++ b/src/app/Scenes/Artwork/Components/ArtworkLotTimer.tsx
@@ -179,7 +179,7 @@ const RenderCountdown: React.FC<AuctionWebsocketWrapperProps> = ({
   // display the timer and progress bar only when duration is positive and lot is biddable
   const shouldShowTimer = !!isBiddableInAuction && moment.duration(duration).milliseconds() > 0
 
-  const isBiddingClosed = !artwork.saleArtwork?.endedAt || timerState === AuctionTimerState.CLOSED
+  const isBiddingClosed = !!artwork.saleArtwork?.endedAt || timerState === AuctionTimerState.CLOSED
 
   return (
     <>
--- a/src/app/Scenes/Artwork/Components/ArtworkLotTimer.tsx
+++ b/src/app/Scenes/Artwork/Components/ArtworkLotTimer.tsx
@@ -1,0 +1,233 @@
+import { ArtworkLotTimer_artwork$data } from "__generated__/ArtworkLotTimer_artwork.graphql"
+import { ArtworkAuctionProgressBar } from "app/Components/Bidding/Components/ArtworkAuctionProgressBar"
+import {
+  AuctionTimerState,
+  currentTimerState,
+  nextTimerState,
+  relevantStateData,
+} from "app/Components/Bidding/Components/Timer"
+import { TimeOffsetProvider } from "app/Components/Bidding/Context/TimeOffsetProvider"
+import { StateManager as CountdownStateManager } from "app/Components/Countdown"
+import { CountdownTimerProps } from "app/Components/Countdown/CountdownTimer"
+import { ModernTicker, SimpleTicker } from "app/Components/Countdown/Ticker"
+import { getTimerInfo } from "app/utils/saleTime"
+import { Time } from "app/utils/useTimer"
+import { AuctionWebsocketContextProvider } from "app/Websockets/auctions/AuctionSocketContext"
+import { useArtworkBidding } from "app/Websockets/auctions/useArtworkBidding"
+import { Flex, Spacer, Text, TimerIcon } from "palette"
+import { Color } from "palette/Theme"
+import { createFragmentContainer, graphql } from "react-relay"
+import { TrackingProp } from "react-tracking"
+
+interface AuctionWebsocketWrapperProps extends CountdownTimerProps {
+  artwork: ArtworkLotTimer_artwork$data
+  hasStarted?: boolean
+  tracking?: TrackingProp
+  biddingEndAt?: string
+  hasBeenExtended?: boolean
+  refetchArtwork: () => void
+  setAuctionTimerState?: (auctionTimerState: string) => void
+}
+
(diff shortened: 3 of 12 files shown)


src/app/Components/Bidding/Components/Timer.tsx
```diff
@@ -162,6 +162,25 @@ export const Countdown: React.FC<CountdownProps> = ({
...
+    return (
+      <Flex alignItems="center" accessibilityLabel="Countdown">
+        <Text variant="xs" weight="medium" color="black60">
+          {label}
+        </Text>
+        {!!extendedBiddingPeriodMinutes && !!cascadingEndTimeFeatureEnabled && (
+          <>
+            <Spacer mt={1} />
+            <Text variant="xs" color="black60" textAlign="center">
+              *Closure times may be extended to accommodate last-minute bids
```
dimatretyak, line 177: I think we won't need to render this label here as it will be displayed in the lot details section.

for-george

src/app/Scenes/Artwork/Components/ArtworkTombstone.tsx
```diff
@@ -1,71 +1,62 @@
...
-@track()
-export class ArtworkTombstone extends React.Component<
-  ArtworkTombstoneProps,
-  ArtworkTombstoneState
-> {
-  state = {
-    showingMoreArtists: false,
-    showAuthenticityCertificateModal: false,
-  }
+export const ArtworkTombstone: React.FC<ArtworkTombstoneProps> = ({ artwork, refetchArtwork }) => {
```
dimatretyak, line 20: **suggestion**: what do you think about making these changes (refactor from class to functional component) in a separate PR?

src/app/Scenes/Artwork/Components/ArtworkTombstone.tsx
```diff
@@ -74,93 +65,90 @@ export class ArtworkTombstone extends React.Component<
...
+      )}
+      {!!isNewArtworkScreenEnabled && (
+        <ArtworkLotTimerFragmentContainer artwork={artwork} refetchArtwork={refetchArtwork} />
+      )}
+      <Flex flexDirection="row" flexWrap="wrap">
+        {artwork.artists?.length === 1
+          ? renderSingleArtist(artwork!.artists![0]!)
+          : renderMultipleArtists()}
+        {!!(artwork.artists?.length === 0 && artwork.culturalMaker) &&
+          renderArtistName(artwork.culturalMaker, null)}
```
dimatretyak, line 112: **question**: maybe we extract put this logic into a separate component?

src/app/Scenes/Artwork/Components/ArtworkLotTimer.tsx
```diff
@@ -0,0 +1,233 @@
...
+    } = props.artwork.sale || {}
+
+    const { endAt: lotEndAt, extendedBiddingEndAt, lotID } = props.artwork.saleArtwork
+
+    const initialBiddingEndAt = extendedBiddingEndAt ?? lotEndAt ?? saleEndAt
+
+    const { currentBiddingEndAt, lotSaleExtended } = useArtworkBidding({
+      lotID,
+      lotEndAt,
+      biddingEndAt: initialBiddingEndAt,
```
dimatretyak, line 65: **suggestion(not-related/not-blocking)**: the lot details section has [similar logic](https://github.com/artsy/eigen/blob/e501105ddb990ccc5096434d517899d2f94a9cef/src/app/Scenes/Artwork/Components/ArtworkLotDetails/LotEndDateTime.tsx#L16-L19). We will have to think about how we can reuse this (for example, store the single instance of `biddingEndAt`in react context)

src/app/Scenes/Artwork/Components/ArtworkTombstone.tsx
gkartalis: Good idea, opening a separate pr today

gkartalis: actually just realized that most of this refactor have been merged on this one https://github.com/artsy/eigen/pull/7551

gkartalis: so rebased and most of the refactor is gone now from the diff

src/app/Components/Bidding/Components/Timer.tsx
gkartalis: good point, removing it

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  We are trying to migrate away from moment towards `luxon`, but found moment imports in the following new files:

- `src/app/Scenes/Artwork/Components/ArtworkLotTimer.tsx`

See [docs](https://moment.github.io/luxon/api-docs/index.html).
  
  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against a4c256b8a906c85f2f18c278c91a01650bc48417
</p>

src/app/Components/Bidding/Components/ArtworkAuctionProgressBar.tsx
```diff
@@ -43,6 +45,10 @@ export const ArtworkAuctionProgressBar: React.FC<ArtworkAuctionProgressBarProps>
   const renderProgressBar = isWithinExtendedBiddingPeriod || hasBeenExtended
 
   return (
-    <>{renderProgressBar && <ProgressBar trackColor="red100" progress={percentComplete * 100} />}</>
+    <>
+      {renderProgressBar && (
+        <ProgressBar height={height} trackColor="red100" progress={percentComplete * 100} />
+      )}
+    </>
```
dimatretyak, line 52: **suggestion**:

```javascript
..

if (isWithinExtendedBiddingPeriod || hasBeenExtended) {
  return <ProgressBar height={height} trackColor="red100" progress={percentComplete * 100} />
}

return null
```

src/app/Scenes/Artwork/Components/ArtworkLotTimer.tsx
```diff
@@ -0,0 +1,247 @@
...
+    <AuctionWebsocketContextProvider
+      channelInfo={{
+        channel: "SalesChannel",
+        sale_id: props.artwork.sale?.internalID,
+      }}
+      enabled={websocketEnabled}
+    >
+      <ArtworkLotTimerWrapper {...props} />
+    </AuctionWebsocketContextProvider>
+  )
```
dimatretyak, line 45: **question**: as far as I can see `AuctionWebsocketContextProvider` is already used on the [Artwork page](https://github.com/artsy/eigen/blob/47703618e33251e869add2a377eb6d8ac4b73a35/src/app/Scenes/Artwork/Artwork.tsx#L442-L448). should we duplicate it again here?

src/app/Scenes/Artwork/Components/ArtworkLotTimer.tsx
```diff
@@ -0,0 +1,247 @@
...
+  artwork,
+  timerState,
+  duration,
+  hasStarted,
+  biddingEndAt,
+  hasBeenExtended,
+}) => {
+  const { isInAuction, sale, isForSale } = artwork
+  const isBiddableInAuction =
+    isInAuction && sale && timerState !== AuctionTimerState.CLOSED && isForSale
```
dimatretyak, line 139: **suggestion**: remove `isInAuction` since it's already checked [here](https://github.com/artsy/eigen/blob/47703618e33251e869add2a377eb6d8ac4b73a35/src/app/Scenes/Artwork/Components/ArtworkLotTimer.tsx#L49)

src/app/Scenes/Artwork/Components/ArtworkLotTimer.tsx
```diff
@@ -0,0 +1,247 @@
...
+
+  const displayAuctionLotLabel =
+    artwork.isInAuction &&
+    artwork.saleArtwork &&
```
dimatretyak, line 143: **suggestion**: we can remove `isInAuction`, `saleArtwork` since it's already checked [here](https://github.com/artsy/eigen/blob/47703618e33251e869add2a377eb6d8ac4b73a35/src/app/Scenes/Artwork/Components/ArtworkLotTimer.tsx#L49)

src/app/Scenes/Artwork/Components/ArtworkLotTimer.tsx
```diff
@@ -0,0 +1,247 @@
...
+      }}
+      enabled={websocketEnabled}
+    >
+      <ArtworkLotTimerWrapper {...props} />
+    </AuctionWebsocketContextProvider>
+  )
+}
+
+export const ArtworkLotTimerWrapper: React.FC<AuctionWebsocketWrapperProps> = (props) => {
+  if (props.artwork.isInAuction && props.artwork.saleArtwork) {
```
dimatretyak, line 49: **suggestion**: i think we can also check `props.artwork.sale` here since it used in `displayAuctionLotLabel`, `isBiddableInAuction`, `shouldShowTimer`

```javascript
const { isInAuction, saleArtwork, sale } = props.artwork ?? {}

if (isInAuction && saleArtwork && sale) {
  ...
}

const isBiddableInAuction = timerState !== AuctionTimerState.CLOSED && isForSale
const displayAuctionLotLabel = artwork.saleArtwork.lotLabel && !artwork.sale.isClosed
...
```

src/app/Scenes/Artwork/Components/ArtworkLotTimer.tsx
```diff
@@ -0,0 +1,247 @@
...
+          </>
+        )}
+      </Flex>
+      {!!shouldShowTimer && (
+        <>
+          <Flex alignItems="flex-end">
+            <Text variant="xs" color="black60">
+              {label}
+            </Text>
+          </Flex>
```
dimatretyak, line 201: **suggestion**: specify `textAlign` prop for `Text`.

```javascript
<Text variant="xs" color="black60" textAlign="right">
  {label}
</Text>
```

src/app/Scenes/Artwork/Components/ArtworkLotTimer.tsx
```diff
@@ -0,0 +1,247 @@
...
+          {...(props as any)}
+        />
+      </TimeOffsetProvider>
+    )
+  } else {
+    return null
+  }
+}
+
+const RenderCountdown: React.FC<AuctionWebsocketWrapperProps> = ({
```
dimatretyak, line 128: **suggestion (blocking)**: in `RenderCountdown` we have to update the `auctionTimerState` which stored in the `Artwork` component, as some components depend on it ([example 1](https://github.com/artsy/eigen/blob/47703618e33251e869add2a377eb6d8ac4b73a35/src/app/Scenes/Artwork/Artwork.tsx#L264), [example 2](https://github.com/artsy/eigen/blob/47703618e33251e869add2a377eb6d8ac4b73a35/src/app/Scenes/Artwork/Artwork.tsx#L137)). We are doing this now in [CommercialInformation](https://github.com/artsy/eigen/blob/47703618e33251e869add2a377eb6d8ac4b73a35/src/app/Scenes/Artwork/Components/CommercialInformation.tsx#L186-L190)

dimatretyak (changes requested):

src/app/Scenes/Artwork/Components/ArtworkLotTimer.tsx
```diff
@@ -0,0 +1,248 @@
...
+      enabled={websocketEnabled}
+    >
+      <ArtworkLotTimerWrapper {...props} />
+    </AuctionWebsocketContextProvider>
+  )
+}
+
+export const ArtworkLotTimerWrapper: React.FC<AuctionWebsocketWrapperProps> = (props) => {
+  const { artwork, refetchArtwork } = props
+  if (artwork.isInAuction && artwork.saleArtwork && artwork.sale) {
```
dimatretyak, line 51: **suggestion(non-blocking)**: 

```javascript
if (!(artwork.isInAuction && artwork.saleArtwork && artwork.sale)) {
  return null
}

return (...)
```

dimatretyak (approved):

dimatretyak:
> we never display [bidding closed](https://www.figma.com/file/9DMfKOi5RuuBD2f49eB9nV/Artwork-page?node-id=2759%3A27264&t=LYIucGe9JcQK7h12-0) message

oh, good catch

dimatretyak:
**question**: @gkartalis can we copy some test cases for timer/countdown from `CommercialInformation.tests.tsx` for the new `ArtworkLotTimer` component?

gkartalis:
@dimatretyak yes! Will do!

dimatretyak:
thank you!

src/app/Scenes/Artwork/Components/ArtworkLotTimer.tsx
```diff
@@ -0,0 +1,255 @@
...
+        {!!displayAuctionLotLabel && (
+          <Text variant="md" color="black100">
+            Lot {artwork.saleArtwork.lotLabel}
+          </Text>
+        )}
+        {!!isBiddingClosed && (
+          <Text variant="sm-display" textAlign="right">
+            Bidding closed
+          </Text>
+        )}
```
dimatretyak, line 184: **question**: should we show this label only when `timerState` is `AuctionTimerState.CLOSED` (as [we are doing](https://github.com/artsy/eigen/blob/e3a0a41427ffe502e196d059f1964341ed7278fd/src/app/Scenes/Artwork/Components/AuctionPrice.tsx#L51-L53) it now) since the end time can be [extended](https://github.com/artsy/eigen/blob/e3a0a41427ffe502e196d059f1964341ed7278fd/src/app/Scenes/Artwork/Components/ArtworkLotTimer.tsx#L65)?

gkartalis: I think using the `endedAt` is the best approach for this since it is null and it is changed to an actual value only when the lot is bidding closed 🤔

dimatretyak: as far as I understand, with this approach, the timer will continue to be displayed if the time has expired. For example:
* the user navigated to the page
* 5 seconds left in the timer
* the user has spent 10 seconds on the page
* the timer will continue to be displayed with 0d 0h 0m 0s

Or am I wrong?

gkartalis: Let me check 👌

dimatretyak: hmm, for [this artwork](https://staging.artsy.net/artwork/6389f9cb69e9c700060288da) i get the following result

for-george

dimatretyak: ok, the timer was hidden

https://user-images.githubusercontent.com/3513494/205301498-239638ec-3c53-4330-a06d-72f4782da102.mp4

dimatretyak: > I think using the endedAt is the best approach for this since it is null and it is changed to an actual value only when the lot is bidding closed

we have to consider `extendedBiddingEndAt`. in the attached screenshot, you can notice that the lot end time is later than the auction end time

for-george-2

dimatretyak: **suggestion**: 

```javascript
...

{shouldShowTimer ? (
  <>
    <Spacer mr={4} />
    <Flex flexDirection="row" alignItems="center">
      {sale?.cascadingEndTimeIntervalMinutes ? (
        <ModernTicker
          duration={duration}
          hasStarted={hasStarted}
          isExtended={hasBeenExtended}
          variant="md"
        />
      ) : (
        <SimpleTicker duration={duration} separator="  " variant="md" />
      )}
      <TimerIcon height={24} width={24} fill={getColor()} ml={1} />
    </Flex>
  </>
) : (
  <Text variant="sm-display" textAlign="right">
    Bidding closed
  </Text>
)}
```

gkartalis: I pushed one change, basically what you suggested but preventing showing the "bidding closed" message when we shouldn't. 

Let me know your thoughts. In the mean time adding some tests ;)

src/app/Scenes/Artwork/Components/ArtworkLotTimer.tsx
```diff
@@ -0,0 +1,265 @@
...
+
+    const timerInfo = getTimerInfo(time, { hasStarted, isExtended: hasBeenExtended })
+
+    return timerInfo.color as Color
+  }
+
+  // display the timer and progress bar only when duration is positive and lot is biddable
+  const shouldShowTimer = !!isBiddableInAuction && moment.duration(duration).milliseconds() > 0
+
+  const isBiddingClosed = !artwork.saleArtwork?.endedAt && timerState === AuctionTimerState.CLOSED
```
dimatretyak, line 182: **question**: as far as I understand we should display "bidding closed" when `endedAt` is **not equal** to `null`? If yes, then in this case there should be such a check

```javascript
const isBiddingClosed = !!artwork.saleArtwork?.endedAt || timerState === AuctionTimerState.CLOSED
```

dimatretyak: I think `||` should be used instead of `&&`

gkartalis: 😅 you are totally right!

dimatretyak:
**question(non-blocking)**: should we hide the timer when the live bidding state started? Now the timer continues to be displayed, but if you go back and navigate the artwork lot page again, the timer will no longer be displayed (this is due to the fact that `moment.duration(duration)` will return the same value until we reload the page [here](https://github.com/artsy/eigen/blob/2f3a39bb6579bb77be7a29543dde10c81f83a294/src/app/Scenes/Artwork/Components/ArtworkLotTimer.tsx#L140)

https://user-images.githubusercontent.com/3513494/205356336-0635ddbf-3b6a-448f-a668-f5d7fdb6cab7.mp4

gkartalis:
> question(non-blocking): should we hide the timer when the live bidding state started? Now the timer continues to be displayed, but if you go back and navigate the artwork lot page again, the timer will no longer be displayed (this is due to the fact that moment.duration(duration) will return the same value until we reload the page [here](https://github.com/artsy/eigen/blob/2f3a39bb6579bb77be7a29543dde10c81f83a294/src/app/Scenes/Artwork/Components/ArtworkLotTimer.tsx#L140). See the second attached video)

Let's leave it as is for now and lets discuss during QA

gkartalis:
@dimatretyak added some simple tests but having a hard time with the duration, even when I pass duration to the rendered component the duration is 0 in my component.

gkartalis:
I am thinking to merge this and fix the tests on a follow up pr to unblock the work, what do you think

dimatretyak (approved):
LGTM. Thank you!

PR #7584, opened 2022-11-03
feat: Initialize artwork title input when skipping artwork selection

olerichter00:
This PR resolves [CX-2728] 

### Description

Initialize artwork title input when skipping the artwork selection in the My Collection artwork flow

### PR Checklist

### To the reviewers 👀

- Initialize artwork title input when skipping artwork selection - ole

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-2728]: https://artsyproduct.atlassian.net/browse/CX-2728?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/MyCollection/Screens/ArtworkForm/Components/ArtworkAutosuggest.tsx
+++ b/src/app/Scenes/MyCollection/Screens/ArtworkForm/Components/ArtworkAutosuggest.tsx
@@ -1,4 +1,5 @@
 import SearchIcon from "app/Icons/SearchIcon"
+import { trim } from "lodash"
 import { Flex, Input, Text, Touchable } from "palette"
 import React, { useState } from "react"
 import { useArtworkForm } from "../Form/useArtworkForm"
@@ -6,7 +7,7 @@ import { ArtworkAutosuggestResultsQueryRenderer } from "./ArtworkAutosuggestResu
 
 interface ArtworkAutosuggestProps {
   onResultPress: (artworkId: string) => void
-  onSkipPress: () => void
+  onSkipPress: (artworkTitle?: string) => void
 }
 
 export const ArtworkAutosuggest: React.FC<ArtworkAutosuggestProps> = ({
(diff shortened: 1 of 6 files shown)


src/app/Scenes/MyCollection/Screens/ArtworkForm/Components/ArtworkAutosuggest.tsx
```diff
@@ -20,7 +21,7 @@ export const ArtworkAutosuggest: React.FC<ArtworkAutosuggestProps> = ({
   const artistSlug = artistSearchResult?.slug || ""
 
   const [artworkQuery, setArtworkQuery] = useState("")
-
+  const trimmedArtworkQuery = trim(artworkQuery)
```
olerichter00, line 24: Trimming the query to avoid whitespaces at the end and the beginning. Thanks @lordkiz for the idea 🙏

lordkiz: @olerichter00 I think you want to do only a `trimStart()` and trim the beginning only because whitespaces in the end of an artist or artwork can be valid and affect result.
In the case of an Artist for eg  a query "Andy" should return an artist named [Andy](https://www.artsy.net/artist/andy) while "Andy " can return Andy Warhol.

MounirDhahri (dismissed):
niice

src/app/Scenes/MyCollection/Screens/ArtworkForm/Components/ArtworkAutosuggest.tsx
olerichter00: Very good point! Although both search terms ("Andy" and "Andy  ") will almost return the same results, there is a slight difference in the search results and I can't tell if it is important. I changed the code to use `trimStart()` instead for the artist autosuggest component and the artwork autosuggest component.

PR #7606, opened 2022-11-07
fix(MOPLAT-490): Fix android unable to build issue

lordkiz:
This PR resolves [MOPLAT-490] 

### Description

- In summary: An update to RN (v0.71) caused lower versions to fetch 0.71 instead of their specified versions. Unfortunately our version still uses the gradle dynamic versioning pattern (i.e using `+` instead of specifying a specific version) so we ended up fetching v0.71. A patch for 0.66.* was released. This PR updates the version to that patch.

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- Fix android unable to build issue - kizito

</details>

[MOPLAT-490]: https://artsyproduct.atlassian.net/browse/MOPLAT-490?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/patches/react-native+0.66.5.patch
+++ b/patches/react-native+0.66.5.patch
@@ -56,7 +56,7 @@ index b4dab8e..2195f8e 100644
  #import "RCTSurfaceView.h"
  #import "RCTTouchEvent.h"
  #import "RCTUIManager.h"
-@@ -112,6 +113,16 @@ RCT_NOT_IMPLEMENTED(-(instancetype)initWithTarget : (id)target action : (SEL)act
+@@ -112,6 +113,16 @@ - (void)_recordNewTouches:(NSSet<UITouch *> *)touches
        continue;
      }
  
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -64,14 +64,14 @@ PODS:
     - Specta (~> 1.0)
   - FBAEMKit (15.0.0):
     - FBSDKCoreKit_Basics (= 15.0.0)
-  - FBLazyVector (0.66.4)
-  - FBReactNativeSpec (0.66.4):
+  - FBLazyVector (0.66.5)
+  - FBReactNativeSpec (0.66.5):
     - RCT-Folly (= 2021.06.28.00-v2)
-    - RCTRequired (= 0.66.4)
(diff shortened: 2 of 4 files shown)


patches/react-native+0.66.5.patch
```diff
@@ -56,7 +56,7 @@ index b4dab8e..2195f8e 100644
  #import "RCTSurfaceView.h"
  #import "RCTTouchEvent.h"
  #import "RCTUIManager.h"
-@@ -112,6 +113,16 @@ RCT_NOT_IMPLEMENTED(-(instancetype)initWithTarget : (id)target action : (SEL)act
+@@ -112,6 +113,16 @@ - (void)_recordNewTouches:(NSSet<UITouch *> *)touches
```
brainbicycle, line 59: this change to the patch looks a little funky at first glance, was this just reapplying after the update?

lordkiz: @brainbicycle the touch related changes seems to be specific to `0.66.5`. I was initially struggling with the first version of the generated patch as it was 2500+ lines long while the original patch we had was only 158 lines long. I had to delete node_modules, pods, clean the build folder and terminate all processes before I could generate this which has all the patch from 0.66.4 + a smaller patch related to UITouch. 
I am not sure where the UITouch part is coming from, but I reckon it is specific for 0.66.5.
Can you pull this branch, delete this patch and try to generate a new one after clearing your node modules etc and see if you get the same thing?

lordkiz:
Do we need Gradle clean and clean build folder on CI for the `test-app-android` and `test-app-ios` steps to pass?

patches/react-native+0.66.5.patch
MounirDhahri: we probably have a description for this fix within the `Hacks.md` if so we can try out if everything is fine then maybe no action is needed

MounirDhahri (approved):
Thanks for fixing this Kizito 🙏

lordkiz:
Works fine at my ends. Will merge. We can always revert if folks find issues.

PR #7611, opened 2022-11-08
fix: prevent markdown lists from breaking the readmore component

gkartalis:
This PR resolves [] 

### Description

[Slack thread](https://artsy.slack.com/archives/C02BAQ5K7/p1667866746452839) for more context.

The ReadMore component broke the app when it was trying to render lists due to a problem with the markdown dependency that we are using (opened an [issue](https://github.com/ariabuckles/simple-markdown/issues/112) a while ago but no one has replied..)

The solution is to prevent the component from rendering the markdown "list" (when it sees dashes it tries to render a markdown list but fails) and default to rendering it as a paragraph which is okay for now (see screenshot below)

simulator_screenshot_4131C009-2056-478E-A025-BFC45923DFD0

### PR Checklist

### To the reviewers 👀

- fixed markdown lists from breaking readmore component - gkartalis

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Components/ReadMore.tsx
+++ b/src/app/Components/ReadMore.tsx
@@ -27,7 +27,6 @@ interface Props {
   color?: ResponsiveValue<Color>
   textStyle?: "sans" | "new"
   testID?: string
-  type?: "show"
   textVariant?: PaletteTextProps["variant"]
   linkTextVariant?: PaletteTextProps["variant"]
 }
@@ -42,7 +41,6 @@ export const ReadMore = React.memo(
     contextModule,
     textStyle = "sans",
     testID,
-    type,
     textVariant = "xs",
     linkTextVariant = "xs",
   }: Props) => {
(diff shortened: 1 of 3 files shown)


src/app/Components/ReadMore.tsx
```diff
@@ -51,27 +49,24 @@ export const ReadMore = React.memo(
...
-            state: SimpleMarkdown.State
-          ) => {
-            return (
-              <TextComponent {...textProps} color={color || "black100"} key={state.key}>
-                {!isExpanded && Number(state.key) > 0 ? ` ${emdash} ` : null}
-                {output(node.content, state)}
-              </TextComponent>
-            )
-          },
+      list: {
```
gkartalis, line 55: we were doing this before only for shows, but removed it and defaulted it to all the lists

jo-rs: Do you remember why it was limited to only shows initially? 🤔 
(Just out of curiosity)

gkartalis: I was the one that did the fix back then, I think the only reason was that this was encountered only in shows but now I think it's better to have it on all surfaces.

gkartalis:
@jo-rs added a test case 😆 let me know what you think

jo-rs (approved):
I appreciate you! 💜

PR #7620, opened 2022-11-09
docs(troubleshooting.md): Add pod repo update item 

damassi:
### Description

This updates the troubleshooting doc with the step that helped me finally install Eigen:

```
bundle exec pod repo update 
```

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -6,7 +6,7 @@ If things are not going right some of the things you can try are
 - `yarn setup:artsy`
 - `yarn install:all`
 - `yarn relay`
-- `cd ios && bundle exec pod repo update && bundle exec pod install`
+- `yarn pod-install-repo-update`
 - `open ios/Artsy.xcworkspace` and build the app again
 - `yarn doctor`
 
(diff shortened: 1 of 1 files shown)


pvinis (changes requested):
good point. maybe replace it with https://github.com/artsy/eigen/blob/main/package.json#L33 since we have a script for this?

docs/troubleshooting.md
```diff
@@ -6,6 +6,7 @@ If things are not going right some of the things you can try are
 - `yarn setup:artsy`
 - `yarn install:all`
 - `yarn relay`
+- `cd ios && bundle exec pod repo update && bundle exec pod install`
```
MounirDhahri, line 9: **suggestion:** I think you can replace this with
```suggestion
- `yarn pod-install-repo-update`
```

MounirDhahri: @pvinis @brainbicycle  should we maybe use `yarn pod-install-repo-update` https://github.com/artsy/eigen/blob/main/scripts/install#L8 and then we will no longer need this change?

PR #7625, opened 2022-11-09
fix(setup): Invoke prettier via npx to avoid error during repo setup

damassi:
### Description

Following the install docs, after running `yarn setup:artsy` the process crashed due to prettier not being installed globally on my system: 

Screenshot 2022-11-09 at 9 14 35 AM

This PR swaps the `yarn` invocation with `npx`, which is designed for these kinds of cases where packages are used outside of the normal package.json flow.


pvinis (approved):
ah good catch. most of us already have it from `yarn install` I guess, but this runs before that on first setup, so npx is a good way.

PR #7626, opened 2022-11-09
chore(podfile): update lock

damassi:
### Description

When running setup steps on a new install, noticed that the `Podfile.lock` was out of date.


brainbicycle:
Ahh this looks like the old churn we used to get because cocoa pods changed the url for the main spec repo. I think you will want to do something along these lines:
https://www.willowtreeapps.com/craft/how-to-fix-unexpected-podfile-lock-diffs-after-pod-install
tl;dr
`pod repo list` 
`# if you have a repo called 'master' or similar`
`pod repo remove master`

damassi:
Yup that fixed it @brainbicycle 👍

PR #7636, opened 2022-11-11
WIP: Add Video support to Artwork Carousel 

damassi:
Super WIP around types, but works! 

Couple things. 
- Tried to refactor this to "do things correctly" in a separate PR -- that is, replace `images` field on Artwork with `figures`, and then narrow the type. This led to a cascade of failures, in addition to the need to refactor the implementation as the ImageCarousel is used in a few different ways around eigen. Aborted that mission.
- The approach in this PR keeps the existing code, and adds a new fragment on the carousel to include `figures` which exclusively extracts video code. Then we shuttle that data along and pass it into the carousel and merge the two together. 
- Works well enough, but ideally everything is refactored to use one consistent form -- `figures`. Ran into type errors doing this, even though the change _should_ have been straight forward and 1:1

### DIFF
--- a/package.json
+++ b/package.json
@@ -179,6 +179,7 @@
     "react-native-share": "7.3.6",
     "react-native-svg": "9.13.3",
     "react-native-view-shot": "3.4.0",
+    "react-native-vimeo-iframe": "1.2.0",
     "react-native-vision-camera": "2.14.1",
     "react-native-webview": "11.3.1",
     "react-relay": "14.0.0",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -179,6 +179,7 @@
     "react-native-share": "7.3.6",
     "react-native-svg": "9.13.3",
     "react-native-view-shot": "3.4.0",
+    "react-native-vimeo-iframe": "1.2.0",
```
pvinis, line 182: did you consider https://www.npmjs.com/package/react-native-video instead? if yes, why did you decide on this. if no, maybe take a look?

damassi: I was actually just going to copy the contents of this lib into eigen. Its a tiny wrapper ontop of react-native-web-view: https://github.com/MetaLabs-inc/react-native-vimeo-iframe/blob/main/src/index.tsx

PR #7664, opened 2022-11-16
feat: keep search filter active when updating query text

dblandin:
Jira: [FX-4457](https://artsyproduct.atlassian.net/browse/FX-4457)

### Description

In the initial implementation of the improved search experience including filter pills, we borrowed the behavior from Spotify's mobile app which is to reset to the Top pill (which aggregates many different result types together) whenever the query text changes.

We've received feedback that this is quite annoying behavior. It's common to start a query, choose a more specific filter pill (like artists) and adjust your query until you see the result you want.

This PR removes the reset behavior altogether.

### PR Checklist

### To the reviewers 👀

- Keep selected search filter active when updating query text

</details>


src/app/Scenes/Search/Search.tests.tsx
```diff
@@ -565,16 +540,6 @@ describe("Search Screen", () => {
...
     })
 
-    it("when the query is changed", () => {
-      const searchInput = screen.getByPlaceholderText("Search artists, artworks, galleries, etc")
-
-      fireEvent(searchInput, "changeText", "12")
-      fireEvent(screen.getByText("Artists"), "press")
-      fireEvent(searchInput, "changeText", "123")
-
-      expect(screen.queryByA11yState({ selected: true })).toHaveTextContent("Top")
```
dimatretyak, line 575: **suggestion**: maybe we can just update test case? Something like this

```
expect(screen.queryByA11yState({ selected: true })).toHaveTextContent("Artists")
```

dimatretyak:
**question**: do you think it would be a bad idea not to disable pills if there are no search results?


https://user-images.githubusercontent.com/3513494/206152850-1cd73b24-194b-4830-9718-ea652afb47ef.mp4

dimatretyak (dismissed):

src/app/Scenes/Search/Search.tests.tsx
dblandin: Yep, sounds good. Updated 👍🏼

dblandin:
> question: do you think it would be a bad idea not to disable pills if there are no search results?

I still like the feedback offered if you stay on the Top pill while you're editing your search query. It makes it pretty clear which filters have content and we added that disable behavior to reduce the chance of hitting an empty state.

If a user is already certain they're looking for a particular type of result, they might tap on a filter pill earlier in the flow. I wish we had some better discovery pathways available on that empty state screen but for now it feels intuitive enough.

PR #7703, opened 2022-11-21
docs: testing docs 🎉

patrinoua:
This PR resolves [] 

### Description

More information about how our tests work, wrappers and preferred patterns.

co-author: @pvinis

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -31,7 +31,7 @@ New fashion: You can use useLazyLoadQuery inside a TestRendered which is the new
 
 New fashion Example: [ArtworkConsignments.tests.tsx](https://github.com/artsy/eigen/blob/6360ddb5304e6c5bb1dc207db13d5bf0a5d7d9b6/src/app/Scenes/Artwork/Components/ArtworkConsignments.tests.tsx#L28-L38)
 
-Old fashioned: You can use a TestRendered to wrap your component with a QueryRenderer and pass it the test query data.
+Old fashioned: You can use a TestRenderer to wrap your component with a QueryRenderer and pass the test query data.
 Old fashion Example: [Inbox.tests.tsx](https://github.com/artsy/eigen/blob/c96dd0807555d69ca2e8655dc68085276d249080/src/app/Containers/Inbox.tests.tsx)
 
 - ⭐️ [`renderWithWrappers`](https://github.com/artsy/eigen/blob/main/src/app/tests/renderWithWrappers.tsx#L88-L103) Our default method for using a component. Can be used on all components. wraps the component with a number of wrappers such as TrackingProvider, GlobalStoreProvider, SafeAreaProvider, etc. Using `testing library`. See the list of wrappers [here](https://github.com/artsy/eigen/blob/main/src/app/tests/renderWithWrappers.tsx#L19-L25).
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -50,7 +50,7 @@ This is an abstraction that is sometimes convenient.
 
 Example of a component with one request [PartnerLocationSection.tests.tsx](https://github.com/artsy/eigen/blob/7703aa01103a06b69b650af11bc9903ab37b1c4b/src/app/Scenes/Partner/Components/PartnerLocationSection.tests.tsx#L21).
 
-`setupTestWrapperTL` can be used to replace the combination of **TestRendered** and **renderWithWrappers**.
+`setupTestWrapperTL` can be used to replace the combination of **TestRenderer** and **renderWithWrappers**.
 
 It returns the function `renderWithRelay` which is similar to `renderWithWrappers`. Both `renderWithRelay` and `renderWithWrappers` use [render](https://testing-library.com/docs/angular-testing-library/api/#render) under the hood. I
 
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -21,7 +21,7 @@ The "new fashion" testing uses > relay v13 and we now prefer to use hooks whenev
 
 - We don’t use snapshot tests; they produce too much churn for too little value.
 
-Our preferred methods to use are marked with a ⭐️, while the ones we'd like to avoid are marked as ❗️ depracated.
+Our preferred methods to use are marked with a ⭐️, while the ones we'd like to avoid are marked as ❗️ deprecated.
 
 ## Test helpers and wrappers, when to use them, and good examples
 
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -25,7 +25,7 @@ Our preferred methods to use are marked with a ⭐️, while the ones we'd like 
 
 ## Test helpers and wrappers, when to use them, and good examples
 
-- `TestRendered` is a component we create, using the component we want to test. We use a TestRendered when for example we want to create a mock query. We don't always need to use a test renderer.
+- `TestRenderer` is a component we create, using the component we want to test. We use a TestRenderer when for example we want to create a mock query. We don't always need to use a test renderer.
 
 New fashion: You can use useLazyLoadQuery inside a TestRendered which is the new relay hooks way for querying data and the way we generally want to swift towards.
 
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -54,7 +54,7 @@ Example of a component with one request [PartnerLocationSection.tests.tsx](https
 
 It returns the function `renderWithRelay` which is similar to `renderWithWrappers`. Both `renderWithRelay` and `renderWithWrappers` use [render](https://testing-library.com/docs/angular-testing-library/api/#render) under the hood. I
 
-t gives you access to all the [queries](https://callstack.github.io/react-native-testing-library/docs/api-queries) such as getBy, getAllBy, queryBy, queryAllBy, findBy, findAllBy, etc.
+It gives you access to all the [queries](https://callstack.github.io/react-native-testing-library/docs/api-queries) such as getBy, getAllBy, queryBy, queryAllBy, findBy, findAllBy, etc.
 
 ❗️ It won't work if there is already a component using a query renderer. [eg?]
 
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -7,7 +7,7 @@ So you want to write tests!
 
 ## Test helpers and wrappers, when to use them, and good examples
 
-- `extractTest` takes a component and returns any string that exists there, whether it's on a title, View or button. You can also check `extractTest.tests.tsx`.
+- `extractText` takes a component and returns any string that exists there, whether it's on a title, View or button. You can also check `extractTest.tests.tsx`.
 
 - `renderWithWrappers` Can be used on all components. Our default method for using a component. wraps the component with a number of wrappers such as TrackingProvider, GlobalStoreProvider, SafeAreaProvider, etc. Using `testing library`. See the list of wrappers [here](https://github.com/artsy/eigen/blob/main/src/app/tests/renderWithWrappers.tsx#L19-L25).
 
@@ -23,7 +23,7 @@ So you want to write tests!
 
 - `resolveMostRecentRelayOperation` resolves the query request. We always put it after rendering a component that has relay requests. Your rendered component makes a request and we use this function in tests to resolve it. Example file: [Inbox.tests.tsx](https://github.com/artsy/eigen/blob/c96dd0807555d69ca2e8655dc68085276d249080/src/app/Containers/Inbox.tests.tsx)
 
-- `rejectMostRecentRelayOperation` for example if network is off / bad / you have a server error. Example: [FollowArtistLink.tests.tsx](**add link here**)
+- `rejectMostRecentRelayOperation` for example if network is off / bad / you have a server error. Example: [FollowArtistLink.tests.tsx](https://github.com/artsy/eigen/blob/cfcdd1429732ea04dc26134e1bf4a4d4cb96f16e/src/app/Scenes/Artwork/Components/FollowArtistLink.tests.tsx)
 
 - `flushPromiseQueue` This is a hack - try to avoid it if possible.
 
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -1,7 +1,67 @@
-Please also check these best practices 
-https://github.com/artsy/eigen/blob/main/docs/best_practices.md#testing
+# Testing
 
-# Testing with Relay [WIP]
+So you want to write tests!
+
+- We use `@testing-library/react-native` and our helper `renderWithWrappers`. In the past we used ReactTestRenderer which we try to now move away from.
+- We refer to this guide 👉 [How should I query?](https://callstack.github.io/react-native-testing-library/docs/how-should-i-query/) from `@testing-library/react-native` for querying components
+
+## Test helpers and wrappers, when to use them, and good examples
+
+- `extractTest` takes a component and returns any string that exists there, whether it's on a title, View or button. You can also check `extractTest.tests.tsx`.
+
+- `renderWithWrappers` Can be used on all components. Our default method for using a component. wraps the component with a number of wrappers such as TrackingProvider, GlobalStoreProvider, SafeAreaProvider, etc. Using `testing library`.
+
+- `renderWithWrappersLEGACY`depracated - avoid this ❗️ uses ReactTestRenderer
+
+- `setupTestWrapperTL` For components that use relay requests. Uses react-native testing library. Renders a test component and resolves the most recent operation. An abstraction that sometimes is convenient. Example [PartnerLocationSection.tests.tsx](https://github.com/artsy/eigen/blob/7703aa01103a06b69b650af11bc9903ab37b1c4b/src/app/Scenes/Partner/Components/PartnerLocationSection.tests.tsx#L21). This function can be used instead of TestRendered and renderWithWrappers. Returns the function `renderWithRelay` which is similar to `renderWithWrappers` and gives you access to a number of properties such as `findWithTestID` etc. It won't work if there is already a component using a query renderer.
+  Better to explicitly use a QueryRenderer to make obvious what is there and minimize complexity.
+
+- `setupTestWrapper` depracated - avoid this ❗️uses ReactTestRenderer . Renders a test component and resolves the most recent operation. An abstraction that sometimes is convenient.
+
+// TODO: rename setupTestWrapperTL to setupTestWrapper and setupTestWrapper to setupTestWrapperLegacy
+// Decide if we want to use this pattern or not.
+
+- `resolveMostRecentRelayOperation` resolves the query request. We always put it after rendering a component that has relay requests. Your rendered component makes a request and we use this function in tests to resolve it. Example file: [Inbox.tests.tsx](https://github.com/artsy/eigen/blob/c96dd0807555d69ca2e8655dc68085276d249080/src/app/Containers/Inbox.tests.tsx)
+
+- `rejectMostRecentRelayOperation` for example if network is off / bad / you have a server error. Example: [FollowArtistLink.tests.tsx](**add link here**)
+
+- `flushPromiseQueue` This is a hack - try to avoid it if possible.
+
+flushPromiseQueue will cancel all promise operations.
+
+This is usually called when promises have been used in a bad way, eg you return the promise instead of resolving it.
+
+Try to make a test run normally and if you are absolutely sure everything else is correct but the test still fails after a promise has been called (eg `resolveMostRecentRelayOperation` or `rejectMostRecentRelayOperation`) try adding this.
+
+// TODO: remove flushPromiseQueue from everywhere [WIP]
+
+- `fetchMockResponseOnce` - do we really need this? What is it used for? 😇
+
+- `mockTrackEvent` - mock data for tracking. We don't actually send this data to our data tracking provider for called analytics function `trackEvent`.
+
+- `mockPostEventToProviders` expects a user behavior (eg button click) and mocks analytics for called analytics function `postEventToProviders`
+
+- `mockFetchNotificationPermissions` mocking for the function `fetchNotificationPermissions`
+
+- `mockTimezone` mocking for time/date eg with moment or luxon. Used for testing the date format.
+
+- `mockNavigate` we test if navigation was called and with which parameters.
+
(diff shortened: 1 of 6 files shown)


docs/testing.md
```diff
@@ -1,7 +1,67 @@
...
 
-# Testing with Relay [WIP]
+So you want to write tests!
+
+- We use `@testing-library/react-native` and our helper `renderWithWrappers`. In the past we used ReactTestRenderer which we try to now move away from.
+- We refer to this guide 👉 [How should I query?](https://callstack.github.io/react-native-testing-library/docs/how-should-i-query/) from `@testing-library/react-native` for querying components
+
+## Test helpers and wrappers, when to use them, and good examples
+
+- `extractTest` takes a component and returns any string that exists there, whether it's on a title, View or button. You can also check `extractTest.tests.tsx`.
```
damassi, line 10: (#minor) It might be worth ordering these functions by most important / most used to least

damassi (approved):
Super helpful 💯

docs/testing.md
```diff
@@ -1,7 +1,67 @@
...
+  Better to explicitly use a QueryRenderer to make obvious what is there and minimize complexity.
+
+- `setupTestWrapper` depracated - avoid this ❗️uses ReactTestRenderer . Renders a test component and resolves the most recent operation. An abstraction that sometimes is convenient.
+
+// TODO: rename setupTestWrapperTL to setupTestWrapper and setupTestWrapper to setupTestWrapperLegacy
+// Decide if we want to use this pattern or not.
+
+- `resolveMostRecentRelayOperation` resolves the query request. We always put it after rendering a component that has relay requests. Your rendered component makes a request and we use this function in tests to resolve it. Example file: [Inbox.tests.tsx](https://github.com/artsy/eigen/blob/c96dd0807555d69ca2e8655dc68085276d249080/src/app/Containers/Inbox.tests.tsx)
+
+- `rejectMostRecentRelayOperation` for example if network is off / bad / you have a server error. Example: [FollowArtistLink.tests.tsx](**add link here**)
```
araujobarret, line 26: I see you linked to the examples, thanks for doing it!

araujobarret (approved):
Awesome!

docs/testing.md
```diff
@@ -1,7 +1,67 @@
...
+So you want to write tests!
+
+- We use `@testing-library/react-native` and our helper `renderWithWrappers`. In the past we used ReactTestRenderer which we try to now move away from.
+- We refer to this guide 👉 [How should I query?](https://callstack.github.io/react-native-testing-library/docs/how-should-i-query/) from `@testing-library/react-native` for querying components
+
+## Test helpers and wrappers, when to use them, and good examples
+
+- `extractTest` takes a component and returns any string that exists there, whether it's on a title, View or button. You can also check `extractTest.tests.tsx`.
+
+- `renderWithWrappers` Can be used on all components. Our default method for using a component. wraps the component with a number of wrappers such as TrackingProvider, GlobalStoreProvider, SafeAreaProvider, etc. Using `testing library`.
```
MounirDhahri, line 12: **suggestion:** maybe we can link to the wrappers here https://github.com/artsy/eigen/blob/main/src/app/tests/renderWithWrappers.tsx#L18

docs/testing.md
```diff
@@ -1,7 +1,67 @@
...
+
+- `renderWithWrappersLEGACY`depracated - avoid this ❗️ uses ReactTestRenderer
+
+- `setupTestWrapperTL` For components that use relay requests. Uses react-native testing library. Renders a test component and resolves the most recent operation. An abstraction that sometimes is convenient. Example [PartnerLocationSection.tests.tsx](https://github.com/artsy/eigen/blob/7703aa01103a06b69b650af11bc9903ab37b1c4b/src/app/Scenes/Partner/Components/PartnerLocationSection.tests.tsx#L21). This function can be used instead of TestRendered and renderWithWrappers. Returns the function `renderWithRelay` which is similar to `renderWithWrappers` and gives you access to a number of properties such as `findWithTestID` etc. It won't work if there is already a component using a query renderer.
```
MounirDhahri, line 16: **issue:**

> Renders a test component and resolves the most recent operation.

I am not sure we do the second part here related to resolving the most recent operation. We do that separately I believe

docs/testing.md
```diff
@@ -1,7 +1,67 @@
...
+
+flushPromiseQueue will cancel all promise operations.
+
+This is usually called when promises have been used in a bad way, eg you return the promise instead of resolving it.
+
+Try to make a test run normally and if you are absolutely sure everything else is correct but the test still fails after a promise has been called (eg `resolveMostRecentRelayOperation` or `rejectMostRecentRelayOperation`) try adding this.
+
+// TODO: remove flushPromiseQueue from everywhere [WIP]
+
+- `fetchMockResponseOnce` - do we really need this? What is it used for? 😇
```
MounirDhahri, line 38: **issue:** maybe we can remove this from here if it's not clear

docs/testing.md
```diff
@@ -1,7 +1,67 @@
...
+
+- `resolveMostRecentRelayOperation` resolves the query request. We always put it after rendering a component that has relay requests. Your rendered component makes a request and we use this function in tests to resolve it. Example file: [Inbox.tests.tsx](https://github.com/artsy/eigen/blob/c96dd0807555d69ca2e8655dc68085276d249080/src/app/Containers/Inbox.tests.tsx)
+
+- `rejectMostRecentRelayOperation` for example if network is off / bad / you have a server error. Example: [FollowArtistLink.tests.tsx](**add link here**)
+
+- `flushPromiseQueue` This is a hack - try to avoid it if possible.
+
+flushPromiseQueue will cancel all promise operations.
+
+This is usually called when promises have been used in a bad way, eg you return the promise instead of resolving it.
```
MounirDhahri, line 32: **issue:** I don't agree with the description of the `flushPromiseQueue` here. I think the method simply waits for a frame or 1/60 of a sec (depending on the js environment configuration where it's being called)

MounirDhahri (commented):
Thanks for adding these docs @patrinoua . Great to see you again contributing here. I left few small comments that I would be happy to explain further if something wasn't clear

docs/testing.md
patrinoua: That's what I tried to do, without being very strict about it, ordering them in order of how often do I see something being used / how important I think they are. But I'm happy to update the order if you have another suggestion 🙏 😊

docs/testing.md
patrinoua: I would love it if anyone that has more context would add comments or make a commit to this PR so we can have more information about this function 🙏 This is why I left this here.

docs/testing.md
patrinoua: Thank you Mounir. Would you please rephrase this with a commit or a comment here in a way you think makes more sense?

docs/testing.md
patrinoua: Thank you Mounir. Would you please rephrase this with a commit or a comment here in a way you think makes more sense?

docs/testing.md
gkartalis, line 10: I think the function here and the file name has a typo, its `extractText` not `extractTest`.

Also I would suggest to remove this function (we use it in two places in eigen).

It works but it was used along with enzyme in order to extract the text from the elements. Now that we use `@testing-library/react-native` there is no need for that anymore since the library itself has the ability to query for text with [byText](https://callstack.github.io/react-native-testing-library/docs/api-queries#bytext)

docs/testing.md
gkartalis: setupTestWrapperTL is great if you don't want access to the relay environment for any request other than the first one. What it does is that it renders the component and resolves the **First** request.

docs/testing.md
```diff
@@ -1,7 +1,67 @@
...
+
+- `renderWithWrappers` Can be used on all components. Our default method for using a component. wraps the component with a number of wrappers such as TrackingProvider, GlobalStoreProvider, SafeAreaProvider, etc. Using `testing library`. See the list of wrappers [here](https://github.com/artsy/eigen/blob/main/src/app/tests/renderWithWrappers.tsx#L19-L25).
+
+- `renderWithWrappersLEGACY`depracated - avoid this ❗️ uses ReactTestRenderer
+
+- `setupTestWrapperTL` For components that use relay requests. Uses react-native testing library. Renders a test component and resolves the most recent operation. An abstraction that sometimes is convenient. Example [PartnerLocationSection.tests.tsx](https://github.com/artsy/eigen/blob/7703aa01103a06b69b650af11bc9903ab37b1c4b/src/app/Scenes/Partner/Components/PartnerLocationSection.tests.tsx#L21). This function can be used instead of TestRendered and renderWithWrappers. Returns the function `renderWithRelay` which is similar to `renderWithWrappers` and gives you access to a number of properties such as `findWithTestID` etc. It won't work if there is already a component using a query renderer.
```
gkartalis, line 16:
```suggestion
- `setupTestWrapperTL` For components that use relay requests. Uses react-native testing library. Renders a test component and resolves the most recent operation. An abstraction that sometimes is convenient. Example [PartnerLocationSection.tests.tsx](https://github.com/artsy/eigen/blob/7703aa01103a06b69b650af11bc9903ab37b1c4b/src/app/Scenes/Partner/Components/PartnerLocationSection.tests.tsx#L21). This function can be used instead of TestRendered and renderWithWrappers. Returns the function `renderWithRelay` which is similar to `renderWithWrappers` (both use render from @testing-library/react-native under the hood) and gives you access to all the [queries](https://callstack.github.io/react-native-testing-library/docs/api-queries) coming from it. It won't work if there is already a component using a query renderer.
```

gkartalis: let me know if you don't agree with the suggested change

docs/testing.md
```diff
@@ -1,7 +1,67 @@
...
+  Better to explicitly use a QueryRenderer to make obvious what is there and minimize complexity.
+
+- `setupTestWrapper` depracated - avoid this ❗️uses ReactTestRenderer . Renders a test component and resolves the most recent operation. An abstraction that sometimes is convenient.
+
+// TODO: rename setupTestWrapperTL to setupTestWrapper and setupTestWrapper to setupTestWrapperLegacy
+// Decide if we want to use this pattern or not.
+
+- `resolveMostRecentRelayOperation` resolves the query request. We always put it after rendering a component that has relay requests. Your rendered component makes a request and we use this function in tests to resolve it. Example file: [Inbox.tests.tsx](https://github.com/artsy/eigen/blob/c96dd0807555d69ca2e8655dc68085276d249080/src/app/Containers/Inbox.tests.tsx)
```
gkartalis, line 24: **suggestion(non-blocking):** up to you
```suggestion
- `resolveMostRecentRelayOperation` resolves the query request. We always use it after rendering a component that has relay requests. Your rendered component makes a request and we use this function in tests to resolve it. Example file: [Inbox.tests.tsx](https://github.com/artsy/eigen/blob/c96dd0807555d69ca2e8655dc68085276d249080/src/app/Containers/Inbox.tests.tsx)

> ℹ️ Note: There might be some cases that you would want to use `flushPromiseQueue` after that in order to wait for the request to be resolved. Use it only if after resolving the request returns the suspense fallback component and not the component. 
```

docs/testing.md
```diff
@@ -1,7 +1,67 @@
...
+
+- `rejectMostRecentRelayOperation` for example if network is off / bad / you have a server error. Example: [FollowArtistLink.tests.tsx](**add link here**)
+
+- `flushPromiseQueue` This is a hack - try to avoid it if possible.
+
+flushPromiseQueue will cancel all promise operations.
```
gkartalis, line 30: **suggestion:** 
```suggestion
flushPromiseQueue will resolve all promise operations.
```

docs/testing.md
gkartalis: I think we can just add the reference to the third party [repo](https://github.com/jefflau/jest-fetch-mock#jest-fetch-mock) for this one since it is not exactly our helper.

docs/testing.md
```diff
@@ -1,7 +1,67 @@
...
+
+- `mockPostEventToProviders` expects a user behavior (eg button click) and mocks analytics for called analytics function `postEventToProviders`
+
+- `mockFetchNotificationPermissions` mocking for the function `fetchNotificationPermissions`
+
+- `mockTimezone` mocking for time/date eg with moment or luxon. Used for testing the date format.
+
+- `mockNavigate` we test if navigation was called and with which parameters.
+
+- `waitUntil` Waits until something happens. RN Testing library has a similar component [waitFor](https://testing-library.com/docs/dom-testing-library/api-async/), that we could potentially replace this with.
```
gkartalis, line 50: I would remove this completely since it is a helper that is legacy

gkartalis (approved):
Great job! Thanks for adding those, left some suggestions on some cases, happy to hear your thoughts! 

Also note that hopefully one day we will update also the relay testing infra and we will need to update some of the above with fresher examples! 

🔥🥇🥳

docs/testing.md
patrinoua: Done, thank you for the suggestion! 🙏

docs/testing.md
pvinis: the note is kinda confusing, and i worry people might just try to add this without knowing why, or worse, when its not needed 😬.

docs/testing.md
pvinis: maybe then we should remove `renderWithWrappersLEGACY`? thought its used in a bunch of places still 🤔

docs/testing.md
gkartalis: happy to skip this

docs/testing.md
gkartalis: + yep, agree

docs/testing.md
patrinoua: I've added this, thank you!

docs/testing.md
patrinoua: I've added this, thank you!

docs/testing.md
patrinoua: I've added this under flushPromiseQueue docs, after waring the devs not to use it. Thank you!

docs/testing.md
patrinoua: I've added that we want to remove this, thank you!

docs/testing.md
patrinoua: done, thank you

docs/testing.md
patrinoua: done, thank you

docs/testing.md
patrinoua: @pvinis @gkartalis thoughts?

docs/testing.md
patrinoua: Added this, thank you

docs/testing.md
MounirDhahri: Sorry for the late reply. I would suggest something like so.

`flushPromiseQueue` is a helper we to wait for the UI for an animation frame. This can be useful in case you need to test an operation that happens asynchronously and that takes less than a frame

docs/testing.md
MounirDhahri: 🙏

docs/testing.md
MounirDhahri: Sorry for the late reply, thanks for helping with the @gkartalis

docs/testing.md
```diff
@@ -1,7 +1,124 @@
...
+
+Therefore we prefer using `screen.getByText`, `screen.getByDisplayValue`, etc, and as a last resort use a `testID` with `screen.getByTestId`.
+
+In our code base there is some "Old fashioned" testing used for Class components for example since they can only be used using the old way in combination with QueryRenderer etc
+
+The "new fashion" testing uses > relay v13 and we now prefer to use hooks whenever possible. We would for example use `useLazyLoadQuery` instead of a `QueryRenderer`.
+
+- [`relay-test-utils`](https://relay.dev/docs/guides/testing-relay-components/) is the framework we use for testing relay components, creating mock environments etc.
+
+- We don’t use snapshot tests; they produce too much churn for too little value. It’s okay to test that a component doesn’t throw when rendered, but use [`extractText`](https://github.com/artsy/eigen/blob/4c7c9be69ab1c2095f4d2fed11a040b1bde6eba8/src/lib/tests/extractText.ts) (or similar) to test the actual component tree.
```
gkartalis, line 22: this can be removed

docs/testing.md
```diff
@@ -1,7 +1,124 @@
...
+
+Our preferred methods to use are marked with a ⭐️, while the ones we'd like to avoid are marked as ❗️ depracated.
```
gkartalis, line 24:
```suggestion
Our preferred methods to use are marked with a ⭐️, while the ones we'd like to avoid are marked as ❗️ deprecated.
```

docs/testing.md
```diff
@@ -1,7 +1,124 @@
...
+
+## Test helpers and wrappers, when to use them, and good examples
+
+- `TestRendered` is a component we create, using the component we want to test. We use a TestRendered when for example we want to create a mock query. We don't always need to use a test renderer.
```
gkartalis, line 28:
```suggestion
- `TestRenderer` is a component we create, using the component we want to test. We use a TestRenderer when for example we want to create a mock query. We don't always need to use a test renderer.
```

docs/testing.md
```diff
@@ -1,7 +1,124 @@
...
+
+New fashion: You can use useLazyLoadQuery inside a TestRendered which is the new relay hooks way for querying data and the way we generally want to swift towards.
+
+New fashion Example: [ArtworkConsignments.tests.tsx](https://github.com/artsy/eigen/blob/6360ddb5304e6c5bb1dc207db13d5bf0a5d7d9b6/src/app/Scenes/Artwork/Components/ArtworkConsignments.tests.tsx#L28-L38)
+
+Old fashioned: You can use a TestRendered to wrap your component with a QueryRenderer and pass it the test query data.
```
gkartalis, line 34:
```suggestion
Old fashioned: You can use a TestRenderer to wrap your component with a QueryRenderer and pass the test query data.
```

docs/testing.md
```diff
@@ -1,7 +1,124 @@
...
+Old fashion Example: [Inbox.tests.tsx](https://github.com/artsy/eigen/blob/c96dd0807555d69ca2e8655dc68085276d249080/src/app/Containers/Inbox.tests.tsx)
+
+- ⭐️ [`renderWithWrappers`](https://github.com/artsy/eigen/blob/main/src/app/tests/renderWithWrappers.tsx#L88-L103) Our default method for using a component. Can be used on all components. wraps the component with a number of wrappers such as TrackingProvider, GlobalStoreProvider, SafeAreaProvider, etc. Using `testing library`. See the list of wrappers [here](https://github.com/artsy/eigen/blob/main/src/app/tests/renderWithWrappers.tsx#L19-L25).
```
gkartalis, line 37:
```suggestion
- ⭐️ [`renderWithWrappers`](https://github.com/artsy/eigen/blob/main/src/app/tests/renderWithWrappers.tsx#L88-L103) Our default method for using a component. Can be used on all components. Wraps the component with a number of wrappers such as TrackingProvider, GlobalStoreProvider, SafeAreaProvider, etc. Using `testing library`. See the list of wrappers [here](https://github.com/artsy/eigen/blob/main/src/app/tests/renderWithWrappers.tsx#L19-L25).
```

docs/testing.md
```diff
@@ -1,7 +1,124 @@
...
+
+Very similar to `renderWithWrappers`. The difference is that it will also resolve the first api request **only**.
+
+If your component has more than one requests you should use `renderWithWrappers` instead, in combination with one of the functions that resolve api requests such as `resolveMostRecentRelayOperation` or `rejectMostRecentRelayOperation` Example: **\_**
+
+This is an abstraction that is sometimes convenient.
+
+Example of a component with one request [PartnerLocationSection.tests.tsx](https://github.com/artsy/eigen/blob/7703aa01103a06b69b650af11bc9903ab37b1c4b/src/app/Scenes/Partner/Components/PartnerLocationSection.tests.tsx#L21).
+
+`setupTestWrapperTL` can be used to replace the combination of **TestRendered** and **renderWithWrappers**.
```
gkartalis, line 53:
```suggestion
`setupTestWrapperTL` can be used to replace the combination of **TestRenderer** and **renderWithWrappers**.
```

docs/testing.md
```diff
@@ -1,7 +1,124 @@
...
+
+It returns the function `renderWithRelay` which is similar to `renderWithWrappers`. Both `renderWithRelay` and `renderWithWrappers` use [render](https://testing-library.com/docs/angular-testing-library/api/#render) under the hood. I
+
+t gives you access to all the [queries](https://callstack.github.io/react-native-testing-library/docs/api-queries) such as getBy, getAllBy, queryBy, queryAllBy, findBy, findAllBy, etc.
```
gkartalis, line 57:
```suggestion
It gives you access to all the [queries](https://callstack.github.io/react-native-testing-library/docs/api-queries) such as getBy, getAllBy, queryBy, queryAllBy, findBy, findAllBy, etc.
```

docs/testing.md
```diff
@@ -1,7 +1,124 @@
...
+
+❗️ It won't work if there is already a component using a query renderer. [eg?]
+
+// TODO: rename setupTestWrapperTL to setupTestWrapper and setupTestWrapper to setupTestWrapperLegacy
+// Decide if we want to use this pattern or not.
+
+Suggestion: Would it make sense to either remove this abstraction alltogether, or suggest it as the main course of action?
```
gkartalis, line 64: Not sure If these lines need to be kept here

src/app/tests/setupTestWrapper.tsx
```diff
@@ -53,6 +53,9 @@ export const setupTestWrapperTL = <T extends OperationType>({
   return { renderWithRelay }
 }
 
+/**
+ * @deprecated avoid using this, use setupTestWrapperTL (above) instead.
```
pvinis, line 57:
```suggestion
 * @deprecated avoid using this, use setupTestWrapperTL instead.
```

gkartalis (commented):
Thanks for the changes, added some comments for typos and small suggestions, other than that when you fix those feel free to merge!

docs/testing.md
pvinis, line 22:
```suggestion
- We don’t use snapshot tests; they produce too much churn for too little value.
```

docs/testing.md
pvinis: lets keep them 🤷. we do that for code I guess :D

PR #7601, opened 2022-11-07
fix: Adjust Submission My Collection artwork image logic

olerichter00:
This PR resolves [CX-3127] 

### Description

This PR stores the ID of the last submission to ensure that only the artwork in My Collection with this ID displays the image of the submission. Without this correction, any artwork in My Collection without an image set would display the image of the last submission.

### PR Checklist

- [ ] I tested my changes on **iOS** / **Android**.
- [ ] I added screenshots or videos to illustrate my changes.
- [ ] I added Tests and Stories for my changes.
- [ ] I added an [app state migration].
- [ ] I hid my changes behind a [feature flag].
- [ ] I have prefixed changes that need to be tested during a release QA with **[NEEDS EXTERNAL QA]** on the changelog.

### To the reviewers 👀

- [ ] I would like **at least one** of the reviewers to run this PR on the simulator or device.

<details><summary>Changelog updates</summary>

#### Cross-platform user-facing changes

- Adjust image logic for My Collection after submitting an artwork - ole

#### iOS user-facing changes

-

#### Android user-facing changes

-

#### Dev changes

-

</details>

Need help with something? Have a look at our [docs], or get in touch with us.

[app state migration]: ../blob/main/docs/adding_state_migrations.md
[feature flag]: ../blob/main/docs/developing_a_feature.md
[docs]: ../blob/main/docs/README.md

[CX-3127]: https://artsyproduct.atlassian.net/browse/CX-3127?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #7710, opened 2022-11-22
fix(CX-2493): Fix Photo Added in Edit mode not updated on MyCollectionArtwork

lordkiz:
This PR resolves [CX-2493] 

### Description

This PR fixes a bug where NoImage Component continues to show after a user adds new images to a MyCollectionArtwork that previously had no images.

#### Before

https://user-images.githubusercontent.com/18648835/203365183-d8fa4b3a-6e22-46d1-96fe-29cae35ba2be.mp4

#### After

https://user-images.githubusercontent.com/18648835/203365834-c3b72ddd-57ad-44a7-ac57-6dadcdb76f3e.mov

### PR Checklist

### To the reviewers 👀

- Fix Photo Added in Edit mode not updated on MyCollectionArtwork - kizito

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-2493]: https://artsyproduct.atlassian.net/browse/CX-2493?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/MyCollection/Components/MyCollectionImageView.tsx
+++ b/src/app/Scenes/MyCollection/Components/MyCollectionImageView.tsx
@@ -13,6 +13,7 @@ export interface MyCollectionImageViewProps {
   aspectRatio?: number
   artworkSlug: string
   artworkSubmissionId?: string | null
+  myCollectionIsRefreshing?: boolean
 }
 
 export const MyCollectionImageView: React.FC<MyCollectionImageViewProps> = ({
@@ -22,6 +23,7 @@ export const MyCollectionImageView: React.FC<MyCollectionImageViewProps> = ({
   aspectRatio,
   artworkSlug,
   artworkSubmissionId,
+  myCollectionIsRefreshing,
 }) => {
   const color = useColor()
   const [localImage, setLocalImage] = useState<LocalImage | null>(null)
@@ -33,7 +35,7 @@ export const MyCollectionImageView: React.FC<MyCollectionImageViewProps> = ({
         setLocalImage(images[0])
       }
     })
-  }, [])
+  }, [myCollectionIsRefreshing])
 
   const {
     photosForMyCollection: { photos },
@@ -52,7 +54,20 @@ export const MyCollectionImageView: React.FC<MyCollectionImageViewProps> = ({
   }, [])
 
   const renderImage = () => {
-    if (localImage) {
+    // prioritise imageURL
+    if (imageURL) {
+      const targetURL = imageURL.replace(":version", "square")
+      return (
+        <OpaqueImageView
+          testID="Image-Remote"
+          imageURL={targetURL}
+          retryFailedURLs
+          height={imageHeight}
+          width={imageWidth}
+          aspectRatio={aspectRatio}
+        />
+      )
+    } else if (localImage) {
       return (
         <RNImage
           testID="Image-Local"
@@ -78,18 +93,6 @@ export const MyCollectionImageView: React.FC<MyCollectionImageViewProps> = ({
           source={{
             uri: localImageConsignments.path,
           }}
-        />
-      )
-    } else if (imageURL) {
-      const targetURL = imageURL.replace(":version", "square")
-      return (
-        <OpaqueImageView
-          testID="Image-Remote"
-          imageURL={targetURL}
-          retryFailedURLs
-          height={imageHeight}
-          width={imageWidth}
-          aspectRatio={aspectRatio}
         />
       )
     } else {
--- a/src/app/Scenes/MyCollection/Screens/Artwork/Components/MyCollectionArtworkHeader.tsx
+++ b/src/app/Scenes/MyCollection/Screens/Artwork/Components/MyCollectionArtworkHeader.tsx
@@ -1,4 +1,5 @@
 import { tappedCollectedArtworkImages } from "@artsy/cohesion"
+import { useNavigation } from "@react-navigation/native"
 import { MyCollectionArtworkHeader_artwork$key } from "__generated__/MyCollectionArtworkHeader_artwork.graphql"
 import { navigate } from "app/navigation/navigate"
 import {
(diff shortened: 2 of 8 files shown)


src/app/Scenes/MyCollection/Components/MyCollectionImageView.tsx
```diff
@@ -52,7 +54,20 @@ export const MyCollectionImageView: React.FC<MyCollectionImageViewProps> = ({
...
+        <OpaqueImageView
+          testID="Image-Remote"
+          imageURL={targetURL}
+          retryFailedURLs
+          height={imageHeight}
+          width={imageWidth}
+          aspectRatio={aspectRatio}
+        />
+      )
+    } else if (localImage) {
```
dariakoko, line 70: We keep changing the order of these conditions 🙈 
https://github.com/artsy/eigen/pull/7601/files#r1015313667

src/app/Scenes/MyCollection/Screens/Artwork/Components/MyCollectionArtworkHeader.tsx
```diff
@@ -45,30 +50,53 @@ export const MyCollectionArtworkHeader: React.FC<MyCollectionArtworkHeaderProps>
...
   const { trackEvent } = useTracking()
 
+  const navigation = useNavigation()
+
+  useEffect(() => {
+    const unsubscribe = navigation?.addListener("focus", () => {
+      handleImages()
+    })
+    return unsubscribe
+  }, [navigation])
```
dariakoko, line 60: Could you, please, explain, what this useEffect does?

dariakoko (commented):
LGTM
I only left a couple of questions

src/app/Scenes/MyCollection/Screens/Artwork/Components/MyCollectionArtworkHeader.tsx
```diff
@@ -45,30 +50,53 @@ export const MyCollectionArtworkHeader: React.FC<MyCollectionArtworkHeaderProps>
...
+  const mapLocalImages = (localImages: LocalImage[]) => {
+    return localImages.map((localImage) => ({
+      url: localImage.path,
+      width: localImage.width,
+      height: localImage.height,
+      deepZoom: null,
+    }))
+  }
+
+  const handleImages = async () => {
```
MounirDhahri, line 75: **suggestion:** I would maybe consider renaming this to something like `updateLocalImagesState` or so and maybe even potentially adding a comment to it

src/app/Scenes/MyCollection/Screens/Artwork/Components/MyCollectionArtworkHeader.tsx
```diff
@@ -77,7 +105,7 @@ export const MyCollectionArtworkHeader: React.FC<MyCollectionArtworkHeaderProps>
   return (
     <Join separator={<Spacer my={1} />}>
       {/* ImageCarousel */}
-      {!!imagesToDisplay ? (
+      {!!imagesToDisplay && imagesToDisplay.length ? (
```
MounirDhahri, line 108: **praise:** 👏

src/app/Scenes/MyCollection/Components/MyCollectionImageView.tsx
```diff
@@ -22,6 +23,7 @@ export const MyCollectionImageView: React.FC<MyCollectionImageViewProps> = ({
   aspectRatio,
   artworkSlug,
   artworkSubmissionId,
+  myCollectionIsRefreshing,
```
MounirDhahri, line 26: **issue:** I am really not a big fan of myCollectionIsRefreshing prop drilling here and would like to see if we can avoid it. I remember at some point we added some events listeners here https://github.com/artsy/eigen/blob/main/src/app/utils/refreshHelpers.ts#L7 have you considered using them

src/app/Scenes/MyCollection/Components/MyCollectionImageView.tsx
lordkiz: 😂 We actually want to avoid localImages as soon as we can

src/app/Scenes/MyCollection/Screens/Artwork/Components/MyCollectionArtworkHeader.tsx
lordkiz: When a user lands on the MyCollectionArtwork Detail Page, we want to ensure they are seeing fresh images. If we don't refresh the image when a user refocuses on this screen, they might end up seeing a stale image. For example, for an artwork without images, when the user adds images and goes back to this page, it will continue to show NoImage Icon. But with the listener, once the screen is focused again, we check the local store if there are images to display.

src/app/Scenes/MyCollection/Screens/Artwork/Components/MyCollectionArtworkHeader.tsx
lordkiz: But we do not always updateLocalImage with this. It just checks whether to switch from unavailable processed images to local or to leave the state as it is. Everything about image whether local or not, is determined by this function. So I think referencing Local in the name might be a misnomer. But happy to change it if it is still confusing.

src/app/Scenes/MyCollection/Components/MyCollectionImageView.tsx
lordkiz: Thanks @MounirDhahri that would have been a cleaner approach. But the components that would rely on this event might be numerous. For example in this case, every `MyCollectionImageView` would need to register a listener to be able to update the image. We already have registered two: MyCollection and Insights. We will probably have another 8 left because we capped it at 10. If every `MyCollectionImageView` registers a listener, we will immediately run out for any Mycollection with more than 8 artworks. While this would only just trigger a warning and will not necessarily result in a crash, we will be using up a lot of resources maintaining the listeners and this would affect performance drastically.

MounirDhahri: I got what you mean now, thanks for explaining

MounirDhahri (approved):

src/app/Scenes/MyCollection/Components/MyCollectionImageView.tsx
dariakoko: 😄 oki)

olerichter00: The previous order was intentional. During the image processing, the image is available (imageURL is present) but the image dimensions are not which leads to the image not being displayed correctly (stretched). 

I've tried to fix this issue by displaying the local image as long as it is available.

olerichter00: The local images are stored by artwork ID and not by image ID. This is probably the reason why this logic fails when the user edits an artwork (adds or removes images).

PR #7720, opened 2022-11-23
fix: handle marketing deeplinks on typescript side

brainbicycle:
This PR resolves [] 

### Description

Follow-up to https://github.com/artsy/eigen/pull/7705

Fixes a bug where marketing deeplinks would open safari if app was in a killed state.

So it turns out that our deeplinks are being handled by both typescript and native code. 
In my testing when the app is killed the deeplink code here gets fired and calls navigate: 
https://github.com/artsy/eigen/blob/f462ca578be55a3e83c946941d5e118c0b80169a/src/app/utils/useDeepLinks.ts#L17

And when the app is in the background it is handled here and calls navigate: 
https://github.com/artsy/eigen/blob/f462ca578be55a3e83c946941d5e118c0b80169a/ios/Artsy/App/ARAppActivityContinuationDelegate.m#L49

I am guessing we can move this all to typescript and it is probably the reason we get double navigation on deeplinks from time to time and need the weird ignoreDebounce code in our navigation. 

But for now just handling the marketing links on the typescript side as well and following redirects. 

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/navigation/navigate.tests.tsx
+++ b/src/app/navigation/navigate.tests.tsx
@@ -122,6 +122,38 @@ describe(navigate, () => {
       `)
   })
 
+  describe("marketing links", async () => {
+    const fetch = jest.fn((_url, _init) =>
+      Promise.resolve({ url: "https://artsy.net/artist/kaws" })
+    )
+    // @ts-ignore
+    global.fetch = fetch
+    beforeEach(() => {
+      fetch.mockClear()
+    })
+
+    it("reroutes marketing links", async () => {
+      await navigate("https://click.artsy.net/artist/kaws")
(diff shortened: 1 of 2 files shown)


MounirDhahri:
Did you try rebasing to fix the build error? that helped yesterday fix the same issue on one of Kizito's PRs

MounirDhahri (dismissed):
LGTM, the only suggestion I have is to maybe add a test to avoid the same regression in the future

brainbicycle:
> LGTM, the only suggestion I have is to maybe add a test to avoid the same regression in the future

yeah good call!

PR #7787, opened 2022-12-06
refactor: finally rename admin menu to dev menu

pvinis:
### Description

we have been calling this the admin menu, and the dev menu. for the longest time we called it admin menu but we know its wrong, and we call it dev menu but we know people wont find it if we call it that, so we kept calling it the wrong thing.

i am very happy to create this PR and finally rename this to dev menu!

i did *not* change the old native code, so the files there are still called `*admin*`, but the will just go away in time, and that's ok.

along with the rename of the menu, I renamed the `adminOverrides`, since they are not actually admin overrides, they are local overrides on the device. no admin-ing takes place for these.

in the same vein, I moved the `environment` the app is using to the `devicePrefs`, as they are a device-located state. they are not attached to the artsy account in any way, so they should not be in the artsyPrefs. 

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>


src/app/store/migration.ts
```diff
@@ -250,6 +252,16 @@ export const artsyAppMigrations: Migrations = {
...
   },
+  [Versions.RenameAdminToLocalOverridesForFeatures]: (state) => {
+    state.artsyPrefs.features.localOverrides = state.artsyPrefs.features.adminOverrides
+    delete state.artsyPrefs.features.adminOverrides
+  },
+  [Versions.MoveEnvironmentToDevicePrefsAndRenameAdminToLocal]: (state) => {
+    state.devicePrefs.environment = state.artsyPrefs.environment
+    delete state.artsyPrefs.environment
+    state.devicePrefs.environment.localOverrides = state.devicePrefs.environment.adminOverrides
+    delete state.devicePrefs.environment.adminOverrides
```
MounirDhahri, line 263: **thought:** is this needed? I believe it's already done above right?

pvinis: i am trying to keep as clean a state as possible, so after assigning it, I also delete the old 🤷

docs/developing_a_feature.md
```diff
@@ -101,12 +101,12 @@ This is marked as unsafe because it will not cause react components to re-render
 
 To enable your feature on a physical or virtual device
 
-- Open the admin menu by pressing `command + control + z` on a mac. If this doesn't work:
+- Open the dev menu by pressing `command + control + z` on a mac. If this doesn't work:
```
patrinoua, line 104: Can we perhaps call it Dev menu / DevMenu / Dev Menu to show it's more like a component, an actual thing?

patrinoua: In debugging.md we refer to it as Dev Menu so maybe we can stick to that

src/app/store/migration.ts
MounirDhahri: makes sense. this affects only devs so it's not a problem anyway

docs/developing_a_feature.md
pvinis: done

MounirDhahri (approved):

gkartalis (approved):
Let's ship this???

pvinis:
i was just waiting for @brainbicycle but i guess we we can yea

PR #7796, opened 2022-12-06
feat: add show only upcoming auction results filter

MounirDhahri:
[THIS CAN BE REVIEWED]

This PR resolves [CX-3222] 

### Description

- Allow filtering for upcoming lots

https://user-images.githubusercontent.com/11945712/207573835-0bd3feb5-8066-4a51-8e64-df238e2f1d24.mov

cc @egdbear 

Blocked by https://github.com/artsy/metaphysics/pull/4580

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-3222]: https://artsyproduct.atlassian.net/browse/CX-3222?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Components/ArtworkFilter/ArtworkFilterOptionsScreen.tsx
+++ b/src/app/Components/ArtworkFilter/ArtworkFilterOptionsScreen.tsx
@@ -168,7 +168,9 @@ export const ArtworkFilterOptionsScreen: React.FC<
                 item={item}
                 count={selectedFiltersCount}
                 onPress={() => {
-                  navigateToNextFilterScreen(item.ScreenComponent)
+                  navigateToNextFilterScreen(
+                    item.ScreenComponent as keyof ArtworkFilterNavigationStack
+                  )
                 }}
               />
             )
@@ -390,7 +392,6 @@ export const filterOptionToDisplayConfigMap: Record<string, FilterDisplayConfig>
   includeUpcoming: {
     displayText: FilterDisplayName.includeUpcoming,
     filterType: "includeUpcoming",
-    // @ts-ignore
     ScreenComponent: "none",
     configType: FilterConfigTypes.FilterScreenCheckboxItem,
   },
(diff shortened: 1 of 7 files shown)


egdbear:
❤️

src/app/Components/ArtworkFilter/ArtworkFilterOptionsScreen.tsx
```diff
@@ -386,6 +387,13 @@ export const filterOptionToDisplayConfigMap: Record<string, FilterDisplayConfig>
     filterType: "estimateRange",
     ScreenComponent: "EstimateRangeOptionsScreen",
   },
+  includeUpcoming: {
+    displayText: FilterDisplayName.includeUpcoming,
+    filterType: "includeUpcoming",
+    // @ts-ignore
+    ScreenComponent: "none",
+    configType: FilterConfigTypes.FilterScreenCheckboxItem,
+  },
```
lordkiz, line 396: Do we need to add this here? I don't think so if this is meant to be used as a local filter. But if you absolutely must, then the ScreenComponent should be the current screen which in this case would be `FilterOptionsScreen` , so that users are not navigated away, I suppose.
For example https://github.com/artsy/eigen/blob/main/src/app/Scenes/MyCollection/utils/localArtworkSortAndFilter.ts#L118

lordkiz (commented):
I left a comment. However I am not sure about the requirements. Is the filtering going to happen in the backend or are we filtering the results of already retrieved auctions?
This looks good so far. 👍

src/app/Components/ArtworkFilter/ArtworkFilterOptionsScreen.tsx
MounirDhahri: thanks for the comment @lordkiz but this is not going to be a local filter, it will be happening in the backend

MounirDhahri: I adjsuted the ts logic a bit to avoid the `ts-ignore` though

MounirDhahri:
Merging this to get it to the release hopefully

PR #7831, opened 2022-12-09
fix: environment swap

pvinis:
This PR resolves [] 

### Description

quick fix after the #7787. the env could not be swapped from stag to prod. the reason was that when we reset the state after logout, we do a complete reset and we just keep some stuff. when the env state was moved to devicePrefs from artsyPrefs, it was not being kept in that reset func.

two other quick changes, one is just the view on devmenu from
Screenshot 2022-12-09 at 12 01 33
to
Screenshot 2022-12-09 at 12 04 58

and finally adding rounded corners on the toast.

both things were things I touched while debugging this.

### PR Checklist

- [x] I have prefixed changes that need to be tested during a release QA with **[NEEDS EXTERNAL QA]** on the changelog.

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Components/Toast/ToastComponent.tsx
+++ b/src/app/Components/Toast/ToastComponent.tsx
@@ -74,6 +74,7 @@ export const ToastComponent = ({
         left={(width - MIDDLE_TOAST_SIZE) / 2}
         backgroundColor={color(backgroundColor)}
         opacity={opacityAnim.interpolate({ inputRange: [0, 1], outputRange: [0, 0.9] })}
+        borderRadius={6}
         overflow="hidden"
         zIndex={999}
       >
--- a/src/app/store/GlobalStoreModel.ts
+++ b/src/app/store/GlobalStoreModel.ts
@@ -86,16 +86,19 @@ export const getGlobalStoreModel = (): GlobalStoreModel => ({
     (a) => a.auth.signOut,
     (actions, _, store) => {
       const {
-        artsyPrefs: existingConfig,
+        artsyPrefs: existingArtsyConfig,
+        devicePrefs: existingDeviceConfig,
         search,
         auth: { userID },
         recentPriceRanges,
       } = store.getState()
 
       // keep existing config state
-      const config = sanitize(existingConfig) as typeof existingConfig
+      const artsyConfig = sanitize(existingArtsyConfig) as typeof existingArtsyConfig
+      const deviceConfig = sanitize(existingDeviceConfig) as typeof existingDeviceConfig
       actions.reset({
-        artsyPrefs: config,
+        artsyPrefs: artsyConfig,
+        devicePrefs: deviceConfig,
         search,
         recentPriceRanges,
         auth: { previousSessionUserID: userID },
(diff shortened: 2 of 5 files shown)


src/app/Components/Toast/ToastComponent.tsx
```diff
@@ -74,6 +74,7 @@ export const ToastComponent = ({
         left={(width - MIDDLE_TOAST_SIZE) / 2}
         backgroundColor={color(backgroundColor)}
         opacity={opacityAnim.interpolate({ inputRange: [0, 1], outputRange: [0, 0.9] })}
+        borderRadius={6}
```
pvinis, line 77: some nice rounded corners

src/app/store/GlobalStoreModel.ts
```diff
@@ -86,16 +86,19 @@ export const getGlobalStoreModel = (): GlobalStoreModel => ({
...
       } = store.getState()
 
       // keep existing config state
-      const config = sanitize(existingConfig) as typeof existingConfig
+      const artsyConfig = sanitize(existingArtsyConfig) as typeof existingArtsyConfig
+      const deviceConfig = sanitize(existingDeviceConfig) as typeof existingDeviceConfig
       actions.reset({
-        artsyPrefs: config,
+        artsyPrefs: artsyConfig,
+        devicePrefs: deviceConfig,
```
pvinis, line 101: we were not keeping this, now its fine and we can swap env again

src/app/store/persistence.ts
```diff
@@ -10,6 +10,9 @@ export const STORAGE_KEY = "artsy-app-state"
 
 export const LEGACY_SEARCH_STORAGE_KEY = "SEARCH/RECENT_SEARCHES"
 
+/**
+ * Removes sessionState and computed properties.
```
pvinis, line 14: quick little doc

src/app/utils/DevMenu.tsx
```diff
@@ -242,7 +244,13 @@ export const DevMenu = ({ onClose = () => dismissModal() }: { onClose(): void })
             }}
           />
           <FeatureFlagMenuItem title={`Sentry release name: "${eigenSentryReleaseName()}"`} />
-          <FeatureFlagMenuItem title={`Device ID: "${getUniqueId()}"`} />
+          <FeatureFlagMenuItem
+            title={`Device ID: ${getUniqueIdSync()}`}
```
pvinis, line 248: changing to the sync version, since it was displaying `[Object object]`, now its the actual id. and it copies when you tap.

src/app/utils/DevMenu.tsx
```diff
@@ -530,7 +538,7 @@ export const FeatureFlagMenuItem: React.FC<{
           </Text>
         </Flex>
         {!!value && (
-          <Flex flex={2} flexDirection="row" alignItems="center">
+          <Flex flex={3} flexDirection="row" alignItems="center">
```
pvinis, line 541: give some more space to the values, so we get `Production` instead of `Produc...`

src/app/store/GlobalStoreModel.ts
MounirDhahri: nice!

MounirDhahri (approved):
Thanks for looking into this Pavlos 🙌

PR #7926, opened 2023-01-04
chore: Clean up AREnableHomeScreenArtworkRecommendations

olerichter00:
This PR resolves [CX-3023] 

### Description
Clean up AREnableHomeScreenArtworkRecommendations.

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-3023]: https://artsyproduct.atlassian.net/browse/CX-3023?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/Home/Home.tsx
+++ b/src/app/Scenes/Home/Home.tsx
@@ -102,7 +102,6 @@ const Home = (props: Props) => {
     relay,
   } = props
 
-  const enableArtworkRecommendations = useFeatureFlag("AREnableHomeScreenArtworkRecommendations")
   const enableMyCollectionHFOnboarding = useFeatureFlag("AREnableMyCollectionHFOnboarding")
   const showUpcomingAuctionResultsRail = useFeatureFlag("ARShowUpcomingAuctionResultsRails")
   const enableLargeNewWorksForYouRail = useLargeNewWorksForYouRail()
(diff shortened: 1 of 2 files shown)


gkartalis (approved):

MounirDhahri:
Speaking of that - I am wondering if we should update the echo model in Eigen to keep the values to whatever they are if they're missing in the remote echo file.

olerichter00:
> Speaking of that @pvinis - I am wondering if we should update the echo model in Eigen to keep the values to whatever they are if they're missing in the remote echo file.

That's a great idea! In the released app we could default to `true` if there is no flag in Echo and keep the warning when developing. 

I just wonder if there would be any risk of forgetting to add the remote flag for features that are not ready to be released 🤔

pvinis:
I would not do much there, since echo should not be used anymore for features, but let's bring it up in mobile-practice and decide.

MounirDhahri:
> I just wonder if there would be any risk of forgetting to add the remote flag for features that are not ready to be released 🤔

That's a valid point, I think we can raise a dev alert in that case. Anyway in my opinion, we should never remove ffs from Echo since after all. If we would like to keep the file small, maybe we can start generating it and separate the old and new flags.

PR #7932, opened 2023-01-04
chore: More closely match designs for content cards on larger devices GRO-1438

jonallured:
This PR introduces a "breakpoint" where larger devices get content cards images that are half the width of the screen. This was something we meant to return to after the cards were launched but are just now getting to. Looks like this:

Screen Shot 2023-01-04 at 2 15 47 PM

I looked at this live with Lois and we picked 700 as the magic number but if anybody has a better idea I'm open to it!

https://artsyproduct.atlassian.net/browse/GRO-1438

/cc @artsy/grow-devs @kathytafel 

[GRO-1438]: https://artsyproduct.atlassian.net/browse/GRO-1438?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pvinis (approved):
nice

PR #7947, opened 2023-01-09
chore: add write SSH key before executing git push commands

artsyit:
By default, the `checkout` command uses a read-only SSH key to clone the repository during builds. When a build needs a write SSH key to push code/tags back to the GitHub repository via `git push`, a job can pull it down via the `add_write_ssh_key` command.


dblandin:
Superseded by https://github.com/artsy/eigen/pull/7948.

PR #7975, opened 2023-01-13
fix(debugging): fix remote debugging; remove reanimated patch 

damassi:
### Description

Remote debugging seems to have broken due to native static methods being called at runtime in `react-native-device-info`. This fixes that. 

See https://github.com/react-native-device-info/react-native-device-info/issues/776

Also removes the reanimated patch since type-checking now passes. 

cc @artsy/mobile-platform 

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>


src/app/utils/DevMenu.tsx
```diff
@@ -553,3 +555,24 @@ export const FeatureFlagMenuItem: React.FC<{
     </Touchable>
   )
 }
+
+const getInfo = () => {
+  // Stub out these values if debugging remotely since `react-native-device-info` leads to a crash
```
brainbicycle, line 560: might be worth an entry in HACKS.md

brainbicycle (dismissed):
nice sleuthing!

damassi:
Ugh, the old issue is back down: 

Screenshot 2023-01-13 at 2 46 35 PM

damassi:
Commenting out the cache seemed to fix it: https://github.com/artsy/eigen/blob/7a253691642b585413866982f91b23a3087a6f9c/metro.config.js#L5

Obviously we don't want to uncomment this, but in any case heads up if you need to debug and run into this issue.

HACKS.md
```diff
@@ -327,12 +327,16 @@ Maybe we don't? We can try to remove it at any point, and see if it works. Try w
 This is needed because xcode 14 says that React-Core-AccessibilityResources and some other pods require a development team.
 We don't really need a dev team for these. Probably some future version of cocoapods will fix this.
 
-## react-native-reanimated patch-package
```
damassi, line 330: @pvinis - this conditions for removal were 👍 , upgrading to the latest minor release and everything typechecked. Anything i'm missing here?

src/app/utils/DevMenu.tsx
```diff
@@ -137,7 +142,10 @@ export const DevMenu = ({ onClose = () => dismissModal() }: { onClose(): void })
 
         <CollapseMenu title="Feature Flags">
           <Flex px={2} mb={1}>
-            <SearchInput onChangeText={setFeatureFlagQuery} placeholder="Search feature flags" />
+            {/* Skip search if debugging remotely. This throws an error in react-native-reanimated */}
+            {!isDebuggingRemotely && (
+              <SearchInput onChangeText={setFeatureFlagQuery} placeholder="Search feature flags" />
+            )}
```
damassi, line 148: Fell into a bit of a rabbit hole here which led me to the patch. For some reason remote js debugging doesn't like this SearchInput? Not too sure, but guarding against it seemed to work. 

In the meantime I removed the patch since type-checking now passes.

src/app/utils/DevMenu.tsx
damassi: done

damassi:
Closing this, need to be triple sure about this log loop / reanimated around the app. Looks like reanimated in general is breaking remote JS debugging. Lets see if there are alternatives. 

Will open a separate PR about removing the patch.

damassi:
And alas, said cache was confusing said patch and a fresh install with a fresh cache blew up the stack; and so: the patch remains a patch, and that's that ☂️ 🚪

pvinis:
reanimated V2 has broken remote debug. it's fixed on v3 which is still in rc. we just need to wait until then I guess.

damassi:
I wonder what we can do till then, I'll def lose my sanity without a debugger, especially in Eigen!

gkartalis (commented):
@damassi are you using [flipper](https://fbflipper.com/)? It works really well for logs/ network requests and has a bunch of open source plugins.

I also **really** miss breakpoints though in eigen in some cases 😅

damassi:
Yup @gkartalis, using flipper, but need proper object inspection (like working on web), breakpoints, etc! Basically, need chrome dev tools.

PR #7990, opened 2023-01-17
chore: Remove stale onboarding gene screens GRO-1510

jonallured:
This PR is some follow up from this thread:

https://artsy.slack.com/archives/C01ADJNCS5D/p1673554200139919

What I'm doing here is removing the gene screens in favor of the collection screens and then updating the names of things.

https://artsyproduct.atlassian.net/browse/GRO-1510

/cc @artsy/grow-devs


mdole (approved):
:chef-kiss:

PR #8006, opened 2023-01-19
feat(CX-2974): Money Input

lordkiz:
This PR resolves [CX-2974] 

### Description

- This PR adds a MoneyInput component modelled after the existing PhoneInput component to Palette
- Provides a base for easily creating and updating Select+Input combination components
- Update the MyCollection Artwork form to use the MoneyInput component

#### Blank
https://user-images.githubusercontent.com/18648835/213443062-eaffa7ae-6efb-4146-8025-0e7a780d200e.mov

#### Prepopulated
https://user-images.githubusercontent.com/18648835/213442514-f333ff2a-97a7-4c43-be2a-170c16272ac9.mov

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-2974]: https://artsyproduct.atlassian.net/browse/CX-2974?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/palette/elements/Input/INTERNALSelectAndInputCombinationBase.tsx
+++ b/src/palette/elements/Input/INTERNALSelectAndInputCombinationBase.tsx
@@ -1,0 +1,105 @@
+import { Input, InputProps, InputRef } from "palette"
+import { forwardRef, useImperativeHandle } from "react"
+import { useEffect, useRef } from "react"
+import { Platform } from "react-native"
+import { Select } from "../Select"
+import { SelectProps } from "../Select/SelectV2"
+
+type TypeForSelect = {
+  [K in keyof SelectProps<string> as `${K}ForSelect`]: SelectProps<string>[K]
+}
+
+export interface ValuePayload {
+  select: {
+    value?: any
+  }
+  input: {
+    value?: string
+  }
+}
+
+/** Underline bar height for text input on android when focused */
+const UNDERLINE_TEXTINPUT_HEIGHT_ANDROID = 1.5
+
+export const INTERNALSelectAndInputCombinationBase = forwardRef<
+  InputRef,
+  {
+    onValueChange: (value: ValuePayload) => void
+    shouldDisplayLocalError?: boolean
+    validate?: () => void
+  } & Omit<InputProps, "onChange" | "onChangeText" | "renderLeftHandSection"> &
+    TypeForSelect
+>(
+  (
+    {
+      inputTextStyle = Platform.select({
+        android: { paddingTop: UNDERLINE_TEXTINPUT_HEIGHT_ANDROID },
+        default: {},
+      }),
+      onValueChange,
+      shouldDisplayLocalError = true,
+      value,
+      validate,
+      // props for Select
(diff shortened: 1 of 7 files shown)


src/palette/elements/Input/INTERNALSelectAndInputCombinationBase.tsx
```diff
@@ -0,0 +1,106 @@
...
+    useEffect(() => {
+      if (isFirstRun.current) {
+        if (value?.length) {
+          validate?.()
+        }
+        isFirstRun.current = false
+        return
+      }
+      validate?.()
+      onValueChange({ select: { value: valueForSelect }, input: { value } })
```
dariakoko, line 72: Could you, please, explain what the `isFirstRun` is doing?

MrSltun (approved):
LGTM!! 🌟 Great work Kizito!
I only have one comment on the font size of the input, it looks a bit smaller, can the variant be changed to `sm`? if it's too complex then we should address it on a Future Friday!

dariakoko (approved):
Looks good, i only left one question

src/palette/elements/Input/INTERNALSelectAndInputCombinationBase.tsx
lordkiz: @dariakoko By default useEffect will fire the first time when run whether or not the dependencies have changed. And would then continue executing the callback whenever the dependencies change. In order to prevent the execution of a part/all of the block of code in the callback on first render, we use a ref (ref and not state because we do not want to reload) with a boolean value, which we switch to opposite on first render, therefore preventing the execution of the that part of the code on first render. Since the ref value has now changed after the first render, the next time the callback is executed, that block of code will run.
In this case we do not want to fire `onValueChange` immediately the component loads. We want to fire it when value actually changes.

dariakoko: Thank you for the explanation @lordkiz

lordkiz:
> LGTM!! 🌟 Great work Kizito!
> 
> I only have one comment on the font size of the input, it looks a bit smaller, can the variant be changed to `sm`? if it's too complex then we should address it on a Future Friday! 

@MrSltun thanks for pointing out this. I just tried passing style, inputTextStyle etc and nothing seems to be able to fix it. Any idea what else I can do?

MrSltun:
@lordkiz it looks a bit complex then, I believe we can merge and address it on a Future Friday :)

MounirDhahri:
Nice PR @lordkiz

MounirDhahri:
Since this wasn't hidden behind a feature flag, the phone variant is broken on the version we will be releasing. Can you fix before the release

MounirDhahri:
Since this wasn't hidden behind a feature flag, the phone variant is broken on the version we will be releasing. Can you please fix before the release

PR #8007, opened 2023-01-20
ci(builder): Wait until tests pass before build

damassi:
### Description

Noticed that we still proceed with the ios and android builds even if tests don't pass; this fixes that. 

cc @artsy/mobile-platform

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -578,6 +578,7 @@ workflows:
                 - beta-android
                 - update-changelog
           requires:
+            - test-js
             - build-test-js
             - horizon/block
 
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -578,6 +578,7 @@ workflows:
                 - beta-android
                 - update-changelog
           requires:
+            - test-js
```
pvinis, line 581: which is what I wanted yesterday. maybe we can have these start and get cancelled with the dynamic config then

pvinis (approved):

.circleci/config.yml
damassi: ah nice, yeah 👍 didn't quite realize what you meant at the time; thought it was wait for this, then for that, but this is just "wait for both"

PR #8015, opened 2023-01-20
feat(FX-4537): Manage top-level activity panel indicator (blue dot) based on if there any entries more recent than the last time the activity panel was opened

dimatretyak:
This PR resolves [FX-4537] 

### Description
We want to handle seen vs unread differently. When the activity panel is opened, we set the most recent notification (top of the list) as the last seen notification. This will inform how we set the top-level bell indicator.

### Acceptance Criteria
* As a user
* Given I have unseen notifications
* When I visit the activity panel
* The top-level indicator goes away
* And I continue to see unread indicators to individual notifications.

### Demo
https://user-images.githubusercontent.com/3513494/214305576-f536bd84-5145-4c50-bcac-4fb54978fa22.mp4

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[FX-4537]: https://artsyproduct.atlassian.net/browse/FX-4537?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/BottomTabs/BottomTabsModel.ts
+++ b/src/app/Scenes/BottomTabs/BottomTabsModel.ts
@@ -36,8 +36,8 @@ export interface BottomTabsModel {
     tabProps: Partial<{ [k in BottomTabType]: object }>
     selectedTab: BottomTabType
   }
-  lastSeendNotificationPublishedAt: string | null
-  setLastSeendNotificationPublishedAt: Action<BottomTabsModel, string | null>
+  lastSeenNotificationPublishedAt: string | null
+  setLastSeenNotificationPublishedAt: Action<BottomTabsModel, string | null>
   syncApplicationIconBadgeNumber: ThunkOn<BottomTabsModel>
   unreadConversationCountChanged: Action<BottomTabsModel, number>
   syncActivityPanelState: Action<
@@ -61,9 +61,9 @@ export const getBottomTabsModel = (): BottomTabsModel => ({
     tabProps: {},
     selectedTab: "home",
   },
-  lastSeendNotificationPublishedAt: null,
-  setLastSeendNotificationPublishedAt: action((state, payload) => {
-    state.lastSeendNotificationPublishedAt = payload
+  lastSeenNotificationPublishedAt: null,
+  setLastSeenNotificationPublishedAt: action((state, payload) => {
+    state.lastSeenNotificationPublishedAt = payload
   }),
   syncApplicationIconBadgeNumber: thunkOn(
     (actions) => [
@@ -199,9 +199,9 @@ export const getBottomTabsModel = (): BottomTabsModel => ({
     const lastNotification = notifications[0] as NotificationNode | undefined
     const lastNotificationPublishedAt = lastNotification?.publishedAt ?? null
     const isLastSeenPublishedAtEmpty =
-      state.lastSeendNotificationPublishedAt === null && lastNotificationPublishedAt
+      state.lastSeenNotificationPublishedAt === null && lastNotificationPublishedAt
     const isNewPublishedAtAvailable = checkIsNewPublishedAt(
-      state.lastSeendNotificationPublishedAt,
+      state.lastSeenNotificationPublishedAt,
       lastNotificationPublishedAt
     )
 
--- a/src/app/Scenes/BottomTabs/BottomTabsModel.ts
+++ b/src/app/Scenes/BottomTabs/BottomTabsModel.ts
@@ -39,7 +39,10 @@ export interface BottomTabsModel {
   lastNotificationPublishedAt: string | null
   syncApplicationIconBadgeNumber: ThunkOn<BottomTabsModel>
   unreadConversationCountChanged: Action<BottomTabsModel, number>
-  actionChanged: Action<BottomTabsModel, { notifications: NotificationNode[]; unreadCount: number }>
+  syncActivityPanelState: Action<
+    BottomTabsModel,
+    { notifications: NotificationNode[]; unreadCount: number }
+  >
   fetchCurrentUnreadConversationCount: Thunk<BottomTabsModel>
   unreadActivityPanelNotificationsCountChanged: Action<BottomTabsModel, number>
   fetchNotificationsInfo: Thunk<BottomTabsModel>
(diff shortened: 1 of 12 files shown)


src/app/Scenes/BottomTabs/BottomTabsModel.ts
```diff
@@ -43,6 +61,10 @@ export const getBottomTabsModel = (): BottomTabsModel => ({
     tabProps: {},
     selectedTab: "home",
   },
+  lastSeendNotificationPublishedAt: null,
```
dimatretyak, line 64: unfortunately we can't get the `userId` of the current user in the `BottomTabsModel`. For this reason, it will not be possible to store `lastSeendNotificationPublishedAt` by `userId`

src/app/Scenes/Home/Components/HomeHeader.tsx
```diff
@@ -3,15 +3,15 @@ import { ArtsyLogoIcon, Box, Flex, Spacer } from "palette"
 import { ActivityIndicator } from "./ActivityIndicator"
 
 export const HomeHeader: React.FC = () => {
-  const hasNotifications = GlobalStore.useAppState(
-    (state) => state.bottomTabs.sessionState.unreadCounts.unreadActivityPanelNotifications > 0
+  const shouldUnreadActivityPanelIndicator = GlobalStore.useAppState(
```
gkartalis, line 6: what about changing the name to
```suggestion
  const shouldDisplayActivityPanelIndicator = GlobalStore.useAppState(
```

src/app/Scenes/BottomTabs/BottomTabsModel.ts
```diff
@@ -94,45 +116,59 @@ export const getBottomTabsModel = (): BottomTabsModel => ({
...
         graphql`
-          query BottomTabsModelFetchAllNotificationsCountsQuery {
+          query BottomTabsModelFetchNotificationsInfoQuery {
             me @principalField {
               unreadConversationCount
               unreadNotificationsCount
             }
+
+            viewer {
+              notificationsConnection(first: 5) {
```
dimatretyak, line 135: **question**: should we query fewer artworks here (note: we then do some [data filtering](https://github.com/artsy/eigen/blob/6f3d9732aecf4e76cabb0fabcb071d5f1346e523/src/app/Scenes/BottomTabs/BottomTabsModel.ts#L192-L198))? Maybe 3 artworks will be enough?

gkartalis (approved):
🔥 left only a suggestion for a variable name change but I leave it up to you 🔥

src/app/Scenes/BottomTabs/BottomTabsModel.ts
gkartalis: This is where we fetch the notifications right? That means that some notifications will be filtered out if they are not "[artwork](https://github.com/artsy/eigen/pull/8015/files#diff-6d4c081dfa6ce30235a724d24103bc3522493f982a4a8c23b4d7f58e873e328cR193)" notifications. 

If my assumption above is true I think that 5 is okay and we shouldn't reduce them 🤔

dimatretyak: > This is where we fetch the notifications right? That means that some notifications will be filtered out if they are not "[artwork](https://github.com/artsy/eigen/pull/8015/files#diff-6d4c081dfa6ce30235a724d24103bc3522493f982a4a8c23b4d7f58e873e328cR193)" notifications.

yep, it's correct

nickskalkin (dismissed):
Looks good! Left one non-blocking question

src/app/Scenes/BottomTabs/BottomTabsModel.ts
```diff
@@ -43,6 +61,10 @@ export const getBottomTabsModel = (): BottomTabsModel => ({
     tabProps: {},
     selectedTab: "home",
   },
+  lastSeenNotificationPublishedAt: null,
```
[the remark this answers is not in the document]
dimatretyak: this is done in order not to display the unread indicator every time when the application is started from "scratch" (when the application was not in the background). Otherwise, with each such launch, the past status "viewed" would be ignored

src/app/Scenes/BottomTabs/BottomTabsModel.ts
```diff
@@ -43,6 +61,10 @@ export const getBottomTabsModel = (): BottomTabsModel => ({
...
+  setLastSeenNotificationPublishedAt: action((state, payload) => {
+    state.lastSeenNotificationPublishedAt = payload
+  }),
```
dimatretyak, line 67: **question**: should we check here that the new `publishedAt` timestamp is more recent than the locally persisted? what do you think about this? Something like this

```javascript
setLastSeenNotificationPublishedAt: action((state, payload) => {
  if (checkIsNewPublishedAt(state.lastSeenNotificationPublishedAt, payload)) {
    state.lastSeenNotificationPublishedAt = payload
  }
})
```

gkartalis: That's a great point, I think it is worth it, might also come in handy for the [next task](https://artsyproduct.atlassian.net/browse/FX-4539) 🤔

gkartalis (dismissed):

dblandin (commented):
LGTM! One suggestion I'd make is to make sure we're using "unread" vs "unseen" consistently. Prior to this change, entries in the activity panel were either read/unread. This additional layer of seen/unseen is new and, for the moment, we're introducing it locally and not persisting that state to our backends.

For entries in the activity panel (and how they're modeled on the backend), and single entry is read either by clicking on the entry or by marking all as read.

For the global indicators, now that we're updating the logic to be more aligned with this "unseen" concept, I think we should shift the language to remain consistent: `hasUnseenNotifications`, `displayUnseenNotificationsIndicator`, etc.

dimatretyak:
@dblandin added changes in [this commit](https://github.com/artsy/eigen/pull/8015/commits/8ab1f3fbb6239d563c4f6ab566cbfcb226219797). Feel free to add comments/remarks if I missed something/misunderstood you

dblandin (approved):
Awesome. LGTM! Thanks for adding those updates 👍🏼

PR #8043, opened 2023-01-24
fix(stack-traces): Disable sentry in __DEV__, fix console and error stack traces

damassi:
### Description

This fixes an issue that surfaced after we got the debugger working in eigen, which is stack and log traces were being obscured by Sentry tapping into the JS pipeline. Now, we disable the hook in `__DEV__`.

**Before**: 

Screenshot 2023-01-24 at 1 12 59 PM

**After**: 

Screenshot 2023-01-24 at 1 08 25 PM

cc @artsy/mobile-platform


pvinis (approved):
nice

PR #8055, opened 2023-01-25
feat: add switchContainer and ff for new search screen

gkartalis:
This PR resolves [FX-4562] 

### Description

- Adds ff `AREnableESOnlySearch` to switch between old and new search
- creates boilerplate code to support the new search screen

### PR Checklist

### To the reviewers 👀

- add switchContainer and ff for new search screen - gkartalis

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[FX-4562]: https://artsyproduct.atlassian.net/browse/FX-4562?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/AppRegistry.tsx
+++ b/src/app/AppRegistry.tsx
@@ -219,8 +219,9 @@ const InnerPageWrapper: React.FC<PageWrapperProps> = ({
   // if we're in a modal, just pass isVisible through
 
   let isVisible = viewProps.isVisible
-  const currentTab = useSelectedTab()
   if (!ignoreTabs) {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    const currentTab = useSelectedTab()
     if (BottomTabOption[viewProps.navStackID as BottomTabType]) {
       // otherwise, make sure it respects the current tab
       isVisible = isVisible && currentTab === viewProps.navStackID
--- a/src/app/AppRegistry.tsx
+++ b/src/app/AppRegistry.tsx
@@ -8,6 +8,8 @@ import {
   WorksForYouScreenQuery,
 } from "app/Components/Containers/WorksForYou"
 import { FadeIn } from "app/Components/FadeIn"
+import { Search2 } from "app/Scenes/Search/Search2"
+import { SearchSwitchContainer } from "app/Scenes/Search/SearchSwitchContainer"
 import React from "react"
 import { AppRegistry, LogBox, Platform, View } from "react-native"
 import { GraphQLTaggedNode } from "relay-runtime"
@@ -217,8 +219,8 @@ const InnerPageWrapper: React.FC<PageWrapperProps> = ({
   // if we're in a modal, just pass isVisible through
 
   let isVisible = viewProps.isVisible
+  const currentTab = useSelectedTab()
   if (!ignoreTabs) {
-    const currentTab = useSelectedTab()
     if (BottomTabOption[viewProps.navStackID as BottomTabType]) {
       // otherwise, make sure it respects the current tab
       isVisible = isVisible && currentTab === viewProps.navStackID
@@ -483,7 +485,9 @@ export const modules = defineModules({
     SellWithArtsyHomeScreenQuery,
   ]),
   SalesNotRootTabView: reactModule(SellWithArtsy),
-  Search: reactModule(SearchScreen, { isRootViewForTabName: "search" }, [SearchScreenQuery]),
+  Search: reactModule(SearchScreen, {}, [SearchScreenQuery]),
+  Search2: reactModule(Search2),
+  SearchSwitchContainer: reactModule(SearchSwitchContainer, { isRootViewForTabName: "search" }),
   Show: reactModule(ShowQueryRenderer, { fullBleed: true }),
   ShowMoreInfo: reactModule(ShowMoreInfoQueryRenderer),
   SavedAddresses: reactModule(SavedAddressesQueryRenderer),
--- a/src/app/AppRegistry.tsx
+++ b/src/app/AppRegistry.tsx
@@ -8,7 +8,7 @@ import {
   WorksForYouScreenQuery,
 } from "app/Components/Containers/WorksForYou"
 import { FadeIn } from "app/Components/FadeIn"
-import { Search2 } from "app/Scenes/Search/Search2"
+import { SearchScreenQuery } from "app/Scenes/Search/Search"
 import { SearchSwitchContainer } from "app/Scenes/Search/SearchSwitchContainer"
 import React from "react"
 import { AppRegistry, LogBox, Platform, View } from "react-native"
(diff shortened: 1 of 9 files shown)


src/app/AppRegistry.tsx
```diff
@@ -217,8 +219,8 @@ const InnerPageWrapper: React.FC<PageWrapperProps> = ({
   // if we're in a modal, just pass isVisible through
 
   let isVisible = viewProps.isVisible
+  const currentTab = useSelectedTab()
```
gkartalis, line 222: (Out of context change) Moved this line above the if because it was breaking the rules of react hooks and it was being called conditionally.

The new `eslint` was complaining so I fixed it in this pr

src/app/routes.ts
```diff
@@ -117,6 +117,7 @@ function getDomainMap(): Record<string, RouteMatcher[] | null> {
     addRoute("/", "Home"),
     addRoute("/sales", "Sales"),
     addRoute("/map", "NewMap"),
+    // TODO: this should be replaced with the new route when ES only Search is ready
     addRoute("/search", "Search"),
```
dimatretyak, line 121: **question**: should we use `addRoute("/search", "SearchSwitchContainer")` now?

src/app/AppRegistry.tsx
gkartalis: good point 🤔 let me raise it we are having a linting discussion [here](https://artsy.slack.com/archives/C02BAQ5K7/p1674663208140039)

cc @damassi

gkartalis:
pushed some changes that might fix [this](https://github.com/artsy/eigen/pull/8055#discussion_r1086845146), might need to test the deeplinking though

src/app/Scenes/Home/Home.tsx
```diff
@@ -91,8 +91,9 @@ const Home = (props: Props) => {
 
   const prefetchUrl = usePrefetch()
 
+  const what = useFeatureFlag("AREnableESOnlySearch")
```
damassi, line 94: 👀 on this variable name

gkartalis: woops, it is still a wip will change it, was playing around with the preloaded hook

gkartalis: 😅

src/app/AppRegistry.tsx
gkartalis: okay, @dimatretyak thank you for the proactive check ❤️ decided on the thread above to add the lint ignore to this one for now

dimatretyak: you're welcome ❤️

src/app/routes.ts
```diff
@@ -117,7 +117,9 @@ function getDomainMap(): Record<string, RouteMatcher[] | null> {
     addRoute("/", "Home"),
     addRoute("/sales", "Sales"),
     addRoute("/map", "NewMap"),
-    addRoute("/search", "Search"),
+    // TODO: this should be replaced with the new route when ES only Search is ready
+    addRoute("/search", "SearchSwitchContainer"),
+    addRoute("/oldsearch", "Search"),
```
dimatretyak, line 122: **question**: maybe it would be better to leave the old name, and add the prefix `new` for the new one? Something like this

```javascript
addRoute("/newsearch", "SearchSwitchContainer"),
addRoute("/search", "Search"),
```

dimatretyak, line 122: **question**: why should we leave the `oldsearch` path? Sorry if this is a stupid question 😅

dimatretyak: maybe these changes will be enough?

```javascript
useEffect(() => {
    !isESOnlySearchEnabled && prefetchUrl("search") // your changes
    prefetchUrl("my-profile")
    prefetchUrl("inbox")
    prefetchUrl("sales")
  }, [])
```

dimatretyak: https://user-images.githubusercontent.com/3513494/214636790-f1b59683-4697-4482-bb24-4a776adfeb8a.mp4

gkartalis: huh, I think you are right 🤔 

I thought I should avoid 
`Search: reactModule(SearchSwitchContainer, { isRootViewForTabName: "search" }, [SearchScreenQuery]),`
for some reason but this might work 👌

gkartalis: +++ You are! Thanks Dima, I thought initially that I should avoid that due to not having the query on the level of the `SearchSwitchContainer` but it works perfectly! 

Will also remove `!isESOnlySearchEnabled && prefetchUrl("search")` since it is not needed

dimatretyak: > Will also remove !isESOnlySearchEnabled && prefetchUrl("search") since it is not needed

don't delete it. this check is needed

dimatretyak: Since we want to [prefetch](https://github.com/artsy/eigen/blob/main/src/app/utils/queryPrefetching.ts#L76) the query for the old `search` screen

dimatretyak: For example:
- when `AREnableESOnlySearch` ff is **disabled**
  - Display the old Search screen with Algolia
  - Prefetch query for this screen in `Home.tsx`
- when `AREnableESOnlySearch` ff is **enabled**
  - Display the new Search screen with ElasticSearch 
  - Ignore prefetch query in `Home.tsx`

dimatretyak: another question came up: would we want to prefetch some query for the new screen with ElasticSearch in the future?

gkartalis: So in my mind, I have the following things:

I am thinking to leave the prefetch('search') as it is now. The reason for this is because this is a boilerplate feature and we would want to prefetch the query all the time since we will not release like that. 

When we are `readyForRelease: true` we can move the initial query on the Switch and this is gonna be the one that is preloaded and it will include the top-level things of both screens (search, search2), we can use an approach like we do in the[ artwork grid](https://github.com/artsy/eigen/blob/main/src/app/Components/ArtworkGrids/InfiniteScrollArtworksGrid.tsx#L474) that avoids fetching something according to a variable (this var could be the feature flag?)

How do the above sound to you? 🤔

src/app/AppRegistry.tsx
```diff
@@ -218,6 +219,7 @@ const InnerPageWrapper: React.FC<PageWrapperProps> = ({
 
   let isVisible = viewProps.isVisible
   if (!ignoreTabs) {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
```
gkartalis, line 222: context on the why we added this comment can be found [here](https://artsy.slack.com/archives/C02BAQ5K7/p1674666001128209?thread_ts=1674663208.140039&cid=C02BAQ5K7)

src/app/AppRegistry.tsx
```diff
@@ -483,7 +485,9 @@ export const modules = defineModules({
     SellWithArtsyHomeScreenQuery,
   ]),
   SalesNotRootTabView: reactModule(SellWithArtsy),
-  Search: reactModule(SearchScreen, { isRootViewForTabName: "search" }, [SearchScreenQuery]),
+  Search: reactModule(SearchSwitchContainer, { isRootViewForTabName: "search" }, [
+    SearchScreenQuery,
+  ]),
```
gkartalis, line 490: This might be enough for now, this is how it works, pre-fetches the search query (search) always

- with `AREnableESOnlySearch: true`: [video](https://www.loom.com/share/8af9296c70204a15b2f772defd56327a)
- with `AREnableESOnlySearch: false`: [video](https://www.loom.com/share/bc42d96a329f46829863bcf5c9b3599c)

dimatretyak (approved):
LGTM 👍

PR #8061, opened 2023-01-25
chore(scripts): Add `yarn navigate-to-route` script

damassi:
### Description

This adds a helpful new command to our `package.json` scripts so that we can navigate directly to a route in eigen:

```
yarn navigate-to-route /artist/andy-warhol
```

cc @artsy/mobile-platform

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "flip_table_extreme": "./scripts/flip-table-extreme",
     "generate-cities-cache": "node scripts/generate-cities-cache.js",
     "generate-cities-objc": "node scripts/generate-cities-objc.js",
+    "navigate-to-route": "./scripts/navigate-to-route",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -31,6 +31,7 @@
     "flip_table_extreme": "./scripts/flip-table-extreme",
     "generate-cities-cache": "node scripts/generate-cities-cache.js",
     "generate-cities-objc": "node scripts/generate-cities-objc.js",
+    "go-to-route": "./scripts/go-to-route",
```
brainbicycle, line 34: nice!!! what do you think about `navigate` as a name instead? just to be more consistent with the function in the code. Love the tool though

damassi: Yup, i can fix 👍 i had some qs about that

PR #8066, opened 2023-01-26
chore: increase app version

MounirDhahri:
### Description
Set app version to **8.7.0**

cc @brainbicycle

#nochangelog


MounirDhahri:
Deploying betas again 🚧

PR #8077, opened 2023-01-27
fix: Refetch artwork after editing

olerichter00:
This PR resolves [CX-3357] 

### Description

Refetch artwork after editing

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-3357]: https://artsyproduct.atlassian.net/browse/CX-3357?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MounirDhahri (commented):
Does returning the artwork as part of the mutation not fix this?

olerichter00:
> Does returning the artwork as part of the mutation not fix this?

We already do it and it doesn't work :/ I've tried to return the required fields directly in the mutation fragment but still the same result.

lordkiz (approved):
Thank you @olerichter00 . I think this is even a better approach - as long as we do not mind the necessary repeat request. Nice work 🙏

olerichter00:
Figuring out why the mutation doesn't change the values directly in the Relay store (like how it works when you save an artwort, for example) would be a great topic to explore at a Future Friday.

MounirDhahri:
agree - we can look into refactoring those screens and the logic there soon

PR #8100, opened 2023-01-31
chore: track Sentry screen navigation

MounirDhahri:
### Description

Screenshot 2023-01-31 at 08 44 46

After noticing the above error and some worrying errors inside Sentry ([here](https://sentry.io/organizations/artsynet/issues/3389526194/?environment=production&project=5867225&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=14d)), I wanted to know which screens this error is happening at. Unfortunately, it turned out that the old tracking was broken and it was also not capturing all screens (for example if you switch tabs.). This PR adds tracking on Intent (when the `navigate` method is used), and also after successful navigation. 

Screenshot 2023-01-31 at 09 40 55

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/metaflags.example.json
+++ b/metaflags.example.json
@@ -3,10 +3,11 @@
   "logAction": true,
   "logDatadog": true,
   "logEventTracked": true,
+  "logNavitation": true,
   "logNotification": true,
   "logOperation": true,
   "logPrefetching": true,
+  "logQueryPath": true,
   "logRelay": true,
-  "logRoute": true,
   "logRunningRequest": true
 }
--- a/metaflags.example.json
+++ b/metaflags.example.json
@@ -3,7 +3,7 @@
   "logAction": true,
   "logDatadog": true,
   "logEventTracked": true,
-  "logNavitation": true,
+  "logNavigation": true,
   "logNotification": true,
   "logOperation": true,
   "logPrefetching": true,
--- a/src/app/Scenes/Home/Home.tsx
+++ b/src/app/Scenes/Home/Home.tsx
@@ -122,7 +122,6 @@ const Home = (props: Props) => {
       title: "New Works for You",
       type: "newWorksForYou",
       data: newWorksForYou,
-      prefetchUrl: "/new-for-you",
     },
     {
       title: "",
--- a/src/app/system/navigation/ModalStack.tsx
+++ b/src/app/system/navigation/ModalStack.tsx
@@ -4,8 +4,10 @@ import {
   createStackNavigator,
   StackCardStyleInterpolator,
 } from "@react-navigation/stack"
+import { Severity, addBreadcrumb } from "@sentry/react-native"
 import { AppModule, modules } from "app/AppRegistry"
 import { __unsafe_mainModalStackRef } from "app/NativeModules/ARScreenPresenterModule"
+import { logNavigation } from "app/utils/loggers"
 import { Platform } from "react-native"
 import { NavStack } from "./NavStack"
 import { useReloadedDevNavigationState } from "./useReloadedDevNavigationState"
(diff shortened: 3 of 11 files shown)


metaflags.example.json
```diff
@@ -3,10 +3,11 @@
   "logAction": true,
   "logDatadog": true,
   "logEventTracked": true,
+  "logNavitation": true,
```
dimatretyak, line 6: **chore**: I think there is a small typo here and you meant `logNavigation`

src/app/system/navigation/NavStack.tsx
```diff
@@ -115,6 +117,22 @@ export const NavStack: React.FC<{
...
+              } `
+            )
+          }
+
+          addBreadcrumb({
+            message: `navigated to ${params.moduleName}`,
+            category: "navigation",
+            data: { ...params },
+            level: Severity.Info,
+          })
```
dimatretyak, line 134: **question**: I noticed that `console.log` and `addBreadcrumb` are called 2 times during navigation. maybe it would be better if we use `onStateChange` for `NavigationContainer` in `ModalStack` for these purposes? For example, like [here](https://reactnavigation.org/docs/screen-tracking/)

https://user-images.githubusercontent.com/3513494/215747045-c5fb052f-f27f-45af-baea-733e543d9ab6.mp4

dimatretyak: Something like this

```javascript
<NavigationContainer
  ref={__unsafe_mainModalStackRef}
  initialState={initialState}
  onStateChange={() => {
    const currentRoute = __unsafe_mainModalStackRef.current?.getCurrentRoute()

    if (currentRoute) {
      const params = currentRoute.params

      if (__DEV__ && logNavitation) {
        console.log(
          `navigated to ${params.moduleName} ${
            params.props ? JSON.stringify(params.props) : ""
          } `
        )
      }

      addBreadcrumb({
        message: `navigated to ${params.moduleName}`,
        category: "navigation",
        data: { ...params },
        level: Severity.Info,
      })
    }
  }}
>...</NavigationContainer>
```

dimatretyak (dismissed):

metaflags.example.json
MounirDhahri: good catch. thanks

src/app/system/navigation/NavStack.tsx
MounirDhahri: That's a better approach. I will change. Thanks Dima

src/app/Scenes/Home/Home.tsx
```diff
@@ -122,7 +122,6 @@ const Home = (props: Props) => {
       title: "New Works for You",
       type: "newWorksForYou",
       data: newWorksForYou,
-      prefetchUrl: "/new-for-you",
```
MounirDhahri, line 125: Removing prefetching the new works for you here since it's already in store as part of the rail.

MounirDhahri: cc @brainbicycle this was causing the sentry issue since it's not specifying a variant for the new works for you model

src/app/utils/queryPrefetching.ts
```diff
@@ -46,12 +46,19 @@ const prefetchQuery = async (query: GraphQLTaggedNode, variables?: Variables) =>
...
+      networkCacheConfig: { force: true },
+    }).toPromise()
+    // this will retain the result in the relay store so it's not garbage collected.
+    environment.retain(operation)
+  } catch (error) {
+    // We don't want to throw an error here because we don't want to block the user from navigating to the page.
+    // We still want to log the error so we can investigate it.
+    if (__DEV__) {
+      console.log(`Prefetching query failed: ${error}`)
+    }
```
MounirDhahri, line 60: @brainbicycle no need for this in Sentry

src/app/system/navigation/ModalStack.tsx
```diff
@@ -21,7 +23,32 @@ const Stack = createStackNavigator()
...
+
+          if (__DEV__ && logNavigation) {
+            console.log(
+              `navigated to ${params.moduleName} ${
+                params.props ? JSON.stringify(params.props) : ""
+              } `
+            )
+          }
+
+          addBreadcrumb({
```
brainbicycle, line 43: https://github.com/artsy/eigen/blob/9c8f9177b9c5daffaa99bd6531abb768f6511918/src/app/system/navigation/navigate.ts#L145 this is where we track tab taps in analytics, maybe it makes sense to track the tab navigate in the same place?

brainbicycle: or if the other approach captures both navigate calls that switch tabs and others it is fine as long as there is no double tracking

MounirDhahri: > if the other approach captures both navigate calls that switch tabs and others it is fine as long as there is no double tracking

Yes, the call in ModalStack is capturing switch tabs and normal navigation events. It's not being double tracked

The main reason I didn't update the logic there is the fact that we are using `tracks.tabScreenView` for analytics only within `switchTab` and I intended my changes only for logging purporses. If we move tracking tabs there, we should probably also move screen tracking also to the same place as well.

brainbicycle: screenTracking happens in useEffects in our screen components I believe but tabs render differently so can't be tracked there. but anyways if there is no double tracking it is fine.

MounirDhahri: great, thanks Brian for the quick feedback 🙌

PR #8107, opened 2023-01-31
fix: load asdf from brew prefix + /libexec + /asdf.sh

erikdstock:
This PR resolves an issue with building the app when node versions are managed with a brew-installed asdf.
It would be good to have another asdf-user test this. It may also require re-installing the package.

For reference, the oh-my-zsh plugin loads asdf [here](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/asdf/asdf.plugin.zsh#L11-L17) from the `/libexec` subdirectory. I suppose we should check that the non-brew method is similarly the same. At first glance, it appears to be the same.

### DIFF
--- a/patches/react-native+0.66.5.patch
+++ b/patches/react-native+0.66.5.patch
@@ -142,9 +142,9 @@ index c498ba6..11338d0 100755
 +if [[ -f "$HOME/.asdf/asdf.sh" ]]; then
 +  # shellcheck source=/dev/null
 +  . "$HOME/.asdf/asdf.sh"
-+elif [[ -x "$(command -v brew)" && -f "$(brew --prefix asdf)/asdf.sh" ]]; then
++elif [[ -x "$(command -v brew)" && -f "$(brew --prefix asdf)/libexec/asdf.sh" ]]; then
 +  # shellcheck source=/dev/null
-+  . "$(brew --prefix asdf)/asdf.sh"
++  . "$(brew --prefix asdf)/libexec/asdf.sh"
 +fi
 +
 +# Set up volta if present


pvinis (changes requested):
im not sure whats up, but please leave everything else as it was in that file. its a diff file, so even silly extra spaces should be there, otherwise it doesn't work as a diff.

patches/react-native+0.66.5.patch
```diff
@@ -142,9 +142,9 @@ index c498ba6..11338d0 100755
 +if [[ -f "$HOME/.asdf/asdf.sh" ]]; then
 +  # shellcheck source=/dev/null
 +  . "$HOME/.asdf/asdf.sh"
-+elif [[ -x "$(command -v brew)" && -f "$(brew --prefix asdf)/asdf.sh" ]]; then
++elif [[ -x "$(command -v brew)" && -f "$(brew --prefix asdf)/libexec/asdf.sh" ]]; then
 +  # shellcheck source=/dev/null
-+  . "$(brew --prefix asdf)/asdf.sh"
++  . "$(brew --prefix asdf)/libexec/asdf.sh"
 +fi
```
erikdstock, line 148: From the [oh-my-zsh plugin](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/asdf/asdf.plugin.zsh#L11-L17) - I think this matches our case.

image

damassi:
@brainbicycle - this impacts you as well

pvinis:
we chatted with Erik, I'll fix up this PR later tonremov reformatting.

pvinis (approved):

pvinis:
trying it out locally, and ill merge if it works first try.

PR #8111, opened 2023-01-31
ci(betterer): Remove betterer, re-add Danger

damassi:
### Description

Due to some developer friction that we've yet to work out, reverting Betterer and adding back Danger. See [this convo](https://artsy.slack.com/archives/C02NAJ2CGLW/p1675190552099359) for more info. 

cc @artsy/mobile-platform


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  We are trying to migrate away from moment towards `luxon`, but found moment imports in the following new files:
- `dangerfile.ts`
See [docs](https://moment.github.io/luxon/api-docs/index.html).
  
  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 386b469aec36f92e5dd5759fe22db7278795e649
</p>

.circleci/config.yml
```diff
@@ -253,17 +253,17 @@ jobs:
...
           command: yarn update-metaphysics
 
-  check-code-improvements:
+  check-code:
     executor:
       name: node/default
       tag: "16.18.0"
     steps:
       - checkout
-      - generate-checksums
```
pvinis, line 262: is this not needed anymore? i feel like this is needed? but maybe the node orb handles that well, in which case, we should use that in all places instead of the install-node-modules.

damassi: Nope, circle ci node orb handles all the caching for us 👍

damassi: (And yup, was gonna refactor that out at some point to slim down this file a bit)

PR #8142, opened 2023-02-02
feat: Artists result screen tab

rajsam003:
This PR resolves [GRO-1272] 

### Acceptance criteria

- [x] User sees artist rails of artists based on the artworks they saved 
- [x] User sees follow artist button 
- [x] User sees bio - limit after the first sentence on mobile
- [x] User sees an artwork rail with the first 25 artworks by the artist 

image

### PR Checklist

- [x] I tested my changes on **iOS** / **Android**.
- [x] I added screenshots or videos to illustrate my changes.
- [ ] I added Tests and Stories for my changes.
- [ ] I added an [app state migration].
- [x] I hid my changes behind a [feature flag].
- [ ] I have prefixed changes that need to be tested during a release QA with **[NEEDS EXTERNAL QA]** on the changelog.

### To the reviewers 👀

- [x] I would like **at least one** of the reviewers to run this PR on the simulator or device.

<details><summary>Changelog updates</summary>

#### Cross-platform user-facing changes

-

#### iOS user-facing changes

-

#### Android user-facing changes

-

#### Dev changes

-

</details>

Need help with something? Have a look at our [docs], or get in touch with us.

[app state migration]: ../blob/main/docs/adding_state_migrations.md
[feature flag]: ../blob/main/docs/developing_a_feature.md
[docs]: ../blob/main/docs/README.md

[GRO-1272]: https://artsyproduct.atlassian.net/browse/GRO-1272?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #8143, opened 2023-02-02
chore: run yarn on main

MounirDhahri:
### Description

- After running yarn on main, I got a different yarn.lock file

#nochangelog


MounirDhahri:
Mergings this as I see it also occurring in other PRs like this https://github.com/artsy/eigen/pull/8142 cc @rajsam003

PR #8152, opened 2023-02-03
tests: camera mock for tests, and less annoying eslint rule

pvinis:
This PR resolves [] 

### Description

I noticed we get this camera errors on tests, so i just added an empty mock for now. Also, while debugging, imports were removed, and we don't like that. The rule will mark the unused imports but it will not "fix" them.

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- mocks and eslint rules

</details>

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -11,6 +11,7 @@ module.exports = {
     "react-hooks",
     "testing-library",
     "unused-imports",
+    "no-autofix",
   ],
   extends: [
     "eslint:recommended",
@@ -52,7 +53,8 @@ module.exports = {
     "import/no-duplicates": ERR,
     "react/jsx-curly-brace-presence": ERR,
     "react-hooks/rules-of-hooks": ERR,
-    "unused-imports/no-unused-imports": ERR,
+    "unused-imports/no-unused-imports": OFF, // look below
+    "no-autofix/unused-imports/no-unused-imports": ERR,
     /**
      * Warnings
      */
--- a/package.json
+++ b/package.json
@@ -280,6 +280,7 @@
     "eslint-import-resolver-typescript": "3.5.3",
     "eslint-plugin-import": "2.27.5",
     "eslint-plugin-jest": "27.2.1",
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -280,6 +280,7 @@
     "eslint-import-resolver-typescript": "3.5.3",
     "eslint-plugin-import": "2.27.5",
     "eslint-plugin-jest": "27.2.1",
+    "eslint-plugin-no-autofix": "1.2.3",
```
gkartalis, line 283: 🔥

src/setupJest.ts
```diff
@@ -92,6 +92,8 @@ jest.mock("@react-native-async-storage/async-storage", () => mockAsyncStorage)
 
 jest.mock("@react-native-community/netinfo", () => mockRNCNetInfo)
 
+jest.mock("react-native-vision-camera", () => {})
```
gkartalis, line 95: omg, was this the thing that was complaining all the time???

gkartalis (approved):

.eslintrc.js
```diff
@@ -52,7 +53,8 @@ module.exports = {
     "import/no-duplicates": ERR,
     "react/jsx-curly-brace-presence": ERR,
     "react-hooks/rules-of-hooks": ERR,
-    "unused-imports/no-unused-imports": ERR,
+    "unused-imports/no-unused-imports": OFF, // look below
+    "no-autofix/unused-imports/no-unused-imports": ERR,
```
damassi, line 57: @pvinis - def disagree strongly with these two additions. It might be an issue with your refactor work (which makes sense since you're literally operating on thousands and thousands of imports), but this kind of thing is only helpful for general product dev. We should file a JIRA ticket and re-enable once done.

src/setupJest.ts
damassi: I tried this a while back and it didn't fix anything 🙁, maybe I didn't add the extra function at the end?

PR #8172, opened 2023-02-09
fix: auction navigation android issue

gkartalis:
This PR resolves [] 

### Description

On android when opening the app and trying to navigate to the `Sale` page we were always getting an error on the first try, and on the second try we were accessing the page correctly.

The reason for that was that the relay and suspense weren't configured correctly for this screen, this pr fixes that.

|Before|After|
|---|---|
|https://www.loom.com/share/9330260275f34efdbddd6bd565ca7104|https://www.loom.com/share/c96813f55e6a4b748ac7960e38749246|

### PR Checklist

### To the reviewers 👀

- 

#### iOS user-facing changes

-

- fix auction navigation android issue - gkartalis

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Scenes/Sales/Sales.tsx
+++ b/src/app/Scenes/Sales/Sales.tsx
@@ -30,7 +30,15 @@ export const SalesScreenQuery = graphql`
   }
 `
 
-export const Sales: React.FC<{ data: SalesQuery["response"] }> = ({ data }) => {
+export const Sales: React.FC = () => {
+  const data = useLazyLoadQuery<SalesQuery>(
+    SalesScreenQuery,
+    {},
+    {
+      fetchPolicy: "store-and-network",
+      networkCacheConfig: { force: true },
+    }
+  )
   const [isRefreshing, setIsRefreshing] = useState(false)
 
   // using max_value because we want CurrentlyRunningAuctions & UpcomingAuctions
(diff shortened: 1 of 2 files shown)


rajsam003 (approved):
Thanks for the change

src/app/Scenes/Sales/Sales.tsx
```diff
@@ -89,14 +97,6 @@ export const Sales: React.FC<{ data: SalesQuery["response"] }> = ({ data }) => {
 }
 
 export const SalesQueryRenderer = () => {
```
dimatretyak, line 99: **suggestion**: rename this from `SalesQueryRenderer` to `SalesScreen`

dimatretyak (approved):

src/app/Scenes/Sales/Sales.tsx
gkartalis: good point, will bring those changes to main as well with this ☝️ thanks @dimatretyak

src/app/Scenes/Sales/Sales.tsx
```diff
@@ -30,7 +30,15 @@ export const SalesScreenQuery = graphql`
...
 `
 
-export const Sales: React.FC<{ data: SalesQuery["response"] }> = ({ data }) => {
+export const Sales: React.FC = () => {
+  const data = useLazyLoadQuery<SalesQuery>(
+    SalesScreenQuery,
+    {},
+    {
+      fetchPolicy: "store-and-network",
+      networkCacheConfig: { force: true },
```
MounirDhahri, line 39: **nitpick(no changes needed):**given that the fetch policy is set to `store-and-network` I suspect that this will ever happen since the `force` is set to `true` inside `networkCacheConfig`. In similar cases, I would suggest to go with `network` as a fetch policy

src/app/Scenes/Sales/Sales.tsx
```diff
@@ -89,14 +97,6 @@ export const Sales: React.FC<{ data: SalesQuery["response"] }> = ({ data }) => {
 }
 
 export const SalesQueryRenderer = () => {
-  const data = useLazyLoadQuery<SalesQuery>(
```
MounirDhahri, line 92: Good catch!

PR #8177, opened 2023-02-09
feat(pageable-artworks): Add ability to swipe between artwork screens within collections

damassi:
This PR resolves [MOPLAT-600]

> ⚠️ Still need to add some tests, but looking for an initial review beforehand!

### Description

This adds the ability to swipe between artworks that are present within various contexts such as rails and grids. I've gone through and updated all the obvious surfaces to support this but Eigen is large so likely something was missed somewhere. We can surface it with some QA. 

https://user-images.githubusercontent.com/236943/218222383-f65ac649-a17d-47d9-ae7b-bc3d2dcf9bec.mp4

It's easy to add a pageable-enabled screen (which can, in the future, be extended to support other _types_ of screens such as swiping between artists, etc): 

```tsx
const SomeCollection = ({ artworks }) => {
  const { enablePageableArtworks } = usePageableArtworks({ artworks }) 
  
  return (
    <>
      {artworks.map(artwork => {
        return <Artwork onPress={enablePageableArtworks} />
      })}
    </>
  )
}
```

Thanks to @dimatretyak for providing the initial swipeable screen implementation which has been brought over from Energy and lightly extended to support Eigen's environment. 

See inline comments for a walk through on how it works. 

cc @artsy/mobile-platform 

TODO: 
- [ ] Add some tests 
- [ ] Investigate weird pixelated placeholders

### PR Checklist

### To the reviewers 👀

- Added ability to swipe between artworks in various collections 

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[MOPLAT-600]: https://artsyproduct.atlassian.net/browse/MOPLAT-600?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/Components/ArtworkGrids/ArtworkGridItem.tsx
```diff
@@ -33,7 +34,7 @@ import { LotProgressBar } from "./LotProgressBar"
 
 const SAVE_ICON_SIZE = 22
 
-export interface ArtworkProps {
+export interface ArtworkProps extends PageableArtworksEvents {
```
damassi, line 37: If we need to extend an existing component with an pageable click event can import this

src/app/Components/ArtworkGrids/ArtworkGridItem.tsx
```diff
@@ -163,11 +165,14 @@ export const Artwork: React.FC<ArtworkProps> = ({
 
   const handleTap = () => {
     if (onPress) {
-      return onPress(artwork.slug)
+      onEnablePageableArtworks?.()
```
damassi, line 168: Which then allows us to mix it in with different click handlers

src/app/Components/ArtworkRail/ArtworkRail.tsx
```diff
@@ -65,6 +68,7 @@ export const ArtworkRail: React.FC<ArtworkRailProps> = ({
           hidePartnerName
           hideArtistName={hideArtistName}
           onPress={() => {
+            enablePageableArtworks()
```
damassi, line 71: Or if the click handler is already defined external to the component, can just invoke the enable function

src/app/Components/PageableScreensView/PageableLazyScreen.tsx
```diff
@@ -0,0 +1,26 @@
+import { PageableScreenEntity } from "app/Components/PageableScreensView/PageableScreensContext"
```
damassi, line 1: The following `PageableScreensView` code has been pulled from Energy (but renamed from Scrollable to Pageable to better reflect the underlying concept / lib ("pages", `react-native-pageable`)

src/app/Scenes/Sale/Components/SaleArtworkList.tsx
```diff
@@ -77,6 +81,7 @@ export const SaleArtworkListContainer = createFragmentContainer(SaleArtworkList,
       edges {
         node {
           id
+          slug
```
damassi, line 84: `usePageableArtworks` expects `artworks` that contain `slugs`, so a few spots had to be updated.

src/app/routes.ts
```diff
@@ -8,102 +8,6 @@ import { parse as parseQueryString } from "query-string"
 import { Platform } from "react-native"
 import { GraphQLTaggedNode } from "relay-runtime"
 
-export function matchRoute(
```
damassi, line 11: I just moved all of this less-critical stuff to the bottom, after the route list

src/app/store/GlobalStoreModel.ts
```diff
@@ -36,20 +37,21 @@ interface GlobalStoreStateModel {
...
+  artworkSubmission: SubmissionModel
+  auth: AuthModel
   bottomTabs: BottomTabsModel
-  search: SearchModel
+  devicePrefs: DevicePrefsModel
   myCollection: MyCollectionModel
-  auth: AuthModel
-  toast: ToastModel
+  native: NativeModel
+  pageable: PageableModel
```
damassi, line 47: Only addition; the rest is alpha

src/app/system/navigation/BackButton.tsx
```diff
@@ -21,6 +23,12 @@ export const BackButton: React.FC<{
       }).start()
     }
   }, [show])
+
+  const handleBackButtonPress = () => {
+    GlobalStore.actions.pageable.resetPageableSlugs()
```
damassi, line 28: On back, reset the slugs. There might be other back button locations? I covered the obvious spots.

src/app/utils/usePageableArtworks.tsx
```diff
@@ -0,0 +1,23 @@
...
+  const slugs = props.artworks.map((artwork) => artwork.slug)
+
+  const enablePageableArtworks = () => {
+    GlobalStore.actions.pageable.setPageableSlugs(slugs)
+  }
+
+  return {
+    enablePageableArtworks,
+  }
+}
```
damassi, line 23: When / if we're ready to add different types to the pager view ("artist", "auction", etc), we can add different hooks that are similar to the one above; eg, `usePageableArtists`.

src/app/AppRegistry.tsx
```diff
@@ -159,7 +158,37 @@ interface ArtworkProps {
...
+        />
+      )
+      // If not within the context of an artwork collection, just render the
+      // individual artwork.
+    } else {
+      return <ArtworkQueryRenderer {...props} />
+    }
+  },
+  (prevProps, nextProps) => {
+    return prevProps.artworkID === nextProps.artworkID
```
damassi, line 189: Really ensure we're not rerendering things unnecessarily

src/app/Components/ArtworkRail/ArtworkRail.tsx
```diff
@@ -43,6 +44,8 @@ export const ArtworkRail: React.FC<ArtworkRailProps> = ({
   showSaveIcon = false,
   trackingContextScreenOwnerType,
 }) => {
+  const { enablePageableArtworks } = usePageableArtworks<{ slug: string }>({ artworks })
```
damassi, line 47: This union type on `artworks` was being funny /shrug

gkartalis:
I might be missing some context so before reviewing had one question: I see from the videos that on pageable artworks there are not Commercial CTA's (attached below an example of the CTAs)

damassi:
Oh interesting @gkartalis - i'm not doing anything different in terms of rendering the artworks (minus being a little more conservative in terms of preventing rerenders. Maybe its below the fold in the screen-cap? In any case I'll give that a look.

gkartalis:
tagging also @dimatretyak here who implemented this behavior for providing more context on the functionality

src/app/store/GlobalStoreModel.ts
dimatretyak: **blocking**: unfortunately, with this approach, there is a chance of overwriting previous `slugs` when navigating from one artwork screen to another

#### Demo
https://user-images.githubusercontent.com/3513494/218545328-08f6f426-b559-45c4-a1f3-c43855f88264.mp4

dimatretyak: maybe we can use navigation params for is this case?

brainbicycle (commented):
The api looks really slick and awesome that this is working so well already. One question about storing the slugs in global state but all looks good to me.

src/app/store/PageableModel.ts
```diff
@@ -0,0 +1,17 @@
+import { action, Action } from "easy-peasy"
+
+export interface PageableModel {
+  pageableSlugs: string[]
+  setPageableSlugs: Action<this, string[]>
+  resetPageableSlugs: Action<this>
+}
```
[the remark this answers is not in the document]
damassi: We need to find a way to lift the state all the way up contextually, which can happen from various places in the app -- from the grid, from a rail, etc. AFAIK, without having a bit of global state there'd be no way to hoist the slugs outside of the overarching context from where the pageable is initialized. 

Also, I'm interested in expanding this model a bit with other categories of pages at some point (for example, artist -> artist -> artist, article -> article -> article, fair -> fair -> fair), so this is a first step.

src/app/system/navigation/BackButton.tsx
damassi: @dimatretyak - would you mind breaking down exactly whats going on in bullet points just so I can follow a bit better? Like, I kinda sorta see whats happening in the screencap but am not exactly following due to my newness in eigen and not knowing all of its behaviors.

src/app/store/GlobalStoreModel.ts
damassi: Oh I see whats happening! interesting. one can go forward as far as they'd like, but not backward. 🤔 🤔 welp, that significantly complicates things!

src/app/store/PageableModel.ts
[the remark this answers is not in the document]
brainbicycle: hmm maybe I am not understanding but all these navigate at some point to an artwork component right, so not quite getting why we can't pass slugs while navigating from whatever collection we are in and if there are multiple slugs show the Pageable component and if there aren't show the regular ArtworkQueryRenderer.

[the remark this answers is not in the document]
damassi: I see, so pass the slugs as additional props via `navigate`. That's certainly possible but not sure that its necessarily more ideal / simpler than hoisting it up a bit and having it exist as part of the overall app shell state. With Dima's comment up above that also makes me think that we'll need something a bit deeper in terms of managing pagable slugs.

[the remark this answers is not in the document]
damassi: I see, so pass the slugs as additional props via `navigate`. That's certainly possible but not sure that its necessarily more ideal / simpler than hoisting it up a bit and having it exist as part of the overall app shell state. With Dima's comment up above that also makes me think that we'll need something a bit deeper in terms of managing pagable slugs.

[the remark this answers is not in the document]
damassi: I see, so pass the slugs as additional props via `navigate`. That's certainly possible but not sure that its necessarily more ideal / simpler than hoisting it up a bit and having it exist as part of the overall app shell state. With Dima's comment up above that also makes me think that we'll need something a bit deeper in terms of managing pagable slugs.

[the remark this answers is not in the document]
brainbicycle: that way we can avoid having to worry about resetting state maybe at the cost of a slightly more complicated UI. or maybe I am missing why the state needs to live beyond the life of the component.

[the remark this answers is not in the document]
damassi: Ooh, actually, if slugs were passed on a per-instance artwork basis, maybe that would _solve_ the problem Dima mentions above, while also the reset thing. Without the global state, when the previous artwork is initialized (via back, or restarting app, etc) those props would be preserved and as such "just work". Yup I think you're on to something, lemme experiment.

src/app/system/navigation/BackButton.tsx
dimatretyak: ### What happens in the video above:
1. I navigate to the artwork screen
2. Artwork slugs are saved in the device storage
3. Removed the app from background processes
4. Launch application again. the app will start completely from scratch (so-called "cold" boat), since I removed the application from background processes in the previous step
5. If you pay attention to the logs, you may notice that artwork slugs will be available immediately (although we haven't navigate to the artwork page yet). this is due to the fact that artwork slugs were saved to the device storage in step 2

You can read a little more about the different state of the application [here](https://reactnative.dev/docs/appstate)

damassi: Thank you, this makes perfect sense 👍

src/app/store/PageableModel.ts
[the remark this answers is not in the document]
damassi: This also solves for the persisted slug issue Dima notes above, lots of birds with one stone, this one! 🙏

src/app/system/navigation/BackButton.tsx
dimatretyak: > Also the user can return to the previous screen using the back button in the bottom navigation bar on android

This happens when the user clicks this button in the screenshot
Image 2023-02-13 22-17-04

src/app/store/GlobalStoreModel.ts
dimatretyak: Small example of the suggested changes: https://github.com/artsy/eigen/commit/c61255aae5db51a59aa8ba3a95d1bc67dc5f1da9

damassi: Amazing 🙏 this solves a lot of different issues. Thank you for the example! I wasn't quite sure how to do this.

src/app/Components/PageableScreensView/PageableScreensView.tsx
```diff
@@ -0,0 +1,52 @@
...
+  return (
+    <PageableScreensContext.Provider value={context}>
+      <PagerView
+        style={{ flex: 1 }}
+        overScrollMode="never"
+        onPageSelected={handlePageSelected}
+        initialPage={activeScreenIndex}
+      >
+        {screens.map((screen, screenIndex) => {
+          const shouldRender = Math.abs(activeScreenIndex - screenIndex) <= prefetchScreensCount
```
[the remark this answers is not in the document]
MounirDhahri: I was afraid that the [memory usage](https://reactnative.dev/docs/profiling) would explode after lots of navigation and multiple artwork screens opened. I think this helps a lot

src/app/system/navigation/BackButton.tsx
MounirDhahri: same comment here, I would suggest listening to [navigation events](https://reactnavigation.org/docs/navigation-events/) here to reset the model state. Also, do we need `pageableSlugs` to persist? maybe they can be part of the `sessionState`

src/palette/atoms/BackButton/BackButton.tsx
```diff
@@ -17,11 +18,16 @@ export const BackButton: React.FC<BackButtonProps> = ({
   containerStyle,
   hitSlop,
 }) => {
+  const handleOnBackPress = () => {
+    GlobalStore.actions.pageable.resetPageableSlugs()
```
MounirDhahri, line 22: Given that we might have different Artwork screens within the navigation stack. resetting the pageable slugs this way will reset it for all previous artwork screens and break the swipe gesture for previous screens. You can see the video for more details
https://user-images.githubusercontent.com/11945712/219325224-6c2ed876-4f79-462b-a646-25fe51e463ae.mov

src/app/store/PageableModel.ts
```diff
@@ -0,0 +1,17 @@
+import { action, Action } from "easy-peasy"
+
+export interface PageableModel {
```
MounirDhahri, line 3: **issue(non-blocker):** should we maybe add a migration for this - although it seems to be working fine without it, we have been adding migrations when declaring new models for a while

MounirDhahri (commented):
Really great work on very little time. I think with just a little bit of polishing, we can get this out. I have few things though that I would like to mention as well.
- I feel the loading skeleton is broken for some reason 🤷‍♂️ I don't know if it's related to your changes (doesn't seem like it)
- After some navigation, I see quite few dropped frames, I am testing on my simulator, and i believe the performance will be much better on device
- I would suggest to maybe hide the changes behind a feature flag, this can help us if we would like to switch this off remotely
- Since the back button is the same on the artwork page, it would be great at a follow up ticket to keep it sticky and only change the content of the screen



_Sorry for the late review, I was off then got sick then forgot to submit my review_

src/palette/atoms/BackButton/BackButton.tsx
MounirDhahri: Should we maybe have `pageableSlugs` by `context or initial artwork id`?

damassi:
Closing in favor of https://github.com/artsy/eigen/pull/8198

PR #8180, opened 2023-02-10
feat: custom eslint rule for preventing use of uselazyloadquery in same component with suspense

gkartalis:
This PR resolves [] 

### Description

A while ago we had some random crashes in eigen in two cases that we were using the `useLazyLoadQuery` hook inside a component that had also `Suspense` in them.

For that reason we wrote a custom eslint rule to prevent this from happening again.

Although I am not sure where is the best place for this to live.
We could have this only in eigen or we could create a different `artsy` custom rules repository that includes this and whatever else we come up with 🤷

In case you want to see how the rule works:

Screenshot 2023-04-19 at 10 31 00

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- custom eslint rule for preventing use of uselazyloadquery in same component with suspense - gkartalis

</details>


src/app/Scenes/Sales/Sales.tsx
```diff
@@ -29,7 +29,6 @@ export const SalesScreenQuery = graphql`
     }
   }
 `
```
gkartalis, line 31:
```suggestion
`

```

eslint-rules/index.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
```
damassi, line 1: Thoughts on moving this to 
```
src/system/devTools/eslint
```
?

From there we can decide where to move it in the future.

damassi (dismissed):
One non-blocking suggestion, looks great 👍

eslint-rules/index.js
gkartalis: Did a quick search for where custom eslint rules should live and seems like the standard is to keep them outside the `src/`  folder in order to prevent having weird issues when building/bundling since eslint is a devDependency 🤷.

So I am thinking to leave it as is 🤔

eslint-rules/no-uselazyloadquery-inside-suspense.js
```diff
@@ -0,0 +1,85 @@
...
+)
+
+module.exports = createRule({
+  name: "no-uselazyloadquery-inside-suspense",
+  meta: {
+    type: "problem",
+    hasSuggestions: true,
+    docs: {
+      description:
+        "In jsx, something like `{someString && <View />}` could produce the empty string (`\"\"`) which is falsy but still produces a value, and that make React Native crash, because it doesn't know how to handle a string that's not inside a `Text`.",
```
gkartalis, line 17:
```suggestion
        "When we use `useLazyLoadQuery` inside a component that has `<Suspense />` in react-native it causes a crash, this rule is for preventing that"
```

eslint-rules/index.js
damassi: Np!

PR #8196, opened 2023-02-14
Feat(CX-3376): Redesign SWA Landing Page [Part 1]

lordkiz:
This PR resolves [CX-3376] 

### Description
Redesign SWA Landing Page [Part 1]

#### iPhone
https://user-images.githubusercontent.com/18648835/218855509-62bfe3ed-9d07-4517-940a-53cd565a47ff.mov

#### Android
[swa-lp-1-android.webm](https://user-images.githubusercontent.com/18648835/218855694-7b78d9d2-c9cd-4ed1-9fa9-2064a5f7cb53.webm)

#### iPad

https://user-images.githubusercontent.com/18648835/218855862-f46eb20a-140c-4dd7-9aa8-73b22712e6b8.mov

### PR Checklist

### To the reviewers 👀

- Redesign SWA Landing Page [Part 1] - kizito

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-3376]: https://artsyproduct.atlassian.net/browse/CX-3376?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MounirDhahri (changes requested):
Niice LGTM. 
I have only one comment about the images size, I feel they're too big for what we need. I would suggest making them as small as possible since they will be part of the carousel, there is no need to have them very clear. I think we can get them to about 1mb maybe for 3x instead of 5mb

src/app/Scenes/SellWithArtsy/Components/SWALandingPageHeroImage.tsx
```diff
@@ -0,0 +1,34 @@
+import { Flex } from "@artsy/palette-mobile"
+import { isPad } from "app/utils/hardware"
+import { Image } from "react-native"
+import { useScreenDimensions } from "shared/hooks"
+
+export const SWALandingPageHeroImage = () => {
```
dariakoko, line 6: You can import the group of images as one image and use it in the Image component

src/app/Scenes/SellWithArtsy/Components/WaysWeSell.tsx
```diff
@@ -0,0 +1,70 @@
...
+      <Flex mx={2}>
+        <Text variant="lg" color={color("white100")} mb={1}>
+          Ways we sell your work
+        </Text>
+        <Text variant="xs" color={color("white100")} mb={2}>
+          We create a tailored strategy to find the optimal sales method for your artwork
+        </Text>
+      </Flex>
+      <FlatList
+        contentContainerStyle={{ marginLeft: 20 }}
```
dariakoko, line 36: **Suggestion:** let's use `const space = useSpace()`?

src/app/Scenes/SellWithArtsy/Components/WaysWeSell.tsx
```diff
@@ -0,0 +1,70 @@
...
+                // because these images are suffixed @1x, @2x etc we cannot `require` outside Image
+                // as this would break tests since in test env the module would not be found without
+                // the suffixes included literally
+                index === 0
+                  ? require("images/ways-we-sell-auction.png")
+                  : index === 1
+                  ? require("images/ways-we-sell-private-sales.png")
+                  : require("images/ways-we-sell-online-storefront.png")
+              }
+              style={{ maxWidth: width - 60 }}
```
dariakoko, line 53: **Suggestion:** and here as well

src/app/Scenes/SellWithArtsy/SellWithArtsyHome.tsx
```diff
@@ -82,6 +85,11 @@ export const SellWithArtsyHome: React.FC<SellWithArtsyHomeProps> = ({
 
             <Spacer y={4} />
 
+            {enableNewSWALandingPage && <Highlights />}
+            {enableNewSWALandingPage && <Spacer y={4} />}
+            {enableNewSWALandingPage && <WaysWeSell />}
+            {enableNewSWALandingPage && <Spacer y={4} />}
```
dariakoko, line 91: **Suggestion:** what do you think about using ff only once and, to avoid using Spacer multiple times, we can wrap it in `<Join /> `

Also, to avoid using Spacer multiple times, we can wrap it in `<Join /> `
```
{enableNewSWALandingPage && (
  <Join separator={<Spacer y={4}  />}>
      <Highlights />
      <WaysWeSell />
  </Join>
)}
```

dariakoko (commented):
Looks good!
I only left several suggestions

lordkiz:
> Niice LGTM.
> I have only one comment about the images size, I feel they're too big for what we need. I would suggest making them as small as possible since they will be part of the carousel, there is no need to have them very clear. I think we can get them to about 1mb maybe for 3x instead of 5mb

@MounirDhahri Ok. Just for the images in the carousel right?
I think we also have to revisit how the header looks on iPads. The design can be better.

src/app/Scenes/SellWithArtsy/SellWithArtsyHome.tsx
lordkiz: Thanks @dariakoko. That is cleaner. But Join will not append spacer below. I think we can fix all this when we are done with part 3

MounirDhahri:
totally agree

src/app/Scenes/SellWithArtsy/Components/SWALandingPageHeroImage.tsx
lordkiz: It appeared stretched to me, no matter resizeMode. Is there another way?

src/app/Scenes/SellWithArtsy/Components/WaysWeSell.tsx
lordkiz: We want to avoid useSpace for now until we have a stable palette-mobile.

src/app/Scenes/SellWithArtsy/Components/SWALandingPageHeroImage.tsx
dariakoko: Have you tried setting height and width and adding resizeMode property?

src/app/Scenes/SellWithArtsy/Components/WaysWeSell.tsx
dariakoko: Why can't we import it from palette?
I'd say it's still better to use useSpace here since it is something very basic and is used in a lot of places already

src/app/Scenes/SellWithArtsy/SellWithArtsyHome.tsx
dariakoko: Okay, as you feel comfortable.
Imo, it's easier to work with components like this, we are automatically avoiding situations when spaces are added to the child components. It's easier to maintain

src/app/Scenes/SellWithArtsy/Components/WaysWeSell.tsx
lordkiz: @dariakoko space is actively being worked on atm. And being replaced in various places in the app. See: https://artsy.slack.com/archives/C02NAJ2CGLW/p1676398423106229
By importing space from `palette` and not mobile-palette, we will end up with merge conflicts. I will replace these soon in later parts of the process.

src/app/Scenes/SellWithArtsy/Components/SWALandingPageHeroImage.tsx
lordkiz: I did. Another thing with all the images mashed in one image is we will usually need a higher quality single image. When you compress the single image to say 500kb and stretch it across the screen end to end, it begins to look blurry. But we can have each of the images compressed to say 150kb and position them without gaining too much app size and still maintain quality.

src/app/Scenes/SellWithArtsy/SellWithArtsyHome.tsx
lordkiz: Absolutely. I agree.

src/app/Scenes/SellWithArtsy/Components/SWALandingPageHeroImage.tsx
lordkiz: I just checked again. While using each image differently has better quality, the quality is negligible compared to the size savings of using a single image.
I will update this. 
Thanks @dariakoko

src/app/Scenes/SellWithArtsy/Components/WaysWeSell.tsx
MounirDhahri: That's a good point, I believe @pvinis is still working on it, I still believe though that we should not optimize for future work and instead we can still use `useSpace` from local `palette` and update the import once `palette-mobile` work is wrapped. I feel like updating the dependency as part of a merge conflict will take less time than keeping an eye on when the work is done, creating a Jira ticket, prioritizing it then following up with a PR (although it's probably one-line of code).

MounirDhahri (dismissed):
Thanks for making the updates. LGTM 🙌

src/app/Scenes/SellWithArtsy/Components/WaysWeSell.tsx
dariakoko: Thanks, @MounirDhahri 
Agree!

src/app/Scenes/SellWithArtsy/Components/SWALandingPageHeroImage.tsx
dariakoko: Thanks @lordkiz!

dariakoko (approved):
Thanks for the updates ⭐

PR #8198, opened 2023-02-14
feat(pageable): Add ability to page through artworks

damassi:
This PR resolves [MOPLAT-600]

### Description

This adds the ability to swipe between artworks that are present within various contexts such as rails and grids. I've gone through and updated all the obvious surfaces to support this but Eigen is large so likely something was missed somewhere. We can surface it with some QA.

Hidden behind a `AREnablePageableArtworkScreens` feature flag. 

### Use 

Pageable artwork screens are easy to add:

```tsx
const SomeRail = ({ someConnection }) => {
  const artworks = extractNodes(someConnection) 
  
  // Initialize hook with artworks that should be pageable
  const { navigateToPageableRoute } = useNavigateToPageableRoute({ artworks })
  
  return (
    <>
      {artworks.map(artwork => {
        return (
          <Artwork onPress={() => {
            // Then navigate. Uses the same underlying code as `navigate`, so 
            // additional props can be passed in a second argument. 
            navigateToPageableRoute(artwork.slug)} 
          />
        }
      }}
    </>
  )
}
```

### PR Checklist

### To the reviewers 👀

- Add ability to swipe between artwork screens in collections

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[MOPLAT-600]: https://artsyproduct.atlassian.net/browse/MOPLAT-600?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/developing_a_feature.md
+++ b/docs/developing_a_feature.md
@@ -68,7 +68,13 @@ First you need to add the new feature block in `src/app/store/config/features.ts
 The `description` property is what makes it possible to override the feature flag from the [Dev Menu](/dev_menu.md).
 
 We also need to add a flag in Echo, our remote feature flags configuration service.
-[Here](https://github.com/artsy/echo/commit/978a103e2c67a8010fabb2184f84aaef31d16f93) is an example PR for how to do that.
+[Here](https://github.com/artsy/echo/commit/978a103e2c67a8010fabb2184f84aaef31d16f93) is an example PR for how to do that. The general feature development flow around adding a flag and updating Echo is like so:
+
+- Add flag to eigen, `readyToRelease: false`
+- Add flag to echo, with `false`
+- Finish dev
+- Change eigen `readyToRelease: true`
+- When the app version is released, swap echo flag to `true` whenever we want users to get it
 
 After adding the echo key, and the PR is merged, update you local copy of echo by running `./scripts/update-echo`.
 
--- a/docs/developing_a_feature.md
+++ b/docs/developing_a_feature.md
@@ -70,10 +70,10 @@ The `description` property is what makes it possible to override the feature fla
 We also need to add a flag in Echo, our remote feature flags configuration service.
 [Here](https://github.com/artsy/echo/commit/978a103e2c67a8010fabb2184f84aaef31d16f93) is an example PR for how to do that. The general feature development flow around adding a flag and updating Echo is like so:
 
-- Add flag to eigen, `readyToRelease: false`
+- Add flag to eigen, `readyForRelease: false`
 - Add flag to echo, with `false`
 - Finish dev
-- Change eigen `readyToRelease: true`
+- Change eigen `readyForRelease: true`
 - When the app version is released, swap echo flag to `true` whenever we want users to get it
 
 After adding the echo key, and the PR is merged, update you local copy of echo by running `./scripts/update-echo`.
--- a/src/app/AppRegistry.tsx
+++ b/src/app/AppRegistry.tsx
@@ -8,13 +8,14 @@ import {
   WorksForYouScreenQuery,
 } from "app/Components/Containers/WorksForYou"
 import { FadeIn } from "app/Components/FadeIn"
+import { PageableScreensView } from "app/Components/PageableScreensView/PageableScreensView"
 import { ArtQuizArtworks } from "app/Scenes/ArtQuiz/ArtQuizArtworks"
 import { ArtQuizNavigation } from "app/Scenes/ArtQuiz/ArtQuizNavigation"
 import { ArtQuizResults } from "app/Scenes/ArtQuiz/ArtQuizResults/ArtQuizResults"
 import { SearchScreenQuery } from "app/Scenes/Search/Search"
 import { SearchScreenQuery as SearchScreenQuery2 } from "app/Scenes/Search/Search2"
 import { SearchSwitchContainer } from "app/Scenes/Search/SearchSwitchContainer"
-import React from "react"
+import React, { useMemo } from "react"
 import { AppRegistry, LogBox, Platform, View } from "react-native"
 import { GraphQLTaggedNode } from "relay-runtime"
 import { SafeAreaInsets, useScreenDimensions } from "shared/hooks"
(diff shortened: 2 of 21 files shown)


src/app/store/config/features.ts
```diff
@@ -49,171 +49,176 @@ function defineFeatures<T extends string>(featureMap: {
 export type FeatureName = keyof typeof features
 
 export const features = defineFeatures({
-  AROptionsPriceTransparency: {
+  ARArtworkRedesingPhase2: {
```
damassi, line 52: The following changes are just alphabetization

src/app/store/config/features.ts
```diff
@@ -49,171 +49,176 @@ function defineFeatures<T extends string>(featureMap: {
...
+  },
+  AREnableOrderHistoryOption: {
+    description: "Enable Order History in settings",
+    echoFlagKey: "AREnableOrderHistoryOption",
     readyForRelease: true,
-    echoFlagKey: "ARReorderSWAArtworkSubmissionFlow",
   },
-  ARArtworkRedesingPhase2: {
-    description: "Enable redesigned artwork page (phase 2)",
+  AREnablePageableArtworkScreens: {
```
damassi, line 156: Only addition

MounirDhahri (approved):
Nice!

src/app/Components/ArtworkGrids/ArtworkGridItem.tsx
```diff
@@ -96,6 +97,7 @@ export const Artwork: React.FC<ArtworkProps> = ({
 
   // This is needed to make sure the filter context is defined
   if (ArtworksFiltersStore.useStore()) {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
```
dimatretyak, line 100: praise 👌

src/app/Components/PageableScreensView/PageableLazyScreen.tests.tsx
```diff
@@ -0,0 +1,21 @@
...
+  })
+
+  it("renders page", async () => {
+    const tree = renderWithWrappers(<PageableLazyScreen screen={screen} shouldRender />)
+
+    await waitFor(() => {
+      expect(tree.queryByText("Screen 1")).toBeDefined()
+    })
+  })
+})
```
dimatretyak, line 21: **praise**: thank you for these test cases!

src/app/system/navigation/useNavigateToPageableRoute.ts
```diff
@@ -0,0 +1,32 @@
...
+
+interface UseNavigateToPagebleRouteProps<T> {
+  artworks: ReadonlyArray<T>
+}
+
+export interface PageableRouteProps {
+  navigateToPageableRoute: (url: string, options?: NavigateOptions) => void
+}
+
+export function useNavigateToPageableRoute<T extends { slug: string }>(
```
dimatretyak, line 12: **question**: should `artworks` be specified here? Maybe it will be enough if we pass only `slugs`? If not, what do you think about renaming `artworks` to something more general (e.g. `entities`, `pagableEntities`)?

src/app/Scenes/Home/Components/ArtworkModuleRail.tsx
```diff
@@ -64,13 +65,19 @@ const ArtworkModuleRail: React.FC<ArtworkModuleRailProps & RailScrollProps> = ({
...
   const showRail = artworks.length
 
   if (!showRail) {
     return null
   }
 
-  return artworks.length ? (
+  if (artworks.length === 0) {
+    return null
+  }
```
dimatretyak, line 78: **praise**: thanks for this little refactor 👌

src/app/AppRegistry.tsx
```diff
@@ -157,9 +158,38 @@ addTrackingProvider("console", ConsoleTrackingProvider)
...
+  const enablePageableArtworkScreens = useFeatureFlag("AREnablePageableArtworkScreens")
+
+  const pageableSlugs = props.pageableSlugs ?? []
+
+  const screens = useMemo(() => {
+    return pageableSlugs.map((slug) => ({
+      name: slug,
+      Component: <ArtworkQueryRenderer {...props} artworkID={slug} />,
+    }))
+  }, [pageableSlugs])
```
dimatretyak, line 174: **question**: it seems to me that we can omit the use of `useMemo` here. Since I don't think it will pass a lot of  `slugs`. Also, if `props.pageableSlugs` is `undefined`, then a new array entity will be used each time and the `useMemo` body will be executed each time

damassi:
@admbtlr - Given the designer question, should I close this PR for now?

src/app/system/navigation/useNavigateToPageableRoute.ts
damassi: Will rename to `items`! Though I like the simplicity of an array of strings, I want to keep it more than just strings because this allows us to automatically point back to the graphql fragment with a nice type error, signaling that we need to add `slug`, if it doesn't exist. With strings we're mostly just hoping for the best 😄

docs/developing_a_feature.md
```diff
@@ -68,7 +68,13 @@ First you need to add the new feature block in `src/app/store/config/features.ts
 The `description` property is what makes it possible to override the feature flag from the [Dev Menu](/dev_menu.md).
 
 We also need to add a flag in Echo, our remote feature flags configuration service.
-[Here](https://github.com/artsy/echo/commit/978a103e2c67a8010fabb2184f84aaef31d16f93) is an example PR for how to do that.
+[Here](https://github.com/artsy/echo/commit/978a103e2c67a8010fabb2184f84aaef31d16f93) is an example PR for how to do that. The general feature development flow around adding a flag and updating Echo is like so:
+
+- Add flag to eigen, `readyToRelease: false`
```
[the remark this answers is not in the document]
damassi: Updated

damassi:
Spoke with @admbtlr and we're going to merge this so that we can have something to interact with while the rest of the design process completes. As its behind a feature flag it should be harmless. 

Addressed everyones comments, but if there's anything else feel free to post back to this PR and i'll tackle in a follow-up. 

Will revisit this once we pick this work stream back up. Thanks for the review 👍

src/app/system/navigation/useNavigateToPageableRoute.ts
dimatretyak: got it. thanks for clarifications!

PR #8227, opened 2023-02-17
fix: increase save button hitslop

gkartalis:
This PR resolves [] 

### Description

Bug that came out of Mobile QA, save button not easily pressable.

- [x] Makes hitslop of save button bigger to be easier to press
- [x] adds customHitslop to palette button
- [x] bell icon wasn't working properly either when on dark button -> wasn't visible so fixed that as well
- [x] placeholder adjustment 

#### Screenshots (with touchable highlighted from RN inspector)

|Before|After|
|---|---|
|IMG_8685|IMG_8684|

|More screenshots|||
|---|---|---|
|IMG_1C12F1291251-2|IMG_1C12F1291251-1|IMG_25A02E0D4E1B-1|

### PR Checklist

### To the reviewers 👀

- increase save button hitslop - gkartalis

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/palette/Button/Button.tsx
+++ b/src/palette/Button/Button.tsx
@@ -76,6 +76,7 @@ export const Button: React.FC<ButtonProps> = ({
   testOnly_state,
   testID,
   textVariant,
+  hitSlop,
   ...rest
 }) => {
   const textVariantBySize = size === "small" ? "xs" : "sm"
@@ -141,7 +142,7 @@ export const Button: React.FC<ButtonProps> = ({
     <Spring native to={to} config={config.stiff}>
       {(springProps: typeof to) => (
         <Pressable
-          hitSlop={rest?.hitSlop}
+          hitSlop={hitSlop}
           testOnly_pressed={testOnly_state === DisplayState.Pressed}
           disabled={testOnly_state === DisplayState.Disabled || disabled}
           onPressIn={() => {
--- a/src/palette/Button/Button.tsx
+++ b/src/palette/Button/Button.tsx
@@ -53,9 +53,6 @@ export interface ButtonProps extends BoxProps {
   testOnly_state?: DisplayState
 
   textVariant?: TextVariant
-
-  /** customHitslop: used only if you want the button to have bigger hitslop than its actual layout */
-  customHitslop?: ButtonProps["hitSlop"]
 }
 
 enum DisplayState {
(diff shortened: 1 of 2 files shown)


dimatretyak (commented):
Unfortunately, the problem is still reproducible on my device. I'll take a look at the problem on Monday and try to fix it

https://user-images.githubusercontent.com/3513494/219748265-79d97cc7-5c7c-4729-bfec-9bd90b9bbf7e.mp4

gkartalis:
Found the reason for this, Pressable inside button does not inherit the rest of the props thus hitslop is not being used anywhere and that's why it is not changing the behavior at all.

Will push a fix here

gkartalis:
@pvinis let me know your thoughts, would we also want to support something like this on external palette as well? 

the text variant button wasn't that easy to press, I guess due to not visible boundaries like the rest of the buttons

src/palette/Button/Button.tsx
```diff
@@ -53,6 +53,9 @@ export interface ButtonProps extends BoxProps {
   testOnly_state?: DisplayState
 
   textVariant?: TextVariant
+
+  /** customHitslop: used only if you want the button to have bigger hitslop than its actual layout */
+  customHitslop?: ButtonProps["hitSlop"]
```
dimatretyak, line 58: **question**: can we rename `customHitslop` to `hitslop`? Sorry if the question is stupid 😅

dimatretyak (approved):

src/palette/Button/Button.tsx
lordkiz: Yeah, I also think this is better named hitslop

gkartalis: also agree 👍 on it

dimatretyak: thank you!

src/palette/Button/Button.tsx
```diff
@@ -141,6 +141,7 @@ export const Button: React.FC<ButtonProps> = ({
     <Spring native to={to} config={config.stiff}>
       {(springProps: typeof to) => (
         <Pressable
+          hitSlop={rest?.hitSlop}
```
pvinis, line 144: id suggest you get `hitSlop` from the props, and just use it without getting it from rest. put it directly on line 77 or something.

pvinis (dismissed):
nice thats a good change. we def have this manually added on folio, so i think adding this to palette-mobile makes sense. maybe even make it default? or use a prop like `bigHitSlop` that adds these?

pvinis:
(your example image reminds me of )

lordkiz (approved):

gkartalis:
also fixed something extra, placeholder adjustment and bell icon background

PR #8232, opened 2023-02-20
feat: Curators Pick Emerging home screen rail

olerichter00:
This PR resolves [CX-3405] 

### Description

Adds the "Curators Pick: Emerging" home screen rail. The grid view will follow in another PR.

### Screenshots

Simulator Screen Shot - iPhone 14 Pro - 2023-02-23 at 12 45 01

### PR Checklist

### To the reviewers 👀

- Curators Pick Emerging home screen rail - ole

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-3405]: https://artsyproduct.atlassian.net/browse/CX-3405?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/Scenes/Home/Home.tsx
```diff
@@ -518,6 +541,12 @@ export const HomeFragmentContainer = createRefetchContainer(
         ...NewWorksForYouRail_artworkConnection
       }
     `,
+    emergingPicksArtworks: graphql`
+      fragment Home_emergingPicksArtworks on Viewer {
+        ...MarketingCollectionRail_viewer
+          @arguments(marketingCollectionID: "curators-picks-emerging")
```
MounirDhahri, line 547: **thought:** would it be possible to check maybe if this name will be the same always and no one will rename it? our logic for the trove broke suddenly before because someone changed its name. 
The only other way around this that I see is having a separate interface for the trove/curators picks in metaphysics but we won't need that if the name is staying as it is

MounirDhahri (commented):
LGTM - I left one small question, please let me know you want to catch up about it to explain further my concerns

src/app/Scenes/Home/Home.tsx
olerichter00: Unfortunately, we don't have that option. But at least we rely on the key in [Force as well](https://github.com/artsy/force/blob/826c74841200cbfff63376d14a133a9f46d1e788/src/Apps/Home/Components/HomeEmergingPicksArtworksRail.tsx#L48) (in many places), so changing the key would probably result in some failed integrity tests.

MounirDhahri: Sounds good - i think we could have had that in a separate interface that returns the artworks from the trove. Let's continue with the same logic as in Force and solve this when the problem arises

MounirDhahri (approved):

PR #8276, opened 2023-02-27
feat: add tracking to home screen rails and improve performance

MounirDhahri:
This PR resolves [CX-3385] 

### Description

This PR adds tracking to the home screen rails.
Bonus: while adding tracking, I noticed few issues related to performance, then more issues then more issues, along the way I decided to fix everything I can fix.
- FPS went from 24 to 47 in boot
- FPS went from 32 to 48 after 3 seconds from boot
- Removed 2 unnecessary rerenders
- Clean up logic a bit

Before changes

https://user-images.githubusercontent.com/11945712/221570249-9ec8d682-6ba9-42fb-9bde-2a4f3c18d704.mov

After changes

https://user-images.githubusercontent.com/11945712/221570185-5ad95c3c-8fbb-40ee-95a1-880f4fbc1b49.mov

Before changes (throttle)

https://user-images.githubusercontent.com/11945712/221570293-690e88d8-2fd3-401e-b8c2-181ad993bfee.mov

After changes (throttling is way shorter and almost not noticeable)

https://user-images.githubusercontent.com/11945712/221570496-0709566b-7fa4-42d6-a957-f0c0cfe308fa.mov

https://user-images.githubusercontent.com/11945712/221570549-f7158835-d538-4dbb-8081-3a3d391045fe.mov

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-3385]: https://artsyproduct.atlassian.net/browse/CX-3385?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Components/ArtworkRail/ArtworkRail.tsx
+++ b/src/app/Components/ArtworkRail/ArtworkRail.tsx
@@ -5,6 +5,7 @@ import { SmallArtworkRail_artworks$data } from "__generated__/SmallArtworkRail_a
 import { ArtworkCardSize, ArtworkRailCard } from "app/Components/ArtworkRail/ArtworkRailCard"
 import { PrefetchFlatList } from "app/Components/PrefetchFlatList"
 import { useFeatureFlag } from "app/store/GlobalStore"
+import { isPad } from "app/utils/hardware"
 import { Schema } from "app/utils/track"
 import { Button } from "palette"
 import React, { ReactElement } from "react"
@@ -53,6 +54,8 @@ export const ArtworkRail: React.FC<ArtworkRailProps> = ({
 }) => {
   const enableBrowseMoreArtworksCard = useFeatureFlag("AREnableBrowseMoreArtworksCard")
 
+  const isTablet = isPad()
+
   return (
     <PrefetchFlatList
       onEndReached={onEndReached}
@@ -73,7 +76,7 @@ export const ArtworkRail: React.FC<ArtworkRailProps> = ({
       // We need to set the maximum number of artists to not cause layout shifts
       // @ts-expect-error
       data={artworks.slice(0, MAX_NUMBER_OF_ARTWORKS)}
-      initialNumToRender={MAX_NUMBER_OF_ARTWORKS}
+      initialNumToRender={isTablet ? 10 : 5}
       contentContainerStyle={{ alignItems: "flex-end" }}
       renderItem={({ item, index }) => (
         <ArtworkRailCard
--- a/src/app/Components/Home/ArtistRails/RecommendedArtistsRail.tsx
+++ b/src/app/Components/Home/ArtistRails/RecommendedArtistsRail.tsx
@@ -1,5 +1,5 @@
 import { ActionType, OwnerType } from "@artsy/cohesion"
-import { Spacer, Flex, SpacingUnit } from "@artsy/palette-mobile"
+import { Spacer, Flex } from "@artsy/palette-mobile"
 import { ArtistCard_artist$data } from "__generated__/ArtistCard_artist.graphql"
 import { RecommendedArtistsRailFollowMutation } from "__generated__/RecommendedArtistsRailFollowMutation.graphql"
 import { RecommendedArtistsRail_me$data } from "__generated__/RecommendedArtistsRail_me.graphql"
@@ -11,7 +11,7 @@ import HomeAnalytics from "app/Scenes/Home/homeAnalytics"
 import { defaultEnvironment } from "app/system/relay/createEnvironment"
 import { extractNodes } from "app/utils/extractNodes"
 import { Spinner } from "palette"
-import React, { useImperativeHandle, useRef, useState } from "react"
+import React, { memo, useImperativeHandle, useRef, useState } from "react"
 import { FlatList, ViewProps } from "react-native"
 import {
   commitMutation,
@@ -30,7 +30,6 @@ interface RecommendedArtistsRailProps extends ViewProps {
   subtitle?: string
   relay: RelayPaginationProp
   me: RecommendedArtistsRail_me$data
-  mb?: SpacingUnit
 }
 
 export const RecommendedArtistsRail: React.FC<RecommendedArtistsRailProps & RailScrollProps> = ({
@@ -39,7 +38,6 @@ export const RecommendedArtistsRail: React.FC<RecommendedArtistsRailProps & Rail
   relay,
   me,
   scrollRef,
-  mb,
 }) => {
   const { trackEvent } = useTracking()
 
@@ -82,7 +80,7 @@ export const RecommendedArtistsRail: React.FC<RecommendedArtistsRailProps & Rail
   }
 
   return (
-    <Flex mb={mb}>
+    <Flex>
       <Flex pl={2} pr={2}>
         <SectionTitle title={title} subtitle={subtitle} />
       </Flex>
@@ -92,6 +90,7 @@ export const RecommendedArtistsRail: React.FC<RecommendedArtistsRailProps & Rail
         prefetchVariablesExtractor={defaultArtistVariables}
         data={artists as any}
         keyExtractor={(artist) => artist.id}
+        initialNumToRender={3}
         onEndReached={loadMoreArtists}
         renderItem={({ item: artist, index }) => {
           return (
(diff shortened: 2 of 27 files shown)


src/app/Components/ArtworkRail/ArtworkRail.tsx
```diff
@@ -73,7 +77,7 @@ export const ArtworkRail: React.FC<ArtworkRailProps> = ({
       // We need to set the maximum number of artists to not cause layout shifts
       // @ts-expect-error
       data={artworks.slice(0, MAX_NUMBER_OF_ARTWORKS)}
-      initialNumToRender={MAX_NUMBER_OF_ARTWORKS}
+      initialNumToRender={isTablet ? 10 : 5}
```
olerichter00, line 80: Nice 👏

src/app/Components/Home/ArtistRails/RecommendedArtistsRail.tsx
```diff
@@ -30,7 +30,6 @@ interface RecommendedArtistsRailProps extends ViewProps {
   subtitle?: string
   relay: RelayPaginationProp
   me: RecommendedArtistsRail_me$data
-  mb?: SpacingUnit
```
olerichter00, line 33: 👏

src/app/Scenes/Home/Home.tsx
```diff
@@ -125,131 +159,336 @@ const Home = (props: Props) => {
...
+          }
+        })
+      }
+    }
+  ).current
 
   const { isRefreshing, handleRefresh, scrollRefs } = useHandleRefresh(relay, modules)
+  // const { isRefreshing, handleRefresh, scrollRefs } = useMemo(() => {
+  //   return getRefreshHelpers(relay, modules)
+  // }, [relay, modules])
```
olerichter00, line 367: 👀

src/app/Scenes/Home/Home.tsx
```diff
@@ -125,131 +159,336 @@ const Home = (props: Props) => {
...
+    articlesConnection,
+    meBelow,
+    meAbove,
+    homePageBelow,
+    homePageAbove,
+    emergingPicksArtworks,
   ])
 
-  modules = modules.filter((module) => !module.hidden && module.data)
+  const onViewableItemsChanged = React.useRef(
```
olerichter00, line 349: **suggestion:** Maybe we can add a comment here to explain shortly what we're doing here.

src/app/Scenes/Home/Home.tsx
```diff
@@ -125,131 +159,336 @@ const Home = (props: Props) => {
...
-      type: "shows",
-      data: showsByFollowedArtists,
-    },
-    {
-      title: "Featured Fairs",
-      subtitle: "See Works in Top Art Fairs",
-      type: "fairs",
-      data: homePageBelow?.fairsModule,
-    },
+  const modules: HomeModule[] = useMemo(() => {
```
olerichter00, line 175: **suggestion:** What do you think about moving this to another file to reduce the file size. 

```
const modules = useHomeScreenModules(props)
```

olerichter00 (approved):
This is great work! Thanks a lot for the performance improvements 🌟 🥳 🌟

src/app/Scenes/Home/Home.tsx
MounirDhahri: I was thinking about that, I will do it now

src/app/Components/LotsByArtistsYouFollowRail/LotsByFollowedArtistsRail.tsx
```diff
@@ -47,13 +46,13 @@ export const LotsByFollowedArtistsRail: React.FC<Props> = ({ title, me, relay, m
...
   return (
-    <Flex mb={mb}>
+    <Flex>
       <Flex mx={2}>
         <SectionTitle title={title} onPress={() => navigate("/lots-by-artists-you-follow")} />
       </Flex>
       <CardRailFlatList
         data={artworks}
-        initialNumToRender={PAGE_SIZE}
+        initialNumToRender={3}
```
lordkiz, line 55: I wonder if 3 is enough to fill up large screens like 15 inch pads.

src/app/Scenes/Home/Components/ArticlesRail.tsx
```diff
@@ -1,30 +1,34 @@
...
-  mb?: SpacingUnit
 }
 
-export const ArticlesRail: React.FC<ArticlesRailProps> = ({ title, articlesConnection, mb }) => {
+export const ArticlesRail: React.FC<ArticlesRailProps> = ({ title, articlesConnection }) => {
   const articles = extractNodes(articlesConnection)
   const tracking = useTracking()
 
+  const { width } = useScreenDimensions()
+  const isTablet = width > 700
```
lordkiz, line 24: We already have `isPad()` utility function

MounirDhahri: I was sure we have an I cound't find 😄. It's because it is called `isPad` and not `isiPad` 😄.

MounirDhahri: thanks I will fix it

src/app/Components/LotsByArtistsYouFollowRail/LotsByFollowedArtistsRail.tsx
MounirDhahri: Good catch, I missed this, I will update it accordingly

src/app/Scenes/Home/Home.tsx
```diff
@@ -264,335 +309,216 @@ const Home = (props: Props) => {
...
-                )
-              default:
-                return null
-            }
-          }}
+          renderItem={renderItem}
           ListHeaderComponent={<HomeHeader />}
           ListFooterComponent={() => <Flex mb={4}>{!!loading && <BelowTheFoldPlaceholder />}</Flex>}
-          keyExtractor={(_item, index) => String(index)}
+          ItemSeparatorComponent={ModuleSeparator}
```
olerichter00, line 320: Sorry, I missed that in the review: Actually, we can't just use `ItemSeparatorComponent` here and we need to add the `mb` property to each component. Some of the components are rendered as `null` when there is no content. This leads to duplicate spacing between the surrounding components.

src/app/Scenes/Home/homeModules.ts
```diff
@@ -0,0 +1,198 @@
...
+import { articlesQueryVariables } from "app/Scenes/Articles/Articles"
+import { isOnboardingVisible } from "app/Scenes/Home/Components/HomeFeedOnboardingRail"
+import { HomeProps } from "app/Scenes/Home/Home"
+import { lotsByArtistsYouFollowDefaultVariables } from "app/Scenes/LotsByArtistsYouFollow/LotsByArtistsYouFollow"
+import { useFeatureFlag } from "app/store/GlobalStore"
+import { isEmpty } from "lodash"
+import { useMemo } from "react"
+import ReactAppboy from "react-native-appboy-sdk"
+
+export const useHomeModules = (props: HomeProps, cards: ReactAppboy.CaptionedContentCard[]) => {
```
damassi, line 11: @MounirDhahri - could we follow-up and rename this file to `useHomeModule` so that it maps 1:1 with the main function export?

damassi:
Nice, so `memo` is what helped perf? We should keep that in mind for the future 👍

src/app/Scenes/Home/homeModules.ts
MounirDhahri: good point, i will do it.

MounirDhahri:
Yes, along with few other things
- having the default windowsize as 1 before the user started to scrolled, meant that the initial scroll would always drop frames - most users are curious to scroll at least a little bit down - I increased the first rendered window size to 2 instead
- we had 2 unnecessary renders that I wasn't able to identify the root cause behind them, wrapping the home component with a memo fixed them.
- the home modules was by far the most computational heavy component, extracting it and making sure that I used a `useMemo` and specified only the right arguments to the dependency array helped
- reducing the number of items we initially in all rails also saved few frames, instead of rendering 10 items on all rails as soon as we render them, now we are rendering half initially - meaning half the items in screen with 14 rails

src/app/Scenes/Home/homeModules.ts
```diff
@@ -0,0 +1,198 @@
...
+      {
+        data: props.homePageAbove?.activeBidsArtworkModule,
+        isEmpty: isEmpty(props.homePageAbove?.activeBidsArtworkModule?.results),
+        title: "Your Active Bids",
+        type: "artwork",
+      },
+      {
+        contextModule: ContextModule.auctionLotsEndingSoonRail,
+        data: props.meAbove,
+        isEmpty: !props.meAbove?.lotsByFollowedArtistsConnectionCount?.totalCount,
```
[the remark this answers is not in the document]
MounirDhahri: I might have an idea to solve that - I am thinking of masking the data.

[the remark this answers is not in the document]
olerichter00: That could work 👍

PR #8302, opened 2023-03-01
chore: upgrade relay deps and fix typing issues

MounirDhahri:
### Description

This PR comes as an effort to bring improvements to the home and my collection tabs. See https://artsyproduct.atlassian.net/browse/CX-3468

Spoiler: I am preparing to introduce `defer` and evaluate updating the way we query the data

Please note that some of the bugs were already there but we were not picking them up, I ignored them for now but I think we should fix them next time we touch those lines

_This was a hard one to get done 😢_
crying-cry

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>


package.json
```diff
@@ -329,6 +328,10 @@
     "typescript": "4.9.5",
     "vscode-apollo-relay": "1.5.2"
   },
+  "resolutions": {
+    "@types/relay-runtime": "14.1.3",
+    "@types/react-relay": "14.1.3"
+  },
```
MounirDhahri, line 334: this was the issue 🥲

src/app/Scenes/Artist/Artist.tsx
```diff
@@ -132,7 +133,7 @@ export const Artist: React.FC<ArtistProps> = (props) => {
 }
 
 interface ArtistQueryRendererProps {
-  environment?: RelayModernEnvironment
+  environment?: RelayModernEnvironment | MockEnvironment
```
olerichter00, line 136: **suggestion:** Not sure whether it's a good idea to integrate types for testing here. What do you think of instead doing something like this in our test files (not sure if this would fix the issue)?

```
environment as RelayModernEnvironment
```

src/app/Components/Artist/ArtistInsights/MarketStats.tests.tsx
```diff
@@ -9,6 +9,7 @@ import { useTracking } from "react-tracking"
 import { createMockEnvironment } from "relay-test-utils"
 import { MarketStatsFragmentContainer, MarketStatsQueryRenderer } from "./MarketStats"
 
+// eslint-disable-next-line react-hooks/rules-of-hooks
 const trackEvent = useTracking().trackEvent
```
gkartalis, line 13: **suggestion(non-blocking):** I think we can just replace this with [`mockTrackEvent`](https://github.com/artsy/eigen/blob/ebf6eee65da533d13f92cd351fc2234101cb7be1/src/app/utils/tests/globallyMockedStuff.ts#L8)

src/app/Scenes/Feature/Feature.tests.tsx
```diff
@@ -4,7 +4,7 @@ import { renderWithWrappersLEGACY } from "app/utils/tests/renderWithWrappers"
 import { createMockEnvironment, MockPayloadGenerator } from "relay-test-utils"
 import { FeatureQueryRenderer } from "./Feature"
 
-let mockRelayEnvironment = defaultEnvironment as ReturnType<typeof createMockEnvironment>
+let mockRelayEnvironment = defaultEnvironment as unknown as ReturnType<typeof createMockEnvironment>
```
olerichter00: 👀 Would this work as well?

```suggestion
let mockRelayEnvironment = defaultEnvironment as unknown
```

src/app/Components/Artist/ArtistInsights/MarketStats.tsx
```diff
@@ -15,6 +15,7 @@ import { ScrollView } from "react-native"
 import { createFragmentContainer, graphql, QueryRenderer } from "react-relay"
 import { useTracking } from "react-tracking"
 import RelayModernEnvironment from "relay-runtime/lib/store/RelayModernEnvironment"
+import { MockEnvironment } from "relay-test-utils"
```
gkartalis, line 18: 👀 why do we need this??

olerichter00 (approved):

src/app/utils/useImageSearch.ts
```diff
@@ -3,7 +3,7 @@ import { defaultEnvironment } from "app/system/relay/createEnvironment"
 import { ReactNativeFile } from "extract-files"
 import { useState } from "react"
 import { Platform } from "react-native"
-import { fetchQuery, graphql } from "react-relay"
+import { fetchQuery, graphql } from "relay-runtime"
```
gkartalis, line 6: had some weird crashes some times when importing from relay-runtime vs react-relay

src/app/Scenes/Feature/Feature.tests.tsx
gkartalis: nope, it will raise a type error when trying to access `mockRelayEnvironment.mock.resolveMostRecentOperation`

src/app/Components/Artist/ArtistInsights/MarketStats.tsx
gkartalis: it is also a devDependency which will break in prod build I think since theoretically it is not included in the bundle

gkartalis:
@MounirDhahri let me know if you wanna pair on avoiding passing the mockenvironment to actual components, happy to help and I've played around with relay testing a bunch

src/app/Scenes/Artist/Artist.tsx
damassi: Or even `as unknown as RelayModernEnvironment`, so we still get code completion. For something as widely used as this I think we could go that route at the lowest level of the stack (if necessary) and still be safe.

src/app/utils/useImageSearch.ts
damassi: Interesting, does `fetchQuery` still type-check? I noticed that the relay 14 jump broke this, which forced us to move to `react-relay`. Maybe the fixed it and now its ok, like it was before.

src/app/Scenes/SellWithArtsy/SubmitArtwork/SubmitArtwork.tests.tsx
```diff
@@ -50,7 +50,9 @@ describe(SubmitArtwork, () => {
   })
 
   describe("Submission Flow", () => {
-    const mockEnvironment = defaultEnvironment as ReturnType<typeof createMockEnvironment>
+    const mockEnvironment = defaultEnvironment as unknown as ReturnType<
+      typeof createMockEnvironment
+    >
```
damassi, line 55: Not for this PR! We should think about wrapping all of these instances up in a single function

src/app/Scenes/Feature/Feature.tests.tsx
olerichter00: What is `as unknown as ReturnType<typeof createMockEnvironment>` actually doing?

src/app/utils/useImageSearch.ts
MounirDhahri: yes, the issue was that `react-relay` and `relay-runtime` were both using different types below the hood that broke things for us.

src/app/Scenes/Feature/Feature.tests.tsx
MounirDhahri: when casting type A to B typescript checks if type A properties are include in type B, if not it raises an error. in this case, I am confident the types are accurate regardless of the typescript warning. Typescript docs suggest in this case to always casting as `unknown` the to the type you need (instead of casting as `any`)

gkartalis (dismissed):

PR #8372, opened 2023-03-16
chore: track artwork views in NWFY and AR rails

MounirDhahri:
This PR resolves [CX-3463] 

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>


src/app/Scenes/Home/Components/ArtworkRecommendationsRail.tsx
```diff
@@ -14,21 +15,27 @@ import { useTracking } from "react-tracking"
...
+> = memo(({ isRailVisible, mb, me, scrollRef, title }) => {
   const { trackEvent } = useTracking()
 
   const { artworkRecommendations } = useFragment(artworksFragment, me)
 
   const railRef = useRef<View>(null)
   const listRef = useRef<FlatList<any>>(null)
 
+  const { onViewableItemsChanged, viewabilityConfig } = useItemsImpressionsTracking({
+    isRailVisible: !!isRailVisible,
```
olerichter00, line 35: nitpick: No need to convert to a boolean here:

```suggestion
    isRailVisible,
```

src/app/Scenes/Home/Components/NewWorksForYouRail.tsx
```diff
@@ -30,6 +32,11 @@ export const NewWorksForYouRail: React.FC<NewWorksForYouRailProps & RailScrollPr
     const railRef = useRef<View>(null)
     const listRef = useRef<FlatList<any>>(null)
 
+    const { onViewableItemsChanged, viewabilityConfig } = useItemsImpressionsTracking({
+      isRailVisible: !!isRailVisible,
```
olerichter00, line 36:
```suggestion
      isRailVisible,
```

src/app/Scenes/Home/Components/useImpressionsTracking.ts
```diff
@@ -0,0 +1,68 @@
+import { ContextModule } from "@artsy/cohesion"
+import HomeAnalytics from "app/Scenes/Home/homeAnalytics"
+import { useFeatureFlag } from "app/store/GlobalStore"
+import { useEffect, useRef, useState } from "react"
+import { ViewToken, ViewabilityConfig } from "react-native"
+import { useTracking } from "react-tracking"
+
+type TrackableItem = { id: string; index: number }
+export const useItemsImpressionsTracking = ({
```
olerichter00, line 9: nitpick: To make it easier to read:

```suggestion
type TrackableItem = { id: string; index: number }

export const useItemsImpressionsTracking = ({
```

src/app/Scenes/Home/Components/useImpressionsTracking.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  const tracking = useTracking()
+
+  const enableItemsViewsTracking = useFeatureFlag("ARImpressionsTrackingHomeItemViews")
+
+  const viewabilityConfig = useRef<ViewabilityConfig>({
+    // Percent of of the item that is visible for a partially occluded item to count as "viewable"
+    itemVisiblePercentThreshold: 60,
+    // @ts-expect-error
+    viewAreaCoveragePercentThreshold: null,
```
olerichter00, line 28: **issue:** `viewAreaCoveragePercentThreshold` expects a number and this could lead to unexpected behaviour 👀

src/app/Scenes/Home/Components/useImpressionsTracking.ts
```diff
@@ -0,0 +1,68 @@
...
+    minimumViewTime: 3000,
```
olerichter00, line 29: **Question:** 3 seconds seems a bit long to me. Did we do any tests beforehand to get a reasonable number?

src/app/store/config/features.ts
```diff
@@ -248,6 +248,12 @@ export const features = defineFeatures({
     showInDevMenu: true,
     echoFlagKey: "AREnableSkeletonAnimation",
   },
+  ARImpressionsTrackingHomeItemViews: {
+    description: "Enable Tracking Items on Home Screen",
```
olerichter00, line 252: **suggestion:** Maybe this is a bit more precise:

```suggestion
    description: "Enable Tracking Item views on Home Screen",
```

olerichter00 (commented):
Great work 🌟

src/app/Scenes/Home/Components/useImpressionsTracking.ts
MounirDhahri: I also thought so but it wasn't the case actually when I was reading the code. React-native [docs](https://reactnative.dev/docs/flatlist#viewareacoveragepercentthreshold) specify that you need one of `itemVisiblePercentThreshold` or `itemVisiblePercentThreshold` to be there and not both. In practice, this didn't work and it was breaking and showing an error message and requiring me to specify only one.

MounirDhahri: let me see if this is still the case, because last time I had this was 3 weeks ago and we didn't update the RN version back there

src/app/store/config/features.ts
MounirDhahri: good idea

src/app/Scenes/Home/Components/useImpressionsTracking.ts
MounirDhahri: yes, it's still happening 

https://user-images.githubusercontent.com/11945712/230440881-13dc984e-1dfa-43d7-b695-a1da072205e4.mov

olerichter00: Then I would expect not setting it at all would give us what we want 🤔 

```suggestion
```

MounirDhahri: I tried doing that in the video above actually

olerichter00: We're setting it to `null` [here](https://github.com/artsy/eigen/blob/a82ebdc08ecb21c635cc37292181b827b87357f6/src/app/Scenes/Home/Home.tsx#L139) as well. https://github.com/artsy/eigen/pull/8476 would fix the types and the field.

MounirDhahri: I am looking for the error in the RN repo and forgot where it was, I can share it here.

olerichter00: I found the issue. The home screen uses `PrefetchFlatlist` which sets `viewAreaCoveragePercentThreshold` ([GitHub](https://github.com/artsy/eigen/blob/ef0d3b2f940a8ab5887106e881320066fa6d3daa/src/app/Components/PrefetchFlatList.tsx#L37)). If we want to use `itemVisiblePercentThreshold` instead we could adjust PrefetchFlatList's viewability config.

MounirDhahri: here it it is https://github.com/facebook/react-native/blob/3506c8d7405a564c4311eb9bfb88edcf8e8ec778/packages/virtualized-lists/Lists/ViewabilityHelper.js#L132

MounirDhahri: they're checking for the value being null instead of undefined or null.

MounirDhahri: Can't patch it now, it's bundled within RN 😅

olerichter00: Interesting 🤔 I don't think we need to patch it. https://github.com/artsy/eigen/pull/8476 fixes the issue without ignoring the config types.

MounirDhahri: never mind my comments- you're right, it was injected by us

MounirDhahri: thanks for the fix 🙏

MounirDhahri: done-spongebob

MounirDhahri: I merged your branch and rebased - should be good now

olerichter00: Nice. What do you think about https://github.com/artsy/eigen/pull/8477 to avoid mixing the config that is passed in via props and the default config? Mixing the config doesn't seem to be a good idea.

PR #8426, opened 2023-03-29
fix(flipper): add react devtools as a dev dep to make flipper work

gkartalis:
This PR resolves [] 

### Description

react dev tools are broken in flipper due to latest react native update, added `react-devtools-core` to devDeps to fix that

|Before|After|
|---|---|
|Screenshot 2023-03-29 at 14 13 01|Screenshot 2023-03-29 at 14 13 26|

### PR Checklist

- [ ] I have tested my changes on **iOS** and **Android**.
- [ ] I hid my changes behind a **[feature flag]**, or they don't need one.
- [ ] I have included **screenshots** or **videos**, or I have not changed the UI.
- [ ] I have added **tests**, or my changes don't require any.
- [ ] I added an **[app state migration]**, or my changes do not require one.
- [ ] I have documented any **follow-up work** that this PR will require, or it does not require any.
- [ ] I have added a **changelog entry** below, or my changes do not require one.

### To the reviewers 👀

- [ ] I would like **at least one** of the reviewers to **run** this PR on the simulator or device.

#nochangelog

PR #8429, opened 2023-03-29
fix: step 2 needs resolution to work properly 😅

gkartalis:
This PR resolves [] 

### Description

https://github.com/artsy/eigen/pull/8426 needs resolution in package json to work 😅

### PR Checklist

### To the reviewers 👀

#nochangelog

### DIFF
--- a/package.json
+++ b/package.json
@@ -331,7 +331,8 @@
   },
   "resolutions": {
     "@types/relay-runtime": "14.1.3",
-    "@types/react-relay": "14.1.3"
+    "@types/react-relay": "14.1.3",
+    "react-devtools-core": "4.27.4"
   },
   "pull-lock": {
     "ios/Podfile.lock": "echo \"🚨 New Podfile.lock detected!, you might need to run yarn install:all\" 🚨",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -331,7 +331,8 @@
   },
   "resolutions": {
     "@types/relay-runtime": "14.1.3",
-    "@types/react-relay": "14.1.3"
+    "@types/react-relay": "14.1.3",
+    "react-devtools-core": "4.27.4"
```
gkartalis, line 335: needs the resolution here explicitly because the react-devtools-core get resolved without it to the react-devtools-core@4.19.1 due to our react native version (which doesn't work)

dimatretyak (approved):
LGTM 👍

MounirDhahri (approved):

package.json
dimatretyak: is it ok that we have the `4.19.1` version in the `yarn.lock` [file](https://github.com/artsy/eigen/blob/901b6bd74a95095de034d0d690a0420aa1d402f7/yarn.lock#L13748)?

gkartalis: I think its due to react native 🤔 react-native has `react-devtools-core` as a dependency but version `4.19.1` and for this reason it is like that on yarn.lock and it overrides our devtools version. 

🤷‍♂️ that's why I added the resolution

damassi:
@gkartalis totally non-blocking but if you think that this might benefit from a HACKS entry, might be good to add there.

PR #8446, opened 2023-04-03
chore: improve home rail views logic

MounirDhahri:
This PR resolves [CX-3385] 

### Description

**Problem:**
- Rails below the fold show position_y as: -1

[5 Whys](https://www.mindtools.com/a3mi00v/5-whys#:~:text=The%20method%20is%20remarkably%20simple,prevent%20the%20issue%20from%20recurring.) **Rails below the fold show position_y as: -1**

- Why? The module visible is not inside `modules` (modules is the result of the query on the home feed)
- Why? Because it didn't yet load
- Why? Because Below the fold query comes later
- Still! Why is it not loaded when the query is solved? because react-native `onViewableItemsChanged` doesn't support dynamic updates so far

**Solution after discussing this with Audrey:**
- Extract the sort and have the same sort everywhere regardless of the item available (we could have done this earlier and it would have made this ticket much easier 🥲)

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-3385]: https://artsyproduct.atlassian.net/browse/CX-3385?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/Scenes/Home/useHomeModules.ts
```diff
@@ -1,44 +1,71 @@
...
 import { isOnboardingVisible } from "app/Scenes/Home/Components/HomeFeedOnboardingRail"
-import { HomeProps } from "app/Scenes/Home/Home"
+import { HomeModule, HomeProps } from "app/Scenes/Home/Home"
 import { lotsByArtistsYouFollowDefaultVariables } from "app/Scenes/LotsByArtistsYouFollow/LotsByArtistsYouFollow"
 import { useFeatureFlag } from "app/store/GlobalStore"
 import { isEmpty } from "lodash"
 import { useMemo } from "react"
 import ReactAppboy from "react-native-appboy-sdk"
 
+export const HOME_RAILS_SORT = [
```
olerichter00, line 11: I don't understand why we need this additional array. Couldn't we just use the order of the array with the modules?

olerichter00: Could we change `useHomeModules` to return the list of the filtered modules and a list of all modules (or just an array of module keys if we only need that (`const moduleKeys = allModules.map((module) => module.key)`). We could use moduleKeys then instead of `HOME_RAILS_SORT`.

olerichter00: Having this array of keys to sort by also seems nice. The only problem I see is that it could lead to very hard to fix problems if these keys don't match (e.g. when adding a new module).

olerichter00 (approved):
Looks great 👍

PR #8459, opened 2023-04-04
fix(artwork): Add fixed pageable artwork screen headers

damassi:
This PR resolves [MOPLAT-662]

### Description

This updates our feature-flagged pageable screen artwork views to now have a fixed header. 

https://user-images.githubusercontent.com/236943/229931082-4ee97a9b-8ce9-44b6-a631-0bcc77b4ca7b.mp4

cc @artsy/mobile-platform 

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>


src/app/AppRegistry.tsx
```diff
@@ -162,40 +161,6 @@ LogBox.ignoreLogs([
...
-        prefetchScreensCount={5}
-      />
-    )
-    // If not within the context of an artwork collection, just render the
-    // individual artwork.
-  } else {
-    return <ArtworkQueryRenderer {...props} />
-  }
-}
-
```
damassi, line 198: Moved this to `Artwork.tsx` to keep everything more contained

src/app/AppRegistry.tsx
```diff
@@ -385,7 +350,7 @@ export const modules = defineModules({
   ArtistArticles: reactModule(ArtistArticlesQueryRenderer),
   ArtistSeries: reactModule(ArtistSeriesQueryRenderer),
   Artwork: reactModule(
-    Artwork,
+    ArtworkPageableScreen,
```
damassi, line 353: Right now this is technically feature-flagged, but we're pretty certain this will launch. If not, will come back here and update.

src/app/Components/PageableScreensView/PageableScreensContext.ts
```diff
@@ -15,9 +15,10 @@ export const PageableScreensContext = createContext(null as unknown as PageableS
 export const usePageableScreensContext = () => {
   const context = useContext<PageableScreensContextValue>(PageableScreensContext)
 
-  if (!context) {
-    throw new Error("usePageableScreensContext must be inside PageableScreensContext.Provider")
-  }
+  // FIXME: Uncomment once we launch pageable screens view
+  // if (!context) {
+  //   throw new Error("usePageableScreensContext must be inside PageableScreensContext.Provider")
+  // }
```
damassi, line 21: To keep things simple while feature-flagged, commented this out so that we can share code between enabled and disabled. Will uncomment this once we launch to make things a bit more strict

gkartalis (approved):
Looks good so far!

Noticed though your comments for removing the feature flag when we are ready to launch (might be missing some context here on the launch plan) but don't we usually release the features **hidden** behind the feature flag and in case something goes really wrong so we can instantly _kill_ it 🤔 ?

Will also do a quick QA in a bit

damassi:
Ah, I was assuming that those flags were mostly for internal use, so that we can test and QA and so on before launch. Never considered that they were effectively kill-switches! I can imagine that for larger / more complex features, but for smaller features like this its probably best to just keep the feature flag density down and remove (but maybe i'm still in a web headspace here). 

What do you think @gkartalis?

gkartalis:
Hmm, good point about keeping the flag density down, 

generally I feel way safer when releasing something like this behind a ff (that includes screen transitioning, fetching data, and also affects the Artwork page which is the most GMV important screen that we have) 

the reason is that if anything goes wrong we **need** to wait some days (best case scenario something less than 1day) till we can deploy a hot-fix (without including the time to fix it).

damassi:
That makes sense!

PR #8498, opened 2023-04-12
feat(CX-3598): add new onboarding cards to home screen

dariakoko:
This PR resolves [CX-3598] 

### Description

https://user-images.githubusercontent.com/36167539/231513604-9d69b90b-cf7e-4707-8009-69b24ba3f216.mp4

### PR Checklist

### To the reviewers 👀

- added new Artsy onboarding rail to the home screen -daria

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-3598]: https://artsyproduct.atlassian.net/browse/CX-3598?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/Scenes/Home/useHomeModules.ts
```diff
@@ -58,6 +60,14 @@ export const useHomeModules = (props: HomeProps, cards: ReactAppboy.CaptionedCon
         title: "Auctions",
         type: "sales",
       },
+      {
+        data: true, // displays static data
```
olerichter00, line 64: **question:** Do we need to pass anything here or can we leave the data empty?

```suggestion
```

olerichter00 (approved):
Looks great 🌟 

I wonder if we could or need to reduce the size of the images 🤔

src/app/Scenes/Home/useHomeModules.ts
dariakoko: Data is the required prop and in the Home.ts file, we check the data https://github.com/artsy/eigen/blob/main/src/app/Scenes/Home/Home.tsx#L208 and display the component only if the data is present.
I thought the way I did it is the easiest

olerichter00: That makes sense 👍 Let's keep it as it is.

dariakoko:
@olerichter00 You mean the images themselves, not the way we display them in the component? Not sure, I downloaded them from figma and they have the same size as the ones on the old cards. You would suggest reducing the size?

olerichter00:
> @olerichter00 You mean the images themselves, not the way we display them in the component? Not sure, I downloaded them from figma and they have the same size as the ones on the old cards. You would suggest reducing the size?

Ok, let's add them as they are. I remember @MounirDhahri was working on the image sizes in Eigen a while ago. @MounirDhahri, do you think it is possible and needed to reduce the size of the images?

dariakoko:
@MounirDhahri @olerichter00 maybe we could document somewhere the optimal image sizes? Do we have a readme for similar purposes?

MounirDhahri:
I don't think it's possible to document that in numbers but it's probably a good idea to write this somewhere in our docs. The rationale behind it comes from some studies that emerged a few years ago about how [app size matters](https://medium.com/googleplaydev/shrinking-apks-growing-installs-5d3fcba23ce2). Overall, the smaller the app size, the more people convert into installing the app. In our case, since we have a fair amount of images in the app, our bundle size has been increasing quickly and it's important for us to try to cut the image size to avoid [affecting our conversion](https://segment.com/blog/mobile-app-size-effect-on-downloads/)

PR #8504, opened 2023-04-13
fix(FX-4728): inconsistency in auction countdown on home feed

dimatretyak:
This PR resolves [FX-4728] 

Slack thread: [link](https://artsy.slack.com/archives/C03N12SR0RK/p1680797397951269)

### Description
Sometimes the incorrect auction countdown time is displayed for artworks on home feed. This is due to the fact that we use the “wrong” endAt [value](https://github.com/artsy/eigen/blob/77d0dbd98b1f3f212ff57d9064edfda4c006bdb4/src/app/Components/ArtworkRail/ArtworkRailCard.tsx#L84) when popcorn bidding is enabled for the auction.

#### Auction with enabled popcorn bidding
| Before | After |
| <video src="https://user-images.githubusercontent.com/3513494/231813938-4ee9ded3-b4f1-40cb-944c-07b13a00e881.mp4" />  | <video src="https://user-images.githubusercontent.com/3513494/231814008-4277024c-acb9-48e5-a632-d468cf4cface.mp4" />  | 

### PR Checklist

### To the reviewers 👀

- ix inconsistency in auction countdown for artworks on home feed - dimatretyak

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[FX-4728]: https://artsyproduct.atlassian.net/browse/FX-4728?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Components/ArtworkRail/ArtworkRailCard.tsx
+++ b/src/app/Components/ArtworkRail/ArtworkRailCard.tsx
@@ -1,4 +1,4 @@
-import { Flex, HeartFillIcon, HeartIcon, Text, useColor } from "@artsy/palette-mobile"
+import { Flex, HeartFillIcon, HeartIcon, Text, useColor, Touchable } from "@artsy/palette-mobile"
 import { themeGet } from "@styled-system/theme-get"
 import {
   ArtworkRailCard_artwork$data,
@@ -12,7 +12,6 @@ import { useSaveArtwork } from "app/utils/mutations/useSaveArtwork"
 import { Schema } from "app/utils/track"
 import { sizeToFit } from "app/utils/useSizeToFit"
 import { compact } from "lodash"
-import { Touchable } from "@artsy/palette-mobile"
 import { useMemo } from "react"
 import { GestureResponderEvent, PixelRatio } from "react-native"
 import { graphql, useFragment } from "react-relay"
@@ -75,14 +74,26 @@ export const ArtworkRailCard: React.FC<ArtworkRailCardProps> = ({
   const fontScale = PixelRatio.getFontScale()
   const artwork = useFragment(artworkFragment, restProps.artwork)
 
-  const { artistNames, date, id, image, internalID, isSaved, partner, slug, title } = artwork
+  const {
+    artistNames,
+    date,
+    id,
+    image,
+    internalID,
+    isSaved,
+    partner,
+    slug,
+    title,
+    sale,
+    saleArtwork,
+  } = artwork
 
   const saleMessage = defaultSaleMessageOrBidInfo({ artwork, isSmallTile: true })
 
-  const urgencyTag =
-    artwork?.sale?.isAuction && !artwork?.sale?.isClosed
-      ? getUrgencyTag(artwork?.sale?.endAt)
-      : null
+  const extendedBiddingEndAt = saleArtwork?.extendedBiddingEndAt
+  const lotEndAt = saleArtwork?.endAt
+  const endAt = extendedBiddingEndAt ?? lotEndAt ?? sale?.endAt
+  const urgencyTag = sale?.isAuction && !sale?.isClosed ? getUrgencyTag(endAt) : null
 
   const primaryTextColor = dark ? "white100" : "black100"
   const secondaryTextColor = dark ? "black15" : "black60"
@@ -414,6 +425,8 @@ const artworkFragment = graphql`
       currentBid {
         display
       }
+      endAt
+      extendedBiddingEndAt
     }
     partner {
       name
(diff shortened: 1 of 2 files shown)


src/app/Components/ArtworkRail/ArtworkRailCard.tsx
```diff
@@ -274,8 +298,24 @@ export const ArtworkRailCard: React.FC<ArtworkRailCardProps> = ({
...
+  const websocketEnabled = !!artwork.sale?.extendedBiddingIntervalMinutes
+  const socketChannelInfo: AuctionWebsocketChannelInfo = {
+    channel: "SalesChannel",
+    sale_id: artwork.sale?.internalID,
+  }
+
+  return (
+    <AuctionWebsocketContextProvider channelInfo={socketChannelInfo} enabled={websocketEnabled}>
+      <ArtworkRailCardInner {...props} artwork={artwork} />
+    </AuctionWebsocketContextProvider>
```
dimatretyak, line 313: Usually we use `AuctionWebsocketContextProvider` **per page** (like for [Artwork](https://github.com/artsy/eigen/blob/07007f404286907636c793c514cade5847c4fdef/src/app/Scenes/Artwork/Artwork.tsx#L469), [Sale](https://github.com/artsy/eigen/blob/07007f404286907636c793c514cade5847c4fdef/src/app/Scenes/Sale/Sale.tsx#L264)). But in this case `AuctionWebsocketContextProvider` is used **per artwork**

src/app/Components/ArtworkRail/ArtworkRailCard.tsx
```diff
@@ -73,16 +82,31 @@ export const ArtworkRailCard: React.FC<ArtworkRailCardProps> = ({
...
+    sale,
+    saleArtwork,
+  } = artwork
 
-  const saleMessage = defaultSaleMessageOrBidInfo({ artwork, isSmallTile: true })
+  const { currentBiddingEndAt } = useArtworkBidding({
+    lotID: saleArtwork?.lotID,
+    lotEndAt: saleArtwork?.endAt,
+    biddingEndAt: saleArtwork?.extendedBiddingEndAt,
+  })
```
dimatretyak, line 105: As far as I understand, "extended" timer state is not supported on other grids (for example, artist artworks grids. see the attached video).

src/app/Components/ArtworkRail/ArtworkRailCard.tsx
lordkiz: Yes, this might be an issue. We use actioncable in the backend backed by redis. With this implementation we will be establishing a connection for each artwork in a sale as opposed to single connection per sale. Assuming 10 artworks per rail x number of users online, it will be too much connections than we can support I think. In dev env redis would typically support about 10 connections, I think. 
I may be wrong. Pinging @mzikherman for input.

dblandin: I might be missing something but is there a way to use the right endAt timestamp (extendedBiddingEndAt), without adding the specific websocket support in this new context? I think that would address the bug without introducing additional risk/complexity.

dimatretyak: > without adding the specific websocket support in this new context

In this case, if someone made bid and the end time was extended, then we will face [the same problem](https://github.com/artsy/eigen/pull/8504#discussion_r1165561664)

dimatretyak: > I don't think every surface is expected to instantaneously respond to dynamically extended end times

ok, if others agree with this, then I don't mind using only [these changes](https://github.com/artsy/eigen/pull/8504/commits/661889173b71efae2a2dacc86840659cda7d7b2d) 👍

lordkiz: > ok, if others agree with this, then I don't mind using only [these changes](https://github.com/artsy/eigen/pull/8504/commits/661889173b71efae2a2dacc86840659cda7d7b2d) 👍

@dimatretyak I think these changes you have highlighted are sufficient. Reacting to extended bid time outside of auction and artwork page is too costly.

dimatretyak: Let's keep only [these changes](https://github.com/artsy/eigen/pull/8504/commits/661889173b71efae2a2dacc86840659cda7d7b2d). Thank y'all for discussion 🙏

PR #8505, opened 2023-04-13
fix(NX-3507): use the right backgroundColor for counterOffer banner

araujobarret:
This PR resolves [NX-3507]

### Description

Fixes the background color of the offer CTA inside a conversation. Also, minor fixes where other areas were using the non-existing color `copper`, changed all to `orange150`.

Before

https://user-images.githubusercontent.com/15792853/231761750-a4f0c1d6-f109-4c89-b1c3-9e1fd5202d30.mov

After

Screenshot 2023-04-13 at 14 38 40

### PR Checklist

### To the reviewers 👀

- fix: use the orange150 color for the offer CTA inside a conversation - araujobarret

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[NX-3507]: https://artsyproduct.atlassian.net/browse/NX-3507?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/Inbox/Components/Conversations/Composer.tests.tsx
+++ b/src/app/Scenes/Inbox/Components/Conversations/Composer.tests.tsx
@@ -1,9 +1,9 @@
 import { Flex } from "@artsy/palette-mobile"
 import { waitFor } from "@testing-library/react-native"
 import { ComposerTestsQuery } from "__generated__/ComposerTestsQuery.graphql"
+import { Button } from "app/Components/Button"
 import { extractText } from "app/utils/tests/extractText"
 import { renderWithWrappersLEGACY } from "app/utils/tests/renderWithWrappers"
-import { Button } from "app/Components/Button"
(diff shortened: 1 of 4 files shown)


lidimayra:
Great work!! I understand `copper100` is a color definition we had in the past and it was removed at some point, right?

I see some other references when [searching](https://github.com/artsy/eigen/search?q=copper100) for `copper100` in eigen, shall we take the chance to fix it everywhere? Or would it be better to handle the other ones separately?

lidimayra (approved):
Approving the PR as it already fixes a critical issue (and the question mentioned before might be out of scope here)

araujobarret:
I think it's safe to move on with this PR and address separately the other copper cases, for example the event message I changed here was using the black fallback, only our banner became transparent instead of black or something else, maybe because the fallback backgroundColor for a button is white.

PR #8524, opened 2023-04-14
fix(webview): Prevent double loads of pages, and fixes back button

damassi:
### Description

Fixes an issue we've had for a while with double WebView loading and broken back button states.

What this does is immediately prevents the subsequent webview link from loading if it meets certain conditions, and will jump backwards in the stack. By the time the user has it the actual back button, the webview has stabilized at the place where it first departed from to launch the native view. 

cc @artsy/mobile-platform 

https://user-images.githubusercontent.com/236943/232156389-00fd5820-157e-4195-b0f3-971822b158a0.mp4

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Components/ArtsyWebView.tsx
+++ b/src/app/Components/ArtsyWebView.tsx
@@ -13,6 +13,7 @@ import { dismissModal, goBack, GoBackProps, navigate } from "app/system/navigati
 import { ArtsyKeyboardAvoidingView } from "app/utils/ArtsyKeyboardAvoidingView"
 import { Schema } from "app/utils/track"
 import { useWebViewCallback } from "app/utils/useWebViewEvent"
+import { debounce } from "lodash"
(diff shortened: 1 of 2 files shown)


damassi:
@brainbicycle - there are likely some missed edge-cases here so if you wouldn't mind pulling down, but i think this points us in the right direction.

admbtlr:
This looks good to me. It doesn't work with the pageable artwork flag on, but that's not a priority for this release, so I would be in favour of merging this and making a ticket to fix that.

admbtlr (approved):

brainbicycle (approved):
Nice! tested this and it seems like a big improvement so I would also be in favor of merging, thanks!

We still have one issue I see on iOS, the view all links because they are webviews don't transition smoothly and override back button so you jump back to home on tapping back

PR #8526, opened 2023-04-17
fix: Update artist auction results initial filters logic

MrSltun:
This PR resolves [] 

### Description

This PR solves the issue of showing 1 applied filter in the Artist's auction results screen when in reality there are no filters applied to the auction results

The issue comes from the initial filters obtained URL query params coming from the Price Database screen

### Screenshots
| Before | After |
| --- | --- |
| image | image |

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- update artist auction results initial filters logic - mrsltun

</details>

### DIFF
--- a/src/app/Components/Artist/ArtistInsights/ArtistInsightsAuctionResults.tsx
+++ b/src/app/Components/Artist/ArtistInsights/ArtistInsightsAuctionResults.tsx
@@ -115,15 +115,12 @@ const ArtistInsightsAuctionResults: React.FC<Props> = ({
   })
 
   useEffect(() => {
-    let filters: FilterArray = []
+    setFilterTypeAction("auctionResult")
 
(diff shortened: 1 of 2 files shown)


olerichter00 (approved):
Thanks for the fix 🙇.

I wonder if we could get rid of the `useEffect` at all by passing the initial filters as prop to the filter provider. I know that this does not work at the moment. But it might be worth investigating why it doesn`t work. Maybe we just need to add an `applyFiltersAction()` when this prop is passed 🤔

MrSltun:
@olerichter00 I liked your suggestions and gave it a few minutes to see if I could do it quickly, but it turns out it needs more time to investigate and fix it properly.
But for now, do you mind if we merge this one and do it in another PR?

PR #8528, opened 2023-04-17
chore: update podfile.lock

MounirDhahri:
This PR resolves [] 

### Description

- getting this diff after running `yarn install:all`

#nochangelog


gkartalis (changes requested):
@MounirDhahri If you run the following the diff should go away:

```bash
cd ios && bundle exec pod update glog && bundle exec pod update RCT-Folly
```

MounirDhahri:
niiice! thanks George. I will

PR #8531, opened 2023-04-17
fix: Update running beta deploys check

MrSltun:
This PR resolves [] 

### Description
This PR updates the curl function to check the CI status of the currently running deploy

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/scripts/deploy-beta-android
+++ b/scripts/deploy-beta-android
@@ -3,12 +3,10 @@ set -euxo pipefail
 
 
 source ./scripts/source-for-bash-env
-source ./.env.shared &> /dev/null
 
 # Check if there is a build in progress
 echo "Checking if there is a build in progress..."
-build_status=$(curl -s -u ${EIGEN_CHECK_STATUS}: https://circleci.com/api/v1.1/project/github/artsy/eigen/tree/beta-android | jq -r '.[0].status')
-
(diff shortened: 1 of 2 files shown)


scripts/deploy-beta-android
```diff
@@ -3,12 +3,10 @@ set -euxo pipefail
 
 
 source ./scripts/source-for-bash-env
-source ./.env.shared &> /dev/null
```
MrSltun, line 6: removing this source command since `.env.shared` doesn't exist in circle ci's context, and the `&> /dev/null` skips printing the error message in the logs

MounirDhahri: **issue:** this was needed locally to make sure that we have the env vars when creating the beta locally. How about we skip this in CI

MounirDhahri (changes requested):

scripts/deploy-beta-android
MrSltun: Actually, we don't need this source since the only env var that we need (`$LOCAL_BRANCH`) is exported from `./scripts/source-for-bash-env` [link to code](https://github.com/artsy/eigen/blob/main/scripts/source-for-bash-env#L7)

MounirDhahri: truee, the status is available without token! thanks

MounirDhahri (approved):

PR #8532, opened 2023-04-17
chore(registry): Alphabetize AppRegistry definition

damassi:
### Description

Noticed that our app registry got a bit out of order so this alphabetizes things so that screens are easier to find through time. 

cc @artsy/mobile-platform


brainbicycle (approved):
🍎 🦓

PR #8548, opened 2023-04-20
fix: small improvements

MounirDhahri:
### Description

Was testing some weirdness in the App and made some small adjustments here that could be nice to have in main. Let me know what you think

#nochanges


src/app/Scenes/Inbox/Components/Conversations/Messages.tsx
```diff
@@ -165,7 +165,7 @@ export const Messages: React.FC<Props> = forwardRef((props, ref) => {
         inverted
         ref={flatList}
         keyExtractor={(group) => {
-          return group[0].id
+          return group[0].__id
```
MounirDhahri, line 168: id was undefined and it was breaking the virtualization logic - potentially leading to performance issues. __id is unique and will be always there from relay

araujobarret (approved):
🙏

PR #8558, opened 2023-04-21
feat(progressive onboarding): context

laurabeth:
This PR resolves [] 

### Description

Creates the context that will govern progressive onboarding behaviors. No UI created as yet. Up to date w/ latest behaviors in force. 

Next steps will be to implement the necessary tooltip-like UI components, then each user workflow: saves, follows and alerts.

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Components/ProgressiveOnboarding/ProgressiveOnboardingContext.tsx
+++ b/src/app/Components/ProgressiveOnboarding/ProgressiveOnboardingContext.tsx
@@ -18,7 +18,7 @@ interface DismissedKeyStatus {
 
 const ProgressiveOnboardingContext = createContext<{
   dismissed: DismissedKey[]
-  dismiss: (key: ProgressiveOnboardingKey | readonly ProgressiveOnboardingKey[]) => void
+  dismiss: (key: ProgressiveOnboardingKey | ProgressiveOnboardingKey[]) => void
   isDismissed: (key: ProgressiveOnboardingKey) => DismissedKeyStatus
 }>({
   dismissed: [],
@@ -56,8 +56,14 @@ export const ProgressiveOnboardingProvider: FC = ({ children }) => {
   const mounted = useDidMount()
 
   const isDismissed = useCallback(
-    (key: DismissedKey) => {
-      return !mounted || dismissed.includes(key)
+    (key: ProgressiveOnboardingKey) => {
+      if (!mounted) return { status: false, timestamp: 0 }
+
+      const dismissedKey = dismissed.find((d) => d.key === key)
+
+      return dismissedKey
+        ? { status: true, timestamp: dismissedKey.timestamp }
+        : { status: false, timestamp: 0 }
     },
     [dismissed, mounted]
   )
@@ -167,7 +173,7 @@ export const __dismiss__ = async (
 ) => {
   const keys = Array.isArray(key) ? key : [key]
 
-  keys.forEach((key) => {
+  keys.forEach(async (key) => {
     const item = await AsyncStorage.getItem(localStorageKey(id))
     const dismissed = parse(item)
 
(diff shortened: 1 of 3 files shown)


src/app/Components/ProgressiveOnboarding/ProgressiveOnboardingContext.tsx
```diff
@@ -0,0 +1,200 @@
...
+  })
+
+  return parse(item)
+}
+
+export const reset = async (id: string) => {
+  await AsyncStorage.removeItem(localStorageKey(id)).catch((e) =>
+    console.log("Error writing reset to local storage", e)
+  )
+}
```
dzucconi, line 200: I understand that most of this is copied out of Force and that these steps will almost certainly change. But I would strongly suggest adding some spec coverage to this. [The specs there](https://github.com/artsy/force/blob/main/src/Components/ProgressiveOnboarding/__tests__/ProgressiveOnboardingContext.jest.tsx) should provide a decent starting point.

src/app/utils/hooks/useDidMount.ts
```diff
@@ -0,0 +1,12 @@
...
+export const useDidMount = () => {
+  const isMountedRef = useRef(false)
+  useEffect(() => {
+    isMountedRef.current = true
+    return () => {
+      isMountedRef.current = false
+    }
+  }, [])
+  return isMountedRef
+}
```
dzucconi, line 12: Why add this hook?

dzucconi (commented):
I don't really have a problem with this, but I'd probably approach this the other way around: get a single simple tooltip displaying first before trying to predict what you might need to support it.

src/app/Components/ProgressiveOnboarding/ProgressiveOnboardingContext.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+const ProgressiveOnboardingContext = createContext<{
+  dismissed: DismissedKey[]
+  dismiss: (key: ProgressiveOnboardingKey | ProgressiveOnboardingKey[]) => void
+  isDismissed: (key: ProgressiveOnboardingKey) => DismissedKeyStatus
+}>({
+  dismissed: [],
+  dismiss: () => {},
+  isDismissed: (_key) => ({ status: false, timestamp: 0 }),
+})
```
damassi, line 27: This is non-blocking, but worth noting that in Eigen we use easy-peasy, and there's a much nicer way to manage local contexts should you wish to use it:

https://easy-peasy.vercel.app/docs/api/create-context-store.html

I would suggest going this route, but can also understand not wanting to refactor.

src/app/Components/ProgressiveOnboarding/ProgressiveOnboardingContext.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+  keys.forEach(async (key) => {
+    const item = await AsyncStorage.getItem(localStorageKey(id))
+    const dismissed = parse(item)
+
+    await AsyncStorage.setItem(
+      localStorageKey(id),
+      JSON.stringify(uniqBy([...dismissed, { key, timestamp }], ({ key }) => key))
+    )
+  })
```
MounirDhahri, line 184: **thought:** I see you are here trying to set a value for the same key (`localStorageKet(id)`) in AsyncStorage many times in sequence. Is it possible to do that only one time when you know the final value?

damassi (changes requested):
We should be using easy-peasy for the context, per @MounirDhahri's note about out of the box persistence, etc. It will simplify this up and down.

src/app/Components/ProgressiveOnboarding/ProgressiveOnboardingContext.tsx
laurabeth: Hey guys! Stepped away from this for future friday, but I'll be back to it shortly. Great feedback and I don't mind the refactor at all now that I've had a little breathing room from the work.

Thanks for taking a look.

PR #8560, opened 2023-04-21
refactor(misc): Cleanup global store imports, speed up type-checking, simplify detect-secrets

damassi:
### Description

This does a few things:
- Cleans up GlobalStore.tsx. Lots of unrelated stuff was in there
- Fixes tsc type-checking speed. We were compiling our sourcecode when all we needed was type-checking. Babel compiles
- Simplifies our detect-secrets config and makes it consistent with rest of artsy

cc @artsy/mobile-platform 

### PR Checklist


src/app/Components/Tabs/NavigationalTabs.tsx
```diff
@@ -20,6 +20,7 @@ export const NavigationalTabs: React.FC<TabsProps> = ({ onTabPress, activeTab, t
   return (
     <TabBarContainer scrollEnabled activeTabIndex={activeTab} tabLayouts={tabLayouts}>
       {tabs.map(({ label, visualClues }, index) => {
+        // eslint-disable-next-line react-hooks/rules-of-hooks
         const { showVisualClue } = useVisualClue()
```
brainbicycle, line 24: is it a problem actually that we are using hooks here?

tsconfig.json
```diff
@@ -5,22 +5,12 @@
     "incremental": true, // improved build speeds.
     "baseUrl": "src", // for imports like `"app/Components"` to resolve correctly.
     "experimentalDecorators": true, // used for class decorators. it would be cool to remove this if we ever get rid of classes.
+    "isolatedModules": false,
+    "noEmit": true,
```
damassi, line 9: These two speed things up quite a bit once the type-checker has warmed up

src/app/Components/Tabs/NavigationalTabs.tsx
damassi: It is according to the rules of hooks, but in practice, no, because the iteration through the loop is in effect creating a new component to return. 

But this does raise the fact that this kind of thing is _everywhere_ in eigen. It's bad, just in the sense that future react versions might actually break on this.

src/app/store/unsafe.ts
```diff
@@ -0,0 +1,101 @@
...
+  const tabState = __unsafe_mainModalStackRef.current
+    ?.getRootState()
+    .routes.find((r) => r.state?.type === "tab")?.state
+  if (!tabState) {
+    return "home"
+  } else {
+    const { index, routes } = tabState
+    return routes[index!].name as BottomTabType
+  }
+}
```
brainbicycle, line 21: this unsafe stuff, my memory may be wrong but I believe we had to use it in certain places because we couldn't use hooks in those contexts. The only reason I bring that up is because if that is the case, maybe it is not possible to remove completely from the codebase and if it isn't maybe a more descriptive place for these would be better? If I am wrong about that ignore me

src/app/Components/Tabs/NavigationalTabs.tsx
brainbicycle: alright, well outside the scope of this pr in any case

src/app/store/unsafe.ts
damassi: Our store init (both in energy and eigen) is fragile, and so this unsafe stuff is needed. Elsewhere where it was used is as a shortcut for things like class components, which don't use hooks (rather than wrapping the class component and injecting hooks). It all needs to be refactored out in a concentrated way, but not in here.

brainbicycle: makes sense!

damassi:
🍏

.lintstagedrc.mjs
```diff
@@ -1,5 +1,5 @@
 export default {
-  "*": ["yarn secrets:check:staged"],
+  "*": "yarn detect-secrets:hook",
```
gkartalis, line 2: 👏

package.json
```diff
@@ -23,6 +23,8 @@
     "danger": "danger",
     "dead-code": "./scripts/check-dead-code.sh",
     "deploy-beta": "./scripts/deploy-beta-both",
+    "detect-secrets:hook": "scripts/detect-secrets.sh hook",
+    "detect-secrets:rebuild": "scripts/detect-secrets.sh rebuild",
```
gkartalis, line 27: Thanks!!! I had added a [card](https://www.notion.so/artsy/Detect-secrets-f491a33d263841f782842e39e5340b56?pvs=4) on the eigen pain points doc this week, happy to see that resolved 👏

damassi: hah nice 😉

PR #8580, opened 2023-04-26
feat: update show rail card

MounirDhahri:
This PR resolves [CX-3631] 

### Description

This PR updates the show rail card to match the new designs
**PS:** The logic for showing the city has been discussed in this slack thread: https://artsy.slack.com/archives/C01B2P6LJUU/p1682410963288209

https://user-images.githubusercontent.com/11945712/234539551-7a293254-4816-4aa8-817a-52333889354d.mov

### PR Checklist

### To the reviewers 👀

- Adjust show rail cards design - mounir

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-3631]: https://artsyproduct.atlassian.net/browse/CX-3631?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Components/ShowCard.tsx
+++ b/src/app/Components/ShowCard.tsx
@@ -1,8 +1,9 @@
-import { Spacer, Flex, Text } from "@artsy/palette-mobile"
+import { Spacer, Flex, Text, Touchable } from "@artsy/palette-mobile"
+import { toTitleCase } from "@artsy/to-title-case"
 import { ShowCard_show$data } from "__generated__/ShowCard_show.graphql"
 import ImageView from "app/Components/OpaqueImageView/OpaqueImageView"
 import { navigate } from "app/system/navigation/navigate"
-import { Touchable } from "@artsy/palette-mobile"
+import { compact } from "lodash"
 import { GestureResponderEvent, ViewProps } from "react-native"
 import { createFragmentContainer, graphql } from "react-relay"
 
@@ -22,6 +23,13 @@ export const ShowCard: React.FC<ShowCardProps> = ({ show, onPress }) => {
     navigate(show.href!)
   }
 
+  const showCity = getShowCity({
+    showName: show.name,
+    showCity: show.city,
+    partnerCities: show.partner?.cities,
+    externalPartnerCity: show.partner?.city,
+  })
+
   return (
     <Flex width={WIDTH}>
       <Touchable haptic onPress={onTap}>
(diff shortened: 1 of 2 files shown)


src/app/Components/ShowCard.tsx
```diff
@@ -22,6 +23,13 @@ export const ShowCard: React.FC<ShowCardProps> = ({ show, onPress }) => {
     navigate(show.href!)
   }
 
+  const showCity = getShowCity({
```
MounirDhahri, line 26: sending props separately instead of the entire show and extracting the function to make it easier to test the component

src/app/Components/ShowCard.tsx
```diff
@@ -32,15 +40,51 @@ export const ShowCard: React.FC<ShowCardProps> = ({ show, onPress }) => {
...
+        return showName.toLowerCase().includes(partnerCity?.toLowerCase())
+      }
+    })
+
+    return matchingCity ? toTitleCase(matchingCity) : null
+  }
+
+  if (externalPartnerCity && showName.toLowerCase().includes(externalPartnerCity?.toLowerCase())) {
+    return toTitleCase(externalPartnerCity)
+  }
```
MounirDhahri, line 83: only one of them should exist and no prioritisation logic is needed here

src/app/Components/ShowCard.tsx
olerichter00: Hmm, intuitively I'd say it would be easier to read to just pass `show` instead of creating a new object for the parameter. But I'm happy to just leave it as it is. 

Another suggestion I'd have to extract logic out of the component is to have a function that returns the formatted string. Something like: `formattedCityAndDates` that returns `compact([showCity, `${show.formattedStartAt}-${show.formattedEndAt}`]).join(" • ")`.

Hmm, intuitively I would say it would be easier to read to just pass `show` instead of creating a new object for the parameter. But I'm happy to just leave it as it is.

Another suggestion I would have for extracting the logic from the component would be a function that returns the formatted string. Something like: `formattedCityAndDate` that returns `compact([showCity, `${show.formattedStartAt}-${show.formattedEndAt}`]).join(" - ")`.

We could than simplify the component to:

```
...
<Text>
  {formatCityAndDate(city)}
</Text>
...
```

olerichter00: Instead of testing the `getShowCity` function, we could test the entire component without having to shape the helper function parameters in a particular way.

olerichter00 (approved):
Looks good 🌟 

I just added some thoughts around implementation details. Feel free to merge 🚂

src/app/Components/ShowCard.tsx
MounirDhahri: the main reason I only sent the arguments needed here was to avoid the need to mock other show fields that are not required for the logic. It was probably too much optimization since they're 4 arguments and once can debate it makes the code less readable. 
I do like `formatCityAndDate` suggested method.  will update to use that to make the code easier

MounirDhahri: I am also writing a test for the show card - this way we can have an integration test along with the unit test for the city logic

MounirDhahri: nice - idea, i did it now

olerichter00 (approved):

PR #8582, opened 2023-04-26
feat(CX-3635):update my collection sticky tab 

MounirDhahri:
This PR resolves [CX-3635] 

### Description

- Update my collection sticky tab 

_At a next ticket, we will need to create some context to tell which tab is open - will leave that for a futher ticket_

https://user-images.githubusercontent.com/11945712/234909997-e0dd276c-e6db-4fa9-bd6c-ac5c65bc4132.mov

### PR Checklist

### To the reviewers 👀

- update my collection sticky tab - mounir

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-3635]: https://artsyproduct.atlassian.net/browse/CX-3635?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lordkiz (approved):
Looks v. good. Nice animations!

src/app/Scenes/MyCollection/Components/MyCollectionStickyHeader.tsx
```diff
@@ -0,0 +1,328 @@
...
+
+const Messages = ({
+  hasMarketSignals,
+  showNewWorksMessage,
+  showSubmissionMessage,
+}: {
+  hasMarketSignals: boolean
+  showNewWorksMessage: boolean
+  showSubmissionMessage: boolean
+}) => {
```
damassi, line 300: #superminor - Any reason these aren't React.FCs?

src/app/utils/hooks/useMeasure.ts
```diff
@@ -0,0 +1,36 @@
...
+  pageX: number
+  pageY: number
+}
+
+export const useMeasure = ({ ref }: { ref: MutableRefObject<any> }) => {
+  const [position, setPosition] = useState<null | Position>(null)
+
+  useEffect(() => {
+    if (ref.current) {
+      ;(ref.current as any).measure(
```
damassi, line 17: the `;(...)` is somewhat prone to breakage; can we assign this to a variable?

damassi:
Animations look pretty good! I wonder if we might be able to shorten the x translation a bit to eliminate some of the opacity / outline overlap: 

Screenshot 2023-04-27 at 10 19 31 AM

A lot of the time with animations a slight nudge in a particular direction can trick the user into thinking that its going further than it seems, like an optical illusion; the trick is getting the timing right. But in this case, i think shortening the distance might be all that's needed. Otherwise, looks great. (And this is just IMO, feel free to ignore)

src/app/Scenes/MyCollection/Components/MyCollectionStickyHeader.tsx
MounirDhahri: no reason actually 😄 I will update it. Thank you

src/app/utils/hooks/useMeasure.ts
MounirDhahri: Sure - will update

src/app/utils/hooks/useMeasure.ts
```diff
@@ -0,0 +1,36 @@
...
+            pageX,
+            pageY,
+          })
+        },
+        () => {
+          console.error("measurement failed")
+        }
+      )
+    }
+  }, [JSON.stringify(position)])
```
olerichter00, line 33: Shouldn't `position` work here too? When we change the position, we create a new object with `setPosition` anyway.

```suggestion
  }, [position])
```

src/app/Scenes/MyCollection/Components/MyCollectionStickyHeader.tsx
```diff
@@ -0,0 +1,328 @@
...
+  showNewWorksMessage,
+  showSeparator,
+}) => {
+  const { showVisualClue } = useVisualClue()
+  const [selectedTab, setSelectedTab] = useState<CollectedTab>(null)
+
+  const showSubmissionMessage = showVisualClue("ArtworkSubmissionMessage")
+  const enableCollectedArtists = useFeatureFlag("AREnableMyCollectionCollectedArtists")
+
+  const showFilters = true
```
olerichter00, line 58: 👀

olerichter00 (approved):
Great work 🌟

src/app/Scenes/MyCollection/Components/MyCollectionStickyHeader.tsx
MounirDhahri: I am removing it now after pushing the changes Chris reported

src/app/utils/hooks/useMeasure.ts
MounirDhahri: that's a good comment - i should have probably explained further why I am tricking this effect. 
The main reason behind me serializing the json is to make sure we're applying a shallow comparison the position object here. In the video attached, you can see the output of the console.log from both approaches.



Uploading Screen Recording 2023-04-28 at 12.52.34.mov…

src/app/Scenes/MyCollection/Components/MyCollectionStickyHeader.tsx
MounirDhahri: this is here to prepare for the next ticket about the filters - but better to be removed maybe until I pick up the other ticket

src/app/utils/hooks/useMeasure.ts
olerichter00: Thanks for the explanation 👍

MounirDhahri:
@damassi here we go. adjusting the padding didn't work - what was possible thought without making the UX worse was to wait for the spill to get blue ( 200 ms - the default spring animation duration) then do the translate X

https://user-images.githubusercontent.com/11945712/235132886-db09ac58-d5e7-4031-b5f9-a8968d261b94.mov

src/app/utils/hooks/useMeasure.ts
MounirDhahri: you're welcome of course 
sorry - the upload didn't work 

https://user-images.githubusercontent.com/11945712/235133362-a5c0b13b-f9c2-4e32-9af3-0c18d1ae2dfa.mov

olerichter00: Isn't there a risk to create an infinite loop when setting position inside of the effect?

MounirDhahri: yes, it's why I am making sure to compare the objects as a string. 
A better approach here would have been using a reference - issue is that the initial position always comes as 0. I can try it out

PR #8594, opened 2023-04-27
fix: only request push notifications once per login session

brainbicycle:
This PR resolves [MOPLAT-737] 

### Description

Adds a check to make sure we only request push permissions once per login session to not spam Android users with the settings dialog when they deny push permissions. 

**Follow-ups:**
Digging into this code revealed it is fairly buggy, probably at least partially my fault but having the logic split across native side and typescript side makes it hard to follow. Going to create a ticket to bring this all over to typescript, write some tests and hopefully fix some of these bugs:
- We are not refreshing tokens when the users push status is authorized
- Login is not the relevant event for our push logic, we should be checking it any time the app is launched
- Not a bug necessarily but now that Android has a similar push permissions model we should make the logic similar

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

- Fix push settings dialog showing many times - Brian 

#### Dev changes

-

</details>

[MOPLAT-737]: https://artsyproduct.atlassian.net/browse/MOPLAT-737?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ios/Artsy/App/ARAppNotificationsDelegate.m
+++ b/ios/Artsy/App/ARAppNotificationsDelegate.m
@@ -194,9 +194,9 @@
         @"action_name" : @"Cancel",
         @"context_screen"  : analyticsContext
     }];
-    [AROptions setBool:YES forOption:ARPushNotificationsAppleDialogueRejected];
 #if (TARGET_IPHONE_SIMULATOR == 0)
     ARErrorLog(@"Error registering for remote notifications: %@", error.localizedDescription);
+    [AROptions setBool:YES forOption:ARPushNotificationsAppleDialogueRejected];
 #endif
 }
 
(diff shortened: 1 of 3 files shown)


damassi (approved):

ios/Artsy/App/ARAppNotificationsDelegate.m
```diff
@@ -194,9 +194,9 @@ - (void)application:(UIApplication *)application didFailToRegisterForRemoteNotif
         @"action_name" : @"Cancel",
         @"context_screen"  : analyticsContext
     }];
-    [AROptions setBool:YES forOption:ARPushNotificationsAppleDialogueRejected];
 #if (TARGET_IPHONE_SIMULATOR == 0)
     ARErrorLog(@"Error registering for remote notifications: %@", error.localizedDescription);
+    [AROptions setBool:YES forOption:ARPushNotificationsAppleDialogueRejected];
```
brainbicycle, line 199: noticed that on simulator this will always fail so this bool gets set making the settings dialog show, which is confusing. Probably a good idea to test push on device always but moving this behind the check should make it a little less confusing.

PR #8606, opened 2023-05-02
feat(MOPLAT-619): Context menu for artwork items

gkartalis:
This PR resolves [MOPLAT-619] 

### Description

- [x] Adds context menu on ArtworkGridItem and ArtworkRailCard
    - [x] user is able to create alert
    - [x] user is able to like / dislike the artwork
    - [x] user is able to view artwork in room
- [x] Behind a FF and enabled only on iOS (for now)
- [x] refactored some tests that were type complaining

### Demo

https://github.com/artsy/eigen/assets/21178754/4d23f60f-950a-4768-8ed9-ddfc005b1948

#### TODO:

1. ✅ (Done) there seems to be something that relay doesn't like: 

```bash
Warning: RelayResponseNormalizer: Invalid record. The record contains two instances of the same id: 
`QXJ0d29yazo2NDM4NzhlYWE0NzBhMzAwMGM3YzRmMGM=` with conflicting field, widthCm and its values: 132 and 
null. If two fields are different but share the same id, one field will overwrite the other.
```

not sure how it affects the app, better to solve it before merging

2. ✅ (Done) Will refactor the ArtworkGridItem.tests.tsx since they are not properly written and all of them break. 

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

- Context menu for artwork items - gkartalis

-

#### Dev changes

-

</details>

[MOPLAT-619]: https://artsyproduct.atlassian.net/browse/MOPLAT-619?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/Components/ArtworkGrids/ArtworkGridItem.tsx
```diff
@@ -102,6 +104,24 @@ export const Artwork: React.FC<ArtworkProps> = ({
...
+      href: artwork.href!,
+      isSaved: artwork.isSaved!,
+      internalID: artwork.internalID!,
+      slug: artwork.slug!,
+      id: artwork.id!,
+      isHangable: artwork.isHangable!,
+      heightCm: artwork.heightCm!,
+      widthCm: artwork.widthCm!,
+    },
+  })
```
damassi, line 123: Any reason these props need to be passed one by one? Why not 
```tsx
const artworkQuickActions = useArtworkItemContextMenu(artwork)
```
?

src/app/Components/ArtworkRail/ArtworkRailCard.tsx
```diff
@@ -74,6 +74,24 @@ export const ArtworkRailCard: React.FC<ArtworkRailCardProps> = ({
...
+      href: artwork.href!,
+      isSaved: artwork.isSaved!,
+      internalID: artwork.internalID!,
+      slug: artwork.slug!,
+      id: artwork.id!,
+      isHangable: artwork.isHangable!,
+      heightCm: artwork.heightCm!,
+      widthCm: artwork.widthCm!,
+    },
+  })
```
damassi, line 93: Ditto

src/app/Components/ContextMenuTouchable/ContextMenuTouchable.tsx
```diff
@@ -0,0 +1,97 @@
+import { Flex, useColor } from "@artsy/palette-mobile"
```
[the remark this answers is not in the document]
gkartalis: sounds great

src/app/Components/ContextMenuTouchable/ContextMenuTouchable.tsx
```diff
@@ -0,0 +1,97 @@
...
+import React from "react"
+import { TouchableHighlight, TouchableHighlightProps } from "react-native"
+import ContextMenu, { ContextMenuAction, ContextMenuProps } from "react-native-context-menu-view"
+import Haptic, { HapticFeedbackTypes } from "react-native-haptic-feedback"
+
+interface ContextAction extends Omit<ContextMenuAction, "subtitletitle"> {
```
damassi, line 7: `subtitletitle` 😆

src/app/Components/ContextMenuTouchable/ContextMenuTouchable.tsx
```diff
@@ -0,0 +1,97 @@
...
+}) => {
+  const color = useColor()
+  const inner =
+    React.Children.count(children) === 1 ? children : <Flex flex={flex}>{children}</Flex>
+
+  const runHaptic = (pressFn: any | undefined) => {
+    if (pressFn !== undefined && haptic !== undefined) {
+      Haptic.trigger(haptic === true ? "impactLight" : haptic)
+    }
+  }
```
damassi, line 41: No need to pass `onPress` as its already within scope as props

src/app/Components/ContextMenuTouchable/ContextMenuTouchable.tsx
```diff
@@ -0,0 +1,97 @@
...
+  haptic,
+  onPress,
+  onLongPress,
+  ...props
+}) => {
+  const color = useColor()
+  const inner =
+    React.Children.count(children) === 1 ? children : <Flex flex={flex}>{children}</Flex>
+
+  const runHaptic = (pressFn: any | undefined) => {
```
damassi, line 37: `triggerHaptic`; corresponds more 1:1 with native method `Haptic.trigger`

src/app/Components/ContextMenuTouchable/ContextMenuTouchable.tsx
```diff
@@ -0,0 +1,97 @@
...
+    if (pressFn !== undefined && haptic !== undefined) {
+      Haptic.trigger(haptic === true ? "impactLight" : haptic)
+    }
+  }
+
+  const onPressWrapped: TouchableHighlightProps["onPress"] = (e) => {
```
damassi, line 43: No need to add additional info to the handler -- On press, handle press. No need to encode more info in the name, as the handler assumes as much, that things are happening in the function. Keeping naming simple / generic / dumb makes future changes easier

```tsx
const handlePress = (event) => {
  runHaptic()
  onPress?.(event)
}
```

src/app/Components/ContextMenuTouchable/ContextMenuTouchable.tsx
```diff
@@ -0,0 +1,97 @@
...
+    runHaptic(onPress)
+    onPress?.(e)
+  }
+
+  const onLongPressFnWrapped: TouchableHighlightProps["onLongPress"] = !isActions(onLongPress)
+    ? (e) => {
+        runHaptic(onLongPress)
+        onLongPress?.(e)
+      }
+    : undefined
```
damassi, line 53: Would suggest making this a little more stripped down and linear:

```tsx
cons handleLongPress = (event) => {
  if (isActions(onLongPress) return 
  
  triggerHaptic()
  onLongPress?.(event) 
}
```

src/app/Components/ContextMenuTouchable/ContextMenuTouchable.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  const contextActions = isActions(onLongPress)
+    ? onLongPress.map((action) => {
+        const { onPress: ignored, ...rest } = action
+        return { subtitletitle: "", ...rest }
+      })
+    : undefined
+
+  const contextOnPress: ContextMenuProps["onPress"] = isActions(onLongPress)
```
damassi, line 62:
```tsx
const handleContextOnPress = (event) => {
  if (!isActions(onLongPress)) return 
  ...
}
```

src/app/Components/ContextMenuTouchable/ContextMenuTouchable.tsx
damassi: Or actually, i see that there are other functions calling this with different types of callback functions being passed. 

Here's a q: why do we need to check this for undefined, etc? Can't we just assume that if a callback (like `handleOnPress` or `handleOnLongPress`) are being called, that's enough to assume that `Haptic.trigger` should be called? Feels a bit over-cautious to check for undefined in here and this function could be stripped down to just 

```tsx
const triggerHaptic = () => {
  Haptic.trigger(haptic === true ? "impactLight" : haptic)
}
```

src/app/Components/ContextMenuTouchable/ContextMenuTouchable.tsx
```diff
@@ -0,0 +1,97 @@
...
+        onLongPress?.(e)
+      }
+    : undefined
+
+  const contextActions = isActions(onLongPress)
+    ? onLongPress.map((action) => {
+        const { onPress: ignored, ...rest } = action
+        return { subtitletitle: "", ...rest }
+      })
+    : undefined
```
damassi, line 60: We should really strive to avoid the ternary or undefined pattern wherever we can, just about everywhere; a variable without a value is, by default, undefined; lean into that 😄. Assigning an `undefined` value is confusing through time. 

Also we're mixing types here which makes for more brittle code. Something like this keeps things consistent:

```tsx
const getContextActions = () => {
  if (!isActions(onLongPress)) return []
  
  const contextActions = onLongPress.map(action => {
    return {
      ...omit(action, "onPress"),
      subtitletitle: "", 
    }
  })
}
```

src/app/Components/ContextMenuTouchable/ContextMenuTouchable.tsx
```diff
@@ -0,0 +1,97 @@
...
+      }
+    : undefined
+
+  return contextActions !== undefined ? (
+    <ContextMenu actions={contextActions} onPress={contextOnPress}>
+      <TouchableHighlight
+        underlayColor={color("white100")}
+        activeOpacity={0.8}
+        {...props}
+        onPress={onPressWrapped}
```
damassi, line 77: `onPress={handlePress}`

src/app/Components/ContextMenuTouchable/ContextMenuTouchable.tsx
```diff
@@ -0,0 +1,97 @@
...
+        onLongPress={onLongPressFnWrapped}
```
damassi, line 78: `onLongPress={handleLongPress}`

src/app/utils/hooks/useArtworkItemContextMenu.ts
```diff
@@ -0,0 +1,120 @@
...
+            openViewInRoom()
+          })
+        },
+      })
+    }
+
+    return artworkQuickActions
+  }
+
+  return getArtworkQuickActions()
```
damassi, line 119: #superminor - for debugging and other code introspection its a helpful habit to assign and return, versus just return:
```tsx
const artworkQuickActions = getArtworkQuickActions()
return artworkQuickActions
```
Now a future dev could drop a console.log in between there, or add a debugger statement or otherwise manipulate things further / expand. 

Of course, some folks like to keep a smaller LOC count, but I've found this to be helpful for my future self in the same way that avoiding implicit returns is helpful for my future self

src/app/utils/hooks/useEnableContextMenu.tsx
```diff
@@ -0,0 +1,19 @@
...
+ * @returns {boolean} shouldShowContextMenu
+ */
+export function useEnableContextMenu() {
+  const isIOS = Platform.OS === "ios"
+  const enableContextMenu = useFeatureFlag("AREnableContextMenu")
+
+  const shouldShowContextMenu = isIOS && enableContextMenu
+
+  return shouldShowContextMenu
+}
```
damassi, line 19: Until this file spreads wider across the codebase, thoughts on moving the code inline into `useArtworkItemContextMenu`? 

Thinking.... we've got a bit of a "hooks problem" in Eigen, where every bit of functionality that's a hook seems to go into `utils` (or now `utils/hooks`), instead of being _promoted_ to `utils/hooks`; meaning, actually generally useful, versus hypothetically generally useful and used > 1-2 places. It's totally fine to colocate hooks files close to the call-site. 

For example:

```
/Components/ContextMenu/ContextMenuArtwork.tsx
/Components/ContextMenu/useContextMenu.tsx
```

src/app/Components/ArtworkGrids/ArtworkGridItem.tsx
damassi: I don't think it's necessary, I think its just a matter of importing the `Artwork_artwork$data` type into the hook! and then it should just work.

src/app/store/config/features.ts
```diff
@@ -291,6 +291,11 @@ export const features = defineFeatures({
     readyForRelease: false,
     showInDevMenu: true,
   },
+  AREnableContextMenu: {
```
damassi, line 294: `AREnableArtworkContextMenu`

src/app/Components/ContextMenu/ContextMenuArtwork.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+export type ContextMenuArtworkProps = Omit<TouchableHighlightProps, "onLongPress"> &
+  ExtraTouchableProps
+
+/**
+ * `haptic` can be used like:
+ * <Touchable haptic />
+ * or
+ * <Touchable haptic="impactHeavy" />
+ */
```
damassi, line 26: We should move this comment (or something like it) directly above the `haptic` prop in the interface; then it will show up in vscode hover, etc

src/app/Components/ContextMenu/ContextMenuArtwork.tsx
```diff
@@ -0,0 +1,112 @@
...
+  onPress,
+  onLongPress,
+  ...props
+}) => {
+  const color = useColor()
+  const inner =
+    React.Children.count(children) === 1 ? children : <Flex flex={flex}>{children}</Flex>
+
+  const triggerHaptic = () => {
+    if (!haptic) return
```
damassi, line 40: 👍

damassi (approved):
Looks great @gkartalis, thanks for working through all of those comments! If there aren't any more big changes incoming feel free to merge when you feel it's ready 👍 🤙

gkartalis:
Thanks for the insightful comments @damassi! 

The only issue that is left with this are some warnings coming from the relay store like this:

```
Warning: RelayResponseNormalizer: Invalid record. The record contains two instances of the same id: 
`QXJ0d29yazo2NDM4NzhlYWE0NzBhMzAwMGM3YzRmMGM=` with conflicting field, widthCm and its values: 132 and 
null. If two fields are different but share the same id, one field will overwrite the other.
```

Not sure if we should merge with this or not 🤷‍♂️. Will spent ~15 minutes in trying to solve them and will update

gkartalis:
reading [this](https://relay.dev/docs/debugging/inconsistent-typename-error/) which looks relevant

damassi:
oh yeah, lets fix this before merge. Seem that come up here and there, but not too sure.

src/app/Components/ArtworkGrids/ArtworkGridItem.tsx
```diff
@@ -427,6 +427,7 @@ export default createFragmentContainer(Artwork, {
   artwork: graphql`
     fragment ArtworkGridItem_artwork on Artwork
     @argumentDefinitions(includeAllImages: { type: "Boolean", defaultValue: false }) {
+      ...useCreateArtworkAlert_artwork
```
damassi, line 430: This is super nitpicky so apologies in advance 😅 but yah super good habit to keep all fragment spreads at the top of the query so they jump right out. A fragment that's spread is instantly more important than every other field because it creates a coupling between parent and child, and this is where stuff goes 💀 (Have had so many moments lost in the weeds where later I found that a fragment was being spread into the bottom and I didn't notice)

damassi (dismissed):

src/app/Components/ArtworkGrids/ArtworkGridItem.tsx
gkartalis: no worries at all 😅 makes sense! Will start doing it as well

brainbicycle (dismissed):

PR #8614, opened 2023-05-03
feat: Query and render real artwork lists for artwork lists screen

rajsam003:
This PR resolves [FX-4747]

This PR replaces the dummy data with real data for artwork lists screen  

- [x] Query and render real artwork lists data like Force does (artwork lists rails)
- [x] Implement infinite scrolling
- [x] Implement pull-to-refresh behaviour
- [x] Display loading skeleton when lists are loading

image

### PR Checklist

### To the reviewers 👀

- Artwork lists on Profile/Saves tab - Sam

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[FX-4747]: https://artsyproduct.atlassian.net/browse/FX-4747?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/Scenes/ArtworkLists/ArtworkListItem.tsx
```diff
@@ -31,21 +35,37 @@ export const ArtworkListItem = ({ artworkList }: ArtworksListsItemProps) => {
...
+const artworkListItemFragment = graphql`
+  fragment ArtworkListItem_collection on Collection {
+    name
+    internalID
+    artworksCount(onlyVisible: true)
+    artworksConnection(first: 4) {
+      edges {
+        node {
+          image {
+            url(version: "square")
```
dblandin, line 65: **suggestion / non-blocking**: What do you think of using `resized` here so we can more intentional with the image size rather than relying on the named template?

```graphql
  image {
    resized(height: 300, width: 300, version: "normalized") {
      url
    }
  }
```

src/app/Scenes/ArtworkLists/ArtworkListItem.tsx
```diff
@@ -31,21 +35,37 @@ export const ArtworkListItem = ({ artworkList }: ArtworksListsItemProps) => {
...
       </Flex>
     </Flex>
   )
 }
+
+const artworkListItemFragment = graphql`
+  fragment ArtworkListItem_collection on Collection {
+    name
+    internalID
+    artworksCount(onlyVisible: true)
```
dblandin, line 60: **question / non-blocking**: Long-term, I wonder if these counts make more sense underneath artworksConnection?

```graphql
artworksConnection(first: 4) {
  counts {
    total // 15
    visible // 10
  }
}
```

dblandin (dismissed):
LGTM overall! Left a couple non-blocking comments.

src/app/Scenes/ArtworkLists/ArtworkLists.tsx
```diff
@@ -1,23 +1,132 @@
...
+  <Suspense fallback={<ArtworkListsPlaceHolder />}>
+    <ArtworkLists />
+  </Suspense>
+)
+
+export const ArtworkListsPlaceHolder = () => {
+  const screen = useScreenDimensions()
+  return (
+    <StickyTabPageScrollView scrollEnabled={false} style={{ paddingTop: 20 }}>
+      <GenericGridPlaceholder width={screen.width - 40} />
```
dimatretyak, line 95: **question**: what do you think about creating a new placeholder component for artwork lists grid? The current one is slightly different (for example, in height). What do you think about this. is it worth spending time on this??

cc: @dblandin

https://user-images.githubusercontent.com/3513494/235911400-797eaa47-11d9-414c-9f33-4c1204c130d3.mp4

dimatretyak (dismissed):
Great job, looks good 🔥

src/app/Scenes/ArtworkLists/ArtworkListItem.tsx
rajsam003: It seems it doesn't have the field `counts` in `artworksConnection`. We can use `artworkCount` for now 🤷🏽‍♂️

image

src/app/Scenes/ArtworkLists/ArtworkLists.tsx
rajsam003: Its from the `<GenericGridPlaceholder/>` component. We would need a custom one.  I would suggest that we work on it on a different PR.

dimatretyak (approved):

anandaroop (dismissed):
Nice!

(I found a minor visual issue, but it's not directly related to your changes here, so I [dropped it in Slack instead](https://artsy.slack.com/archives/C9SATFLUU/p1683208799535019?thread_ts=1682539233.102409&cid=C9SATFLUU))

src/app/Scenes/ArtworkLists/ArtworkListItem.tsx
damassi: (:ceiling-cat: I think adding counts on a connection is a single helper away in MP, but i could be wrong)

PR #8513, opened 2023-04-14
fix: reanimated crashes fix pt2

gkartalis:
This PR resolves [] 

### Description

Fixed the reanimated crashes weirdness

### PR Checklist

- [x] I have tested my changes on **iOS** and **Android**.
- [x] I hid my changes behind a **[feature flag]**, or they don't need one.
- [x] I have included **screenshots** or **videos**, or I have not changed the UI.
- [x] I have added **tests**, or my changes don't require any.
- [x] I added an **[app state migration]**, or my changes do not require one.
- [x] I have documented any **follow-up work** that this PR will require, or it does not require any.
- [x] I have added a **changelog entry** below, or my changes do not require one.

### To the reviewers 👀

- [ ] I would like **at least one** of the reviewers to **run** this PR on the simulator or device.

<details><summary>Changelog updates</summary>

#### Cross-platform user-facing changes

- reanimated crashes - gkartalis

#### iOS user-facing changes

-

#### Android user-facing changes

-

#### Dev changes

-

</details>

Need help with something? Have a look at our [docs], or get in touch with us.

[app state migration]: ../blob/main/docs/adding_state_migrations.md
[feature flag]: ../blob/main/docs/developing_a_feature.md
[docs]: ../blob/main/docs/README.md

PR #8624, opened 2023-05-04
fix: remove all implicit returns in reanimated hooks

brainbicycle:
This PR resolves [] 

### Description

We believe that implicit returns are not playing well with babel resulting in our js functions in hooks not being converted to "worklets" which causes crashes when reanimated tries to run them on the wrong thread.

This converts all our reanimated hooks to have explicit returns.

https://artsy.slack.com/archives/C02NAJ2CGLW/p1683148878257549

props @gkartalis @damassi for sleuthing

# Follow-up:
- There is a function useCode() from reanimated used throughout our code that still has implicit returns, if we keep seeing issues maybe this needs updating as well

### PR Checklist

### To the reviewers 👀

- Hopefully fix reanimated crashes - Brian, Chris, George 

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/utils/animations/useSkeletonAnimation.tsx
+++ b/src/app/utils/animations/useSkeletonAnimation.tsx
@@ -26,11 +26,13 @@ export function useSkeletonAnimation({
     }
   }, [])
 
-  const animatedStyle = useAnimatedStyle(() => ({
-    opacity: enableSkeletonAnimation
-      ? interpolate(shared.value, [0, 1], [targetOpacityValue, 1])
-      : 1,
-  }))
+  const animatedStyle = useAnimatedStyle(() => {
+    return {
+      opacity: enableSkeletonAnimation
+        ? interpolate(shared.value, [0, 1], [targetOpacityValue, 1])
+        : 1,
+    }
+  })
 
   return animatedStyle
 }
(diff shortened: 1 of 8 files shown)


gkartalis:
I think `useCode` is okay, it was from reanimated 1, the problematic ones are the hooks from version 2

src/app/utils/animations/useSkeletonAnimation.tsx
```diff
@@ -26,11 +26,13 @@ export function useSkeletonAnimation({
...
-    opacity: enableSkeletonAnimation
-      ? interpolate(shared.value, [0, 1], [targetOpacityValue, 1])
-      : 1,
-  }))
+  const animatedStyle = useAnimatedStyle(() => {
+    return {
+      opacity: enableSkeletonAnimation
+        ? interpolate(shared.value, [0, 1], [targetOpacityValue, 1])
+        : 1,
+    }
```
gkartalis, line 34: I remember in one of the initial fixes I did there was also an issue coming from something like [this](https://github.com/artsy/eigen/pull/8513/files), what do you think of changing this to:

```suggestion
    if (enableSkeletonAnimation) {
      return {
        opacity: interpolate(shared.value, [0, 1], [targetOpacityValue, 1])
      }
    }
    return {
      opacity: 1
    }
```

just to be on the safe side of things

gkartalis (approved):

damassi (approved):

PR #8628, opened 2023-05-04
refactor(legacy-screen): Point at palette-mobile

damassi:
Related:
- https://github.com/artsy/palette-mobile/pull/99

### Description

This updates Eigen to use the `LegacyScreen` component from palette-mobile. Eventually this will be refactored to use the new `Screen`, but for now this eliminates some code duplication and will make the delete process cleaner. 

The one place we couldn't use it is in the `ArtQuizArtworks` screen, which pointed at the _new_-legacy screen from palette mobile, instead of the one internal to eigen. I think this was to make use of the `RawHeader`, which we can eliminate and work into the [new `Screen.Header` `rightElements` prop](https://github.com/artsy/eigen/blob/1cb81dd71083f393fd1d12ab7efc30a1aee19296/src/app/Scenes/ArtQuiz/ArtQuizArtworks.tsx#L171). Because of this, we're using the _actual_ new screen component in this view. Still need to test on the device but seems ok. 

Follow-up PR from here should be to refactor `LegacyScreen` from Eigen and point it at `Screen`, then delete `LegacyScreen` from `palette-mobile`.  Seems like it should just be a matter of updating the imports. 

cc @artsy/mobile-platform 

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>


gkartalis (approved):
looks good to go 👏 I will run the app locally to make sure everything is smooth

gkartalis:
Tested it out! lets merge 🔥

PR #8637, opened 2023-05-05
refactor(context): Decouple create alert and colocate context menu code

damassi:
### Description

This is a proposal to update the context menu stuff with a few changes:

- Decoupled the ArtworkAlert stuff from the context menu code. There are some interdependencies there, but when it comes down to it the only thing we need is a callback function signaling that we want to trigger save alert from menu. From there we can call up. 
- Created a `CreateArtworkAlertModal` component that has all of its data dependencies self contained so we don't need to have a hook separate from the component, which will allow us to spread `CreateArtworkAlertModal_artwork` in the same fashion that we'd expect with other relay components 
- Moved all of the code from `useArtworkContext` hook directly into `ContextMenuArtwork` as separating the two made less sense after we decoupled the alert stuff from the hook 

Let me know what you think, if this feels good, otherwise happy to close. 

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>


src/app/Components/ContextMenu/ContextMenuArtwork.tsx
```diff
@@ -1,112 +1,187 @@
...
+      disabled={!shouldDisplayContextMenu}
     >
-      {inner}
-    </TouchableHighlight>
+      <TouchableHighlight
+        underlayColor={color("white100")}
+        activeOpacity={0.8}
+        {...touchableProps}
+        onPress={handleTouchablePress}
+        onLongPress={handleTouchableLongPress}
```
damassi, line 181: I tried and I tried to get this to work, thinking we could trigger some bit of state and add padding around the children of the context menu while it expands, but no go :( For some reason the context menu seems to be capturing the long press touch event.

Tried lots of other things too, like using a setTimeout and then animating in using Moti, and other hacky attempts. Nothing worked.

src/app/Components/ContextMenu/ContextMenuArtwork.tsx
```diff
@@ -1,112 +1,187 @@
...
   haptic,
   onPress,
   onLongPress,
-  ...props
+  onCreateAlertActionPress,
+  ...touchableProps
 }) => {
   const color = useColor()
-  const inner =
-    React.Children.count(children) === 1 ? children : <Flex flex={flex}>{children}</Flex>
```
damassi, line 37: https://react.dev/reference/react/legacy. But more generally, would suggest avoiding stuff like this unless strictly necessary and to keep it simple. We can pass a prop (like `flex={1}`) and test for that props existence vs making things too implicit via hidden layers of view configuration logic

src/app/Components/ArtworkGrids/ArtworkGridItem.tsx
```diff
@@ -202,10 +194,11 @@ export const Artwork: React.FC<ArtworkProps> = ({
   return (
     <>
       <ContextMenuArtwork
-        haptic
+        onCreateAlertActionPress={() => setShowCreateArtworkAlertModal(true)}
+        artwork={artwork}
         onPress={handleTap}
         testID={`artworkGridItem-${artwork.title}`}
-        onLongPress={artworkQuickActions}
```
damassi, line 208: This is no longer needed now that we're computing the actions internal to the context menu component, but thinking here: props are cheap! since we're creating actions to pass, we could do 
```tsx
actions={contextMenuActions}
onLongPress={handleLongPress}
```
and keep categories completely separate (as well as types), vs merging the two together and then having to disambiguate inside the component

src/app/Components/ContextMenu/ContextMenuArtwork.tsx
```diff
@@ -1,112 +1,187 @@
...
+      onPress={handleContextPress}
+      disabled={!shouldDisplayContextMenu}
     >
-      {inner}
-    </TouchableHighlight>
+      <TouchableHighlight
+        underlayColor={color("white100")}
+        activeOpacity={0.8}
+        {...touchableProps}
+        onPress={handleTouchablePress}
```
damassi, line 180: Still feel a bit 🤔 about combining touchable behavior inside of the context menu code, since pressing on an artwork should be one thing, and the context menu is another, eg:
```tsx
<ContextMenuArtwork>
  <TouchableHighlight onPress={navigateToArtwork}>
    ...
  </TouchableHighlight>
</ContextMenuArtwork>
```

The context menu ideally wouldn't need to worry about anything related to presses -- unless its a long press -- except when said presses _have_ to be overwritten (via something like `event.stopPropagation()`) in order for the context menu to work. Since that isn't the case, combining the two leads to some indirection.

gkartalis (approved):
Thanks! We def need to check if we are proceeding with this as is or if we want to try out a different approach as the library you mentioned (https://github.com/enesozturk/react-native-hold-menu)

src/app/Components/ContextMenu/ContextMenuArtwork.tsx
gkartalis: I also tried different approaches but they weren't successful 🤷‍♂️

gkartalis:
@damassi lets merge this to the `gkartalis/MOPLAT-619` ?

PR #8646, opened 2023-05-09
fix: Android back button takes you back 

gkartalis:
This PR resolves [MOPLAT-761] 

### Description

It was reported that the back button on android (also the gesture) is not working properly.

Created a hook that overrides the behavior and takes you back to the previous screen.

Added it in the following places:
-  the **Artwork** modals (that were previously taking the user on mycollection screen when pressing back)
    - modals medium 
    - attributionClass 
    - authenticity 
- BottomTabNavigator (to reenable it across the app

### Demo Videos

https://github.com/artsy/eigen/assets/21178754/d1631959-9c8e-458f-9633-29ad412ec3b1

https://github.com/artsy/eigen/assets/21178754/300d18f5-4573-4107-9359-5e12448bc62e

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

- Android back button takes you back on Artwork Artist Sale screens - gkartalis

#### Dev changes

-

</details>

[MOPLAT-761]: https://artsyproduct.atlassian.net/browse/MOPLAT-761?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/BottomTabs/BottomTabsNavigator.tsx
+++ b/src/app/Scenes/BottomTabs/BottomTabsNavigator.tsx
@@ -1,6 +1,7 @@
 import { createBottomTabNavigator } from "@react-navigation/bottom-tabs"
 import { AppModule } from "app/AppRegistry"
 import { NavStack } from "app/system/navigation/NavStack"
+import { useAndroidGoBack } from "app/utils/hooks/useBackHandler"
 import { BottomTabType } from "./BottomTabType"
 import { BottomTabs } from "./BottomTabs"
 
@@ -11,6 +12,7 @@ const TabContent = ({
 }: {
   route: { params: { tabName: BottomTabType; rootModuleName: AppModule } }
 }) => {
+  useAndroidGoBack()
   // TODO: simplify this, we probably can get rid of NavStack
   return <NavStack id={route.params.tabName} rootModuleName={route.params.rootModuleName} />
 }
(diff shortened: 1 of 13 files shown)


damassi (commented):
Here's a question -- in general, are we _always_ going to want the android go back functionality? If so, is there any way we can hoist this up to the router level as a default prop (true) as configured inside of the `AppRegistry` file? https://github.com/artsy/eigen/blob/main/src/app/AppRegistry.tsx#L519. 

Another thought would be to place it in `Main`: https://github.com/artsy/eigen/blob/main/src/app/App.tsx#L57

I worry that if we have to add this manually it will lead to chasing bugs as devs overlook, but if its part of the router / framework / app-shell level, then we should always be good.

gkartalis:
Good point, not sure how it will behave in screens like Home 🤔 where it can't take you back but I can try it out

src/app/Scenes/BottomTabs/BottomTabsNavigator.tsx
```diff
@@ -11,6 +12,7 @@ const TabContent = ({
 }: {
   route: { params: { tabName: BottomTabType; rootModuleName: AppModule } }
 }) => {
+  useAndroidGoBack()
```
damassi, line 15: Adding this here wasn't sufficient for covering the whole app? Those other spots were still needed?

What's the pattern / how do we know when it should be used vs not?

PR #8651, opened 2023-05-09
fix: deeplink analytics issues

brainbicycle:
This PR resolves [MOPLAT-684] 

### Description

- Removes the screen tracking provider from the home view and instead manually tracks screen event on focus
- ~Adds visibility checks based on focus state for other home analytics~
- Removes deep link tracking from native side since we are tracking in typescript

#### Follow-ups:

 - There are still some analytics on Home Screen that will fire erroneously when coming from a deeplink when the app is killed, they are: rail views, myCollectionOnboarding screen view, experiment views, probably they can be fixed in a similar way by checking the screen is focused but my first attempt was buggy so keeping out of the PR for now

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- Fix deeplink and home analytics issues - Brian 

</details>

[MOPLAT-684]: https://artsyproduct.atlassian.net/browse/MOPLAT-684?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/Scenes/Home/Home.tsx
```diff
@@ -710,6 +715,7 @@ export const HomeQueryRenderer: React.FC<HomeQRProps> = ({ environment }) => {
       maybeReportExperimentVariant({
         experimentName: RECOMMENDATION_MODEL_EXPERIMENT_NAME,
         enabled: worksForYouRecommendationsModel.enabled,
+        visible: isFocused,
```
MounirDhahri, line 718: **issue:** I believe we should also add `isFocused` to the dependencies array here. 
Only problem is that it might get triggered everytime the screen is active again, but I think `maybeReportExperimentVariant` doesn't report the same events again so we should be fine.

gkartalis (commented):
https://github.com/artsy/eigen/pull/8646/files#diff-f0b7be7d3f96d4dea763f784715b2e780549606efdf041184568e74c4e559d04L118-L126

This will fix the failing tests

ios/Artsy/App/ARAppDelegate.mm
```diff
@@ -53,7 +52,10 @@
 #import "AREmission.h"
 #import "ARNotificationsManager.h"
 #import <React/RCTLinkingManager.h>
+<<<<<<< HEAD:ios/Artsy/App/ARAppDelegate.m
+=======
 #import <React/RCTAppSetupUtils.h>
+>>>>>>> origin/main:ios/Artsy/App/ARAppDelegate.mm
```
brainbicycle, line 58: whoops

ios/Artsy/App/ARAppDelegate.mm
```diff
@@ -268,10 +270,22 @@ - (BOOL)application:(UIApplication *)app openURL:(NSURL *)url options:(NSDiction
...
+        notification = [NSNotification notificationWithName:kAFApplicationLaunchedWithURLNotification
+                                                     object:nil
+                                                   userInfo:@{kAFApplicationLaunchOptionsURLKey : url}];
+
+        [[NSNotificationCenter defaultCenter] postNotification:notification];
+        return YES;
+    }
+=======
     [self trackDeeplinkWithTarget:url referrer:_referralURLRepresentation];
+>>>>>>> origin/main:ios/Artsy/App/ARAppDelegate.mm
```
brainbicycle, line 288: nailed the merge

damassi (approved):

PR #8654, opened 2023-05-10
feat(CX-3645): create artists grid view

MounirDhahri:
This PR resolves [CX-3645] 

### Description

This PR creates the artists grid view inside my collection. This is the grid that gets visible when the users select the **Artists** pill.

https://github.com/artsy/eigen/assets/11945712/014691c3-b4ec-47f8-8221-aa74634f6129

After pulling the latest changes from main 

https://github.com/artsy/eigen/assets/11945712/e7571e0a-2cb7-43ea-b19f-2541a072746f

### PR Checklist

### To the reviewers 👀

- create collected artists grid view - mounir

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-3645]: https://artsyproduct.atlassian.net/browse/CX-3645?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/Scenes/MyCollection/Components/MyCollectionCollectedArtistItem.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { MyCollectionCollectedArtistItem_artist$key } from "__generated__/MyCollectionCollectedArtistItem_artist.graphql"
+import { ArtistListItemContainer } from "app/Components/ArtistListItem"
+import { MyCollectionTabsStore } from "app/Scenes/MyCollection/State/MyCollectionTabsStore"
+import { useMemo } from "react"
+import { graphql, useFragment } from "react-relay"
+
+interface ArtistItem {
+  artist: MyCollectionCollectedArtistItem_artist$key
+  // TODO: Implement compact version of artists grid
+  compact?: boolean
```
MounirDhahri, line 11: this is being tracked in a separate ticket. I decided not to do it here because the PR is getting really big https://artsyproduct.atlassian.net/browse/CX-3702

src/app/Scenes/MyCollection/Components/MyCollectionBottomSheetModals/MyCollectionBottomSheetModals.tsx
```diff
@@ -1,29 +1,31 @@
...
+  const setViewKind = MyCollectionTabsStore.useStoreActions((actions) => actions.setViewKind)
+  const view = MyCollectionTabsStore.useStoreState((state) => state.viewKind)
+  const id = MyCollectionTabsStore.useStoreState((state) => state.id)
 
   const snapPoints = useMemo(() => ["50%"], [])
 
   const handleSheetChanges = useCallback((index: number) => {
     if (index === -1) {
-      setView(null)
+      setViewKind({ viewKind: null })
```
MounirDhahri, line 21: I might refactor this to be reset view in a future step if we end up adding any more logic to it

src/app/Scenes/MyCollection/Components/MyCollectionBottomSheetModals/MyCollectionBottomSheetModalArtist.tsx
```diff
@@ -0,0 +1,16 @@
+import { Flex, Text } from "@artsy/palette-mobile"
+import { BottomSheetView } from "@gorhom/bottom-sheet"
+
+export const MyCollectionBottomSheetModalArtist: React.FC<{ artistId: string }> = ({
```
MounirDhahri, line 4: preparing the modal here for future work by Daria

src/app/Components/ArtistListItem.tsx
```diff
@@ -45,115 +46,106 @@ export const formatTombstoneText = (
...
 
-@track()
-export class ArtistListItem extends React.Component<Props, State> {
-  static defaultProps = {
-    withFeedback: false,
-    containerStyle: {},
-  }
-
-  state = { isFollowedChanging: false }
+const ArtistListItem: React.FC<Props> = ({
```
MounirDhahri, line 49: I mainly refactored this component to be a functional component and made it possible to have a custom right button instead of always having the follow button

src/app/Scenes/MyCollection/Components/MyCollectionCollectedArtistsView.tsx
```diff
@@ -0,0 +1,115 @@
...
+        onComplete: () => {
+          setRefreshing(false)
+        },
+      }
+    )
+  }
+
+  const collectedArtists = extractNodes(data.myCollectionInfo?.collectedArtistsConnection)
+
+  if (!collectedArtists.length) {
```
MounirDhahri, line 61: I don't have any fancy logic here for when there are no artists because if there were no artists, my collection empty page will be visible and will take care of this. this is mainly here to avoid weird ts casting

src/app/Scenes/MyCollection/Components/MyCollectionSearchBar.tsx
```diff
@@ -155,49 +155,61 @@ export const MyCollectionSearchBar: React.FC<MyCollectionSearchBarProps> = ({
...
-            </Flex>
+            <ViewAsIcons onViewOptionChange={onViewOptionChange} viewOption={viewOption} />
           </Flex>
         </Flex>
       )}
     </Flex>
   )
 }
+
+export const ViewAsIcons: React.FC<{
```
MounirDhahri, line 166: Extracting this since it's used for both artworks and artists - will move to a separate file as I start the work on the grid.

src/app/Components/MenuItem.tsx
```diff
@@ -86,7 +86,7 @@ export const MenuItem: React.FC<{
 
         <Flex flexDirection="row" justifyContent="flex-end" flex={1} height="100%">
           {!!value && (
-            <Flex flex={1}>
+            <Flex overflow="hidden" minWidth={200}>
```
olerichter00, line 89: **question:** Why do we need `overflow="hidden"`?

MounirDhahri: no, sorry, i was about to remove that commit after i finish the QA call in 12 minutes. I fixed it separately in a cleaner way here https://github.com/artsy/eigen/pull/8675

src/app/Components/ArtistListItem.tsx
```diff
@@ -45,115 +46,106 @@ export const formatTombstoneText = (
...
-                <EntityHeader
-                  mr={1}
-                  name={name}
-                  meta={formatTombstoneText(nationality, birthday, deathday) ?? undefined}
-                  imageUrl={imageURl ?? undefined}
-                  initials={initials ?? undefined}
-                />
-              </Flex>
+    return undefined
+  }, [nationality, birthday, deathday, uploadsCount])
```
olerichter00, line 108: I am not sure about this pattern. If this leads to a real (and necessary) performance improvement, I think it's fine. Otherwise, it could lead to potential bugs. For example, forgetting to add something to the array (this happened to me yesterday on the home screen).

src/app/Components/ArtistListItem.tsx
```diff
@@ -233,3 +225,24 @@ export const ArtistListItemPlaceholder = () => (
     </Flex>
   </Flex>
 )
+
+const tracks = {
```
olerichter00, line 229: 👏

src/app/Scenes/MyCollection/Components/MyCollectionCollectedArtistsRail.tsx
```diff
@@ -98,90 +85,49 @@ export const MyCollectionCollectedArtistsRail: React.FC<MyCollectionCollectedArt
...
-  image,
+export const Artist: React.FC<{ artist: MyCollectionCollectedArtistsRail_artist$key }> = ({
+  artist,
 }) => {
+  const data = useFragment(artistFragment, artist)
+
   return (
     <Flex mr={1} width={ARTIST_CIRCLE_DIAMETER}>
-      <Avatar initials={initials} src={image} size="sm" />
+      <Avatar initials={data.initials!} src={data?.image?.url || undefined} size="sm" />
```
lordkiz, line 95: I think you mean `Avatar initials={data.initials} ` or `initials={data.initials || undefined}`

src/app/Scenes/MyCollection/Components/MyCollectionCollectedArtistsView.tsx
olerichter00: True, this case should never happen.

src/app/Scenes/MyCollection/Components/MyCollectionStickyHeader.tsx
```diff
@@ -58,6 +58,14 @@ export const MyCollectionStickyHeader: React.FC<MyCollectionStickyHeaderProps> =
...
   const showSubmissionMessage = showVisualClue("ArtworkSubmissionMessage")
   const enableCollectedArtists = useFeatureFlag("AREnableMyCollectionCollectedArtists")
+  const selectedTab = MyCollectionTabsStore.useStoreState((state) => state.selectedTab)
+
+  const showArtworkFilters = useMemo(() => {
+    if (!enableCollectedArtists) {
+      return !!hasArtworks
+    }
+    return selectedTab === "Artworks"
+  }, [selectedTab, hasArtworks])
```
lordkiz, line 68: `enableCollectedArtists` might be needed in the dependency array, I think 🤷

src/app/Scenes/MyCollection/Components/MyCollectionSearchBar.tsx
```diff
@@ -155,49 +155,61 @@ export const MyCollectionSearchBar: React.FC<MyCollectionSearchBarProps> = ({
...
+  onViewOptionChange: (viewOption: ViewOption) => void
+  viewOption: ViewOption
+}> = ({ onViewOptionChange, viewOption }) => {
+  const space = useSpace()
+  return (
```
olerichter00, line 171:
```suggestion
  const space = useSpace()

  return (
```

olerichter00 (dismissed):
Great work, @MounirDhahri! 🌟 

I found two small issues:

1. Looking at the screenshot in the ticket, I don't see this border between the filters and the artist list.

Screenshot 2023-05-12 at 17 32 20

2. If the artist doesn't has meta information, the design looks weird:

Screenshot 2023-05-12 at 17 32 44

src/app/Scenes/MyCollection/State/MyCollectionTabsStore.tsx
```diff
@@ -1,23 +1,48 @@
...
+        break
+
+      case "Add":
+        state.viewKind = "Add"
+        break
+
+      default:
+        state.viewKind = payload.viewKind
+        state.id = payload.id
+        break
```
lordkiz, line 44: We may not need this switch statement if we type payload.viewKind. With this, one can set viewKind to anything.

src/app/utils/pluralize.ts
```diff
@@ -0,0 +1,18 @@
...
+  if (count === 1) {
+    return word
+  }
+
+  if (customPlural) {
+    return customPlural
+  }
+
+  return `${word}s`
+}
```
lordkiz, line 18: Maybe one day we can create a better pluralisation logic by using this as a guide: https://github.com/rails/rails/blob/main/activesupport/lib/active_support/inflections.rb

lordkiz (commented):
Nice PR. I don't see any major blockers. 👏

src/app/Scenes/MyCollection/Components/MyCollectionCollectedArtistsRail.tsx
MounirDhahri: sure - that also works. my assumption is that the initials are always there so I casted it but not the avatar. Will update

lordkiz (approved):

lordkiz (dismissed):

src/app/Components/ArtistListItem.tsx
MounirDhahri: Usually there is a warning telling you something is messy but I can see that it's disabled in Eigen for some reason. I will update this to use a method instead

src/app/utils/pluralize.ts
MounirDhahri: wow! powerful

src/app/Components/ArtistListItem.tsx
olerichter00: That would solve the problem with the possible bugs. 

I am still not convinced about using `useMemo` etc. too much. I once read a great article about how it actually can degrade performance in some cases, but I can`t find it anymore.

olerichter00: I think I found it: https://blog.logrocket.com/when-not-to-use-usememo-react-hook/

MounirDhahri: thanks for sharing this article. I understand the rationale behind it. my initial thought on optimising here is because the artist items are rendered within lists and if the list items are using a prop that might change. using a `useMemo` (when sure it won't break something), can theoretically lead to performance improvements. For now I switched to using a method anyway given it's not performing bad and can switch back in case it breaks

src/app/utils/pluralize.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * Method that returns the plural of a word
+ * @param word
+ * @param count
+ * @param customPlural
+ * @returns string
+ */
+export const pluralize = (word: string, count: number, customPlural?: string) => {
```
dariakoko, line 8: 👏 hooraaay

src/app/Scenes/MyCollection/Components/MyCollectionCollectedArtistsRail.tsx
```diff
@@ -29,29 +28,24 @@ const AddMoreButton = () => {
...
-    setIsLoading(true)
-
-    relay.loadMore(20, (err) => {
-      setIsLoading(false)
-
-      if (err) {
-        console.error(err)
-      }
-    })
+    loadNext(10)
```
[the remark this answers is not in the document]
MounirDhahri: For no particularly big reason honestly. Only thing I can think of is that 20 is a big number to load first imo and most users have less than 10 artists.

PR #8668, opened 2023-05-11
fix: navigate to quiz page only when navigation is ready

MounirDhahri:
This PR resolves an issue making the app break after starting the quiz (tried after signing up but it might be also happening when starting the quiz from a different flow)

**Before** 

https://github.com/artsy/eigen/assets/11945712/1a8fa74d-5a6e-426f-a67a-4abe264a79e6

**After**

https://github.com/artsy/eigen/assets/11945712/896e40b7-4afd-454b-b5ac-7c1abc947588

### Description

Once the user signs up, we switch the navigation stack they're on to be within the main navigation container. the latter takes a while to be ready regardless of the fact that the first screen (home) is already mounted. The existing logic for the quiz is built in such a way that the user gets directly navigated to the art quiz once they're in home with incomplete quiz. This means there is a chance that they get navigated to the art quiz before the navigation container is ready. 

**Suggested next steps:**
- the art quiz should be the first route you get to if you want to go to the art quiz from the onboarding. (Not home then navigate to art quiz)
- Somehow we forget to think about the navigation `onReady` prop. I believe that will fix tons of deep linking issues
- Some methods that `navigate` is using, use `requestAnimationFrame` to wait for the navigation stack to be ready, we might be able to remove that delay in favour of not allowing navigation until it's ready or waiting for the navigation container to be ready

### PR Checklist

### To the reviewers 👀

- fix navigate to quiz page only when navigation is ready - mounir

#### iOS user-facing changes

-

-

#### Dev changes

- add isNavigationReady to global store - mounir

</details>


src/app/system/navigation/ModalStack.tsx
```diff
@@ -22,10 +23,14 @@ const Stack = createStackNavigator()
  */
 export const ModalStack: React.FC = ({ children }) => {
   const initialState = useReloadedDevNavigationState("main_modal_stack", __unsafe_mainModalStackRef)
+  const { setSessionState: setNavigationReady } = GlobalStore.actions
   return (
     <NavigationContainer
       ref={__unsafe_mainModalStackRef}
       initialState={initialState}
+      onReady={() => {
+        setNavigationReady({ isNavigationReady: true })
```
damassi, line 32: Thoughts on why this hasn't manifested before?

damassi:
@MounirDhahri - this is a nonblocking suggestion, but now that we have two bits of required "boot" state in order for the app to proceed, can we wrap those up into a hook and perhaps check it in `Main.tsx`? 

Similar to how we do it in Energy:
- https://github.com/artsy/energy/blob/main/src/system/hooks/useSystemIsDoneBooting.ts
- https://github.com/artsy/energy/blob/main/src/Navigation.tsx#L60-L62

src/app/system/navigation/ModalStack.tsx
MounirDhahri: tbh, I have no solid answer to your question. could be because of a package upgrade that this showed up

damassi: All good, yah, was just curious. /such-is-life

brainbicycle (approved):
🙏 thanks for catching and fixing!

rajsam003:
I didnt get this issue while working on it. Thanks for fixing it @MounirDhahri

MounirDhahri:
I will have a look later at why this broke- I think I have an idea now. will share my findings here

PR #8678, opened 2023-05-12
feat: add artwork lists context for artwork list screen

dimatretyak:
This PR resolves [] 

### Demo
| Before | After |
| <video src="https://github.com/artsy/eigen/assets/3513494/f5d9b7d8-d960-4bba-b3d9-7cff32b46403" />  | <video src="https://github.com/artsy/eigen/assets/3513494/22ed13f7-02e1-4201-8e86-ce134df8dda2" />  | 

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- add artwork lists context for artwork list screen - dimatretyak

</details>

### DIFF
--- a/src/app/Components/ArtworkLists/views/SelectArtworkListsForArtworkView/useUpdateArtworkListsForArtwork.ts
+++ b/src/app/Components/ArtworkLists/views/SelectArtworkListsForArtworkView/useUpdateArtworkListsForArtwork.ts
@@ -2,6 +2,7 @@ import {
   useUpdateArtworkListsForArtworkMutation,
   useUpdateArtworkListsForArtworkMutation$data,
 } from "__generated__/useUpdateArtworkListsForArtworkMutation.graphql"
+import { refreshOnArtworkSave } from "app/utils/refreshHelpers"
 import { UseMutationConfig, graphql, useMutation } from "react-relay"
 import { Disposable, RecordSourceSelectorProxy } from "relay-runtime"
 
@@ -31,6 +32,10 @@ export const useUpdateArtworkListsForArtwork = (artworkID: string): MutationResu
     return initialCommit({
       ...config,
       updater: updater(artworkID),
+      onCompleted: (...args) => {
+        refreshOnArtworkSave()
+        config.onCompleted?.(...args)
+      },
     })
   }
 
@@ -107,24 +112,15 @@ export const getArtworkListsCountByType = (artworkLists: ArtworkListEntity) => {
 }
 
 const SaveArtworkListsMutation = graphql`
-  mutation useUpdateArtworkListsForArtworkMutation(
-    $artworkID: String!
-    $input: ArtworksCollectionsBatchUpdateInput!
-  ) {
+  mutation useUpdateArtworkListsForArtworkMutation($input: ArtworksCollectionsBatchUpdateInput!) {
     artworksCollectionsBatchUpdate(input: $input) {
       responseOrError {
         ... on ArtworksCollectionsBatchUpdateSuccess {
           addedToArtworkLists: addedToCollections {
-            internalID
             default
-            ...ArtworkListItem_item @arguments(artworkID: $artworkID)
-            ...ArtworkListItem_collection
           }
           removedFromArtworkLists: removedFromCollections {
-            internalID
             default
-            ...ArtworkListItem_item @arguments(artworkID: $artworkID)
-            ...ArtworkListItem_collection
           }
         }
         ... on ArtworksCollectionsBatchUpdateFailure {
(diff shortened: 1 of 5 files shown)


src/app/Components/ArtworkLists/views/SelectArtworkListsForArtworkView/useUpdateArtworkListsForArtwork.ts
```diff
@@ -107,22 +112,15 @@ export const getArtworkListsCountByType = (artworkLists: ArtworkListEntity) => {
...
-    $input: ArtworksCollectionsBatchUpdateInput!
-  ) {
+  mutation useUpdateArtworkListsForArtworkMutation($input: ArtworksCollectionsBatchUpdateInput!) {
     artworksCollectionsBatchUpdate(input: $input) {
       responseOrError {
         ... on ArtworksCollectionsBatchUpdateSuccess {
           addedToArtworkLists: addedToCollections {
-            internalID
             default
-            ...ArtworkListItem_item @arguments(artworkID: $artworkID)
```
dimatretyak, line 120: There is no longer a need to query this to update the data in the relay store, as [refreshOnArtworkSave](https://github.com/artsy/eigen/blob/f689e1457bcd8a5afc5609b5cacd71856dc6ec13/src/app/Components/ArtworkLists/views/SelectArtworkListsForArtworkView/useUpdateArtworkListsForArtwork.ts#L36) will be called, which will cause all `Saves` pages to be [refetched](https://github.com/artsy/eigen/blob/f689e1457bcd8a5afc5609b5cacd71856dc6ec13/src/app/Scenes/ArtworkLists/ArtworkLists.tsx#L59-L65).

dimatretyak: > correct logic for `optimisticUpdater`/`updater`

WIP: https://github.com/artsy/eigen/compare/dimatretyak/sadnbox/artwork-lists-improvements

src/app/Scenes/ArtworkLists/ArtworkLists.tsx
```diff
@@ -55,6 +56,14 @@ export const ArtworkLists = () => {
     )
   }
 
+  useEffect(() => {
+    RefreshEvents.addListener(FAVORITE_ARTWORKS_REFRESH_KEY, handleRefresh)
+
+    return () => {
+      RefreshEvents.removeListener(FAVORITE_ARTWORKS_REFRESH_KEY, handleRefresh)
+    }
+  }, [])
```
dimatretyak, line 65: The disadvantage of this approach is that pagination will be lost


https://github.com/artsy/eigen/assets/3513494/bd8e0b92-b892-4594-bc7b-e0b5fb5ff3ee

src/app/Components/ArtworkLists/views/SelectArtworkListsForArtworkView/useUpdateArtworkListsForArtwork.ts
dimatretyak: ### Update relay store/cache via `updater`
[Another approach](https://github.com/artsy/eigen/compare/dimatretyak/sadnbox/artwork-lists-improvements) using `optimisticUpdater` and `updater`, in which we don't have to do [refetch](https://github.com/artsy/eigen/blob/f689e1457bcd8a5afc5609b5cacd71856dc6ec13/src/app/Scenes/ArtworkLists/ArtworkLists.tsx#L59-L65) for the `Saves` page (see video).

IMHO such code will be more difficult to maintain and any changes to the query (for example, passing new arguments) will break the code in updater. What do you think? 🤔

### Demo
https://github.com/artsy/eigen/assets/3513494/a0a8c267-ac34-42e9-bf4a-aabeffca9e5c

src/app/Scenes/ArtworkLists/ArtworkLists.tsx
anandaroop: Good callout. Can you remind me of the sort order here?  

If it's in descending order of modification date I'd say this is fine. If it's creation date, then it's not ideal, but maybe still acceptable.

src/app/Components/ArtworkLists/views/SelectArtworkListsForArtworkView/useUpdateArtworkListsForArtwork.ts
dimatretyak: to tell the truth, I prefer the 2 approach (we have the same behaviour in Force)

src/app/Scenes/ArtworkLists/ArtworkLists.tsx
anandaroop: I think it might be time to [reconsider that](https://artsy.slack.com/archives/C9SATFLUU/p1684353474289759), in which case I feel like this problem would be minimized. (And I don't think it's blocker for this PR)

anandaroop (approved):

dimatretyak:
After discussion, it was decided to use [approach 2 ](https://github.com/artsy/eigen/pull/8713) (with modification data in relay store/cache). If we need, then we will return to the current approach 1 (using `refetch')  in the future

PR #8684, opened 2023-05-15
RFC: Add Geolocation service

olerichter00:
This PR addresses [CX-3630](https://artsyproduct.atlassian.net/browse/CX-3630)

### Description

To include shows near the user's location on the home screen's "Shows for You" rail, we need to add a package to detect the location of the device.

I am thinking about using [react-native-geolocation-service](https://www.npmjs.com/package/react-native-geolocation-service) because it's quite small compared to others.

Screenshot 2023-05-15 at 13 22 30

Another popular alternative could be [@react-native-community/geolocation](https://www.npmjs.com/package/@react-native-community/geolocation). 

Screenshot 2023-05-15 at 13 23 28

---

In addition to the location service and because we want to fall back to IP-based geolocation, I've also added the package [react-native-network-info](https://www.npmjs.com/package/react-native-network-info). This allows us to fall back to IP-based geolocation, and send the user's IP address in the GraphQL query.

---

I intend to use the package like this:

```typescript
// Returns the user's location if available, otherwise returns the user's IP address
export const useLocationOrIpAddress = () => {
  const [isLoading, setIsLoading] = useState(true)
  const [location, setLocation] = useState<Location | null>(null)
  const [ip, setIp] = useState<string | null>(null)

  const getIp = async () => {
    try {
      // Get IPv4 IP (priority: WiFi first, cellular second)
      const ipv4Address = "85.214.123.221" || (await NetworkInfo.getIPV4Address())

      // setIp(ipv4Address)
      // TODO: Remove me
      setIp(ipv4Address)
    } catch (error) {
      console.log("Failed to get IPv4 address.", error)
    } finally {
      setIsLoading(false)
    }
  }

  const getLocation = async () => {
    Geolocation.getCurrentPosition(
      (position: any) => {
        setLocation({ lat: position.coords.latitude, lng: position.coords.longitude })
        setIsLoading(false)
      },
      (error: any) => {
        console.log("Failed to get current position.", error)
        // Get IP address as a fallback
        getIp()
      },
      { timeout: 15000, maximumAge: 10000 }
    )
  }

  useEffect(() => {
    getLocation()
  }, [])

  return { isLoading, location, ip }
}
```
[CX-3630]: https://artsyproduct.atlassian.net/browse/CX-3630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gkartalis:
[@react-native-community/geolocation](https://www.npmjs.com/package/@react-native-community/geolocation) is double the size 🤯.

Some thoughts: Other than the bundle size difference I think @react-native-community/geolocation is better for the following reasons:
- its the official geolocation package backed from rn community
- gets more frequent updates (last updated last month vs the other library which was updated 8 months ago)
- TS support vs [react-native-geolocation-service](https://www.npmjs.com/package/react-native-geolocation-service) being a JS library


> In addition to the location service and because we want to fall back to IP-based geolocation, I've also added the package [react-native-network-info](https://www.npmjs.com/package/react-native-network-info). This allows us to fall back to IP-based geolocation, and send the user's IP address in the GraphQL query.

Are there any other packages out there that do the same job? My concern here is that this library hasn't been updated for 4 years 😓

olerichter00:
> Some thoughts: Other than the bundle size difference I think @react-native-community/geolocation is better for the following reasons:

These are very good reasons. I agree, [@react-native-community/geolocation](https://www.npmjs.com/package/@react-native-community/geolocation) looks like the better choice 👍  If there are no objections I'll use this one instead. 

> Are there any other packages out there that do the same job? My concern here is that this library hasn't been updated for 4 years 😓

I'll have a look. But it seems to work and it's only getting the devices IP address.

> Don't we have in gravity the users IP address? Could we expose it and use it in the app?

Doesn't seem to be that easy. There are some discussions around this here: https://github.com/artsy/metaphysics/pull/4992#discussion_r1185962912

MounirDhahri:
To your point @gkartalis about getting the IP from the backend, I saw something like that [here](https://github.com/teamairship/react-native-public-ip/blob/master/index.js)

olerichter00:
I'll merge to move on with https://github.com/artsy/eigen/pull/8665

PR #8689, opened 2023-05-15
feat: support lazy loading inside artist and my collection sticky tab pages [WIP]

MounirDhahri:
This PR resolves [CX-3706] 

### Description

Over the past few weeks we have noticed that diffusion is breaking quite often due to increased traffic, a lot of this traffic can be avoided if lazy load tabs that are using diffusion but are barely opened. This PR will give us some time to improve diffusion to handle more traffic and we can potentially no longer lazy load any of this tabs.

This PR does not change any of the exiting logic, I mainly refactored the tabs to be separate query renderers and fixed tests that broke. 

It's tricky to get the skeleton right on some of the tabs because we don't know if the user has data or not and what to draw there. I am open to suggestions there

https://github.com/artsy/eigen/assets/11945712/aa108a34-0fe6-4b8c-9575-5a1a5e106f3d

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- lazy load my collection and artist screens - mounir

</details>

[CX-3706]: https://artsyproduct.atlassian.net/browse/CX-3706?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/Scenes/Inbox/Screens/Conversation.tsx
```diff
@@ -1,5 +1,5 @@
-import { InfoCircleIcon, Flex, Text } from "@artsy/palette-mobile"
-import NetInfo from "@react-native-community/netinfo"
+import { InfoCircleIcon, Flex, Text, Touchable } from "@artsy/palette-mobile"
```
MounirDhahri, line 1: i will revert this, it's here by mistake

egdbear:
Thanks for taking this @MounirDhahri 🚀

src/app/Scenes/MyProfile/MyProfileHeaderMyCollectionAndSavedWorks.tsx
```diff
@@ -43,6 +32,7 @@ export const MyProfileHeaderMyCollectionAndSavedWorks: React.FC<{
       <SafeAreaView edges={["top"]} style={{ flex: 1 }}>
         <StickyTabPage
           disableBackButtonUpdate
+          lazyLoadTabs
```
olerichter00, line 35: **question:** Following up on our conversation in Knowledge Share: Could we remove this flag? The reason I'm asking is, that setting the flag means that the user sees a loading screen after switching to another tab for the first time, which we should try to avoid in my opinion.

Have you tested if the problem this PR solves would be solved without setting this flag here? Using a query renderer per tab instead of fragments (that get all resolved with the same query) would probably already fix the issue of slow queries. We're doing this (without lazy loading) in [AboveTheFoldQueryRenderer](https://github.com/artsy/eigen/blob/main/src/app/utils/AboveTheFoldQueryRenderer.tsx#L13).

MounirDhahri: I think your point here is valid, my rationale here is to not load the tab content unless the tab is visited to avoid the extra traffic on Diffusion. Your point here is about UX and I definitely agree that it's a lot better as UX if the tab content is already visible. Separating the QueryRenderers and removing this flag, will help load the data separately from the my collection tab and maybe have it ready when the user lands there. I think that's eventually where we want to get if diffusion was able to handle the extra traffic.
How about we load the content of the Saves and Follows tab and only lazy load the slow insights tab query?

olerichter00: Alright, if firing the queries at the same time is resulting in timeouts, lazy loading the other queries sounds like a good idea.

An alternative could be to load the data for the opened tab immediately and the data for the other tabs after a timeout ([example](https://github.com/artsy/eigen/blob/main/src/app/utils/AboveTheFoldQueryRenderer.tsx#L101)).

MounirDhahri: nice approach. Hopefully we can do as you suggested later when diffusion can handle more traffic

src/app/utils/hooks/withSuspense.tsx
```diff
@@ -0,0 +1,19 @@
+import { Spinner } from "@artsy/palette-mobile"
+import { ProvidePlaceholderContext } from "app/utils/placeholders"
+import { Suspense } from "react"
+
+export const withSuspense =
```
MounirDhahri, line 5: Rational about this in this [thread](https://artsy.slack.com/archives/C02BAQ5K7/p1655222755848999?thread_ts=1655219870.652769&cid=C02BAQ5K7)

src/app/Components/Artist/ArtistAbout/ArtistAbout.tsx
```diff
@@ -1,23 +1,28 @@
...
 import { ArtistAboutShowsFragmentContainer } from "./ArtistAboutShows"
 
 interface Props {
-  artist: ArtistAbout_artist$data
+  artist: ArtistAbout_artist$key
 }
 
-export const ArtistAbout: React.FC<Props> = ({ artist }) => {
+export const ArtistAbout: React.FC<Props> = ({ artist: artistProp }) => {
+  const artist = useFragment(artistAboutFragment, artistProp)
```
dimatretyak, line 24: **praise**: thank you for refactoring and using relay hooks

src/app/Components/Artist/ArtistInsights/ArtistInsights.tests.tsx
```diff
@@ -57,7 +53,7 @@ describe("ArtistInsights", () => {
     resolveMostRecentRelayOperation(mockEnvironment)
 
     setImmediate(() => {
-      expect(trackEvent).toHaveBeenCalledWith({
+      expect(mockTrackEvent).toHaveBeenCalledWith({
```
dimatretyak, line 56: nice to see it 👍

src/app/Components/Artist/ArtistInsights/MarketStats.tsx
```diff
@@ -189,11 +189,10 @@ export const MarketStatsFragmentContainer = createFragmentContainer(MarketStats,
 
 export const MarketStatsQueryRenderer: React.FC<{
   artistInternalID: string
-  environment: RelayModernEnvironment
-}> = ({ artistInternalID, environment }) => {
+}> = ({ artistInternalID }) => {
   return (
     <QueryRenderer<MarketStatsQuery>
-      environment={environment}
+      environment={getRelayEnvironment()}
```
dimatretyak, line 195: Thank you. with the old approach, there could be problems that are difficult to debug

src/app/Components/StickyTabPage/StickyTabPagePlaceholder.tsx
```diff
@@ -0,0 +1,16 @@
+import { Flex, Spinner } from "@artsy/palette-mobile"
+import { StickyTabPageScrollView } from "app/Components/StickyTabPage/StickyTabPageScrollView"
+
+export const StickTabPagePlaceholder: React.FC = () => {
```
dimatretyak, line 4: **nitpick**: small typo in name. Please rename from `StickTabPagePlaceholder` to `StickyTabPagePlaceholder`

dimatretyak (approved):

src/app/Components/StickyTabPage/StickyTabPage.tsx
```diff
@@ -75,6 +77,12 @@ export const StickyTabPage: React.FC<StickyTabPageProps> = ({
     Array<JSX.Element | null>
   >([])
 
+  const enableLazyLoading = useFeatureFlag("AREnableStickyTabsLazyLoading")
+
+  const [viewedTabLabels, setViewedTabLabels] = useState(
+    enableLazyLoading && lazyLoadTabs ? [tabs[initialTabIndex].title] : tabs.map((tab) => tab.title)
```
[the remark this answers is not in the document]
MounirDhahri: changed as well. thanks 🙌

MounirDhahri:
Just address your review comments @dimatretyak and fixed broken tests. I moved the filter test to `ArtistArtworks` since I believe it belongs there

src/app/Components/StickyTabPage/StickyTabPage.tsx
```diff
@@ -128,8 +149,14 @@ export const StickyTabPage: React.FC<StickyTabPageProps> = ({
         stickyHeaderHeight,
         headerOffsetY,
         tabLabels: tabs.map((tab) => tab.title),
+        viewedTabLabels,
         tabVisualClues: tabs.map((tab) => tab.visualClues),
         setActiveTabIndex(index) {
+          const tabTitle = tabs[index].title
+          if (!viewedTabLabels.includes(tabTitle)) {
```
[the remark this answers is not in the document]
dimatretyak: thanks!

olerichter00 (approved):

egdbear:
I'm not sure if this issue can be addressed in this particular pull request or if it should be handled separately, but it seems that `auctionResultsByFollowedArtists` is causing a heavy load on Diffusion's database.

The query used in this component triggers a count operation for most of the provided artist IDs. While I'm not deeply familiar with the design of the component that utilizes this query, ideally, we should find a way to avoid using the totalCount altogether.


Convos [here](https://artsy.slack.com/archives/C0HP61PUJ/p1684116169579189) and [here](https://artsy.slack.com/archives/C03D5MXDAN7/p1684240515683009).

MounirDhahri:
Thanks for raising this @egdbear - we are aware of its performance issues. We will work on addressing it in a separate PR though the next sprint.

src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
```diff
@@ -309,3 +317,35 @@ export default createPaginationContainer(
     `,
   }
 )
+
+const artistArtworksQuery = graphql`
+  query ArtistArtworksQuery($artistID: String!, $input: FilterArtworksInput) {
+    artist(id: $artistID) {
+      ...ArtistArtworks_artist @arguments(input: $input)
```
dimatretyak, line 324: **question**: do you get this error when navigated to artist screen?

https://github.com/artsy/eigen/assets/3513494/931a292a-6668-4ee8-98ec-f8e8961dbfd9

dimatretyak: thank you!

brainbicycle:
this is great work! I just want to raise that MoPlat is currently working on replacing the stickyTabs with the simpler implementation from Energy, I don't think this should be a problem but if CX is planning more work on the tabs soon we should probably coordinate! cc @damassi @gkartalis

olerichter00:
> The query used in this component triggers a count operation for most of the provided artist IDs. While I'm not deeply familiar with the design of the component that utilizes this query, ideally, we should find a way to avoid using the totalCount altogether.

@egdbear @MounirDhahri found a solution to avoid `totalCount`: https://github.com/artsy/eigen/pull/8721

MounirDhahri:
Closing this PR since I cherry picked most of the changes into different other PRs. this was a nice one 💟  thanks all for reviewing

PR #8697, opened 2023-05-16
fix: hot reload when app is run using yarn start

MounirDhahri:
This PR fixes hot reloading issues on Eigen

### Description

https://github.com/artsy/eigen/assets/11945712/29e9491f-928b-4fc1-9b1f-9142f8d436f4

### PR Checklist

- [x] I have tested my changes on **iOS** and **Android**.
- [x] I hid my changes behind a **[feature flag]**, or they don't need one.
- [x] I have included **screenshots** or **videos**, or I have not changed the UI.
- [x] I have added **tests**, or my changes don't require any.
- [x] I added an **[app state migration]**, or my changes do not require one.
- [x] I have documented any **follow-up work** that this PR will require, or it does not require any.
- [x] I have added a **changelog entry** below, or my changes do not require one.

### To the reviewers 👀

- [ ] I would like **at least one** of the reviewers to **run** this PR on the simulator or device.

<details><summary>Changelog updates</summary>

#### Cross-platform user-facing changes

-

#### iOS user-facing changes

-

#### Android user-facing changes

-

#### Dev changes

-

</details>

Need help with something? Have a look at our [docs], or get in touch with us.

[app state migration]: ../blob/main/docs/adding_state_migrations.md
[feature flag]: ../blob/main/docs/developing_a_feature.md
[docs]: ../blob/main/docs/README.md

PR #8701, opened 2023-05-17
Revert "fix: hot reload when app is run using yarn start"

MounirDhahri:
Reverts artsy/eigen#8697


MounirDhahri:
sorry I reverted this before going on holidays quickly. raised an error in the main PR that was happening to all users and wanted to revert it while investigating the issue. Thanks for fixing it

MounirDhahri:
Done - just added some description 🙏

MounirDhahri:
Something seems to be broken when it comes to the query map generation 

https://github.com/artsy/eigen/assets/11945712/c26d08b3-060a-4557-8dc5-e82554d91c8a

dimatretyak:
@MounirDhahri thank you for checking!

damassi:
Once we re-add query files to git + queryMap -- https://github.com/artsy/eigen/blob/28ddd4dbc9e474bc7b59c8529c8fbda4a5e18963/.gitignore#L86 -- this should be fixed.

PR #8717, opened 2023-05-19
fix: Disable Shows for You location feature flag

olerichter00:
### Description

This PR disables the "Shows for You" rail location feature flag and fixes the logic. The fix is need to prevent the rail from accessing the devices location if the feature flag is disabled.

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Scenes/Home/Components/ShowsRail.tests.tsx
+++ b/src/app/Scenes/Home/Components/ShowsRail.tests.tsx
@@ -48,7 +48,7 @@ describe("ShowsRailContainer", () => {
     })
   })
 
-  describe("with location enabled", () => {
+  describe("with location disabled", () => {
     const environment = createMockEnvironment()
 
     it("renders the title and the shows", async () => {
--- a/src/app/Scenes/Home/Components/ShowsRail.tests.tsx
+++ b/src/app/Scenes/Home/Components/ShowsRail.tests.tsx
@@ -20,26 +20,56 @@ jest.mock("react-native-network-info", () => ({
 }))
 
 describe("ShowsRailContainer", () => {
-  const environment = createMockEnvironment()
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
 
-  it("renders the title and the shows", async () => {
-    const { getByText } = renderWithHookWrappersTL(
-      <ShowsRailContainer title="Shows for You" />,
-      environment
-    )
+  describe("with location enabled", () => {
+    const environment = createMockEnvironment()
 
-    environment.mock.resolveMostRecentOperation((operation) =>
-      MockPayloadGenerator.generate(operation, { Me: () => meResponse })
-    )
+    it("renders the title and the shows", async () => {
+      const { getByText } = renderWithHookWrappersTL(
+        <ShowsRailContainer title="Shows for You" />,
+        environment
+      )
 
-    expect(Geolocation.getCurrentPosition).toHaveBeenCalled()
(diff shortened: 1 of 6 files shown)


src/app/Scenes/Home/Components/ShowsRail.tsx
```diff
@@ -109,24 +109,22 @@ export const tracks = {
 
 interface ShowsRailContainerProps {
   title: string
-  withLocation?: boolean
+  disableLocation?: boolean
```
olerichter00, line 112: Renaming the prop because it's enabled by default.

src/app/Scenes/Home/Components/ShowsRail.tsx
```diff
@@ -109,24 +109,22 @@ export const tracks = {
...
 
 export const ShowsRailContainer: React.FC<ShowsRailContainerProps> = ({
-  withLocation = true,
+  disableLocation = false,
   ...restProps
 }) => {
-  const { location, ipAddress, isLoading } = useLocationOrIpAddress()
-
-  const locationProps = withLocation ? { location, ipAddress } : {}
+  const { location, ipAddress, isLoading } = useLocationOrIpAddress(disableLocation)
```
olerichter00, line 119: Passing `disableLocation` to prevent the hook from getting the location of the device.

src/app/Scenes/Home/Components/ShowsRail.tsx
MrSltun: **Suggestion:** Should we have it disabled by default instead? and enabled when we enable the feature flag?


I would even ask if the user wants to enable it even after we enable the feature flag, maybe by having a small icon or button next to the shows rail title?

MrSltun (approved):
Looks great!! Thank you Ole! 🙌 
I just left a small non-blocking suggestion :)

src/app/Scenes/Home/Components/ShowsRail.tests.tsx
```diff
@@ -20,26 +20,56 @@ jest.mock("react-native-network-info", () => ({
...
+
+      await flushPromiseQueue()
+
+      expect(getByText("Shows for You")).toBeDefined()
+
+      expect(getByText("Leeum Collection: Beyond Space")).toBeDefined()
+    })
+  })
+
+  describe("with location enabled", () => {
```
dimatretyak, line 51: i think here we need to use `with location disabled`

dimatretyak (dismissed):
LGTM 👍

src/app/Scenes/Home/Components/ShowsRail.tsx
olerichter00: > Suggestion: Should we have it disabled by default instead? and enabled when we enable the feature flag?

I thought it would be easier to clean up the feature flag logic if `disableLocation` is enabled by default. If it would be disabled by default, removing the feature flag would result in the rail not adding shows near the user's location.

```
<ShowsRailContainer title={item.title} disableLocation={!enableShowsForYouLocation} />
```

-->

```
<ShowsRailContainer title={item.title} />
// location is disabled
```

> I would even ask if the user wants to enable it even after we enable the feature flag, maybe by having a small icon or button next to the shows rail title?

That is a very good point. Already working on designs for this feature. Until then, we decided not to ask the user for permission on the home screen.

MrSltun (approved):

PR #8727, opened 2023-05-22
chore: make artist list item clickable

MounirDhahri:
### Description

This PR addresses the following QA bugs reported by Pierluca:
- Show the artist preview modal instead of navigating to the user press
- Adds underlay color to artwork rows for better UX

https://github.com/artsy/eigen/assets/11945712/989e016c-f167-4ef3-aa69-0b086c96ebaa

### PR Checklist

### To the reviewers 👀

#nochanges

### DIFF
--- a/src/app/Components/ArtistListItem.tsx
+++ b/src/app/Components/ArtistListItem.tsx
@@ -14,16 +14,17 @@ import RelayModernEnvironment from "relay-runtime/lib/store/RelayModernEnvironme
 
 interface Props {
   artist: ArtistListItem_artist$data
-  relay: RelayProp
   Component?: any
-  contextModule?: string
-  withFeedback?: boolean
   containerStyle?: StyleProp<ViewStyle>
+  contextModule?: string
   disableNavigation?: boolean
   onFollowFinish?: () => void
-  uploadsCount?: number | null
+  onPress?: () => void
+  relay: RelayProp
   RightButton?: JSX.Element
   showFollowButton?: boolean
+  uploadsCount?: number | null
+  withFeedback?: boolean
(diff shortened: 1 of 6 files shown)


src/app/Components/ArtistListItem.tsx
```diff
@@ -118,6 +120,11 @@ const ArtistListItem: React.FC<Props> = ({
         <Touchable
           noFeedback={!withFeedback}
           onPress={() => {
+            if (onPress) {
+              onPress()
+              return
+            }
+
```
dariakoko, line 127: Why do we need this check here? In what case will it be false?

dariakoko: aha, because of the showArtistPreview method?

src/app/Scenes/MyCollection/Components/MyCollectionCollectedArtistsRail.tsx
```diff
@@ -57,7 +57,7 @@ export const MyCollectionCollectedArtistsRail: React.FC<MyCollectionCollectedArt
   if (!collectedArtists) return <></>
 
   return (
-    <Flex mx={-2}>
+    <Flex>
```
dariakoko, line 60: 🙏

dariakoko (approved):
LGTM ⭐

src/app/Components/ArtistListItem.tsx
MounirDhahri: yes, I would like to trigger only the onPress if it's there without navigating to the page.
I could have also had this in an `if else` block

PR #8731, opened 2023-05-22
fix: hide context menu on android

gkartalis:
This PR resolves [MOPLAT-619] 

### Description

Hides the long press feature on android.

Turns out the disabled prop on ContextMenu is working only on iOS so I did an early return if it is android return only the children, otherwise wrap with the ContextMenu

### PR Checklist

### To the reviewers 👀

#nochangelog

[MOPLAT-619]: https://artsyproduct.atlassian.net/browse/MOPLAT-619?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


brainbicycle (approved):
Good catch!

PR #8741, opened 2023-05-23
refactor(WIP): remove bundle-js job and move bundling to fastlane #doNotMerge

gkartalis:
This PR resolves [MOPLAT-771] 

### Description

- MOPLAT-771 -> get rid of bundle-js job which bundles ios for`build-test-js` (which is **NOT** needed) and might speed things up when running CI checks
- Bonus: Attempt to fix ios hermes sourcemaps

### PR Checklist

- [x] I have tested my changes on **iOS** and **Android**.
- [x] I hid my changes behind a **[feature flag]**, or they don't need one.
- [x] I have included **screenshots** or **videos**, or I have not changed the UI.
- [x] I have added **tests**, or my changes don't require any.
- [x] I added an **[app state migration]**, or my changes do not require one.
- [x] I have documented any **follow-up work** that this PR will require, or it does not require any.
- [x] I have added a **changelog entry** below, or my changes do not require one.

### To the reviewers 👀

- [ ] I would like **at least one** of the reviewers to **run** this PR on the simulator or device.

<details><summary>Changelog updates</summary>

#### Cross-platform user-facing changes

-

#### iOS user-facing changes

-

#### Android user-facing changes

-

#### Dev changes

-

</details>

Need help with something? Have a look at our [docs], or get in touch with us.

[app state migration]: ../blob/main/docs/adding_state_migrations.md
[feature flag]: ../blob/main/docs/developing_a_feature.md
[docs]: ../blob/main/docs/README.md

[MOPLAT-771]: https://artsyproduct.atlassian.net/browse/MOPLAT-771?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #8659, opened 2023-05-10
feat(iOS): Enable hermes on iOS

gkartalis:
This PR resolves [MOPLAT-763] 

### Description

Enables hermes on iOS 🔥

### PR Checklist

- [x] I have tested my changes on **iOS** and **Android**.
- [x] I hid my changes behind a **[feature flag]**, or they don't need one.
- [x] I have included **screenshots** or **videos**, or I have not changed the UI.
- [x] I have added **tests**, or my changes don't require any.
- [x] I added an **[app state migration]**, or my changes do not require one.
- [x] I have documented any **follow-up work** that this PR will require, or it does not require any.
- [x] I have added a **changelog entry** below, or my changes do not require one.

### To the reviewers 👀

- [ ] I would like **at least one** of the reviewers to **run** this PR on the simulator or device.

<details><summary>Changelog updates</summary>

#### Cross-platform user-facing changes

-

#### iOS user-facing changes

- enable hermes on iOS - gkartalis

#### Android user-facing changes

-

#### Dev changes

-

</details>

Need help with something? Have a look at our [docs], or get in touch with us.

[app state migration]: ../blob/main/docs/adding_state_migrations.md
[feature flag]: ../blob/main/docs/developing_a_feature.md
[docs]: ../blob/main/docs/README.md

[MOPLAT-763]: https://artsyproduct.atlassian.net/browse/MOPLAT-763?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #8768, opened 2023-05-26
Revert: "feat(iOS): Enable hermes on iOS"

gkartalis:
Reverts artsy/eigen#8659

due to broken sourcemaps on iOS only:

- [Sentry issue that we opened for this](https://github.com/getsentry/sentry-react-native/issues/3087)
- Another related PR https://github.com/artsy/eigen/pull/8741


brainbicycle (approved):
🙏 thanks George!

PR #8788, opened 2023-05-31
feat: open "Saves" tab when "View Saves" CTA was pressed

dimatretyak:
This PR resolves [] 

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- open "Saves" tab when "View Saves" CTA was pressed - dimatretyak

</details>

### DIFF
--- a/src/app/Scenes/MyProfile/MyProfileHeaderMyCollectionAndSavedWorks.tsx
+++ b/src/app/Scenes/MyProfile/MyProfileHeaderMyCollectionAndSavedWorks.tsx
@@ -3,7 +3,7 @@ import { VisualClueDot, VisualClueText } from "@artsy/palette-mobile"
 import { MyProfileHeaderMyCollectionAndSavedWorksQuery } from "__generated__/MyProfileHeaderMyCollectionAndSavedWorksQuery.graphql"
 import { MyProfileHeaderMyCollectionAndSavedWorks_me$data } from "__generated__/MyProfileHeaderMyCollectionAndSavedWorks_me.graphql"
 import { useCheckIfArtworkListsEnabled } from "app/Components/ArtworkLists/useCheckIfArtworkListsEnabled"
-import { StickyTabPage } from "app/Components/StickyTabPage/StickyTabPage"
+import { StickyTabPage, TabProps } from "app/Components/StickyTabPage/StickyTabPage"
 import { ArtworkListsQR } from "app/Scenes/ArtworkLists/ArtworkLists"
 import { FavoriteArtworksQueryRenderer } from "app/Scenes/Favorites/FavoriteArtworks"
 import { MyCollectionBottomSheetModals } from "app/Scenes/MyCollection/Components/MyCollectionBottomSheetModals/MyCollectionBottomSheetModals"
@@ -16,11 +16,13 @@ import {
   MyCollectionTabsStore,
   MyCollectionTabsStoreProvider,
 } from "app/Scenes/MyCollection/State/MyCollectionTabsStore"
+import { GlobalStore } from "app/store/GlobalStore"
 import { getRelayEnvironment } from "app/system/relay/defaultEnvironment"
 import { renderWithPlaceholder } from "app/utils/renderWithPlaceholder"
 import { ProvideScreenTrackingWithCohesionSchema } from "app/utils/track"
 import { screen } from "app/utils/track/helpers"
 import { compact } from "lodash"
+import { useMemo } from "react"
 import { SafeAreaView } from "react-native-safe-area-context"
 import { QueryRenderer, createRefetchContainer } from "react-relay"
 import { graphql } from "relay-runtime"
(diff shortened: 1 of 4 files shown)


src/app/Scenes/MyProfile/MyProfileHeaderMyCollectionAndSavedWorks.tsx
```diff
@@ -114,9 +119,14 @@ export const MyProfileHeaderMyCollectionAndSavedWorksScreenQuery = graphql`
 `
 
 export const MyProfileHeaderMyCollectionAndSavedWorksQueryRenderer: React.FC = () => {
+  const tabProps = useMyProfileTabProps()
+  const key = useMemo(() => `${Date.now()}`, [tabProps])
```
dimatretyak, line 123: This hack helps us to always open the "Saves" tab, even if the user has previously visited the screen

### Demo
| without hack | with hack |
| Content Cell | Content Cell |

dimatretyak: alternatively, we can use this hack **only** for `Saves' tab

```javascript
const initialTab = tabProps?.initialTab ?? Tab.collection
const key = useMemo(() => {
  if (initialTab === Tab.savedWorks) {
    return `${Date.now()}`
  }

  return "MyProfileScreen"
}, [tabProps, initialTab])

...

return (
  <ProvideScreenTrackingWithCohesionSchema
    ...
    key={key}
  >
    ...
  </ProvideScreenTrackingWithCohesionSchema>
)
```

dblandin: I imagine this is tricker because the screen is likely already loaded because it's a primary screen, associated with the navigation bar? We can't simply rely on passProps, the way we do for other screen that tend to load/render fresh upon navigation.

Might be a good question for #mobile-practice.

dimatretyak: > We can't simply rely on passProps, the way we do for other screen that tend to load/render fresh upon navigation

As an alternative approach, we can navigate to **the new screen** instead of navigating to **the tab**. Something like this


https://github.com/artsy/eigen/assets/3513494/97ccaef2-2f80-4393-8c21-3e8cb5c9cad9

src/app/routes.ts
```diff
@@ -142,6 +142,10 @@ function getDomainMap(): Record<string, RouteMatcher[] | null> {
 
     addRoute("/artwork-certificate-of-authenticity", "ArtworkCertificateAuthenticity"),
     addRoute("/artwork-classifications", "ArtworkAttributionClassFAQ"),
+    addRoute("/artwork-lists", "MyProfile", (params) => ({
+      ...params,
+      initialTab: "Saves",
+    })),
```
MounirDhahri, line 148: **suggestion:** because this route has a specific logic - it might be useful to add a test for it

src/app/Scenes/MyProfile/MyProfileHeaderMyCollectionAndSavedWorks.tsx
MounirDhahri: Smart

MounirDhahri: I would have thought about using the instructions [here](https://reactnavigation.org/docs/params/#passing-params-to-a-previous-screen) on passing params to an already opened route - might be tricky though because you are using a different route for the profile tab instead of `my-profile?initialTab=saves`

MounirDhahri (dismissed):

src/app/Scenes/MyProfile/MyProfileHeaderMyCollectionAndSavedWorks.tsx
MounirDhahri: no need to implement such a feature there @dimatretyak - we're getting rid of the sticky tabs anyway the next sprint or so and it's not worth it to investigate that route currently.

MounirDhahri: no need to implement such a feature there @dimatretyak - we're getting rid of the sticky tabs anyway the next sprint or so and it's not worth it to investigate that route currently.

dimatretyak: ahh, got it. thank you for info!

src/app/routes.ts
dimatretyak: > not sure if you would like to allow users on android to deep link to this route 

yes, I think there is no need for deep links support for artwork lists

dimatretyak: > because this route has a specific logic - it might be useful to add a test for it

Addressed here: https://github.com/artsy/eigen/pull/8788/commits/7cb6c1280204d1f5fc810e44957e212ba0d2e0e8

src/app/Scenes/MyProfile/MyProfileHeaderMyCollectionAndSavedWorks.tsx
dimatretyak: > Are we using it to [force a resetting of the state](https://react.dev/learn/preserving-and-resetting-state#option-2-resetting-state-with-a-key) of the displayed tabs?

yes, you're right

PR #8791, opened 2023-06-01
fix: partner double header

gkartalis:
This PR resolves [] 

### Description

During QA we found that when navigating to a partner from search results we were displaying double header.

The reason for that was the following:

Screenshot 2023-06-01 at 12 42 23

When querying for searchableItem we get a `slug` / `href` without the `/partner` prefix and this causes the `navigateToPartner` function to not actually use the Partner component to render the `Partner` but the `VanityURLEntity`.

This PR fixes this.

⚠️ Before sticky tabs this wasn't a problem since we didn't pass hidesBackButton [here](https://github.com/artsy/eigen/blob/febd1b641e00f14a0802d98a686ef9bc782ffb9e/src/app/AppRegistry.tsx#L490) and the react module options were the same as [VanityURLEntity](https://github.com/artsy/eigen/blob/febd1b641e00f14a0802d98a686ef9bc782ffb9e/src/app/AppRegistry.tsx#L534) so we never noticed 🤷‍♂️

| Before | After |
|---|---|
|https://github.com/artsy/eigen/assets/21178754/12aa1974-9f62-41c9-967c-0ff64206c15a|https://github.com/artsy/eigen/assets/21178754/7c41efc2-2440-49db-8fd1-7351fd16f0be|

### PR Checklist

### To the reviewers 👀

- fix: partner double header bug from search results - gkartalis

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>


src/app/system/navigation/navigate.ts
```diff
@@ -221,7 +221,9 @@ export enum SlugType {
 }
 
 export function navigateToPartner(slug: string) {
-  navigate(slug, { passProps: { entity: EntityType.Partner, slugType: SlugType.ProfileID } })
+  navigate(`partner/${slug}`, {
```
brainbicycle, line 224: good find! If we know that the entity is a partner it makes sense to navigate directly to the partner page but in theory the VanityURLEntity should be able to render partners correctly after a redirect, I think this is good to merge as is but I am wondering if we still have a bug with the vanityURLEntity rendering double headers in some cases?

brainbicycle (approved):

src/app/system/navigation/navigate.ts
```diff
@@ -221,7 +221,9 @@ export enum SlugType {
 }
 
 export function navigateToPartner(slug: string) {
```
dimatretyak, line 223: we also need to pass `slug` instead of `href` in `PartnerCard` ([proof](https://github.com/artsy/eigen/blob/4ac7169544098482b1ff8ee9dcd0064bf8e25723/src/app/Scenes/Artwork/Components/PartnerCard.tsx#L50)). otherwise we will face this


https://github.com/artsy/eigen/assets/3513494/594b299a-805a-4052-b75a-1dc57f8c36c3

dimatretyak (changes requested):

src/app/system/navigation/navigate.ts
gkartalis: true true, addressing it! Thanks Dima

brainbicycle: and maybe as an additional safety check in case we miss any spots, check if the string passed contains `partner/` before adding

src/app/system/navigation/navigate.ts
gkartalis: I don't think so 🤔 The main issue was that we were navigating to the partner page in two different components:
- VanityURLEntity from search
- Partner from deeplinks

It is also used in Fairs but we didn't change anything there since it didn't have sticky tabs.

The issue started because we added a `Screen` and `Screen.Header` to the Partner page while the `VanityURLEntity` component was rendering its own header (in AppRegistry we were hiding the back button but didn't work)

gkartalis: I don't think so 🤔 The main issue was that we were navigating to the partner page in two different components:
- VanityURLEntity from search
- Partner from deeplinks

It is also used in Fairs but we didn't change anything there since it didn't have sticky tabs.

The issue started because we added a `Screen` and `Screen.Header` to the Partner page while the `VanityURLEntity` component was rendering its own header (in AppRegistry we were hiding the back button but didn't work)

brainbicycle: in theory any entity can get routed to from the VanityURLEntity screen I believe if a shortcut url exists for it, so I do think it is still possible to get to a partner screen through VanityURLEntity, but in any case it sounds like this will address the major issue so we can address that seperately

gkartalis: perfect

dimatretyak (approved):
Thank you!

src/app/system/navigation/navigate.ts
```diff
@@ -220,8 +220,22 @@ export enum SlugType {
   FairID = "fairID",
 }
 
+export function getPartnerSlug(slug: string) {
+  const regex = /partner\//
```
dimatretyak, line 224: **nitpick**: move out of the scope of `getPartnerSlug`

dimatretyak: feel free to ignore if don't agree with it

src/app/system/navigation/navigate.tests.tsx
```diff
@@ -309,3 +309,15 @@ describe(navigate, () => {
...
+
+describe("getPartnerSlug", () => {
+  it('should remove "partner/" if it exists in the slug', () => {
+    const slug = "partner/my-partner"
+    expect(getPartnerSlug(slug)).toBe("partner/my-partner")
+  })
+
+  it('should not modify the slug if "partner/" is not present', () => {
+    const slug = "my-partner"
+    expect(getPartnerSlug(slug)).toBe(`partner/${slug}`)
```
brainbicycle, line 321: 🙌

brainbicycle (approved):

damassi (approved):

PR #8833, opened 2023-06-08
fix(lotsByFollowedArtists): make rail more defensive

gkartalis:
This PR resolves [] 

### Description

Some people reported not being able to load the app and getting unable to load, while testing with @cavvia's account found the issue to be on the `lotsByFollowedArtistsRail` on Home screen.

The issue is that on the rail we render saleArtworks but the `lotsByFollowedArtistsConnection` wasn't defensive.

As you can see above, it returns data but saleArtworks are null for all.

In this PR we hide the whole rail if the hasSaleArtworks is false and also adds some additional checks on the RailCard component.

### PR Checklist

### To the reviewers 👀

- fix(lotsByFollowedArtists): make rail more defensive - gkartalis

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Components/LotsByArtistsYouFollowRail/LotsByFollowedArtistsRail.tsx
+++ b/src/app/Components/LotsByArtistsYouFollowRail/LotsByFollowedArtistsRail.tsx
@@ -35,9 +35,9 @@ export const LotsByFollowedArtistsRail: React.FC<Props> = ({
 
   const { navigateToPageableRoute } = useNavigateToPageableRoute({ items: artworks })
 
-  const hasArtworks = artworks?.length
+  const hasSaleArtworks = artworks?.some((artwork) => artwork?.saleArtwork)
 
-  if (!hasArtworks) {
+  if (!hasSaleArtworks) {
     return null
   }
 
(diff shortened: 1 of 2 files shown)


gkartalis:
Question: now that I hide the rail if we don't have any saleArtworks the Home screen looks like this:

\Screenshot 2023-06-08 at 12 37 56

Are the paddings okay or we do need any further adjustment?

lordkiz:
Hmmmmm. How do you have lots without sale_artworks? This should not happen at all. Maybe the place to look should be in Gravity or MP.

cavvia:
The GraphQL API was returning lots which are associated with an unpublished sale, in case that helps to clarify the situation. That issue was fixed in the auction lot recs backend [here](https://github.com/artsy/motion/pull/538) and the data in the backend database (vortex) has been rebuilt. There may still be some caching at play in the API though (I'm not sure).

gkartalis:
@MounirDhahri should we close this then? Or should we keep it in place just to make it more defensive?

MounirDhahri:
I would rather close this to make sure similar regressions do not happen gracefully in the future. I understand the rationale behind being less defensive here but if we were, we wouldn't have been able to tell if it's worth prioritizing looking into it or not and it wouldn't have made it to Sentry for example

gkartalis:
yep, sounds good. Closing

brainbicycle:
@gkartalis @MounirDhahri how about we add a sentry error when this occurs but keep the code defensive? I get we want to be aware of issues like this but we shouldn't rely on user reports to get that awareness.

MounirDhahri:
Linking [this comment](https://github.com/artsy/gravity/pull/16431#pullrequestreview-1472416420) from @joeyAghion here.

damassi:
We should def open this back up and merge. This error [just appeared again](https://artsy.slack.com/archives/C02BAQ5K7/p1686327593510739?thread_ts=1686327509.477009&amp;cid=C02BAQ5K7), for the second time this week. 

Not sure where the debate is here, but: Defensive coding should always be front and center, as should prioritizing resilient systems. If we would like info around delicate areas in the code we use logging, either to the app or to a third party system like sentry; we don't let stuff bring down the whole UX, which leads to users uninstalling and never coming back.

damassi:
Thanks for the fix @gkartalis - @MounirDhahri, if there's some follow up we'll be happy to handle it or assist so that we can still get visibility into where things go wrong. For the time being however, lets now play wack a mole with every unpublished sale artwork that comes through.

PR #8834, opened 2023-06-08
feat(CX-3640): add collected artists screen

MounirDhahri:
This PR resolves [CX-3640] 

### Description

The PR is getting big and I would rather have it merged and continue the work in a different PR before it gets too big

**Still needs to be done:** 
- Adding some tests
- Adding mutations logic
___
**Adding collected artists**

https://github.com/artsy/eigen/assets/11945712/e750f774-b987-4404-b84c-b5df92bc02d7

___
**Adding an artist through the add artwork screen still works**

https://github.com/artsy/eigen/assets/11945712/815ce5fb-e2ab-4359-8d58-ec2f1c7e161b

___
**Adding an artist through add artists works well**

https://github.com/artsy/eigen/assets/11945712/2aff5132-87f2-4368-8bc7-5f7c08748bf1

### PR Checklist

### To the reviewers 👀

- add artists you collect screen - mounir

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[CX-3640]: https://artsyproduct.atlassian.net/browse/CX-3640?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/App.tsx
```diff
@@ -45,6 +46,9 @@ if (__DEV__) {
   NativeModules.DevSettings.setIsShakeToShowDevMenuEnabled(false)
 }
 
+// Required in order to use sets inside easy-peasy stores
+enableMapSet()
```
damassi, line 50: Do we _have_ to use Sets? I ask because by using a set, we instantly disqualify the store from being JSON serializable, and need to introduce a translation layer. 

Its cool that sets ensure uniqueness, but we should avoid fancy features if the same can be achieved with a `uniq` or some other lodash method.

MounirDhahri: No, actually we don't have to use sets here, I can tweak the code to use an array instead and rely on `uniq`. I hesitated before enabling this but eventually did it because we already use our own layer to manage parts of the `localStorage` and it helped making things work without changes, but I understand the rationale behind avoiding it in case we would like to get rid of our custom logic towards a shared setup as Energy. I will change it 🫡 thank you

src/app/Scenes/MyCollection/Screens/MyCollectionAddCollectedArtists/MyCollectionAddCollectedArtistsAutosuggest.tsx
```diff
@@ -0,0 +1,289 @@
...
+}
+
+const ARTIST_LIST_ITEM_HEIGHT = 100
+
+const MyCollectionAddCollectedArtistsAutosuggestPlaceholder: React.FC<{}> = ({}) => {
+  const isTablet = isPad()
+
+  const Circle = () => (
+    <Flex alignItems="center" justifyContent="center" width={isTablet ? "20%" : "50%"} mt={2}>
+      {/* <PlaceholderBox width={100} height={100} borderRadius={50} backgroundColor="black10"></PlaceholderBox> */}
```
olerichter00, line 190: 👀

olerichter00 (dismissed):
Great work 🌟

PR #8847, opened 2023-06-09
fix: patch react-native-scrollable-tab-view to fix android crashes

gkartalis:
This PR resolves [MOPLAT-811] 

### Description

Fixes `Tried to register two views with the same name RNCViewPager` crash that we found from playstore [pre-launch report](https://play.google.com/console/developers/6449739225222972501/app/4975007939329818983/pre-launch-report/overview?artifactId=4859817822073730743) and [sentry](https://artsynet.sentry.io/issues/4218775411/events/8b2928403d1c4d6daaa73bbb59cc8f6e/)

By looking into issues of `react-native-scrollable-tab-view` found [this solution](https://github.com/ptomasroos/react-native-scrollable-tab-view/issues/1133#issuecomment-800745239) that worked.

The issue was preventing users to access inbox on android devices.

#### Follow up steps

Replace `react-native-scrollable-tab-view` deprecated library with the new collapsible tabs from palette (might need design input due to different design).

The derprecated library is only used in two places, `City.tsx` and `Inbox.tsx`.

### PR Checklist

- [x] I have tested my changes on **iOS** and **Android**.
- [x] I hid my changes behind a **[feature flag]**, or they don't need one.
- [x] I have included **screenshots** or **videos**, or I have not changed the UI.
- [x] I have added **tests**, or my changes don't require any.
- [x] I added an **[app state migration]**, or my changes do not require one.
- [x] I have documented any **follow-up work** that this PR will require, or it does not require any.
- [x] I have added a **changelog entry** below, or my changes do not require one.

### To the reviewers 👀

- [ ] I would like **at least one** of the reviewers to **run** this PR on the simulator or device.

<details><summary>Changelog updates</summary>

#### Cross-platform user-facing changes

-

#### iOS user-facing changes

-

#### Android user-facing changes

-

#### Dev changes

- patch react-native-scrollable-tab-view to fix android crashes - gkartalis

</details>

Need help with something? Have a look at our [docs], or get in touch with us.

[app state migration]: ../blob/main/docs/adding_state_migrations.md
[feature flag]: ../blob/main/docs/developing_a_feature.md
[docs]: ../blob/main/docs/README.md

[MOPLAT-811]: https://artsyproduct.atlassian.net/browse/MOPLAT-811?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #8851, opened 2023-06-09
fix: main patch issues

gkartalis:
This PR resolves [] 

### Description

Attempt to fix main patch issues due to https://github.com/artsy/eigen/pull/8847. 

More context: merging the PR above #8847 caused some issues with ci as you can see on the checks but only when merged to main.

Then I opened https://github.com/artsy/eigen/pull/8848 for refreshing the cache hoping to fix the issue but the issue remained on main only 🤔.

### Other fixes @brainbicycle
This ended up requiring a few fixes:
- updated our android images to non-deprecated version
- added ruby orb to ensure we have correct ruby installed on machines, was causing a bundle install failure (aside, it seems we run ruby and gem installation too often worth looking into)
- removed `set -u` from our scripts around android, this causes hard fails for unbounded vars and apparently there is one in rvm in the ruby orb

### PR Checklist

### To the reviewers 👀

#nochangelog


dimatretyak (dismissed):

dimatretyak:
oops, can't apply patch locally

image

gkartalis:
because theoretically you don't need to apply the package locally 🤔 running `yarn install:all` should automatically apply it for you

dimatretyak:
> @dimatretyak what if you do `rm -rf node_modules` first and then just run `yarn install:all`

this time I didn't face this problem when applying the patch after I initially made `rm -rf node_modules`

gkartalis:
amazing! Thanks for the confirmation! Will ping people at #practice-mobile to run `rm -rf node_modules` first

PR #8856, opened 2023-06-09
feat: Prefer default images in widget GRO-1756

jonallured:
Turns out another thing we should do is prefer the image that has been marked as `is_default` so this PR just updates the logic like so:

* if there's a default image, use that
* if not then find the image with the lowest position
* if there still isn't an image then use a fallback

Note: we shouldn't even reach that third branch but it's there just in case.

https://artsyproduct.atlassian.net/browse/GRO-1756

/cc @artsy/grow-devs @isakelaris


brainbicycle (approved):
looks good! ⚡

PR #8878, opened 2023-06-14
feat: remove lists from release for 8.14.0

gkartalis:
This PR resolves [] 

### Description

⚠️ this is a PR to reduce the future work to us and @artsy/fx-devs (not clear exactly on what the status is for the lists right now so opened it to have it ready in case we want to remove this release)

Disable `AREnableArtworkLists` from `8.14.0` in order to not to have to refresh the feature flag.

Refreshing the feature flag means that if this release goes live like this we will need to open two follow-up PRs:
- eigen to introduce a new ff for the feature
- echo to introduce a new ff for the feature to toggle remotely

### PR Checklist

### To the reviewers 👀

#nochangelog

### DIFF
--- a/src/app/store/config/features.ts
+++ b/src/app/store/config/features.ts
@@ -231,10 +231,9 @@ export const features: { [key: string]: FeatureDescriptor } = {
     showInDevMenu: true,
   },
   AREnableArtworkLists: {
-    readyForRelease: true,
+    readyForRelease: false,
     description: "Enable Artwork Lists",
     showInDevMenu: true,
-    echoFlagKey: "AREnableArtworkLists",
   },
   AREnableNewAuctionsRailCard: {
     description: "Enable New Auctions Home Rail Card",
--- a/src/app/Scenes/Artwork/Components/ArtworkActions.tests.tsx
+++ b/src/app/Scenes/Artwork/Components/ArtworkActions.tests.tsx
@@ -137,21 +137,46 @@ describe("ArtworkActions", () => {
   })
 
   describe("Save button", () => {
-    it("should trigger save mutation when user presses save button", () => {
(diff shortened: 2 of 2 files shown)


src/app/store/config/features.ts
```diff
@@ -231,10 +231,9 @@ export const features: { [key: string]: FeatureDescriptor } = {
     showInDevMenu: true,
   },
   AREnableArtworkLists: {
-    readyForRelease: true,
+    readyForRelease: false,
     description: "Enable Artwork Lists",
     showInDevMenu: true,
-    echoFlagKey: "AREnableArtworkLists",
```
dblandin, line 237: Since this was previously marked as `readyForRelease` and toggled remotely via Echo in 8.13.0, should we consider adding a new feature flag to the next public-facing release in 8.15.0 (hopefully)?

The `AREnableArtworkLists` flag as it currently exists only affects internal Artsy users, so maybe we keep it that way until we're ready to remove the feature flags entirely post launch?

gkartalis: We could do it like this.

The issue though is that in the [release candidate diff](https://github.com/artsy/eigen/tree/gkartalis/8.14.0-release-candidate) the `useCheckIfArtworkListsEnabled` is not there anymore.

That means that in order to hide it from this release (8.14.0) we should either:
- `flip it to false in echo` && refresh the feature flag as you said
or
- disable it with this pr from readyForRelease

dblandin: Ahh, I see. Yeah, let's go ahead and disable with this PR for now then 👍🏼 

We can explain that anyone looking for the feature can enable it explicitly in the admin menu.

dblandin (approved):

PR #8934, opened 2023-06-27
feat(PRO-159): add initiator ID to self-initiated IDV requests

mdole:
This PR resolves [PRO-159] 

### Description

Companion to https://github.com/artsy/gravity/pull/16451. We want to store the ID of the user who initiated an IDV request.

I didn't test this one personally since it's a small change and I haven't run Eigen in a while, so figured I'd get some help rather than fighting with sims for a day. @gkartalis thanks for volunteering to help with that! LMK if you either a) have changed your mind and want me to do it, or b) want to do it together 🙂 

### PR Checklist

### To the reviewers 👀

#### Dev changes

- add initiator ID to self-initiated IDV requests

</details>

[PRO-159]: https://artsyproduct.atlassian.net/browse/PRO-159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/MyProfile/useHandleVerification.tsx
+++ b/src/app/Scenes/MyProfile/useHandleVerification.tsx
@@ -4,36 +4,19 @@ import { verifyEmail } from "app/utils/verifyEmail"
 import { verifyID } from "app/utils/verifyID"
 import { useCallback, useState } from "react"
 
-type verificationType = "ID" | "Email"
-
-const IDVerificationBlockingStates = ["passed", "failed", "watchlist_hit"]
+const VERIFICATION_BANNER_TIMEOUT = 6000
+const ID_VERIFICATION_BLOCKING_STATES = ["passed", "failed", "watchlist_hit"]
 
 export const useHandleEmailVerification = () => {
-  return useHandleVerification("Email")
-}
-
-export const useHandleIDVerification = () => {
-  return useHandleVerification("ID")
-}
-
-const VERIFICATION_BANNER_TIMEOUT = 6000
-
-const useHandleVerification = (type: verificationType) => {
   const [showVerificationBanner, setShowVerificationBanner] = useState(false)
 
   const handleVerification = useCallback(async () => {
     try {
-      const response = await verify(type)
+      await verifyEmail(getRelayEnvironment())
 
-      // if response as state from ID verification
-      const IDVerificationState = !(response && IDVerificationBlockingStates.includes(response))
-
-      if (type === "Email" || IDVerificationState) {
-        setShowVerificationBanner(true)
-      }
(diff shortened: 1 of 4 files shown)


src/app/Scenes/MyProfile/useHandleVerification.tsx
```diff
@@ -1,39 +1,50 @@
-import { captureException } from "@sentry/react-native"
 import { getRelayEnvironment } from "app/system/relay/defaultEnvironment"
 import { verifyEmail } from "app/utils/verifyEmail"
 import { verifyID } from "app/utils/verifyID"
 import { useCallback, useState } from "react"
 
-type verificationType = "ID" | "Email"
-
-const IDVerificationBlockingStates = ["passed", "failed", "watchlist_hit"]
+const VERIFICATION_BANNER_TIMEOUT = 6000
```
mdole, line 6: The code here felt difficult to read and unnecessarily joined, and would have added more complexity with the addition of a parameter that only applied to one method. Felt like a simplification to make them individual methods, albeit with some duplication.

src/app/Scenes/MyProfile/useHandleVerification.tsx
```diff
@@ -1,39 +1,50 @@
...
+      const IdvHasValidState =
+        typeof verificationState === "string" &&
+        !ID_VERIFICATION_BLOCKING_STATES.includes(verificationState)
 
-      if (type === "Email" || IDVerificationState) {
+      if (IdvHasValidState) {
         setShowVerificationBanner(true)
       }
     } catch (error) {
-      captureException(error)
```
mdole, line 35: I _think_ this may never have been tested/might not work as expected. When I tried using it from a test env with an error I copy/pasted from an actual mutation error response from Insomnia, the error was an array with several properties, and the call to `captureException` was like "hey hold on this should be a string!"

Not sure if this is better; maybe we should just try to do a better job parsing the error from MP and passing a string rather than an array 🤷

gkartalis (approved):
Looks great @mdole!

src/app/Scenes/MyProfile/useHandleVerification.tsx
iskounen: You sort of flipped `useHandleVerification` inside out with this change, and I agree that is a better way to read and organize this code. It allows for the mechanics that both methods share to be extracted and modularized in a more reusable way 👍

iskounen (approved):
This looks good to me. Thanks for wrapping this work up 🎉

I'll reassign to you in case you would like to check the "tested on Android and iOS" box first.

PR #8941, opened 2023-06-28
feat: add hero unit rail back GRO-1749

jonallured:
This PR adds a rail to the top of the home feed for hero units. Props to everybody that helped me get this together!! 😅 

The goal is for this to be a case of no visual change but a new source of the content. Previously we were using the Braze SDK to populate a rail of Content Cards but now we can hit MP as normal and get back the same stuff. Looks like this:

Screenshot 2023-06-28 at 5 02 22 PM

https://artsyproduct.atlassian.net/browse/GRO-1749

/cc @artsy/grow-devs


src/app/Scenes/Home/Home.tsx
```diff
@@ -135,8 +138,6 @@ const Home = memo((props: HomeProps) => {
   const prefetchUrl = usePrefetch()
   const tracking = useTracking()
 
-  const { cards } = useContentCards()
-
```
gkartalis, line 139: 👋

src/app/Scenes/Home/useHomeModules.ts
```diff
@@ -18,6 +17,14 @@ export const useHomeModules = (props: HomeProps, cards: ReactAppboy.CaptionedCon
...
     const allModules: Array<HomeModule> = [
       // Above-The-Fold Modules
+      {
+        data: props.heroUnits,
+        isEmpty: isEmpty(props.heroUnits),
+        key: "heroUnitsRail",
+        prefetchUrl: "",
+        title: "",
+        type: "heroUnits",
+      },
```
gkartalis, line 27: **question:** Order matters here to display the rails, do we want the content cards to appear as the first rail or do we want them to be where they were before the change? (I know that I moved it further up but it was only for debugging purposes, eitherway I leave the decision up to you @jonallured )

src/app/Scenes/Home/Components/HeroUnitsRail.tsx
```diff
@@ -0,0 +1,121 @@
+import { Box, Flex, Spacer, Text, Button, Touchable } from "@artsy/palette-mobile"
+import {
+  HeroUnitsRail_heroUnitsConnection$data,
+  HeroUnitsRail_heroUnitsConnection$key,
+} from "__generated__/HeroUnitsRail_heroUnitsConnection.graphql"
+import OpaqueImageView from "app/Components/OpaqueImageView/OpaqueImageView"
```
gkartalis, line 6: One suggestion here, lets use the OpaqueImageView2 here since its the new image component that we use on eigen (on vscode this might be more easily detectable since the other image component is marked as deprecated)

```suggestion
import { OpaqueImageView } from "app/Components/OpaqueImageView2"
```

src/app/Scenes/Home/Components/HeroUnitsRail.tsx
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  return (
+    <Touchable key={item.internalID} onPress={handlePress}>
+      <Flex bg="black100" flexDirection="row" height={CARD_HEIGHT} width={screenWidth}>
+        <OpaqueImageView
+          height={CARD_HEIGHT}
+          imageURL={item.image?.imageURL}
+          resizeMode="cover"
+          width={cardImageWidth}
```
gkartalis, line 45: The api of the new image is the same (already checked locally that it doesn't break the design) but it doesn't need the `resizeMode`

```suggestion
        <OpaqueImageView
          height={CARD_HEIGHT}
          imageURL={item.image?.imageURL}
          width={cardImageWidth}
```

gkartalis (approved):
Looks 🔥 left some suggestions but other than that I leave it up to you 👏

src/app/Scenes/Home/useHomeModules.ts
jonallured: Oh good call! I am confirming this order with @jpoczik and will probably move it back down once I hear from her. 👍

src/app/Scenes/Home/Components/HeroUnitsRail.tsx
jonallured: Done! 🙌

damassi: Use the `Image` component from Palette!

src/app/Scenes/Home/Components/HeroUnitsRail.tsx
damassi: `import { Image } from '@artsy/palette-mobile'`

damassi (commented):
Two suggestions would be to use the new Image component from Palette, which we're trying to use more widely, instead of the OpaqueImageView

src/app/Scenes/Home/Components/HeroUnitsRail.tsx
gkartalis: Oh good point! Is it ready to be used everywhere @damassi ?

damassi: Yep, seems good to go! If there's something that's missing we should add it over there to be used here, but seems fine as is right now.

gkartalis: Amazing!

gkartalis: https://github.com/artsy/eigen/pull/8946 opened this to mark it as deprecated and point people to use the new one 🌟

src/app/Scenes/Home/useHomeModules.ts
jonallured: Confirmed: this should move back to where it was - under new works for you. 👍

src/app/Scenes/Home/Components/HeroUnitsRail.tsx
jonallured: Ok got it - moved to using palette mobile's Image component 👍

src/app/Scenes/Home/Components/HeroUnitsRail.tsx
```diff
@@ -0,0 +1,116 @@
...
+}
+
+export const homeFragment = graphql`
+  fragment HeroUnitsRail_heroUnitsConnection on HeroUnitConnection {
+    edges {
+      node {
+        internalID
+        body
+        credit
+        image {
```
damassi, line 104: I (think) you'll also need to pass the aspectRatio

jonallured: I ran this through the sim and didn't need aspectRatio 🤷

damassi: even better

PR #8955, opened 2023-07-05
feat: add more performant masonry grid #doNotMerge

gkartalis:
This PR resolves [PHIRE-3] 

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[PHIRE-3]: https://artsyproduct.atlassian.net/browse/PHIRE-3?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
+++ b/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
@@ -1,5 +1,18 @@
 import { ActionType, ContextModule, OwnerType } from "@artsy/cohesion"
-import { Spacer, Box, Text, Button, Tabs, BellIcon } from "@artsy/palette-mobile"
+
+import {
+  Text,
+  Button,
+  Tabs,
+  BellIcon,
+  Spacer,
+  Box,
+  Flex,
+  useSpace,
+  useScreenDimensions,
+} from "@artsy/palette-mobile"
+import { useScreenScrollContext } from "@artsy/palette-mobile/dist/elements/Screen/ScreenScrollContext"
+import { MasonryFlashList } from "@shopify/flash-list"
 import { ArtistArtworks_artist$data } from "__generated__/ArtistArtworks_artist.graphql"
 import { ArtistArtworksFilterHeader } from "app/Components/Artist/ArtistArtworks/ArtistArtworksFilterHeader"
 import { useShowArtworksFilterModal } from "app/Components/Artist/ArtistArtworks/hooks/useShowArtworksFilterModal"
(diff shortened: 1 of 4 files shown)


src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
```diff
@@ -193,8 +240,88 @@ const ArtistArtworksContainer: React.FC<ArtworksGridProps & ArtistArtworksContai
...
+          viewAreaCoveragePercentThreshold: 20,
+        }}
+        onViewableItemsChanged={onViewableItemsChanged}
+        contentContainerStyle={{
+          paddingHorizontal: space(2),
+          paddingBottom: space(4),
+          backgroundColor: "purple",
+        }}
+        // onEndReached={() => console.warn("onEndReached")}
+        // onEndReachedThreshold={0.1}
```
gkartalis, line 320: **onEndReached prop: (generally the correct way of fetching more data on an infinite scroll list)** 

If wrapped with Flex the onEndReached is getting called on the **first** render only.

if wrapped with Flex with 1 and another one with the height (apparently Masonry really likes being wrapped with a [**fixed** height](https://shopify.github.io/flash-list/docs/known-issues#1-flashlists-rendered-size-is-not-usable-warning) parent ) the onEndreached is getting called correctly **BUT** breaks the scrolling functionality of the sticky tabs.

src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
```diff
@@ -193,8 +240,88 @@ const ArtistArtworksContainer: React.FC<ArtworksGridProps & ArtistArtworksContai
...
+        }}
+        // onLoad={({ elapsedTimeInMs }) => {
+        //   console.log("List Load Time", elapsedTimeInMs)
+        // }}
+        viewabilityConfig={{
+          // itemVisiblePercentThreshold: 50,
+          waitForInteraction: true,
+          viewAreaCoveragePercentThreshold: 20,
+        }}
+        onViewableItemsChanged={onViewableItemsChanged}
```
gkartalis, line 313: **`onViewableItemsChanged` + `viewabilityConfig`** workaround:

tried the following workaround:

```js
viewabilityConfig={{
  waitForInteraction: true,
  viewAreaCoveragePercentThreshold: 20, // played around with different values here
}}

onViewableItemsChanged={
  useCallback(

({ changed }) => {


// check if the last item of each page is visible to fetch more
// artworksPageSize is basically :
// const artworks = extractNodes(artist.artworks)
// const artworksLength = artworks?.length ?? 0
// const artworksPageSize = artworksLength / 10
// in order to have a hacky way of tracking the page to fetch more items when 
// we need to in order to have an infinite scroll experience

const isEnd = changed.some(
(obj) => obj.index === PAGE_SIZE * artworksPageSize - 1 && obj.isViewable
)



relay.loadMore(PAGE_SIZE)

}

},

[artworksPageSize]

)
}
```


The issue with the above was that the `onViewableItemsChanged` callback is not aware of any state changes / variable changes within the component. 

This results in artworksPageSize getting incremented BUT within the context of `onViewableItemsChanged` it is always 1.

src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
```diff
@@ -193,8 +240,88 @@ const ArtistArtworksContainer: React.FC<ArtworksGridProps & ArtistArtworksContai
...
+        contentContainerStyle={{
+          paddingHorizontal: space(2),
+          paddingBottom: space(4),
+          backgroundColor: "purple",
+        }}
+        // onEndReached={() => console.warn("onEndReached")}
+        // onEndReachedThreshold={0.1}
+      />
```
gkartalis, line 321: Workaround with ListFooterComponent:

Having a 'load more' button on the end of the list to fetch more data solves our issues but breaks the infinite scroll experience 😅

gkartalis:
Another approach would be to add Masonry to the collapsible-tabs library that we use but I highly doubt that it would work due to already having some layout issues with flashlist as a [known issue](https://github.com/PedroBern/react-native-collapsible-tab-view/issues/335) for FlashList (<MasonryFlashList is two flashlists under the hood).

In order to also add Masonry to the collapsible tabs we need to do something similar to [this](https://github.com/PedroBern/react-native-collapsible-tab-view/blob/main/src/FlashList.tsx) but the issue is that Flashlist and MasonryFlashList onScroll events work a bit differently (have already opened an [issue](https://github.com/Shopify/flash-list/issues/689) about that on flashlist repo).

And that makes me believe that this won't work either way.

gkartalis:
Feel free to review or play around with it and let me know if I missed something. Not ready but open for feedback eitherway on my comments mostly.

araujobarret (commented):
Thanks for all the exploration here first of all!

Seeing how many issues/hacks are necessary I don't feel super confident in using it, might become more of a problem than a helper.
Another downside that limits the usage of the Masonry is to need of having a parent View with a fixed height, not all the time we will have this value.

gkartalis:
Waiting for design feedback to see how we proceed with that. Will close this till then and will reopen a new pr once we have an actual guideline

PR #8961, opened 2023-07-06
feat(DIA-40): change editorial in Artist/About tab

araujobarret:
This PR resolves [DIA-40] 

### Description

Changes the Editorial Rails in Artist -> About tab to the new design.

https://github.com/artsy/eigen/assets/15792853/d2c483e6-b1b3-460d-8edc-315ac1125984

### PR Checklist

### To the reviewers 👀

- change the Editorial section in Artist/About tab - araujobarret

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[DIA-40]: https://artsyproduct.atlassian.net/browse/DIA-40?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Components/Artist/Articles/Article.tsx
+++ b/src/app/Components/Artist/Articles/Article.tsx
@@ -1,7 +1,8 @@
-import { Flex, Image, Spacer, Text, Touchable } from "@artsy/palette-mobile"
+import { Flex, Spacer, Text, Touchable } from "@artsy/palette-mobile"
 import { Articles_article$key } from "__generated__/Articles_article.graphql"
 import { navigate } from "app/system/navigation/navigate"
 import { useFeatureFlag } from "app/utils/hooks/useFeatureFlag"
+import FastImage from "react-native-fast-image"
 import { useFragment, graphql } from "react-relay"
 
 interface ArticleProps {
@@ -24,11 +25,19 @@ export const Article: React.FC<ArticleProps> = ({ article, headline = false }) =
   return (
     <Touchable onPress={handleOnPress}>
       <Flex width="100%" overflow="hidden">
-        <Image
-          src={data.thumbnailImage?.url ?? ""}
-          aspectRatio={1.5}
-          width={headline ? 350 : 175}
-        />
+        <Flex width="100%" style={{ aspectRatio: 1.5 }}>
+          {/* TODO: palette Image doesn't work with percentages, we can change it
+              as soon as the palette element works
+           */}
+          <FastImage
+            source={{ uri: data.thumbnailImage?.url ?? "" }}
+            style={{
+              width: "100%",
+              height: "100%",
+            }}
+            resizeMode="cover"
+          />
+        </Flex>
         <Spacer y={1} />
         <Text variant="xs" color="black100">
           {data.vertical}
--- a/src/app/Components/Artist/Articles/Article.tsx
+++ b/src/app/Components/Artist/Articles/Article.tsx
@@ -28,7 +28,6 @@ export const Article: React.FC<ArticleProps> = ({ article, headline = false }) =
           src={data.thumbnailImage?.url ?? ""}
           aspectRatio={1.5}
           width={headline ? 350 : 175}
-          backgroundColor="blue10"
         />
         <Spacer y={1} />
         <Text variant="xs" color="black100">
--- a/src/app/Components/Artist/Articles/Article.tsx
+++ b/src/app/Components/Artist/Articles/Article.tsx
@@ -1,0 +1,58 @@
+import { Flex, Image, Spacer, Text, Touchable } from "@artsy/palette-mobile"
+import { Articles_article$key } from "__generated__/Articles_article.graphql"
+import { navigate } from "app/system/navigation/navigate"
+import { useFeatureFlag } from "app/utils/hooks/useFeatureFlag"
+import { useFragment, graphql } from "react-relay"
+
+interface ArticleProps {
+  article: Articles_article$key
+  headline?: boolean
+}
+
+export const Article: React.FC<ArticleProps> = ({ article, headline = false }) => {
+  const data = useFragment(articleQuery, article)
+  const enableNativeArticleView = useFeatureFlag("AREnableNativeArticleView")
+
+  const handleOnPress = () => {
+    if (enableNativeArticleView) {
+      navigate(`/article2/${data.internalID}`)
+    } else {
+      navigate(data.href!)
+    }
+  }
+
+  return (
+    <Touchable onPress={handleOnPress}>
+      <Flex width="100%" overflow="hidden">
+        <Image
+          src={data.thumbnailImage?.url ?? ""}
+          aspectRatio={1.5}
+          width={headline ? 350 : 175}
+          backgroundColor="blue10"
+        />
+        <Spacer y={1} />
+        <Text variant="xs" color="black100">
+          {data.vertical}
+        </Text>
+        <Spacer y={1} />
+        <Text variant={headline ? "lg-display" : "sm-display"}>{data.thumbnailTitle}</Text>
+        <Spacer y={1} />
+        <Text variant="xs" color="black60">{`By ${data.byline}`}</Text>
+        {!headline && <Spacer y={4} />}
+      </Flex>
+    </Touchable>
+  )
+}
+
+const articleQuery = graphql`
+  fragment Articles_article on Article {
+    internalID
+    href
+    thumbnailImage {
+      url(version: "large")
+    }
+    thumbnailTitle
+    vertical
+    byline
+  }
+`
--- a/src/app/Components/Artist/Articles/Articles.tests.tsx
+++ b/src/app/Components/Artist/Articles/Articles.tests.tsx
@@ -33,10 +33,12 @@ describe("Articles", () => {
   it("renders", () => {
     renderWithRelay()
 
-    expect(screen.getByText(`Artsy Editorial Featuring <mock-value-for-field-"name">`)).toBeTruthy()
-    expect(screen.getByText(`<mock-value-for-field-"vertical">`)).toBeTruthy()
-    expect(screen.getByText(`<mock-value-for-field-"thumbnailTitle">`)).toBeTruthy()
-    expect(screen.getByText(`By <mock-value-for-field-"byline">`)).toBeTruthy()
+    expect(
+      screen.getByText(`Artsy Editorial Featuring <mock-value-for-field-"name">`)
+    ).toBeOnTheScreen()
+    expect(screen.getByText(`<mock-value-for-field-"vertical">`)).toBeOnTheScreen()
+    expect(screen.getByText(`<mock-value-for-field-"thumbnailTitle">`)).toBeOnTheScreen()
+    expect(screen.getByText(`By <mock-value-for-field-"byline">`)).toBeOnTheScreen()
   })
 
   it("navigates to a specific article via webview", () => {
--- a/src/app/Components/Artist/Articles/Articles.tests.tsx
+++ b/src/app/Components/Artist/Articles/Articles.tests.tsx
@@ -1,0 +1,65 @@
+import { screen, fireEvent } from "@testing-library/react-native"
+import { ArticlesTestsQuery } from "__generated__/ArticlesTestsQuery.graphql"
+import { __globalStoreTestUtils__ } from "app/store/GlobalStore"
+import { navigate } from "app/system/navigation/navigate"
+import { extractNodes } from "app/utils/extractNodes"
+import { setupTestWrapper } from "app/utils/tests/setupTestWrapper"
+import { graphql } from "react-relay"
+import { Articles } from "./Articles"
+
+describe("Articles", () => {
+  const { renderWithRelay } = setupTestWrapper<ArticlesTestsQuery>({
+    Component: ({ artist, articlesConnection }) => {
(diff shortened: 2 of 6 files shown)


src/app/Components/Artist/Articles/Article.tsx
```diff
@@ -0,0 +1,58 @@
...
+  article: Articles_article$key
+  headline?: boolean
+}
+
+export const Article: React.FC<ArticleProps> = ({ article, headline = false }) => {
+  const data = useFragment(articleQuery, article)
+  const enableNativeArticleView = useFeatureFlag("AREnableNativeArticleView")
+
+  const handleOnPress = () => {
+    if (enableNativeArticleView) {
```
araujobarret, line 17: that comes from ArticleCard component, @damassi was implementing it and didn't want to break his current idea.

Chriss, maybe you can say why the idea of the ff here?

gkartalis: @araujobarret the reason for this is that we worked recently on replacing the previous articles implementation from webviews to native views. 

It is possible to be released in this release waiting for feedback on that, but this is why the feature flag exists

araujobarret: was confused because in the other PR `AREnableNativeArticleView` was never put in `features.ts` and Echo, so I thought it could be something that was stopped in the middle of the implementation.

gkartalis: https://github.com/artsy/eigen/blob/main/src/app/store/config/features.ts#L223 it is in both [features](https://github.com/artsy/eigen/blob/main/src/app/store/config/features.ts#L223) and [echo](https://github.com/artsy/echo/blob/main/Echo.json5#L79) though on main 🤔, oh maybe you meant via the gitblame that took you to the PR

src/app/Components/Artist/ArtistAbout/ArtistAbout.tsx
```diff
@@ -103,7 +104,7 @@ export const ArtistAboutContainer = createFragmentContainer(ArtistAbout, {
           }
         }
       }
-      articles: articlesConnection(first: 10, inEditorialFeed: true) {
+      articlesConnection(first: 5, inEditorialFeed: true) {
```
gkartalis, line 107: 👏

src/app/Components/Artist/Articles/Article.tsx
araujobarret: The ones you sent they're different, from features is `AREnableNativeArticleView` and from Echo is `AREnableNativeArticles`.

src/app/Components/Artist/ArtistAbout/ArtistAbout.tests.tsx
```diff
@@ -57,4 +57,25 @@ describe("ArtistAbout", () => {
       expect(screen.UNSAFE_queryByType(ArtistAboutShowsFragmentContainer)).toBeTruthy()
     })
   })
+
+  describe("ArtistAboutEditorial", () => {
+    it("renders editorial section", () => {
+      renderWithRelay({
+        ArticleConnection: () => ({
+          edges: [{ node: { internalID: "article-id" } }],
+        }),
```
gkartalis, line 66: I think we don't need these lines since in order for https://github.com/artsy/eigen/pull/8961/files#diff-feb8d0132d497c61ec3870b4eba400f3658f642859bbcb3dcaccc45f5ef50512R70 to be truthy we only need the name of the artist

src/app/Components/Artist/ArtistAbout/ArtistAbout.tests.tsx
```diff
@@ -57,4 +57,25 @@ describe("ArtistAbout", () => {
...
+
+      expect(screen.getByText("Artsy Editorial Featuring Andy Warhol")).toBeTruthy()
+    })
+
+    it("does not render when there are no articles", () => {
+      renderWithRelay({
+        ArticleConnection: () => ({ edges: null }),
+      })
+
+      expect(screen.queryByText("Artsy Editorial Featuring", { exact: false })).toBeNull()
```
gkartalis, line 78: **question:** any reason why we need `{ exact: false }` here?

src/app/Components/Artist/Articles/Articles.tests.tsx
```diff
@@ -0,0 +1,65 @@
...
+          }
+        }
+      }
+    `,
+  })
+
+  it("renders", () => {
+    renderWithRelay()
+
+    expect(screen.getByText(`Artsy Editorial Featuring <mock-value-for-field-"name">`)).toBeTruthy()
```
gkartalis, line 36: **suggestion(super minor and non-blocking):**  but there is a "newer" way of using `toBeTruthy` `toBeNull` coming from `@testing-library/jest-native` that we use that does the same thing but it is more in sync with the web testing library which is the [`toBeOnTheScreen`](https://github.com/testing-library/jest-native#tobeonthescreen()) or `not.toBeOnTheScreen()`

gkartalis (commented):
Looks great! Left some suggestions/questions but great job

src/app/Components/Artist/Articles/Articles.tests.tsx
```diff
@@ -0,0 +1,65 @@
...
+    expect(screen.getByText(`<mock-value-for-field-"vertical">`)).toBeTruthy()
+    expect(screen.getByText(`<mock-value-for-field-"thumbnailTitle">`)).toBeTruthy()
```
gkartalis, line 38: **question:** should we use the [resolver](https://github.com/artsy/eigen/pull/8961/files#diff-601fdf18fb8f1fdb833ee754cbb41ff576da1c02079809d91f97cf7f9ac43a6bR34) in order to have the actual thumbnailTitle value and not the autoGenerated one since we assert on those?

src/app/Components/Artist/Articles/Article.tsx
gkartalis: if you see though the [line 227](https://github.com/artsy/eigen/blob/main/src/app/store/config/features.ts#L227) on features.ts you'll see that the `echoFlagKey` which is the one that "connects" the flag with echo is `AREnableNativeArticles` which is the same with echo 😅

Sorry for the confusion there, but the plan is to get rid of the feature flag theoretically before releasing in order to also have deeplinks working with articles, but it is still behind the ff till we address some launch blocking feedback. Will mark it as false in eigen EOD today if we don't have news on it!

src/app/Components/Artist/Articles/Article.tsx
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  return (
+    <Touchable onPress={handleOnPress}>
+      <Flex width="100%" overflow="hidden">
+        <Image
+          src={data.thumbnailImage?.url ?? ""}
+          aspectRatio={1.5}
+          width={headline ? 350 : 175}
+          backgroundColor="blue10"
```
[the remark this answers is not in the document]
araujobarret: good catch, it wasn't!

src/app/Components/Artist/Articles/Article.tsx
araujobarret: ahhh I got confused but was my fault since most of the other ffs we use the same `key` and `echoFlagKey` 🙇 

So I assume it's same to keep the ff check here, right?

gkartalis: no worries ❤️ yep lets keep it and I will clean it up when we are releasing 🧹

src/app/Components/Artist/Articles/Articles.tests.tsx
gkartalis: Exactly! Its the exact replacement of that because it didn't previously exist on the library ⚡️

src/app/Components/Artist/Articles/Articles.tests.tsx
araujobarret: imo there's no need for adding a mocked value to replace them unless we strongly feel to have more human-readable mocked data. That case already have value, just the value itself is kinda mocked ugly boy

src/app/Components/Artist/Articles/Articles.tsx
```diff
@@ -1,44 +1,59 @@
...
+
+      <Article article={articlesData[0]} headline />
+
+      <Spacer y={2} />
+
+      <Flex flex={1} flexShrink={0} flexGrow={1}>
+        <MasonryStatic
+          data={articlesData.slice(1, articlesData.length)}
+          renderItem={({ item }) => <Article article={item} />}
+          numColumns={2}
```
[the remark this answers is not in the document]
araujobarret: We don't have tablet sizes in the design unfortunately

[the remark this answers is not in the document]
gkartalis: We do have some users though that use tablets, it might be worth a quick ipad simulator check that it is not really ugly when rendered on a tablet 🤷 but I leave it up to you to follow up **if you want** with design 🌟

src/app/Components/Artist/Articles/Articles.tests.tsx
gkartalis: no strong preference from my side either 🤷 feel free to ignore the comment

src/app/Components/Artist/ArtistAbout/ArtistAbout.tests.tsx
araujobarret: to be more false positive proof, in case someone changes something and instead of `Andy Warhol` uses `mocked-data-blah`, with this we're really safe to check when we render or not

src/app/Components/MasonryStatic.tsx
```diff
@@ -0,0 +1,49 @@
...
+}
+
+/**
+ * Mansory static list, ideal for not soo long lists using the masonry layout
+ * do not use this component for infinite loading lists or a too large dataset
+ */
+export function MasonryStatic<T>({
+  data,
+  renderItem,
+  numColumns = 1,
```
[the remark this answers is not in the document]
araujobarret: Yeah, it makes sense, lemme change

src/app/Components/Artist/ArtistAbout/ArtistAbout.tests.tsx
gkartalis: oh gotcha! Then maybe its better to use due to this reference in the library [docs](https://callstack.github.io/react-native-testing-library/docs/api-queries#precision)


```suggestion
      expect(screen.queryByText(/Artsy Editorial Featuring/)).not.toBeOnTheScreen()
```

(the TLDR; of it is: `In most cases using a regex instead of a string gives you more control over fuzzy matching and should be preferred over { exact: false }.`)

src/app/Components/Artist/Articles/Articles.tsx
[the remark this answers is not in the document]
araujobarret: With the latest commit the percentages are way better in different size and density devices:
image

src/app/Components/Artist/Articles/Articles.tests.tsx
araujobarret: done in https://github.com/artsy/eigen/pull/8961/commits/e735d02ca008d8505893020ed177b41abb303073

src/app/Components/Artist/ArtistAbout/ArtistAbout.tests.tsx
araujobarret: done in https://github.com/artsy/eigen/pull/8961/commits/e735d02ca008d8505893020ed177b41abb303073

src/app/Components/Artist/ArtistAbout/ArtistAbout.tests.tsx
araujobarret: done in https://github.com/artsy/eigen/pull/8961/commits/f73636c57c42ec17b24756c89c1b6d4d62a33775

src/app/Components/MasonryStatic.tsx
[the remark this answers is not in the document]
araujobarret: done in https://github.com/artsy/eigen/pull/8961/commits/e735d02ca008d8505893020ed177b41abb303073

araujobarret:
@gkartalis thanks again for the review! all the comments have been addressed, please let me know if still missing something 👍

src/app/Components/MasonryStatic.tsx
```diff
@@ -0,0 +1,49 @@
...
+ */
+export function MasonryStatic<T>({
+  data,
+  renderItem,
+  numColumns = 2,
+  columnSeparator,
+  columnKey,
+}: MasonryStaticProps<T>) {
+  const itemsColumns: T[][] = []
+  Array.from(Array(numColumns)).forEach(() => itemsColumns.push([]))
```
kulak91, line 24: Just curious. Wouldn't be better to join these lines? `const itemsColumns: T[][] = Array.from({ length: numColumns }, () => []);`

araujobarret: Thanks for the contribution! done in https://github.com/artsy/eigen/pull/8961/commits/73ba9ba5ad7e119349ee92266829871f511afa0a

gkartalis (approved):

PR #8965, opened 2023-07-07
fix(home): missing Home_heroUnits spread fragment

araujobarret:
### Description

No data changes only fixes the error from below which was caused because we never spread the fragment `Home_heroUnits`. 
Small renaming in the query to remove the alias, I don't know why I see a lot of aliasing to match UI from GraphQL, I strongly believe we should use what comes from MP, if the naming is bad there we should go on MP and change instead of aliasing. Alias in Relay should be used to avoid conflicting fields from different fragments/queries, for example 2 fragments relying on `date` but each one using a different format `param`.

Screenshot 2023-07-07 at 10 43 22

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- spread Home_heroUnits fragment in Home - araujobarret

</details>

### DIFF
--- a/src/app/Scenes/Home/Home.tsx
+++ b/src/app/Scenes/Home/Home.tsx
@@ -126,7 +126,7 @@ export interface HomeProps extends ViewProps {
   meBelow: Home_meBelow$data | null
   relay: RelayRefetchProp
   emergingPicks: Home_emergingPicks$data | null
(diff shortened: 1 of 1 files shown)


gkartalis (approved):
aaaaamazing! Thanks!

PR #8370, opened 2023-03-15
feat: integrate Sift SDK (PRO-87)

mdole:
This PR resolves [PRO-87]

### Description

Uses [Sift's RN wrapper](https://github.com/SiftScience/sift-react-native) to send "device information and app life cycle events" to Sift ([docs](https://sift.com/developers/docs/curl/javascript-api/overview)).

Noting that I've already updated the .env.shared that lives in S3, as well as CircleCI env, with the new variables.

This is unfortunately pretty much black-box; I can see what payloads show up in our Sift sandbox but there doesn't appear to be any other way to get at what's triggered when etc.

Event payloads look like:

```json
{
  "data": [
    {
      "user_id": "5744b8888b3b8145d7005304",
      "installation_id": "81F4B083-DC04-4047-8304-2969997400A0",
      "time": 1678895687765,
      "ios_app_state": {
        "network_addresses": [
          "fe80::aede:48ff:fe00:1122",
          "172.16.100.10",
          "fe80::e850:c6ff:fe8e:f01b",
          "fe80::e850:c6ff:fe8e:f01b",
          "fe80::296b:e36f:9541:34a2",
          "fe80::3c9e:888c:92a7:a2fc",
          "fe80::ce81:b1c:bd2c:69e",
          "10.8.0.94"
        ],
        "window_root_view_controller_titles": [
          "RNSScreen"
        ],
        "sdk_version": "v2.1.1"
      }
    },
    {
      "user_id": "5744b8888b3b8145d7005304",
      "installation_id": "81F4B083-DC04-4047-8304-2969997400A0",
      "time": 1678895807977,
      "ios_app_state": {
        "network_addresses": [
          "fe80::aede:48ff:fe00:1122",
          "172.16.100.10",
          "fe80::e850:c6ff:fe8e:f01b",
          "fe80::e850:c6ff:fe8e:f01b",
          "fe80::296b:e36f:9541:34a2",
          "fe80::3c9e:888c:92a7:a2fc",
          "fe80::ce81:b1c:bd2c:69e",
          "10.8.0.94"
        ],
        "sdk_version": "v2.1.1"
      }
    }
  ],
  "beacon_key": "ab6ecc6664"
}


```

### PR Checklist

- [x] I have tested my changes on **iOS** and **Android**.
- [x] I hid my changes behind a **[feature flag]**, or they don't need one.
- [x] I have included **screenshots** or **videos**, or I have not changed the UI.
- [x] I have added **tests**, or my changes don't require any.
[...]

PR #8991, opened 2023-07-13
feat: send additional Sift events on Android

mdole:
### Description

Followup to https://github.com/artsy/eigen/pull/8370.

Sift's React Native package doesn't correctly upload events without manual calls. This means we've had a few Sift events on Android (on app open + user login), but not as many as we'd like. The [recommendation](https://github.com/SiftScience/sift-react-native/pull/23#issuecomment-1630984250) from the folks who worked on it is to add these calls on navigate.

### PR Checklist

### To the reviewers 👀

#### Dev changes

- Adds additional Sift event calls on Android - mdole

</details>

### DIFF
--- a/patches/sift-react-native+0.1.8.patch
+++ b/patches/sift-react-native+0.1.8.patch
@@ -1,0 +1,12 @@
+diff --git a/node_modules/sift-react-native/lib/typescript/src/index.d.ts b/node_modules/sift-react-native/lib/typescript/src/index.d.ts
+index e6bcb83..6582ec8 100644
+--- a/node_modules/sift-react-native/lib/typescript/src/index.d.ts
++++ b/node_modules/sift-react-native/lib/typescript/src/index.d.ts
+@@ -1,6 +1,7 @@
+ declare type SiftReactNativeType = {
+     setSiftConfig(accountId: string, beaconKey: string, disallowCollectingLocationData: boolean, serverUrlFormat: string): void;
+     setUserId(userId: string): void;
++    setPageName(pageName: String): void;
+     unsetUserId(): void;
+     upload(): void;
+ };
--- a/src/app/store/config/features.ts
+++ b/src/app/store/config/features.ts
@@ -260,6 +260,12 @@ export const features: { [key: string]: FeatureDescriptor } = {
     readyForRelease: false,
     showInDevMenu: true,
   },
+  AREnableAdditionalSiftAndroidTracking: {
+    description: "Send additional events to Sift on Android",
+    readyForRelease: true,
+    showInDevMenu: true,
+    echoFlagKey: "AREnableAdditionalSiftAndroidTracking",
+  },
 }
 
 export interface DevToggleDescriptor {
--- a/src/app/system/navigation/ModalStack.tsx
+++ b/src/app/system/navigation/ModalStack.tsx
@@ -38,8 +38,6 @@ export const ModalStack: React.FC = ({ children }) => {
   useEffect(() => {
     if (trackSiftAndroid) {
       const initialRouteName = routeNameRef.current
-      // Expect error becuase their typing doesn't include setPageName, but it's there
-      // @ts-expect-error
       SiftReactNative.setPageName(`screen_${initialRouteName}`)
       SiftReactNative.upload()
     }
@@ -75,7 +73,6 @@ export const ModalStack: React.FC = ({ children }) => {
           })
 
           if (trackSiftAndroid) {
-            // @ts-expect-error
             SiftReactNative.setPageName(`screen_${currentRoute.name}`)
             SiftReactNative.upload()
           }
--- a/src/app/system/navigation/ModalStack.tsx
+++ b/src/app/system/navigation/ModalStack.tsx
@@ -8,8 +8,11 @@ import { Severity, addBreadcrumb } from "@sentry/react-native"
 import { AppModule, modules } from "app/AppRegistry"
 import { __unsafe_mainModalStackRef } from "app/NativeModules/ARScreenPresenterModule"
 import { GlobalStore } from "app/store/GlobalStore"
+import { useFeatureFlag } from "app/utils/hooks/useFeatureFlag"
 import { logNavigation } from "app/utils/loggers"
+import { useEffect, useRef } from "react"
 import { Platform } from "react-native"
+import SiftReactNative from "sift-react-native"
 import { NavStack } from "./NavStack"
 import { useReloadedDevNavigationState } from "./useReloadedDevNavigationState"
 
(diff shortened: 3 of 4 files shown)


mdole:
@MounirDhahri mind pulling this down and confirming you're still seeing events correctly fired? Would be greatly appreciated 🙂

src/app/system/navigation/ModalStack.tsx
```diff
@@ -24,6 +27,24 @@ const Stack = createStackNavigator()
...
+  // Code for Sift tracking; needs to be manually fired on Android
+  // See https://github.com/SiftScience/sift-react-native/pull/23#issuecomment-1630984250
+  const enableAdditionalSiftAndroidTracking = useFeatureFlag(
+    "AREnableAdditionalSiftAndroidTracking"
+  )
+  const trackSiftAndroid = Platform.OS === "android" && enableAdditionalSiftAndroidTracking
+  const routeNameRef = useRef()
+  useEffect(() => {
+    if (trackSiftAndroid) {
+      const initialRouteName = routeNameRef.current
```
MounirDhahri, line 40: **issue:** I believe you forgot to set the `routeNameRef` value within the `onReady` block inside the navigation container.

src/app/system/navigation/ModalStack.tsx
```diff
@@ -24,6 +27,24 @@ const Stack = createStackNavigator()
...
+      // Expect error becuase their typing doesn't include setPageName, but it's there
+      // @ts-expect-error
```
MounirDhahri, line 42: **nitpick:** I suggest patching the package to include this type or even better make a PR at the fork we're using to add it. I see it's already exported here https://github.com/SanoopC/sift-react-native/blob/02aff9c411e26f9303cdc2c6d41584e057754173/android/src/main/java/com/siftreactnative/SiftReactNativeModule.kt#L60

src/app/store/config/features.ts
```diff
@@ -260,6 +260,12 @@ export const features: { [key: string]: FeatureDescriptor } = {
     readyForRelease: false,
     showInDevMenu: true,
   },
+  AREnableAdditionalSiftAndroidTracking: {
```
MounirDhahri, line 263: **praise:** 🙌

MounirDhahri (commented):
Nice! I tested this again and it's working well. I left a minor comment then I believe we can merge this

https://github.com/artsy/eigen/assets/11945712/d54f9acb-a3ed-4ec9-9e61-614a8239780d

src/app/system/navigation/ModalStack.tsx
mdole: ah! thanks @MounirDhahri. updated in https://github.com/artsy/eigen/pull/8991/commits/105f3038a1625b7c1dd3cba6019d2f889af0afc3; that look right? 

Also a bonus question for you - in [their example](https://github.com/SanoopC/sift-react-native/blob/screen_navigation_example/example/root.js) they do a comparison of the last route name and the current route name and only send an event if the route name changed. I don't think that's super necessary (just might mean we're firing a little more often I suppose). Do you agree or do you think I should add that?

src/app/system/navigation/ModalStack.tsx
mdole: Good call, thanks for suggesting! Added a patch in https://github.com/artsy/eigen/pull/8991/commits/70fb134010a5070b63a6ee244b700c54a3e43dae; there's been an [issue open in the repo](https://github.com/SiftScience/sift-react-native/issues/6) for a while and it appears nobody is working on it 😬  but I'll add a comment there

patches/sift-react-native+0.1.8.patch
```diff
@@ -0,0 +1,12 @@
...
+--- a/node_modules/sift-react-native/lib/typescript/src/index.d.ts
++++ b/node_modules/sift-react-native/lib/typescript/src/index.d.ts
+@@ -1,6 +1,7 @@
+ declare type SiftReactNativeType = {
+     setSiftConfig(accountId: string, beaconKey: string, disallowCollectingLocationData: boolean, serverUrlFormat: string): void;
+     setUserId(userId: string): void;
++    setPageName(pageName: String): void;
+     unsetUserId(): void;
+     upload(): void;
+ };
```
gkartalis, line 12: Hey! Can you also add an entry to [Hacks.md](https://github.com/artsy/eigen/blob/main/HACKS.md) with why this is needed in order to keep track of the patched packages?

src/app/system/navigation/ModalStack.tsx
MounirDhahri: I agree as well, I don't think we should add it. I believe in their case, it might be that they're trying to restrict the events that are happening within the "same" screen and firing it only once for some reason. In our case, that would be like not sending events when the user is still in the Artwork screen, and that might be most of the user session. Our users will most likely keep jumping from one artwork to another suggested artwork for the duration of their session. That's why I find it useful to still track it.

https://github.com/artsy/eigen/assets/11945712/7541ac2f-d543-432a-8902-bed069454036

MounirDhahri: This will also allow us to follow what docs suggest when setting up Android [[here](https://github.com/SiftScience/sift-android#add-sift-to-your-application-flow)] by tracking unique page views.

src/app/system/navigation/ModalStack.tsx
MounirDhahri: awesome, thanks Matt

patches/sift-react-native+0.1.8.patch
mdole: Yessir! done in https://github.com/artsy/eigen/pull/8991/commits/014c20d780e52bc0d142b78f59b133fe0541cd52

src/app/system/navigation/ModalStack.tsx
mdole: Ah yeah good thought! That makes sense, thank you :)

src/app/system/navigation/ModalStack.tsx
```diff
@@ -24,11 +27,28 @@ const Stack = createStackNavigator()
...
+  )
+  const trackSiftAndroid = Platform.OS === "android" && enableAdditionalSiftAndroidTracking
+  const routeNameRef = useRef<string>()
+  useEffect(() => {
+    if (trackSiftAndroid) {
+      const initialRouteName = routeNameRef.current
+      SiftReactNative.setPageName(`screen_${initialRouteName}`)
+      SiftReactNative.upload()
+    }
+  }, [])
```
MounirDhahri, line 44: **issue(blocker):** I think you need to trigger this only after the navigation is ready, otherwise it will be empty initially in case the navigation is not yet ready.

MounirDhahri: Actually, even better, you might want to move this to be within `onReady` prop in the `NavigationContainer` to avoid watching the `onReady` global state

MounirDhahri (changes requested):
Sorry, I missed something before. I added it now

src/app/system/navigation/ModalStack.tsx
mdole: Good catch, thank you! Updated in https://github.com/artsy/eigen/pull/8991/commits/3d08ba54b8a9461623d1dd43c68421044699cf0b

MounirDhahri (approved):

PR #9056, opened 2023-07-25
chore: less strict error handling pt 1

MounirDhahri:
This PR resolves [PHIRE-130] [PHIRE-131]

### Description

Unable to load screens exploration.

So far:

- replaced the old middleware with a new one that throws **only** for principalField errors, really similar to force
- removed optionalField directive from homescreen
- adds section to track errors that are not coming from principal fields and opt in query system to register screens to the new one to avoid breaking other non defensive screens.
- ff to toggle between the two middlewares (⚠️ the middleware is marked as `unsafe` due to its place in the codebase. Even if we toggle the featureflag remotely the users would have to restart the app for the new value to take effect! ⚠️)

### PR Checklist

### To the reviewers 👀

- display less Unable to Load screens (behind a ff) - gkartalis

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[PHIRE-130]: https://artsyproduct.atlassian.net/browse/PHIRE-130?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PHIRE-131]: https://artsyproduct.atlassian.net/browse/PHIRE-131?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/Scenes/Home/Home.tsx
```diff
@@ -767,27 +766,28 @@ export const HomeQueryRenderer: React.FC<HomeQRProps> = ({ environment }) => {
...
-              @optionalField {
+            articlesConnection(first: 10, sort: PUBLISHED_AT_DESC, inEditorialFeed: true) {
               ...Home_articlesConnection
             }
           }
         `,
         variables: {
           version: worksForYouRecommendationsModel.payload || "B",
+          // do we need this? see also https://github.com/artsy/eigen/blob/45e9269079465d5fd65dd8262e27bbbaaba97cbb/src/app/system/relay/middlewares/errorMiddleware.ts#L158
+          useNewErrorMiddleware: true,
```
gkartalis, line 790: That's a nice idea for exploration, we can opt in with a variable different screens incrementally if we don't want to do this in once try to monitor the situation better

src/app/Scenes/Home/Home.tsx
```diff
@@ -767,27 +766,28 @@ export const HomeQueryRenderer: React.FC<HomeQRProps> = ({ environment }) => {
...
               ...Home_featured
             }
-            me @optionalField {
+            me {
               ...Home_meBelow
               ...RecommendedArtistsRail_me
             }
-            articlesConnection(first: 10, sort: PUBLISHED_AT_DESC, inEditorialFeed: true)
-              @optionalField {
+            articlesConnection(first: 10, sort: PUBLISHED_AT_DESC, inEditorialFeed: true) {
```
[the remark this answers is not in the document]
gkartalis: FYI: if you connect to metaphysics locally you need also to connect to staging vpn for diffusion, otherwise some requests aren't resolved and a weird laggy behavior is encountered in the app till the request times out

src/app/system/relay/middlewares/errorMiddleware.ts
```diff
@@ -80,51 +84,94 @@ const trackError = (
...
+  return res
+}
+
+export const errorMiddleware = () => (next: MiddlewareNextFn) => async (req: GraphQLRequest) => {
+  const res = await next(req)
+
+  const useNewErrorMiddlewareFeatureFlag = unsafe_getFeatureFlag("ARUseNewErrorMiddleware")
+
+  // Do we want to differenciate between screens with a flag in the query variables?
+  // const isScreenUsingNewErrorMiddleware = req.variables.useNewErrorMiddleware === true
```
gkartalis, line 167: questions from george

src/app/system/relay/middlewares/errorMiddleware.ts
```diff
@@ -80,51 +84,94 @@ const trackError = (
...
+
+  const enableNewErrorMiddleware = useNewErrorMiddlewareFeatureFlag
+  //  || isScreenUsingNewErrorMiddleware do we want to use this???
+
+  if (!!enableNewErrorMiddleware) {
+    return newErrorMiddleware(req, res)
+  }
```
damassi, line 174: This might be a naive question, but why do we need anything in this file? Why don't we just start from scratch and copy exactly what Force uses verbatim?

brainbicycle: discussed in slack but putting summary here for posterity: https://artsy.slack.com/archives/C02BAQ5K7/p1692115479088329
Goal is to have this file eventually just be the newMiddleWare and close to what force has, right now it is just big because we want to keep the old middleware around while transitioning in case we run into problems and need to turn off the new middleware.

src/app/Scenes/Home/Home.tsx
```diff
@@ -764,21 +763,20 @@ export const HomeQueryRenderer: React.FC<HomeQRProps> = ({ environment }) => {
...
-            emergingPicks: marketingCollection(slug: "curators-picks-emerging") @optionalField {
+            emergingPicks: marketingCollection(slug: "curators-picks-emerging") {
               ...Home_emergingPicks
             }
-            featured: viewingRooms(featured: true) @optionalField {
+            featured: viewingRooms(featured: true) {
               ...Home_featured
             }
-            me @optionalField {
+            me {
```
brainbicycle, line 775: a tricky bit which I hadn't really considered, if we remove optional field and the feature flag is disabled for the new middleware we are essentially making our error handling stricter. I wonder can we keep it and it will have no affect with the new middleware and remove later?

gkartalis: Great catch, reverted the home.tsx. changes and tested out with the `@optionalField` and we're okay to proceed 🎉

src/app/store/config/features.ts
```diff
@@ -280,6 +280,11 @@ export const features: { [key: string]: FeatureDescriptor } = {
     showInDevMenu: true,
     echoFlagKey: "AREnableFallbackToGeneratedAlertNames",
   },
+  ARUseNewErrorMiddleware: {
+    description: "Use updated error middleware",
+    readyForRelease: false,
+    showInDevMenu: true,
```
MrSltun, line 286: We need to update this to match Echo's PR artsy/echo#524

```suggestion
    description: "Use updated error middleware",
    readyForRelease: false,
    showInDevMenu: true,
    echoFlagKey: "ARUseNewErrorMiddleware",
```

gkartalis: https://artsyproduct.atlassian.net/browse/PHIRE-226

Its gonna be in a different pR 🎉

MounirDhahri:
Thanks for pushing this through @gkartalis 👏👏👏
I lost track of the conversation here after getting sick and going on holidays, do you have a few minutes to catch up about the upcoming changes?

gkartalis:
Sure thing! Feel free to drop some time on my calendar!

MounirDhahri:
Great, done ✅

PR #9164, opened 2023-08-16
chore(ONYX-13, ONYX-12): remove old saves implementation

dariakoko:
This PR resolves [ONYX-13] [ONYX-12]

### Description

Remove old saves implementation

### PR Checklist

### To the reviewers 👀

- remove old saves implementation -daria

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[ONYX-13]: https://artsyproduct.atlassian.net/browse/ONYX-13?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ONYX-12]: https://artsyproduct.atlassian.net/browse/ONYX-12?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ArtsyOpenSource:

- Cross-platform user-facing changes (remove old saves implementation -daria)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 0f8060448698826a5b235301811f084dac70e929
</p>

src/app/Components/ArtworkLists/useSaveArtworkToArtworkLists.ts
```diff
@@ -90,17 +76,12 @@ export const useSaveArtworkToArtworkLists = (options: Options) => {
   }
 
   const saveArtworkToLists = () => {
-    if (!isArtworkListsEnabled) {
-      legacySaveArtworkToDefaultArtworkList()
-      return
-    }
```
dariakoko, line 96: Is it the right move to remove this condition?

src/app/Components/ArtworkLists/useSaveArtworkToArtworkLists.ts
```diff
@@ -32,27 +28,17 @@ export const useSaveArtworkToArtworkLists = (options: Options) => {
   }
   let isSaved = artwork.isSaved
 
-  if (isArtworkListsEnabled) {
```
dariakoko, line 35: And this one too

src/app/Components/ArtworkLists/useSaveArtworkToArtworkLists.ts
nickskalkin: `isArtworkListsEnabled` takes into account `options.saveToDefaultCollectionOnly` (take a look at how the variable is defined [here](https://github.com/artsy/eigen/blob/main/src/app/Components/ArtworkLists/useSaveArtworkToArtworkLists.ts#L18)). We pass `saveToDefaultCollectionOnly` as `true` when we want to user to experience "old" behaviour, when an artwork is being saved only to default collection. You can trigger this behaviour when clicking on "Watch lot", for instance. We still want to have this "old" behaviour, that's why `saveToDefaultCollectionOnly` should be taken into account here.

The easiest change I can imagine is simply to define `isArtworkListsEnabled` like this:

```ts
const isArtworkListsEnabled = !options.saveToDefaultCollectionOnly
```

But personally I would change all places which check `isArtworkListsEnabled` to `!options.saveToDefaultCollectionOnly`

dariakoko: Yehhh, that's what I thought initially. But the saveArtworkToDefaultArtworkList() opens the modal with the collections anyways. The only difference is that when we like the artwork the default list is already selected, but when we like the lot none of the lists is selected. I thought we don't need to show the modal in case we're saving an auction artwork. Or should we?

src/app/Scenes/Artwork/Components/ArtworkActions.tests.tsx
```diff
@@ -137,9 +135,7 @@ describe("ArtworkActions", () => {
   })
 
   describe("Save button", () => {
-    // TODO: remove skip when lists feature flag is readyForRelease: true
-    describe.skip("whith lists feature flag enabled", () => {
-      __globalStoreTestUtils__?.injectFeatureFlags({ AREnableArtworksLists: true })
+    describe("whith lists feature flag enabled", () => {
```
anandaroop, line 138: **question (non-blocking):** Wondering if we even need to wrap the test with this `describe` block?

anandaroop (approved):
Looking good!

src/app/Components/ArtworkLists/useSaveArtworkToArtworkLists.ts
```diff
@@ -90,17 +78,16 @@ export const useSaveArtworkToArtworkLists = (options: Options) => {
   }
 
   const saveArtworkToLists = () => {
-    if (!isArtworkListsEnabled) {
-      legacySaveArtworkToDefaultArtworkList()
-      return
+    if (options.saveToDefaultCollectionOnly) {
+      saveArtworkToDefaultArtworkList()
```
nickskalkin, line 82: *blocking*: this code has two issue:

- there should be a return statement as well, otherwise the code below will be executed as well
- saveArtworkToDefaultArtworkList always opens the drawer with artwork lists. When `saveToDefaultCollectionOnly` options is passed (when clicking on Watch lot button, for instance), the drawer shouldn't be opened - an artwork just gets added to a default list silently.

This can be achieved by adding the code below to the optimisticUpdater in saveArtworkToDefaultArtworkList:

```ts
      if (options.saveToDefaultCollectionOnly) {
        return
      }
```

nickskalkin (changes requested):

src/app/Components/ArtworkLists/useSaveArtworkToArtworkLists.ts
dariakoko: Right, thank you @nickskalkin

src/app/utils/mutations/useSaveArtwork.ts
```diff
@@ -47,6 +48,7 @@ export const useSaveArtwork = ({
       onCompleted: () => {
         clearPrevCommit()
         onCompleted?.(nextSavedState)
+        refreshOnArtworkSave()
```
dariakoko, line 51: I reused `refreshOnArtworkSave()` method we used for the old saves implementation. This method refreshes Saved tab. If we keep Saved tab open, go to Home screen and save the artwork, when we are back to the Saved tab we see the updated lists. Without this method we would need to refresh the Saved tab ourselves.
The only thing is - if we keep the Artwork **list** open the method doesn't update it as it updates only the Saves tab. Would be nice to be able to refresh individual lists as well imo

https://github.com/artsy/eigen/assets/36167539/8c942751-35c6-4970-a63f-a930e9172ed1

anandaroop (dismissed):
LGTM 👍🏽 

(though I still think [this](https://github.com/artsy/eigen/pull/9164/files?file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx&show-viewed-files=true&show-deleted-files=false#r1296278208) could be cleaned up)

dariakoko:
Right, @anandaroop, sorry, missed this. Cleaned up now

PR #9168, opened 2023-08-21
fix: Update saved alerts list screen after creating alert

olerichter00:
### Description

Update saved alerts list screen after creating alert.

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Scenes/SavedSearchAlert/SavedSearchAlertForm.tsx
+++ b/src/app/Scenes/SavedSearchAlert/SavedSearchAlertForm.tsx
@@ -3,6 +3,7 @@ import { Dialog, quoteLeft, quoteRight, useTheme } from "@artsy/palette-mobile"
 import { NavigationProp, useNavigation } from "@react-navigation/native"
 import { SearchCriteriaAttributes } from "app/Components/ArtworkFilter/SavedSearch/types"
 import { goBack, navigate } from "app/system/navigation/navigate"
+import { refreshSavedAlerts } from "app/utils/refreshHelpers"
 import { FormikProvider, useFormik } from "formik"
 import React, { useEffect, useState } from "react"
 import { Alert, ScrollView, StyleProp, ViewStyle } from "react-native"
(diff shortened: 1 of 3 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 6350142bd1ae7efb9559a62d4604d2221b7b8d1d
</p>

src/app/Scenes/SavedSearchAlert/SavedSearchAlertForm.tsx
```diff
@@ -162,6 +163,7 @@ export const SavedSearchAlertForm: React.FC<SavedSearchAlertFormProps> = (props)
       }
 
       navigation.navigate("ConfirmationScreen", { searchCriteriaID: result.id })
+      refreshSavedAlerts()
```
olerichter00, line 166: Ideally, the mutation should return `me`, which would theoretically allow us to refetch the entire saved alerts connection. 

However, we haven't actually implemented something like this in Eigen yet. It would be great to explore how reliably updating connections in the Relay store with mutations works. 

The pattern I use here is already used for My Collection and Saved artworks.

nickskalkin: Interesting, I didn't know about this approach!

This may be a stupid question, since I'm not that familiar with relay internals, but in Force sometimes we update relay cache manually upon a mutation success. Here is [an example](https://github.com/artsy/force/blob/main/src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/Mutations/useCreateCollection.ts#L9). Is this something we could / should do here instead?

olerichter00: This could be another approach to explore. Looks quite complex but probably better than these event handlers 🤔

nickskalkin (approved):
👍

PR #9221, opened 2023-09-06
fix: make alert name multiline on android

MounirDhahri:
This PR resolves [] 

### Description

This CI makes the alert name input span across multiple lines in Android. This is required to make sure that text doesn't cut because of the input height. Ideally, we would ellipsize the text but this is not yet supported in RN on Android. Another solution that I tried to work on is to make to use the placeholders array that we are also using inside the search tab, but this turned out to be not possible because it assumes we know the possible placeholder strings (which is obviously not the case here)

Screenshot 2023-09-06 at 12 33 57

### PR Checklist

### To the reviewers 👀

- make alert name multiline on android - mounir

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Scenes/SavedSearchAlert/Components/SavedSearchNameInput.tsx
+++ b/src/app/Scenes/SavedSearchAlert/Components/SavedSearchNameInput.tsx
@@ -6,6 +6,7 @@ import { getRelayEnvironment } from "app/system/relay/defaultEnvironment"
 import { useFormikContext } from "formik"
 import { omit } from "lodash"
 import { useEffect, useState } from "react"
+import { Platform } from "react-native"
 import { QueryRenderer } from "react-relay"
 import { graphql } from "relay-runtime"
 interface SavedSearchNameInputProps {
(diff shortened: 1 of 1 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (make alert name multiline on android - mounir)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 484ca6ebc0614a20972437bbe0288c9383d7420f
</p>

dariakoko (approved):
Thank you for the fix!

I've just noticed that we use Input from palette-mobile only for this screen. Not sure why it is like this. I'll cherrypick this one to the release anyways. Will reice the question in practice mobile

dblandin:
@MounirDhahri Sorry for the late ping on this but could we use [the `ellipsizeMode` Text property](https://reactnative.dev/docs/text#ellipsizemode) here?

> On Android, when numberOfLines is set to a value higher than 1, only tail value will work correctly.

MounirDhahri:
I tried that on my phone actually but it didn't work unfortunately. It only worked on iOS :( there are also threads in react-native where people complain it's not working

PR #9239, opened 2023-09-11
feat: add create alert at end of artworks list

MounirDhahri:
This PR resolves [ONYX-295] 

### Description
**[This is already reviewable]**
**[Erin still needs to update the copy before merging this PR]**

This PR adds support to creating an alert at the end of the list of artworks in Artist page.
| Tracking event                                                                                                                                    | Create alert Message                                                                                                                              |
|---------------------------------------------------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------|
| Simulator Screenshot - iPhone 14 - 2023-09-11 at 11 54 05 | Simulator Screenshot - iPhone 14 - 2023-09-11 at 11 50 59 |

### PR Checklist

### To the reviewers 👀

- add create alert at the end of artworks list grid - mounir

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[ONYX-295]: https://artsyproduct.atlassian.net/browse/ONYX-295?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
+++ b/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
@@ -185,7 +185,7 @@ const ArtworksGrid: React.FC<ArtworksGridProps> = ({
     if (showCreateAlertAtEndOfList && !relay.hasMore()) {
       return (
         <Message
-          title="Get notified when new works are added."
+          title="Get notified when works you're looking for are added."
           containerStyle={{ width: Dimensions.get("window").width, left: -space(2), mt: 2 }}
           IconComponent={() => {
             return <CreateAlertButton />
--- a/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
+++ b/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
@@ -10,6 +10,7 @@ import {
   Spacer,
   BellIcon,
   Spinner,
+  Message,
 } from "@artsy/palette-mobile"
 import { ArtistArtworks_artist$data } from "__generated__/ArtistArtworks_artist.graphql"
 import { ArtistArtworksFilterHeader } from "app/Components/Artist/ArtistArtworks/ArtistArtworksFilterHeader"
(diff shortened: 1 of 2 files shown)


src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
```diff
@@ -162,6 +173,31 @@ const ArtworksGrid: React.FC<ArtworksGridProps> = ({
...
+          <Spinner />
+        </Flex>
+      )
+    }
+
+    if (showCreateAlertAtEndOfList && !relay.hasMore()) {
+      return (
+        <Message
+          title="Get notified when new works are added."
+          containerStyle={{ width: Dimensions.get("window").width, left: -space(2) }}
```
MounirDhahri, line 189: 1. Going full screen here because there is not enough space for the copy and the button otherwise
Simulator Screenshot - iPhone 14 - 2023-09-11 at 11 47 24

MounirDhahri: Although would have been great if this was already full width, but the width is coming from the parent component and we would need to change the padding in all componennts if we would want to avoid a negative value on left. My judgement, this is not needed now

dariakoko: Sounds good! I think this is not the only place where we have a negative value for the message component. Are there any strong cons to using negative values in styles, or is this just not a good practice?

dariakoko:
Looks like the spacing is way too small. Could you, please, adjust it?
image

src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
MounirDhahri: negative values in styles are a lot of the times references to mistakes designing (architecturally )components. Inner components should be ideally included in parent components borders. This makes it easier to guess how a component will look like when used, quicker to debug, and quicker to adjust styles without any spacings gymnastics 😄

nickskalkin (approved):
👍

MounirDhahri:
Updated the copy to what Erin asked, thanks all for reviewing. will add #squashongreen

PR #9248, opened 2023-09-12
fix(Artist): useCallback to avoid extra re-renders

araujobarret:
### Description

That's a temporary fix, the real fix should be in palette-mobile to use the `BelowTitleHeaderComponent` as a ReactNode pattern.

Props to @MounirDhahri to pointing the quick fix super fast 🚀 

### PR Checklist

### To the reviewers 👀

- 

#### iOS user-facing changes

-

-

#### Dev changes

- removes the extra re-render when following artist - araujobarret

</details>

### DIFF
--- a/src/app/Components/Artist/ArtistHeader.tsx
+++ b/src/app/Components/Artist/ArtistHeader.tsx
@@ -12,10 +12,10 @@ const ARTIST_HEADER_SCROLL_MARGIN = 100
 interface Props {
   artist: ArtistHeader_artist$data
   relay: RelayProp
-  onLayout?: ViewProps["onLayout"]
+  onLayoutChange?: ViewProps["onLayout"]
 }
 
-export const ArtistHeader: React.FC<Props> = ({ artist, onLayout }) => {
+export const ArtistHeader: React.FC<Props> = ({ artist, onLayoutChange }) => {
   const { width } = useScreenDimensions()
   const { updateScrollYOffset } = useScreenScrollContext()
   const isTablet = isPad()
@@ -46,7 +46,7 @@ export const ArtistHeader: React.FC<Props> = ({ artist, onLayout }) => {
   const handleOnLayout = ({ nativeEvent, ...rest }: LayoutChangeEvent) => {
     if (nativeEvent.layout.height > 0) {
       updateScrollYOffset(nativeEvent.layout.height - ARTIST_HEADER_SCROLL_MARGIN)
-      onLayout?.({ nativeEvent, ...rest })
+      onLayoutChange?.({ nativeEvent, ...rest })
     }
   }
 
--- a/src/app/Components/Artist/ArtistHeader.tsx
+++ b/src/app/Components/Artist/ArtistHeader.tsx
@@ -12,10 +12,10 @@ const ARTIST_HEADER_SCROLL_MARGIN = 100
 interface Props {
   artist: ArtistHeader_artist$data
   relay: RelayProp
-  onLayout?: ViewProps["onLayout"]
+  handleLayoutChange?: ViewProps["onLayout"]
 }
 
-export const ArtistHeader: React.FC<Props> = ({ artist, onLayout }) => {
+export const ArtistHeader: React.FC<Props> = ({ artist, handleLayoutChange }) => {
   const { width } = useScreenDimensions()
   const { updateScrollYOffset } = useScreenScrollContext()
   const isTablet = isPad()
@@ -46,7 +46,7 @@ export const ArtistHeader: React.FC<Props> = ({ artist, onLayout }) => {
   const handleOnLayout = ({ nativeEvent, ...rest }: LayoutChangeEvent) => {
     if (nativeEvent.layout.height > 0) {
       updateScrollYOffset(nativeEvent.layout.height - ARTIST_HEADER_SCROLL_MARGIN)
-      onLayout?.({ nativeEvent, ...rest })
+      handleLayoutChange?.({ nativeEvent, ...rest })
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/app/Scenes/Artist/Artist.tsx
```diff
@@ -119,15 +119,19 @@ export const Artist: React.FC<ArtistProps> = (props) => {
...
+              <ArtistHeaderFragmentContainer
+                artist={artistAboveTheFold!}
+                onLayout={({ nativeEvent }) => {
+                  if (headerHeight !== nativeEvent.layout.height) {
+                    setHeaderHeight(nativeEvent.layout.height)
+                  }
+                }}
+              />
+            ),
+            []
```
MounirDhahri, line 134: **thought:** do you intentionally not want to have `artistAboveTheFold` as part of the dependencies array 👀. Nothing there seems to be updating so it's fine maybe for now https://www.figma.com/file/lVbIxbnH2IUk1K1FoxyFHV/Filters-and-Alerts-redesign?type=design&node-id=250-35496&mode=design&t=Fcs9b5FYCufRIERE-0

ArtsyOpenSource:

- Dev changes (removes the extra re-render when following artist - araujobarret)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 8d5662303e438c3d7823b01f9fa024f5087eeda0
</p>

MounirDhahri (dismissed):
Thanks for the PR 🙏

src/app/Scenes/Artist/Artist.tsx
araujobarret: yes because it's only a bandaid for now, plus `artistAboveTheFold` it's an object, if if doesn't change any data from time to time can re-calculate the function signature again, but not sure if in this case would still cause the re-render.
When I'm back from ooo I can address this on palette-mobile.

src/app/Scenes/Artist/Artist.tsx
```diff
@@ -119,15 +119,19 @@ export const Artist: React.FC<ArtistProps> = (props) => {
...
-            <ArtistHeaderFragmentContainer
-              artist={artistAboveTheFold!}
-              onLayout={({ nativeEvent }) => {
-                if (headerHeight !== nativeEvent.layout.height) {
-                  setHeaderHeight(nativeEvent.layout.height)
-                }
-              }}
-            />
+          // TODO: fix how we pass the element in palette-mobile, probably needs to be ReactNode type/pattern
+          BelowTitleHeaderComponent={useCallback(
```
damassi, line 123: This screams rule of hooks error 😄 

Can this be moved up and into a variable? If ever the slightest logic is added above this component, it will error out.

damassi (commented):
It's pretty easy to work between palette mobile and eigen with the watcher. Any reason we're going the bandaid route vs fixing things in palette mobile? 

(This solution seems fine btw, just wondering)

araujobarret:
Sorry for the quick fix, I was leaving to catch my plane, had not much time to check all deps nor fix in Palette.

src/app/Scenes/Artist/Artist.tsx
MounirDhahri: sorry missed this during review 😢

gkartalis (commented):
@MounirDhahri lets merge this before the release? And when Carlos is back can follow up with the rest 🤔

MounirDhahri:
sure, I will address the comment and we can merge this

MounirDhahri:
I am also adding the fix the palette-mobile as well to make sure we don't suffer from this in the future

MounirDhahri:
No need to do anything here, I will make the same fix to palette mobile. I will close this

MounirDhahri:
okay, I tried to understand the code again and see why we are using `headerHeight` which is the reason why we have the re-render and it's actually not being used anywhere. I will remove it and then merge this. 

The fix on Palette-mobile isn't required anymore.

MounirDhahri:
So the update is not happening because of the component in palette but instead it's happening because from within the component, we update the state of the parent using `onChange`. I believe the the current fix is enough and we can already merge this later today if no one else minds.

src/app/Scenes/Artist/Artist.tsx
```diff
@@ -95,6 +95,20 @@ export const Artist: React.FC<ArtistProps> = (props) => {
     })
   }
 
+  const renderBelowTheHeaderComponent = useCallback(
+    () => (
+      <ArtistHeaderFragmentContainer
+        artist={artistAboveTheFold!}
+        handleLayoutChange={({ nativeEvent }) => {
```
MounirDhahri, line 102: renamed this to `handleLayoutChange` because `onLayout` confused me

src/app/Components/Artist/ArtistHeader.tsx
```diff
@@ -12,10 +12,10 @@ const ARTIST_HEADER_SCROLL_MARGIN = 100
 interface Props {
   artist: ArtistHeader_artist$data
   relay: RelayProp
-  onLayout?: ViewProps["onLayout"]
+  handleLayoutChange?: ViewProps["onLayout"]
```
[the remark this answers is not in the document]
MounirDhahri: The reason behind my naming is to avoid clashes on the component itself and I got too used to a specific pattern for naming methods. I mainly follow this pattern that I've seen recommended in other areas and it's been fine so far
```typescript
const Foo = ({ onClick }) => {
  const handleClick = (event) => {
    doSomethingElseHere();
    onClick(event);
  }
  
  return (
    <button onClick={handleClick}>Bar</button>
  )
}
```
Naming is subjective always anyway 😄 I will revert it.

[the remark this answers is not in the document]
damassi: Note that props passed in as callback props are idiomatically `onThingChanged` and functions internal to components are `handleThingChanged`. That gives us information about the direction of the data flow. So: `onLayoutChange`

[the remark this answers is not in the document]
MounirDhahri: makes sense. I will update 🙏

PR #9305, opened 2023-09-21
fix(DIA-168): onboarding quiz crash

MounirDhahri:
This PR resolves [DIA-168] 

### Description
This PR resolves a race condition that resolved in a crash if you move through the art quiz fast. The solution here is to decouple the swiper from the handler in order to avoid relying on a state value for the active index that might be outdated. I also moved the API logic to happen before the state logic. 

**Note:** I believe the logic can be improved even further here buy having the `activeIndex` getting incremented only after a successful mutation and creating a queue of mutations. It's alright though as it is right now and it should be good enough in its current state.

**Before** 

https://github.com/artsy/eigen/assets/11945712/35e360e5-cb68-4f64-a4c5-879141059c30

**After**

https://github.com/artsy/eigen/assets/11945712/6178ad9b-eb6b-4f32-83af-3316a4368b58

### PR Checklist

### To the reviewers 👀

- fix onboarding quiz crash - mounir

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[DIA-168]: https://artsyproduct.atlassian.net/browse/DIA-168?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/ArtQuiz/ArtQuizArtworks.tsx
+++ b/src/app/Scenes/ArtQuiz/ArtQuizArtworks.tsx
@@ -46,17 +46,17 @@ const ArtQuizArtworksScreen = () => {
 
   const handleSwipe = (swipeDirection: "left" | "right") => {
     handleNext(swipeDirection === "right" ? "Like" : "Dislike", activeCardIndex)
-    setActiveCardIndex(activeCardIndex + 1)
+
+    // No need to swipe through the last card, just navigate to the results screen
+    if (activeCardIndex + 1 < artworks.length) {
+      setActiveCardIndex(activeCardIndex + 1)
+    }
   }
 
   const handleNext = (action: "Like" | "Dislike", activeIndex: number) => {
     popoverMessage.hide()
 
     const currentArtwork = artworks[activeIndex]
-
-    if (!currentArtwork) {
-      return
-    }
 
     if (action === "Like") {
       submitSave({
--- a/src/app/Scenes/ArtQuiz/ArtQuizArtworks.tsx
+++ b/src/app/Scenes/ArtQuiz/ArtQuizArtworks.tsx
@@ -10,6 +10,7 @@ import { ArtQuizLoader } from "app/Scenes/ArtQuiz/ArtQuizLoader"
 import { GlobalStore } from "app/store/GlobalStore"
 import { goBack, navigate } from "app/system/navigation/navigate"
 import { extractNodes } from "app/utils/extractNodes"
+import { isEmpty } from "lodash"
 import { Suspense, useEffect, useState } from "react"
 import { Image } from "react-native"
 
@@ -57,6 +58,10 @@ const ArtQuizArtworksScreen = () => {
     popoverMessage.hide()
 
     const currentArtwork = artworks[activeIndex]
+
+    if (isEmpty(currentArtwork)) {
+      return
+    }
 
     if (action === "Like") {
       submitSave({
--- a/src/app/Scenes/ArtQuiz/ArtQuizArtworks.tsx
+++ b/src/app/Scenes/ArtQuiz/ArtQuizArtworks.tsx
@@ -21,6 +21,7 @@ const ArtQuizArtworksScreen = () => {
   const { width } = useScreenDimensions()
   const space = useSpace()
   const artworks = extractNodes(queryResult.me?.quiz.quizArtworkConnection)
+
   const lastInteractedArtworkIndex = queryResult.me?.quiz.quizArtworkConnection?.edges?.findIndex(
     (edge) => edge?.interactedAt === null
   )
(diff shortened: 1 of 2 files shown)


brainbicycle (dismissed):
nice 🦾 thank you!

MounirDhahri:
@brainbicycle I will do a small exploration about this for few minutes. Thanks for the quick review 🙏

ArtsyOpenSource:

- Cross-platform user-facing changes (fix onboarding quiz crash - mounir)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 3f41858d6211e3ed04f5bd624f607b71d2e86708
</p>

MounirDhahri:
I feel like there is something wrong with the logic although it looks straightforward

src/app/Scenes/ArtQuiz/ArtQuizArtworks.tsx
```diff
@@ -43,16 +44,20 @@ const ArtQuizArtworksScreen = () => {
...
   }
 
   const handleNext = (action: "Like" | "Dislike", activeIndex: number) => {
     popoverMessage.hide()
 
     const currentArtwork = artworks[activeIndex]
 
+    if (!currentArtwork) {
+      return
+    }
```
gkartalis, line 59: **question(non-blocking):** How would you feel about changing this to use isEmpty from lodash?
```suggestion
    if (isEmpty(currentArtwork)) {
      return
    }
```

Reason I'm suggesting this is that sometimes we might have an empty object as currentArtwork and `!{} = false` while `isEmpty(currentArtwork) = true`

src/app/Scenes/ArtQuiz/ArtQuizArtworks.tsx
```diff
@@ -43,16 +44,20 @@ const ArtQuizArtworksScreen = () => {
     }
   }, [])
 
-  const handleSwipe = (swipeDirection: "left" | "right", activeIndex: number) => {
-    setActiveCardIndex(activeIndex + 1)
-    handleNext(swipeDirection === "right" ? "Like" : "Dislike", activeIndex)
+  const handleSwipe = (swipeDirection: "left" | "right") => {
+    handleNext(swipeDirection === "right" ? "Like" : "Dislike", activeCardIndex)
+    setActiveCardIndex(activeCardIndex + 1)
```
gkartalis, line 49: 👏👏👏

gkartalis (dismissed):
Nice and simplified! Agree that it can be improved further (maybe we could create a ticket in Diamond board for this to not get forgoten?)

✨✨✨✨

gkartalis:
oh sorry, just saw the wip, feel free to ignore and will review again when its ready!

src/app/Scenes/ArtQuiz/ArtQuizArtworks.tsx
MounirDhahri: I can do that, it's probably not needed in this case though since I believe we do that in MP. I think the failure comes from the fact we are getting an invalid index. I am pushing a comment and a better fix now

MounirDhahri:
Just pushed a better fix now - please feel free to review

src/app/Scenes/ArtQuiz/ArtQuizArtworks.tsx
```diff
@@ -43,9 +44,13 @@ const ArtQuizArtworksScreen = () => {
...
-  const handleSwipe = (swipeDirection: "left" | "right", activeIndex: number) => {
-    setActiveCardIndex(activeIndex + 1)
-    handleNext(swipeDirection === "right" ? "Like" : "Dislike", activeIndex)
+  const handleSwipe = (swipeDirection: "left" | "right") => {
+    handleNext(swipeDirection === "right" ? "Like" : "Dislike", activeCardIndex)
+
+    // No need to swipe through the last card, just navigate to the results screen
+    if (activeCardIndex + 1 < artworks.length) {
+      setActiveCardIndex(activeCardIndex + 1)
+    }
```
MounirDhahri, line 53: @brainbicycle @gkartalis the crash came from the fact that at the last index, we still do increment the `activeCardIndex` which leads to using a card index that's not in the array

gkartalis (dismissed):

src/app/Scenes/ArtQuiz/ArtQuizArtworks.tsx
brainbicycle: nice catch! I would be in favor of still leaving in the check if currentArtwork is undefined, I think we should get in the habit of being defensive about this stuff, if we introduce some other logic changing the index the problem might come back otherwise

MounirDhahri: good idea. I will do it 🙌

PR #9340, opened 2023-09-27
fix(ProgressiveOnboarding): adjust the logic to display the first Popover of saved Artwork

araujobarret:
### Description

Fix a weird behavior once you enter a screen with an ArtworkGridItem at the bottom and try to show the Popover even when the respective element anchoring it is not on the screen yet. On Android that was breaking and becoming unresponsive.

Added back to Android since got back to work with the fix(a04e87ab9fdaa3980f5e43dc95dca532d37fc31c)

iOS

https://github.com/artsy/eigen/assets/15792853/e0832062-3062-4376-b695-888c6197a9aa

Android

https://github.com/artsy/eigen/assets/15792853/52f71a3f-cde2-4ff7-842c-024ecc782fa3

### PR Checklist

### To the reviewers 👀

- fix the first progressive onboarding alert on the Artwork screen - araujobarret

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Components/ProgressiveOnboarding/ProgressiveOnboardingSaveArtwork.tsx
+++ b/src/app/Components/ProgressiveOnboarding/ProgressiveOnboardingSaveArtwork.tsx
@@ -16,36 +16,39 @@ export const ProgressiveOnboardingSaveArtwork: React.FC = ({ children }) => {
 
   const savedArtworks = data?.me.counts.savedArtworks
   const isDismissed = _isDismissed("save-artwork").status
-  const isDisplayable = !isDismissed && savedArtworks === 0
+  const isDisplayable = !isDismissed && savedArtworks === 0 && isInView
 
   const handleDismiss = () => {
     setIsVisible(false)
     dismiss("save-artwork")
   }
 
-  if (!isDisplayable) {
-    if (isInView) {
-      return <>{children}</>
-    }
-
-    return <ElementInView onVisible={() => setIsInView(true)}>{children}</ElementInView>
+  // all conditions met we show the popover
+  if (isDisplayable) {
+    return (
+      <Popover
+        visible={isVisible}
+        onDismiss={handleDismiss}
+        onPressOutside={handleDismiss}
+        title={
+          <Text weight="medium" color="white100">
+            Like what you see?
+          </Text>
+        }
+        content={<Text color="white100">Hit the heart to save an artwork.</Text>}
+      >
+        <Flex>{children}</Flex>
+      </Popover>
+    )
   }
 
-  return (
-    <Popover
-      visible={isVisible}
-      onDismiss={handleDismiss}
-      onPressOutside={handleDismiss}
-      title={
-        <Text weight="medium" color="white100">
-          Like what you see?
-        </Text>
-      }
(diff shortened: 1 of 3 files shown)


src/app/Components/ProgressiveOnboarding/ProgressiveOnboardingSaveArtwork.tsx
```diff
@@ -16,36 +16,39 @@ export const ProgressiveOnboardingSaveArtwork: React.FC = ({ children }) => {
...
     setIsVisible(false)
     dismiss("save-artwork")
   }
 
-  if (!isDisplayable) {
-    if (isInView) {
-      return <>{children}</>
-    }
+  // all conditions met we show the popover
+  if (isDisplayable) {
```
araujobarret, line 27: no real changes in the if/else, just organized it better to avoid nesting and improve readability

src/app/Components/ProgressiveOnboarding/ProgressiveOnboardingSaveArtwork.tsx
```diff
@@ -16,36 +16,39 @@ export const ProgressiveOnboardingSaveArtwork: React.FC = ({ children }) => {
 
   const savedArtworks = data?.me.counts.savedArtworks
   const isDismissed = _isDismissed("save-artwork").status
-  const isDisplayable = !isDismissed && savedArtworks === 0
+  const isDisplayable = !isDismissed && savedArtworks === 0 && isInView
```
araujobarret, line 19: that's the fix, we forgot to consider if the element appeared in the view to show the Popover, that was fixed before but somehow got back again

ArtsyOpenSource:

- Cross-platform user-facing changes (fix the first progressive onboarding alert on the Artwork screen - araujobarret)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against a04e87ab9fdaa3980f5e43dc95dca532d37fc31c
</p>

MrSltun (approved):
Looks great! Thank you for the fix Carlos!! 🌟

MounirDhahri (approved):
Nice fix! I think that solves it perfectly. I do believe though that at some point, we would need to be careful with the popovers and potentially queue them in case we would be adding more popovers that will be showing at the same time

araujobarret:
>Nice fix! I think that solves it perfectly. I do believe though that at some point, we would need to be careful with the popovers and potentially queue them in case we would be adding more popovers that will be showing at the same time

@MounirDhahri yes, I can't agree more; the main issue is when the content is scrollable, it's hard to calculate where to position the element exactly, also can't have double modals rendered at the same time. The API of Popovers needs to be 
more sturdy, maybe some scrolling to the popover once it's seen to avoid it being rendered but half visible to the screen. 

Unfortunately, this is the best solution I've found so far for Popovers in RN ecosystem, hope the future of anchoring elements to a View gets better.

MounirDhahri:
rn popover docs state that they support a js mode for showing modals, we can use that in the future. For now, what we have is largely sufficient. Thanks again for fixing this 🙏

PR #9387, opened 2023-10-05
feat: create alerts filters screen

MounirDhahri:
This PR resolves [ONYX-393] [ONYX-376] [ONYX-373] 

### Description

This PR adds the foundational work required to add future alerts to the new alert screen. It consists of the following
- [x] New filters store, based on the learning from the previous versions that we had of the store
- [x] Your alerts section
- [x] New filter screen
- [x] An example of rarity filtering

https://github.com/artsy/eigen/assets/11945712/9296bda9-e1ee-4023-8b3f-0ae70cfd4596

## Note for reviewers
- If you would like to see the progression of this PR, you **can** review it commit by commit. I would suggest though refraining from commenting at early commits because I kept on refactoring the store as I progressed with the work
- Was it possible to divide this into more PRs? I wish I knew how, but until I built the first filter and displayed the right data, the store shape was ambiguous to me although I knew where I wanted to get eventually.

### PR Checklist

### To the reviewers 👀

- add filter screen to create alert modal - mounir
- add your filters section to the new filter screen - mounir
- add rarity section to the new filter screen - mounir
- add clear all button to the new filter screen

#### iOS user-facing changes

-

-

#### Dev changes

- add docs about how to add a fitler to the new filters screen - mounir

</details>

[ONYX-393]: https://artsyproduct.atlassian.net/browse/ONYX-393?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ONYX-376]: https://artsyproduct.atlassian.net/browse/ONYX-376?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ONYX-373]: https://artsyproduct.atlassian.net/browse/ONYX-373?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against f74f626b5ad9ee06398b1624cccedd0e2c4a0097
</p>

src/app/Components/NewArtworkFilter/NewArtworkFilterAppliedFilters.tsx
```diff
@@ -0,0 +1,68 @@
...
+  }, [selectedFilters, entity])
+
+  return (
+    <Flex px={2}>
+      <Text variant="sm" fontWeight={500}>
+        Your Filters
+      </Text>
+
+      <Flex flexDirection="row" flexWrap="wrap" mt={1} mx={-0.5}>
+        {/* TODO: Adapt useSavedSearchPills to work here with little coupling from saved searches */}
```
MounirDhahri, line 42: Ticketed this here https://artsyproduct.atlassian.net/browse/ONYX-418

dblandin (commented):
Looking good!

A couple high-level comments/questions:
- Since we're only targeting the alert experience with this iteration, would it be more intention-revealing to organize and name these components to reflect that hierarchy — rather than suggesting that these might soon become a new way of filtering artwork grids generally? I think the goals are very different and will require different approaches. With alerts, we want to give collectors the ability to set a criteria based on work that might not already be on the marketplace. That means we won't have the same coupling to search result aggregations that currently inform options for grid filters. Maybe we put any new components under `/Scenes/SavedSearchAlert/Components/Filters`?
- Instead of creating and maintaining a new `NewArtworkFilterStore`, can we accomplish the UX here by utilizing the existing `SavedSearchStore`? I'm not super clear on what we gain by maintaining a separate in-between store (`NewArtworkFilterStore`).

MounirDhahri:
Thanks for the quick review @dblandin 

> Since we're only targeting the alert experience with this iteration, would it be more intention-revealing to organize and name these components to reflect that hierarchy — rather than suggesting that these might soon become a new way of filtering artwork grids generally? I think the goals are very different and will require different approaches. With alerts, we want to give collectors the ability to set a criteria based on work that might not already be on the marketplace. That means we won't have the same coupling to search result aggregations that currently inform options for grid filters. Maybe we put any new components under /Scenes/SavedSearchAlert/Components/Filters?

Your comment makes sense, I probably got too excited and added these separately. My rationale here is based on the conversations we had initially with Eva and Barney and how we would like to port these filters to ALL surfaces and not just saved searches. Because of that, I separated the files and looked into ways on how I would decouple the saved searches from the "new filters". About `aggergations`, they're currently not used in the store, I expected to need them but I didn't. I will remove them.

> Instead of creating and maintaining a new NewArtworkFilterStore, can we accomplish the UX here by utilizing the existing SavedSearchStore? I'm not super clear on what we gain by maintaining a separate in-between store (NewArtworkFilterStore).

I believe the logic can be combined in this case without much issues other than further coupling the filters and saved searches. I did however here as well prefer to keep the logic separate in order to decouple the filters and the saved searches. We currently inject the filter params from the normal filters, and both can exist without each other in any screen. We would still be injecting the new filter params to the saved search but from the new filter screen. By having the new filters store logic inside the the saved searches store, we will not be able to migrate with ease the other filters in the future to match the new one (again based on that our mid-term vision would be to migrate filters to the new easy way)

MounirDhahri:
Had a call with @dblandin to make sure we're aligned here. I will be moving forward toward uniting the filters store with the saved searches store. The reason behind that is that my initial assumption isn't accurate and we only want the filters in the context of saved searches. I am working on a quick adjustment now.

src/app/Scenes/SavedSearchAlert/screens/AddFilterPill.tsx
```diff
@@ -0,0 +1,28 @@
+import { Pill, PillProps } from "@artsy/palette-mobile"
+import { useEffect, useState } from "react"
+import useDebounce from "react-use/lib/useDebounce"
+
+export const FillPill: React.FC<PillProps> = (props) => {
```
olerichter00, line 5: The name of the component and the file name are inconsistent. What do you think about naming the component `FilterPill`?

src/app/Scenes/SavedSearchAlert/screens/AddFilterPill.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+  useDebounce(
+    () => {
+      if (selected !== props.selected) {
+        props.onPress?.()
+      }
+    },
+    200,
+    [selected]
+  )
```
olerichter00, line 16: Why do we need a debounce here? 🤔

src/app/Scenes/SavedSearchAlert/screens/AddFiltersScreen.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { useNavigation } from "@react-navigation/native"
+import { SearchCriteria } from "app/Components/ArtworkFilter/SavedSearch/types"
+import { FancyModalHeader } from "app/Components/FancyModal/FancyModalHeader"
+import { SavedSearchStore } from "app/Scenes/SavedSearchAlert/SavedSearchStore"
+import { AddFiltersScreenAppliedFilters } from "app/Scenes/SavedSearchAlert/screens/AddFiltersScreenAppliedFilters"
+import { AddFiltersScreenRarity } from "app/Scenes/SavedSearchAlert/screens/AddFiltersScreenRarity"
+import { MotiView } from "moti"
+import { Alert } from "react-native"
+
+export const AddFiltersScreen: React.FC<{}> = () => {
```
olerichter00, line 11: Not sure if `AddFiltersScreen` is the best name for this component. Although the CTA that navigates the user to this screen says 'Add Filters' this screen can also be used to remove (or clear all) filters.

What do you think about renaming this component to `SavedSearchFilterScreen`  (and components with a related name)?

src/app/Scenes/SavedSearchAlert/screens/AddFiltersScreenAppliedFilters.tsx
```diff
@@ -0,0 +1,51 @@
+import { Flex, Pill, Text } from "@artsy/palette-mobile"
+import { SearchCriteria } from "app/Components/ArtworkFilter/SavedSearch/types"
+import { SavedSearchStore } from "app/Scenes/SavedSearchAlert/SavedSearchStore"
+import { useSavedSearchPills } from "app/Scenes/SavedSearchAlert/useSavedSearchPills"
+import { MotiView } from "moti"
+
+export const AddFiltersScreenAppliedFilters: React.FC<{}> = ({}) => {
```
olerichter00, line 7: We probably want to move all these components and tests that are not screens into `.../SavedSearchAlert/Components`.

src/app/Scenes/SavedSearchAlert/screens/AddFiltersScreenAppliedFilters.tsx
```diff
@@ -0,0 +1,51 @@
...
+            <Pill
+              m={0.5}
+              variant="filter"
+              accessibilityLabel={pill.label}
+              disabled={
+                pill.paramName === SearchCriteria.artistID ||
+                pill.paramName === SearchCriteria.artistIDs
+              }
+              onPress={() => {
+                // Add remove
```
olerichter00, line 37: I guess, this todo can be removed.

```suggestion
                // Add remove
```

src/app/Scenes/SavedSearchAlert/screens/AddFiltersScreenRarity.tsx
```diff
@@ -0,0 +1,68 @@
...
+      })
+    } else {
+      const newValues = (selectedAttributes || []).concat(value)
+
+      setValueToAttributesByKeyAction({
+        key: SearchCriteria.attributionClass,
+        value: newValues,
+      })
+    }
+  }
```
olerichter00, line 39: This might be easier to read 🤷 

```suggestion
  const handlePress = (value: string) => {
    const isSelected = isValueSelected({
      selectedAttributes,
      value: value,
    })

    if (isSelected) {
      removeValueFromAttributesByKeyAction({
        key: SearchCriteria.attributionClass,
        value: value,
      })
    } else {
      const newValues = (selectedAttributes || []).concat(value)

      setValueToAttributesByKeyAction({
        key: SearchCriteria.attributionClass,
        value: newValues,
      })
    }
  }
```

olerichter00 (commented):
This is great work 🌟

src/app/Scenes/SavedSearchAlert/screens/AddFilterPill.tsx
MounirDhahri: what's even FillPill 😄 - I had FilterPill in mind like the file name

src/app/Scenes/SavedSearchAlert/screens/AddFilterPill.tsx
MounirDhahri: good question, this was actually the reason why I created a new pill. `onPress` events cause lots of state updates on both the visible screen and the main create alert screen. The UX ended up being either laggy or it would just get not responsive. After the debouncing `onPress` this got solved and an interaction like the one in the video below became possible.

https://github.com/artsy/eigen/assets/11945712/5b0899e7-37c6-4159-99cb-0955b86cf895

src/app/Scenes/SavedSearchAlert/screens/AddFiltersScreen.tsx
MounirDhahri: fair comment, should have renamed after Devon's previous comment

src/app/Scenes/SavedSearchAlert/screens/AddFiltersScreenAppliedFilters.tsx
MounirDhahri: good idea

src/app/Scenes/SavedSearchAlert/screens/AddFiltersScreenAppliedFilters.tsx
MounirDhahri: yuup

src/app/Scenes/SavedSearchAlert/screens/AddFiltersScreenRarity.tsx
MounirDhahri: agree

src/app/Scenes/SavedSearchAlert/screens/AddFilterPill.tsx
olerichter00: It would be great if we could use [startTransition](https://react.dev/blog/2022/03/29/react-v18#new-feature-transitions) here...

An immediate state update within the component and a delayed store update with `useEffect` should work (sorry, I cannot test it right now, because I cannot build the app...):

```
export const SavedSearchFilterPill: React.FC<PillProps> = (props) => {
  const [selected, setSelected] = useState(props.selected)

  useEffect(() => {
    props.onPress?.()
  }, [selected])

  return (
    <Pill {...props} mt={1} mr={1} selected={selected} onPress={() => setSelected(!selected)} />
  )
}
```

MounirDhahri: good idea and nice place to try it out first

olerichter00 (dismissed):

src/app/Scenes/SavedSearchAlert/screens/AddFilterPill.tsx
MounirDhahri:
```
  useEffect(() => {
    props.onPress?.()
  }, [selected])
```
is slow though in my case, if you press few times in a row you could see the values getting added and removed with a delay.

MounirDhahri: `startTransition` is not yet supported in our react version

MounirDhahri: I went with your first suggest, and we can update if the performance is bad.

olerichter00: > is slow though in my case, if you press few times in a row you could see the values getting added and removed with a delay.

I see, debouncing the update seems like a good solution to avoid this. Could we decrease the debounce timeout (maybe to 100)?

You could also try to [set `leading` to true](https://lodash.com/docs/4.17.15#debounce) to avoid the timeout when the user clicks for the first time 🤔

dblandin:
Love how this turned out!! Nice work!! 💯 💜 🚀

PR #7759, opened 2022-11-30
fix: webview should move us to native screens if we have them

pvinis:
This PR resolves [MOPLAT-524] 

### Description

- `parseURL` was throwing `deprecated` issues, so I replaced it with the recommendation of just `URL`, and added the polyfill since RN doesn't have the complete thing.
- moved tests to new helpers, and fixed some tests that were green but not testing actually, since things were misconfigured, like `waitFor(() => { expect(..`
- moved the logic from `onShouldStartLoadWithRequest` to `onNavigationStateChange`. Some context [here](https://artsy.slack.com/archives/C02NAJ2CGLW/p1669726228778479), but the main idea is that artsy.net is an SPA, so in the webview the `onShouldStartLoadWithRequest` is mostly not called because of that. The right place to intercept navigation is on `onNavigationStateChange`, which is called always, for every navigation.

### PR Checklist

- [ ] I tested my changes on **iOS** / **Android**.
- [ ] I added screenshots or videos to illustrate my changes.
- [ ] I added Tests and Stories for my changes.
- [ ] I added an [app state migration].
- [ ] I hid my changes behind a [feature flag].
- [ ] I have prefixed changes that need to be tested during a release QA with **[NEEDS EXTERNAL QA]** on the changelog.

### To the reviewers 👀

- [ ] I would like **at least one** of the reviewers to run this PR on the simulator or device.

<details><summary>Changelog updates</summary>

#### Cross-platform user-facing changes

-

#### iOS user-facing changes

-

#### Android user-facing changes

-

#### Dev changes

-

</details>

Need help with something? Have a look at our [docs], or get in touch with us.

[app state migration]: ../blob/main/docs/adding_state_migrations.md
[feature flag]: ../blob/main/docs/developing_a_feature.md
[docs]: ../blob/main/docs/README.md

[MOPLAT-524]: https://artsyproduct.atlassian.net/browse/MOPLAT-524?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #9389, opened 2023-10-06
feat: prevent backward navigation when clicking a link in a webview (EMI-1395)

iskounen:
This PR resolves [EMI-1395]

### Description

When users tap on external links (such as help center articles) in the checkout flow, and then return to the checkout flow by tapping the back button, they are navigated to the previous step of the checkout flow. This is a poor user experience.

This PR addresses that by ensuring that the WebView's backward navigation event (`goBack`) isn't called for taps on external links.

### PR Checklist

#### iOS user-facing changes

- Opening help center links during checkout won't return users to the previous step

</details>

[EMI-1395]: https://artsyproduct.atlassian.net/browse/EMI-1395?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Components/ArtsyWebView.tests.tsx
+++ b/src/app/Components/ArtsyWebView.tests.tsx
@@ -5,6 +5,7 @@ import { goBack, navigate } from "app/system/navigation/navigate"
 import { appJson } from "app/utils/jsonFiles"
 import { renderWithWrappers } from "app/utils/tests/renderWithWrappers"
 import mockFetch from "jest-fetch-mock"
+import { debounce } from "lodash"
 import { stringify } from "query-string"
 import Share from "react-native-share"
 import WebView, { WebViewProps } from "react-native-webview"
@@ -49,6 +50,11 @@ jest.mock("react-native-webview", () => {
   }
 })
 
+jest.mock("lodash", () => ({
+  ...jest.requireActual("lodash"),
+  debounce: jest.fn(),
+}))
+
 const mockOnNavigationStateChange: WebViewNavigation = {
   navigationType: "click",
   url: "https://gooooogle.com",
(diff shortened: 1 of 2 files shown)


ArtsyOpenSource:

- iOS user-facing changes (Opening help center links during checkout won't return users to the previous step)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 558af8a1b6fa3563f3ca9142083c927799f531eb
</p>

iskounen:
@damassi I've asked for your review on this PR because it makes some changes to logic that you added in #7759 to make sure that routing happens correctly for artwork pages. We believe that logic was causing users to be navigated back one step in the checkout flow when they clicked on help center links.

We removed some debouncing and `!_TEST_` guards because we weren't sure that they were necessary. Please let us know if you have thoughts about that change in particular.

damassi:
👀

MrSltun (dismissed):
Looks great!! 🌟🌟🌟

src/app/Components/ArtsyWebView.tests.tsx
```diff
@@ -220,6 +254,63 @@ describe("ArtsyWebViewPage", () => {
       })
       expect(navigate).toHaveBeenCalledWith("https://google.com")
     })
+
+    describe("the inner WebView's goBack method", () => {
+      it("is called when the URL matches a ReactWebView, ModalWebView, or VanityURLEntity route", () => {
```
brainbicycle, line 259: confused about this test case, don't the other test cases contradict it? 

e.g: `it("is called when the URL matches a ReactWebView, ModalWebView, or VanityURLEntity route"` but `it("is not called when the URL matches a ModalWebView route"`

brainbicycle: should it be something like `"it("is not called when the URL matches a ModalWebView route to an external url"` and `it("is called when the URL matches a ReactWebView, ModalWebView, or VanityURLEntity route to a matching route"`?

gkartalis (approved):
We did some QA during mobile OH and it looks good to go!

src/app/Components/ArtsyWebView.tests.tsx
brainbicycle: 👍 makes sense, thanks for the explanation!

brainbicycle (approved):

PR #9390, opened 2023-10-06
docs: testing docs 

gkartalis:
This PR resolves [] 

### Description

Updates our testing docs

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- updates testing docs - gkartalis

</details>

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -1,173 +1,138 @@
-# Testing
+# Testing best practices
 
-So you want to write tests!
+## Table of Contents
 
-- We primarily use `@testing-library/react-native` and our helper `setupTestWrapperTL`
+- [How Should I Query?](#how-should-i-query)
+- [Testing examples for:](#examples)
+  - [useLazyLoadQuery](#testing-a-component-that-has-uselazyloadquery)
+  - [useFragment](#testing-a-component-with-usefragment)
+- [setupTestWrapper](#setuptestwrapper)
+  - [setupTestWrapper params](#setuptestwrapper-params)
+  - [renderWithRelay function](#renderwithrelay-function)
+  - [renderWithRelay params](#renderwithrelay-params)
+  - [renderWithRelay return values](#renderwithrelay-return-values)
 
-- We also use `renderWithWrappers`. In the past we used ReactTestRenderer which we try to now move away from.
+## How should I query
 
-- Get started https://callstack.github.io/react-native-testing-library/docs/getting-started/
-- API Queries https://callstack.github.io/react-native-testing-library/docs/api-queries/
+Under the hood we are using `@testing-library/react-native` to render the component for the testing environment. This means that you can use the query methods that you would use with `@testing-library/react-native`.
 
-- When we want to select an element we refer to this guide 👉 [How should I query?](https://callstack.github.io/react-native-testing-library/docs/how-should-i-query/) from `@testing-library/react-native` for querying components
+Remember in your tests to always try to **resemble how the user interacts with the UI** as much as possible.
 
-Based on the [Guiding Principles](https://testing-library.com/docs/guiding-principles/), your test should resemble how users interact with your code (component, page, etc.) as much as possible.
+You can read more on how to query on the official docs [here](https://callstack.github.io/react-native-testing-library/docs/how-should-i-query).
 
-Therefore we prefer using `screen.getByText`, `screen.getByDisplayValue`, etc, and as a last resort use a `testID` with `screen.getByTestId`.
+## Examples
 
-In our code base there is some "Old fashioned" testing used for Class components for example since they can only be used using the old way in combination with QueryRenderer etc
-
-The "new fashion" testing uses > relay v13 and we now prefer to use hooks whenever possible. We would for example use `useLazyLoadQuery` instead of a `QueryRenderer`.
-
-- [`relay-test-utils`](https://relay.dev/docs/guides/testing-relay-components/) is the framework we use for testing relay components, creating mock environments etc.
-
-- We don’t use snapshot tests; they produce too much churn for too little value.
-
-Our preferred methods to use are marked with a ⭐️, while the ones we'd like to avoid are marked as ❗️ deprecated.
-
-## Test helpers and wrappers, when to use them, and good examples
-
-- `TestRenderer` is a component we create, using the component we want to test. We use a TestRenderer when for example we want to create a mock query. We don't always need to use a test renderer.
-
-New fashion: You can use useLazyLoadQuery inside a TestRendered which is the new relay hooks way for querying data and the way we generally want to swift towards.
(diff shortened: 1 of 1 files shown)


gkartalis:
Still need to write an overview of our helpers and add more examples

docs/testing.md
```diff
@@ -1,173 +1,148 @@
...
 
-- We also use `renderWithWrappers`. In the past we used ReactTestRenderer which we try to now move away from.
+## How should I query
 
-- Get started https://callstack.github.io/react-native-testing-library/docs/getting-started/
-- API Queries https://callstack.github.io/react-native-testing-library/docs/api-queries/
+Under the hood we are using `@testing-library/react-native` to render the component for the testing environment. This means that you can use the query methods that you would use with `@testing-library/react-native`.
 
-- When we want to select an element we refer to this guide 👉 [How should I query?](https://callstack.github.io/react-native-testing-library/docs/how-should-i-query/) from `@testing-library/react-native` for querying components
+Remember in your tests to always try to **resemble how the user interacts with the UI** as much as possible.
```
araujobarret, line 20: Personally, I like this kind of tip, but makes the docs more chunky and people can skip important areas having more things than necessary.
From what I imagine when writing unit tests to UI that's intrinsic to it, we don't need to reminder engineers some more higher level practices/guidelines here, just my thought, happy to hear what the others have to say.

docs/testing.md
```diff
@@ -1,173 +1,148 @@
...
+    Me: () => ({ name: "name" }),
+  })
 
-#### Components not using relay
+  await waitForElementToBeRemoved(() => screen.getByTestId("loading"))
 
-- [CustomSizeInputs.tests.tsx](/src/app/Components/ArtworkFilter/Filters/CustomSizeInputs.tests.tsx)
-- [SizesOptionsScreen.tests.tsx](/src/app/Components/ArtworkFilter/Filters/SizesOptionsScreen.tests.tsx)
+  expect(screen.getByText("name")).toBeTruthy()
+})
```
araujobarret, line 63: simple and clear 👍

docs/testing.md
```diff
@@ -1,173 +1,148 @@
...
+    query BarTestQuery @relay_test_operation {
+      me {
+        ...Bar_me
+      }
+    }
+  `,
+})
 
-We have 3 ways of testing relay components:
+it("works", () => {
```
araujobarret, line 93: maybe `it("renders", () => {` to keep more consistent to all the first render tests we have in the repo

docs/testing.md
```diff
@@ -1,173 +1,148 @@
...
+| ----------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `Component` | The React component you want to test. This component should be typed with the expected response shape for your GraphQL query.                                                                                            |
+| `preloaded` | A boolean flag that indicates whether you want to simulate preloaded data. When set to true, you must provide a query as well. Preloading is useful when you want to test scenarios where data is available immediately. |
+| `query`     | The Relay query you want to test. This query should be typed with the expected response shape for your GraphQL query.                                                                                                    |
+| `variables` | The variables to be used with the GraphQL query. You can provide default values if needed.                                                                                                                               |
```
araujobarret, line 115: I have the opinion we should get rid of `variables`; the reason is you can write in the query the variable value and it's easier to read.
Never found an example where we needed to tweak this value from the tests that required this extra prop.

example:

```jsx
query: graphql`
      query ArtistShows2TestsQuery($artistID: String!) @relay_test_operation {
        artist(id: $artistID) {
          slug
          ...ArtistShows2_artist
        }
      }
    `,
    variables: {
      artistID: "andy-warhol"
    }
```

vs
```jsx
query: graphql`
      query ArtistShows2TestsQuery @relay_test_operation {
        artist(id: "andy-warhol") {
          slug
          ...ArtistShows2_artist
        }
      }
    `,
```

docs/testing.md
damassi: This general RTL guidance I also very much disagree with and favor the test-id approach, since that is a) less code, and b) instant and direct.

docs/testing.md
damassi: Though I don't personally use it, we should keep the pipeline similar to just about every other GraphQL client api, which takes `variables`. (There are many situations where this feature would be necessary.)

docs/testing.md
```diff
@@ -1,173 +1,148 @@
...
-setupTestWrapper_LEGACY : abstract some of the boilerplate of Relay
-Component,
-Query
-Variables
+The renderWithRelay function is returned by setupTestWrapper. It can be used to render your component and control the Relay environment for testing purposes.
 
-_example file: ItemArtwork.tests.tsx_
+```tsx
+const { renderWithRelay } = setupTestWrapper({
+  /* configuration */
```
damassi, line 123: Might as well fill this out. Just a couple more lines for absolute clarity. (Thinking, i've been totally ruined by all of the unnecessarily truncated examples in the Relay docs.)

damassi (approved):
<3 <3 <3 way better in every way

docs/testing.md
MounirDhahri: I would even advocate for avoiding generic descriptions here and instead recommend using something more descriptive. All the tests are supposed to describe that the component is working and is rendered properly. I would suggest instead something like
```typescript
it("shows the profile name", () => {
})  
```

MounirDhahri (approved):

docs/testing.md
```diff
@@ -1,173 +1,148 @@
-# Testing
+# Testing best practices
```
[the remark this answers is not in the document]
gkartalis: yep, makes sense!

docs/testing.md
gkartalis: that sounds reasonable

docs/testing.md
gkartalis: good point, will add this +++

docs/testing.md
gkartalis: I am a huge believer of favoring anything other than test-ID honestly.

I feel like they make the people writing tests a) writing more accessible components if they use this method by default, b) test-IDs are not that resilient to UI changes, c) using getByText / getByLabelText improves the maintainability of the tests as well in my opinion and finally d) more simplified debugging its easier to understand why a test is failing when you are referencing visible elements

gkartalis: @araujobarret I think that many people are not used in writing tests with react-testing-library ways and this makes it easier to have a reference on how to query for the things that they want to query, happy to move it further down though if you think its better to have the actual examples first rather than the how-should-i-query section

damassi: (my comments are def non-blocking!)

PR #9395, opened 2023-10-09
feat(DIA-113): add missing track to some areas of artist screen

araujobarret:
This PR resolves [DIA-113] 

### Description

Investigating the artist screen in Force and comparing it to Eigen we noticed some events we track in Force are not tracked in Eigen, even though we have the mobile events in Cohesion.
This PR adds the missing events, summing up:
- adds `tappedArticleGroup` when tapping an article on the Artist screen
- adds `tappedArticleGroup` when tapping `View All` on the articles section on the Artist screen
- adds `tappedShowGroup` when tapping a show on the Artist screen
- adds `tappedArtistGroup` when tapping a related Artist on the Artist screen
- refactored `Show` to `ArtistShow` and to use relay hooks instead of `createFragmentContainer`

### PR Checklist

### To the reviewers 👀

- add some missing events in the Artist screen - araujobarret

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[DIA-113]: https://artsyproduct.atlassian.net/browse/DIA-113?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Components/Artist/ArtistShows/VariableSizeShowsList.tsx
+++ b/src/app/Components/Artist/ArtistShows/VariableSizeShowsList.tsx
@@ -5,7 +5,7 @@ import { chunk } from "lodash"
 import { Component } from "react"
 import { LayoutChangeEvent, StyleSheet, View } from "react-native"
 import { createFragmentContainer, graphql } from "react-relay"
-import { ArtistShowFragmentContainer } from "./ArtistShow"
+import { ArtistShow } from "./ArtistShow"
 
 interface Props {
   shows: VariableSizeShowsList_shows$data
@@ -65,8 +65,8 @@ class ShowsList extends Component<Props, State> {
         <Stack>
           {chunk(this.props.shows, this.numberOfColumns()).map((shows, index) => (
             <Stack horizontal key={index} style={{ flex: 0 }}>
-              {shows.map((show) => (
-                <ArtistShowFragmentContainer show={show} styles={showStyles} key={show.id} />
+              {shows.map((show, index) => (
+                <ArtistShow show={show} styles={showStyles} key={show.id} index={index} />
               ))}
             </Stack>
           ))}
--- a/src/app/Components/Artist/RelatedArtistsRailCell.tests.tsx
+++ b/src/app/Components/Artist/RelatedArtistsRailCell.tests.tsx
@@ -1,19 +1,23 @@
 import { fireEvent, screen } from "@testing-library/react-native"
 import { RelatedArtistsRailCellTestQuery } from "__generated__/RelatedArtistsRailCellTestQuery.graphql"
 import { navigate } from "app/system/navigation/navigate"
+import { mockTrackEvent } from "app/utils/tests/globallyMockedStuff"
 import { setupTestWrapper } from "app/utils/tests/setupTestWrapper"
 import { graphql } from "react-relay"
-import { RelatedArtistsRailCell } from "./RelatedArtistsRailCell"
+import { RelatedArtistsRailCell, tracks } from "./RelatedArtistsRailCell"
 
 describe("RelatedArtistsRail", () => {
   const { renderWithRelay } = setupTestWrapper<RelatedArtistsRailCellTestQuery>({
-    Component: ({ artist }) => {
-      return <RelatedArtistsRailCell artist={artist} />
+    Component: ({ relatedArtist, artist }) => {
+      return <RelatedArtistsRailCell relatedArtist={relatedArtist} artist={artist} index={0} />
     },
     query: graphql`
       query RelatedArtistsRailCellTestQuery @relay_test_operation {
-        artist(id: "artist-id") @required(action: NONE) {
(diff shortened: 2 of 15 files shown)


src/app/Components/Artist/ArtistShows/VariableSizeShowsList.tsx
```diff
@@ -65,8 +65,8 @@ class ShowsList extends Component<Props, State> {
         <Stack>
           {chunk(this.props.shows, this.numberOfColumns()).map((shows, index) => (
             <Stack horizontal key={index} style={{ flex: 0 }}>
-              {shows.map((show) => (
-                <ArtistShowFragmentContainer show={show} styles={showStyles} key={show.id} />
+              {shows.map((show, index) => (
+                <ArtistShow show={show} styles={showStyles} key={show.id} index={index} />
```
MounirDhahri, line 69: **note(unrelated):** we should probably look into improving this logic in the future towards in favour of avoiding to pass the styles as a prop

src/app/Components/Artist/RelatedArtistsRailCell.tests.tsx
```diff
@@ -64,16 +68,26 @@ describe("RelatedArtistsRail", () => {
...
   formattedNationalityAndBirthday: "American, 01-01-01",
+  slug: "andy-warhol",
   name: "Andy Warhol",
   href: "artist-href",
   isFollowed: false,
   image: {
     url: "image-url",
   },
-}
+} as any
```
MounirDhahri, line 93: in order to avoid this, we can maybe update `tracking.tappedArtistGroup` method to not include relay injected fragments. I would suggest using `CleanRelayFragment` helper to do that. Something like this 👇  would probably work:
```typescript
export const tracks = {
  tappedArtistGroup: (
    relatedArtist: CleanRelayFragment<NonNullable<RelatedArtistsRailCell_relatedArtist$data>>,
    artist: CleanRelayFragment<NonNullable<RelatedArtistsRailCell_artist$data>>,
    index: number
  ): TappedArtistGroup => ({
    ...
  }),
}

```

MounirDhahri (commented):
LGTM, one minor suggestion to avoid `any` and this should be clear to go

src/app/Components/Artist/ArtistShows/VariableSizeShowsList.tsx
araujobarret: agreed, saw it too but seemed too out of scope to include a refactoring to remove it

src/app/Components/Artist/RelatedArtistsRailCell.tests.tsx
araujobarret: I share a different opinion on these relay typings of the tests. 😅 

Let's remember with the test utils I don't need to mock all properties that I want from the object, so I could assign an artist as `Artist: () => ({ id: "mocked-id" })` and it would work, but if I assign the type when I'm declaring the mocked object it's going to force me or to implement all missing keys, or to play around with TS in order to avoid errors, increases the verbosity for almost no gain since the `MockPayloadGenerator.generate` ignores the types and auto-fills any missing property that the component needs.

I remember doing a lot of refactoring on the Order app in the tests and one of the annoying points there was were exactly huge mocked objects the types of fragments of the component where we would use just 1 or 2 keys but because of the types the tests' size were huge plus the verbosity of the types, again for no gain in terms of what the typings do for us.

That said I'd prefer having the any in the mocked object rather than making the type there strict (relay tests only). @damassi do you share any ideas here too?

araujobarret: The error is valid but not soo valid because the property is going to be auto-filled by the mocker, and if we don't wanna make it custom there's no problem leaving the job to the mocker imo. I guess using the types provided by the relay compiler in the tests we're using something that not even the docs(which are poor) say to do it, they recommend using the types in the component, not in the testing-related documentation.

> If we would like to avoid this, and avoid also having a huge inline object, we can potentially pick the properties we need otherwise there is no use for the type.

that's true, but not all the time we want to inline if it's something we're going to reuse across multiple test cases, so strong types there can be more pain than helpers.

re snapshot tests, we should discuss more, particularly since I don't see the need and other folks like @jonallured  share the same opinion https://artsy.slack.com/archives/C02BC3HEJ/p1669730278300759?thread_ts=1669718588.443139&cid=C02BC3HEJ

MounirDhahri: good point, I also don't see value in using snapshot tests to test components. I see only value in them for testing method arguments in contexts like this.

MounirDhahri: Fwiw, my comment is a nitpick, it's not blocking to merge the PR.

MounirDhahri (approved):

PR #9444, opened 2023-10-18
feat: add catch blocks to fetch requests + reporting

gkartalis:
This PR resolves [PHIRE-325] 

### Description

Adds catch blocks to fetch requests + captureMessage for error reporting on Sentry.

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- add catch blocks to fetch requests + reporting - gkartalis

</details>

[PHIRE-325]: https://artsyproduct.atlassian.net/browse/PHIRE-325?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Components/ShareSheet/ShareSheet.tsx
+++ b/src/app/Components/ShareSheet/ShareSheet.tsx
@@ -15,6 +15,7 @@ import {
   WhatsAppAppIcon,
 } from "@artsy/palette-mobile"
 import Clipboard from "@react-native-clipboard/clipboard"
+import { captureMessage } from "@sentry/react-native"
 import { FancyModal } from "app/Components/FancyModal/FancyModal"
 import { FancyModalHeader } from "app/Components/FancyModal/FancyModalHeader"
 import { useShareSheet } from "app/Components/ShareSheet/ShareSheetContext"
@@ -91,11 +92,17 @@ export const ShareSheet = () => {
   const handleMorePress = async () => {
     const details = shareContent(data)
 
-    const resp = await RNFetchBlob.config({
-      fileCache: true,
-    }).fetch("GET", currentImageUrl)
+    let resp
 
-    const base64RawData = await resp.base64()
+    try {
+      resp = await RNFetchBlob.config({
+        fileCache: true,
+      }).fetch("GET", currentImageUrl)
+    } catch (error) {
+      captureMessage(`Error fetching RNFetchBlob imgUrl: ${error}`, "error")
+    }
+
+    const base64RawData = await resp?.base64()
     const base64Data = `data:image/png;base64,${base64RawData}`
 
     const shareOptions = {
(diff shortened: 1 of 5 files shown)


ArtsyOpenSource:

- Dev changes (add catch blocks to fetch requests + reporting - gkartalis)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 281eeaa43c8bbc0b6e71a58279982bdbc34e5a85
</p>

MrSltun (approved):
⛑️

src/app/utils/PushNotification.ts
```diff
@@ -79,8 +80,16 @@ export const saveToken = (token: string, ignoreSameTokenCheck = false) => {
...
+
+        let res
+
+        try {
+          res = await fetch(request)
+        } catch (error) {
+          captureMessage(`Push Notification Error while fetching token: ${error}`, "error")
+        }
+
+        const response = await res?.json()
```
MounirDhahri, line 92: **Q:** any reason why you extracted `res` as a separate variable here? I would be in favor of wrapping both asyc statements with a try catch statement

src/app/utils/googleMaps.ts
```diff
@@ -53,10 +54,14 @@ export const getLocationDetails = async ({
...
-  )
-  const data = await response.json()
+  let response
+
+  try {
+    response = await fetch("https://maps.googleapis.com/maps/api/place/details/json?" + queryString)
+  } catch (error) {
+    captureMessage(`Error fetching getLocationDetails: ${error}`, "error")
+  }
+  const data = await response?.json()
```
MounirDhahri, line 64: same here

MounirDhahri (commented):
thanks for adding this. I see you didn't throw an error on failures and I think that makes sense since we wouldn't want to break the UX here. I would suggest though to show a toast though in case something didn't go as expected. 
For example, you would press on a link that matches `click.artsy.net`, instead of failing silently and not navigating the user, it could be nice to show a hint about it. We can use `showToast` with a red background color

src/app/utils/PushNotification.ts
gkartalis: ++ makes sense will do that

src/app/Components/ShareSheet/ShareSheet.tsx
```diff
@@ -91,11 +92,17 @@ export const ShareSheet = () => {
...
-    const base64RawData = await resp.base64()
+    try {
+      resp = await RNFetchBlob.config({
+        fileCache: true,
+      }).fetch("GET", currentImageUrl)
+    } catch (error) {
+      captureMessage(`Error fetching RNFetchBlob imgUrl: ${error}`, "error")
+    }
+
+    const base64RawData = await resp?.base64()
```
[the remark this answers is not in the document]
araujobarret: Another unrelated point, I see on line `111` we don't need these `RNFetchBlob.fetch` for cases different than `sale`, isn't better to encapsulate it in a function to only return `{ url: base64Data }` once it's the accepted case?

brainbicycle (approved):
🦺

gkartalis:
I will actually close this one and reopen - refactored variant of this with a wrapper function around fetch that has also captureMessage for sentry, but will take your comments into account and will re-add you as reviewers

PR #9504, opened 2023-10-27
fix(ArtistHeader): fix Artist image disappearing

araujobarret:
### Description

This PR addresses one issue when navigating to a Partner with the verified Artist pill, from the Artist screen, would override the graphql property on Relay store `Artist#coverArtwork#image#url` to `undefined`. Thread [here](https://artsy.slack.com/archives/C05EEBNEF71/p1698412082704999).

Other changes:
  - refactors the ArtistHeader to use Relay hooks
  - ~removes the Artworks counter due to an overlapping bug(no quick fix there)~ fixed in https://github.com/artsy/eigen/pull/9504/commits/1c2bb975efa4915f52452f906767f65e3688c601

https://github.com/artsy/eigen/assets/15792853/e6c83d2f-403f-48c0-9d7d-5e9dd3dce09f

### PR Checklist

### To the reviewers 👀

- fix: artist image in the artist screen disappears no more - araujobarret

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
+++ b/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
@@ -287,12 +287,19 @@ const ArtworksGrid: React.FC<ArtworksGridProps> = ({
         // be visible above list content
         ListHeaderComponentStyle={{ zIndex: 1 }}
         ListHeaderComponent={
-          <Tabs.SubTabBar>
-            <ArtistArtworksFilterHeader
-              artist={artist}
-              showCreateAlertModal={() => setIsCreateAlertModalVisible(true)}
-            />
-          </Tabs.SubTabBar>
+          <>
+            <Tabs.SubTabBar>
+              <ArtistArtworksFilterHeader
+                artist={artist}
+                showCreateAlertModal={() => setIsCreateAlertModalVisible(true)}
+              />
+            </Tabs.SubTabBar>
+            <Flex pt={1}>
+              <Text variant="xs" weight="medium">{`${artist.counts?.artworks} Artwork${
+                artist.counts?.artworks > 1 ? "s" : ""
+              }:`}</Text>
+            </Flex>
+          </>
         }
         ListFooterComponent={<ListFooterComponenet />}
       />
--- a/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
+++ b/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
@@ -292,11 +292,6 @@ const ArtworksGrid: React.FC<ArtworksGridProps> = ({
               artist={artist}
               showCreateAlertModal={() => setIsCreateAlertModalVisible(true)}
             />
-            <Flex px={2} pt={1}>
-              <Text variant="xs" weight="medium">{`${artist.counts?.artworks} Artwork${
-                artist.counts?.artworks > 0 ? "s" : ""
-              }:`}</Text>
-            </Flex>
           </Tabs.SubTabBar>
         }
         ListFooterComponent={<ListFooterComponenet />}
--- a/src/app/Components/Artist/ArtistHeader.tsx
+++ b/src/app/Components/Artist/ArtistHeader.tsx
@@ -11,14 +11,17 @@ import {
   useSpace,
 } from "@artsy/palette-mobile"
 import { useScreenScrollContext } from "@artsy/palette-mobile/dist/elements/Screen/ScreenScrollContext"
-import { ArtistHeader_artist$data } from "__generated__/ArtistHeader_artist.graphql"
-import { ArtistHeader_me$data } from "__generated__/ArtistHeader_me.graphql"
+import {
+  ArtistHeader_artist$data,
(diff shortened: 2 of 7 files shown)


src/app/Components/Artist/ArtistHeader.tsx
```diff
@@ -56,14 +58,20 @@ export const ArtistHeader: React.FC<Props> = ({ artist, me, onLayoutChange }) =>
   const { updateScrollYOffset } = useScreenScrollContext()
   const showArtistsAlertsSetFeatureFlag = useFeatureFlag("ARShowArtistsAlertsSet")
   const tracking = useTracking()
+  const artistData = useFragment(artistFragment, artist)
+  const meData = useFragment(meFragment, me)
+
+  if (!artistData || !meData) {
+    return null
+  }
```
gkartalis, line 66: **question(non-blocking):** do we want to have this a bit further down in the component? In order to avoid hiding everything `artistData` related i.e. if `meData` fails only 🤔

src/app/Components/Artist/ArtistHeader.tsx
```diff
@@ -56,14 +58,20 @@ export const ArtistHeader: React.FC<Props> = ({ artist, me, onLayoutChange }) =>
   const { updateScrollYOffset } = useScreenScrollContext()
   const showArtistsAlertsSetFeatureFlag = useFeatureFlag("ARShowArtistsAlertsSet")
   const tracking = useTracking()
+  const artistData = useFragment(artistFragment, artist)
+  const meData = useFragment(meFragment, me)
```
gkartalis, line 62: 👏 bonus clap for the hooks migration

src/app/Components/Artist/ArtistHeader.tsx
```diff
@@ -175,41 +183,41 @@ export const ArtistHeader: React.FC<Props> = ({ artist, me, onLayoutChange }) =>
...
-          url(version: "large")
-        }
+const artistFragment = graphql`
+  fragment ArtistHeader_artist on Artist {
+    slug
+    birthday
+    coverArtwork {
+      title
+      image {
+        url(version: "larger")
```
gkartalis, line 193: **question(curiosity):** What's the difference on large vs larger? Why did we go for larger?

src/app/Components/ArtistListItem.tsx
```diff
@@ -220,10 +219,10 @@ export const ArtistListItemContainer = createFragmentContainer(ArtistListItem, {
...
       birthday
       deathday
-      coverArtwork {
-        image {
-          url
-        }
+      # TOFIX: we must use coverArtwork#image here instead, this replacement is fixing
+      # an Artist#coverImage got replaced by this component data with wrong data
+      image {
+        url
```
gkartalis, line 225: 👀 Let's create a ticket around that? Just to make sure that we don't forget about this

src/app/Components/Artist/ArtistHeader.tsx
araujobarret: looks the same though

src/app/Components/ArtistListItem.tsx
araujobarret: Ticket created in https://artsyproduct.atlassian.net/browse/DIA-303

src/app/Components/Artist/ArtistHeader.tsx
araujobarret: I don't think so, we don't have a partial render logic of the screen plus rendering the screen depends on both fragment s' data, so this kind of practice is very common.

src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
```diff
@@ -292,11 +292,6 @@ const ArtworksGrid: React.FC<ArtworksGridProps> = ({
               artist={artist}
               showCreateAlertModal={() => setIsCreateAlertModalVisible(true)}
             />
-            <Flex px={2} pt={1}>
-              <Text variant="xs" weight="medium">{`${artist.counts?.artworks} Artwork${
-                artist.counts?.artworks > 0 ? "s" : ""
-              }:`}</Text>
-            </Flex>
```
egdbear, line 299: I noticed that we are removing this from the PR description. What is the issue here?
I see that we have the counts in the design:

image

src/app/Components/ArtistListItem.tsx
gkartalis: thanks!

src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
araujobarret: It has a bug, if you scroll this counter becomes sticky and overlaps with the Grid, there was no easy solution to fix it in this PR, I will try to make a fix for this release but I'd need to go to Palette first then come back to Eigen 😞

egdbear: Thanks @araujobarret for solving this! 🚀

gkartalis (approved):
thank you! Lets 🚢

PR #9652, opened 2023-12-11
feat(EMI-1549): Handling deep linking and push notifications for Partner Offers

MrSltun:
This PR resolves [EMI-1549] 

### Description

This PR adds a new handler for deep linking and push notifications for Partner Offers

### Videos

| Scenario | Android | iOS |
|---|---|---|
| Valid Offer | <video src="https://github.com/artsy/eigen/assets/20655703/663cf46f-83ad-4f6c-bb2a-903c32c87ffb" /> | <video src="https://github.com/artsy/eigen/assets/20655703/8ba6e9cd-3f44-4159-8b31-87f1a874b248" /> |
| Offer expired / artwork sold | <video src="https://github.com/artsy/eigen/assets/20655703/12219e07-427a-4875-9cc7-f324b32823fe" /> | <video src="https://github.com/artsy/eigen/assets/20655703/a8f4acbc-bb2e-4a70-ade0-e806603bf118" /> |
| Invalid offer | <video src="https://github.com/artsy/eigen/assets/20655703/7fbbb7d2-95d0-48e6-a709-db114d2b1dca" /> | <video src="https://github.com/artsy/eigen/assets/20655703/6cf5b45d-524a-4358-bd41-d7031aaa8f2b" /> |

### PR Checklist

### To the reviewers 👀

- Handle deep linking and push notifications for Partner Offers - mrsltun

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[EMI-1549]: https://artsyproduct.atlassian.net/browse/EMI-1549?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ArtsyOpenSource:

- Cross-platform user-facing changes (Handle deep linking and push notifications for Partner Offers - mrsltun)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against f954fd4112aefb0967319feb7c47f40b2ed6ed7b
</p>

MounirDhahri:
Thanks for the thorough videos @MrSltun - 

**Feedback unrelated to the PR**
In the first video, I noticed that iOS is prompting us to "Open Artsy", although I believe that we have Universal links set up and **it should open the link without a prompt**. @brainbicycle @gkartalis  Would you mind guys if you check this out at some point in the future. Also when opening Banksy artist page from Chrome in my phone, I wasn't prompted to open it in the App.

src/app/Scenes/PartnerOffer/PartnerOfferContainer.tsx
```diff
@@ -0,0 +1,51 @@
...
+export const PartnerOfferContainer: React.FC<{ partnerOfferID: string }> = ({ partnerOfferID }) => {
+  const handleRedirect = (response: usePartnerOfferCheckoutMutation$data) => {
+    const orderOrError = response.commerceCreatePartnerOfferOrder?.orderOrError
+
+    if (orderOrError?.error) {
+      const { code, data } = orderOrError.error
+      const artworkId = JSON.parse(data?.toString() ?? "")?.artwork_id
+
+      if (code === "expired_partner_offer") {
+        // TODO: add params to show expired banner (EMI-1606)
```
MounirDhahri, line 17: 👍

src/app/Scenes/PartnerOffer/PartnerOfferContainer.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { usePartnerOfferMutation } from "app/Scenes/PartnerOffer/mutations/usePartnerOfferCheckoutMutation"
+import { goBack, navigate } from "app/system/navigation/navigate"
+import { useEffect } from "react"
+
+export const PartnerOfferContainer: React.FC<{ partnerOfferID: string }> = ({ partnerOfferID }) => {
+  const handleRedirect = (response: usePartnerOfferCheckoutMutation$data) => {
+    const orderOrError = response.commerceCreatePartnerOfferOrder?.orderOrError
+
+    if (orderOrError?.error) {
+      const { code, data } = orderOrError.error
```
MounirDhahri, line 13: too late for this but `error_code` would have been a nice name for this. 
```suggestion
      const { code: errorCode, data: maybeArtwork } = orderOrError.error
```

src/app/Scenes/PartnerOffer/PartnerOfferContainer.tsx
```diff
@@ -0,0 +1,51 @@
...
+      const artworkId = JSON.parse(data?.toString() ?? "")?.artwork_id
```
MounirDhahri, line 14: Since `JSON.parse("")` will throw an exception, I guess you probably want something like ``
```suggestion
      const artworkId = JSON.parse(data?.toString() ?? "{}")?.artwork_id
```

src/app/Scenes/PartnerOffer/PartnerOfferContainer.tsx
MounirDhahri: maybe `data:artwork` could also be better here if we're sure there will be an artwork returned - since the offered artwork might get deleted or unpublished

src/app/Scenes/PartnerOffer/PartnerOfferContainer.tsx
```diff
@@ -0,0 +1,51 @@
...
+        navigate(`/artwork/${artworkId}`, { replaceActiveScreen: true })
+        return
+      }
+
+      if (code === "not_acquireable") {
+        // TODO: add params to show unavailable/sold banner (EMI-1606)
+        navigate(`/artwork/${artworkId}`, { replaceActiveScreen: true })
+        return
+      }
+    }
```
MounirDhahri, line 27: I think we might need to throw a return here in case the error code is not in the list above or have an else statement here to make sure we don't go to the next block if we have a new error that is not supported in the app. Something like this 👇 
```suggestion
    } else {
```

src/app/Scenes/PartnerOffer/PartnerOfferContainer.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+    if (orderOrError?.order) {
+      // we need to go back to the home screen before navigating to the orders screen
+      // to prevent the user from closing the modal and navigating back to the this screen
```
MounirDhahri, line 31: good thinking - I think though that you are assuming they are in a home tab and it's probably the case for most users but it's possible as well that they might be in the app when they got the notification and in a different tab. You might want to consider using one of the helpers we have that reset the navigation state in the app like `popToRootAndScrollToTop` or so (not sure)

MounirDhahri: oh actually we can even use `replaceActiveModal` if we also set this screen as a modal 👀

MrSltun: I tried to update the code with this one, but it doesn't work as expected.
The idea here is to add the `PartnerOfferContainer` to the stack, and replace it depending on the result of the mutation, regardless of where the collector is :)

brainbicycle (approved):
Looks good!

brainbicycle:
> Thanks for the thorough videos @MrSltun - 
> 
> 
> 
> **Feedback unrelated to the PR**
> 
> In the first video, I noticed that iOS is prompting us to "Open Artsy", although I believe that we have Universal links set up and **it should open the link without a prompt**. @brainbicycle @gkartalis  Would you mind guys if you check this out at some point in the future. Also when opening Banksy artist page from Chrome in my phone, I wasn't prompted to open it in the App.

Hmm I am not sure why the open with artsy is showing in the first video but universal links work as far as I can tell, opening an artsy.net link from say slack or an email opens the ios app without prompting. I think when you are in the artsy mobile web site on chrome or safari it is not going to open the app because of our react routing. That said I am not sure we want it to anyway because if you are browsing the site you don't want to be kicked to the app when switching pages in my opinion. We do have the smart banners in safari that show "open in artsy" which i think is working as intended.

MounirDhahri:
I was going to answer here but I believe my comment will diverge from the PR, I will move the conversation to Slack

PR #9688, opened 2024-01-03
feat: Add new works for you screen + fixes

MounirDhahri:
Reverts artsy/eigen#9670

This PR makes the following changes:
- Move the view mode toggle button to the top of the screen
- Add bottom padding to the scroll view (and it's not broken this time 😄)
- Wrap the screen with a suspense (this was why it was sometimes broken)

https://github.com/artsy/eigen/assets/11945712/083d882b-2cec-40bd-bc1e-c6269d3a37fa

https://github.com/artsy/eigen/assets/11945712/f0e8fd42-99b3-4ad3-8bd4-244d0603945d

### DIFF
--- a/src/app/Components/ArtworkRail/ArtworkRailCard.tsx
+++ b/src/app/Components/ArtworkRail/ArtworkRailCard.tsx
@@ -159,7 +159,7 @@ export const ArtworkRailCard: React.FC<ArtworkRailCardProps> = ({
           return imageDimensions.width
         }
       case "fullWidth":
-        return imageDimensions.width
+        return Dimensions.get("window").width
 
       default:
         assertNever(size)
@@ -376,7 +376,6 @@ const ArtworkRailCardImage: React.FC<ArtworkRailCardImageProps> = ({
 }) => {
   const color = useColor()
   const EXTRALARGE_RAIL_CARD_IMAGE_WIDTH = useExtraLargeWidth()
-  const { height: screenHeight } = useScreenDimensions()
 
   if (!containerWidth) {
     return null
@@ -403,12 +402,14 @@ const ArtworkRailCardImage: React.FC<ArtworkRailCardImageProps> = ({
     {
       width: isRecentlySoldArtwork
         ? EXTRALARGE_RAIL_CARD_IMAGE_WIDTH
+        : size === "fullWidth"
+        ? Dimensions.get("screen").width
         : ARTWORK_LARGE_RAIL_CARD_IMAGE_WIDTH,
       height: ARTWORK_RAIL_CARD_IMAGE_HEIGHT[size],
     }
   )
 
-  const aspectRatio = imageDimensions.width / imageDimensions.height
+  const aspectRatio = image?.aspectRatio || imageDimensions.width / imageDimensions.height
 
   const getImageHeight = () => {
     let adjustedHeight = 0
@@ -423,17 +424,17 @@ const ArtworkRailCardImage: React.FC<ArtworkRailCardImageProps> = ({
       return adjustedHeight
     }
 
-    const MINIMUM_IMAGE_HEIGHT = screenHeight * 0.2
-    const MAXIMUM_IMAGE_HEIGHT = screenHeight * 0.7
+    const MINIMUM_ASPECT_RATIO = 0.2
+    const MAXIMUM_ASPECT_RATIO = 0.7
 
     // Make sure that the image height is not too little or not too big
     // See https://artsy.slack.com/archives/C05EQL4R5N0/p1701167802957189?thread_ts=1701164551.999919&cid=C05EQL4R5N0
-    if (height / aspectRatio < MINIMUM_IMAGE_HEIGHT) {
-      return MINIMUM_IMAGE_HEIGHT
-    } else if (height / aspectRatio > MAXIMUM_IMAGE_HEIGHT) {
-      return MAXIMUM_IMAGE_HEIGHT
+    if (aspectRatio < MINIMUM_ASPECT_RATIO) {
+      return Dimensions.get("screen").width / MINIMUM_ASPECT_RATIO
+    } else if (aspectRatio > MAXIMUM_ASPECT_RATIO) {
+      return Dimensions.get("screen").width / aspectRatio
     } else {
-      return height / aspectRatio
+      return Dimensions.get("screen").width / aspectRatio
     }
   }
 
@@ -441,7 +442,7 @@ const ArtworkRailCardImage: React.FC<ArtworkRailCardImageProps> = ({
     <Flex>
       <Flex width={containerWidth}>
         <OpaqueImageView
-          style={{ maxHeight: ARTWORK_RAIL_CARD_IMAGE_HEIGHT[size] }}
+          // style={{ maxHeight: ARTWORK_RAIL_CARD_IMAGE_HEIGHT[size] }}
           imageURL={src}
           height={getImageHeight()}
           width={containerWidth}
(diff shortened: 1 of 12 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 36303b78d8dd8694082cdac476edf52c5e77b2fd
</p>

src/app/Components/ArtworkRail/ArtworkRailCard.tsx
```diff
@@ -62,6 +72,7 @@ export const ArtworkRailCard: React.FC<ArtworkRailCardProps> = ({
   lotLabel,
   lowEstimateDisplay,
   highEstimateDisplay,
+  metaContainerStyles,
```
olerichter00, line 75: I wonder if there is a better name. Could we just name it `containerStyles`? Because it's applied to the same container as `containerWidth`.

src/app/Components/ArtworkRail/ArtworkRailCard.tsx
```diff
@@ -356,6 +377,10 @@ const ArtworkRailCardImage: React.FC<ArtworkRailCardImageProps> = ({
   const color = useColor()
   const EXTRALARGE_RAIL_CARD_IMAGE_WIDTH = useExtraLargeWidth()
 
+  if (!containerWidth) {
+    return null
+  }
```
olerichter00, line 382: **Question:** When can this case happen? And how would it look like if `containerWidth` is `0` or `null`?

src/app/Components/ArtworkRail/ArtworkRailCard.tsx
```diff
@@ -377,23 +402,50 @@ const ArtworkRailCardImage: React.FC<ArtworkRailCardImageProps> = ({
     {
       width: isRecentlySoldArtwork
         ? EXTRALARGE_RAIL_CARD_IMAGE_WIDTH
+        : size === "fullWidth"
+        ? Dimensions.get("screen").width
```
olerichter00, line 406: Maybe we can use a variable instead of using a nested ternary operator. 

I guess it's okay like this, but it can be quite hard to read.

src/app/Components/ArtworkRail/ArtworkRailCard.tsx
```diff
@@ -377,23 +402,50 @@ const ArtworkRailCardImage: React.FC<ArtworkRailCardImageProps> = ({
...
+      return Dimensions.get("screen").width / aspectRatio
+    }
+  }
+
   return (
     <Flex>
       <Flex width={containerWidth}>
         <OpaqueImageView
-          style={{ maxHeight: ARTWORK_RAIL_CARD_IMAGE_HEIGHT[size] }}
+          // style={{ maxHeight: ARTWORK_RAIL_CARD_IMAGE_HEIGHT[size] }}
```
olerichter00, line 445: 👀

src/app/Scenes/NewWorksForYou/Components/NewWorksForYouGrid.tsx
```diff
@@ -0,0 +1,133 @@
...
+            />
+          </Flex>
+        </Animated.ScrollView>
+        {headerElement}
+      </SafeAreaView>
+    </ProvideScreenTrackingWithCohesionSchema>
+  )
+}
+
+export const oldNewWorksForYouGridFragment = graphql`
```
olerichter00, line 76: 👀 

```suggestion
export const newWorksForYouGridFragment = graphql`
```

src/app/Scenes/NewWorksForYou/Components/NewWorksForYouGrid.tsx
```diff
@@ -0,0 +1,133 @@
...
+  ({ version, maxWorksPerArtist }) => {
+    const data = useLazyLoadQuery<NewWorksForYouGridQuery>(newWorksForYouGridQuery, {
+      version,
+      maxWorksPerArtist,
+    })
+
+    // This won't happen because the query would fail thanks to the @principalField
+    // Adding it here to make TS happy
+    if (!data.viewer) {
+      return <Text>Something went wrong</Text>
```
olerichter00, line 127:
```suggestion
      return <Text>Something went wrong.</Text>
```

olerichter00 (commented):
Nice work 🌟

src/app/Components/ArtworkRail/ArtworkRailCard.tsx
MounirDhahri: I think `containerStyles` can be misleading here since it's only applied to the `meta` of the artwork. `containerWidth` is also applied there but that's because we need to know the width when the image width is too small to limit it.

src/app/Components/ArtworkRail/ArtworkRailCard.tsx
MounirDhahri: good idea

src/app/Components/ArtworkRail/ArtworkRailCard.tsx
MounirDhahri: it should actually never happen, I will update it

src/app/Components/ArtworkRail/ArtworkRailCard.tsx
```diff
@@ -31,11 +39,12 @@ export const ARTWORK_RAIL_CARD_IMAGE_HEIGHT = {
   small: 230,
   large: 320,
   extraLarge: 400,
+  fullWidth: Dimensions.get("window").height,
```
dariakoko, line 42: Would using `useScreenDimensions()` work here?
`const { height } = useScreenDimensions()`

src/app/Components/ArtworkRail/ArtworkRailCard.tsx
```diff
@@ -140,6 +158,8 @@ export const ArtworkRailCard: React.FC<ArtworkRailCardProps> = ({
         } else {
           return imageDimensions.width
         }
+      case "fullWidth":
+        return Dimensions.get("window").width
```
dariakoko, line 162: and here as well

src/app/Scenes/NewWorksForYou/NewWorksForYou.tsx
```diff
@@ -1,184 +1,168 @@
...
+                <SkeletonBox key={i} width="100%" height={400} />
+              </Flex>
+              <Spacer y={1} />
+              <SkeletonText>David Hockey</SkeletonText>
+              <Spacer y={0.5} />
+              <SkeletonText>Mercy from the Virtues H9-13 </SkeletonText>
+              <Spacer y={0.5} />
+              <SkeletonText>Berg Contemporary</SkeletonText>
+              <Spacer y={0.5} />
+              <SkeletonText>£38,000</SkeletonText>
```
dariakoko, line 161: **Question**: Was it a requirement to use the price in the SkeletonText here? I'm thinking, wouldn't it be better to stay consistent and not show this data on the loading screen?

dariakoko (approved):
Looks great ⭐

src/app/Scenes/NewWorksForYou/NewWorksForYou.tsx
MounirDhahri: Great question. The skeleton text doesn't actually show the data- the text is only used to measure how big the skeleton box should be. 
The implementation goal is to do this:
```typescript
<SkeletonText>£38,000</SkeletonText>
```

instead of guessing the width of a container as in here since we calculate the width based on the injected text

```typescript
<PlaceholderText width={170} />
```

src/app/Components/ArtworkRail/ArtworkRailCard.tsx
MounirDhahri: Good point, the reason behind using this is because ` useScreenDimensions()` gets the dimensions from the `screen`, however in the case, I need the height of the `window` instead. They're different in Android and it leads to calculating the menu bar and messing up the calculation here

PR #9735, opened 2024-01-17
feat: add fancy collapsible header logic

MounirDhahri:
### Description
This PR adds logic to support a fancy collapsible header. This is needed for the upcoming work on the `Activity` screen and potentially for other similar screens. Ideally, we can have a similar logic for all the screens 🤞

### How to use
Screenshot 2024-01-17 at 11 34 03

### Customization

Untitled-2023-04-27-1400

### Demo

https://github.com/artsy/eigen/assets/11945712/7a198180-9646-4073-bf2b-03bacd4d086a

### PR Checklist

### To the reviewers 👀

- add fancy collapsible header logic - mounir

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/utils/useFancyHeader.tsx
+++ b/src/app/utils/useFancyHeader.tsx
@@ -1,0 +1,227 @@
+import {
+  BackButton,
+  Flex,
+  Separator,
+  Text,
+  useScreenDimensions,
+  useSpace,
+} from "@artsy/palette-mobile"
+import { TEXT_VARIANTS } from "@artsy/palette-tokens/dist/typography/v3"
+import { useMeasure } from "app/utils/hooks/useMeasure"
+import { useMemo, useRef } from "react"
+import { View } from "react-native"
+import Animated, {
+  useAnimatedProps,
+  useAnimatedScrollHandler,
+  useAnimatedStyle,
+  useSharedValue,
+  withTiming,
+} from "react-native-reanimated"
+
+const DYNAMIC_CONTENT_HEIGHT_MAX = 120
+const DYNAMIC_CONTENT_HEIGHT_MIN = 56
+const STICKY_CONTENT_HEIGHT = 50
+const DYNAMIC_CONTENT_TITLE_FONT_SIZE = parseInt(TEXT_VARIANTS["lg-display"].fontSize, 10)
+const NAVIGATION_TITLE_FONT_SIZE = parseInt(TEXT_VARIANTS["sm"].fontSize, 10)
+const DYNAMIC_CONTENT_TITLE_INCREASE_SIZE = 4
+
+const HORIZONTAL_PADDING = 20
+
+export const useFancyHeader = ({
+  title = "Activity",
+  rightButton,
+  stickyContent,
+}: {
+  title: string
+  rightButton: JSX.Element
+  stickyContent?: JSX.Element
+}) => {
+  const space = useSpace()
+  const { width: screenWidth } = useScreenDimensions()
+
+  const titleRef = useRef<Text>(null)
+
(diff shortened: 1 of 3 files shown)


src/app/utils/useFancyHeader.tsx
```diff
@@ -0,0 +1,227 @@
...
+const DYNAMIC_CONTENT_HEIGHT_MAX = 120
+const DYNAMIC_CONTENT_HEIGHT_MIN = 56
+const STICKY_CONTENT_HEIGHT = 50
+const DYNAMIC_CONTENT_TITLE_FONT_SIZE = parseInt(TEXT_VARIANTS["lg-display"].fontSize, 10)
+const NAVIGATION_TITLE_FONT_SIZE = parseInt(TEXT_VARIANTS["sm"].fontSize, 10)
+const DYNAMIC_CONTENT_TITLE_INCREASE_SIZE = 4
+
+const HORIZONTAL_PADDING = 20
+
+export const useFancyHeader = ({
```
MounirDhahri, line 30: I need to look into a less lazy name here

gkartalis (commented):
**question:** since we do have the Screen and Screen.AnimatedHeader components in palette are we thinking of embedding this logic there as well?

ArtsyOpenSource:

- Cross-platform user-facing changes (add fancy collapsible header logic - mounir)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 410c2cd63be8884412ba78997628d4c230031229
</p>

olerichter00 (commented):
This looks great! 🌟 

Doesn't [PageWithSimpleHeader](https://github.com/artsy/eigen/blob/0d20151ef900df94801f37a79b091c8794059c12/src/app/Components/PageWithSimpleHeader.tsx) has the same functionality (just without this great animation)?

MounirDhahri:
> since we do have the Screen and Screen.AnimatedHeader components in palette are we thinking of embedding this logic there as well?

Nice idea, I think it should be possible to add this there as well. I had this here initially because it's only used in the `Activity` screen but I will talk to design to see if this is something we would like to add everywhere to include it in palette.

MounirDhahri:
> This looks great! 🌟

Thank you!

> Doesn't [PageWithSimpleHeader](https://github.com/artsy/eigen/blob/0d20151ef900df94801f37a79b091c8794059c12/src/app/Components/PageWithSimpleHeader.tsx) has the same functionality (just without this great animation)?

Thanks for the feedback as well. Yes, this is accurate, but it's unfortunately not as nice as a UX as I believe this is. My goal here is to slightly elevate the UX and hopefully we can have this in many other screens once Barney approves it

olerichter00:
> Thanks for the feedback as well. Yes, this is accurate, but it's unfortunately not as nice as a UX as I believe this is. My goal here is to slightly elevate the UX and hopefully we can have this in many other screens once Barney approves it.

Yes, your header component looks way better 🤩. Then, I will approve this. Feel free to add it to the new Activity screen. 

But anyway, I think we **should** update [PageWithSimpleHeader](https://github.com/artsy/eigen/blob/0d20151ef900df94801f37a79b091c8794059c12/src/app/Components/PageWithSimpleHeader.tsx) and the Palette's `Screen` component to use this new animation/implementation. If we only add it to the Activity screen, we will have an inconsistent header throughout the app, which isn't ideal.

MounirDhahri:
I am on it 🤝

MounirDhahri:
I will close this in favor of this one https://github.com/artsy/palette-mobile/pull/176

PR #9748, opened 2024-01-22
feat(ONYX-619): implement filter pills

MounirDhahri:
This PR resolves [ONYX-619] 

### Description

https://github.com/artsy/eigen/assets/11945712/a449218e-7053-4810-9011-c2018ecf4be8

Blocked by: https://github.com/artsy/palette-mobile/pull/176

### PR Checklist

### To the reviewers 👀

- implement new activity screen filter pills - mounir

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[ONYX-619]: https://artsyproduct.atlassian.net/browse/ONYX-619?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/Activity/ActivityContainer.tsx
+++ b/src/app/Scenes/Activity/ActivityContainer.tsx
@@ -1,0 +1,49 @@
+import { ActivityContainerQuery } from "__generated__/ActivityContainerQuery.graphql"
+import { useMarkNotificationsAsSeen } from "app/Scenes/Activity/hooks/useMarkNotificationsAsSeen"
+import { Suspense } from "react"
+import { graphql, useLazyLoadQuery } from "react-relay"
+import { ActivityList } from "./ActivityList"
+import { ActivityTabPlaceholder } from "./ActivityTabPlaceholder"
+import { NotificationType } from "./types"
+import { getNotificationTypes } from "./utils/getNotificationTypes"
+
+interface ActivityProps {
+  type: NotificationType
+}
+
+export const ActivityContainer: React.FC<ActivityProps> = (props) => {
+  return (
+    <Suspense fallback={<ActivityTabPlaceholder />}>
+      <ActivityContent {...props} />
+    </Suspense>
+  )
+}
+
+export const ActivityContent: React.FC<ActivityProps> = ({ type }) => {
+  const types = getNotificationTypes(type)
+  const queryData = useLazyLoadQuery<ActivityContainerQuery>(
+    activityContainerQuery,
+    {
+      count: 10,
+      types,
+    },
+    {
+      fetchPolicy: "store-and-network",
+    }
+  )
+
+  useMarkNotificationsAsSeen()
+
+  return <ActivityList viewer={queryData.viewer} me={queryData.me} type={type} />
+}
+
+const activityContainerQuery = graphql`
+  query ActivityContainerQuery($count: Int, $after: String, $types: [NotificationTypesEnum]) {
+    viewer {
+      ...ActivityList_viewer @arguments(count: $count, after: $after, types: $types)
+    }
+    me {
+      ...ActivityList_me
+    }
+  }
+`
--- a/src/app/Scenes/Activity/ActivityList.tsx
+++ b/src/app/Scenes/Activity/ActivityList.tsx
@@ -29,7 +29,7 @@ export const ActivityList: React.FC<ActivityListProps> = ({ viewer, type, me }) 
         top: { value: 0 },
       }
     : // Although this breaks the rule of hooks, it's safe to do it here
-      // Because the feature flag is only updates on app start thanks to unsafe_getFeatureFlag
+      // Because the feature flag is only updates on screen mount thanks to unsafe_getFeatureFlag
       // The above check will be removed once the feature flag is removed
       // It seems reasonable here to do this to avoid duplicating code and messing up valuable
       // Git history
--- a/src/app/Scenes/Activity/ActivityList.tsx
+++ b/src/app/Scenes/Activity/ActivityList.tsx
@@ -21,9 +21,11 @@ interface ActivityListProps {
 }
 
 export const ActivityList: React.FC<ActivityListProps> = ({ viewer, type, me }) => {
-  const showPartnerOffersInActivity = unsafe_getFeatureFlag("ARShowPartnerOffersInActivity")
+  const enableNavigateToASingleNotification = unsafe_getFeatureFlag(
+    "AREnableSingleActivityPanelScreen"
+  )
 
-  const headerMeasurements = showPartnerOffersInActivity
+  const headerMeasurements = enableNavigateToASingleNotification
     ? {
         height: { value: 0 },
         top: { value: 0 },
@@ -90,7 +92,9 @@ export const ActivityList: React.FC<ActivityListProps> = ({ viewer, type, me }) 
       headerOffset = -headerMeasurements.height.value
     }
 
-    const ScrollViewComponent = showPartnerOffersInActivity ? Screen.ScrollView : Tabs.ScrollView
+    const ScrollViewComponent = enableNavigateToASingleNotification
+      ? Screen.ScrollView
+      : Tabs.ScrollView
     return (
       <ScrollViewComponent
         refreshControl={<RefreshControl onRefresh={handleRefresh} refreshing={refreshing} />}
@@ -102,7 +106,7 @@ export const ActivityList: React.FC<ActivityListProps> = ({ viewer, type, me }) 
     )
   }
 
-  const FlatlistComponent = showPartnerOffersInActivity ? Screen.FlatList : Tabs.FlatList
+  const FlatlistComponent = enableNavigateToASingleNotification ? Screen.FlatList : Tabs.FlatList
 
   return (
     <FlatlistComponent
--- a/src/app/Scenes/Activity/ActivityList.tsx
+++ b/src/app/Scenes/Activity/ActivityList.tsx
@@ -1,8 +1,8 @@
-import { Flex, Spinner, Separator, Tabs } from "@artsy/palette-mobile"
+import { Flex, Screen, Separator, Spinner, Tabs } from "@artsy/palette-mobile"
 import { ActivityList_me$key } from "__generated__/ActivityList_me.graphql"
 import { ActivityList_viewer$key } from "__generated__/ActivityList_viewer.graphql"
-import { ActivityQuery } from "__generated__/ActivityQuery.graphql"
 
+import { unsafe_getFeatureFlag } from "app/store/GlobalStore"
 import { extractNodes } from "app/utils/extractNodes"
 import { useState } from "react"
 import { RefreshControl } from "react-native"
@@ -21,13 +21,26 @@ interface ActivityListProps {
 }
 
 export const ActivityList: React.FC<ActivityListProps> = ({ viewer, type, me }) => {
-  const headerMeasurements = useHeaderMeasurements()
+  const showPartnerOffersInActivity = unsafe_getFeatureFlag("ARShowPartnerOffersInActivity")
+
+  const headerMeasurements = showPartnerOffersInActivity
+    ? {
+        height: { value: 0 },
+        top: { value: 0 },
+      }
+    : // Although this breaks the rule of hooks, it's safe to do it here
+      // Because the feature flag is only updates on app start thanks to unsafe_getFeatureFlag
+      // The above check will be removed once the feature flag is removed
+      // It seems reasonable here to do this to avoid duplicating code and messing up valuable
+      // Git history
+      // eslint-disable-next-line react-hooks/rules-of-hooks
+      useHeaderMeasurements()
   const [refreshing, setRefreshing] = useState(false)
 
-  const { data, hasNext, isLoadingNext, loadNext, refetch } = usePaginationFragment<
-    ActivityQuery,
-    ActivityList_viewer$key
-  >(notificationsConnectionFragment, viewer)
+  const { data, hasNext, isLoadingNext, loadNext, refetch } = usePaginationFragment(
+    notificationsConnectionFragment,
+    viewer
+  )
 
   const meData = useFragment(meFragment, me)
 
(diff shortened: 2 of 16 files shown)


MounirDhahri:
I will fix the merge conflicts then mark as ready for review

ArtsyOpenSource:

- Cross-platform user-facing changes (implement new activity screen filter pills - mounir)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 1900a5277e53dfe7df25b8f423f0543c099df6ab
</p>

src/app/Scenes/Activity/NewActivityScreen.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+export const NewActivityScreen: React.FC = () => {
+  return (
+    <ActivityScreenStore.Provider>
+      <NewActivityContent />
+    </ActivityScreenStore.Provider>
+  )
+}
+
+export type ActivityType = "all" | "alerts"
```
nickskalkin, line 29: There is a new type of notifications for which pill should be displays - _Offers_. The trick is the pill for offers has to be displayed conditionally - if user has offers.

Technically it is possible to perform such query:

```graphql
notificationsConnection(first: 1, notificationTypes: [PARTNER_OFFER_CREATED]) {
    edges {
      node {
        publishedAt
      }
    }
}
```

and then check if there is an item with `expiresAt` < 1 month ago (backend cleans up older notifications)

nickskalkin (commented):
Looks good! Left one comment about _Offer_ pill

src/app/Scenes/Activity/NewActivityScreen.tsx
nickskalkin: or maybe it is possible to simply use `totalCount`. I've tried with `totalCount` on staging and it showed me the total number of such notifications ever created - but this might be because of a "weird' staging data sync (we don't sync notifications)

src/app/Scenes/Activity/ActivityList.tsx
```diff
@@ -21,13 +21,28 @@ interface ActivityListProps {
...
+  const headerMeasurements = enableNavigateToASingleNotification
+    ? {
+        height: { value: 0 },
+        top: { value: 0 },
+      }
+    : // Although this breaks the rule of hooks, it's safe to do it here
+      // Because the feature flag is only updates on screen mount thanks to unsafe_getFeatureFlag
+      // The above check will be removed once the feature flag is removed
+      // It seems reasonable here to do this to avoid duplicating code and messing up valuable
+      // Git history
```
dariakoko, line 37: 👀

src/app/Scenes/Activity/NewActivityScreen.tsx
```diff
@@ -0,0 +1,40 @@
...
+  type: ActivityType
+  setType: Action<this, ActivityType>
+}
+
+export const ActivityScreenStore = createContextStore<ActivityScreenStoreModel>({
+  type: "all",
+  setType: action((state, payload) => {
+    state.type = payload
+  }),
+})
```
dariakoko, line 40: **minor:** What do you think about moving store operations to a separate file?

src/app/Scenes/Activity/ActivityContainer.tsx
```diff
@@ -0,0 +1,49 @@
...
+export const ActivityContent: React.FC<ActivityProps> = ({ type }) => {
+  const types = getNotificationTypes(type)
+  const queryData = useLazyLoadQuery<ActivityContainerQuery>(
+    activityContainerQuery,
+    {
+      count: 10,
+      types,
+    },
+    {
+      fetchPolicy: "store-and-network",
```
dariakoko, line 31: **question:** why don't we use the default `store-or-network` fetchPolicy here?

src/app/Scenes/Activity/components/NewActivityHeader.tsx
```diff
@@ -0,0 +1,25 @@
...
+    <Flex px={2} pb={2}>
+      <Flex flexDirection="row">
+        <Pill variant="default" selected={type === "all"} mr={0.5} onPress={() => setType("all")}>
+          All
+        </Pill>
+        <Pill
+          variant="default"
+          mr={0.5}
+          selected={type === "alerts"}
+          onPress={() => setType("alerts")}
```
dariakoko, line 18: **minor:** Let's export constant values to reuse them in other places and avoid issues resulted by typos?
```
export alertsType = "alerts"
export allType = "all"
```
Or do what @tam-kis did [here](https://github.com/artsy/force/pull/13392/files#diff-5fd1a0cbb6ba35e2cc8301c83c4d66968e0bfb67bad6b5c22f5e5f1b85dcfaa3R7-R11)

dariakoko (dismissed):
Thank you for improving the tests and adding comments to the new pieces of code!
Looks good ⭐ 
I only left several minot comments

src/app/Scenes/Activity/ActivityList.tsx
dariakoko: Line 37

src/app/Scenes/Activity/ActivityList.tsx
```diff
@@ -21,13 +21,28 @@ interface ActivityListProps {
...
+      // eslint-disable-next-line react-hooks/rules-of-hooks
+      useHeaderMeasurements()
   const [refreshing, setRefreshing] = useState(false)
 
-  const { data, hasNext, isLoadingNext, loadNext, refetch } = usePaginationFragment<
-    ActivityQuery,
-    ActivityList_viewer$key
-  >(notificationsConnectionFragment, viewer)
+  const { data, hasNext, isLoadingNext, loadNext, refetch } = usePaginationFragment(
```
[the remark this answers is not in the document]
MounirDhahri: yes, the last relay updates interpreted the types properly 🎉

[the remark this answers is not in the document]
gkartalis: oh so we don't need to add the types anymore there?

src/app/Scenes/Activity/ActivityList.tsx
```diff
@@ -21,13 +21,28 @@ interface ActivityListProps {
...
+        height: { value: 0 },
+        top: { value: 0 },
+      }
+    : // Although this breaks the rule of hooks, it's safe to do it here
+      // Because the feature flag is only updates on screen mount thanks to unsafe_getFeatureFlag
+      // The above check will be removed once the feature flag is removed
+      // It seems reasonable here to do this to avoid duplicating code and messing up valuable
+      // Git history
+      // eslint-disable-next-line react-hooks/rules-of-hooks
+      useHeaderMeasurements()
```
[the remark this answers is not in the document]
MounirDhahri: thanks for the feedback here. Unfortunately this won't solve the issue I am trying to fix here and which is not calling `useHeaderMeasurements`. The latters calls some values from tabs context that are not available when in the new design which is not tab based.

src/app/Scenes/Activity/ActivityList.tsx
MounirDhahri: pardon me, I didn't get what you suggested here. should I move it to the line above?

src/app/Scenes/Activity/ActivityContainer.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+export const ActivityContainer: React.FC<ActivityProps> = (props) => {
+  return (
+    <Suspense fallback={<ActivityTabPlaceholder />}>
+      <ActivityContent {...props} />
+    </Suspense>
+  )
+}
+
+export const ActivityContent: React.FC<ActivityProps> = ({ type }) => {
```
olerichter00, line 22: This export might not be needed, but also ok to leave it.
```suggestion
const ActivityContent: React.FC<ActivityProps> = ({ type }) => {
```

src/app/Scenes/Activity/NewActivityScreen.tsx
```diff
@@ -0,0 +1,40 @@
...
+  return (
+    <Screen>
+      <Screen.AnimatedHeader onBack={goBack} title="Activity" />
+      <Screen.StickySubHeader title="Activity">
+        <NewActivityHeader />
+      </Screen.StickySubHeader>
+      <ActivityContainer type={type} />
+    </Screen>
+  )
+}
```
olerichter00, line 19: What's the thinking behind extracting this into its own component? I wonder if it would be easier to read if we just had one component (removing `NewActivityContent` and just having `NewActivityScreen` in this file).

olerichter00 (dismissed):
Looks great 👍

src/app/Scenes/Activity/NewActivityScreen.tsx
```diff
@@ -0,0 +1,40 @@
+import { Screen } from "@artsy/palette-mobile"
+import { ActivityContainer } from "app/Scenes/Activity/ActivityContainer"
+import { NewActivityHeader } from "app/Scenes/Activity/components/NewActivityHeader"
+import { goBack } from "app/system/navigation/navigate"
+import { Action, action, createContextStore } from "easy-peasy"
+
+export const NewActivityContent: React.FC = () => {
```
damassi, line 7: This is probably already noted, but we should add a comment to remove the `New` in the name since there's a handful of (old) new components littered around.

MounirDhahri: good catch - i will

src/app/Scenes/Activity/NewActivityScreen.tsx
MounirDhahri: i can add that as well since we're going to start sending offers

src/app/Scenes/Activity/ActivityList.tsx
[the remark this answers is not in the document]
MounirDhahri: yes, they worked without the types

src/app/Scenes/Activity/ActivityList.tsx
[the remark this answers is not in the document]
gkartalis: ooooh got what you mean ~ sure then lets do it like this

src/app/Scenes/Activity/ActivityContainer.tsx
MounirDhahri: Usually, I would do as you suggest since we probably already made the query on the home screen in case the user opened the app normally (not using a deep link).In this case though, because we uncovered some caching issues during the QA session and have no short-term plans of fixing those caching issues, I took the short safe path of getting the data from the store, while also making the network request

src/app/Scenes/Activity/NewActivityScreen.tsx
MounirDhahri: my main reason here is to be able to use `const type = ActivityScreenStore.useStoreState((state) => state.type)` because the store will be ready by then. Let me see if I can do this differently ;)

MounirDhahri: the other way to avoid this, would be to move the hook to the `ActivityContainer` but I dont' want to do that (for now), since the same component is used in the existing screen and is a expecting the `type` prop to be injected there.  
The other alternative would be to have the provider inside `Activity.tsx` which also didn't seem like a better option for me. I am interested in hearing if you have thoughts about that.

src/app/Scenes/Activity/NewActivityScreen.tsx
MounirDhahri: small note to make sure we're on the same page, i believe the pill should be displayed but disabled as we do on the `Search` tab. please let me know if I didn't get this properly from design.

src/app/Scenes/Activity/NewActivityScreen.tsx
olerichter00: Hmm, but wouldn't it be good to remove the 'type' prop from 'ActivityContainer' and read the type from the store within 'ActivityContainer' instead? This way, we would avoid prop drilling, and it should be possible to read from the store within the component, right?

src/app/Scenes/Activity/NewActivityScreen.tsx
nickskalkin: I think it should be hidden (not disabled) when there are not offers. We had a call with Barney awhile ago and he said it should be hidden (he even pointed me to a comment in figma, but I cannot find it anymore...). Or have you seen a spec with the disabled pill state?

src/app/Scenes/Activity/components/NewActivityHeader.tsx
MounirDhahri: I understand the rationale for exporting the values as constants, I don't believe though this will not add much value here. The reason behind that is that `setType` arguments are typed
Screenshot 2024-01-23 at 10 40 28

src/app/Scenes/Activity/NewActivityScreen.tsx
MounirDhahri: yes, that I believe will be the first thing we will need to do when we do the feature flag cleanup. I didn't do that now because `ActivityContainer` is used by the old layout which is not wrapped by the store. Do you suggest I wrap the old screen with a provider and do the refactor? I didn't do that because I wanted to keep the logic the same and not make this PR bigger. what do you think about it?

src/app/Scenes/Activity/ActivityList.tsx
MounirDhahri: maybe I should have emphasized this differently, actually `Git history` here was a factor in taking this path. I kept this file the same to avoid creating a new file where I am the author and there would be no straightforward way for me and other devs in the future to take advantage of quick git blame to get addition context from original PRs

dariakoko: Get it now, thank you for explaining!

MounirDhahri: with pleasure, anytime of course

src/app/Scenes/Activity/NewActivityScreen.tsx
MounirDhahri: I will tag you in the slack thread

src/app/Scenes/Activity/NewActivityScreen.tsx
olerichter00: Thanks for the explanation. Sorry, I hadn't read the code thoroughly enough. Let's do it when we clean up the old code. 👍

olerichter00 (dismissed):

src/app/Scenes/Activity/NewActivityScreen.tsx
MounirDhahri: no need to apologize, I can explain my rationale with pleasure <3

PR #9832, opened 2024-02-09
feat: Improve Activity screen performance

MounirDhahri:
This PR resolves [ONYX-738] 

### Description

This PR makes some updates to the activity screen in order to improve it's performance. These updates include:
- Using memo for activity item components
- Setting a header height to avoid the layout calculation and jumps
- Lazy load images thanks to using `LazyFlatlist` component from `palette-mobile`.
- Setting Flex height and width in image containers to avoid layout calculation 
 
Group 4 (1)

**I recommend to hide white spaces when reviewing this**

https://github.com/artsy/eigen/assets/11945712/18e0574f-20ab-4249-a0fc-b00923d6818b

### PR Checklist

### To the reviewers 👀

- Improve Activity screen performance - mounir

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[ONYX-738]: https://artsyproduct.atlassian.net/browse/ONYX-738?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/Activity/ActivityItem.tsx
+++ b/src/app/Scenes/Activity/ActivityItem.tsx
@@ -12,6 +12,7 @@ import { navigateToActivityItem } from "app/Scenes/Activity/utils/navigateToActi
 import { navigate } from "app/system/navigation/navigate"
 import { extractNodes } from "app/utils/extractNodes"
 import { useFeatureFlag } from "app/utils/hooks/useFeatureFlag"
+import { memo } from "react"
 import { TouchableOpacity } from "react-native"
 import { graphql, useFragment } from "react-relay"
 import { useTracking } from "react-tracking"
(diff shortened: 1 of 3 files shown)


src/app/utils/hooks/useHasSeenItem.ts
```diff
@@ -0,0 +1,37 @@
+import { useRef, useState } from "react"
+import { ViewToken, ViewabilityConfig } from "react-native"
+
+// This is a custom hook that is used to determine if a user has seen an item in a list or not
+export const useHasSeenItem = <T>({ keyExtractor }: { keyExtractor: (item: T) => string }) => {
```
MounirDhahri, line 5: this is the hook that we. can use

src/app/Scenes/Activity/ActivityList.tsx
```diff
@@ -45,6 +46,12 @@ export const ActivityList: React.FC<ActivityListProps> = ({ viewer, type, me })
     return true
   })
 
+  const { viewabilityConfig, onViewableItemsChanged, hasSeenItem, viewedItems } = useHasSeenItem<
```
MounirDhahri, line 49: this is the only change I made, the rest is all basic react improvements that unfortunately didn't add any noticeable performance improvement

src/app/Scenes/Activity/ActivityItem.tsx
```diff
@@ -20,108 +21,181 @@ import { isArtworksBasedNotification } from "./utils/isArtworksBasedNotification
...
+
+            <Flex flexDirection="row" alignItems="center">
+              {artworks.map((artwork) => {
+                return (
+                  <Flex
+                    key={`${item.internalID}-${artwork.internalID}`}
+                    mr={1}
+                    accessibilityLabel="Activity Artwork Image"
+                    height={ARTWORK_IMAGE_SIZE}
+                    width={ARTWORK_IMAGE_SIZE}
```
MounirDhahri, line 178: making this change to give react a break from calculating how big this View is going to be

src/app/Scenes/Activity/ActivityList.tsx
```diff
@@ -94,7 +110,7 @@ export const ActivityList: React.FC<ActivityListProps> = ({ viewer, type, me })
         )
       }
       onEndReached={handleLoadMore}
-      renderItem={({ item }) => <ActivityItem notification={item} />}
+      renderItem={renderItem}
```
damassi, line 113: Wow, the viewableItemsConfig has a lot of features! RN seems to have thought of everything. 

I wonder if we might be able to create a wrapper API that would require less wiring for devs, but using all of the parts you set up above? This might be a great use-case for renderProps:

```tsx
<LazyFlatList maybeSomeConfig={foo}>
  {({ viewabilityConfig, onViewableItemsChanged, hasSeenItem, viewedItems }) => {
    return (
      <FlatList 
        viewabilityConfig={viewabilityConfig}
        onViewableItemsChanged={onViewableItemsChanged}
        renderItem={({ item }) => {
          return <SomeItem isVisible={hasSeenItem(item) />
        }}
      />
    )
  }}}
/>
```

And then from here, everything can be automatically setup from within `LazyFlatList` and devs don't have to worry about any of it - it becomes more a matter of wrapping the flat list and then opting in, with minimal changes to the original implementation.  

Then in the future, we can create a set of LazyFoo components (ie, `LazyMasonry`) and raise the shared stuff up a level via composition.

damassi: Either that or a straight up `LazyFlatList` component that is exactly like `FlatList` and merges props in, which we could add to `palette-mobile`. That could be even simpler.

MounirDhahri: Nice idea, this is a good addition and an adequate usage for the render props design pattern. I will make the change.

damassi (approved):

PR #9836, opened 2024-02-14
chore: add animations playground

MounirDhahri:
### Description

This PR adds docs on learning animations in Eigen and updates our docs to link to it. It also makes it possible to navigate quickly to it from the Dev Settings menu

#nochangelog


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  ✅ **No changelog changes**
  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against b661e235f94ced2e2005f92adef0f19084b7b29c
</p>

src/app/Scenes/Home/HomeContainer.tsx
```diff
@@ -1,11 +1,14 @@
 import { HomeQueryRenderer } from "app/Scenes/Home/Home"
+import { Playground } from "app/Scenes/Playground/Playground"
```
MounirDhahri, line 2: thinking of maybe lazy importing this in the future in case we add more stuff to it to avoid affecting the bundle size, I don't think this is worth doing now.

damassi: Its tiny, and lib is already in, so nbd!

src/app/Scenes/Playground/Animation/L0-Reanimated/index.tsx
```diff
@@ -0,0 +1,43 @@
+import { Flex, Join, Separator, Text } from "@artsy/palette-mobile"
+import { ScrollView } from "react-native-gesture-handler"
+
+export default function App() {
```
damassi, line 4: This is minor and non-blocking, but in general we should always avoid default exports because that leads to 1:2 potential; eg:

```tsx
import Foo from 'path/to/bar'
```
vs
```tsx
import { Foo } from 'path/to/bar' // leads to error
```
while 
```
import { Bar } from 'path/to/bar'
```
can only be 1:2'd if one does 
```
import { Bar as Foo } from 'path/to/bar'
```
which doesn't happen often, and requires extra steps. With named exports everything is in the static import map graph and leads to a much more locked-down codebase.

damassi (approved):
Thanks for adding 👍 Now we've got a handy lil reference

src/app/Scenes/Playground/Animation/L0-Reanimated/index.tsx
MounirDhahri: forgot to rename this - will update it

PR #9922, opened 2024-03-02
fix(ONYX-822): fix broken ArtsyWebView paddings

dariakoko:
This PR resolves [ONYX-822] 

### Description

The ArtsyWebView uses SafeAreaView. SafeAreaView adds paddings to the top and bottom of the page to make sure the active area of the screen can be assessed. We do not need this padding if we are not presenting the screen modally. When we do not the screen modally we can use a react-native View instead of SafeAreaView

| Before iOS | After iOS | Before Android | After Android |
| --- | --- | --- | --- |
| image | image | Screenshot_1709343815 | Screenshot_1709344209 |

### PR Checklist

### To the reviewers 👀

- in ArtsyWebView use SafeAreaView only if presented modally - daria

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[ONYX-822]: https://artsyproduct.atlassian.net/browse/ONYX-822?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/routes.ts
+++ b/src/app/routes.ts
@@ -271,7 +271,6 @@ function getDomainMap(): Record<string, RouteMatcher[] | null> {
       alwaysPresentModally: true,
       safeAreaEdges: ["bottom"],
     }),
-    addWebViewRoute("/artists"),
     addWebViewRoute("/buy-now-feature-faq"),
     addWebViewRoute("/buyer-guarantee"),
     addWebViewRoute("/categories"),
@@ -280,7 +279,6 @@ function getDomainMap(): Record<string, RouteMatcher[] | null> {
       safeAreaEdges: ["bottom"],
     }),
     addWebViewRoute("/identity-verification-faq"),
-    addWebViewRoute("/galleries"),
     addWebViewRoute("/meet-the-specialists"),
     addWebViewRoute("/orders/:orderID", {
       mimicBrowserBackButton: true,
--- a/src/app/routes.ts
+++ b/src/app/routes.ts
@@ -271,6 +271,7 @@ function getDomainMap(): Record<string, RouteMatcher[] | null> {
       alwaysPresentModally: true,
       safeAreaEdges: ["bottom"],
     }),
+    addWebViewRoute("/artists"),
     addWebViewRoute("/buy-now-feature-faq"),
     addWebViewRoute("/buyer-guarantee"),
     addWebViewRoute("/categories"),
@@ -279,6 +280,7 @@ function getDomainMap(): Record<string, RouteMatcher[] | null> {
       safeAreaEdges: ["bottom"],
     }),
     addWebViewRoute("/identity-verification-faq"),
+    addWebViewRoute("/galleries"),
     addWebViewRoute("/meet-the-specialists"),
     addWebViewRoute("/orders/:orderID", {
       mimicBrowserBackButton: true,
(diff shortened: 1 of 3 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (in ArtsyWebView use SafeAreaView only if presented modally - daria)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against e791ce5c64a0cc2fb73f29379baa6f1d78fee665
</p>

src/app/routes.ts
```diff
@@ -279,6 +280,7 @@ function getDomainMap(): Record<string, RouteMatcher[] | null> {
       safeAreaEdges: ["bottom"],
     }),
     addWebViewRoute("/identity-verification-faq"),
+    addWebViewRoute("/galleries"),
```
MounirDhahri, line 283: Nice idea on adding the routes here, but this leads me to think the issue affects any route that is not here. I would therefore look into another approach to get the same behaviour that we get from `addWebViewRoute` to all our deep urls.

MounirDhahri (changes requested):
This is great, I don't think though we should be merging this at its current state since I find this as an opportunity to fix other similar URL routes that we haven't included in `routes.ts` - especially because now we know what is broken

olerichter00 (commented):
Thanks a lot for the fix 🙏 

Although this fix works with all routes that are added with `addWebViewRoute("/artists")`, routes that are not added like this would be interpreted as vanity URLs (e.g., `/shows`), which results in the Back button being rendered on top of the page title. I've created https://github.com/artsy/eigen/pull/9924 as an attempt to solve this issue as well.

src/app/routes.ts
olerichter00: https://github.com/artsy/eigen/pull/9924 would fix the issue by using [ArtsyWebViewPage](https://github.com/artsy/eigen/blob/4a10245a845a6059bebf89a295724268d13ec486/src/app/Components/ArtsyWebView.tsx#L51) for all URLs that down't mach a specific route.

MounirDhahri:
We thought about the same thing @olerichter00 -  Looking into your PR now

src/app/routes.ts
olerichter00: Please note that my PR would be a fix on top of this PR because this change is still needed: https://github.com/artsy/eigen/pull/9922/files#diff-edc2fa1c2b97a3ddeaf270d241e4054ba768be253b35b4e75e63914b6b1c20cdR257

olerichter00: It works without adding the routes but it seems that https://github.com/artsy/eigen/pull/9922/files#diff-edc2fa1c2b97a3ddeaf270d241e4054ba768be253b35b4e75e63914b6b1c20cdR257 is needed, no?

MounirDhahri: perfect, do. you mind if you don't commit the route changes then and leave that before merging

olerichter00: Good idea not to commit the route changes. @dariakoko I'd say, let's merge https://github.com/artsy/eigen/pull/9924 into this PR, then remove the route changes, test again, and then merge into main.

dariakoko: Good idea, doing it now

dariakoko:
Addressed your comments @MounirDhahri @olerichter00 and tested - looking good!
image

olerichter00 (approved):
🌟

MounirDhahri (approved):
Perfect - our webviews across the app will be much better now!

MounirDhahri:
Will merge this to get it to the release

PR #9923, opened 2024-03-02
fix: enable scroll only when there is more than one elements in the a…

dariakoko:
This PR resolves [] 

### PR Checklist

Enable scroll on the activity screen only when there is more than one elements in the array

### To the reviewers 👀

- enable scroll on the activity screen only when there is more than one elements in the array - daria

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Scenes/Activity/ActivityEmptyView.tsx
+++ b/src/app/Scenes/Activity/ActivityEmptyView.tsx
@@ -1,10 +1,12 @@
 import { Flex, Spacer, Text, Touchable } from "@artsy/palette-mobile"
 import { navigate } from "app/system/navigation/navigate"
 import { useFeatureFlag } from "app/utils/hooks/useFeatureFlag"
+import { ScrollView } from "react-native-gesture-handler"
 import { NotificationType } from "./types"
 
 interface ActivityEmptyViewProps {
   type: NotificationType
+  refreshControl?: React.ReactElement
 }
 
 const artistLink = (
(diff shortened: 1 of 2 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (enable scroll on the activity screen only when there is more than one elements in the array - daria)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against b484b97b5b2e82cdda1c242b982a70256ac4ebca
</p>

olerichter00 (commented):
@MounirDhahri, this bug might be related to the new [Screen.StickySubheader](https://github.com/artsy/eigen/blob/e7dc477974216318f0f9376134e51790cd20954a/src/app/Scenes/Activity/NewActivityScreen.tsx#L54) component. It seems like the sub header is collapsing when scrolling and then abruptly jumping back. I'm fine with addressing this bug as proposed in this PR, but we could see the same issue later with a similar layout: Screen header + FlatlistComponent and content that doesn't fill the entire screen. Do you have any ideas on how we could resolve the root issue instead of just disabling scrolling?

MounirDhahri:
Thanks Daria for the fix but I agree with @olerichter00 I think this should be fixed in `palette`. I will work on it today

MounirDhahri:
I am still trying to come up with a good fix for this but with no success - i will push a temporary fix here since this PR breaks the refresh control

PR #9933, opened 2024-03-04
fix: parsing of Google Places API details request

anandaroop:
This PR resolves [ONYX-812]

### Description

The Sell With Artsy intake form was observed to be preventing users from submitting an artwork because of an invalid location, even when the location _appeared_ to be a valid result from Google.

This has to do with how we parse the response that we get from `https://maps.googleapis.com/maps/api/place/details`

After the fix, the previously failing **Sydney** example now works:

https://github.com/artsy/eigen/assets/140521/7968e187-fc99-4bd2-a28a-a2cbedb7a54d

### PR Checklist

### To the reviewers 👀

- Improve reliability of Location input field - anandaroop

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[ONYX-812]: https://artsyproduct.atlassian.net/browse/ONYX-812?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/utils/googleMaps.tests.ts
+++ b/src/app/utils/googleMaps.tests.ts
@@ -1,8 +1,8 @@
 import { fetchMockResponseOnce } from "app/utils/tests/fetchMockHelpers"
-import { getLocationPredictions } from "./googleMaps"
+import { getLocationPredictions, getLocationDetails } from "./googleMaps"
 
-describe("queryLocation()", () => {
-  it("queries the google maps api and returns some formatted results", async () => {
+describe(getLocationPredictions, () => {
+  it("queries the google maps autocomplete api and returns some formatted results", async () => {
     fetchMockResponseOnce(
       JSON.stringify({
         predictions: [
@@ -24,3 +24,56 @@ describe("queryLocation()", () => {
     ])
   })
 })
+
+describe(getLocationDetails, () => {
+  it("queries the google maps place details api and returns some formatted results", async () => {
+    fetchMockResponseOnce(JSON.stringify(placeDetailsResponse))
+
+    const result = await getLocationDetails({
+      id: "sydney_place_id_4242",
+      name: "Sydney",
+    })
+
+    expect(result).toEqual({
+      city: "Sydney",
+      coordinates: [-33.8688197, 151.2092955],
+      country: "Australia",
+      countryCode: "AU",
+      id: "sydney_place_id_4242",
+      name: "Sydney",
+      postalCode: undefined,
+      state: "New South Wales",
+      stateCode: "NSW",
+    })
+  })
+})
+
+const placeDetailsResponse = {
+  result: {
+    place_id: "sydney_place_id_4242",
+    name: "Sydney",
+    address_components: [
+      {
+        long_name: "Sydney",
+        short_name: "Sydney",
+        types: ["colloquial_area", "locality", "political"],
+      },
+      {
+        long_name: "New South Wales",
+        short_name: "NSW",
+        types: ["administrative_area_level_1", "political"],
+      },
+      {
+        long_name: "Australia",
+        short_name: "AU",
+        types: ["country", "political"],
+      },
+    ],
+    geometry: {
+      location: {
+        lat: -33.8688197,
+        lng: 151.2092955,
+      },
+    },
+  },
+}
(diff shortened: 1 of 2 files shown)


src/app/utils/googleMaps.tests.ts
```diff
@@ -1,8 +1,8 @@
 import { fetchMockResponseOnce } from "app/utils/tests/fetchMockHelpers"
-import { getLocationPredictions } from "./googleMaps"
+import { getLocationPredictions, getLocationDetails } from "./googleMaps"
 
-describe("queryLocation()", () => {
-  it("queries the google maps api and returns some formatted results", async () => {
+describe(getLocationPredictions, () => {
+  it("queries the google maps autocomplete api and returns some formatted results", async () => {
```
anandaroop, line 5: Just updating an existing test here — note how the describe block got out of sync with the function name. This is why I tend to use the function name's symbol itself in the describe block, rather than a string version of it which can get out of sync. (Please remind me if there are any objections to this pattern in Eigen.)

ArtsyOpenSource:

- Cross-platform user-facing changes (Improve reliability of Location input field - anandaroop)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against f22c3a0d958daf0bee10d2a1ef7fd47b0a7c17db
</p>

src/app/utils/googleMaps.ts
```diff
@@ -71,12 +71,12 @@ export const getLocationDetails = async ({
...
-  const statePlace = address_components.find(
-    (comp) => comp.types[0] === "administrative_area_level_1"
+  const cityPlace = address_components.find((comp) => comp.types.includes("locality"))
+  const statePlace = address_components.find((comp) =>
+    comp.types.includes("administrative_area_level_1")
   )
-  const countryPlace = address_components.find((comp) => comp.types[0] === "country")
-  const postalCodePlace = address_components.find((comp) => comp.types[0] === "postal_code")
+  const countryPlace = address_components.find((comp) => comp.types.includes("country"))
+  const postalCodePlace = address_components.find((comp) => comp.types.includes("postal_code"))
```
anandaroop, line 79: This is the crux of the issue — we were assuming that the address component type that we are interested in was at position 0 of the array in the response from Google, but that assumption does not always hold.

Some cities will have a less standard location hierarchy and we may find a less common type at position 0, as you can see in the (realistic) mock data added in the test here.

For comparison, note that Gravity does similar processing of Google place results, but it does it via a dedicated gem called Geocoder. You can see how Geocoder avoids the assumption that the type we are interested in is at position 0, and instead just checks for inclusion anywhere in the array:

https://github.com/alexreisner/geocoder/blob/c41b9065ae85907fbc33c11bff3ed4eee1d2f36d/lib/geocoder/results/google.rb#L112-L114

This PR essentially makes Eigen's parsing of Google data consistent with that^

src/app/utils/googleMaps.tests.ts
```diff
@@ -24,3 +24,56 @@ describe("queryLocation()", () => {
...
+const placeDetailsResponse = {
+  result: {
+    place_id: "sydney_place_id_4242",
+    name: "Sydney",
+    address_components: [
+      {
+        long_name: "Sydney",
+        short_name: "Sydney",
+        types: ["colloquial_area", "locality", "political"],
+      },
```
anandaroop, line 60: This mock reflects the actual data we get back from Google for the **Sydney** request.

Note that `locality` is the _second_, not first, item in this `result.address_components.types` entry

MounirDhahri (approved):
Awesome! great choice to go with this. It is the one with the highest impact 👏 Also thanks for the regression test

src/app/utils/googleMaps.ts
olerichter00: I've just checked [the code in Force](https://github.com/artsy/force/blob/61292f3af5eedc82002aab5098c46d7b40459b46/src/Components/LocationAutocompleteInput.tsx#L208), and it seems to be consistent with your fix. 🎉

olerichter00 (approved):
Thanks for the fix 🙏

PR #10039, opened 2024-04-05
fix: photo picker crash

brainbicycle:
This PR resolves [] 

### Description

our patch for finding the root window in the photo stopped working reliably, this updates the window finding to a more robust method. 

It is still not totally clear why / under what conditions this doesn't work, it might be an iOS update or another dep update that broke it. 

### Follow-ups:
- There are issues in the upstream repo about windows not being found? Maybe we could contribute this back if this is the root of the problem?

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

- Fix photo picker crash - Brian + George 

-

#### Dev changes

-

</details>

### DIFF
--- a/patches/react-native-image-crop-picker+0.38.1.patch
+++ b/patches/react-native-image-crop-picker+0.38.1.patch
@@ -1,8 +1,8 @@
 diff --git a/node_modules/react-native-image-crop-picker/ios/src/ImageCropPicker.m b/node_modules/react-native-image-crop-picker/ios/src/ImageCropPicker.m
-index 3b867d1..6a19214 100644
+index 7101410..2fd5436 100644
 --- a/node_modules/react-native-image-crop-picker/ios/src/ImageCropPicker.m
 +++ b/node_modules/react-native-image-crop-picker/ios/src/ImageCropPicker.m
-@@ -123,8 +123,19 @@ - (void) setConfiguration:(NSDictionary *)options
+@@ -124,8 +124,21 @@ RCT_EXPORT_MODULE();
      }
  }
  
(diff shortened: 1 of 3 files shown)


patches/react-native-image-crop-picker+0.38.1.patch
```diff
@@ -1,23 +1,25 @@
...
 -- (UIViewController*) getRootVC {
 -    UIViewController *root = [[[[UIApplication sharedApplication] delegate] window] rootViewController];
 +- (UIViewController *) getRootVC {
 +    UIWindow *presentingWindow;
-+    for (UIWindow *window in [[UIApplication sharedApplication] windows]) {
-+        if ([window isKindOfClass:NSClassFromString(@"ARWindow")]) {
-+            presentingWindow = window;
++    if (!presentingWindow) {
++        for (UIWindow *window in [[UIApplication sharedApplication] windows]) {
++            if (![window isHidden] && [window windowLevel] == UIWindowLevelNormal) {
```
brainbicycle, line 15: better check that the window is not hidden and is at level normal, rather than checking our weird subclass that doesn't seem to work reliably

patches/react-native-image-crop-picker+0.38.1.patch
```diff
@@ -1,23 +1,25 @@
...
++                presentingWindow = window;
++            }
 +        }
 +    }
++
 +    if (!presentingWindow) {
-+        NSAssert(false, @"Unable to find ARWindow did it get renamed?");
```
brainbicycle, line 19: the assert here was causing the crash in dev mode, we don't really need it, if this stops working photo picker stops working

gkartalis (dismissed):

patches/react-native-image-crop-picker+0.38.1.patch
```diff
@@ -1,23 +1,25 @@
...
  }
  
 -- (UIViewController*) getRootVC {
 -    UIViewController *root = [[[[UIApplication sharedApplication] delegate] window] rootViewController];
 +- (UIViewController *) getRootVC {
 +    UIWindow *presentingWindow;
-+    for (UIWindow *window in [[UIApplication sharedApplication] windows]) {
-+        if ([window isKindOfClass:NSClassFromString(@"ARWindow")]) {
-+            presentingWindow = window;
++    if (!presentingWindow) {
```
brainbicycle, line 13: this if is unnecessary will always be nil at this point

PR #10072, opened 2024-04-15
fix: My Collection artist keyword search

olerichter00:
This PR resolves [ONYX-895] 

### Description

My Collection artist keyword search.

https://github.com/artsy/eigen/assets/4691889/91c47398-9448-4a58-b35d-ff402ac65e18

### PR Checklist

### To the reviewers 👀

- Fix My Collection artist keyword search by fetching all artists when searching - ole

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[ONYX-895]: https://artsyproduct.atlassian.net/browse/ONYX-895?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/MyCollection/Components/MyCollectionCollectedArtistsRail.tsx
+++ b/src/app/Scenes/MyCollection/Components/MyCollectionCollectedArtistsRail.tsx
@@ -160,7 +160,7 @@ export const Artist: React.FC<{
 
 const collectedArtistsPaginationFragment = graphql`
   fragment MyCollectionCollectedArtistsRail_me on Me
-  @argumentDefinitions(count: { type: "Int", defaultValue: 10 }, after: { type: "String" })
+  @argumentDefinitions(count: { type: "Int", defaultValue: 20 }, after: { type: "String" })
   @refetchable(queryName: "MyCollectionCollectedArtistsRail_myCollectionInfoRefetch") {
     userInterestsConnection(
       first: $count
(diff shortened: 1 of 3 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (Fix My Collection artist keyword search by prefetching all artists when searching - ole)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 3f375e4f0c61eaa28cfc0aebc14a92b1e88b0149
</p>

MounirDhahri (commented):
Thanks for the fix - last time we did the same thing for the artworks, I was personally against this, but we ended up doing it because of time constraints, we eventually found out later it wasn't as much work as we expected it to be (doing it in backend) and later we ended up building on top of that by adding local filters for my collection artworks (more tech debt).

I feel like the same thing applies here and I find this is an opportunity to adapt the logic and support keyword filtering in MP to use it in Force without fetching all artists.

olerichter00:
@MounirDhahri, yes, backend filtering would be ideal! There is some more context about what we would have to do in this [Slack thread](https://artsy.slack.com/archives/C05EQL4R5N0/p1712923228479259) and in this[ tech plan](https://www.notion.so/artsy/My-Collection-Artwork-Filter-Sorting-51cf7140ed164c489b7ca00c8db4bb65?pvs=4).  

My thinking is that it wouldn't be worth it to put so much effort into implementing the filtering in the backend in this case because this feature is probably not used very often. However, I agree that we could benefit from implementing support for collecting artworks because we could then eliminate the client-side filtering you've mentioned.

So, IMHO we should merge this PR anyway because it's an easy fix for the keyword filter that doesn't work at the moment with more than 10 artists. At the same time, let's bring up this project regarding the filter support for collected artists and discuss it with the team 👍

MounirDhahri:
@olerichter00 sounds good to me - thanks for sharing the tech plan, it's been so long I forgot about the existence of that tech plan

PR #10085, opened 2024-04-16
feat(EMI-1766): Show partner offer details on Artwork screeen

MrSltun:
This PR resolves [EMI-1766] 

> [!NOTE]
> This work is hidden behind `AREnablePartnerOfferV1Improvements` feature flag

### Description
This PR adds the partner offer details in the artwork screen
It also removes the `Exclude shipping and taxes` label from the artwork screen
And refactors a few tests 

#### Removing excl. shipping and taxes label

| | Before | After |
|---|---|---|
| Android | image | image |
| iOS | image | image |

#### New Partner Offer details statuses

| | Android | iOS |
|---|---|---|
| Exact Price | image | image |
| Expiring soon + range | image | image |
| Hidden price | image | image |
| Large price display | image | image |

### PR Checklist

- [X] I have tested my changes on **iOS** and **Android**.
- [x] I hid my changes behind a **[feature flag]**, or they don't need one.
- [X] I have included **screenshots** or **videos**, or I have not changed the UI.
- [X] I have added **tests**, or my changes don't require any.
- [X] I added an **[app state migration]**, or my changes do not require one.
- [X] I have documented any **follow-up work** that this PR will require, or it does not require any.
- [X] I have added a **changelog entry** below, or my changes do not require one.

### To the reviewers 👀

- [ ] I would like **at least one** of the reviewers to **run** this PR on the simulator or device.

<details><summary>Changelog updates</summary>

#### Cross-platform user-facing changes

- Adds partner offer details to artwork screen (behind FF: AREnablePartnerOfferV1Improvements) - mrsltun

#### iOS user-facing changes

-

#### Android user-facing changes

-

#### Dev changes

-

@artsy/emerald-devs 

</details>

Need help with something? Have a look at our [docs], or get in touch with us.

[app state migration]: ../blob/main/docs/adding_state_migrations.md
[feature flag]: ../blob/main/docs/developing_a_feature.md
[docs]: ../blob/main/docs/README.md

[EMI-1766]: https://artsyproduct.atlassian.net/browse/EMI-1766?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #10087, opened 2024-04-16
feat: [EMI-1753] artwork screen partner offer button behavior

erikdstock:
This PR resolves [EMI-1753] 

### Description
This PR changes the behavior of the purchase button when a partner offer is available. It was a little tricky because the button is also used in our partner offer notification panel.

I started out by following the patterns we used in artsy/force#13647 and its followup PRs, but ran into some typing issues due to the re-use of the button and the need to add our own checkout link (gravity attaches one to the notification which is separate from the object we are querying here). Ended up piggybacking off of work in #10085 and #10117 to make sure we didn't duplicate each other's work.

In the process I also converted some order creation mutations to the `useMutation => Promise` style for more consistent response handling, which also means that the creation action doesn't require navigating to a special screen that creates the order in a side effect. **This affects our existing, non-feature-flagged behavior in the activity panel, so needs QA.**

### PR Checklist

- [x] I have tested my changes on **iOS** [x] and **Android** [ ]. _iOS only_

### To the reviewers 👀

<details><summary>Example of the partner offer expiring and using the list price (also works with partner offer)</summary>

</details>

-

#### iOS user-facing changes

- UX improvements

-

#### Dev changes

- Add feature-flagged partner offer checkout to artwork screen purchase button when available
- Partner offer order creation uses a mutation rather than navigating to a special screen that creates the mutation as a side effect.

</details>

[EMI-1753]: https://artsyproduct.atlassian.net/browse/EMI-1753?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/Activity/components/PartnerOfferCreatedNotification.tsx
+++ b/src/app/Scenes/Activity/components/PartnerOfferCreatedNotification.tsx
@@ -20,7 +20,7 @@ export const PartnerOfferCreatedNotification: React.FC<PartnerOfferCreatedNotifi
 }) => {
   const notificationData = useFragment(PartnerOfferCreatedNotificationFragment, notification)
 
-  const { headline, item, notificationType, artworksConnection, targetHref } = notificationData
+  const { headline, item, notificationType, artworksConnection } = notificationData
 
   const { hasEnded } = getTimer(item?.partnerOffer?.endAt || "")
   const noLongerAvailable = !item?.partnerOffer?.isAvailable
@@ -80,7 +80,6 @@ export const PartnerOfferCreatedNotification: React.FC<PartnerOfferCreatedNotifi
               internalID: item?.partnerOffer?.internalID || "",
               endAt: item?.partnerOffer?.endAt || "",
               isAvailable: item?.partnerOffer?.isAvailable || false,
-              targetHref: targetHref,
               note: item?.partnerOffer?.note || "",
               id: item?.partnerOffer?.internalID || "",
             }}
--- a/src/app/Scenes/Activity/components/PartnerOfferCreatedNotification.tsx
+++ b/src/app/Scenes/Activity/components/PartnerOfferCreatedNotification.tsx
@@ -77,6 +77,7 @@ export const PartnerOfferCreatedNotification: React.FC<PartnerOfferCreatedNotifi
               priceWithDiscountMessage: item?.partnerOffer?.priceWithDiscount?.display || "",
             }}
             partnerOffer={{
+              internalID: item?.partnerOffer?.internalID || "",
               endAt: item?.partnerOffer?.endAt || "",
               isAvailable: item?.partnerOffer?.isAvailable || false,
               targetHref: targetHref,
@@ -101,6 +102,7 @@ export const PartnerOfferCreatedNotificationFragment = graphql`
         expiresAt
         available
         partnerOffer {
+          internalID
           note
           internalID
           endAt
--- a/src/app/Scenes/Activity/components/PartnerOfferCreatedNotification.tsx
+++ b/src/app/Scenes/Activity/components/PartnerOfferCreatedNotification.tsx
@@ -81,7 +81,6 @@ export const PartnerOfferCreatedNotification: React.FC<PartnerOfferCreatedNotifi
               endAt: item?.partnerOffer?.endAt || "",
               isAvailable: item?.partnerOffer?.isAvailable || false,
               note: item?.partnerOffer?.note || "",
-              id: item?.partnerOffer?.internalID || "",
             }}
             showArtworkCommercialButtons
           />
--- a/src/app/Scenes/Activity/components/PartnerOfferCreatedNotification.tsx
+++ b/src/app/Scenes/Activity/components/PartnerOfferCreatedNotification.tsx
@@ -100,7 +100,6 @@ export const PartnerOfferCreatedNotificationFragment = graphql`
         expiresAt
         available
         partnerOffer {
-          internalID
           note
           internalID
           endAt
--- a/src/app/Scenes/Artwork/Components/ArtworkCommercialButtons.tsx
+++ b/src/app/Scenes/Artwork/Components/ArtworkCommercialButtons.tsx
@@ -1,7 +1,7 @@
 import { Spacer, Flex, Join } from "@artsy/palette-mobile"
 import { ArtworkCommercialButtons_artwork$key } from "__generated__/ArtworkCommercialButtons_artwork.graphql"
 import { ArtworkCommercialButtons_me$key } from "__generated__/ArtworkCommercialButtons_me.graphql"
-import { ArtworkCommercialButtons_partnerOfferToCollector$key } from "__generated__/ArtworkCommercialButtons_partnerOfferToCollector.graphql"
+import { ArtworkCommercialButtons_partnerOffer$key } from "__generated__/ArtworkCommercialButtons_partnerOffer.graphql"
 import { AuctionTimerState } from "app/Components/Bidding/Components/Timer"
 import { ArtworkStore } from "app/Scenes/Artwork/ArtworkStore"
 import { BuyNowButton } from "app/Scenes/Artwork/Components/CommercialButtons/BuyNowButton"
@@ -14,7 +14,7 @@ import { BidButtonFragmentContainer } from "./CommercialButtons/BidButton"
 interface ArtworkCommercialButtonsProps {
   artwork: ArtworkCommercialButtons_artwork$key
   me: ArtworkCommercialButtons_me$key
-  partnerOfferToCollector: ArtworkCommercialButtons_partnerOfferToCollector$key
+  partnerOffer: ArtworkCommercialButtons_partnerOffer$key
 }
 
 const RowContainer: React.FC = ({ children }) => {
@@ -31,12 +31,12 @@ const RowContainer: React.FC = ({ children }) => {
 
 export const ArtworkCommercialButtons: React.FC<ArtworkCommercialButtonsProps> = ({
   artwork,
-  partnerOfferToCollector,
+  partnerOffer,
   me,
 }) => {
   const artworkData = useFragment(artworkFragment, artwork)
   const meData = useFragment(meFragment, me)
-  const partnerOfferData = useFragment(partnerOfferFragment, partnerOfferToCollector)
+  const partnerOfferData = useFragment(partnerOfferFragment, partnerOffer)
   const selectedEditionId = ArtworkStore.useStoreState((state) => state.selectedEditionId)
   const auctionState = ArtworkStore.useStoreState((state) => state.auctionState)
   const isBiddableInAuction = artworkData.isInAuction && artworkData.sale
@@ -156,7 +156,7 @@ const meFragment = graphql`
 `
 
 const partnerOfferFragment = graphql`
-  fragment ArtworkCommercialButtons_partnerOfferToCollector on PartnerOfferToCollector {
+  fragment ArtworkCommercialButtons_partnerOffer on PartnerOfferToCollector {
     internalID
   }
 `
--- a/src/app/Scenes/Artwork/Components/ArtworkCommercialButtons.tsx
+++ b/src/app/Scenes/Artwork/Components/ArtworkCommercialButtons.tsx
@@ -1,18 +1,20 @@
 import { Spacer, Flex, Join } from "@artsy/palette-mobile"
 import { ArtworkCommercialButtons_artwork$key } from "__generated__/ArtworkCommercialButtons_artwork.graphql"
 import { ArtworkCommercialButtons_me$key } from "__generated__/ArtworkCommercialButtons_me.graphql"
+import { ArtworkCommercialButtons_partnerOfferToCollector$key } from "__generated__/ArtworkCommercialButtons_partnerOfferToCollector.graphql"
 import { AuctionTimerState } from "app/Components/Bidding/Components/Timer"
 import { ArtworkStore } from "app/Scenes/Artwork/ArtworkStore"
+import { BuyNowButton } from "app/Scenes/Artwork/Components/CommercialButtons/BuyNowButton"
+import { InquiryButtonsFragmentContainer } from "app/Scenes/Artwork/Components/CommercialButtons/InquiryButtons"
+import { MakeOfferButtonFragmentContainer } from "app/Scenes/Artwork/Components/CommercialButtons/MakeOfferButton"
 import { Children } from "react"
 import { useFragment, graphql } from "react-relay"
 import { BidButtonFragmentContainer } from "./CommercialButtons/BidButton"
-import { BuyNowButton } from "./CommercialButtons/BuyNowButton"
-import { InquiryButtonsFragmentContainer } from "./CommercialButtons/InquiryButtons"
-import { MakeOfferButtonFragmentContainer } from "./CommercialButtons/MakeOfferButton"
 
 interface ArtworkCommercialButtonsProps {
   artwork: ArtworkCommercialButtons_artwork$key
   me: ArtworkCommercialButtons_me$key
+  partnerOfferToCollector: ArtworkCommercialButtons_partnerOfferToCollector$key
 }
 
 const RowContainer: React.FC = ({ children }) => {
@@ -29,10 +31,12 @@ const RowContainer: React.FC = ({ children }) => {
 
 export const ArtworkCommercialButtons: React.FC<ArtworkCommercialButtonsProps> = ({
   artwork,
+  partnerOfferToCollector,
   me,
 }) => {
   const artworkData = useFragment(artworkFragment, artwork)
   const meData = useFragment(meFragment, me)
+  const partnerOfferData = useFragment(partnerOfferFragment, partnerOfferToCollector)
   const selectedEditionId = ArtworkStore.useStoreState((state) => state.selectedEditionId)
   const auctionState = ArtworkStore.useStoreState((state) => state.auctionState)
   const isBiddableInAuction = artworkData.isInAuction && artworkData.sale
@@ -57,7 +61,12 @@ export const ArtworkCommercialButtons: React.FC<ArtworkCommercialButtonsProps> =
             auctionState={auctionState as AuctionTimerState}
             variant="outline"
           />
-          <BuyNowButton artwork={artworkData} editionSetID={selectedEditionId} renderSaleMessage />
+          <BuyNowButton
+            partnerOffer={partnerOfferData}
+            artwork={artworkData}
+            editionSetID={selectedEditionId}
+            renderSaleMessage
+          />
         </RowContainer>
       )
     }
@@ -79,13 +88,23 @@ export const ArtworkCommercialButtons: React.FC<ArtworkCommercialButtonsProps> =
           editionSetID={selectedEditionId}
           variant="outline"
         />
-        <BuyNowButton artwork={artworkData} editionSetID={selectedEditionId} />
+        <BuyNowButton
+          partnerOffer={partnerOfferData}
+          artwork={artworkData}
+          editionSetID={selectedEditionId}
+        />
       </RowContainer>
     )
   }
 
   if (artworkData.isAcquireable) {
-    return <BuyNowButton artwork={artworkData} editionSetID={selectedEditionId} />
+    return (
+      <BuyNowButton
+        partnerOffer={partnerOfferData}
+        artwork={artworkData}
+        editionSetID={selectedEditionId}
+      />
+    )
   }
 
   if (artworkData.isInquireable && artworkData.isOfferable) {
@@ -133,5 +152,11 @@ const artworkFragment = graphql`
 const meFragment = graphql`
   fragment ArtworkCommercialButtons_me on Me {
     ...BidButton_me
+  }
+`
+
+const partnerOfferFragment = graphql`
+  fragment ArtworkCommercialButtons_partnerOfferToCollector on PartnerOfferToCollector {
+    internalID
   }
 `
--- a/src/app/Scenes/Artwork/Components/ArtworkCommercialButtons.tsx
+++ b/src/app/Scenes/Artwork/Components/ArtworkCommercialButtons.tsx
@@ -158,5 +158,6 @@ const meFragment = graphql`
 const partnerOfferFragment = graphql`
   fragment ArtworkCommercialButtons_partnerOffer on PartnerOfferToCollector {
     internalID
+    endAt
   }
 `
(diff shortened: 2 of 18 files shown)


ArtsyOpenSource:

- Dev changes (Add feature-flagged partner offer checkout to artwork screen purchase button when available)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 7fcf0c85d40974e2459f24635fc89c8a0e6a0b79
</p>

src/app/Scenes/Artwork/usePartnerOfferOnArtwork.tsx
```diff
@@ -0,0 +1,88 @@
...
+      }, interval)
+
+      return () => {
+        clearTimeout(timeout)
+      }
+    }
+  }, [endAt, timer])
+
+  return timer
+}
```
erikdstock, line 44: My approximation of [force's more sophisticated useTimer() implementation.](https://github.com/artsy/force/blob/f1be8714692190684efa8579392476ba6fb0e984/src/Utils/Hooks/useTimer.tsx#L25-L56)

src/app/Scenes/Artwork/usePartnerOfferOnArtwork.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+const partnerOfferCheckoutHref = (partnerOffer: { internalID: string }) =>
+  `partner-offer/${partnerOffer.internalID}/checkout`
+
```
erikdstock, line 48: BuyNowButton needs this, which is nice because it's easier to initiate checkout via a route than writing a whole duplicate mutation for it. I didn't touch that interface, and this file is specific to the artwork screen.

May be useful for #10085

src/app/Scenes/Artwork/usePartnerOfferOnArtwork.tsx
```diff
@@ -0,0 +1,88 @@
...
+    isAvailable: !!partnerOfferData.isAvailable,
+    endAt: partnerOfferData.endAt || "",
+    timer: partnerOfferData.endAt ? timer : null,
+    targetHref: partnerOfferCheckoutHref(partnerOfferData),
+  }
+}
+
+const PARTNER_OFFERS_FRAGMENT = graphql`
+  fragment usePartnerOfferOnArtwork_me on Me @argumentDefinitions(artworkID: { type: "String!" }) {
+    partnerOffersConnection(artworkID: $artworkID, first: 1) {
```
erikdstock, line 74: The hook itself uses the fragment to request and build the partner offer type. Some fallback values are necessary to fulfill the BuyNowButton's prop type.

erikdstock (commented):
Small tour.

src/app/Scenes/Artwork/usePartnerOfferOnArtwork.tsx
erikdstock: One thing that could become painful is if we need to request additional or differently-formatted properties in the query that fuels this hook - like [here](https://github.com/artsy/eigen/pull/10085/files#diff-1878884bed2847a33d6cdbbe376ad06b42c82cc6451fb8b7264acc47cb505a3bR121-R123).

src/app/Scenes/Artwork/Components/ArtworkCommercialButtons.tsx
```diff
@@ -43,6 +44,10 @@ export const ArtworkCommercialButtons: React.FC<ArtworkCommercialButtonsProps> =
     isBiddableInAuction
   const noEditions = !artworkData.editionSets || artworkData.editionSets.length === 0
 
+  const partnerOffer = usePartnerOfferOnArtwork(meData) || undefined
```
damassi, line 47: Rather than `|| undefined`, just type is as possibly undefined from the hook. If nothing is there or comes back, that's that -- its already undefined. 

(Superminor pedantic note -- undefined isn't something that's assigned, but rather `null`)

src/app/Scenes/Artwork/usePartnerOfferOnArtwork.tsx
```diff
@@ -0,0 +1,88 @@
...
+  priceWithDiscount: any
+  isAvailable: boolean
+  note?: string | null
+  targetHref: string
+  timer: ReturnType<typeof getTimer> | null
+}
+
+const THE_PAST = new Date(0).toISOString()
+
+const useTimer = (endAt?: string | null) => {
```
damassi, line 20: Might as well move this into its own file

src/app/Scenes/Artwork/usePartnerOfferOnArtwork.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const [timer, setTimer] = useState(getTimer(endAt || THE_PAST))
```
damassi, line 21: No need to pass in THE_PAST. Can assign it as a default argument on `getTimer`, so that if `endAt` is null or undefined, it falls back to the constant defined at the top of the file and thus in scope.

src/app/Scenes/Artwork/usePartnerOfferOnArtwork.tsx
```diff
@@ -0,0 +1,88 @@
...
+      let interval = 1000 // default to 1 second
+
+      if (parseInt(timer.time.days) > 0) {
+        interval = 60 * 60 * 1000 // update every hour
+      } else if (parseInt(timer.time.hours) > 0) {
+        interval = 60 * 1000 // update every minute
+      }
+
+      const timeout = setTimeout(() => {
+        setTimer(getTimer(endAt))
```
damassi, line 34: Unless you need multiple instances of the timer, can 
```
const initialTimer = getTimer(endAt)
``` 
and then pass `initialTimer` into use state and here

src/app/Scenes/Artwork/usePartnerOfferOnArtwork.tsx
damassi: if this function is only used once, and not exported, no need for a function. Just assign directly:

```tsx
href: `partner-offer/${partnerOffer.internalID}/checkout`
```

src/app/Scenes/Artwork/usePartnerOfferOnArtwork.tsx
```diff
@@ -0,0 +1,88 @@
...
+  return timer
+}
+
+const partnerOfferCheckoutHref = (partnerOffer: { internalID: string }) =>
+  `partner-offer/${partnerOffer.internalID}/checkout`
+
+export const usePartnerOfferOnArtwork = (
+  me: usePartnerOfferOnArtwork_me$key
+): PartnerOffer | null => {
+  const partnerOfferVisibilityEnabled = useFeatureFlag("AREnablePartnerOfferV1Improvements")
```
damassi, line 52: If the flag is named one thing, no need to name the variable something else entirely different: 
```
const enablePartnerOfferV1Improvements = useFeatureFlag("AREnablePartnerOfferV1Improvements")
```

src/app/Scenes/Artwork/Components/ArtworkCommercialButtons.tsx
```diff
@@ -43,6 +44,10 @@ export const ArtworkCommercialButtons: React.FC<ArtworkCommercialButtonsProps> =
...
+  const validPartnerOffer =
```
damassi, line 48: The naming of this doesn't make much sense -- a `validPartnerOffer` isn't something that can be `undefined`. 

Also, i'm sorry i'm flagging this 1:2 stuff everywhere, but please try to make it a habit 😄  If there's a prop on a component called `partnerOffer`, we shouldn't pass it something called `validPartnerOffer`. Just pass it `partnerOffer`. Complexity is defined by the amount of information in a system. Simplicity is sticking to only what is _strictly necessary_.

src/app/Scenes/Artwork/Components/ArtworkCommercialButtons.tsx
```diff
@@ -43,6 +44,10 @@ export const ArtworkCommercialButtons: React.FC<ArtworkCommercialButtonsProps> =
...
+    partnerOffer?.isAvailable && !partnerOffer.timer?.hasEnded ? partnerOffer : undefined
```
damassi, line 49: All behavior can be defined by the return value of the hook, vs intermediate variables around the hook, designed to look like the return value of the hook. Ie:

```tsx
const partnerOffer = usePartnerOffer(endAt)
```

The contents of `usePartnerOffer` (and associated type) define what is valid:

```tsx
const usePartnerOffer = (endAt): PartnerOffer | null => {
  if (!endAt) {
    return null
  }

  if (!isAvailable) {
    return null
  }
  
  if (!timer.hasEnded) {
    return null 
  }
  
  return { ... } 
}
```
From here you can test the hook directly, vs having to test the hook and other parts of the code, etc. It also disambiguates things.

src/app/Scenes/Artwork/Components/ArtworkCommercialButtons.tsx
```diff
@@ -57,7 +62,12 @@ export const ArtworkCommercialButtons: React.FC<ArtworkCommercialButtonsProps> =
             auctionState={auctionState as AuctionTimerState}
             variant="outline"
           />
-          <BuyNowButton artwork={artworkData} editionSetID={selectedEditionId} renderSaleMessage />
+          <BuyNowButton
+            partnerOffer={partnerOffer}
```
damassi, line 66: But all that said, we don't even need to pass `partnerOffer` to the BuyNow button. Just pass `me` fragment key, and then let the `BuyNow` button manage its own data in an entirely co-located way via relay, vs passing data around and building up abstractions that only seem to be used once. 

Doing this, we can eliminate the `usePartnerOffer` hook completely. Instead we just 
- create a `useTimer` hook and ideally copy whats in force exactly; 
- and reference that in the `BuyNowButton`, which plucks out `partnerOffer` from the connection off of `me`.

src/app/Scenes/Artwork/usePartnerOfferOnArtwork.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+export interface PartnerOffer {
+  internalID: string
+  endAt: string
+  priceWithDiscount: any
+  isAvailable: boolean
+  note?: string | null
+  targetHref: string
+  timer: ReturnType<typeof getTimer> | null
+}
```
damassi, line 16: It breaks best practices to create or redefine data interfaces from things that can be plucked right out auto-generated relay types. ie, this is mostly a node on an edge on a `partnerOfferConnection`; grab that directly, and then `extend` it with whatever is further necessary.

src/app/Scenes/Artwork/usePartnerOfferOnArtwork.tsx
```diff
@@ -0,0 +1,88 @@
...
+    }
+  }, [endAt, timer])
+
+  return timer
+}
+
+const partnerOfferCheckoutHref = (partnerOffer: { internalID: string }) =>
+  `partner-offer/${partnerOffer.internalID}/checkout`
+
+export const usePartnerOfferOnArtwork = (
```
damassi, line 49: Mentioned [here](https://github.com/artsy/eigen/pull/10087/files#r1568163418), but i think we can avoid this hook completely and bring complexity way down

damassi (commented):
@erikdstock - I know this is still a draft PR, but its taken on enough shape to where I wanted to go ahead and call out a few items before any more time is spent. 

Specifically:
- Timer code can live in `useTimer.tsx`. We should copy force's more thought out and reusable implementation, vs writing something new (unless strictly necessary). 
- We don't need a `usePartnerOffer` hook at all, since all we're doing with it is passing the return value of that into a `BuyNowButton` prop
- We can colocate all partner offer data and request it directly from within the `BuyNowButton` 
- All that `BuyNowButton` needs is a `me` relay key prop; in the parent, we simply spread `BuyNowButton_me`; in BuyNowButton` we pluck what we need off of me via `useFragment`. This additionally keeps things flexible in the future
- Inside of BuyNowButton, ParterOffer data can be "computed" directly in the function body of the component or in an associated util, but looking closely at that code, that's not necessary

Somewhat strongly think that this should be reworked and simplified before merge and totally happy to pair if any of the comments are confusing! 

(It also might be easiest to work through the above from a fresh branch, vs refactoring it in here with all of the already-existing comments that might make future review more complicated than necessary.)

src/app/Scenes/Artwork/Components/ArtworkCommercialButtons.tsx
damassi: Also, lets please stick to 1:1 naming. If `usePartnerOfferOnArtwork` returns a `partnerOffer`, just name the hook `usePartnerOffer` 🙏

erikdstock:
Thanks for the early feedback @damassi - on this point:

> * We don't need a `usePartnerOffer` hook at all, since all we're doing with it is passing the return value of that into a `BuyNowButton` prop

I'm in favor of reworking and simplifying - really the only reason for the hook was to combine the timer with the PO object. The reason for it (and why i dont know if it should be moved directly into the button) is two fold:
- Notification panel borrows the BuyNowButton component from artwork and passes a different object than the normal PartnerOffer type [here](https://github.com/artsy/eigen/blob/573f87d77aabc56441938d2584a04059b45377a3/src/app/Scenes/Activity/components/PartnerOfferCreatedNotification.tsx#L71-L83).
- There are other parts of the artwork screen that have to adjust in the presence of a partner offer, particularly those in flight for #10085. Since we are dealing with a connection type that needs to become a single optional value and the offer may expire while the user is on the screen I thought keeping this logic somewhere would be best.

My thinking was that the hook would encapsulate this logic. I'll work with Sultan to get it sorted out asap.

damassi:
> Notification panel borrows the BuyNowButton component from artwork and passes a different object than the normal PartnerOffer type

This is fine! There's a few different options here:
- If we're on a different type of screen its ok to allow for an optional `partnerOffer` argument on the button. 
- If that is found, we skip the fragment request on `me`, similar to the following 
```graphql
...BuyNowButton_me @relay(skip: $shouldSkipForSomeReason)
``` 
- Another option is to add a "me enabled" wrapper around the BuyNowButton:
```tsx
const CollectorBuyNowButton = ({ me }) => {
  const meData = useFragment(..., me) 
  const partnerOffer = ...
  
  return <BuyNowButton partnerOffer={partnerOffer} ... />
} 
```

The link posted from [here](https://github.com/artsy/eigen/blob/573f87d77aabc56441938d2584a04059b45377a3/src/app/Scenes/Activity/components/PartnerOfferCreatedNotification.tsx#L79-L83) is a good example of only doing only whats strictly necessary. From what I can tell here, this is what's needed for a partner offer. Three props, two from relay, one an href. Perfect. In this PR there's vastly more code, more types, more everything. Yes, a timer is present, but that's just another prop on the partnerOffer being passed in. Is `timer` a necessary property of a partner offer? If not, separate it, simplify up and down. A timer isn't a partner offer. We don't need to spread hooks and so on and so forth if all that's actually needed is `endAt`, `isAvailable` and `href`.  Pass those in. Stick to what is strictly necessary 😉. 

Sorry for being a bit fussy here about this, and happy to help out. (Again, would suggest tackling a second pass of this in a new branch.)

src/app/Scenes/Artwork/Components/ArtworkCommercialButtons.tsx
erikdstock: oh yes, maybe a better way to write it but this was for the optional prop of the BuyNowButton - the hook as written returns `null` for the reason you mentioned, so i need to convert it back to undefined. Could probably do it in different ways to achieve the same ends.

src/app/Scenes/Artwork/usePartnerOfferOnArtwork.tsx
erikdstock: Yes I'd like to do this if there's support and to match the force API if possible. This was just intended as an experiment since I wasn't sure if it already existed. I'll work with mobile practice on this - I think my ideal would be to make the full `Hooks/` expression a followup PR just to not block work figuring out that if-possible parts.

damassi: I don't think we need to work with mobile practice and adding additional process, and holding up work on the PR. Its just a matter of adding a nice useable hook to the codebase that wasn't there before, but which is useful.

src/app/Scenes/Artwork/Components/ArtworkCommercialButtons.tsx
erikdstock: Mentioned in my main comment above but there are two reasons I don't want to do this:
- BuyNowButton is already used outside the Artwork screen by the notifications panel which passes it a partner offer with a slightly different type (because it comes from a notification object). The fact it's used outside the artwork screen puts some limits on what we can do elegantly.
- The artwork screen has to respond to the presence of the partner offer in other components.

I figured as part of #10085 we might move the usage of this hook to a higher level on the artwork screen or otherwise memoize it so that all components can share the same hook result.

src/app/Scenes/Artwork/Components/CommercialButtons/BuyNowButton.tsx
```diff
@@ -46,68 +53,96 @@ export const BuyNowButton = ({
...
+    try {
+      const response = await partnerOfferMutation.commitMutation({
+        partnerOfferId: partnerOffer.internalID,
+      })
+      const orderOrError = response.commerceCreatePartnerOfferOrder?.orderOrError
+      if (orderOrError?.error) {
+        const { code: errorCode, data: artwork } = orderOrError.error
+        const artworkId = JSON.parse(artwork?.toString() ?? "{}")?.artwork_id
+
+        if (errorCode === "expired_partner_offer") {
```
erikdstock, line 71: Response handling adapted from PartnerOfferContainer route below.

src/app/Scenes/PartnerOffer/mutations/usePartnerOfferCheckoutMutation.tsx
```diff
@@ -6,24 +6,25 @@ import {
 } from "__generated__/usePartnerOfferCheckoutMutation.graphql"
 import { graphql, useMutation } from "react-relay"
 
-export const usePartnerOfferMutation = (
-  onMutationComplete?: (response: usePartnerOfferCheckoutMutation$data) => void
-) => {
+export const usePartnerOfferMutation = () => {
```
erikdstock, line 9: promisified - affects PartnerOfferContainer which creates the partner offer order when routed from an email/push.

src/app/Scenes/Artwork/hooks/useCreateOrder.ts
```diff
@@ -0,0 +1,49 @@
+import {
+  useCreateOrderMutation,
+  useCreateOrderMutation$data,
+  CommerceCreateOrderWithArtworkInput,
+} from "__generated__/useCreateOrderMutation.graphql"
+import { graphql, useMutation } from "react-relay"
+
+export const useCreateOrder = () => {
```
erikdstock, line 8: hook/promisified for BuyNowButton. Makes it easier to give the button two different actions to take if they use similar call signatures.

src/app/Scenes/Artwork/Components/CommercialButtons/BuyNowButton.tsx
```diff
@@ -46,68 +53,96 @@ export const BuyNowButton = ({
...
+        navigate(`/orders/${orderOrError.order.internalID}`, {
+          passProps: { title: "Purchase" },
+        })
+      }
+    } catch (e) {
+      onMutationError(e as Error)
+    } finally {
+      setIsCommittingCreateOrderMutation(false)
     }
+  }
```
MrSltun, line 124: **Praise:** Refactoring the create order mutation! 👏👏👏

MrSltun (dismissed):
Great work Erik! 🌟🌟
I just left one comment

src/app/Scenes/Artwork/usePartnerOfferOnArtwork.tsx
erikdstock: moved to a new branch to handle separately

src/app/Scenes/Artwork/Components/CommercialButtons/BuyNowButton.tsx
```diff
@@ -46,68 +53,96 @@ export const BuyNowButton = ({
...
+          })
+
+          return
+        }
+
+        if (errorCode === "not_acquireable") {
+          navigate(`/artwork/${artworkId}`, {
+            replaceActiveScreen: true,
+            passProps: { artworkOfferUnavailable: true },
+          })
```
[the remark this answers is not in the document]
erikdstock: nope i think we talked about this in passing last week - it's copied from the usage in partner offer container, i believe. so we don't want that here... I will adjust.

src/app/Scenes/PartnerOffer/PartnerOfferContainer.tsx
```diff
@@ -1,55 +1,63 @@
...
+
+          return
+        } else {
+          handleGenericError()
+        }
+      } else if (orderOrError?.order) {
+        // we need to go back to the home screen before navigating to the orders screen
+        // to prevent the user from closing the modal and navigating back to the this screen
+        goBack()
+        navigate(`/orders/${orderOrError.order?.internalID}`)
```
damassi, line 48: Is it possible that these two actions might get into a race condition in some fashion? Seems like a potential future bug.

src/app/Scenes/PartnerOffer/mutations/usePartnerOfferCheckoutMutation.tsx
damassi: Can we rename this file so that it matches the name of the primary named export 1:1? 
`/mutations/usePartnerOfferMutation.tsx`

damassi (dismissed):
Much, much simpler 👍 Thanks for working through all of these changes. Just a couple minor questions but nothing blocking!

src/app/Scenes/PartnerOffer/PartnerOfferContainer.tsx
damassi: no prob 👍

src/app/Scenes/Activity/components/PartnerOfferCreatedNotification.tsx
```diff
@@ -101,6 +100,7 @@ export const PartnerOfferCreatedNotificationFragment = graphql`
         expiresAt
         available
         partnerOffer {
+          internalID
```
erikdstock, line 103:
```suggestion
```

src/app/Scenes/PartnerOffer/mutations/usePartnerOfferCheckoutMutation.tsx
damassi: Chatted in slack about colliding fragment names -- all good!

PR #10122, opened 2024-04-22
fix: Prevent Activity panel from crashing when Activities miss images

olerichter00:
### Description

This allows the Activity panel and rail items to display without images. Without this fix, the app was crashing. with this fix a gray box is shown in case the image is missing.

Usually, all activities should have images. I just noticed this bug in staging where some data is missing.

### PR Checklist

### To the reviewers 👀

- Prevent Activity panel from crashing when Activities miss images - ole

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Scenes/Activity/ActivityItem.tsx
+++ b/src/app/Scenes/Activity/ActivityItem.tsx
@@ -59,7 +59,7 @@ export const ActivityItem: React.FC<ActivityItemProps> = memo(
               <Flex flexDirection={showAsRow ? "row" : "column"}>
                 <Flex flexDirection="row" alignItems="center">
                   {item.previewImages.map((image) => {
-                    if (!image.url) return null
+                    if (!image?.url) return null
 
                     return (
                       <Flex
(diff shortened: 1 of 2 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (Prevent Activity panel from crashing when Activities miss images - ole)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 53d4001cc1b81b7c19f033a51578d21550c8c1ca
</p>

MounirDhahri (approved):

src/app/Scenes/Activity/ActivityItem.tsx
```diff
@@ -59,7 +59,7 @@ export const ActivityItem: React.FC<ActivityItemProps> = memo(
               <Flex flexDirection={showAsRow ? "row" : "column"}>
                 <Flex flexDirection="row" alignItems="center">
                   {item.previewImages.map((image) => {
-                    if (!image.url) return null
+                    if (!image?.url) return null
```
MounirDhahri, line 62: was the type generated wrong? if so maybe it's worth it to fix this in MP as well

olerichter00: Well, the type is correct, and `previewImages` is an array of non-nullable images. Now, I am not able to reproduce the error anymore. 🤔

I could imagine that the `Image` resolver failed and in this case an image could be `null` because in theory all fields are optional.

olerichter00: I'll merge this anyway to make sure this is not breaking the whole screen, but you're right, in theory it should not...

PR #10140, opened 2024-04-24
chore: update required fields for submission

MounirDhahri:
This PR resolves [ONYX-903] 

### Description
This PR updates the required fields for submissions in-app to be (artist, title and category)

### PR Checklist

### To the reviewers 👀

- update required fields for submissions - mounir

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[ONYX-903]: https://artsyproduct.atlassian.net/browse/ONYX-903?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/SellWithArtsy/SubmitArtwork/ArtworkDetails/ArtworkDetailsForm.tsx
+++ b/src/app/Scenes/SellWithArtsy/SubmitArtwork/ArtworkDetails/ArtworkDetailsForm.tsx
@@ -56,6 +56,13 @@ export const ArtworkDetailsForm: React.FC = () => {
         accessibilityLabel="Title"
       />
       <StandardSpace />
+      <CategoryPicker<AcceptableCategoryValue | null>
+        handleChange={(category) => setFieldValue("category", category)}
+        options={categories}
+        required
+        value={values.category}
+      />
+      <StandardSpace />
       <Input
         title="Year"
         placeholder="YYYY"
@@ -67,13 +74,7 @@ export const ArtworkDetailsForm: React.FC = () => {
       />
       <StandardSpace />
       <Spacer y={2} />
-      <CategoryPicker<AcceptableCategoryValue | null>
-        handleChange={(category) => setFieldValue("category", category)}
-        options={categories}
-        required={false}
-        value={values.category}
-      />
-      <StandardSpace />
+
       <Input
         title="Materials"
         placeholder={[
--- a/src/app/Scenes/SellWithArtsy/SubmitArtwork/ArtworkDetails/ArtworkDetailsForm.tsx
+++ b/src/app/Scenes/SellWithArtsy/SubmitArtwork/ArtworkDetails/ArtworkDetailsForm.tsx
@@ -15,8 +15,8 @@ import { Select, SelectOption } from "app/Components/Select"
 import { CategoryPicker } from "app/Scenes/MyCollection/Screens/ArtworkForm/Components/CategoryPicker"
 import { GlobalStore } from "app/store/GlobalStore"
 import { artworkRarityClassifications } from "app/utils/artworkRarityClassifications"
-import { useExperimentFlag } from "app/utils/experiments/hooks"
 import { LocationWithDetails } from "app/utils/googleMaps"
+import { useFeatureFlag } from "app/utils/hooks/useFeatureFlag"
 import { useFormikContext } from "formik"
 import React, { useEffect, useRef, useState } from "react"
 import { InfoModal } from "./InfoModal/InfoModal"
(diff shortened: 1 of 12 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (update required fields for submissions - mounir)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 04797e780323e77bfdae286b91a6964bb093cc40
</p>

src/app/Scenes/SellWithArtsy/SubmitArtwork/ArtworkDetails/ArtworkDetailsForm.tsx
```diff
@@ -56,6 +56,13 @@ export const ArtworkDetailsForm: React.FC = () => {
         accessibilityLabel="Title"
       />
       <StandardSpace />
+      <CategoryPicker<AcceptableCategoryValue | null>
```
olerichter00, line 59: nitpick: Would it make sense to add `null` as a possible value to `AcceptableCategoryValue`?

src/app/Scenes/SellWithArtsy/mutations/updateConsignSubmissionMutation.ts
```diff
@@ -18,7 +19,7 @@ export const updateConsignSubmission = (input: UpdateSubmissionMutationInput) =>
         }
       `,
       variables: {
-        input,
+        input: omitBy(input, isEmpty),
```
olerichter00, line 22: I;m not sure whether this is a good idea! With this change, it wouldn't be possible to set a field in the form to an empty value (`false, null, [], {}` etc.). Even if this isn't a problem right now, it could lead to issues in the future that are very hard to debug.

olerichter00 (commented):
Looks good! I'm just not sure about this `omitBy(input, isEmpty),`....

src/app/Scenes/SellWithArtsy/mutations/updateConsignSubmissionMutation.ts
MounirDhahri: This is accurate - I thought about this as well and find it a valid concern. The reason why I didn't think of optimizing in favor of this is that I know we are changing towards not sending any submission fields anymore and instead, we will have the submission artwork-related fields, as a **standalone artwork** in a few sprints time. 

The dilemma here is whether should we optimize for a future where this is modeled separately or for a future where we would support updating submissions. I opted for the second but I don't feel like the first option would have been a wrong choice as well

olerichter00: Hmm, if this change isn't needed, I would not `omitBy` because it's not clear how long this code will be here.

dblandin: Another potential issue: you can currently go back and resubmit previous steps (for example go back from "Photos" to "Artwork Details"), wouldn't this empty input filtering disregard any edits to remove input from a field?

src/app/Scenes/SellWithArtsy/SubmitArtwork/ArtworkDetails/validation.ts
```diff
@@ -88,17 +92,17 @@ export const artworkDetailsValidationSchema = Yup.object().shape({
     is: limitedEditionValue,
     then: Yup.string().required().trim(),
   }),
-  dimensionsMetric: Yup.string().required(),
-  height: Yup.string().required().trim(),
-  width: Yup.string().required().trim(),
+  dimensionsMetric: Yup.string(),
+  height: Yup.string().trim(),
+  width: Yup.string().trim(),
```
dblandin, line 97: For dimensions specifically, we might have to wrap the new optional status in a feature flag so we have the option to make it required again remotely without a new app release. I'll create a new ticket with the context/requirements. In the meantime, I think we can keep dimensions required in this PR so it's not blocking.

dblandin (commented):
Looking good! Also agree that the omitBy filtering might not be needed and I added a note about the dimensions fields specifically.

src/app/Scenes/SellWithArtsy/mutations/updateConsignSubmissionMutation.ts
MounirDhahri: Good point - this is a valid point that warrants updating this. Will do that

src/app/Scenes/SellWithArtsy/SubmitArtwork/ArtworkDetails/ArtworkDetailsForm.tsx
MounirDhahri: I wanted to keep my changes to a minimum since this was there already and I only moved it down - I will see if it's a quick change

olerichter00 (approved):

PR #10149, opened 2024-04-25
fix(artwork): only show partner bio if enabled

damassi:
### Description

Support for private artworks: This fixes things so we only display the partner bio section if isUnlisted is true and displayPartnerBio isn't false. Also updated the tests to use latest relay patterns. 

cc @artsy/amber-devs


jpotts244 (approved):
🙏🏻 !!

Mitch-Henson (approved):

PR #10060, opened 2024-04-11
feat(ONYX-803): add recommended auction lots to homepage

tam-kis:
This PR resolves [ONYX-803](https://artsyproduct.atlassian.net/browse/ONYX-803)

### Description

Add new "Auction Lots for You" on Homepage and remove "Auction Lots for You Ending Soon". We are replacing the old rail with the new one that displays recommended auction lots.

https://github.com/artsy/eigen/assets/17580625/cc4368bb-defc-404a-9d7d-c0a83c441ecb

### PR Checklist

- [ ] I have tested my changes on **iOS** and **Android**.
- [ ] I hid my changes behind a **[feature flag]**, or they don't need one.
- [x] I have included **screenshots** or **videos**, or I have not changed the UI.
- [ ] I have added **tests**, or my changes don't require any.
- [ ] I added an **[app state migration]**, or my changes do not require one.
- [ ] I have documented any **follow-up work** that this PR will require, or it does not require any.
- [ ] I have added a **changelog entry** below, or my changes do not require one.

### To the reviewers 👀

- [ ] I would like **at least one** of the reviewers to **run** this PR on the simulator or device.

<details><summary>Changelog updates</summary>

#### Cross-platform user-facing changes

- add Auction Lots for You on Homepage - tam-kis

#### iOS user-facing changes

-

#### Android user-facing changes

-

#### Dev changes

-

</details>

Need help with something? Have a look at our [docs], or get in touch with us.

[app state migration]: ../blob/main/docs/adding_state_migrations.md
[feature flag]: ../blob/main/docs/developing_a_feature.md
[docs]: ../blob/main/docs/README.md

[ONYX-803]: https://artsyproduct.atlassian.net/browse/ONYX-803?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #10164, opened 2024-04-29
fix: Avoid displaying wrong cached data when rendering artworks for you

olerichter00:
This PR resolves [ONYX-932] 

### Description

Some cached data persists when navigating between “New Works for You” and “Auction Lots for You” This change avoid displaying wrong cached data when rendering artworks for you by changing the Relay cache policy.

### PR Checklist

### To the reviewers 👀

- Avoid displaying wrong cached data etween “New Works for You” and “Auction Lots for You” - ole

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[ONYX-932]: https://artsyproduct.atlassian.net/browse/ONYX-932?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Components/WorksForYouArtworks.tsx
+++ b/src/app/Components/WorksForYouArtworks.tsx
@@ -1,7 +1,7 @@
 import { ContextModule, OwnerType } from "@artsy/cohesion"
 import { Flex, Screen, SimpleMessage, Text } from "@artsy/palette-mobile"
-import { NewWorksForYouArtworksQuery } from "__generated__/NewWorksForYouArtworksQuery.graphql"
-import { NewWorksForYouArtworks_viewer$key } from "__generated__/NewWorksForYouArtworks_viewer.graphql"
+import { WorksForYouArtworksQuery } from "__generated__/WorksForYouArtworksQuery.graphql"
+import { WorksForYouArtworks_viewer$key } from "__generated__/WorksForYouArtworks_viewer.graphql"
 import { MasonryInfiniteScrollArtworkGrid } from "app/Components/ArtworkGrids/MasonryInfiniteScrollArtworkGrid"
 import { NewWorksForYouPlaceholder } from "app/Scenes/NewWorksForYou/NewWorksForYou"
 import { GlobalStore } from "app/store/GlobalStore"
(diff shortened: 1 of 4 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (Avoid displaying wrong cached data etween “New Works for You” and “Auction Lots for You” - ole)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 3cc64797e26d0643e80cdd09c3c902922dd5244f
</p>

src/app/Components/WorksForYouArtworks.tsx
```diff
@@ -14,10 +14,10 @@ import { graphql, useLazyLoadQuery, usePaginationFragment } from "react-relay"
 export const PAGE_SIZE = 100
 
 interface NewWorksForYouProps {
-  viewer: NewWorksForYouArtworks_viewer$key
+  viewer: WorksForYouArtworks_viewer$key
 }
 
-export const NewWorksForYouArtworks: React.FC<NewWorksForYouProps> = ({ viewer }) => {
+export const WorksForYouArtworks: React.FC<NewWorksForYouProps> = ({ viewer }) => {
```
olerichter00, line 20: Renaming this component to `WorksForYouArtworks` and moving it to `src/app/Components` because it's used in more than one screen.

src/app/Components/WorksForYouArtworks.tsx
```diff
@@ -106,23 +106,24 @@ export const newWorksForYouArtworksQuery = graphql`
...
       newWorksForYouArtworksQuery,
       {
         version,
         maxWorksPerArtist,
         onlyAtAuction,
       },
       {
-        fetchPolicy: "store-and-network",
+        // This is necessary to avoid displaying cached because this component is used in more than one place and Relay returns data from another screen before the new data is loaded.
+        fetchPolicy: "network-only",
```
olerichter00, line 126: Not sure why Relay displays the cached data even if the parameters change `onlyAtAuction = true` vs `onlyAtAuction = false` 🤷

tam-kis: Could it be related to removing the `prefetchUrl` in [this PR](https://github.com/artsy/eigen/pull/10060/files#diff-397d8c9ffea709ba3b5a563e84d7619a624a75bef28d06591f2a0f7f52e7532cL65), because it wasn't happening before, for `lotsByArtistsYouFollow`?

tam-kis (approved):
Thanks for resolving this quickly! Let's also get a review from an Eigen dev.

src/app/Components/WorksForYouArtworks.tsx
olerichter00: Prefetching the data could prevent the bug as well, but I think the main reason is that now both screens use the same component for the artwork grid, which wasn't happening before. So it might be a good idea to fix this issue anyway.

araujobarret: really weird the relay store considers `onlyAtAuction` with different values the same entry to re-use it as cached value.
Maybe moving from QueryRenderer into a fully hook approach could fix it(follow-up PR)?

olerichter00: Actually, it already uses hooks for the query and the fragments...https://github.com/artsy/eigen/blob/6b5c85c9121b7e8cb4c42b1e56f316a928717a81/src/app/Components/WorksForYouArtworks.tsx#L117

PR #10226, opened 2024-05-12
fix: broken profile header when navigating fast

araujobarret:
### Description

The glitch of not showing the header on the profile screen was gone once I:
- removed the `@refetechable` annotation from the fragment(I don't know why someone would do it, this annotation should be only used with useRefetachableFragment hook, not with useFragment)
- force the query to run on the network

There are some extra things to be removed but it's a bit out of the scope of this fix, I'm handling it on behalf of Diamond but just to clarify it wasn't introduced by us.

Highly appreciate it if someone could test and merge this PR(tomorrow I will be roaming around some trains and then go for holidays)

https://github.com/artsy/eigen/assets/15792853/ae747c5e-e460-4977-a456-022846d9d425

https://github.com/artsy/eigen/assets/15792853/b333177e-11d6-4021-803c-c9fc1e6bf9c1

### PR Checklist

### To the reviewers 👀

- fix: my profile header missing when opening just after bootstrap quickly - araujobarret

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Scenes/MyProfile/MyProfileHeader.tsx
+++ b/src/app/Scenes/MyProfile/MyProfileHeader.tsx
@@ -35,9 +35,9 @@ import { normalizeMyProfileBio } from "./utils"
 
 const ICON_SIZE = 14
 
-export const MyProfileHeader: React.FC<{ data: MyProfileHeader_me$key }> = ({ data }) => {
+export const MyProfileHeader: React.FC<{ meProp: MyProfileHeader_me$key }> = ({ meProp }) => {
   const { fetchKey, refetch } = useRefetch()
-  const me = useFragment<MyProfileHeader_me$key>(myProfileHeaderFragment, data)
+  const me = useFragment<MyProfileHeader_me$key>(myProfileHeaderFragment, meProp)
 
   const localImage = useLocalImageStorage("profile", undefined, undefined, fetchKey)
   const userProfileImagePath = localImage?.path || me?.icon?.url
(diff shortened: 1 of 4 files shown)


src/app/Scenes/MyProfile/MyProfileHeader.tsx
```diff
@@ -35,18 +35,17 @@ import { normalizeMyProfileBio } from "./utils"
 
 const ICON_SIZE = 14
 
-export const MyProfileHeader: React.FC<{ me: MyProfileHeader_me$key }> = (props) => {
-  const newCollectorSettings = useFeatureFlag("AREnableNewCollectorSettings")
-  const space = useSpace()
+export const MyProfileHeader: React.FC<{ data: MyProfileHeader_me$key }> = ({ data }) => {
   const { fetchKey, refetch } = useRefetch()
```
araujobarret, line 39: this is a terrible pattern, won't touch it here but we def should touch the store to update things rather than forcing a component to re-render....

ArtsyOpenSource:

- Cross-platform user-facing changes (fix: my profile header missing when opening just after bootstrap quickly - araujobarret)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 7af12cd501f9bed0475e5adcd7312e04b19f23bf
</p>

src/app/Scenes/MyProfile/MyProfileHeader.tsx
araujobarret: on Line 63 we prop drill it until the mutation callback to force a parent screen(Profile) to be refreshed, this is not how we work with Relay

olerichter00 (dismissed):
Thanks for the fix! 🌟 

This is a strange bug. It looks like Relay does not request the data because it thinks it is already in the store or the data has already been requested (this might be the reason why it only doesn't load the data when navigating to the screen quickly).

`network-only` and `store-and-network` both work as fetch policies).

MounirDhahri (dismissed):
Good catch!

dariakoko (approved):

PR #10261, opened 2024-05-17
fix(APPL-565): Fix external url issue in Android webview

MrSltun:
This PR resolves [APPL-565] 

### Description
This PR fixes the issue in Android when tapping an external URL inside the webview
Where the user gets redirected to the browser, but the webview goes to a blank page

### Videos
| Before | After |
|---|---|
| <video src="https://github.com/artsy/eigen/assets/20655703/a9310489-c1ac-4221-b945-c7c17cfd6577" /> | <video src="https://github.com/artsy/eigen/assets/20655703/c6c9f359-db9a-4b8f-a037-91cb68b1345c" /> |

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

- Fix the issue with the webview having a blank page after tapping an external URL - mrsltun

#### Dev changes

-

</details>

[APPL-565]: https://artsyproduct.atlassian.net/browse/APPL-565?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Components/ArtsyWebView.tsx
+++ b/src/app/Components/ArtsyWebView.tsx
@@ -14,7 +14,7 @@ import { useWebViewCallback } from "app/utils/useWebViewEvent"
 import { debounce } from "lodash"
 import { parse as parseQueryString } from "query-string"
 import { forwardRef, useEffect, useImperativeHandle, useRef, useState } from "react"
-import { View } from "react-native"
+import { Platform, View } from "react-native"
 import { Edge, SafeAreaView, useSafeAreaInsets } from "react-native-safe-area-context"
 import Share from "react-native-share"
 import WebView, { WebViewNavigation, WebViewProps } from "react-native-webview"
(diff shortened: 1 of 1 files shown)


ArtsyOpenSource:

- Android user-facing changes (Fix the issue with the webview having a blank page after tapping an external URL - mrsltun)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against aa8e1c8098bbbde8f7846b8f477643ac0887c670
</p>

brainbicycle (approved):
nice! this looks good, I wonder if we should also maybe add help.artsy.net to our handled routes so it stays in the webview rather than opening the browser?

gkartalis (approved):
👏 Thanks for fixing it, also agree with @brainbicycle

MrSltun:
@gkartalis @brainbicycle I like the idea too, but it looks a bit off with the artsy logo at the top
image

I'm not sure if we can remove that like we do in Force

MounirDhahri (commented):
Glad to see this working well. I wonder if you tried this solution from the community just for Android - many reported good results about it and I feel like it's worth trying out as well
```typescript
      onContentProcessDidTerminate={() => webViewRef.current.reload()}
```

brainbicycle:
> @gkartalis @brainbicycle I like the idea too, but it looks a bit off with the artsy logo at the top image
> 
> I'm not sure if we can remove that like we do in Force

we can do separately, thanks for looking into!

PR #10321, opened 2024-05-31
fix(ONYX-1026): navigate to the previous screen only after the keyboard is closed

dariakoko:
This PR resolves [ONYX-1026] 

### Description

When we go back from a modal where we used a keyboard to the previous screen, it takes time for the keyboard to close. We need to make sure the keyboard is closed before navigating back.

In this PR we added a keyboard listener that allows us to know when the keyboard is closed. After the keyboard is closed, we navigate back to the previous screen.

### PR Checklist

### To the reviewers 👀

- 

#### iOS user-facing changes

-

- navigate to the previous screen only after the keyboard is closed -daria

#### Dev changes

-

</details>

[ONYX-1026]: https://artsyproduct.atlassian.net/browse/ONYX-1026?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/SellWithArtsy/ArtworkForm/Components/SubmitArtworkTopNavigation.tsx
+++ b/src/app/Scenes/SellWithArtsy/ArtworkForm/Components/SubmitArtworkTopNavigation.tsx
@@ -5,13 +5,14 @@ import { GlobalStore } from "app/store/GlobalStore"
 import { goBack } from "app/system/navigation/navigate"
 import { useFeatureFlag } from "app/utils/hooks/useFeatureFlag"
 import { useFormikContext } from "formik"
-import { useEffect } from "react"
-import { Alert, LayoutAnimation } from "react-native"
+import { useEffect, useState } from "react"
+import { Alert, Keyboard, LayoutAnimation } from "react-native"
 
 export const SubmitArtworkTopNavigation: React.FC<{}> = () => {
   const enableSaveAndExit = useFeatureFlag("AREnableSaveAndContinueSubmission")
   const currentStep = SubmitArtworkFormStore.useStoreState((state) => state.currentStep)
   const hasCompletedForm = currentStep === "CompleteYourSubmission"
+  const [backPressed, setBackPressed] = useState(false)
 
   const { values } = useFormikContext<ArtworkDetailsFormModel>()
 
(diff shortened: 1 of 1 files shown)


ArtsyOpenSource:

- Android user-facing changes (navigate to the previous screen only after the keyboard is closed -daria)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against c159fc2b96a1c289745f93db9a69ca5baeee3b0b
</p>

araujobarret (approved):
Thanks for tackling it!

olerichter00 (commented):
Hey @dariakoko, 

With this change, the back button (x) on the "Select Artist" step isn't working anymore after returning from a step further in the flow (by clicking "Back" at the bottom navigation until reaching the "Select Artist" step again). I've created a [PR](https://github.com/artsy/eigen/pull/10324) to revert the changes. Would you have time to check if you can fix the keyboard issue differently (maybe by adding a timeout)? Please let me know if you need any support.

olerichter00:
Or maybe we can fix this by configuring [NavigationContainer](https://github.com/artsy/eigen/blob/de017de3c3aa67a28d0cf6b1903e260a883b3423/src/app/Scenes/SellWithArtsy/ArtworkForm/SubmitArtworkForm.tsx#L128): https://reactnavigation.org/docs/stack-navigator/#keyboardhandlingenabled

dariakoko:
I will work on it today, thank you for spotting bug @olerichter00!

MounirDhahri:
Did you maybe make a recording of the before and after for this issue? I am not sure we need to do this since we dismiss the keyboard already after each step. I believe we do that in the main component

damassi:
FWIW, we have this util in folio, and this screen transition wait variable from palette-mobile: https://github.com/artsy/energy/blob/main/src/utils/waitForScreenTransition.ts

dariakoko:
We chatted with @MounirDhahri about the bug and he proposed this solution https://github.com/artsy/palette-mobile/pull/233.
Should be sed like this <Screen.Body fullwidth disableKeyboardAvoidance> on he Sell screen
It makes the sticky cta stay in place even when the keyboard is open
The keyboard will still be visible on the Sell screen tho. If we want it to disappear first, we should call Keyboard.dismiss() before navigating. I believe, Mounir will add it to the PR along with some other minor changes

PR #10363, opened 2024-06-06
chore: improve validation performance

MounirDhahri:
### Description

This PR is an attempt to improve the performance of validation in the submit artwork flow.

**We have tested this on the QA session yesterday**

The idea behind this is the following:
- Formik validates values on change. This validation is done in the wrapper component `SubmitArtworkForm`. Given the massive children tree, it might take a few seconds for changes to see the validation taking effect. Hence the performance issues on the bottom navigation component. Since the bottom navigation component, is the **only** component that needs that validation, I am updating the logic to only validate the values there.

**Before**

https://github.com/artsy/eigen/assets/11945712/194e598e-da9b-4884-ac85-d04c1904f33e

**After**

https://github.com/artsy/eigen/assets/11945712/9aee3d8e-cf8a-4c75-bc6c-1e2ad29705b3

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Scenes/SellWithArtsy/ArtworkForm/Components/SubmitArtworkAddTitle.tsx
+++ b/src/app/Scenes/SellWithArtsy/ArtworkForm/Components/SubmitArtworkAddTitle.tsx
@@ -6,6 +6,7 @@ import { ArtworkDetailsFormModel } from "app/Scenes/SellWithArtsy/ArtworkForm/Ut
 import { ProvideScreenTrackingWithCohesionSchema } from "app/utils/track"
 import { screen } from "app/utils/track/helpers"
 import { useFormikContext } from "formik"
+import { ScrollView } from "react-native"
 
 export const SubmitArtworkAddTitle = () => {
   const { handleChange, values } = useFormikContext<ArtworkDetailsFormModel>()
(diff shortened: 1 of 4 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 0e06aec9a89e683e20e6e6f3ae757ca008286793
</p>

olerichter00 (approved):
Thanks for this improvement. Although this is working, I would like to explore other ways to make the submission flow more performant. This change adds even more complexity and would prevent us from displaying error messages within the form, I guess.

Let's merge this PR, but talk later about how we can simplify the whole form to avoid having to add these kinds of hacks. A structure [like this](https://github.com/artsy/force/pull/13927) that has a single isolated route and Formik context for each step could be worth exploring.

MounirDhahri:
True, I thought about that initially but the issue with such structure is that, if the formik context is screen specific, then we will need another store value to inject within the bottom navigation since it won't have access to `isValid`

olerichter00:
> True, I thought about that initially but the issue with such structure is that, if the formik context is screen specific, then we will need another store value to inject within the bottom navigation since it won't have access to isValid

Now that we don't use the sliding transition animation anymore, we could move the bottom navigation into the screens. This would also allow us to make this component simpler or have different bottom navigation components.

PR #10424, opened 2024-06-24
[Spike] chore: refactor submit artwork flow [Do not merge]

MounirDhahri:
### Description

This PR was an attempt to do some refactoring for the new submission flow. My goal here was the following: more decoupling of screens + easier navigation logic.

- I eventually found out I can simplify the navigation logic by moving the components to withing the navigation container (this wasn't possible at the beginning because our architecture was slightly different)
- I started the decoupling work of the screens in favor of having a separate provider by screen. However
   - This meant I should be querying for the values for each screen separately **OR** query once on the `SubmitArtworkForm` wrapper mounted, and keep updating its values to keep them in sync with the screen values. Both approaches led to a performance hit to the screen. The reason behind that is that the bottom buttons are shared between all screens, and they need to tell what screen is active and if the form is valid. Which is different in Force.
 

How I suggest to proceed next:
- Separate the navigation refactoring commit into a separate PR.
- No need to have a formik provider per screen currently.
- We can improve the `createOrUpdate` mutation logic to only include the payload that changed. We currently are using the old mutation that does not support that, it's time to change it.

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against da7a712127599fa4cd668dab86dd6d66db473a07
</p>

PR #10446, opened 2024-07-02
[WIP] feat(DIA-561): CompleteMyProfile flow

araujobarret:
This PR resolves [DIA-561] 

### Description

WIP, still missing some unit tests and one cleanup in the MyProfileEdit screen.

### PR Checklist

### To the reviewers 👀

- feat: add complete my profile flow - araujobarret

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>


ArtsyOpenSource:

- Cross-platform user-facing changes (feat: add complete my profile flow - araujobarret)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 20a26b101c437e2e81de41af0df7d8d21cb78b8b
</p>

src/app/Scenes/CompleteMyProfile/AvatarStep.tsx
```diff
@@ -0,0 +1,163 @@
...
+    </Screen>
+  )
+}
+
+interface ImageSelectorProps {
+  src?: State["iconUrl"]
+  onImageSelect: (props: { localPath: string; geminiUrl: string }) => void
+}
+
+const ImageSelector: FC<ImageSelectorProps> = ({ src, onImageSelect }) => {
```
egdbear, line 70: **nitpick:** Should this be in the `ImageSelector`.tsx file?

egdbear (commented):
Great stuff 🚀 . Curious to see how it works on android!

src/app/Scenes/CompleteMyProfile/AvatarStep.tsx
araujobarret: It's not mandatory, I usually do if things become long in the file, do you feel it is hard to read because gets longer with the 2 elements in one?

src/app/Scenes/CompleteMyProfile/IdentityVerificationStep.tsx
```diff
@@ -0,0 +1,72 @@
...
+export const IdentityVerificationStep = () => {
+  const space = useSpace()
+  const { goNext, isCurrentRouteDirty, field, setField } = useCompleteProfile()
+  const { user } = useCompleteMyProfileContext()
+  const { handleVerification } = useHandleIDVerification(user?.internalID ?? "")
+
+  const handleSendVerification = () => {
+    handleVerification()
+    setField(true)
+  }
```
iskounen, line 18: If this inspires the user to go to their inbox to start the ID verification flow, there is a risk that the flow will be abandoned, and any changes made in the last 3 steps will be lost.

I think we should save the profile when users tap "Continue" so that we don't need to worry about losing those changes if the user exits the flow unexpectedly in subsequent steps.

src/app/Scenes/CompleteMyProfile/useCompleteProfile.ts
```diff
@@ -0,0 +1,139 @@
...
+type StateNormalizedForMutation = Pick<State, "location" | "profession"> & { iconUrl?: string }
+
+const filterMutationInputFields = (progressState: State): StateNormalizedForMutation => {
+  return Object.keys(progressState).reduce((acc, key) => {
+    // iconUrl state is an object with localPath and geminiUrl, needs more handling than the other fields
+    if (key === "iconUrl") {
+      return { ...acc, iconUrl: progressState["iconUrl"]?.geminiUrl }
+    }
+    // filter out the isIdentityVerified field for the mutation
+    if (key !== "isIdentityVerified") {
```
iskounen, line 133: Thank you for explicitly handling this case. It would be very bad if a user were accidentally marked as having been verified.

I wonder if we can be more defensive about that happening by explicitly declaring the input object to the mutation instead of manipulating the context's state object. So instead of:

```
const input = filterMutationInputFields({
  // order matters, first the context state then the hook field state
  ...progressStateWithoutUndefined,
  ...fieldKeyValue,
})
```

we would do:

```
const input = {
  location: {
    ...
  },
  profession: "...",
  iconUrl: "..."
}
```

That way there is no risk of `isIdentityVerified` being updated by accident when this code changes (and it's a bit easier to read).

Alternatively, you can assert that `isIdentityVerified` is not included in the input passed to the mutation in your existing tests to give us some protection for accidental changes:

```
expect(commitMutationMock).toHaveBeenCalledWith(
  expect.objectContaining({
    variables: objectContaining({
      input: object.not.objectContaining({
        isIdentityVerified: true,
      }),
    }),
  })
)
```

src/app/Scenes/CompleteMyProfile/LocationStep.tsx
```diff
@@ -0,0 +1,67 @@
...
+          allowCustomLocation
+          aria-label="Enter your primary location"
+          autoFocus
+          title="Primary location"
+          placeholder="Primary location"
+          displayLocation={buildLocationDisplay(field)}
+          onChange={handleOnChange}
+          enableClearButton
+          onClear={handleOnClear}
+        />
```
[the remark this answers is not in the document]
araujobarret: I don't think it's necessary imo, the Location component is already an abstraction over the Input on palette, another abstraction would be a 3 level inheritance, I'm more in the side of duplication of props for this case 😅

src/app/Scenes/CompleteMyProfile/ChangesSummary.tsx
```diff
@@ -0,0 +1,180 @@
...
+        <Flex py={1}>
+          <Touchable onPress={goBack}>
+            <Text variant="xs">Back</Text>
+          </Touchable>
+        </Flex>
+
+        <Flex flexDirection="row" alignItems="center" gap={space(1)}>
+          <ProgressBar
+            height={4}
+            trackColor="green100"
```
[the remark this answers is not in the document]
araujobarret: not sure, will send the question to Ryan 🤔

src/app/Scenes/CompleteMyProfile/ProfessionStep.tsx
```diff
@@ -0,0 +1,43 @@
...
+        <Input
+          aria-label="Profession"
+          placeholder="Profession"
+          title="Profession"
+          autoFocus
+          value={field}
+          onChangeText={handleOnChange}
+        />
+
+        <Footer isFormDirty={isCurrentRouteDirty} onGoNext={goNext} />
```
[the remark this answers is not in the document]
araujobarret: The context is used solely to preserve the data when navigating back(the screens on top are destroyed).
The hook state is not preserved between screens, so in the end each component calling the hook has an individual dirtiness control.

Not sure how's the best way to detach the dirty control from it, some navigation fns checks also what you have in the hook's state before navigating, maybe that wasn't just clear enough from the hook(I can add some docs to the hook)

src/app/Scenes/CompleteMyProfile/ChangesSummary.tsx
```diff
@@ -0,0 +1,180 @@
...
+export const ChangesSummary = () => {
+  const space = useSpace()
+  const { goBack, saveAndExit, isLoading } = useCompleteProfile()
+  const { progressState, progressStateWithoutUndefined, steps } = useCompleteMyProfileContext()
+
+  const handleAddArtistsToMyCollection = () => {
+    if (!isLoading) {
+      navigate(`my-collection/collected-artists/new`)
+    }
+  }
```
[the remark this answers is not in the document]
araujobarret: The progress won't be lost unless we have some navigation reset if they complete the Add new Artist flow.

I had the same question when developing it, I don't have stronger opinions here, my intention was more to leave it to decide during QA to see what folks think.

If we save it before moving there's some things we will need to tweak, we'll need to save navigate back and trigger a subsequent navigation to this route, I don't know if the route navigation batching works fine or when we navigate to the profile screen it kills the next navigation call.

src/app/Scenes/CompleteMyProfile/useCompleteProfile.ts
araujobarret: I liked the simplicity of the first suggestion, the reduce was a bit complex and an overkill, there's no problem passing a few extra args as undefined in the mutation payload 👍 
done in https://github.com/artsy/eigen/pull/10446/commits/05cdac9b94db664eab264e65f072c42df884d969

src/app/Scenes/CompleteMyProfile/useCompleteMyProfileSteps.ts
```diff
@@ -0,0 +1,89 @@
...
+      profession
+      location {
+        display
+      }
+      isIdentityVerified
+    }
+  }
+`
+
+type Loading = "loading"
```
gkartalis, line 41: 👀

gkartalis (commented):
**(non-blocking):** Another question (already discussed that in #practice-mobile) but raising here for visibility, 

would it make sense to use a similar patern here as we do on sell with Artsy flow (having the progress bar separated from the screen steps) in order to take advantage of the smoother animations on the progressbar?

araujobarret:
@gkartalis thanks for the review, good catch on the route missing!
I addressed the changes you requested here https://github.com/artsy/eigen/pull/10446/commits/ff4e9d417e6241bdbaf60112fdc608fe3499fa5a

We have a [ticket](https://artsyproduct.atlassian.net/browse/DIA-744) to address this nice UI behaviour, it wasn't predicted by design but I'm sure it's a nice addition and will be consistent with Sell With Artsy flow too 👍

src/app/Scenes/MyProfile/MyProfileHeader.tsx
```diff
@@ -176,7 +182,7 @@ export const MyProfileHeader: React.FC<{ meProp: MyProfileHeader_me$key }> = ({
       <Flex height={70} width={70} borderRadius={35} backgroundColor={color("black10")}>
         <TouchableOpacity
           onPress={() => {
-            navigate("/my-profile/edit", {
+            navigate("", {
```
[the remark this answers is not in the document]
araujobarret: nice catch, was a leftover before holidays 🙈 https://github.com/artsy/eigen/pull/10446/commits/ff4e9d417e6241bdbaf60112fdc608fe3499fa5a

src/app/Scenes/CompleteMyProfile/AvatarStep.tsx
```diff
@@ -0,0 +1,163 @@
...
+import { Header } from "app/Scenes/CompleteMyProfile/Header"
+import { useCompleteProfile } from "app/Scenes/CompleteMyProfile/useCompleteProfile"
+import { getConvertedImageUrlFromS3 } from "app/utils/getConvertedImageUrlFromS3"
+import { showPhotoActionSheet } from "app/utils/requestPhotos"
+import { FC, useState } from "react"
+
+export const AvatarStep = () => {
+  const { goNext, isCurrentRouteDirty, setField, field } = useCompleteProfile<State["iconUrl"]>()
+
+  const handleOnImageSelect = async ({
```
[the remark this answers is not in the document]
araujobarret: not at all, fixed here https://github.com/artsy/eigen/pull/10446/commits/ff4e9d417e6241bdbaf60112fdc608fe3499fa5a

src/app/Scenes/CompleteMyProfile/LocationStep.tsx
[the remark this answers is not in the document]
iskounen: OK. That makes sense too 👍

src/app/Scenes/CompleteMyProfile/useCompleteProfile.ts
iskounen: I see that you're still spreading `fieldKeyValue` into the input. What keys are going to be included in that?

src/app/Scenes/CompleteMyProfile/IdentityVerificationStep.tsx
iskounen: OK. Let's see what Ryan has to say. Just for context, the scenario that I'm thinking about is where a user force-closing the app at some point in the flow, and we lose the changes they made in previous steps.

src/app/Scenes/CompleteMyProfile/ChangesSummary.tsx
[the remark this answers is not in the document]
iskounen: > The progress won't be lost unless we have some navigation reset if they complete the Add new Artist flow.

How does that work? Does the "Add New Artist" flow get a copy of the context state from the "Complete My Profile" flow?

> we'll need to save navigate back and trigger a subsequent navigation to this route

I don't think we'll need to do this if we break up the `saveAndExit` method and create a `save` method just to request the mutation (if there are changes).

src/app/Scenes/CompleteMyProfile/useCompleteProfile.ts
araujobarret: that's intentional, if the field is only on the fly(haven't updated the context yet) we want to preserve it too, and send it to the mutation.

For performance's sake, we only update the context state once we move to the next screen to avoid heavy re-processing in the upper-level context

src/app/Scenes/CompleteMyProfile/ChangesSummary.tsx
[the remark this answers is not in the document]
araujobarret: > How does that work? Does the "Add New Artist" flow get a copy of the context state from the "Complete My Profile" flow?

the navigation is stacked in this case, we preserve the previous stack and start a new one on top of the complete my profile

>I don't think we'll need to do this if we break up the saveAndExit method and create a save method just to request the mutation (if there are changes).

the save isn't the problem, the problem will be the user navigating back and seeing the previous complete my profile stack of screens, that's why I said we'd need to save -> navigate to the profile screen -> navigate to the collection screen

src/app/Scenes/CompleteMyProfile/CompleteMyProfile.tsx
```diff
@@ -0,0 +1,68 @@
...
+}
+
+const Stack = createStackNavigator()
+
+const CompleteMyProfileNavigator = () => {
+  const { steps } = useCompleteMyProfileContext()
+
+  return (
+    <NavigationContainer independent>
+      <Stack.Navigator screenOptions={{ headerShown: false }}>
```
MounirDhahri, line 38: did you consider having the header here? it should help you make the transition smoother between the screens by keeping the header
```suggestion
    <NavigationContainer independent>
      <Header />
      <Stack.Navigator screenOptions={{ headerShown: false }}>
```

https://github.com/artsy/eigen/assets/11945712/cf87cfd4-7b34-4d0e-b023-74e1778d0bea

araujobarret: After some testing, I noticed some jankiness when using it(tried some perf improvs like useCallback for the header and stuff, same result):

https://github.com/artsy/eigen/assets/15792853/304e060b-aa4f-4c0c-b16d-383e84409ee7

src/app/Scenes/CompleteMyProfile/useCompleteProfile.ts
iskounen: Now I understand why `fieldKeyValue` is being spread there 👍 

It looks like `isIdentityVerified: false` will be spread when the user taps "Save & Exit" on the identity verification step since that is the `ROUTE_ACTION_TYPE` for that step, but it is never updated. That works in this iteration, but I am worried that future engineers may set that to `true` by accident.

src/app/Scenes/CompleteMyProfile/ChangesSummary.tsx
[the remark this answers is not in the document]
iskounen: I assumed that the user exits the "Complete My Profile" flow when they enter the "Add New Artist" flow, but it sounds like that's not the case. What I'm hearing is that the user is still in the "Complete My Profile" flow while they are in the "Add New Artist" flow, and are returned to it when the user completes the "Add New Artist" flow. Is that correct?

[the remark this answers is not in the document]
araujobarret: Exactly 👍 
Basically starts another stack navigator on top of the current one, navigation in Eigen is a bit hybrid we still have native parts that is why some parts are not 100% flexible,

src/app/Scenes/CompleteMyProfile/useCompleteProfile.ts
araujobarret: that's right, I think we still need the filter function that I removed then 😓 

I'm not 100% sure what happens if we send `isIdentityVerified` to Gravity, I don't think we expose this field to be changed by the user from an endpoint, will double-check

src/app/Scenes/CompleteMyProfile/CompleteMyProfile.tsx
araujobarret: It looks better now, there's some minor padding to adjust when having the keyboard open in Android but should be good to go I believe(we'll address more stuff during QA too)

https://github.com/artsy/eigen/assets/15792853/87564dc3-975b-4cde-8346-8516457a8cc3

src/app/Scenes/CompleteMyProfile/useCompleteProfile.ts
araujobarret: I decided to improve the test to avoid anything extra on the payload and kept the extractor function, the progress state can't be mapped 1:1 fully to the mutation due to some app constraints such as the local path of the image to be stored in the relay state.

src/app/Scenes/CompleteMyProfile/CompleteMyProfile.tsx
MounirDhahri: awesome!

MounirDhahri: Btw, we faced the same issue of the keyboard in the submit artwork flow and we had to wrap the bottom buttons with a special container to solve it. 
https://github.com/artsy/eigen/blob/f32e39db641275331c084ff41da2f9dff7ec0015/src/app/Scenes/SellWithArtsy/ArtworkForm/Components/SubmitArtworkBottomNavigation.tsx#L192

MounirDhahri:
I have a few design comments that are not blocking to this PR - but I want to share them to make our UX is aligned across the app.
1. Do we intentionally don't want to have a `borderTop` at the top of the bottom navigation. We do that in the submit artwork flow design
Screenshot 2024-07-08 at 12 23 02

2. I see we show the step number here, but we don't show that in the submit artwork flow - should we align on one approach in both areas to avoid the confusion. 

3. Can we present the complete your profile modally instead of pushing it as a screen. That will help avoid triggering the swipe back gesture to quit the screen - (we would also need to disable the same gesture btw as a modal, we have a prop for that in `AppRegistry`)

araujobarret:
thanks for taking a look @MounirDhahri !

1. I think it's intentional but I'll double-check with Ryan anyways 👍 
2. Will also put him in the loop, no strong feelings from my side
3. I have a different view here, I see the swipe back as a nice feature to have, right now the swipe back returns to the previous step or the home screen if it's the 1st step and you can close the whole stack just tapping the `Close & Save` button. I don't think this screen should be a modal unless we'd have some hard technical limitations

src/app/Scenes/CompleteMyProfile/AvatarStep.tsx
```diff
@@ -0,0 +1,162 @@
...
+  State,
+  useCompleteMyProfileContext,
+} from "app/Scenes/CompleteMyProfile/CompleteMyProfileProvider"
+import { Footer } from "app/Scenes/CompleteMyProfile/Footer"
+import { useCompleteProfile } from "app/Scenes/CompleteMyProfile/useCompleteProfile"
+import { getConvertedImageUrlFromS3 } from "app/utils/getConvertedImageUrlFromS3"
+import { showPhotoActionSheet } from "app/utils/requestPhotos"
+import { FC, useState } from "react"
+
+export const AvatarStep = () => {
```
damassi, line 26: #minor React.FC

src/app/Scenes/CompleteMyProfile/AvatarStep.tsx
damassi: +1. Since it contains a good amount of self-contained functionality, its more discoverable being in its own file.

src/app/Scenes/CompleteMyProfile/ChangesSummary.tsx
```diff
@@ -0,0 +1,154 @@
...
+  Text,
+  useSpace,
+  CheckCircleFillIcon,
+  CheckCircleIcon,
+} from "@artsy/palette-mobile"
+import { useCompleteMyProfileContext } from "app/Scenes/CompleteMyProfile/CompleteMyProfileProvider"
+import { useCompleteProfile } from "app/Scenes/CompleteMyProfile/useCompleteProfile"
+import { navigate } from "app/system/navigation/navigate"
+
+export const ChangesSummary = () => {
```
damassi, line 15: React.Fc

src/app/Scenes/CompleteMyProfile/CompleteMyProfile.tsx
```diff
@@ -0,0 +1,77 @@
...
+export type Routes = keyof CompleteMyProfileNavigationRoutes
+
+export type NavigationPayloadField = {
+  field: string
+  value: string | boolean | Partial<LocationWithDetails>
+}
+
+const Stack = createStackNavigator()
+
+const CompleteMyProfileNavigator = () => {
```
damassi, line 34: Ditto here and elsewhere, to match patterns throughout the rest of the app

src/app/Scenes/CompleteMyProfile/CompleteMyProfileProvider.tsx
```diff
@@ -0,0 +1,104 @@
...
+  progressState: {},
+  setProgressState: () => {},
+  progressStateWithoutUndefined: {},
+})
+
+interface CompleteMyProfileProviderProps {
+  children: React.ReactNode
+}
+
+export const CompleteMyProfileProvider: React.FC<CompleteMyProfileProviderProps> = ({
```
damassi, line 33: Can we avoid vanilla contexts and instead use https://easy-peasy.vercel.app/docs/api/create-context-store.html?

src/app/Scenes/CompleteMyProfile/useUpdateMyProfile.tsx
```diff
@@ -0,0 +1,15 @@
+import { graphql, useMutation } from "react-relay"
```
damassi, line 1: Since there are many hooks, can we move this into a `CompleteMyProfile/Hooks` folder, which matches other idioms?

PR #10468, opened 2024-07-10
feat(POC): add error boundary to withSuspense

gkartalis:
This PR resolves [PHIRE-1019] and [PHIRE-1109]

### Description

Introduced an errorFallback component in case something goes wrong, devs can override by passing their custom `errorfallback`.

We also introduced `react-error-boundary` dep that is recommended from react official docs over the standard react one  and also utilized the `onError` callback to report the errors on Sentry.

### How to test

You can connect local metaphysics to eigen locally and throw an error from the query from any point that we do use the `withSuspense` 

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- add error boundary to withSuspense - gkartalis

</details>

[PHIRE-1019]: https://artsyproduct.atlassian.net/browse/PHIRE-1019?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PHIRE-1109]: https://artsyproduct.atlassian.net/browse/PHIRE-1109?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ArtsyOpenSource:

- Dev changes (add error boundary to withSuspense - gkartalis)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 8b2502c9d057e92451de9215516851bc9148c64a
</p>

src/app/utils/hooks/withSuspense.tsx
```diff
@@ -9,18 +10,25 @@ export const withSuspense =
       <Flex flex={1} justifyContent="center" alignItems="center">
         <Spinner />
       </Flex>
+    ),
+    ErrorFallback: React.FC<any> = () => (
+      <Flex flex={1} justifyContent="center" alignItems="center">
+        <Spinner />
```
brainbicycle, line 16: wonder if it makes sense to use the fallback passed in for the placeholder? I haven't looked elsewhere but I think in a lot of places this is a skeleton loader which might look a little better than the spinner?

brainbicycle (approved):
this looks good to me. I would actually be in favor of testing and merging because I believe as it stands today these network errors are essentially unhandled and lead to render error that then bubble up to our app wide error boundary which essentially blanks out the whole screen with the unable to load screen. In an ideal world we have thoughtful error states everywhere but I still think this is better than where we are at now. 

As far as testing I wonder if there is somewhere we can just arbitrarily throw an error, maybe in relay code when using the withSuspense hook, rather than messing with connecting local metaphysics and throwing for specific queries.

brainbicycle:
This has been discussed a bit more and I think we should merge this, it is an improvement on what happens today.

MounirDhahri:
Looks like the tests are breaking 👀

MounirDhahri:
thanks @gkartalis for looking into this.

src/app/Scenes/SavedSearchAlert/screens/AlertPriceRangeScreen.tests.js
```diff
@@ -0,0 +1,181 @@
...
+                    {
+                        count: 483,
+                        name: "50000",
+                        value: "50000",
+                    },
+                ],
+            },
+        ],
+    },
+};
```
damassi, line 181: Is this compiled code?

damassi (approved):
One q, but otherwise looks great 👍

src/app/Scenes/SavedSearchAlert/screens/AlertPriceRangeScreen.tests.js
gkartalis: 😅 I guess I run `yarn tsc src/app/Scenes/SavedSearchAlert/screens/AlertPriceRangeScreen.tests.js` instead of `yarn test src/app/Scenes/SavedSearchAlert/screens/AlertPriceRangeScreen.tests.js` and committed the diff

PR #10475, opened 2024-07-12
chore: register to bid ux/ui improvemnts

MounirDhahri:
[Future Friday]

### Description

This PR makes the following changes:

- Users were able to quit the flow directly without warning - now they will be warned about it
If you swipe down, the modal will close and you will use your progress
- we probably don't want that to happen accidentally, I made the modal sheet not dismissible by swiping down, it will only be dismissed by click the X button
If you press on Conditions of sale, we used top open it in a WebView where you could easily get lost
- now, we don't distract you from registering and open that externally so you are right where you left - just like we do in the SWA flow
- Use the new credit card input component
- The font sizes were a mess, custom line heights, custom font sizes etc.. I updated them to look more or less like what we do in other surfaces
- The paddings were massive inside the billing address screen - I made them a bit smaller to match our spacings in other areas.
- I fixed the small extra padding missing in the country picker
Instead of manually going through the fields in billing address, you can simply use the keyboard "Next" to be navigated through them faster

### Before

https://github.com/user-attachments/assets/08dfb578-e941-42c2-8d75-39833493e371

___

### After

https://github.com/user-attachments/assets/7bb0e6a9-6bca-45ff-bb7a-abf58fc71606

https://github.com/user-attachments/assets/c6b0b60e-9e44-4528-b3b1-1763c369ddde

https://github.com/user-attachments/assets/6701d86e-f0b7-442f-84af-922e6cee9643

### PR Checklist

### To the reviewers 👀

- register to bid ux/ui improvements - mounir

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Components/Bidding/Screens/BillingAddress.tsx
+++ b/src/app/Components/Bidding/Screens/BillingAddress.tsx
@@ -106,11 +106,11 @@ export const BillingAddress: React.FC<BillingAddressProps> = ({
       </FancyModalHeader>
       <ScrollView
         ref={scrollViewRef}
-        contentContainerStyle={{ padding: 20, paddingBottom: 50 }}
+        contentContainerStyle={{ paddingHorizontal: 20, paddingBottom: 50 }}
         keyboardDismissMode="on-drag"
         keyboardShouldPersistTaps="handled"
       >
-        <Stack spacing={4}>
+        <Stack spacing={2}>
           <Input
             ref={fullNameRef}
             title="Full name"
@@ -122,6 +122,7 @@ export const BillingAddress: React.FC<BillingAddressProps> = ({
             value={address.fullName}
             onChangeText={(e) => handleInputTextChange("fullName", e)}
             testID="input-full-name"
+            onSubmitEditing={() => addressLine1Ref.current?.focus()}
           />
 
           <Input
@@ -135,6 +136,7 @@ export const BillingAddress: React.FC<BillingAddressProps> = ({
             value={address.addressLine1}
             onChangeText={(e) => handleInputTextChange("addressLine1", e)}
             testID="input-address-1"
+            onSubmitEditing={() => addressLine2Ref.current?.focus()}
           />
 
           <Input
@@ -152,6 +154,7 @@ export const BillingAddress: React.FC<BillingAddressProps> = ({
             value={address.addressLine2}
             onChangeText={(e) => handleInputTextChange("addressLine2", e)}
             testID="input-address-2"
+            onSubmitEditing={() => cityRef.current?.focus()}
           />
 
           <Input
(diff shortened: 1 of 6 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (register to bid ux/ui improvements - mounir)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 7504cecd0901c85562e25af60ce026c9a08d4e3c
</p>

src/app/Components/Bidding/Screens/BillingAddress.tsx
```diff
@@ -122,6 +122,7 @@ export const BillingAddress: React.FC<BillingAddressProps> = ({
             value={address.fullName}
             onChangeText={(e) => handleInputTextChange("fullName", e)}
             testID="input-full-name"
+            onSubmitEditing={() => addressLine1Ref.current?.focus()}
```
olerichter00, line 125: 👏

src/app/Components/Bidding/Screens/Registration.tsx
```diff
@@ -116,11 +116,11 @@ export class Registration extends React.Component<RegistrationProps, Registratio
   }
 
   onPressGeneralTermsAndConditionsOfSale = () => {
-    navigate("/terms")
+    Linking.openURL("https://www.artsy.net/terms")
```
olerichter00, line 119: Is this opening the URL in the browser instead of the app?

olerichter00 (approved):
This is great! 🌟  🌟  🌟

src/app/Components/Bidding/Screens/Registration.tsx
brainbicycle: curious about why we want this, isn't it better to keep them in a webview in the app? see there is the cookie consent stuff showing up as well which I think in theory shouldn't show up in our own webviews?

brainbicycle: oh i see the justification above but don't agree with this one, we set the user agent in our webviews to avoid some of the stuff you see in the browser like the app download banner, cookie consent etc.

src/app/Components/Bidding/Screens/Registration.tsx
```diff
@@ -116,11 +116,11 @@ export class Registration extends React.Component<RegistrationProps, Registratio
...
   }
 
   onPressConditionsOfSale = () => {
-    navigate("/conditions-of-sale")
+    Linking.openURL("https://www.artsy.net/conditions-of-sale")
```
brainbicycle, line 123: same feedback

brainbicycle (changes requested):
overall this looks great! but I disagree that we should prefer browser over webview when linking out to one off pages, we get more control over webviews, if we need to make it harder to get lost we should think about updating the webview for this change IMO

src/app/Components/Bidding/Screens/Registration.tsx
MounirDhahri: addresssed

MounirDhahri: thanks for the feedbacl

brainbicycle (approved):
thank you!

PR #10491, opened 2024-07-15
refactor: update InquiryModal to use latest Relay patterns

iskounen:
This PR refactors the `InquiryModal` component to fetch its data using a `useFragment` hook instead of `createFragmentContainer`. It also adds a `useMutation` hook (instead of `commitMutation`) to declare the mutation to submit an inquiry.

### PR Checklist

#### Dev changes

- Inquiry modal uses a useFragment hook to populate artwork data
- Inquiry modal uses a useMutation hook to submit inquiries

</details>

### DIFF
--- a/src/app/Scenes/Artwork/Components/CommercialButtons/useSubmitInquiryRequest.tsx
+++ b/src/app/Scenes/Artwork/Components/CommercialButtons/useSubmitInquiryRequest.tsx
@@ -1,0 +1,23 @@
+import { useSubmitInquiryRequestMutation } from "__generated__/useSubmitInquiryRequestMutation.graphql"
+import { graphql, useMutation } from "react-relay"
+
+export const useSubmitInquiryRequest = () => {
+  return useMutation<useSubmitInquiryRequestMutation>(graphql`
+    mutation useSubmitInquiryRequestMutation($input: SubmitInquiryRequestMutationInput!) {
+      submitInquiryRequestMutation(input: $input) {
+        inquiryRequest {
+          inquireable {
+            __typename
+            ... on Artwork {
+              title
+              artists {
+                name
+                birthday
+              }
+            }
+          }
+        }
+      }
+    }
+  `)
+}
--- a/src/app/Scenes/Artwork/Components/CommercialButtons/InquiryButtons.tests.tsx
+++ b/src/app/Scenes/Artwork/Components/CommercialButtons/InquiryButtons.tests.tsx
@@ -9,7 +9,7 @@ import { graphql } from "react-relay"
 
 jest.mock("app/Scenes/Artwork/Components/CommercialButtons/InquiryModal", () => {
   return {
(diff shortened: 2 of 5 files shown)


ArtsyOpenSource:

- Dev changes (Inquiry modal uses a useFragment hook to populate artwork data,Inquiry modal uses a useMutation hook to submit inquiries)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 9a095c974265edb15f28362bf91782269e25b00e
</p>

src/app/Scenes/Artwork/Components/CommercialButtons/useSubmitInquiryRequest.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { graphql, useMutation } from "react-relay"
+
+export const useSubmitInquiryRequest = () => {
+  return useMutation<useSubmitInquiryRequestMutation>(graphql`
+    mutation useSubmitInquiryRequestMutation($input: SubmitInquiryRequestMutationInput!) {
+      submitInquiryRequestMutation(input: $input) {
+        inquiryRequest {
+          inquireable {
+            __typename
+            ... on Artwork {
```
araujobarret, line 11: do we need to return data here? I don't see the response being used anywhere

araujobarret (dismissed):

src/app/Scenes/Artwork/Components/CommercialButtons/useSubmitInquiryRequest.tsx
iskounen: I agree that we don't do anything with this data, but the mutation requires that we declare some field to be returned, so I'm not sure what to return when I don't need any data.

I see that there's another mutation in Eigen that just returns `clientMutationId`. Is that the correct pattern to follow when you don't need a return value from a mutation?

araujobarret: I think something like `clientMutationId` is fine 👍

araujobarret: or even just `__typename`

araujobarret: There's a hidden relay field embedded always which I forgot the name  🤔 
@damassi do you know the name of the field that's always present, something comes to my mind as `__id` but not sure

damassi: `id` is immediate global identifier for a parent field, but __id is also ringing a bell. Not sure. Totally cool to just return `id`.

iskounen: Thank you. I'll go with `clientMutationId`, then 👍

PR #10504, opened 2024-07-22
chore: fair screen fixes

MounirDhahri:
This PR resolves [] 

### Description

Since we have a major upcoming fair - I thought it's about time to fix the issues in the screen as well as allow users to share the fair link. Below is a list of changes:
- Fix broken tap on **Read more**
- Add Share fair URL
- Fix broken padding between sections
- Remove additional separator 
- Add `nestedScrollEnabled` for a small performance improvements (I doubt it will be noticeable but better have it than not)
- Align sort and filter experience with other screens (show number of artwork in bold instead of `Showing X artwork`)
- Use `Screen` component in both `Fair` and `FairInfo`screens.
- Fix back button covering content in both `Fair` and `FairInfo` screens

| Before | After |
|--------|--------|
| <Video src="https://github.com/user-attachments/assets/fda7c4b7-8c52-434b-9097-2eb99b3803a1" /> |<Video src="https://github.com/user-attachments/assets/d16ddd69-20bf-491c-ae25-bfe10d6f2911" /> |

### PR Checklist

- [x] I have tested my changes on **iOS** and **Android**.
- [x] I hid my changes behind a **[feature flag]**, or they don't need one.
- [x] I have included **screenshots** or **videos**, or I have not changed the UI.
- [x] I have added **tests**, or my changes don't require any.
- [x] I added an **[app state migration]**, or my changes do not require one.
- [x] I have documented any **follow-up work** that this PR will require, or it does not require any.
- [x] I have added a **changelog entry** below, or my changes do not require one.

### To the reviewers 👀

- [ ] I would like **at least one** of the reviewers to **run** this PR on the simulator or device.

<details><summary>Changelog updates</summary>

#### Cross-platform user-facing changes

- fix fair screen outstanding issues - mounir

#### iOS user-facing changes

-

#### Android user-facing changes

-

#### Dev changes

-

</details>

Need help with something? Have a look at our [docs], or get in touch with us.

[app state migration]: ../blob/main/docs/adding_state_migrations.md
[feature flag]: ../blob/main/docs/developing_a_feature.md
[docs]: ../blob/main/docs/README.md

PR #10506, opened 2024-07-22
refactor(Fair): Add Masonry to the Fair and FairFollowedArtists screen

araujobarret:
### Description

This PR refactors the Fair and FairFollowedArtists screens, which the 1st was breaking on Android and supper laggy to incorporate the new Masonry element as a grid viewer. The refactoring on the 2nd screen was necessary because it used the fragment container from the 1st screen.

The 1st commit contains the refactor and the 2nd is the pulling of #10504 adapted to the refactor.

There are follow-ups to be made re polishing the UI, and some shifting happening around but we should prioritize shipping this boost, especially because of the crashes, and to work on them later.

<details>
<summary>Android</summary>

https://github.com/user-attachments/assets/e017d029-05e0-4f0f-a6e8-da8995ec08db

</details>

<details>
<summary>iPhone</summary>

https://github.com/user-attachments/assets/536d2637-f3a0-41da-9d87-01bbc9aa650e

</details>

### PR Checklist

### To the reviewers 👀

- refactor: revamp Fair screen with Masonry and a better layout to support it - aruajobarret

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Components/ArtworkFilter/useArtworkFilters.ts
+++ b/src/app/Components/ArtworkFilter/useArtworkFilters.ts
@@ -11,7 +11,13 @@ import {
 } from "./ArtworkFilterHelpers"
 import { ArtworksFiltersStore, selectedOptionsUnion } from "./ArtworkFilterStore"
 
-interface UseArtworkFiltersOptions {
+// Relay doesn't export their helper type(KeyType), so we have to redefine it here
+type RelayData = { " $data"?: unknown; " $fragmentSpreads": FragmentType } | null | undefined
+
+type UseArtworkFiltersOptions<T extends RelayData> = {
+  /**
+   * @deprecated use the prop `refetch` that is returned from relay hooks instead of HoC relay prop
+   */
   relay?: RelayPaginationProp
   aggregations?: unknown
   pageSize?: number
@@ -21,13 +27,10 @@ interface UseArtworkFiltersOptions {
   onApply?: () => void
   onRefetch?: (error?: Error | null) => void
   refetchRef?: MutableRefObject<() => void>
+  refetch?: RefetchFnDynamic<OperationType, T>
 }
 
-/**
- * @deprecated Use useArtworkFiltersV2 instead
- * and refactor the component to use Relay hooks instead of containers
- */
-export const useArtworkFilters = ({
+export const useArtworkFilters = <T extends RelayData>({
   relay,
   aggregations,
   pageSize = PAGE_SIZE,
@@ -37,7 +40,8 @@ export const useArtworkFilters = ({
   onRefetch,
   type = "filter",
   refetchRef,
-}: UseArtworkFiltersOptions) => {
+  refetch,
+}: UseArtworkFiltersOptions<T>) => {
   const setAggregationsAction = ArtworksFiltersStore.useStoreActions(
     (state) => state.setAggregationsAction
   )
@@ -52,10 +56,13 @@ export const useArtworkFilters = ({
     setAggregationsAction(aggregations)
   }, [])
 
-  const refetch = () => {
-    if (relay !== undefined) {
-      const filterParams = filterArtworksParams(appliedFilters, filterType)
+  const _refetch = () => {
+    const filterParams = filterArtworksParams(appliedFilters, filterType)
+    const refetchArgs = refetchVariables ?? {
+      input: prepareFilterArtworksParamsForInput(filterParams),
+    }
 
+    if (!!relay) {
       relay.refetchConnection(
         pageSize,
         (error) => {
@@ -69,18 +76,36 @@ export const useArtworkFilters = ({
             throw new Error(errorMessage)
           }
         },
-        refetchVariables ?? { input: prepareFilterArtworksParamsForInput(filterParams) }
+        refetchArgs
+      )
+      return
+    }
+
+    if (!!refetch) {
+      refetch(
+        { ...refetchArgs, count: pageSize },
+        {
+          onComplete: (error) => {
+            onRefetch?.(error)
+            if (error) {
+              const errorMessage = componentPath
+                ? `${componentPath} ${type} error: ${error.message}`
+                : error.message
+              throw new Error(errorMessage)
+            }
+          },
+        }
       )
     }
   }
 
   if (refetchRef) {
-    refetchRef.current = refetch
+    refetchRef.current = _refetch
   }
 
   useEffect(() => {
     if (applyFilters) {
-      refetch()
+      _refetch()
 
       if (onApply) {
         onApply()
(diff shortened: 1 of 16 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (refactor: revamp Fair screen with Masonry and a better layout to support it - aruajobarret)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 14fe154540cf65096ccd8dd9699a65507352f563
</p>

src/app/Scenes/Fair/Components/FairArtworks.tsx
```diff
@@ -1,44 +1,213 @@
...
+        exitModal={handleFilterToggle}
+        closeModal={handleFilterToggle}
+      />
+    </>
+  )
+}
+
+// Performant version of FairArtworks that doesn't use tabs
+// left in the same file intentionally to make it easier to compare the two
+// forcing changes in both components
```
damassi, line 199: 👍

src/app/Components/ArtworkFilter/useArtworkFilters.ts
```diff
@@ -119,3 +120,71 @@ export const useSelectedFiltersCount = () => {
...
+  UseArtworkFiltersOptions,
+  "refetchRef" | "relay"
+> & {
+  refetch: RefetchFnDynamic<OperationType, T>
+}
+
+/**
+ * Replaces useArtworkFilters for components using relay hooks
+ */
+export const useArtworkFiltersForHooks = <T extends RelayData>({
```
damassi, line 136: The rename doesn't quite make sense, just in the sense that we don't use class components anymore so the previous name was fine. thoughts?

araujobarret: Agree that doesn't make sense 😂 
The component is a small variant of useArtworkFilters, but doesn't expect the relay prop, that's the main difference, it expects the refetch function exported from Pagination and Refetchable containers.

damassi: If so, we should encode that in the name, as an extension of relay. (Hooks is an overloaded term in this regard -- maybe `useRelayArtworkFilters`, or something of the kind.)

araujobarret: I think it's better, but the other one also uses relay but an outdated HoC concept, in the end, I think becomes confusing.
How about useArtworkFiltersV2 and deprecating the other one asking to people refactor components using it to use hooks instead of Containers?

PS: I think I need a kid to start getting better with naming 😂

damassi: hah yes yes :) and +1 to V2 + deprecation. That sends a stronger signal! and gives us a place to do find and replaces to see the refactor burden (whats left to migrate, whats using the new thing, then just sweep over and fix.)

araujobarret: done in https://github.com/artsy/eigen/pull/10506/commits/b3eb35f79a928a12999609077c018f371374aa25

MounirDhahri: tbh I think we can avoid creating a new hook here - what we want to do is to call `refetchConnection` directly instead of `relay.refetchConnection` - I think we can avoid that by doing something like this https://github.com/artsy/eigen/pull/10506#discussion_r1687740528 or already updating the instances of `useArtworkFilters` to send `refetchConnection` function instead of `relay`. The reason why I am nitpicking this is because the artwork filters are one the most complicated business pieces in Eigen and the less code we have related to them the more we can set ourselves for sucess

iskounen (commented):
I read through your changes to learn how the Fairs page is structured in case Diamond works on it in the future. `FairArtworks` is still a monster (and rightfully so) but the rest of the `Fair` components are well organized and easy to understand 👍

src/app/Components/ArtworkFilter/useArtworkFilters.ts
```diff
@@ -119,3 +124,71 @@ export const useSelectedFiltersCount = () => {
...
+  UseArtworkFiltersOptions,
+  "refetchRef" | "relay"
+> & {
+  refetch: RefetchFnDynamic<OperationType, T>
+}
+
+/**
+ * Replaces useArtworkFilters for components using patterns with relay hooks
+ */
+export const useArtworkFiltersV2 = <T extends RelayData>({
```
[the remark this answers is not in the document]
araujobarret: Compromise made! we don't touch this element so much in Diamond so it makes total sense to keep it as you'd like since Onyx works the most with it 👍 
I only deprecated the `relay` prop, so folks adding it to newer screens will get the nudge about using the other prop 👍

src/app/Components/ArtworkFilter/useArtworkFilters.ts
damassi: Agree with @MounirDhahri; if we can get away with 1, lets go with 1. I'll unresolve this comment, but lets not worry about refactoring this now while that crash is out in the wild. We can follow up!

src/app/Components/ArtworkFilter/useArtworkFilters.ts
```diff
@@ -10,7 +11,13 @@ import {
 } from "./ArtworkFilterHelpers"
 import { ArtworksFiltersStore, selectedOptionsUnion } from "./ArtworkFilterStore"
 
-interface UseArtworkFiltersOptions {
+// Relay doesn't export their helper type(KeyType), so we have to redefine it here
+type RelayData = { " $data"?: unknown; " $fragmentSpreads": FragmentType } | null | undefined
```
damassi, line 15: Also for a follow-up PR / discussion, but I kinda think this is a (minor) red flag, just in the sense that relay works _so good_. They've thought of everything; and if they didn't export this or define this pattern at the library level, they probably avoided it for a reason. 

In this case we should trust the friction and come up with another approach that's more idiomatic.

src/app/Components/ArtworkFilter/useArtworkFilters.ts
araujobarret: The main question is, who will be tackling this change? Does it impact all surfaces we have around.
Asking to flag a ticket, happy to host it on Diamond but the filters were heavily worked by FX if I recall it right

damassi: I would say as a general rule of thumb: If a team opts to create a duplicate function to achieve a certain end, but said duplicate function could be shared (or better yet avoided) with a small amount of refactoring, then that falls on the team needing to make the change. Otherwise refactoring work would never get done, and code would just be append only, rather than maintained.

araujobarret: I will create on 💎, if any team feels compelled to grab it we can move, but it makes sense to have some mobile work done, especially debt.

damassi: thank you @araujobarret 👍

PR #10524, opened 2024-07-29
feat: add New Works for You (and potentially other artwork-based rails) to new home view screen

nickskalkin:
### Description

Depends on https://github.com/artsy/metaphysics/pull/5865

This is a small proof of concept to validate how easy it is to add (in this case reuse) components to the new home view. I've mostly re-used existing [NewWorksForYouRail](https://github.com/artsy/eigen/blob/main/src/app/Scenes/Home/Components/NewWorksForYouRail.tsx) (removed some code for simplicity, but maybe it should be brought back).

TODO:
- [ ] Use `href` from metaphysics
- [ ] Aforementioned `href` ideally should include parameters which MP uses by default as `artworksForUser` arguments (`/new-for-you?version=C&maxWorksPerArtist=3`). See this [comment](https://github.com/artsy/metaphysics/pull/5865#discussion_r1693706818) for more context. Currently Eigen / Force hardcode some parameters (version, maxWorksPerArtist), but now Metaphysics becomes the source of truth. If we want to display same artworks on home view **rail** and on the **screen** (when user clicks on the view all chevron), screen should be aware of which parameters MP used to render the rail.

### PR Checklist

### To the reviewers 👀

#nochangelog

</details>

### DIFF
--- a/data/schema.graphql
+++ b/data/schema.graphql
@@ -2253,12 +2253,6 @@ type ArtistsRailHomeViewSection implements GenericHomeViewSection & Node {
 
   # A type-specific ID likely used as a database ID.
   internalID: ID!
-
-  # A stable identifier for this section
-  key: String!
-
-  # A display title for this section
-  title: String!
 }
 
 type Artwork implements Node & Searchable & Sellable {
@@ -3073,12 +3067,6 @@ type ArtworksRailHomeViewSection implements GenericHomeViewSection & Node {
 
   # A type-specific ID likely used as a database ID.
   internalID: ID!
-
-  # A stable identifier for this section
-  key: String!
-
-  # A display title for this section
-  title: String!
 }
 
 enum AssetType {
@@ -10375,12 +10363,6 @@ interface GenericHomeViewSection {
 
   # A type-specific ID likely used as a database ID.
   internalID: ID!
-
-  # A stable identifier for this section
-  key: String!
-
-  # A display title for this section
-  title: String!
 }
 
 type GravityMutationError {
@@ -10726,6 +10708,11 @@ type HomePageSalesModule {
 
 # Experimental schema for new home view
 type HomeView {
+  # A home view section
+  section(
+    # The ID of the section
+    id: String!
+  ): HomeViewSection
   sectionsConnection(
     after: String
     before: String
@@ -10736,17 +10723,8 @@ type HomeView {
 
 # A component specification
 type HomeViewComponent {
-  # Which component type to render
-  type: HomeViewComponentType
-}
-
-# Known component types for use in home view
-enum HomeViewComponentType {
-  # A standard rail of artists
-  ARTISTS_RAIL
-
-  # A standard rail of artworks
-  ARTWORKS_RAIL
+  # A display title for this section
+  title: String!
 }
 
 union HomeViewSection = ArtistsRailHomeViewSection | ArtworksRailHomeViewSection
--- a/src/app/AppRegistry.tsx
+++ b/src/app/AppRegistry.tsx
@@ -141,7 +141,7 @@ import {
   ViewingRoomsListScreen,
   viewingRoomsListScreenQuery,
 } from "./Scenes/ViewingRoom/ViewingRoomsList"
-import { GlobalStore } from "./store/GlobalStore"
+import { GlobalStore, unsafe_getFeatureFlag } from "./store/GlobalStore"
 import { propsStore } from "./store/PropsStore"
 import { DevMenu } from "./system/devTools/DevMenu/DevMenu"
 import { Schema, screenTrack } from "./utils/track"
@@ -488,7 +488,7 @@ export const modules = defineModules({
     hidesBackButton: true,
   }),
   Home: reactModule(HomeContainer, {
-    isRootViewForTabName: "home",
+    fullBleed: unsafe_getFeatureFlag("ARUseNewHomeView") ? true : false,
   }),
   HomeView: reactModule(HomeViewScreen, { hidesBackButton: true }),
   Inbox: reactModule(InboxQueryRenderer, { isRootViewForTabName: "inbox" }, [InboxScreenQuery]),
--- a/src/app/Scenes/HomeView/HomeView.tsx
+++ b/src/app/Scenes/HomeView/HomeView.tsx
@@ -1,6 +1,6 @@
 import { Flex, Screen, Separator, Spacer, Text } from "@artsy/palette-mobile"
 import { HomeViewQuery } from "__generated__/HomeViewQuery.graphql"
-import { NewWorksForYouSection } from "app/Scenes/HomeView/Sections/NewWorksForYouSection"
+import { ArtworksRailHomeViewSection } from "app/Scenes/HomeView/Sections/ArtworksRailHomeViewSection"
 import { goBack } from "app/system/navigation/navigate"
 import { extractNodes } from "app/utils/extractNodes"
 import { Suspense } from "react"
@@ -38,8 +38,11 @@ export const HomeView: React.FC = () => {
 const Section: React.FC<{ section: any }> = (props) => {
   const { section } = props
 
-  if (props.section.key === "NEW_WORKS_FOR_YOU") {
-    return <NewWorksForYouSection section={section} />
+  console.log("[Debug] section.component.")
+
+  // or should we check for __typename? (which is ArtworksRailHomeViewSection)
+  if (section.component.type === "ARTWORKS_RAIL") {
+    return <ArtworksRailHomeViewSection section={section} />
   }
 
   return (
@@ -89,7 +92,7 @@ export const homeViewScreenQuery = graphql`
               }
             }
 
-            ...NewWorksForYouSection_section
+            ...ArtworksRailHomeViewSection_section
           }
         }
       }
--- a/src/app/Scenes/HomeView/HomeView.tsx
+++ b/src/app/Scenes/HomeView/HomeView.tsx
@@ -1,5 +1,6 @@
 import { Flex, Screen, Separator, Spacer, Text } from "@artsy/palette-mobile"
 import { HomeViewQuery } from "__generated__/HomeViewQuery.graphql"
+import { NewWorksForYouSection } from "app/Scenes/HomeView/Sections/NewWorksForYouSection"
 import { goBack } from "app/system/navigation/navigate"
 import { extractNodes } from "app/utils/extractNodes"
 import { Suspense } from "react"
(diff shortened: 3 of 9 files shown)


data/schema.graphql
```diff
@@ -3098,6 +3098,8 @@ enum ArtworkSorts {
 
 # An artwork rail section in the home view
 type ArtworksRailHomeViewSection implements GenericHomeViewSection & Node {
+  artworksConnection(after: String, before: String, first: Int, last: Int): ArtworkConnection
```
nickskalkin, line 3101: Manually synced this field with local MP. Can be removed once https://github.com/artsy/metaphysics/pull/5865 is merged.

src/app/Scenes/HomeView/HomeView.tsx
```diff
@@ -12,7 +13,7 @@ export const HomeView: React.FC = () => {
   const sections = extractNodes(queryData.homeView.sectionsConnection)
 
   return (
-    <Screen>
+    <Screen mt={0}>
```
nickskalkin, line 16: Don't know why this needed, but without `mt={0}` I have a huge empty space on top of the screen.

src/app/Scenes/HomeView/HomeView.tsx
```diff
@@ -37,8 +38,12 @@ export const HomeView: React.FC = () => {
 const Section: React.FC<{ section: any }> = (props) => {
   const { section } = props
 
+  if (props.section.key === "NEW_WORKS_FOR_YOU") {
```
nickskalkin, line 41: This is where Eigen decides which component to render for which section.

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  ✅ **No changelog changes**
  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 8d60a9eba685096d5c7dec32fad140cd30acd626
</p>

src/app/Scenes/HomeView/HomeView.tsx
```diff
@@ -72,7 +77,7 @@ export const HomeViewScreen: React.FC = () => (
 export const homeViewScreenQuery = graphql`
   query HomeViewQuery {
     homeView {
-      sectionsConnection(first: 3) {
+      sectionsConnection(first: 4) {
```
nickskalkin, line 80: Increased to 4, because New Works for You section is 4th

src/app/Scenes/HomeView/Sections/NewWorksForYouSection.tsx
```diff
@@ -0,0 +1,61 @@
+import { Flex } from "@artsy/palette-mobile"
```
nickskalkin, line 1: Almost exactly the same as https://github.com/artsy/eigen/blob/main/src/app/Scenes/Home/Components/NewWorksForYouRail.tsx

src/app/Scenes/HomeView/HomeView.tsx
olerichter00: Could we, instead of using the key, use the [HomeViewComponentType](https://github.com/artsy/metaphysics/blob/3390f03de49bc99765ba6d62cb283a1b2a085753/_schemaV2.graphql#L11382) to decide which component to render? Like this, we would only need an "Artwork Rail" component instead of a dedicated "New works for you" component. This approach would allow us to easily create new artwork sections on the home view without the need to update the client.

olerichter00: With some additional fields/attributes like `size` or `variant` in the schema, we could then determine how a section is rendered.

nickskalkin: Yes, @olerichter00, this totally makes sense. I will try to do it today, otherwise @MounirDhahri maybe you could finish it when you are back

joeyAghion: (Just discussed in person.) I agree that clients should ideally switch between _types_ of sections, not the individual section identifiers. That gives us flexibility to introduce new or different instances of common data shapes (lists of artworks, individual artworks, activity, lists of artists, static content, etc.).

If a client doesn't have the latest-and-greatest implementation of a component, it can still handle the server-driven data as long as it knows _a way_ to handle that shape.

src/app/Scenes/HomeView/Sections/ArtworksRailHomeViewSection.tsx
```diff
@@ -0,0 +1,64 @@
...
+  section: ArtworksRailHomeViewSection_section$key
+}
+
+export const ArtworksRailHomeViewSection: React.FC<ArtworksRailHomeViewSectionProps> = ({
+  section,
+}) => {
+  const data = useFragment(fragment, section)
+  const title = data.title
+  const artworks = extractNodes(data.artworksConnection)
+  const componentHref = "" // TODO: should be in schema
```
nickskalkin, line 20: MP doesn't support this field yet - we should implement it soon

src/app/Scenes/HomeView/HomeView.tsx
nickskalkin: @olerichter00 @joeyAghion now I'm branching based on section type (ArtworksRail, ArtistsRail, etc): https://github.com/artsy/eigen/pull/10524/commits/db0c3041d3db1a4892bbe5335f9bb47c5abebc7c

Screen looks "ugly" now because MP returns dummy data for other artwork-based rails:

olerichter00: Looks great! The new `ArtworksRailHomeViewSection` could check for artworks in the `artworkConnection` and not render anything if there are no artworks (Make sure not to use `ItemSeparator` if items can return `null` but add bottom padding to all items to avoid double spacing between items) 🤔

src/app/Scenes/HomeView/HomeView.tsx
araujobarret: probably it comes from the safe area insets, you can use the prop `safeArea={false}`, I think it's more accurate

data/schema.graphql
```diff
@@ -10736,17 +10723,11 @@ type HomeView {
...
-  # Which component type to render
-  type: HomeViewComponentType
-}
-
-# Known component types for use in home view
-enum HomeViewComponentType {
-  # A standard rail of artists
-  ARTISTS_RAIL
+  # A URL to navigate to when clicked
+  href: String
```
dblandin, line 10727: This `href` field doesn't yet exist in the schema. Might it be something you added locally?

src/app/Scenes/HomeView/HomeView.tsx
```diff
@@ -1,62 +1,30 @@
...
-      <Screen.StickySubHeader
-        title={SCREEN_TITLE}
-        separatorComponent={<Separator borderColor="black5" />}
-      />
-
       <Screen.Body fullwidth>
         <Screen.FlatList
           data={sections}
-          keyExtractor={(item) => `${item.title}`}
+          keyExtractor={(item) => `${item.internalID || ""}`}
```
dblandin, line 17: This `internalID` should always be present as it's defined with `GraphQLNonNull` [here](https://github.com/artsy/metaphysics/blob/8e21ecbaac17ea3b45885aa3f3ccc351e5a79215/src/schema/v2/object_identification.ts#L231-L238). Is this `|| ""` bit necessary?

src/app/Scenes/HomeView/HomeView.tsx
```diff
@@ -72,16 +40,18 @@ export const HomeViewScreen: React.FC = () => (
...
           cursor
           node {
+            __typename
             ... on GenericHomeViewSection {
-              key
-              title
-              component {
-                type
-              }
+              internalID
```
dblandin, line 49: **question**: I'm still unclear if we need to repeat `internalID` under the fragment spreads if we're including it for each node above.

src/app/Scenes/HomeView/Sections/ArtworksRailHomeViewSection.tsx
```diff
@@ -0,0 +1,63 @@
...
+}
+
+export const ArtworksRailHomeViewSection: React.FC<ArtworksRailHomeViewSectionProps> = ({
+  section,
+}) => {
+  const data = useFragment(fragment, section)
+
+  const artworks = extractNodes(data.artworksConnection)
+
+  if (!data.component || !artworks || artworks.length === 0) {
```
dblandin, line 21: **comment / non-blocking**: This `component` property is currently nullable but I think we should make it non-null at the schema level to reduce the need for checks like this.

I wonder also if we can query for and rely on the totalCount from the artworksConnection as another check to whether we should skip the rendering of this section?

src/app/Scenes/HomeView/Sections/Section.tsx
```diff
@@ -0,0 +1,27 @@
...
+  switch (section.__typename) {
+    case "ArtworksRailHomeViewSection":
+      return <ArtworksRailHomeViewSection section={section} />
+    case "GenericHomeViewSection":
+      return <GenericHomeViewSection section={section} />
+    default:
+      if (__DEV__) {
+        return <Text>Non supported section: {section.__typename}</Text>
+      }
+      return null
```
dblandin, line 25: **praise**: Nice 👏🏼

src/app/Scenes/HomeView/Sections/ArtworksRailHomeViewSection.tsx
MounirDhahri: Good thought about improving the component nullability - I will PR that in MP

MounirDhahri: https://github.com/artsy/metaphysics/pull/5877

src/app/Scenes/HomeView/HomeView.tsx
araujobarret: isn't the `internalID` the `id` in reality?
image

src/app/Scenes/HomeView/Sections/ArtworksRailHomeViewSection.tsx
araujobarret: what's the logic of componentHref? the link to the resource on eigen? 
If yes, we should be able to understand the type of component to render and mount it's URI inside Eigen itself, not coming from MP if I get it right

src/app/Scenes/HomeView/HomeView.tsx
dblandin: @araujobarret The value is the same, yes. I'm curious about the nullability given the union interface used here. Would rather this be non-nullable so that's something to explore.

src/app/AppRegistry.tsx
```diff
@@ -488,7 +488,7 @@ export const modules = defineModules({
     hidesBackButton: true,
   }),
   Home: reactModule(HomeContainer, {
-    isRootViewForTabName: "home",
```
brainbicycle, line 491: was this removed on purpose? this will make it so tab doesn't switch when navigating to home

PR #10551, opened 2024-08-02
feat: add home view artists rail

MounirDhahri:
This PR resolves [ONYX-1167] 

### Description
This PR adds the artists generic rail to `Section`. This rail's goal is to render `ArtistsRailHomeViewSection`.

https://github.com/user-attachments/assets/f35cda56-9c0b-4083-9478-4ca6e4269723

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[ONYX-1167]: https://artsyproduct.atlassian.net/browse/ONYX-1167?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Components/Home/ArtistRails/ArtistCard.tsx
+++ b/src/app/Components/Home/ArtistRails/ArtistCard.tsx
@@ -16,7 +16,7 @@ interface ArtistCardProps {
   onPress?: () => void
 }
 
-const IMAGE_MAX_HEIGHT = 180
+export const IMAGE_MAX_HEIGHT = 180
 
 export const ArtistCard: React.FC<ArtistCardProps> = ({ artist, onDismiss, onFollow, onPress }) => {
   const color = useColor()
--- a/src/app/Components/Home/ArtistRails/ArtistCard.tsx
+++ b/src/app/Components/Home/ArtistRails/ArtistCard.tsx
@@ -1,5 +1,6 @@
 import { CloseIcon, Flex, FollowButton, Text, Touchable, useColor } from "@artsy/palette-mobile"
 import { ArtistCard_artist$data } from "__generated__/ArtistCard_artist.graphql"
+import { useFollowArtist } from "app/Components/Artist/useFollowArtist"
 import { OpaqueImageView } from "app/Components/OpaqueImageView2"
 
 import { navigate } from "app/system/navigation/navigate"
@@ -14,12 +15,29 @@ interface ArtistCardProps {
   onDismiss?: () => void
   onFollow?: () => void
   onPress?: () => void
+  // Use the default follow button instead of the injected onFollow button
+  showDefaultFollowButton?: boolean
 }
 
 export const IMAGE_MAX_HEIGHT = 180
 
-export const ArtistCard: React.FC<ArtistCardProps> = ({ artist, onDismiss, onFollow, onPress }) => {
+export const ArtistCard: React.FC<ArtistCardProps> = ({
+  artist,
+  onDismiss,
+  onFollow,
+  onPress,
+  showDefaultFollowButton = false,
+}) => {
   const color = useColor()
+  const { handleFollowToggle } = useFollowArtist(artist)
+
+  if (__DEV__) {
+    if (showDefaultFollowButton && onFollow) {
+      console.warn(
+        "ArtistCard: onFollow and showDefaultFollowButton are both set, onFollow will be ignored"
+      )
+    }
+  }
 
   const artistImages = extractNodes(artist.filterArtworksConnection)
     .filter((artwork) => {
@@ -56,11 +74,15 @@ export const ArtistCard: React.FC<ArtistCardProps> = ({ artist, onDismiss, onFol
               </Text>
             )}
           </Flex>
-          {!!onFollow && (
-            <Flex>
-              <FollowButton isFollowed={!!artist.isFollowed} onPress={onFollow} />
-            </Flex>
-          )}
+          {!!onFollow ||
+            (showDefaultFollowButton && (
+              <Flex>
+                <FollowButton
+                  isFollowed={!!artist.isFollowed}
+                  onPress={handleFollowToggle ?? onFollow}
+                />
+              </Flex>
+            ))}
         </Flex>
         {!!onDismiss && (
           <Flex
@@ -190,6 +212,7 @@ export const ArtistCardContainer = createFragmentContainer(ArtistCard, {
           }
         }
       }
+      ...useFollowArtist_artist
     }
   `,
 })
--- a/src/app/Components/Home/ArtistRails/RecommendedArtistsRail.tsx
+++ b/src/app/Components/Home/ArtistRails/RecommendedArtistsRail.tsx
@@ -85,7 +85,7 @@ export const RecommendedArtistsRail: React.FC<RecommendedArtistsRailProps & Rail
       </Flex>
       <CardRailFlatList<ArtistCard_artist$data>
         listRef={listRef}
-        prefetchUrlExtractor={(item) => item?.href || undefined}
+        prefetchUrlExtractor={(item) => item?.href}
         prefetchVariablesExtractor={defaultArtistVariables}
         data={artists as any}
         keyExtractor={(artist) => artist.id}
--- a/src/app/Components/Home/ArtistRails/RecommendedArtistsRail.tsx
+++ b/src/app/Components/Home/ArtistRails/RecommendedArtistsRail.tsx
@@ -85,7 +85,7 @@ export const RecommendedArtistsRail: React.FC<RecommendedArtistsRailProps & Rail
       </Flex>
       <CardRailFlatList<ArtistCard_artist$data>
         listRef={listRef}
-        prefetchUrlExtractor={(item) => item?.href!}
+        prefetchUrlExtractor={(item) => item?.href || undefined}
         prefetchVariablesExtractor={defaultArtistVariables}
         data={artists as any}
         keyExtractor={(artist) => artist.id}
@@ -180,7 +180,7 @@ export const tracks = {
   }),
 }
 
-const followOrUnfollowArtist = (followArtist: ArtistCard_artist$data) => {
+export const followOrUnfollowArtist = (followArtist: ArtistCard_artist$data) => {
   return new Promise<void>((resolve, reject) => {
     commitMutation<RecommendedArtistsRailFollowMutation>(getRelayEnvironment(), {
       mutation: graphql`
--- a/src/app/Scenes/HomeView/Sections/ArtistsRailHomeViewSection.tsx
+++ b/src/app/Scenes/HomeView/Sections/ArtistsRailHomeViewSection.tsx
@@ -1,41 +1,47 @@
 import { Flex, Spacer, Spinner } from "@artsy/palette-mobile"
-import {
-  ArtistsRailHomeViewSection_section$data,
-  ArtistsRailHomeViewSection_section$key,
-} from "__generated__/ArtistsRailHomeViewSection_section.graphql"
+import { ArtistsRailHomeViewSection_section$data } from "__generated__/ArtistsRailHomeViewSection_section.graphql"
 import {
   IMAGE_MAX_HEIGHT as ARTIST_RAIL_IMAGE_MAX_HEIGHT,
   ArtistCardContainer,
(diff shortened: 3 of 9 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against a788a025da34303482c4d1941190a678bb8f4909
</p>

src/app/Components/Home/ArtistRails/ArtistCard.tsx
```diff
@@ -16,7 +16,7 @@ interface ArtistCardProps {
   onPress?: () => void
 }
 
-const IMAGE_MAX_HEIGHT = 180
+export const IMAGE_MAX_HEIGHT = 180
```
olerichter00, line 19: Might be good to choose a name related to the artist rail if we export it. Could be something like:

```suggestion
export const ARTIST_RAIL_MAX_IMAGE_SIZE = 180
```

src/app/Components/Home/ArtistRails/RecommendedArtistsRail.tsx
```diff
@@ -85,7 +85,7 @@ export const RecommendedArtistsRail: React.FC<RecommendedArtistsRailProps & Rail
       </Flex>
       <CardRailFlatList<ArtistCard_artist$data>
         listRef={listRef}
-        prefetchUrlExtractor={(item) => item?.href!}
+        prefetchUrlExtractor={(item) => item?.href || undefined}
```
olerichter00, line 88: Instead of adding `|| undefined`, we can probably also accept `null` [here](https://github.com/artsy/eigen/blob/db2e0f09ed52ba8927fb0fa2b9a054a9fb3491a9/src/app/Components/PrefetchFlatList.tsx#L11).

src/app/Components/Home/ArtistRails/RecommendedArtistsRail.tsx
```diff
@@ -180,7 +180,7 @@ export const tracks = {
   }),
 }
 
-const followOrUnfollowArtist = (followArtist: ArtistCard_artist$data) => {
+export const followOrUnfollowArtist = (followArtist: ArtistCard_artist$data) => {
```
olerichter00, line 183:
```suggestion
const followOrUnfollowArtist = (followArtist: ArtistCard_artist$data) => {
```

src/app/Scenes/HomeView/Sections/ArtistsRailHomeViewSection.tsx
```diff
@@ -0,0 +1,107 @@
...
+              )
+            }
+            return null
+          }}
+          renderItem={({ item: artist, index }) => {
+            return (
+              <View style={{ flexDirection: "row" }}>
+                <ArtistCardContainer
+                  artist={artist}
+                  // onFollow={() => followOrUnfollowArtist(artist)}
```
olerichter00, line 75: 👀

src/app/Components/Home/ArtistRails/RecommendedArtistsRail.tsx
olerichter00: https://github.com/artsy/eigen/pull/10551/files#diff-26a95c958ab313802cb7c4d321bccd0d993c935cd97d0d32acae54b4ac9975a8R75

src/app/Scenes/HomeView/Sections/ArtistsRailHomeViewSection.tsx
```diff
@@ -0,0 +1,107 @@
...
+        <CardRailFlatList<Artist>
+          data={artists}
+          keyExtractor={(artist) => artist.internalID}
+          // I noticed that sometimes FlatList seemed to get confused about where cards should be
+          // and making this explicit fixes that.
+          getItemLayout={(_data, index) => ({
+            index,
+            offset: index * (CARD_WIDTH + INTER_CARD_PADDING),
+            length: CARD_WIDTH + INTER_CARD_PADDING,
+          })}
```
olerichter00, line 53: Is this something we could also do in `CardRailFlatList.tsx`? Looks like something we should apply everywhere and not here 🤔

src/app/Scenes/HomeView/Sections/ArtistsRailHomeViewSection.tsx
```diff
@@ -0,0 +1,107 @@
...
+          onEndReached={onEndReached}
+          ListFooterComponent={() => {
+            if (!!hasNext && isLoadingNext) {
```
olerichter00, line 56:
```suggestion
            if (hasNext && isLoadingNext) {
```

src/app/Scenes/HomeView/Sections/ArtistsRailHomeViewSection.tsx
```diff
@@ -0,0 +1,107 @@
...
+                  width={IMAGE_MAX_HEIGHT / 2}
+                  height={IMAGE_MAX_HEIGHT}
+                  justifyContent="center"
+                  alignItems="center"
+                >
+                  <Spinner />
+                </Flex>
+              )
+            }
+            return null
```
olerichter00, line 68:
```suggestion
            }

            return null
```

src/app/Scenes/HomeView/Sections/ArtistsRailHomeViewSection.tsx
```diff
@@ -0,0 +1,107 @@
...
+          renderItem={({ item: artist, index }) => {
+            return (
+              <View style={{ flexDirection: "row" }}>
+                <ArtistCardContainer
+                  artist={artist}
+                  // onFollow={() => followOrUnfollowArtist(artist)}
+                />
+                {index === artists.length - 1 ? null : (
+                  <View style={{ width: INTER_CARD_PADDING }} />
+                )}
```
olerichter00, line 79: Would this be accurate / equivalent? 

```suggestion
                {(index < artists.length - 1)  && <View style={{ width: INTER_CARD_PADDING }} />}

```

src/app/Scenes/HomeView/Sections/ArtistsRailHomeViewSection.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+type Artist = ExtractNodeType<ArtistsRailHomeViewSection_section$data["artistsConnection"]>
+export const ArtistsRailHomeViewSection: React.FC<ArtworksRailHomeViewSectionProps> = ({
+  section,
+}) => {
+  const { data, hasNext, loadNext, isLoadingNext } = usePaginationFragment(artistsFragment, section)
+
+  const title = data.component?.title
+
+  const onEndReached = useCallback(() => {
```
araujobarret, line 28: don't think we need the `useCallback` here since we don't use `onEndReached` in any `useEffect`

src/app/Components/Home/ArtistRails/RecommendedArtistsRail.tsx
araujobarret: I think this shouldn't be here anymore, [idea] we could move it to: 
```
srr/app/Components/Home/mutations/followOrUnfollowArtist.ts
```

src/app/Components/Home/ArtistRails/ArtistCard.tsx
MounirDhahri: I feel like it's kind of clear here since it's already insde the `ArtistCard` component. I will update it in the import thought to be something like
```typescript
import { IMAGE_MAX_HEIGHT as ARTIST_RAIL_MAX_IMAGE_SIZE } from ...
```

src/app/Components/Home/ArtistRails/RecommendedArtistsRail.tsx
MounirDhahri: we do actually have a helper for that in [here](https://github.com/artsy/eigen/blob/f993f231211b0797e2755ca00966522442fd9843/src/app/utils/mutations/useFollowArtist.ts#L3)
https://github.com/artsy/eigen/blob/f993f231211b0797e2755ca00966522442fd9843/src/app/utils/mutations/useFollowArtist.ts#L3
but I am avoiding making big changes here because we are not keeping any of the rails and mainly trying to hook the data as an example on how to consume the new HomeView API

MounirDhahri (commented):
thanks for the early comments - I pushed this code last week to make sure it can make it to the Home View updates call before wrapping up the PR.

src/app/Scenes/HomeView/Sections/ArtistsRailHomeViewSection.tsx
olerichter00: I agree if this is a hack/fix specifically for only this component. But if it is not, I would suggest applying it in the `CardRailFlatList`. If we don't, wouldn't the next one using `CardRailFlatList` most likely forget to test for this bug and add the fix 🤔

src/app/Scenes/HomeView/Sections/ArtistsRailHomeViewSection.tsx
```diff
@@ -0,0 +1,92 @@
...
+  return (
+    <Flex>
+      <Flex pl={2} pr={2}>
+        <SectionTitle title={title} />
+      </Flex>
+      <CardRailFlatList<Artist>
+        data={artists}
+        keyExtractor={(artist) => artist.internalID}
+        onEndReached={onEndReached}
+        ItemSeparatorComponent={() => <Spacer x={`${INTER_CARD_PADDING}px`} />}
```
araujobarret, line 49: We shouldn't be using this `INTER_CARD_PADDING`(sorry was hard to follow up all the PRs once this was introduced) at all, we have our spacing defined in the DS level and we should use it instead of abstractions on top of it.
Another important thing is the value of this variable is not fine, it should be 10 or 20, we don't support 15.

src/app/Scenes/HomeView/Sections/ArtistsRailHomeViewSection.tsx
```diff
@@ -0,0 +1,107 @@
...
+            return null
+          }}
+          renderItem={({ item: artist, index }) => {
+            return (
+              <View style={{ flexDirection: "row" }}>
+                <ArtistCardContainer
+                  artist={artist}
+                  // onFollow={() => followOrUnfollowArtist(artist)}
+                />
+                {index === artists.length - 1 ? null : (
```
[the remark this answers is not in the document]
araujobarret: 😢

src/app/Scenes/HomeView/Sections/ArtistsRailHomeViewSection.tsx
MounirDhahri: that's a good point - I totally agree here.
I am trying to avoid making any changes to existing components because I don't have enough context on why it was made that way. This component is mainly a copy paste to recommended artists rails. Design is coming up with new rails soon

MounirDhahri: I will update the value to `1` though and hope design doesn't complain about it 😄

araujobarret: 🤞🤞🤞 
Usually, that's a smoke signal they've made the mockups inconsistent with our DS(using a spacing measure that doesn't exist)

araujobarret: that's the problem of reusing the existing elements which I mentioned before, it snowballs these elements' debt even more

src/app/Components/Home/ArtistRails/ArtistCard.tsx
olerichter00: I guess, it's ok like it is 👍 I'm just concerned about exporting things with a generic name because it could confuse editor's auto-import feature...

MounirDhahri:
Let's merge this and move to hooks later - @olerichter00 thanks for the fixes you made.

PR #10599, opened 2024-08-13
feat: add add my collection artists prompt

araujobarret:
This PR resolves [DIA-717] 

### Description

This PR introduces the Collected Artists prompt, which has not yet been wired up to any screens.
Main changes:
- Adds the BottomSheet prompt to Eigen 
- reuses some areas from Collected Artists 
- refactors the Collected Artists logic to share the code with the prompt

Android:

https://github.com/user-attachments/assets/05c8666b-fb31-41fc-affb-aafb79f2afd9

iOS 

https://github.com/user-attachments/assets/2ebe3b5e-3179-48c5-888c-983a88c5731f

### PR Checklist

### To the reviewers 👀

- feat: add collected artists prompt(not wiring it to any screen yet) - araujobarret

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[DIA-717]: https://artsyproduct.atlassian.net/browse/DIA-717?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/MyCollection/Components/MyCollectionArtistsPrompt/MyCollectionArtistsAutosuggest.tsx
+++ b/src/app/Scenes/MyCollection/Components/MyCollectionArtistsPrompt/MyCollectionArtistsAutosuggest.tsx
@@ -1,0 +1,99 @@
+import { Flex, Input, Spacer, useScreenDimensions, useSpace } from "@artsy/palette-mobile"
+import { useBottomSheet } from "@gorhom/bottom-sheet"
+import { MyCollectionArtistsAutosuggestQuery } from "__generated__/MyCollectionArtistsAutosuggestQuery.graphql"
+import SearchIcon from "app/Components/Icons/SearchIcon"
+import { MyCollectionArtistsAutosuggestItem } from "app/Scenes/MyCollection/Components/MyCollectionArtistsPrompt/MyCollectionArtistsAutosuggestItem"
+import { MyCollectionArtistsAutosuggestListHeader } from "app/Scenes/MyCollection/Components/MyCollectionArtistsPrompt/MyCollectionArtistsAutosuggestListHeader"
+import { FOOTER_HEIGHT } from "app/Scenes/MyCollection/Components/MyCollectionArtistsPrompt/MyCollectionArtistsPromptFooter"
+import { extractNodes } from "app/utils/extractNodes"
+import { useDebouncedValue } from "app/utils/hooks/useDebouncedValue"
+import { FC, useState } from "react"
+import { FlatList, Keyboard, KeyboardAvoidingView } from "react-native"
+import { graphql, useLazyLoadQuery } from "react-relay"
+
+export const MyCollectionArtistsAutosuggest: FC = () => {
+  const space = useSpace()
+  const {
+    safeAreaInsets: { bottom },
+  } = useScreenDimensions()
+  const { expand, collapse, animatedIndex } = useBottomSheet()
+  const [query, setQuery] = useState("")
+
+  const { debouncedValue: debouncedQuery } = useDebouncedValue({ value: query, delay: 250 })
+
+  const data = useLazyLoadQuery<MyCollectionArtistsAutosuggestQuery>(QUERY, {
+    query: debouncedQuery,
+  })
+
+  const matches = extractNodes(data?.matchConnection)
+
+  const handleOnFocus = () => {
+    if (animatedIndex.value === 0) {
+      expand()
+    }
+  }
+
+  const handleOnBlur = () => {
+    if (matches.length === 0) {
+      collapse()
+    }
+  }
+
+  const handleOnClear = () => {
+    if (!Keyboard.isVisible()) {
+      collapse()
+    }
+  }
+
+  return (
+    <KeyboardAvoidingView behavior="padding">
+      <Flex gap={space(2)} height="100%">
+        <Flex>
+          <Input
+            placeholder="Search for artists on Artsy"
+            icon={<SearchIcon width={18} height={18} />}
+            enableClearButton
+            autoCorrect={false}
+            value={query}
+            onChangeText={setQuery}
+            onClear={handleOnClear}
+            onFocus={handleOnFocus}
+            onBlur={handleOnBlur}
+          />
+        </Flex>
+
+        <FlatList
+          data={matches}
+          keyExtractor={(item) => `MyCollectionArtistsAutosuggestItem_${item.internalID}`}
+          renderItem={({ item }) => <MyCollectionArtistsAutosuggestItem artist={item} />}
+          ListHeaderComponent={() => (
+            <MyCollectionArtistsAutosuggestListHeader
+              query={query}
+              debouncedQuery={debouncedQuery}
+              resultsLength={matches.length}
+              isLoading={!data}
+            />
+          )}
+          ListFooterComponent={() => <Flex height={FOOTER_HEIGHT + bottom} />}
+          ItemSeparatorComponent={() => <Spacer y={2} />}
+        />
+      </Flex>
+    </KeyboardAvoidingView>
+  )
+}
+
+const QUERY = graphql`
+  query MyCollectionArtistsAutosuggestQuery($query: String!) {
+    matchConnection(term: $query, entities: ARTIST, mode: AUTOSUGGEST, first: 7) {
+      edges {
+        node {
+          ...MyCollectionArtistsAutosuggestItem_artist
+          __typename
+          ... on Artist {
+            internalID @required(action: NONE)
+          }
+        }
+      }
+    }
+  }
+`
--- a/src/app/Scenes/MyCollection/Components/MyCollectionArtistsPrompt/MyCollectionArtistsAutosuggestItem.tsx
+++ b/src/app/Scenes/MyCollection/Components/MyCollectionArtistsPrompt/MyCollectionArtistsAutosuggestItem.tsx
@@ -1,0 +1,79 @@
+import { Avatar, Button, CheckIcon, Flex, Text, useSpace } from "@artsy/palette-mobile"
+import { MyCollectionArtistsAutosuggestItem_artist$key } from "__generated__/MyCollectionArtistsAutosuggestItem_artist.graphql"
+import { MyCollectionAddCollectedArtistsStore } from "app/Scenes/MyCollection/Screens/MyCollectionAddCollectedArtists/MyCollectionAddCollectedArtistsStore"
+import { MotiView } from "moti"
+import { FC } from "react"
+import { graphql, useFragment } from "react-relay"
+
+interface MyCollectionArtistsAutosuggestItemProps {
+  artist: MyCollectionArtistsAutosuggestItem_artist$key
+}
+
+export const MyCollectionArtistsAutosuggestItem: FC<MyCollectionArtistsAutosuggestItemProps> = ({
+  artist,
+}) => {
+  const space = useSpace()
+  const data = useFragment(QUERY, artist)
+
+  const artistIds = MyCollectionAddCollectedArtistsStore.useStoreState((state) => state.artistIds)
+  const addOrRemoveArtist = MyCollectionAddCollectedArtistsStore.useStoreActions(
+    (actions) => actions.addOrRemoveArtist
+  )
+
+  if (!data) {
+    return null
+  }
+
+  const selected = artistIds.includes(data.internalID)
+  const displayAvatar = !!data.coverArtwork?.image?.url || !!data.initials
+
+  const handleOnPress = () => {
+    addOrRemoveArtist(data.internalID)
+  }
+
+  return (
+    <MotiView
+      from={{ opacity: 0 }}
+      animate={{ opacity: 1 }}
+      transition={{ type: "timing", duration: 200 }}
+    >
+      <Flex flexDirection="row" justifyContent="space-between" alignItems="center">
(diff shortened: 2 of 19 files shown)


src/app/Scenes/MyCollection/Screens/MyCollectionAddCollectedArtists/MyCollectionAddCollectedArtists.tsx
```diff
@@ -1,108 +1,29 @@
...
+  const { submit, isSubmitting: isLoading } = useSubmitMyCollectionArtists(
+    "MyCollectionAddCollectedArtists"
+  )
 
   const artistIds = MyCollectionAddCollectedArtistsStore.useStoreState((state) => state.artistIds)
   const customArtists = MyCollectionAddCollectedArtistsStore.useStoreState(
     (state) => state.customArtists
   )
 
-  const addUserInterests = async () => {
```
araujobarret, line 30: no changes here so far, just moved the logic to a hook to be used by the prompt too

ArtsyOpenSource:

- Cross-platform user-facing changes (feat: add collected artists prompt(not wiring it to any screen yet) - araujobarret)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against ea822d3a8cf9239a5702c5a27000f79230a8fef7
</p>

src/app/Scenes/MyCollection/Screens/MyCollectionAddCollectedArtists/MyCollectionAddCollectedArtistsAutosuggest.tsx
```diff
@@ -159,7 +159,7 @@ export const MyCollectionAddCollectedArtistsAutosuggest: React.FC<{}> = ({}) =>
 const myCollectionAddCollectedArtistsAutosuggestQuery = graphql`
   query MyCollectionAddCollectedArtistsAutosuggestQuery {
     me {
-      userInterestsConnection(first: 100) {
+      userInterestsConnection(first: 100, interestType: ARTIST) {
```
araujobarret, line 162: autosuggest here should be filtering the interest type by Artist

src/app/Scenes/MyCollection/Screens/MyCollectionAddCollectedArtists/MyCollectionAddCollectedArtists.tsx
```diff
@@ -1,108 +1,26 @@
...
-      })
-    )
-
-    const numberOfFailedResults = results.filter((result) => !result).length
-
-    if (numberOfFailedResults > 0) {
-      toast.show("Some artists could not be created.", "bottom", { backgroundColor: "red100" })
-    }
-  }
+  const count = MyCollectionAddCollectedArtistsStore.useStoreState((state) => state.count)
```
araujobarret, line 20: only change so far, moved the count to be a computed property of the store rather than derivating it here

src/app/Scenes/MyCollection/Components/MyCollectionArtistsPrompt/MyCollectionArtistsAutosuggestItem.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+interface MyCollectionArtistsAutosuggestItemProps {
+  artist: MyCollectionArtistsAutosuggestItem_artist$key
+}
+
+export const MyCollectionArtistsAutosuggestItem: FC<MyCollectionArtistsAutosuggestItemProps> = ({
+  artist,
+}) => {
+  const space = useSpace()
+  const data = useFragment(QUERY, artist)
```
olerichter00, line 16: Suggestion to be able to call the artist from the fragment `artist`. This might be easier to read than the more generic `data`.

```suggestion
export const MyCollectionArtistsAutosuggestItem: FC<MyCollectionArtistsAutosuggestItemProps> = (props) => {
  const space = useSpace()
  const artist = useFragment(QUERY, props.artist)
```

src/app/Scenes/MyCollection/Components/MyCollectionArtistsPrompt/MyCollectionArtistsPromptBody.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { refreshMyCollection } from "app/utils/refreshHelpers"
+import { FC } from "react"
+
+export const MyCollectionArtistsPromptBody: FC = () => {
+  const toast = useToast()
+  const { submit, isSubmitting } = useSubmitMyCollectionArtists("MyCollectionArtistsPrompt")
+  const { close } = useBottomSheet()
+
+  const handleSubmit = async () => {
+    const response = await submit()
```
olerichter00, line 16: What happens if submitting the data fails? Can we wrap this in try-catch-finally blocks and show a toast in case submitting the data fails?

olerichter00: Ah, catching errors already happens in `useSubmitMyCollectionArtists`, right?

src/app/Scenes/MyCollection/Components/MyCollectionArtistsPrompt/MyCollectionArtistsAutosuggest.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { MyCollectionArtistsAutosuggestItem } from "app/Scenes/MyCollection/Components/MyCollectionArtistsPrompt/MyCollectionArtistsAutosuggestItem"
+import { MyCollectionArtistsAutosuggestListHeader } from "app/Scenes/MyCollection/Components/MyCollectionArtistsPrompt/MyCollectionArtistsAutosuggestListHeader"
+import { FOOTER_HEIGHT } from "app/Scenes/MyCollection/Components/MyCollectionArtistsPrompt/MyCollectionArtistsPromptFooter"
+import { extractNodes } from "app/utils/extractNodes"
+import { useDebouncedValue } from "app/utils/hooks/useDebouncedValue"
+import { FC, useState } from "react"
+import { FlatList, Keyboard, KeyboardAvoidingView } from "react-native"
+import { graphql, useLazyLoadQuery } from "react-relay"
+
+export const MyCollectionArtistsAutosuggest: FC = () => {
```
olerichter00, line 14: Would it be possible to use the already existing [ArtistAutosuggest](https://github.com/artsy/eigen/blob/c1656fdfed905512e8ddff7615eb7e17fdb5e472/src/app/Components/ArtistAutosuggest/ArtistAutosuggest.tsx#L16) and re-use the code?

src/app/Scenes/MyCollection/Components/MyCollectionArtistsPrompt/MyCollectionArtistsPromptBody.tsx
araujobarret: Yep, I didn't change any behavior already present only moved them to the hook to make it shareable between the prompt and the existing screen.
But the way it is set the functions called inside `submit`, respectively `createArtists` and `addUserInterestsArtists` handle the errors internally so `submit` doesn't need to care. Here we only handle the success case if we have a `response`.
`createUser` error https://github.com/artsy/eigen/pull/10599/files#diff-8cd6fc2d95de4deaeb045e828da47611054a706d66667f7644474de12c1bb468R35
`addUserInterestsArtists` errors https://github.com/artsy/eigen/pull/10599/files#diff-8fa3b3f92bf0e1989f64d429143339aad210ef03e5dba5e425055647b9066e58R37-R48

src/app/Scenes/MyCollection/Components/MyCollectionArtistsPrompt/MyCollectionArtistsAutosuggestItem.tsx
araujobarret: I think data is an okay pattern, we have it like this in our workshop docs, all apps use the same pattern calling the `key` the type name and it is like Relay uses in their docs, example https://relay.dev/docs/api-reference/use-fragment/#usefragment

src/app/Scenes/MyCollection/Components/MyCollectionArtistsPrompt/MyCollectionArtistsAutosuggest.tsx
araujobarret: I tried all that I could, so I reused the stores and shared the behavior via hooks.
The UI elements are different, the list items, for example, and the input interaction with the bottom sheet. Abstracting the existing component would make it a little monster and hard to maintain.

src/app/Scenes/MyCollection/Components/MyCollectionArtistsPrompt/MyCollectionArtistsAutosuggestItem.tsx
olerichter00: The pattern is ok and working, but in my opinion, it has two significant drawbacks:

1. It creates unnecessary ambiguity when accessing the data further down in the component, making it more difficult to understand. (e.g. `data.internalID` vs `artist.internalID`).
2. Refactoring/renaming is needed when adding more fragment to the component ([example](https://github.com/artsy/eigen/blob/12b44d2e419b55e1ae95466df7b6eac870e1a056/src/app/Scenes/MyCollection/Screens/Artwork/Components/ArtworkInsights/RequestForPriceEstimate/RequestForPriceEstimateBanner.tsx#L23)).

olerichter00: In Eigen's codebase, there can be found a great mix of both patterns: 59 x `data = useFragment`, 54 x `somethingData = useFragment`` and 65 x `something = useFragment`.

src/app/Scenes/MyCollection/hooks/useCreateArtists.ts
```diff
@@ -0,0 +1,40 @@
...
+          return await createArtist({
+            displayName: customArtist.name,
+            birthday: customArtist.birthYear,
+            deathday: customArtist.deathYear,
+            nationality: customArtist.nationality,
+            isPersonalArtist: true,
+          })
+        } catch (error) {
+          return null
+        }
```
egdbear, line 28: **nitpick:** readability -- this loop could go before the deferring.

olerichter00: You could use `Promise.allSettled` or `Promise.all` here instead, which would be cleaner and easier to understand, and more performant because the requests would run in parallel (https://www.w3schools.com/jsref/jsref_promise_allsettled.asp).

egdbear: +1 for Promise.all. Although, this is old code that’s just been moved here, I don’t want to make this blocking. But since we’re already at it, it might be worth giving it some love.

src/app/Scenes/MyCollection/Components/MyCollectionArtistsPrompt/MyCollectionArtistsAutosuggest.tsx
araujobarret: I will rename all the files bc will be inconsistent if I do only for this one, it'll be come something like:
```
MyCollectioArtistsPropmt
|--Autosuggest
|--AutosuggestItem
|--AutosuggestListHeader
```
is this pattern better?

olerichter00: Hmm, I was thinking more of removing anything related to My Collection from the name and moving it to `src/app/Components`. Similar to the general [ArtistAutosuggest component](https://github.com/artsy/eigen/blob/main/src/app/Components/ArtistAutosuggest/ArtistAutosuggest.tsx#L16), which is used in My Collection but also in other places. 

If this component is very specific to and integrated in My Collection and cannot be used elsewhere, then I would be okay with leaving it as it is.

src/app/Scenes/MyCollection/hooks/useCreateArtists.ts
araujobarret: I will try but since it's unrelated to this PR if it requires a lot of testing I won't change it 😬

src/app/Scenes/MyCollection/Components/MyCollectionArtistsPrompt/MyCollectionArtistsAutosuggest.tsx
araujobarret: That's not a generic component for sure, very specific towards MyCollection + BottomSheet, so I think it's better to leave it as it is for now since it is accommodated in the rightful place (`MyCollectioArtistsPrompt`), thoughts?

olerichter00: Thanks for clarifying 🙏  Let's leave it as it is then 👍

olerichter00 (approved):

src/app/Scenes/MyCollection/hooks/useCreateArtists.ts
araujobarret: nice addition, the testing was simple and it is a nice feature to use!

PR #10601, opened 2024-08-13
fix: fix screens tracking on submission flow

dariakoko:
This PR resolves [] 

### Description

Resolved [Notion card](https://www.notion.so/artsy/App-When-we-click-on-Add-Additional-Information-events-for-all-previous-screens-are-fired-d8cec042ab2044e99845e85bcf15cea1?pvs=4): When we click on Add Additional Information, events for all previous screens are fired

Track every screen using UseEffect

https://github.com/user-attachments/assets/93001122-63e7-4cc4-8944-90f1e568a0c6

### PR Checklist

### To the reviewers 👀

- fix screens tracking on submission flow -daria

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Scenes/SellWithArtsy/ArtworkForm/Components/SubmitArtworkAddDetails.tsx
+++ b/src/app/Scenes/SellWithArtsy/ArtworkForm/Components/SubmitArtworkAddDetails.tsx
@@ -8,14 +8,13 @@ import { SubmitArtworkStackNavigation } from "app/Scenes/SellWithArtsy/ArtworkFo
 import { useNavigationListeners } from "app/Scenes/SellWithArtsy/ArtworkForm/Utils/useNavigationListeners"
 import { useSubmissionContext } from "app/Scenes/SellWithArtsy/ArtworkForm/Utils/useSubmissionContext"
 import { SubmissionModel } from "app/Scenes/SellWithArtsy/ArtworkForm/Utils/validation"
-import { useSubmitArtworkTracking } from "app/Scenes/SellWithArtsy/Hooks/useSubmitArtworkTracking"
 import {
   AcceptableCategoryValue,
   acceptableCategoriesForSubmission,
 } from "app/Scenes/SellWithArtsy/SubmitArtwork/ArtworkDetails/utils/acceptableCategoriesForSubmission"
 import { createOrUpdateSubmission } from "app/Scenes/SellWithArtsy/SubmitArtwork/ArtworkDetails/utils/createOrUpdateSubmission"
 import { useFormikContext } from "formik"
-import { useEffect, useRef } from "react"
+import { useRef } from "react"
 import { ScrollView } from "react-native"
 
 export const SubmitArtworkAddDetails = () => {
@@ -26,14 +25,9 @@ export const SubmitArtworkAddDetails = () => {
   const setCurrentStep = SubmitArtworkFormStore.useStoreActions((actions) => actions.setCurrentStep)
   const navigation = useNavigation<NavigationProp<SubmitArtworkStackNavigation, "AddDetails">>()
 
-  const { currentStep } = useSubmissionContext()
-  const { trackSubmissionStepScreen } = useSubmitArtworkTracking()
+  const { currentStep, useSubmitArtworkScreenTracking } = useSubmissionContext()
 
-  useEffect(() => {
-    if (currentStep === "AddDetails") {
-      trackSubmissionStepScreen(currentStep, values.submissionId || undefined)
-    }
-  }, [currentStep])
+  useSubmitArtworkScreenTracking("AddDetails")
 
   useNavigationListeners({
     onNextStep: async () => {
--- a/src/app/Scenes/SellWithArtsy/ArtworkForm/Components/SubmitArtworkAddDetails.tsx
+++ b/src/app/Scenes/SellWithArtsy/ArtworkForm/Components/SubmitArtworkAddDetails.tsx
@@ -1,4 +1,3 @@
-import { OwnerType } from "@artsy/cohesion"
 import { Flex, Input, Join, Spacer, Text } from "@artsy/palette-mobile"
 import { NavigationProp, useNavigation } from "@react-navigation/native"
 import { SelectOption } from "app/Components/Select"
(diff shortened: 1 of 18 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (fix screens tracking on submission flow -daria)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 60471a6e14ffa7a04e0fe9af530191ecba7b2058
</p>

src/app/Scenes/SellWithArtsy/ArtworkForm/Components/SubmitArtworkAddDetails.tsx
```diff
@@ -9,26 +8,33 @@ import { SubmitArtworkStackNavigation } from "app/Scenes/SellWithArtsy/ArtworkFo
...
   const navigation = useNavigation<NavigationProp<SubmitArtworkStackNavigation, "AddDetails">>()
 
+  const { currentStep } = useSubmissionContext()
+  const { trackSubmissionStepScreen } = useSubmitArtworkTracking()
+
+  useEffect(() => {
+    if (currentStep === "AddDetails") {
+      trackSubmissionStepScreen(currentStep, values.submissionId || undefined)
+    }
+  }, [currentStep])
```
olerichter00, line 36: Great idea to use an effect to do the screen tracking! 🤩 

To avoid duplicating this code for every step, do you think we could move the effect into the context (while removing the `if (currentStep === CURRENT_STEP)` check)?

src/app/Scenes/SellWithArtsy/Hooks/useSubmitArtworkTracking.ts
```diff
@@ -145,6 +169,8 @@ const getOwnerType = (currentStep: SubmitArtworkScreen): OwnerType => {
       return OwnerType.submitArtworkStepAddPhoneNumber
     case "CompleteYourSubmission":
       return OwnerType.submitArtworkStepCompleteYourSubmission
+    case "CompleteYourSubmissionPostApproval":
+      return OwnerType.submitArtworkStepCompleteYourSubmissionPostApproval
```
olerichter00, line 173: To ensure we have a case for **every** step, we could use this trick to make TypeScript complain: [never type for exhaustive checks](https://graphite.dev/guides/typescript-switches#using-the-object-object-type-for-exhaustive-checks) ([Example](https://github.com/artsy/force/blob/aa445446f620f9925bd4d570d495f2e0e55a6f0c/src/Apps/Sell/Hooks/useSubmissionTracking.ts#L160)).

```
    default:
      // https://graphite.dev/guides/typescript-switches#using-the-object-object-type-for-exhaustive-checks
      const exhaustiveCheck: never = currentStep
      throw new Error(`Unhandled case: ${exhaustiveCheck}`)
```

olerichter00 (commented):
I like your solution! But it would be great if we could avoid the code duplication by moving the effect into the context.

src/app/Scenes/SellWithArtsy/ArtworkForm/Components/SubmitArtworkAddDetails.tsx
dariakoko: I would prefer to have it on the screen for visibility and I think if we move to the context the issue will not be solved

We need the `if (currentStep === CURRENT_STEP)` check to make sure we are firing  the tracking event only for this screen, otherwise it will be fired for all screen of the stack

src/app/Scenes/SellWithArtsy/Hooks/useSubmitArtworkTracking.ts
dariakoko: I think it should be fine as it is because we have this line: `(currentStep: SubmitArtworkScreen): OwnerType `
if I remove one of the cases it complains, and if I add a random case it complains as well
I guess, if we remove `| null` from [here](https://github.com/artsy/force/blob/aa445446f620f9925bd4d570d495f2e0e55a6f0c/src/Apps/Sell/Hooks/useSubmissionTracking.ts#L132) we won'e need this check either

src/app/Scenes/SellWithArtsy/ArtworkForm/Components/SubmitArtworkAddDetails.tsx
olerichter00: You're right, moving the effect to the context wouldn't work because the context does not know which step is currently rendered. 

To avoid the code duplication, which IMO, we should definitely do, we could extract the screen tracking code into a hook:

```
  const useSubmitArtworkScreenTracking = (step: keyof SubmitArtworkStackNavigation) => {
    useEffect(() => {
      if (currentStep === step) {
        trackSubmissionStepScreen(currentStep, values.submissionId || undefined)
      }
    }, [currentStep])
  }
```

and call it like this within the step component.
```
  useSubmitArtworkScreenTracking("AddTitle")
```

olerichter00 (commented):
I've tested the "Finish previous submission" flow and still see the "Artist Step" screen event firing even when continuing the submission from a later step.

I'd say let's use your new method of screen tracking in every screen that's within the [submit artwork steps Stack Navigator](https://github.com/artsy/eigen/blob/60471a6e14ffa7a04e0fe9af530191ecba7b2058/src/app/Scenes/SellWithArtsy/ArtworkForm/SubmitArtworkForm.tsx#L168).



Screenshot 2024-08-14 at 14 44 39

src/app/Scenes/SellWithArtsy/Hooks/useSubmitArtworkTracking.ts
olerichter00: That's way better 🤩 

https://github.com/artsy/force/pull/14342

dariakoko:
Good catch @olerichter00 , thanks

src/app/Scenes/SellWithArtsy/ArtworkForm/Components/SubmitArtworkAddDetails.tsx
dariakoko: Ah oki, get it now
Will address 👍

olerichter00 (approved):

PR #10623, opened 2024-08-19
fix: generalize and animate AnimatedMasonryListFooterComponent

gkartalis:
This PR resolves [PHIRE-1124] 

### Description

During the mobile QA we noticed that when doing infinite scroll in any `Tabs.Masonry` surface the spinner was never displayed for the `first` refetch. 

This PR fixes that. Note that in order to do it we had to introduce a boolean state variable on the artworkGrids that use the legacy relay containers.

In surfaces where we use the hooks (i.e. Fairs), it is **not needed** since the `isLoadingNext` is working as expected.

Unified also the page size for all surfaces and the spinner and **bonus** fix was that with the new `AnimatedMasonryListFooterComponent` the content is no longer jumpy after the next page is rendered ⭐ .

Tested on all surfaces on both Android and iOS - Attaching before and after videos for `CollectionArtworks` and `ArtistArtworks`

|Before|After|
|---|---|
|<video src="https://github.com/user-attachments/assets/68a780e3-0285-4cf4-9323-40e9a7b2d006" width="400" />|<video src="https://github.com/user-attachments/assets/c2abc4c7-78b4-48eb-b206-06d6426390ca" width="400" />|
|<video src="https://github.com/user-attachments/assets/ecc09c86-cc6e-45c1-a124-92cbfa9fa44c" width="400" />|<video src="https://github.com/user-attachments/assets/ff9b229d-f8be-4a88-b0be-cbfc34909961" width="400" />|

### PR Checklist

### To the reviewers 👀

- fix spinner not being displayed for first refetch on Tabbed Masonry - gkartalis

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[PHIRE-1124]: https://artsyproduct.atlassian.net/browse/PHIRE-1124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
+++ b/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
@@ -32,6 +32,7 @@ import { useFeatureFlag } from "app/utils/hooks/useFeatureFlag"
 import {
   AnimatedMasonryListFooterComponent,
   ESTIMATED_MASONRY_ITEM_SIZE,
+  MASONRY_LIST_PAGE_SIZE,
   NUM_COLUMNS_MASONRY,
   ON_END_REACHED_THRESHOLD_MASONRY,
 } from "app/utils/masonryHelpers"
@@ -75,6 +76,7 @@ const ArtworksGrid: React.FC<ArtworksGridProps> = ({
     relay,
     aggregations: artist.aggregations?.aggregations,
     componentPath: "ArtistArtworks/ArtistArtworks",
+    pageSize: MASONRY_LIST_PAGE_SIZE,
   })
 
   const setInitialFilterStateAction = ArtworksFiltersStore.useStoreActions(
@@ -139,7 +141,7 @@ const ArtworksGrid: React.FC<ArtworksGridProps> = ({
   const loadMore = useCallback(() => {
     if (relay.hasMore() && !relay.isLoading()) {
       setIsLoading(true)
-      relay.loadMore(10, () => {
+      relay.loadMore(MASONRY_LIST_PAGE_SIZE, () => {
         setIsLoading(false)
       })
     }
--- a/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
+++ b/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
@@ -140,6 +140,9 @@ const ArtworksGrid: React.FC<ArtworksGridProps> = ({
 
   const loadMore = useCallback(() => {
     if (relay.hasMore() && !relay.isLoading()) {
+      // IMPORTANT: this is a workaround to show the spinner concistently between refetches of pages
+      // and it is not needed for grids that use relay hooks since isLoadingNext works better than the
+      // legacy container API. See FairArtworks.tsx for an example of how to use with relay hooks.
       setIsLoading(true)
       relay.loadMore(MASONRY_LIST_PAGE_SIZE, () => {
         setIsLoading(false)
--- a/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
+++ b/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
@@ -30,12 +30,12 @@ import { Props as InfiniteScrollGridProps } from "app/Components/ArtworkGrids/In
 import { extractNodes } from "app/utils/extractNodes"
 import { useFeatureFlag } from "app/utils/hooks/useFeatureFlag"
 import {
-  AnimatedMasonryListFooterComponent,
   ESTIMATED_MASONRY_ITEM_SIZE,
   MASONRY_LIST_PAGE_SIZE,
   NUM_COLUMNS_MASONRY,
   ON_END_REACHED_THRESHOLD_MASONRY,
 } from "app/utils/masonryHelpers"
+import { AnimatedMasonryListFooter } from "app/utils/masonryHelpers/AnimatedMasonryListFooter"
 import { Schema } from "app/utils/track"
 import React, { useCallback, useEffect, useMemo, useRef, useState } from "react"
 import { RelayPaginationProp, createPaginationContainer, graphql } from "react-relay"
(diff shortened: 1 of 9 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (fix spinner not being displayed for first refetch on Tabbed Masonry - gkartalis)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 11e06edb2c007e75234d3c8166a3d851dfb94e9a
</p>

src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
```diff
@@ -31,7 +30,9 @@ import { Props as InfiniteScrollGridProps } from "app/Components/ArtworkGrids/In
 import { extractNodes } from "app/utils/extractNodes"
 import { useFeatureFlag } from "app/utils/hooks/useFeatureFlag"
 import {
+  AnimatedMasonryListFooterComponent,
```
araujobarret, line 33: looking at how big this file is becoming we could consider breaking it down, this new component seems to be a good candidate to be in another area

src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
```diff
@@ -133,11 +135,15 @@ const ArtworksGrid: React.FC<ArtworksGridProps> = ({
     // TODO: Get the new count of the artist alerts
   }
 
-  const shouldDisplaySpinner = !!artworks.length && !!relay.isLoading() && !!relay.hasMore()
+  const shouldDisplaySpinner =
+    !!isLoading && !!artworks.length && !!relay.isLoading() && !!relay.hasMore()
```
araujobarret, line 139: `artworks.length` is not needed here, if the result is `0` the `hasMore()` check will also return false, isn't it?

src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
```diff
@@ -133,11 +135,15 @@ const ArtworksGrid: React.FC<ArtworksGridProps> = ({
...
 
   const loadMore = useCallback(() => {
     if (relay.hasMore() && !relay.isLoading()) {
-      relay.loadMore(10)
+      setIsLoading(true)
```
araujobarret, line 143: maybe tag this with some comment because it shouldn't be the standard way of checking if is loading on paginated data

araujobarret (commented):
If it's for a hotfix happy to approve and move on, but wonder if we could spend a bit more time on it if we don't need to ship in this release 😅 

My main concern is that we need this extra `isLoading` state when other apps where we use pagination hooks/containers we don't

src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
gkartalis: I would say that in this PR I made it a tiny bit smaller - I wouldn't touch it more since it belongs to another team and it would be out of context for this PR

src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
gkartalis: comment sounds good

src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
gkartalis: if `artworks.length` is 0 it will exit early - same if the `isLoading` is false won't check the rest. Do you suggest a different approach?

gkartalis:
> If it's for a hotfix happy to approve and move on, but wonder if we could spend a bit more time on it if we don't need to ship in this release 😅

It is not a hotfix. We can def spend more time on it but what do you have in mind exactly?

> My main concern is that we need this extra isLoading state when other apps where we use pagination hooks/containers we don't

This behavior has to do with how different relay legacy `isLoading()` and `hasMore()` versus the new from hooks `isLoadingNext` and `hasNext`. 

I am happy to add comments but what else do you suggest as an improvement or change?

src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
araujobarret: Sorry, I meant the `app/utils/masonryHelpers`, I think we could slice some stuff from there, I'd def leave ArtistsArtworks alone 😂

gkartalis: What do you mean exactly by `another area`? Like another file? Would this be for the components? ConstantS?

src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
araujobarret: Good point about exiting earlier, my comment was oriented on reducing the number of checks to make it easier to read(3 checks instead of 4), perf-wise don't think it's a huge deal, but DX it could improve, will leave it to you 👍

araujobarret:
@gkartalis I see I see, checked the FairArtworks which has really hooks and we don't need it, so it's Relay's old remains causing the issue rather than our component setup asking for extra state control.

Fair enough from my side, just in the comment there add the reference to try to remove it as soon as these components are migrated to relay hooks, what do you think?

araujobarret (dismissed):

gkartalis:
@araujobarret Sounds good, will add the comment on all cases 🌟

src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
araujobarret: Everything there is under `index.tsx`, I think we could leave the **index** as it is for now and just add the new element under:
```
|--app/utils/masonryHelpers
|---------------------------- index.tsx
|---------------------------- AnimatedMasonryListFooterComponent.tsx
```

gkartalis: done!

gkartalis:
Done! Added comments and moved the component to its own file under masonryHelpers - I did remove the `Component` part from the name as well since it felt too much

araujobarret:
It looks solid, thanks for the changes!

PR #10633, opened 2024-08-21
feat(DIA-707): wire up collected artists prompt to inquiry

araujobarret:
This PR resolves [DIA-707]

### Description

Continuing of the prompt work, after inquiring on an Artwork if the collector hasn't been prompted in the last 30 days and has no items in his collection, we show the prompt.

Main changes:
- refactored the InquiryModal to be simpler by using the hook to control its behavior isolatedly(easier to test and read)
- changed the order of the prompts(agreed with the product) now we show the edit profile prompt before the collection(it's easier for the collector to fill in their information rather than initially having a collection)

https://github.com/user-attachments/assets/ae887f90-e87d-4f22-b53d-5fbbd6bd2df1

### PR Checklist

### To the reviewers 👀

- feat: wire collected artists prompt to inquiry(ff hidden) - araujobarret

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[DIA-707]: https://artsyproduct.atlassian.net/browse/DIA-707?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/Artwork/Artwork.tsx
+++ b/src/app/Scenes/Artwork/Artwork.tsx
@@ -772,7 +772,9 @@ export const ArtworkContainer = createRefetchContainer(
     me: graphql`
       fragment Artwork_me on Me @argumentDefinitions(artworkID: { type: "String!" }) {
         ...ArtworkCommercialButtons_me
-        ...InquiryModal_me
+        ...useSendInquiry_me
+        ...MyProfileEditModal_me
+        ...BidButton_me
         partnerOffersConnection(artworkID: $artworkID, first: 1) {
           edges {
             node {
--- a/src/app/Scenes/Artwork/Components/ArtworkCommercialButtons.tsx
+++ b/src/app/Scenes/Artwork/Components/ArtworkCommercialButtons.tsx
@@ -1,7 +1,9 @@
 import { Spacer, Flex, Join } from "@artsy/palette-mobile"
 import { ArtworkCommercialButtons_artwork$key } from "__generated__/ArtworkCommercialButtons_artwork.graphql"
-import { ArtworkCommercialButtons_me$key } from "__generated__/ArtworkCommercialButtons_me.graphql"
 import { ArtworkCommercialButtons_partnerOffer$key } from "__generated__/ArtworkCommercialButtons_partnerOffer.graphql"
+import { BidButton_me$key } from "__generated__/BidButton_me.graphql"
+import { MyProfileEditModal_me$key } from "__generated__/MyProfileEditModal_me.graphql"
+import { useSendInquiry_me$key } from "__generated__/useSendInquiry_me.graphql"
 import { AuctionTimerState } from "app/Components/Bidding/Components/Timer"
 import { ArtworkStore } from "app/Scenes/Artwork/ArtworkStore"
 import { BuyNowButton } from "app/Scenes/Artwork/Components/CommercialButtons/BuyNowButton"
@@ -15,7 +17,7 @@ import { BidButtonFragmentContainer } from "./CommercialButtons/BidButton"
 
 interface ArtworkCommercialButtonsProps {
   artwork: ArtworkCommercialButtons_artwork$key
-  me: ArtworkCommercialButtons_me$key
+  me: MyProfileEditModal_me$key & useSendInquiry_me$key & BidButton_me$key
   partnerOffer: ArtworkCommercialButtons_partnerOffer$key
 }
 
@@ -206,7 +208,8 @@ const artworkFragment = graphql`
 const meFragment = graphql`
   fragment ArtworkCommercialButtons_me on Me {
     ...BidButton_me
-    ...InquiryModal_me
+    ...useSendInquiry_me
+    ...MyProfileEditModal_me
   }
 `
 
--- a/src/app/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tests.tsx
+++ b/src/app/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tests.tsx
@@ -1,4 +1,10 @@
-import { fireEvent, screen, waitFor } from "@testing-library/react-native"
+import {
+  act,
+  fireEvent,
+  screen,
+  waitFor,
+  waitForElementToBeRemoved,
+} from "@testing-library/react-native"
 import { InquiryModalTestsQuery } from "__generated__/InquiryModalTestsQuery.graphql"
 import { InquiryModal } from "app/Scenes/Artwork/Components/CommercialButtons/InquiryModal"
 import { AUTOMATED_MESSAGES } from "app/Scenes/Artwork/Components/CommercialButtons/constants"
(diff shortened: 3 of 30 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (feat: wire collected artists prompt to inquiry(ff hidden) - araujobarret)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 540936984d08597387291f7ec4277959fa970e1b
</p>

src/app/Scenes/Artwork/Artwork.tsx
```diff
@@ -772,7 +772,9 @@ export const ArtworkContainer = createRefetchContainer(
     me: graphql`
       fragment Artwork_me on Me @argumentDefinitions(artworkID: { type: "String!" }) {
         ...ArtworkCommercialButtons_me
-        ...InquiryModal_me
+        ...useSendInquiry_me
```
araujobarret, line 775: removing the `me` fragment from InquiryModal required these changes everywhere, InquiryModal doesn't use `me`, so no fragment should be defined there.

src/app/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tests.tsx
```diff
@@ -114,253 +122,56 @@ describe("inquiry modal", () => {
     })
   })
 
-  describe("sending the inquiry", () => {
```
araujobarret, line 117: since a lot of logic was moved from `InquiryModal` to `useSendInquiry` the same applies to the tests. They were moved/refactored too.

src/app/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
```diff
@@ -28,141 +25,62 @@ import { ShippingModal } from "./ShippingModal"
...
 
-  useEffect(() => {
+  const resetAndExit = () => {
     setMessage(randomAutomatedMessage())
-  }, [setMessage])
+    dispatch({ type: "resetForm", payload: null })
+    dispatch({ type: "setInquiryModalVisible", payload: false })
+  }
 
-  useEffect(() => {
```
araujobarret, line 54: none of these useEffects were necessary, they were not ported to the useSendInquiry

src/app/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
```diff
@@ -28,141 +25,62 @@ import { ShippingModal } from "./ShippingModal"
...
   const scrollViewRef = useRef<ScrollView>(null)
-  const [commit] = useSubmitInquiryRequest()
   const tracking = useTracking()
 
-  const artworkData = useFragment(artworkFragment, artwork)
-  const meData = useFragment(meFragment, me)
+  const artwork = useFragment(artworkFragment, _artwork)
 
-  const [message, setMessage] = useState<string>("")
+  const [message, setMessage] = useState<string>(() => randomAutomatedMessage())
```
araujobarret, line 39: that's how you lazily start state, not with an useEffect with empty deps(setMessage as dep is not necessary since it's a state setter)

src/app/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
```diff
@@ -252,7 +170,22 @@ export const InquiryModal: React.FC<InquiryModalProps> = ({ artwork, me }) => {
         />
       </FancyModal>
       <InquirySuccessNotification />
-      {!!profilePromptIsEnabled && <CompleteProfilePrompt artwork={artworkData} me={meData} />}
+
+      {!!profilePromptIsEnabled && (
+        <CompleteProfilePrompt
+          artwork={artwork}
+          me={me}
+          visible={state.profilePromptVisible}
```
araujobarret, line 178: @iskounen I had to lift these props up, the elements using this prompt need to have the control over the visibility and when it was dismissed

src/app/Scenes/MyProfile/MyProfileEditModal.tsx
```diff
@@ -18,14 +18,14 @@ import { useTracking } from "react-tracking"
...
-  onClose: () => void
+  onDismiss: () => void
   visible: boolean
 }
 
 export const MyProfileEditModal: React.FC<MyProfileEditModalProps> = ({
   me,
   message,
-  onClose,
+  onDismiss,
```
araujobarret, line 28: wording to be consistent with BottomSheet props

src/app/utils/ArtworkInquiry/ArtworkInquiryStore.tsx
```diff
@@ -62,11 +61,6 @@ export const artworkInquiryStateReducer = (
         ...inquiryState,
         profilePromptVisible: action.payload,
       }
-    case "setMyCollectionArtistsPromptVisible":
```
araujobarret, line 65: oversight from past PR, we already had one state in place for the collection prompt

src/app/utils/collectorPromptHelpers.tests.ts
```diff
@@ -0,0 +1,134 @@
+import {
```
araujobarret, line 1: made sense adding coverage for this file, the tests we had were not covering it 100%

src/app/Scenes/MyCollection/Components/MyCollectionArtistsPrompt/__tests__/MyCollectionArtistsPromptFooter.tests.tsx
```diff
@@ -59,7 +63,32 @@ describe("MyCollectionArtistsPromptFooter", () => {
     expect(onPress).not.toHaveBeenCalled()
   })
 
-  it("calls close when text is pressed", () => {
+  it("does call close when text is pressed given an error", () => {
```
iskounen, line 66: It looks like this test asserts that it does **not** call close when there is an error, and the following test asserts that it does call close when there is **no** error. If that's correct, then the test description should be updated.

src/app/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
iskounen: I am curious about this change because if the `InquiryModal` and the prompts are all aware of the same `ArtworkInquiryContext`, then they can communicate visibility and dismissals using its event dispatches. I found that to be a better way of communicating than sharing props.

Is there a scenario where using dispatch events didn't work?

iskounen (commented):
I think the `useSendInquiry` hook introduces a nice abstraction to the already complicated `InquiryModal` component. Thanks for tying this all together!

src/app/Scenes/MyCollection/Components/MyCollectionArtistsPrompt/__tests__/MyCollectionArtistsPromptFooter.tests.tsx
araujobarret: oh yeah, nice catch! 🎣

araujobarret:
>I am curious about this change because if the InquiryModal and the prompts are all aware of the same ?>ArtworkInquiryContext, then they can communicate visibility and dismissals using its event dispatches. I found that to be a better way of communicating than sharing props.
>
>Is there a scenario where using dispatch events didn't work?

It wasn't fully necessary here but will be in the next ticket of wiring it up to the activity card, I included this small update because I had to do the same for the collection prompt in order to leave it abstract to be set visible from other surfaces.

It enables the prompt to be used on other surfaces, before the `visible` was linked directly to a state variable of the the inquiry context, if we use the same element in the Notifications for example, won't work since Notifications has no InquiryContext wrapping it up.

src/app/Scenes/Artwork/Components/ArtworkCommercialButtons.tsx
```diff
@@ -15,7 +17,7 @@ import { BidButtonFragmentContainer } from "./CommercialButtons/BidButton"
 
 interface ArtworkCommercialButtonsProps {
   artwork: ArtworkCommercialButtons_artwork$key
-  me: ArtworkCommercialButtons_me$key
+  me: MyProfileEditModal_me$key & useSendInquiry_me$key & BidButton_me$key
```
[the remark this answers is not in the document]
araujobarret: It's a bit of annoying when the level of nesting is too much between the elements and then we have 3 types in an intersection drilled in multiple elements, but Artwork surface is a huge one maybe that's why is kinda hard to not be soo verbose here

araujobarret:
Merging it, can address follow-ups later 👍

PR #10672, opened 2024-08-28
feat: add activity card for collector update notification

araujobarret:
This PR resolves [DIA-696] [DIA-703] [DIA-708] [DIA-710]

### Description

This PR implements the Activity Items for the collector profile update notifications.

https://github.com/user-attachments/assets/9ca13dfc-71c7-4251-a851-32bfdbfe4995

### PR Checklist

### To the reviewers 👀

- feat: add activity items given collector update notifications - araujobarret

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[DIA-696]: https://artsyproduct.atlassian.net/browse/DIA-696?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DIA-708]: https://artsyproduct.atlassian.net/browse/DIA-708?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DIA-703]: https://artsyproduct.atlassian.net/browse/DIA-703?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DIA-710]: https://artsyproduct.atlassian.net/browse/DIA-710?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/Activity/components/CollectorUpdateNotification.tsx
+++ b/src/app/Scenes/Activity/components/CollectorUpdateNotification.tsx
@@ -18,7 +18,7 @@ export const CollectorUpdateNotification: FC<CollectorUpdateNotificationProps> =
   notification: _notification,
   item: _item,
 }) => {
-  const [visible, setVisible] = useState(false)
+  const [promptVisible, setPromptVisible] = useState(false)
   const notification = useFragment(NOTIFICATION_FRAGMENT, _notification)
   const item = useFragment(ITEM_FRAGMENT, _item)
 
@@ -27,20 +27,23 @@ export const CollectorUpdateNotification: FC<CollectorUpdateNotificationProps> =
   }
 
   const handleOnDismiss = () => {
-    setVisible(false)
+    setPromptVisible(false)
   }
 
   const itemInfo = getItemInfo(item)
 
   return (
     <>
-      <Touchable onPress={() => setVisible(true)}>
+      <Touchable onPress={() => setPromptVisible(true)}>
         <Flex flex={1} py={2} pr={2}>
-          <Text fontWeight={500}>{itemInfo.title}</Text>
+          <Text variant="sm-display" fontWeight={500}>
+            {itemInfo.title}
+          </Text>
           <Text variant="xs">{itemInfo.body}</Text>
 
-          <Text variant="xs">
-            Artsy Message •<Text variant="xs">{` ${notification.publishedAt}`}</Text>
+          <Text variant="xs" fontWeight={500}>
+            Artsy Message •
+            <Text variant="xs" fontWeight="normal">{` ${notification.publishedAt}`}</Text>
           </Text>
         </Flex>
       </Touchable>
@@ -48,15 +51,11 @@ export const CollectorUpdateNotification: FC<CollectorUpdateNotificationProps> =
       {itemInfo.prompt === "AddArtistsToCollection" ? (
         <MyCollectionBottomSheetModalArtistsPrompt
           title="Tell us about the artists in your collection."
-          visible={visible}
+          visible={promptVisible}
           onDismiss={handleOnDismiss}
         />
       ) : (
-        <CollectorProfilePrompt
-          me={item.me}
-          visible={visible}
-          onDismiss={() => setVisible(false)}
-        />
+        <CollectorProfilePrompt me={item.me} visible={promptVisible} onDismiss={handleOnDismiss} />
       )}
     </>
   )
--- a/src/app/Scenes/Activity/components/CollectorUpdateNotification.tsx
+++ b/src/app/Scenes/Activity/components/CollectorUpdateNotification.tsx
@@ -1,0 +1,108 @@
+import { Flex, Text, Touchable } from "@artsy/palette-mobile"
+import {
+  CollectorUpdateNotification_item$data,
+  CollectorUpdateNotification_item$key,
+} from "__generated__/CollectorUpdateNotification_item.graphql"
+import { CollectorUpdateNotification_notification$key } from "__generated__/CollectorUpdateNotification_notification.graphql"
+import { MyCollectionBottomSheetModalArtistsPrompt } from "app/Scenes/MyCollection/Components/MyCollectionBottomSheetModals/MyCollectionBottomSheetModalArtistsPrompt"
+import { FC, useState } from "react"
+import { graphql, useFragment } from "react-relay"
+import { CollectorProfilePrompt } from "./CollectorProfilePrompt"
+
+interface CollectorUpdateNotificationProps {
+  notification: CollectorUpdateNotification_notification$key
+  item: CollectorUpdateNotification_item$key
+}
+
+export const CollectorUpdateNotification: FC<CollectorUpdateNotificationProps> = ({
+  notification: _notification,
+  item: _item,
+}) => {
+  const [visible, setVisible] = useState(false)
+  const notification = useFragment(NOTIFICATION_FRAGMENT, _notification)
+  const item = useFragment(ITEM_FRAGMENT, _item)
+
+  if (!notification || !item) {
+    return null
+  }
+
+  const handleOnDismiss = () => {
+    setVisible(false)
+  }
+
+  const itemInfo = getItemInfo(item)
+
+  return (
+    <>
+      <Touchable onPress={() => setVisible(true)}>
+        <Flex flex={1} py={2} pr={2}>
+          <Text fontWeight={500}>{itemInfo.title}</Text>
+          <Text variant="xs">{itemInfo.body}</Text>
+
+          <Text variant="xs">
+            Artsy Message •<Text variant="xs">{` ${notification.publishedAt}`}</Text>
+          </Text>
+        </Flex>
+      </Touchable>
+
+      {itemInfo.prompt === "AddArtistsToCollection" ? (
+        <MyCollectionBottomSheetModalArtistsPrompt
+          title="Tell us about the artists in your collection."
+          visible={visible}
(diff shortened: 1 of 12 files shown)


src/app/Scenes/Activity/components/ExpiresInTimer.tsx
```diff
@@ -15,7 +16,9 @@ const WatchIcon: FC<{ fill?: string }> = ({ fill = "red100" }) => {
 }
 
 export const ExpiresInTimer: FC<ExpiresInTimerProps> = ({ item }) => {
+  // @ts-ignore: fix ExpiresInTimer fragment data
```
araujobarret, line 19: I won't fix these too, they're legit ts complaints because the component is using relay data/types in the wrong way.
Whoever comes back here from PartnerOffer work should fix it

src/app/Scenes/Activity/components/CollectorUpdateNotification.tsx
```diff
@@ -0,0 +1,108 @@
...
+interface CollectorUpdateNotificationProps {
+  notification: CollectorUpdateNotification_notification$key
+  item: CollectorUpdateNotification_item$key
+}
+
+export const CollectorUpdateNotification: FC<CollectorUpdateNotificationProps> = ({
+  notification: _notification,
+  item: _item,
+}) => {
+  const [visible, setVisible] = useState(false)
```
nickskalkin, line 21: **nitpick**: I would specify what exactly is visible (bottomSheetVisible), because it's not clear without looking at the bottom of the file

src/app/Scenes/Activity/components/ExpiresInTimer.tsx
nickskalkin: Interesting, quite sure I wrote this but linter was not complaining...

src/app/Scenes/Activity/components/CollectorUpdateNotification.tsx
```diff
@@ -0,0 +1,108 @@
...
+    setVisible(false)
+  }
+
+  const itemInfo = getItemInfo(item)
+
+  return (
+    <>
+      <Touchable onPress={() => setVisible(true)}>
+        <Flex flex={1} py={2} pr={2}>
+          <Text fontWeight={500}>{itemInfo.title}</Text>
```
nickskalkin, line 39: **suggestion**: I think it is sm-display and bold for other types of notifications

nickskalkin (approved):

egdbear (dismissed):

src/app/Scenes/Activity/components/CollectorUpdateNotification.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  if (!notification || !item) {
+    return null
+  }
+
+  const handleOnDismiss = () => {
+    setVisible(false)
+  }
+
+  const itemInfo = getItemInfo(item)
```
iskounen, line 33: **suggestion (non-blocking):** I think having copy in JSX code makes it easier to read React component because using `{hasEmptyCollection ? "foo" : "bar"}` makes the variations (and the logic that determines them) easier to see.

src/app/Scenes/Activity/components/CollectorUpdateNotification.tsx
```diff
@@ -0,0 +1,108 @@
...
+            Artsy Message •<Text variant="xs">{` ${notification.publishedAt}`}</Text>
+          </Text>
+        </Flex>
+      </Touchable>
+
+      {itemInfo.prompt === "AddArtistsToCollection" ? (
+        <MyCollectionBottomSheetModalArtistsPrompt
+          title="Tell us about the artists in your collection."
+          visible={visible}
+          onDismiss={handleOnDismiss}
```
iskounen, line 55: **nitpick:** The `onDismiss` handler for both prompts do the same thing, but this one uses a named function, and the profile prompt uses an anonymous one. I think this one can be an anonymous function too.

src/app/Scenes/Activity/components/CollectorUpdateNotification.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { graphql, useFragment } from "react-relay"
+import { CollectorProfilePrompt } from "./CollectorProfilePrompt"
+
+interface CollectorUpdateNotificationProps {
+  notification: CollectorUpdateNotification_notification$key
+  item: CollectorUpdateNotification_item$key
+}
+
+export const CollectorUpdateNotification: FC<CollectorUpdateNotificationProps> = ({
+  notification: _notification,
```
iskounen, line 18: **nitpick (non-blocking):** I see that you're using the underscore to be able to reuse the variable name when we've fetched the data, but we usually use underscores for unused parameters in our codebase, so this was confusing at first.

src/app/Scenes/Activity/components/ExpiresInTimer.tsx
araujobarret: it's because we added another type spread in `item` in the `ActivityItem` element, so the autogenerated type there can be a different type now. 
I tried something like
```jsx
if (!item.__typename !== "PartnerOfferCreatedNotificationItem") {
  return null
}
```
but didn't fix the issue, and other minor attempts too, so I pushed it to the next fortunate engineer touching it xD

araujobarret:
thanks for the review both!
I addressed the visual changes, it made sense to keep them consistent, our Figma mockups might be wrong on our end. 🙇

src/app/Scenes/MyCollection/Components/MyCollectionArtistsPrompt/MyCollectionArtistsPromptFooter.tsx
```diff
@@ -12,40 +10,20 @@ interface MyCollectionArtistsPromptFooterProps {
...
-          res.updateCollectorProfile?.collectorProfileOrError?.mutationError?.message ?? e
-        if (error) {
-          handleError(error)
-          return
-        }
-        close()
-      },
-      onError: handleError,
-    })
+    console.log("handleOnClose")
```
iskounen, line 23: **nitpick:** 🙈

src/app/Scenes/Activity/components/CollectorUpdateNotification.tsx
araujobarret: bottomSheet is kinda abstract to who's using the element, so I tried `promptVisible` hope that's a bit better 😂

src/app/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
```diff
@@ -78,6 +80,25 @@ export const InquiryModal: React.FC<InquiryModalProps> = ({ artwork: _artwork, m
...
+    commit({
+      variables: { input: { promptedForUpdate: true } },
+      onCompleted: (res, e) => {
+        const error =
+          res.updateCollectorProfile?.collectorProfileOrError?.mutationError?.message ?? e
+        if (error) {
+          console.error(
+            "[InquiryModal MyCollectionArtistsPromptFooter updateCollectorProfile] error:",
+            error
+          )
```
iskounen, line 92: **suggestion:** I liked how this was DRY'ed up with a `handleError` method in its old location.

src/app/Scenes/Home/Components/ActivityRailItem.tsx
```diff
@@ -44,6 +44,10 @@ export const ActivityRailItem: React.FC<ActivityRailItemProps> = (props) => {
   const isPartnerOffer = item.notificationType === "PARTNER_OFFER_CREATED"
   const isEditorial = item.notificationType === "ARTICLE_FEATURED_ARTIST"
 
+  if (item.item?.__typename === "CollectorProfileUpdatePromptNotificationItem") {
+    return null
+  }
```
iskounen, line 49: **question:** Should we exit a bit earlier, before `handlePress` is declared, to get a minor performance improvement when there is a `CollectorProfileUpdatePromptNotificationItem` in the feed (which it can be for 4 weeks)?

src/app/Scenes/Activity/components/CollectorUpdateNotification.tsx
araujobarret: personal-taste from me, I prefer less anonymous functions on the callbacks on the components, usually, they increase verbosity when reading the component tree of a specific element, but changing it in order to keep consistent, anonymous won 👍

src/app/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx
araujobarret: at this stage it'd be only to handle a console log statement, before had a popover show function being called 🙈

src/app/Scenes/Home/Components/ActivityRailItem.tsx
araujobarret: It won't be a problem because the commit is going to be sent to the event loop as an async call, we don't block but I'll change the order anyways 👍

PR #10693, opened 2024-08-30
feat: add rail viewed tracking

MounirDhahri:
This PR resolves [] 

### Description

This PR adds rails impressions tracking to the home view.

Since RN doesn't support changing `onViewabilityConfig` props on the fly, it wasn't possible (for now) to get right scroll position since the method will only have access to the old values. I removed that for now and will discuss with Dayane if that's a prop that is critical for us. 

_I will try to add that next week, but I think it should be shippable as it is now_

https://github.com/user-attachments/assets/89525d36-d95d-4187-aebe-319df2a8a1b1

### PR Checklist

### To the reviewers 👀

- add rail viewed tracking - mounir

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Scenes/HomeView/HomeView.tsx
+++ b/src/app/Scenes/HomeView/HomeView.tsx
@@ -1,10 +1,20 @@
 import { Flex, Screen, Spacer, Spinner, Text } from "@artsy/palette-mobile"
 import { HomeViewQuery } from "__generated__/HomeViewQuery.graphql"
-import { HomeViewSectionsConnection_viewer$key } from "__generated__/HomeViewSectionsConnection_viewer.graphql"
+import {
+  HomeViewSectionsConnection_viewer$data,
+  HomeViewSectionsConnection_viewer$key,
+} from "__generated__/HomeViewSectionsConnection_viewer.graphql"
 import { Section } from "app/Scenes/HomeView/Sections/Section"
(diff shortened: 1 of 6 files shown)


nickskalkin (approved):
Everything looks good!

> it wasn't possible (for now) to get right scroll position

I think that index of a section is more important than scroll position. Since sections can be shown in different order for different users - does it makes sense to add this info to analytic event payload?

src/app/Scenes/HomeView/HomeView.tsx
```diff
@@ -73,36 +90,60 @@ const sectionsFragment = graphql`
...
+                totalCount
+              }
               ...ArticlesRailHomeViewSection_section
               ...ArticlesCardsHomeViewSection_section
             }
             ... on ArtworksRailHomeViewSection {
               internalID
+              _artworksConnection: artworksConnection {
+                totalCount
+              }
```
olerichter00, line 110: Wouldn't this lead to a second call from Metaphysics to Gravity because we're requesting the artworksConnection without any arguments here (`_artworksConnection`) and then with different arguments in the [`ArtworksRailHomeViewSection_section` fragment](https://github.com/artsy/eigen/pull/10693/files#diff-02d75af5582539c09ef32b302e53e0348e5e3de01cb069392c40f2ffec91351fR52)?

PR #10717, opened 2024-09-05
[WIP] feat(EMI-2042): Create artworks grid and rail context

MrSltun:
This PR resolves [EMI-2042] 

> [!NOTE]
> This PR is still WIP, but would love some early feedback on it! 

### Description
This PR adds a context for the Artwork rails or grids for the `ArtworkGridItem` and `ArtworkRailCard` to know where they are

This is a preliminary work for the "Curator's Pick" and "Increased Interest" signals that we are going to add to `ArtworkGridItem` and `ArtworkRailCard` to show them only in specific places, not everywhere in the app

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[EMI-2042]: https://artsyproduct.atlassian.net/browse/EMI-2042?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/Activity/components/NotificationArtworkList.tsx
+++ b/src/app/Scenes/Activity/components/NotificationArtworkList.tsx
@@ -4,7 +4,7 @@ import { NotificationArtworkList_artworksConnection$key } from "__generated__/No
 import { PriceOfferMessage } from "app/Components/ArtworkGrids/ArtworkGridItem"
 import { MasonryInfiniteScrollArtworkGrid } from "app/Components/ArtworkGrids/MasonryInfiniteScrollArtworkGrid"
 import { CommercialButtonsQueryRenderer } from "app/Scenes/Activity/components/NotificationCommercialButtons"
-import { ArtworksGridAOrRailProvider } from "app/utils/ArtworksContext/ArtworksGridAndRailContext"
+import { ArtworksGridRailProvider } from "app/utils/ArtworksContext/ArtworksGridRailContext"
 import { extractNodes } from "app/utils/extractNodes"
 import { FC } from "react"
 import { ImageBackground } from "react-native"
@@ -38,7 +38,7 @@ export const NotificationArtworkList: FC<NotificationArtworkListProps> = (props)
 
   return (
     <Flex minHeight={400}>
-      <ArtworksGridAOrRailProvider currentGridOrRail="ACTIVITY_PANEL_ARTWORKS_GRID">
+      <ArtworksGridRailProvider currentGridRail="ACTIVITY_PANEL_ARTWORKS_GRID">
         <MasonryInfiniteScrollArtworkGrid
           animated
           artworks={artworks}
@@ -52,7 +52,7 @@ export const NotificationArtworkList: FC<NotificationArtworkListProps> = (props)
           priceOfferMessage={priceOfferMessage}
           style={{ paddingBottom: 120 }}
         />
-      </ArtworksGridAOrRailProvider>
+      </ArtworksGridRailProvider>
       {!!showArtworkCommercialButtons && (
         <CommercialButtonsQueryRenderer
           artworkID={artworks[0].internalID}
--- a/src/app/Scenes/Activity/components/NotificationArtworkList.tsx
+++ b/src/app/Scenes/Activity/components/NotificationArtworkList.tsx
@@ -4,7 +4,7 @@ import { NotificationArtworkList_artworksConnection$key } from "__generated__/No
 import { PriceOfferMessage } from "app/Components/ArtworkGrids/ArtworkGridItem"
 import { MasonryInfiniteScrollArtworkGrid } from "app/Components/ArtworkGrids/MasonryInfiniteScrollArtworkGrid"
 import { CommercialButtonsQueryRenderer } from "app/Scenes/Activity/components/NotificationCommercialButtons"
-import { ArtworksGridRailProvider } from "app/utils/ArtworksContext/ArtworksGridRailContext"
+import { ArtworksGridRailContextStore } from "app/utils/ArtworksContext/ArtworksGridRailContext"
 import { extractNodes } from "app/utils/extractNodes"
 import { FC } from "react"
 import { ImageBackground } from "react-native"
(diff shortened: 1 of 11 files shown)


src/app/Scenes/Home/Home.tsx
```diff
@@ -104,6 +105,7 @@ export interface HomeModule extends ArtworkActionTrackingProps {
   subtitle?: string
   title: string
   type: string
+  currentGridOrRail?: GridAndRailType
```
MrSltun, line 108: Would love to hear your thoughts about adding the rail key to the home module type to use it when adding a provider to the rails

src/app/utils/ArtworksContext/ArtworksGridAndRailContext.tsx
MrSltun, line 1: would also love to hear how you would do this differently

MrSltun (commented):
Would love to have some early feedback on this implementation

src/app/Scenes/Activity/components/NotificationArtworkList.tsx
```diff
@@ -4,6 +4,7 @@ import { NotificationArtworkList_artworksConnection$key } from "__generated__/No
 import { PriceOfferMessage } from "app/Components/ArtworkGrids/ArtworkGridItem"
 import { MasonryInfiniteScrollArtworkGrid } from "app/Components/ArtworkGrids/MasonryInfiniteScrollArtworkGrid"
 import { CommercialButtonsQueryRenderer } from "app/Scenes/Activity/components/NotificationCommercialButtons"
+import { ArtworksGridAOrRailProvider } from "app/utils/ArtworksContext/ArtworksGridAndRailContext"
```
damassi, line 7: #superminor - The "ThingOrFoo" pattern is something that one typically doesn't see when naming variables. We should pick one general thing that encompasses both and go with it (ArtworkCollectionContext) or remove the "or" (ArtworkGridRailProvider)

src/app/utils/ArtworksContext/ArtworksGridAndRailContext.tsx
```diff
@@ -0,0 +1,92 @@
...
+interface ArtworksGridOrRailContextState {
+  currentGridOrRail: GridAndRailType
+  shouldShow: ShouldShowType
+}
+
+interface ArtworksGridAndRailProviderProps {
+  currentGridOrRail: GridAndRailType
+}
+
+const ArtworksGridOrRailContext = createContext<ArtworksGridOrRailContextState>({
```
damassi, line 15: Can we add an [eslint rule that warns against the importing](https://github.com/artsy/volt/blob/main/.eslintrc.js#L126-L130) of vanilla createContext so that the team stops hand rolling state management from scratch, which always comes with inconsistent implementations? 

We should be using easy peasy's create-context-store where we can! (for all previous reasons mentioned historically through time and for all of the additional [ones here](https://github.com/artsy/volt-v2/blob/main/docs/best_practices.md#state). Its more performant; its consistent; there's baked in typescript support -- and so on.

src/app/Scenes/Home/Home.tsx
dariakoko: The solution should also be somewhat compatible with the new Home View implementation, right?

src/app/utils/ArtworksContext/ArtworksGridAndRailContext.tsx
damassi: Ahh sorry, wrong docs location: https://github.com/artsy/volt/blob/main/doc/javascript/managing-state.md

src/app/Scenes/Home/Home.tsx
MrSltun: Well, I believe that the new Home View implementation would be done in MP

src/app/utils/ArtworksContext/ArtworksGridAndRailContext.tsx
araujobarret: ++ for easy-peasy, since it's a surface used in many places we need to have performance in mind

src/app/utils/ArtworksContext/ArtworksGridAndRailContext.tsx
```diff
@@ -0,0 +1,92 @@
...
+        return { LIMITED_TIME_OFFER: false }
+
+      default:
+        return {
+          LIMITED_TIME_OFFER: false,
+        }
+    }
+  }
+
+  return (
```
araujobarret, line 57: so, from what I see here we are adding this context too all Rails in the app to show or not show the elements, is that right?

MrSltun: Yes, and the elements are the signals on the artwork grid items/rail cards, specifically "Curator's Pick" and "Increased Interest"

src/app/Scenes/Activity/components/NotificationArtworkList.tsx
MrSltun: We can rename it, but it will be used across the app

src/app/Scenes/Activity/components/NotificationArtworkList.tsx
```diff
@@ -38,7 +38,9 @@ export const NotificationArtworkList: FC<NotificationArtworkListProps> = (props)
 
   return (
     <Flex minHeight={400}>
-      <ArtworksGridRailProvider currentGridRail="ACTIVITY_PANEL_ARTWORKS_GRID">
+      <ArtworksGridRailContextStore.Provider
```
damassi, line 41: Rather than importing the bare context / provider, we can export a component `ArtworkGridRailProvider` and provide our own API:
- https://github.com/artsy/volt/blob/main/app/javascript/apps/conversations/ConversationsContext.tsx#L17-L27
- https://github.com/artsy/volt/blob/main/app/javascript/apps/demand/DemandContext.tsx#L105-L114

This more closely matches how we use providers around and in force and so on too. Makes this layer "middleware", not necessarily bound to a specific library or pattern, so very slightly more flexible at the cost of an additional export.

damassi:
One last minor request @MrSltun - can we rename `ArtworksGridRailContextStoreProvider` to something a bit more generic? That way it doesn't leak implementation details ("ContextStore") which is tied to easy-peasy. Something like `ArtworksGridRailProvider` keeps things slender and generic.

MrSltun:
A little update
Thanks to the feedback I got during Mobile Office hours, I decided to go with a different approach, which is passing props down to the elements to show the signals instead of having a context on each rail or grid we have on the app, mainly due to the amount of change and maintainability that would be needed for the providers across the app.

I'll update the PR and continue with the other approach, but please let me know if you have any concerns

PR #10720, opened 2024-09-05
chore: enable sentry performance metrics

brainbicycle:
This PR resolves [PHIRE-1142] 

### Description

This enables support for starting to collect tracing data in sentry by hooking into our react navigation setup following the docs [here](https://docs.sentry.io/platforms/react-native/tracing/instrumentation/react-navigation/).
 
You can see some metrics in our performance dashboard [here](https://artsynet.sentry.io/performance/?isDefaultQuery=false&metricSearchSetting=transactionsOnly&project=5867225&referrer=performance-transaction-summary&statsPeriod=7d&unselectedSeries=p100%28%29&unselectedSeries=avg%28%29).

### Follow ups:

- [ ] Go over the performance dashboard in mobile practice
- [ ] Look into enabling profiler on core screens 
- [ ] Review bootstrap set up, if it is necessary, should we do differently

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- enable navigation performance metrics in sentry - brian 

</details>

[PHIRE-1142]: https://artsyproduct.atlassian.net/browse/PHIRE-1142?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gkartalis (commented):
This looks great! 

Thank you for adding it to all NavigationContainer components! 👏👏👏

src/app/Scenes/Onboarding/OnboardingQuiz/OnboardingQuiz.tsx
```diff
@@ -46,9 +48,17 @@ export const OnboardingQuiz = () => {
     })
   }
 
+  const navContainerRef = useRef(null)
```
gkartalis, line 51: **question(non-blocking):** Do we need to explicitly declare the type here or is it automatically taken from the context?

src/app/Scenes/Home/HomeContainer.tsx
```diff
@@ -38,3 +39,5 @@ export const HomeContainer = () => {
 
   return <HomeQueryRenderer />
 }
+
+export const HomeContainer = Sentry.withProfiler(InnerHomeContainer)
```
brainbicycle, line 43: not 100% this is doing anything currently, double check, this should give us more react profile style metrics in home screen transactions

src/app/App.tsx
```diff
@@ -154,6 +154,7 @@ const Main = () => {
 const InnerApp = () => (
   <Providers>
     <AsyncStorageDevtools />
+    <SentrySetup />
```
brainbicycle, line 157: a bit weird but we need sentry set up before the Main nav stack is setup which happens early in the lifecycle, making this a parent component does that plus will refresh if the dev toggle or env is changed

src/app/system/errorReporting/setupSentry.ts
```diff
@@ -0,0 +1,104 @@
...
+        beforeNavigate: (context: TransactionContext) => {
+          /**
+           * This is a hack because our navigation setup is weird.
+           * We have a main modal stack at the root and then every screen presented
+           * in a tab gets its own stack. Sentry is registering every screen in a tab as
+           * a screen transaction for the root screen. For example, tapping on an
+           * artwork in the home screen will fire as `screen:home` instead of `screen:artwork`.
+           * To fix it we find the focused route and update the context name to the correct screen name.
+           * This should be updated if we change our navigation setup.
+           */
```
brainbicycle, line 85: funkiness do to our nav set up, let me know if the comment makes sense, the good news is most screens fire and get registered as transactions but we need to update the names manually

src/app/App.tsx
```diff
@@ -163,4 +164,5 @@ const InnerApp = () => (
   </Providers>
 )
 
-export const App = codePush(codePushOptions)(InnerApp)
+const SentryApp = Sentry.wrap(InnerApp)
```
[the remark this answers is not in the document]
brainbicycle: this is recommended by the docs though: https://docs.sentry.io/platforms/react-native/tracing/instrumentation/automatic-instrumentation/

index-common.js
```diff
@@ -17,7 +17,7 @@ if (__DEV__) {
   } catch {}
 }
 
-require("./src/app/system/errorReporting/sentrySetup").setupSentry({ environment: "bootstrap" })
+require("./src/app/system/errorReporting/setupSentry").setupSentry({ environment: "bootstrap" })
```
brainbicycle, line 20: I think we should probably remove this, I will do as separate pr to avoid too much in this one. I am guessing the intention was to initialized sentry early to catch errors early in app lifecycle, but I think we pay for it with complexity in our reporting having and extra env, and initializing 2x might be leading to subtler bugs in our reporting.

src/app/Scenes/Onboarding/OnboardingQuiz/OnboardingQuiz.tsx
brainbicycle: we shouldn't do this at all actually, registering the other breaks the main nav capturing events, the good news is that somehow most of our nav events end up going through the main nav

ArtsyOpenSource:

- Dev changes (enable navigation performance metrics in sentry - brian)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 2c1695efd465ac3996a061d30b3a96e0d406c325
</p>

src/app/store/config/features.ts
```diff
@@ -330,8 +332,29 @@ export const devToggles: { [key: string]: DevToggleDescriptor } = {
...
+        environment: environment.env,
+        debug: value,
+      })
+
+      if (value) {
+        toast.show("Sentry debugging enabled", "middle")
+      } else {
+        toast.show("Sentry debugging disabled", "middle")
+      }
+    },
```
brainbicycle, line 357: since sentry set up is no longer in a hook because it needs to be earlier in the lifecycle, call it manually on change.

gkartalis (approved):
I like the idea of the dev toggle to change env, thanks for adding that!

src/app/system/errorReporting/setupSentry.ts
```diff
@@ -0,0 +1,105 @@
...
+          if (__unsafe_mainModalStackRef.current) {
```
gkartalis, line 87: should we do this like `?.` to be safer?
```suggestion
          if (__unsafe_mainModalStackRef?.current) {
```

gkartalis (approved):

PR #10738, opened 2024-09-10
feat: adding display of showing collector signal to artwork page

oxaudo:
This PR resolves [https://artsyproduct.atlassian.net/browse/EMI-2021]

### Description
The info looks like this:

Screen Shot 2024-09-09 at 6 07 08 PM

### PR Checklist

- [ x ] I hid my changes behind a **[feature flag]**, or they don't need one.
- [ x ] I have included **screenshots** or **videos**, or I have not changed the UI.
- [ x ] I added an **[app state migration]**, or my changes do not require one.
- [ x ] I have documented any **follow-up work** that this PR will require, or it does not require any.
- [ x ] I have added a **changelog entry** below, or my changes do not require one.

### To the reviewers 👀

- Added display for the show/fair the artwork is in on the artwork details page

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>


src/app/Scenes/Artwork/Artwork.tsx
```diff
@@ -301,6 +302,13 @@ export const Artwork: React.FC<ArtworkProps> = (props) => {
         excludeVerticalMargin: true,
       })
 
+      sections.push({
+        key: "artworkCollectorSignals",
```
oxaudo, line 306: Not 100% sure about this naming as I think we might have some other signal display in some other area on the page? Not sure what is the plan here. Can name it more concrete as `artworkShowingNowCollectorSignal`. What you think?

src/app/Scenes/Artwork/Artwork.tsx
```diff
@@ -301,6 +302,13 @@ export const Artwork: React.FC<ArtworkProps> = (props) => {
...
+        element: <ArtworkCollectorSignalsFragmentContainer artwork={artworkAboveTheFold} />,
+        excludeSeparator: true,
+        excludeVerticalMargin: true,
```
oxaudo, line 309: Not completely sure if we want to exclude this VerticalMargin here. But was following the pattern...

ArtsyOpenSource:

- Cross-platform user-facing changes (Added display for the show/fair the artwork is in on the artwork details page)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against a427f38c63af9136d55acc8906785d37b3aa1953
</p>

src/app/Scenes/Artwork/Components/ArtworkCollectorSignals.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+const ArtworkCollectorSignals: React.FC<ArtworkCollectorSignalsProps> = ({ artwork }) => {
+  const { collectorSignals } = artwork
+  const runningShow = collectorSignals?.runningShow
+
+  if (!runningShow) {
+    return null
+  }
+
+  const showStartsAt: string = DateTime.fromISO(runningShow.startAt || "").toFormat("MMM d")
```
oxaudo, line 19: Had to add this strange `|| ""` as the startAt and endAt defined as potentially nil though pretty sure that if the show exists as a running show it needs to have start/end and href. Not sure if there is a better way to go around this.

src/app/Scenes/Artwork/Components/ArtworkCollectorSignals.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const showEndAt: string = DateTime.fromISO(runningShow.endAt || "").toFormat("MMM d")
+
+  return (
+    <Box mt={4} mb={2}>
```
oxaudo, line 23: The margins seem to be upped by 2 in this case but that is how design seems to look: https://www.figma.com/design/22RFPSkUu97g0ska3azFL2/Social-proof-%26-urgency-signals?node-id=2315-28859&node-type=canvas&m=dev

src/app/Scenes/Artwork/Components/ArtworkCollectorSignals.tsx
```diff
@@ -0,0 +1,57 @@
...
+      <Flex testID="artwork-collector-ssignals" flexDirection="row" alignContent="center">
+        <FairIcon mr={0.5} fill="black60" height={25} width={25} />
+        <Flex testID="artwork-collector-ssignals" flexDirection="column" alignContent="left">
+          <Text variant="sm" color="black100">
+            Showing now{"  •  "}
```
oxaudo, line 28: Think it's safe to just concat it as string here as I believe all of those will be available if the show is running.

oxaudo (commented):
@MrSltun - adding some comments inline as a newbie here. Want to see your opinion before I add specs here.

src/app/Scenes/Artwork/Artwork.tsx
MrSltun: I think the name should be changed like you suggested `artworkShowingNowCollectorSignal` :)

src/app/Scenes/Artwork/Artwork.tsx
MrSltun: I think it works here, since if you had vertical margin, you would have 40px above and below the component ([reference](https://github.com/artsy/eigen/blob/main/src/app/Scenes/Artwork/Artwork.tsx#L553))

src/app/Scenes/Artwork/Components/ArtworkCollectorSignals.tsx
```diff
@@ -0,0 +1,57 @@
...
+const ArtworkCollectorSignals: React.FC<ArtworkCollectorSignalsProps> = ({ artwork }) => {
+  const { collectorSignals } = artwork
+  const runningShow = collectorSignals?.runningShow
+
+  if (!runningShow) {
+    return null
+  }
+
+  const showStartsAt: string = DateTime.fromISO(runningShow.startAt || "").toFormat("MMM d")
+  const showEndAt: string = DateTime.fromISO(runningShow.endAt || "").toFormat("MMM d")
```
MrSltun, line 20: **Suggestion**: I believe that there's no need to add the type here since it infers it from the result of the function :)
```suggestion
  const showEndAt = DateTime.fromISO(runningShow.endAt || "").toFormat("MMM d")
```

src/app/Scenes/Artwork/Components/ArtworkCollectorSignals.tsx
MrSltun: In my opinion, I think it should be symmetrical so it looks neat, I also noted that there's a design with a symmetrical one where there's 40px vertical margin :)

src/app/Scenes/Artwork/Components/ArtworkCollectorSignals.tsx
MrSltun: It's okay to concat strings here, but we could use the `bullet` variable from palette mobile ([example](https://github.com/artsy/eigen/blob/main/src/app/Components/Artist/ArtistInsights/ArtistInsightsAuctionResults.tsx#L255))

src/app/Scenes/Artwork/Components/ArtworkCollectorSignals.tsx
```diff
@@ -0,0 +1,57 @@
...
+            endAt
+            href
+            name
+            startAt
+          }
+        }
+      }
+    `,
+  }
+)
```
MrSltun, line 57: **Suggestion**: Maybe we could use the new hook method instead of the fragment container?
here's [an example](https://github.com/artsy/eigen/blob/main/src/app/Scenes/Activity/components/NotificationArtworkList.tsx) of a component that uses `useFragment`

src/app/Scenes/Artwork/Components/ArtworkCollectorSignals.tsx
oxaudo: I started from 40ox each side but looks like the component that follows include 20px at the top so the bottom margin becomes too big. If we add 40 on both sides we should just remove that `excludeVerticalMargin` that you've mentioned before. No? But in this case space at the bottom of this `showing now` becomes too big...

src/app/Scenes/Artwork/Components/ArtworkCollectorSignals.tsx
oxaudo: I'll try to change it as per example. Though not exactly sure what is the tradeoffs of one vs the other?

src/app/__fixtures__/ArtworkFixture.ts
```diff
@@ -669,5 +669,9 @@ export const ArtworkFixture = {
       ],
     },
   },
-  collectorSignals: null,
+  collectorSignals: {
```
oxaudo, line 672: I thought that this might be the reason for tests fail as now collectorSignals is in the object format. But does not seem to work either :(

src/app/Scenes/Artwork/Components/ArtworkCollectorSignals.tsx
damassi: Yah at this point in 2024 we should be using hooks everywhere, including web!

src/app/Scenes/Artwork/Components/ArtworkShowingNowCollectorSignal.tests.tsx
```diff
@@ -0,0 +1,52 @@
...
+        artwork(id: "example") {
+          ...ArtworkShowingNowCollectorSignal_artwork
+        }
+      }
+    `,
+  })
+
+  beforeEach(() => {
+    // __globalStoreTestUtils__?.injectFeatureFlags({ AREnableCuratorsPicksAndInterestSignals: true })
+  })
```
MrSltun, line 21: Maybe we should remove this? 👀

src/app/Scenes/Artwork/Components/ArtworkShowingNowCollectorSignal.tests.tsx
```diff
@@ -0,0 +1,52 @@
+import { screen } from "@testing-library/react-native"
+import { ArtworkShowingNowCollectorSignal } from "app/Scenes/Artwork/Components/ArtworkShowingNowCollectorSignal"
+// import { __globalStoreTestUtils__ } from "app/store/GlobalStore"
```
MrSltun, line 3: as well as this one?

src/app/Scenes/Artwork/Components/ArtworkShowingNowCollectorSignal.tests.tsx
```diff
@@ -0,0 +1,52 @@
...
+      Artwork: () => ({
+        collectorSignals: {
+          runningShow: null,
+        },
+      }),
+    })
+
+    expect(screen.queryByText("Showing now")).not.toBeOnTheScreen()
+  })
+})
```
MrSltun, line 52: 👏👏👏👏

MrSltun (approved):
Looks great! I just added some comments :)

src/app/Scenes/Artwork/Components/ArtworkShowingNowCollectorSignal.tests.tsx
oxaudo: Oh... yea - I was wondering if I should add the flag just to test what happens if it's enabled but then figured will do this one in the next pr...

MrSltun (approved):

PR #10749, opened 2024-09-11
feat(EMI-2027): Add Curators' Pick and Increased Interest signals

MrSltun:
This PR resolves [EMI-2027] 

> [!NOTE]
> This work is hidden behind `AREnableCuratorsPicksAndInterestSignals` feature flag

### Description
This PR adds the new Increased Interest and Curator's Pick signals on artwork rail cards, grid items, and artwork details page

### Screenshots
| | Android | iOS |
|---|---|---|
| Artwork screen - increased interest | image | image |
| Artwork screen - curator's pick | image | image |
| Artwork rail - increased interest | image | image |
| Artwork rail - curator's pick | image | image |
| Artwork grid - increased interest | image | image |
| Artwork grid - curator's pick | image | image |

### PR Checklist

### To the reviewers 👀

- add new Increased Interest and Curator's Pick signals (hidden behind FF) - mrsltun

#### iOS user-facing changes

-

-

#### Dev changes

-

cc @artsy/emerald-devs 

</details>

[EMI-2027]: https://artsyproduct.atlassian.net/browse/EMI-2027?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Components/ArtworkGrids/ArtworkGridItem.tsx
+++ b/src/app/Components/ArtworkGrids/ArtworkGridItem.tsx
@@ -19,6 +19,7 @@ import { CreateArtworkAlertModal } from "app/Components/Artist/ArtistArtworks/Cr
 import { filterArtworksParams } from "app/Components/ArtworkFilter/ArtworkFilterHelpers"
 import { ArtworksFiltersStore } from "app/Components/ArtworkFilter/ArtworkFilterStore"
 import { ArtworkAuctionTimer } from "app/Components/ArtworkGrids/ArtworkAuctionTimer"
+import { ArtworkSocialSignal } from "app/Components/ArtworkGrids/ArtworkSocialSignal"
 import { useSaveArtworkToArtworkLists } from "app/Components/ArtworkLists/useSaveArtworkToArtworkLists"
 import { ContextMenuArtwork } from "app/Components/ContextMenu/ContextMenuArtwork"
 import { DurationProvider } from "app/Components/Countdown"
@@ -80,6 +81,8 @@ export interface ArtworkProps extends ArtworkActionTrackingProps {
   /** allows for artwork to be added to recent searches */
   updateRecentSearchesOnTap?: boolean
   urgencyTagTextStyle?: TextProps
+  hideIncreasedInterestSignal?: boolean
+  hideCuratorsPickSignal?: boolean
 }
 
 export const Artwork: React.FC<ArtworkProps> = ({
@@ -110,6 +113,8 @@ export const Artwork: React.FC<ArtworkProps> = ({
   trackTap,
   updateRecentSearchesOnTap = false,
   urgencyTagTextStyle,
+  hideIncreasedInterestSignal = false,
+  hideCuratorsPickSignal = false,
 }) => {
   const itemRef = useRef<any>()
   const color = useColor()
@@ -355,6 +360,13 @@ export const Artwork: React.FC<ArtworkProps> = ({
                     </Text>
                   </Box>
                 )}
+                {!isAuction && !displayLimitedTimeOfferSignal && !!collectorSignals && (
+                  <ArtworkSocialSignal
+                    collectorSignals={collectorSignals}
+                    hideCuratorsPick={hideCuratorsPickSignal}
+                    hideIncreasedInterest={hideIncreasedInterestSignal}
+                  />
+                )}
                 {!!showLotLabel && !!artwork.saleArtwork?.lotLabel && (
                   <>
                     <Text variant="xs" numberOfLines={1} caps {...lotLabelTextStyle}>
@@ -592,6 +604,7 @@ export default createFragmentContainer(Artwork, {
           lotClosesAt
         }
         ...ArtworkAuctionTimer_collectorSignals
+        ...ArtworkSocialSignal_collectorSignals
       }
       ...useSaveArtworkToArtworkLists_artwork
     }
--- a/src/app/Components/ArtworkGrids/ArtworkGridItem.tsx
+++ b/src/app/Components/ArtworkGrids/ArtworkGridItem.tsx
@@ -123,6 +123,9 @@ export const Artwork: React.FC<ArtworkProps> = ({
   const showBlurhash = useFeatureFlag("ARShowBlurhashImagePlaceholder")
   const AREnablePartnerOfferSignals = useFeatureFlag("AREnablePartnerOfferSignals")
   const AREnableAuctionImprovementsSignals = useFeatureFlag("AREnableAuctionImprovementsSignals")
+  const AREnableCuratorsPicksAndInterestSignals = useFeatureFlag(
+    "AREnableCuratorsPicksAndInterestSignals"
+  )
 
   let filterParams: any = undefined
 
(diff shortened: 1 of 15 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (add new Increased Interest and Curator's Pick signals (hidden behind FF) - mrsltun)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against bd00b3c6a67663c6455aa18ad009c6a25f29a989
</p>

src/app/Scenes/Artwork/Components/ArtworkCuratorsPickIncreasedInterestCollectorSignal.tsx
```diff
@@ -0,0 +1,58 @@
...
+  let SignalIcon = VerifiedIcon
+
+  if (increasedInterest) {
+    singalTitle = "Increased Interest"
+    signalDescription = "Based on collector activity in the past 14 days"
+    SignalIcon = TrendingIcon
+  }
+
+  return (
+    <Flex flexDirection="row" mt={4} mb={2}>
```
oxaudo, line 35: Curious how would this look if artwork has both curator's pic and showing now in some show. Wonder if we need to account for that.

src/app/Scenes/Artwork/Artwork.tsx
```diff
@@ -672,6 +687,7 @@ export const ArtworkContainer = createRefetchContainer(
         ...ArtworkPartnerOfferNote_artwork
         ...ArtworkPrice_artwork
         ...ArtworkDimensionsClassificationAndAuthenticity_artwork
+        ...ArtworkCuratorsPickIncreasedInterestCollectorSignal_artwork
```
oxaudo, line 690: So strange. How this is not failing Artwork.spec and my change which is basically the same does:(

src/app/Components/ArtworkGrids/ArtworkGridItem.tsx
```diff
@@ -355,6 +363,16 @@ export const Artwork: React.FC<ArtworkProps> = ({
...
+                {!isAuction &&
+                  !displayLimitedTimeOfferSignal &&
+                  !!collectorSignals &&
+                  !!AREnableCuratorsPicksAndInterestSignals && (
+                    <ArtworkSocialSignal
+                      collectorSignals={collectorSignals}
+                      hideCuratorsPick={hideCuratorsPickSignal}
+                      hideIncreasedInterest={hideIncreasedInterestSignal}
+                    />
+                  )}
```
erikdstock, line 375: I wonder why the partner offer signal is [treated differently](https://github.com/artsy/eigen/pull/10749/files#diff-5f044e56abd8cbe34fe5015f877c3a670c61603f43cdfb481391269a965ac370R271-R276) and whether these can all go in a single label signal.

Edit: See comment below

src/app/Components/ArtworkGrids/ArtworkSocialSignal.tests.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  describe("increased interest signal", () => {
+    it("renders the increased interest signal", () => {
+      renderWithRelay({
+        Artwork: () => ({
+          collectorSignals: {
+            increasedInterest: true,
+            curatorsPick: false,
+          },
+        }),
```
erikdstock, line 31: You should be able to use `collectorSignals.primaryLabel` to get a string enum of `'limited_time_offer' | 'increased_interest' | etc`. Can we do that here? Then you don't have to worry about prioritization.

src/app/Components/ArtworkGrids/ArtworkSocialSignal.tsx
```diff
@@ -0,0 +1,70 @@
...
+  hideIncreasedInterest,
+  dark = false,
+}) => {
+  const { curatorsPick, increasedInterest } = useFragment(fragment, collectorSignals)
+
+  if ((hideIncreasedInterest && hideCuratorsPick) || (!curatorsPick && !increasedInterest)) {
+    return null
+  }
+
+  const primaryColor = dark ? "white100" : "black100"
```
erikdstock, line 23: Is this early return necessary?

src/app/Components/ArtworkGrids/ArtworkSocialSignal.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  switch (true) {
+    case increasedInterest && !hideIncreasedInterest:
+      return (
```
erikdstock, line 27: I think this could be `switch (collectorSignals.primaryLabel)` - but in that case we might need some additional fallback logic if the primary label is hidden but another is available.

erikdstock (commented):
Nothing blocking, but I would be interested to see if we can use `primaryLabel` here eventually, as well as combining the partner offer label into this one component.

However, with the new requirements to suppress labels from above I now wonder how primary label should handle that - it seems like the prioritization might wind up on the client regardless (unless we parameterize the MP `artwork.collectorSignals` field)... For this reason I think all of these comments are non-blocking.

src/app/Components/ArtworkGrids/ArtworkSocialSignal.tsx
damassi: `switch (true)` is a nice secret weapon, and powerful for pattern matching like seen the above 😎

src/app/Scenes/Artwork/Artwork.tsx
MrSltun: I think because my work is wrapped behind a feature flag, which is false by default, so it doesn't change the tests

src/app/Components/ArtworkGrids/ArtworkSocialSignal.tsx
araujobarret: never thought of using it in this way, I liked it!

src/app/Scenes/Artwork/Components/ArtworkCuratorsPickIncreasedInterestCollectorSignal.tsx
```diff
@@ -0,0 +1,58 @@
...
+interface Props {
+  artwork: ArtworkCuratorsPickIncreasedInterestCollectorSignal_artwork$key
+}
+
+export const ArtworkCuratorsPickIncreasedInterestCollectorSignal: React.FC<Props> = ({
+  artwork,
+}) => {
+  const { collectorSignals } = useFragment(fragment, artwork)
+
+  if (!collectorSignals) {
```
araujobarret, line 14: better for readability:
```suggestion
  if (!collectorSignals || (!collectorSignals.increasedInterest && !collectorSignals.curatorsPick)) {
    return null
  }
```
then you won't need the check on line 20.

src/app/Scenes/Artwork/Components/ArtworkCuratorsPickIncreasedInterestCollectorSignal.tsx
araujobarret: let's avoid margins here, it makes hard to compose it and controlling the spacing from the parent, lesgo with padding

src/app/Scenes/Collection/Screens/CollectionArtworks.tsx
```diff
@@ -27,6 +27,8 @@ interface CollectionArtworksProps {
   relay: RelayPaginationProp
 }
 
+const CURATORS_PICKS_SLUGS = ["curators-picks-emerging-artists", "curators-picks-blue-chip-artists"]
```
araujobarret, line 30: shouldn't this come from MP or Gravity?

src/app/Components/ArtworkGrids/ArtworkSocialSignal.tsx
MrSltun: we can remove it :)

src/app/Scenes/Collection/Screens/CollectionArtworks.tsx
MrSltun: Not sure if there's a service for it already, but we are using it to check if the current collection is one of these, where we don't want to show the signals in these grids

src/app/Components/ArtworkGrids/ArtworkSocialSignal.tests.tsx
MrSltun: we can do that and refactor everything to use the `primaryLabel`, but let's do it in a follow-up PR

PR #10760, opened 2024-09-12
feat: prefetch home view query

MounirDhahri:
### Description

This PR adds prefetching to the new home view.

This saves us about `400ms`of waiting the home query.

https://github.com/user-attachments/assets/878de37d-98ab-4897-a248-e7488b2422ec

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- prefetch new home view query - mounir

</details>

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -10,6 +10,7 @@ import { setupSentry } from "app/system/errorReporting/setupSentry"
 import { ModalStack } from "app/system/navigation/ModalStack"
 import { usePurgeCacheOnAppUpdate } from "app/system/relay/usePurgeCacheOnAppUpdate"
 import { useDevToggle } from "app/utils/hooks/useDevToggle"
+import { useFeatureFlag } from "app/utils/hooks/useFeatureFlag"
 import { addTrackingProvider } from "app/utils/track"
 import {
   SEGMENT_TRACKING_PROVIDER,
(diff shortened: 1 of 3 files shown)


src/app/App.tsx
```diff
@@ -135,6 +140,10 @@ const Main = () => {
 
   useEffect(() => {
     if (isLoggedIn) {
+      if (shouldDisplayNewHomeView) {
+        prefetchUrl("/")
+      }
```
olerichter00, line 145: To avoid nested if statements and make this easier to read:
```suggestion
    if (isLoggedIn && shouldDisplayNewHomeView) {
      prefetchUrl("/")
```

src/app/AppRegistry.tsx
```diff
@@ -488,9 +489,15 @@ export const modules = defineModules({
...
-    isRootViewForTabName: "home",
-  }),
+  Home: reactModule(
+    HomeContainer,
+    {
+      isRootViewForTabName: "home",
+    },
+    ArtsyNativeModule.isBetaOrDev && !unsafe_getFeatureFlag("ARPreferLegacyHomeScreen")
+      ? [homeViewScreenQuery]
+      : undefined
```
olerichter00, line 499: I think we can just add the new home view query here because it's not used anywhere else.

```suggestion
    [homeViewScreenQuery]
```

olerichter00 (commented):
Amazing! 🌟

src/app/App.tsx
MounirDhahri: nice idea but I would have to check again for `isLoggedIn` since that's important for line 147 where we call `savePendingToken`

Do you mean something like

```typescript
if (isLoggedIn && shouldDisplayNewHomeView) {
	 prefetchUrl("/")
}
if (isLoggedIn) {
	 savePendingToken()
}
```

src/app/AppRegistry.tsx
MounirDhahri: True - I wanted this as a safe guard just in case we would accidentally call the prefetch home query somewhere and we would end up exposing a query with an interface that is still changing  and break clients. Maybe I went too far here since I am also checking before calling `prefetchURL("/")`

src/app/App.tsx
olerichter00: oh, I haven't that there is more in the first if-statement...

src/app/AppRegistry.tsx
olerichter00: Yeah, both should be fine, and we can clean this up when we remove the feature flag. 👍

PR #10763, opened 2024-09-12
feat: Implement Loading Skeletons for all home view sections

MounirDhahri:
This PR resolves [ONYX-1155] 

### Description
This PR adds the loading skeletons to the home view sections.
Here is exactly what I did:
- For each section, convert the section into a query renderer
- then, implement the placeholder
- then remove the query from the homeview level

After making my changes, I noticed that the homeview screen performance dropped, so I switched to `Flashlist` for out of the box performance improvements. 

https://github.com/user-attachments/assets/17aa19b3-ccf2-4f9a-9684-0bbc6e4d0b31

### PR Checklist

### To the reviewers 👀

- Add loading skeletons to home view sections - mounir

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[ONYX-1155]: https://artsyproduct.atlassian.net/browse/ONYX-1155?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -1,6 +1,6 @@
 import { GoogleSignin } from "@react-native-google-signin/google-signin"
 import * as Sentry from "@sentry/react-native"
-import { NUMBER_OF_SECTIONS_TO_LOAD } from "app/Scenes/HomeView/HomeView"
+import { homeViewScreenQueryVariables } from "app/Scenes/HomeView/HomeView"
 import { GlobalStore, unsafe__getEnvironment, unsafe_getDevToggle } from "app/store/GlobalStore"
 import { codePushOptions } from "app/system/codepush"
 import { AsyncStorageDevtools } from "app/system/devTools/AsyncStorageDevTools"
@@ -142,9 +142,7 @@ const Main = () => {
   useEffect(() => {
     if (isLoggedIn) {
       if (shouldDisplayNewHomeView) {
-        prefetchUrl("/", {
-          count: NUMBER_OF_SECTIONS_TO_LOAD,
-        })
+        prefetchUrl("/", homeViewScreenQueryVariables())
       }
 
       savePendingToken()
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -1,5 +1,6 @@
 import { GoogleSignin } from "@react-native-google-signin/google-signin"
 import * as Sentry from "@sentry/react-native"
+import { NUMBER_OF_SECTIONS_TO_LOAD } from "app/Scenes/HomeView/HomeView"
 import { GlobalStore, unsafe__getEnvironment, unsafe_getDevToggle } from "app/store/GlobalStore"
 import { codePushOptions } from "app/system/codepush"
 import { AsyncStorageDevtools } from "app/system/devTools/AsyncStorageDevTools"
@@ -141,7 +142,9 @@ const Main = () => {
   useEffect(() => {
     if (isLoggedIn) {
       if (shouldDisplayNewHomeView) {
-        prefetchUrl("/")
+        prefetchUrl("/", {
+          count: NUMBER_OF_SECTIONS_TO_LOAD,
+        })
       }
 
       savePendingToken()
--- a/src/app/Scenes/HomeView/HomeView.tsx
+++ b/src/app/Scenes/HomeView/HomeView.tsx
@@ -97,6 +97,7 @@ export const HomeView: React.FC = () => {
             ) : null
           }
           refreshControl={<RefreshControl refreshing={isRefreshing} onRefresh={handleRefresh} />}
+          onEndReachedThreshold={3}
         />
       </Screen.Body>
     </Screen>
@@ -177,7 +178,6 @@ const sectionsFragment = graphql`
             }
             ... on HomeViewSectionGalleries {
               internalID
-              ...HomeViewSectionGalleries_section
             }
           }
         }
--- a/src/app/Scenes/HomeView/HomeView.tsx
+++ b/src/app/Scenes/HomeView/HomeView.tsx
@@ -153,7 +153,6 @@ const sectionsFragment = graphql`
             }
             ... on HomeViewSectionArtists {
               internalID
-              ...HomeViewSectionArtists_section
             }
             ... on HomeViewSectionAuctionResults {
               internalID
--- a/src/app/Scenes/HomeView/HomeView.tsx
+++ b/src/app/Scenes/HomeView/HomeView.tsx
@@ -1,5 +1,6 @@
 import { ContextModule, OwnerType } from "@artsy/cohesion"
 import { Flex, Screen, Spinner, Text } from "@artsy/palette-mobile"
+import { FlashList } from "@shopify/flash-list"
 import { HomeViewQuery } from "__generated__/HomeViewQuery.graphql"
 import { HomeViewSectionsConnection_viewer$key } from "__generated__/HomeViewSectionsConnection_viewer.graphql"
 import { HomeView_me$key } from "__generated__/HomeView_me.graphql"
@@ -30,7 +31,7 @@ const NUMBER_OF_SECTIONS_TO_LOAD = 100
 export const HOME_VIEW_SECTIONS_SEPARATOR_HEIGHT = "30px"
 
 export const HomeView: React.FC = () => {
-  const flatlistRef = useBottomTabsScrollToTop("home")
+  const flashlistRef = useBottomTabsScrollToTop("home")
   const [isRefreshing, setIsRefreshing] = useState(false)
 
   const queryData = useLazyLoadQuery<HomeViewQuery>(homeViewScreenQuery, {
@@ -80,8 +81,8 @@ export const HomeView: React.FC = () => {
   return (
     <Screen safeArea={false}>
       <Screen.Body fullwidth>
-        <Screen.FlatList
-          innerRef={flatlistRef}
+        <FlashList
+          ref={flashlistRef}
           data={sections}
           keyExtractor={(item) => `${item.internalID || ""}`}
           renderItem={({ item }) => {
@@ -89,6 +90,7 @@ export const HomeView: React.FC = () => {
           }}
           onEndReached={() => loadNext(10)}
           ListHeaderComponent={<HomeHeader />}
+          estimatedItemSize={500}
           ListFooterComponent={
             hasNext ? (
               <Flex width="100%" justifyContent="center" alignItems="center" height={200}>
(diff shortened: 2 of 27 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (Add loading skeletons to home view sections - mounir)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 67cf1cebf7b3c2bf78291bb1208c422761a94f19
</p>

src/app/Scenes/HomeView/Sections/HomeViewSectionHeroUnits.tsx
```diff
@@ -39,6 +39,10 @@ export const HomeViewSectionHeroUnits: React.FC<HomeViewSectionHeroUnitsProps> =
 
   const { width } = useScreenDimensions()
 
+  if (!heroUnits || heroUnits.length === 0) {
+    return null
```
anandaroop, line 43: **praise:** 🙏🏽 I just noticed this yesterday

anandaroop (dismissed):
Amazing! 🤩

src/app/Scenes/HomeView/HomeView.tsx
```diff
@@ -107,53 +110,39 @@ const sectionsFragment = graphql`
             }
             ... on HomeViewSectionActivity {
```
dblandin, line 111: **suggestion**: One anticipated benefit of this change was being able to remove any named fragment spreads from this initial query. Since we're already querying for internalID within the `HomeViewSectionGeneric` fragment spread, can we now remove all of these others?

src/app/Scenes/HomeView/Sections/HomeViewSectionArtworks.tsx
```diff
@@ -107,3 +115,57 @@ const fragment = graphql`
...
+const HomeViewSectionArtworksPlaceholder: React.FC = () => {
+  const randomValue = useMemoizedRandom()
+  return (
+    <Skeleton>
+      <Flex mx={2} my={HOME_VIEW_SECTIONS_SEPARATOR_HEIGHT}>
+        <SkeletonText variant="lg-display">Arwtworks Rail</SkeletonText>
+        <Spacer y={2} />
+
+        <Flex flexDirection="row">
+          <Join separator={<Spacer x="15px" />}>
```
dblandin, line 138: **suggestion**: I see `15px` in few different spots in this diff. Shouldn't we stick with standard spacing values in the design system?

dblandin (commented):
Awesome! This is looking really promising. Great work!! 💯

src/app/Scenes/HomeView/HomeView.tsx
```diff
@@ -10,7 +10,7 @@ import { Suspense, useState } from "react"
 import { RefreshControl } from "react-native"
 import { fetchQuery, graphql, useLazyLoadQuery, usePaginationFragment } from "react-relay"
 
-const NUMBER_OF_SECTIONS_TO_LOAD = 10
+const NUMBER_OF_SECTIONS_TO_LOAD = 100
```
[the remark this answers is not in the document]
dblandin: A value of `100` feels to me like avoiding pagination altogether. Shouldn't we stick with 10, or even reduce to 5 if the speed has improved enough to support it?

src/app/Scenes/HomeView/Sections/HomeViewSectionActivity.tsx
```diff
@@ -131,6 +137,40 @@ const homeViewSectionActivityQuery = graphql`
   }
 `
 
+export const HomeViewSectionActivityPlaceholder: React.FC = () => {
+  const randomValue = useMemoizedRandom()
+
+  return (
+    <Flex ml={2} mr={2} my={2}>
+      <SkeletonText variant="lg-display">Latest Activity</SkeletonText>
```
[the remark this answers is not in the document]
MounirDhahri: I tried that initially but in case the component is not going to be available (no activities), you will get a loading placeholder with "Latest Activity" disappearing next. I felt it's less awkward if a Skeleton disappears than a section text disappearing with it

src/app/Scenes/HomeView/HomeView.tsx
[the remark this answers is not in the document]
MounirDhahri: oh yeah - this needs to be reverted - good catch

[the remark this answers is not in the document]
MounirDhahri: I needed it for testing

src/app/Scenes/HomeView/HomeView.tsx
```diff
@@ -51,15 +52,16 @@ export const HomeView: React.FC = () => {
...
+        <FlashList
+          ref={flashlistRef}
           data={sections}
           keyExtractor={(item) => `${item.internalID || ""}`}
           renderItem={({ item }) => {
             return <Section section={item} />
           }}
           onEndReached={() => loadNext(10)}
           ListHeaderComponent={<HomeHeader />}
+          estimatedItemSize={500}
```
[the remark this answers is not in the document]
MounirDhahri: Good question, `Flashlist` does some computation under the hood and needs estimates to be able to render components initially - I needed to come up with a value and here I chose the Artworks Rail size by using "large artwork rail image height" + "metadata height" + "section title". Unfortunately there is no way to give an estimate by component type so I had to choose something.

src/app/Scenes/HomeView/HomeView.tsx
MounirDhahri: download

src/app/Scenes/HomeView/Sections/HomeViewSectionArtworks.tsx
MounirDhahri: That was a weird thing for me as well but I wanted to keep the components spacings as it is until we discuss it with Design. I see a commit here from Pavlos where we were intentional about it https://github.com/artsy/eigen/pull/8165 and it's used in all artwork rails currently. Maybe we can address that separately.

damassi:
This is looking great! Couple qs:

- What would it look like to have the first two sections ("above the fold") in the main query, and then lazy load the rest on scroll? We need to keep performance top-of-mind, but thinking that we also need to keep "native app" even closer. (If the performance is terrible, better to lean into performant solutions first, and then improve towards native experiences incrementally)
- Can we move the lazy load trigger threshold way further down? By the time the content scrolls into view, it would (ideally) be loaded so that the user doesn't see any skeletons.

dblandin:
> What would it look like to have the first two sections ("above the fold") in the main query

I think the challenge is that if we want to fetch the entire section, including its type-specific data, we might have to bring back all of [these named fragment spreads](https://github.com/artsy/eigen/pull/10763/files#diff-46149b9dcf5250a5d5962c4035c90028926b990e80e9d4a62fe4e882166b8c75L108-L157) into the main query — which I'd very much like to avoid as that introduces too much coupling between this top-level query and concrete types.

I think @MounirDhahri's other pull request here attempts to [prefetch this query](https://github.com/artsy/eigen/pull/10760) so the lazy loading can start earlier. Open to other approaches but without knowing the type of section that might appear in those first two or so slots (above the fold), it would be difficult to load section-specific data here.

dblandin:
> Can we move the lazy load trigger threshold way further down?

Very much in favor of trying this 👍🏼

src/app/Scenes/HomeView/Sections/HomeViewSectionArtworks.tsx
dblandin: Gotcha. Yeah, I'm fine to defer but let's be sure to bring it up with design/Barney.

anandaroop:
> What would it look like to have the first two sections ("above the fold") in the main query

Tricky, because we can't know with certainty what the first two sections will be (MP will decide that on the fly, per user). We might be able to make some assumptions, but I worry then that we'd lose the flexibility that is one of the main design goals here.

> Can we move the lazy load trigger threshold 

Big +1 to this!

MounirDhahri:
> Can we move the lazy load trigger threshold

Let's definitely move it down - it will definitely help with the performance since we will be rendering less items at a time. I am also planning a parallel discovery [here](https://github.com/artsy/eigen/pull/10773) to avoid rendering more than we need.

MounirDhahri:
That was a nice idea - video after changes 👇 

https://github.com/user-attachments/assets/9a26eb8f-724a-47ae-81dc-4b6c4353555c

src/app/Scenes/HomeView/Sections/HomeViewSectionActivity.tsx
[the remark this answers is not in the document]
dblandin: > but in case the component is not going to be available (no activities), you will get a loading placeholder with "Latest Activity" disappearing next.

I think that's okay for the short-term. For components with no empty state design, it would be interesting for MP to make that decision. The section within MP could query for count of available activities and decide to not send down the section at all if `0`.

damassi:
> Tricky, because we can't know with certainty what the first two sections will be (MP will decide that on the fly, per user)

Makes total sense. NP 👍 

Re intersection points: this is much better! 👍 Moving it as far as one can would be good, since once the app is booted its booted, and stuff is going on in the background, below the fold. Might be worth pushing it down even more.

anandaroop:
Looking great!

**thought (non-blocking):** I realize now that the suspense placeholder that says "Loading home view…" is still present. Makes sense that we can't draw accurate skeletons at that stage, so I wonder what we should do with that one. (Not for this PR)

src/app/App.tsx
```diff
@@ -141,7 +142,9 @@ const Main = () => {
   useEffect(() => {
     if (isLoggedIn) {
       if (shouldDisplayNewHomeView) {
-        prefetchUrl("/")
+        prefetchUrl("/", {
+          count: NUMBER_OF_SECTIONS_TO_LOAD,
+        })
```
olerichter00, line 147: To ensure these variables always match the variables from the screen, I've introduced [this](https://github.com/artsy/eigen/blob/fbad3147589a0d19c68dddb2e6070f6f7c6191c4/src/app/Scenes/RecommendedAuctionLots/RecommendedAuctionLots.tsx#L65) pattern to make sure updates don't break the refetching. Can we use this pattern here as well?

MounirDhahri: sure - good idea to be aligned.

MounirDhahri:
@anandaroop I added a [commit](https://github.com/artsy/eigen/pull/10763/commits/d76b8280e5a6dabfd6dc3214be264c4440e8f43e) to update the placeholder as well - I am not so happy with the default transition on the simulator. I will wait for the build to take another look at it

MounirDhahri:
https://github.com/user-attachments/assets/e4d5e821-dc19-442f-b201-8e3d181420c1


It looks great now after updating the home view skeleton as well. I will merge the PR before I had to fix conflicts again on 34 commits 😄

PR #10778, opened 2024-09-13
fix: wait for animations to avoid dead deeplinks

brainbicycle:
This PR resolves [] 

### Description

Fixes deeplinks not being followed from killed app state. Wraps push request in a requestAnimationFrame, I believe the main thread is overloaded causing the call to be ignored otherwise.

<details>
<summary>
Before
</summary>

https://github.com/user-attachments/assets/083ad429-eca9-4aa7-a163-9c17e5c94f73

</details>

<details>
<summary>
After
</summary>

https://github.com/user-attachments/assets/952e589e-0acb-4c87-9e06-6c86f4f60c1a

</details>

#### Follow-ups

- [ ] It is probably worth seeing if we can understand all that is happening on main thread on startup from killed state, maybe we are slowing ourselves down needlessly?

### PR Checklist

### To the reviewers 👀

- fix broken push deeplink follow from killed state - brian 

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/system/navigation/navigate.ts
+++ b/src/app/system/navigation/navigate.ts
@@ -157,7 +157,7 @@ export async function navigate(url: string, options: NavigateOptions = {}) {
         requestAnimationFrame(pushView)
       }, timeoutTime)
     } else {
-      pushView()
+      requestAnimationFrame(pushView)
     }
   }
 }
--- a/src/app/system/navigation/navigate.tests.tsx
+++ b/src/app/system/navigation/navigate.tests.tsx
@@ -36,6 +36,14 @@ describe(navigate, () => {
   beforeEach(() => {
     Linking.openURL = jest.fn()
     LegacyNativeModules.ARScreenPresenterModule.pushView = jest.fn()
+    jest.spyOn(global, "requestAnimationFrame").mockImplementation((cb) => {
+      cb(0)
+      return 0
+    })
+  })
+
+  afterEach(() => {
+    jest.restoreAllMocks()
   })
 
   describe("routes to various screens", () => {


ArtsyOpenSource:

- Cross-platform user-facing changes (fix broken push deeplink follow from killed state - brian)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 26c57a783a3975a854a09b0ac9eeb45e3d49760d
</p>

src/app/system/navigation/navigate.ts
```diff
@@ -157,7 +157,7 @@ export async function navigate(url: string, options: NavigateOptions = {}) {
         requestAnimationFrame(pushView)
       }, timeoutTime)
     } else {
-      pushView()
+      requestAnimationFrame(pushView)
```
gkartalis, line 160: does this need to be like this? I guess not since it works 🤷 
```suggestion
      requestAnimationFrame(pushView())
```

gkartalis (approved):
awesome!

src/app/system/navigation/navigate.ts
brainbicycle: My favorite robot says (paraphrased) that `requestNavigationFrame(pushView())` will execute pushView immediately and pass the result to requestNavigationFrame, whereas `requestNavigationFrame(pushView)` passes the function reference and gives requestNavigationFrame control of when to execute pushView, which is what we want in this case. I didn't know that before though I just copied it from the similar call above so thanks for asking!

PR #10838, opened 2024-09-24
chore: only track unable to loads from app wide error boundary

brainbicycle:
This PR resolves [] 

### Description

This updates our tracking of unable to loads to only track those that came from the app wide error boundary. 
We use this error boundary in some screens as well as app wide. Those in screens are mostly okay, some improvement maybe can be made to the error screens but priority is the app wide error boundary since when users hit this they have no access to navigation. So hopefully by removing these from tracking we get a better sense of real problem and it is easier to find the culprits. 

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- only track unable to loads from app wide error boundary - brian 

</details>


ArtsyOpenSource:

- Dev changes (only track unable to loads from app wide error boundary - brian)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 7e2adc4f1e94a352066e08a25ac05accb6eced1d
</p>

MounirDhahri (approved):
Let's try this out. Even if it means we will get less errors reported. 
I have a follow-up question though, will this affect thrown errors from mutations etc. will we still get those reported?

brainbicycle:
> Let's try this out. Even if it means we will get less errors reported. I have a follow-up question though, will this affect thrown errors from mutations etc. will we still get those reported?

After testing some more, I think I need to make this a little more nuanced, will check on mutations (I think they will still get reported) but I found other unable to loads that aren't app wide but are also dead ends with no nav.

MounirDhahri:
Sounds good to me - thanks for looking into it

PR #10839, opened 2024-09-24
feat(EMI-2112): failed payment order history

erikdstock:
This PR resolves [EMI-2112]
### Description

This PR adds a `payment failed` state to the order history along with a CTA to fix the failed payment. The work to expose that new payment failed state is still in progress, so there are no user-facing changes yet.

The `getOrderStatus()` method now returns a `string | null` rather than the string `'unknown'`. Unhandled (null) values are not rendered in the app - this should never happen.

### PR Checklist

<details><summary>📱Order history (failed payment state simulated)</summary>
<p>

Simulator Screenshot - iPhone 15 Pro - 2024-09-24 at 13 48 20

image

</p>
</details> 

<details><summary>📱After clicking CTA</summary>
<p>

Simulator Screenshot - iPhone 15 Pro - 2024-09-24 at 13 48 16

image

</p>
</details> 

### To the reviewers 👀

- Enhancements to the order history screen when an order has a failed payment

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[EMI-2112]: https://artsyproduct.atlassian.net/browse/EMI-2112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/OrderHistory/OrderHistoryRow.tests.tsx
+++ b/src/app/Scenes/OrderHistory/OrderHistoryRow.tests.tsx
@@ -54,7 +54,7 @@ describe("OrderHistoryRow", () => {
   it("displays the artist name", () => {
     renderWithRelay({ CommerceOrder: () => mockOrder })
 
-    expect(screen.getByTestId("artist-names")).toHaveTextContent("Torbjørn Rødland")
+    expect(screen.getByText("Torbjørn Rødland")).toBeOnTheScreen()
   })
 
   it("displays the partner name", () => {
@@ -66,19 +66,19 @@ describe("OrderHistoryRow", () => {
   it("displays the order creation date", () => {
     renderWithRelay({ CommerceOrder: () => mockOrder })
 
-    expect(screen.queryByTestId("date")).toHaveTextContent("5/18/2021")
+    expect(screen.getByText("5/18/2021")).toBeOnTheScreen()
   })
 
   it("displays the price", () => {
     renderWithRelay({ CommerceOrder: () => mockOrder })
 
-    expect(screen.getByTestId("price")).toHaveTextContent("11,200")
+    expect(screen.getByText("11,200")).toBeOnTheScreen()
   })
 
   it("displays the display state", () => {
     renderWithRelay({ CommerceOrder: () => mockOrder })
 
-    expect(screen.getByTestId("order-status")).toHaveTextContent("pending")
+    expect(screen.getByText("pending")).toBeOnTheScreen()
   })
 
   describe("artwork image", () => {
--- a/src/app/Scenes/OrderHistory/OrderHistoryRow.tests.tsx
+++ b/src/app/Scenes/OrderHistory/OrderHistoryRow.tests.tsx
@@ -220,7 +220,7 @@ describe("OrderHistoryRow", () => {
 
       expect(navigate).toHaveBeenCalledWith("/orders/internal-id/payment/new", {
         modal: true,
-        passProps: { orderID: "internal-id", title: "Update Payment Method" },
+        passProps: { orderID: "internal-id", title: "Update Payment Details" },
       })
     })
     it("is not visible when the displayState is not PAYMENT_FAILED", () => {
--- a/src/app/Scenes/OrderHistory/OrderHistoryRow.tests.tsx
+++ b/src/app/Scenes/OrderHistory/OrderHistoryRow.tests.tsx
@@ -60,7 +60,7 @@ describe("OrderHistoryRow", () => {
   it("displays the partner name", () => {
     renderWithRelay({ CommerceOrder: () => mockOrder })
 
-    expect(screen.getByTestId("partner-name")).toHaveTextContent("Andrea Festa Fine Art")
+    expect(screen.getByText("Andrea Festa Fine Art")).toBeOnTheScreen()
   })
 
   it("displays the order creation date", () => {
--- a/src/app/Scenes/OrderHistory/OrderHistoryRow.tests.tsx
+++ b/src/app/Scenes/OrderHistory/OrderHistoryRow.tests.tsx
@@ -1,4 +1,4 @@
-import { fireEvent } from "@testing-library/react-native"
+import { fireEvent, screen } from "@testing-library/react-native"
 import { OrderHistoryRowTestsQuery } from "__generated__/OrderHistoryRowTestsQuery.graphql"
 import { navigate } from "app/system/navigation/navigate"
 import { extractText } from "app/utils/tests/extractText"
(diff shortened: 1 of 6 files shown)


src/app/utils/getOrderStatus.ts
```diff
@@ -13,11 +13,13 @@ export function getOrderStatus(displayState: CommerceOrderDisplayStateEnum) {
     case "PROCESSING":
       return "processing"
     case "PROCESSING_APPROVAL":
-      return "payment processing"
+      return "processing approval"
```
erikdstock, line 16: oops

src/app/utils/getOrderStatus.ts
```diff
@@ -13,11 +13,13 @@ export function getOrderStatus(displayState: CommerceOrderDisplayStateEnum) {
...
     case "IN_TRANSIT":
       return "in transit"
     case "CANCELED":
       return "canceled"
+    case "PAYMENT_FAILED":
+      return "payment failed"
     default:
       return "unknown"
```
erikdstock, line 24: 1: Need to check where else this function is used to make sure this text won't render in an unexpected place.
2: The 'unknown' text feels not ideal. I considered suppressing it on the OrderHistoryRow and just rendering no status in those cases.

src/app/utils/getOrderStatus.ts
erikdstock:
```suggestion
      return "payment processing"
```

ArtsyOpenSource:

- Cross-platform user-facing changes (Enhancements to the order history screen when an order has a failed payment)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 6c3460447419e3b0656fe26da7b71744ecada9c9
</p>

oxaudo (approved):
This looks good to me. Only had couple of neat picky comments here. Though def defer to our app specialists to weight in on best practices.

src/app/Scenes/OrderHistory/OrderHistoryRow.tests.tsx
```diff
@@ -52,44 +52,44 @@ describe("OrderHistoryRow", () => {
...
-    expect(tree.queryByTestId("price")?.children[0]).toBe("11,200")
+    expect(screen.getByTestId("price")).toHaveTextContent("11,200")
   })
 
   it("displays the display state", () => {
-    const tree = renderWithRelay({ CommerceOrder: () => mockOrder })
+    renderWithRelay({ CommerceOrder: () => mockOrder })
 
-    expect(tree.queryByTestId("order-status")?.children[0]).toBe("pending")
+    expect(screen.getByTestId("order-status")).toHaveTextContent("pending")
```
MrSltun, line 81: Thank you for refactoring this! 🙏

src/app/Scenes/OrderHistory/OrderHistoryRow.tests.tsx
```diff
@@ -52,44 +52,44 @@ describe("OrderHistoryRow", () => {
...
-    expect(tree.queryByTestId("artist-names")?.children[0]).toBe("Torbjørn Rødland")
+    expect(screen.getByTestId("artist-names")).toHaveTextContent("Torbjørn Rødland")
   })
 
   it("displays the partner name", () => {
-    const tree = renderWithRelay({ CommerceOrder: () => mockOrder })
+    renderWithRelay({ CommerceOrder: () => mockOrder })
 
-    expect(tree.queryByTestId("partner-name")?.children[0]).toBe("Andrea Festa Fine Art")
+    expect(screen.getByTestId("partner-name")).toHaveTextContent("Andrea Festa Fine Art")
```
MrSltun, line 63: **Suggestion (non-blocking)**: Maybe we could use a different way to test?
something like this

```suggestion
    expect(screen.getByText("Andrea Festa Fine Art")).toBeOnTheScreen()
```

src/app/Scenes/OrderHistory/OrderHistoryRow.tsx
```diff
@@ -18,8 +18,17 @@ export const OrderHistoryRow: React.FC<OrderHistoryRowProps> = ({ order }) => {
...
-  const orderIsInactive = orderStatus === "canceled" || orderStatus === "refunded"
+  const orderStatusColor = ["canceled", "payment failed"].includes(orderStatus)
+    ? "red100"
+    : "black60"
+
+  const hasFailedPayment = orderStatus === "payment failed"
+  const isActive =
+    !hasFailedPayment &&
+    orderStatus &&
+    !(["canceled", "refunded"] as string[]).includes(orderStatus)
```
MrSltun, line 31: **Question**: do we need the type casting `as string[]` here?

src/app/Scenes/OrderHistory/OrderHistoryRow.tsx
```diff
@@ -62,15 +71,17 @@ export const OrderHistoryRow: React.FC<OrderHistoryRowProps> = ({ order }) => {
...
-            <Text
-              textAlign="right"
-              variant="xs"
-              color={orderStatus === "canceled" ? "red100" : "black60"}
-              style={{ textTransform: "capitalize" }}
-              testID="order-status"
-            >
-              {orderStatus}
-            </Text>
+            {!!orderStatus && (
```
MrSltun, line 74: This will always return true, since `getOrderStatus` always returns a string that is not empty, maybe we could change the `getOrderStatus` return to give back `null` or an empty string?

src/app/utils/getOrderStatus.ts
```diff
@@ -1,6 +1,6 @@
 import { CommerceOrderDisplayStateEnum } from "__generated__/OrderDetailsHeader_info.graphql"
 
-export function getOrderStatus(displayState: CommerceOrderDisplayStateEnum) {
+export function getOrderStatus(displayState: CommerceOrderDisplayStateEnum): string {
```
MrSltun, line 3: **Question**: Do we need the type for the return type to be `string`? I believe Typescript infers a type based on the returning value of the function on its own

MrSltun (commented):
Looks good! 🌟 I just added a few comments

MrSltun:
@erikdstock can we also test these changes on Android and add screenshots?

MrSltun (dismissed):

src/app/Scenes/OrderHistory/OrderHistoryRow.tsx
```diff
@@ -104,6 +115,21 @@ export const OrderHistoryRow: React.FC<OrderHistoryRowProps> = ({ order }) => {
             {isViewOffer ? "View Offer" : "View Order"}
           </Button>
         )}
+        {!!hasFailedPayment && (
+          <Button
+            block
+            variant="fillDark"
+            onPress={() =>
+              navigate(`/orders/${order.internalID}/payment/new`, {
+                modal: true,
```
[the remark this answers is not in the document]
erikdstock: This is a style of system-level navigation where instead of adding a view to a stack you are popping up a single web view that acts independently.

src/app/utils/getOrderStatus.ts
erikdstock: Another relic of when I added types because it was `string | null` - when I changed it I also changed the types.

erikdstock:
> @erikdstock can we also test these changes on Android and add screenshots?

I did test them on android - will see if i can get some screenshots together after responding to feedback

src/app/utils/getOrderStatus.ts
```diff
@@ -13,12 +13,14 @@ export function getOrderStatus(displayState: CommerceOrderDisplayStateEnum) {
...
+      return "processing approval"
     case "IN_TRANSIT":
       return "in transit"
     case "CANCELED":
       return "canceled"
+    case "PAYMENT_FAILED":
+      return "payment failed"
     default:
-      return "unknown"
+      return null
```
MrSltun, line 24: **nitpick**: Can we change this to an empty string `""` to remove the type casting for the function and the OrderHistoryRow component? 👀

MrSltun: Thank you!! 🙏

erikdstock: Fine with the change but I actually think `null` is the correct type here if we don't want to display the message - it tells TS to make us handle a non-message value in a way that an empty string does not - the developer has to know to check before rendering an empty text field.  

The official type definition of `Array.includes()` doesn't permit types that _could_ match the array type:
image
The signature should be `Array<T>.includes<U extends T>(searchElement: U, fromIndex?: number): boolean` to include values that might match the array type since the method itself handles the actual logic for the boolean. This seems wrong to me, so casting in this one place would be the right way to work around it while keeping the other developer ergonomic benefits.

PR #10847, opened 2024-09-26
chore: better DX around for @cacheable

MounirDhahri:
### Description

This PR makes a few adjustments to follow up on `@cacheable` [PR](https://github.com/artsy/eigen/pull/10842#issuecomment-2373992261). The goal here is to make it clear to developers when `@cacheable` is not going to take effect

| Use case | Screenshot |
|--------|--------|
| Dev added `@cacheable` to a query where `force` is set to `true` | Simulator Screenshot - iPhone 16 Pro - 2024-09-26 at 09 49 25 |
| Dev added `@cacheable` to a query where `nocache` is present in the URL  | Simulator Screenshot - iPhone 16 Pro - 2024-09-26 at 09 48 05 |
| Dev added `@cacheable` to a query where a specialized argument is present | Simulator Screenshot - iPhone 16 Pro - 2024-09-26 at 09 52 38 | 

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- Improve DX in `@cacheable` - mounir

</details>

### DIFF
--- a/src/app/routes.ts
+++ b/src/app/routes.ts
@@ -164,13 +164,25 @@ export const getCurrentURL = () => {
 
   let { route: path } = currentModule
 
+  const queryParams = [] as { [key: string]: string }[]
+
   if (currentModuleProps) {
     Object.entries(currentModuleProps).map(([key, value]) => {
-      path = path.replace(`:${key}`, value as string)
+      if (path.includes(`:${key}`)) {
+        path = path.replace(`:${key}`, value as string)
+      } else {
+        queryParams.push({ [key]: value as string })
+      }
     })
   }
 
-  return encodeURI(`${webURL}${path}`)
+  const queryParamsString = queryParams.map((params) =>
+    Object.entries(params)
+      .map(([key, value]) => `${key}=${value}`)
+      .join("&")
+  )
+
+  return encodeURI(`${webURL}${path}?${queryParamsString}`)
 }
 
 export function getDomainMap(): Record<string, RouteMatcher[] | null> {
--- a/src/app/system/relay/helpers/cacheHeaderMiddlewareHelpers.ts
+++ b/src/app/system/relay/helpers/cacheHeaderMiddlewareHelpers.ts
@@ -19,7 +19,7 @@ export const hasNoCacheParamPresent = (url: string) => {
   return false
 }
 
-export const SKIP_CACHE_ARGUMENTS = ["includeArtworksByFollowedArtists", "collectionID"]
+export const SKIP_CACHE_ARGUMENTS = ["includeArtworksByFollowedArtists"]
 // Important - Add any new personalized argument checks to this list. That way, logged-in queries
 // _without_ this argument can still be `@cacheable`, and when queries include this argument,
 // those queries will not be cached.
--- a/src/app/system/relay/helpers/cacheHeaderMiddlewareHelpers.ts
+++ b/src/app/system/relay/helpers/cacheHeaderMiddlewareHelpers.ts
@@ -19,11 +19,11 @@ export const hasNoCacheParamPresent = (url: string) => {
   return false
 }
 
+export const SKIP_CACHE_ARGUMENTS = ["includeArtworksByFollowedArtists", "collectionID"]
 // Important - Add any new personalized argument checks to this list. That way, logged-in queries
 // _without_ this argument can still be `@cacheable`, and when queries include this argument,
 // those queries will not be cached.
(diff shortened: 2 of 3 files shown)


src/app/routes.ts
```diff
@@ -164,13 +164,25 @@ export const getCurrentURL = () => {
...
+      if (path.includes(`:${key}`)) {
+        path = path.replace(`:${key}`, value as string)
+      } else {
+        queryParams.push({ [key]: value as string })
+      }
     })
   }
 
-  return encodeURI(`${webURL}${path}`)
+  const queryParamsString = queryParams.map((params) =>
```
MounirDhahri, line 179: the query params where not present in the helper so I added them here

src/app/system/relay/middlewares/cacheHeaderMiddleware.ts
```diff
@@ -16,18 +17,34 @@ export const shouldSkipCDNCache = (req: GraphQLRequest) => {
   //     - a known personalized argument is present in the query (for example `include_artworks_by_followed_artists`)
   //     - nocache param is present in the url
 
-  if (req.cacheConfig?.force === true) {
```
MounirDhahri, line 19: I moved this down since it will be covered by the default case anyway

ArtsyOpenSource:

- Dev changes (Improve DX in `@cacheable` - mounir)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 42b7c68a92d3462ec9d17cb976e7b36191caccb4
</p>

src/app/system/relay/middlewares/cacheHeaderMiddleware.ts
```diff
@@ -16,18 +17,34 @@ export const shouldSkipCDNCache = (req: GraphQLRequest) => {
...
-    return true
-  }
-
   if (isRequestCacheable(req)) {
+    if (req.cacheConfig?.force === true) {
+      if (__DEV__) {
+        console.warn(
```
araujobarret, line 23: can we use info here? These warnings in the simulators/emulators are pretty annoying

src/app/system/relay/helpers/cacheHeaderMiddlewareHelpers.ts
```diff
@@ -19,11 +19,11 @@ export const hasNoCacheParamPresent = (url: string) => {
   return false
 }
 
+export const SKIP_CACHE_ARGUMENTS = ["includeArtworksByFollowedArtists", "collectionID"]
```
araujobarret, line 22: isn't better to add some comments around it to give engineers direction and/or maybe add it to some docs about cache in Eigen?

araujobarret (dismissed):
Looks good, left small comments only

src/app/system/relay/helpers/cacheHeaderMiddlewareHelpers.ts
MounirDhahri: that's a valid point - we need something like [this](https://github.com/artsy/force/blob/cbc411bc93466aac3e43a9126b0b98fd6af12174/docs/caching2.md) but for Eigen. I will draft something

src/app/system/relay/middlewares/cacheHeaderMiddleware.ts
MounirDhahri: I tried both actually but they both seems equally ugly 🥲 except that the warn shows the count better

https://github.com/user-attachments/assets/b24bb177-d32e-41a7-9bdf-90eb76ed7d54

araujobarret: 😢  then I'd vote only to console.log, these bubbles are bad for DX because they block the screen and all the time we have to close them

MounirDhahri: I agree that console.warns can be annoying, but that's only if you can't dismiss them or don't know why they're occuring. Here, the action item should be clear, for example you need to remove `@cacheable`. I do though think it's bad DX to declare your query as cacheable, expecting it to be cacheable, but it's not cacheable because something changed in the arguments and you aren't aware of it

araujobarret: yeah, after a 2nd thought I'm fine with it, it should really pester someone in this case

damassi (approved):

PR #10851, opened 2024-09-26
feat: render new Homeview for artsy people

MounirDhahri:
This PR resolves [ONYX-1330] 

### Description

This PR exposes the new homeview to artsy people.

**For easier review, I suggest you hide white spaces**
 

### PR Checklist

- [x] I have tested my changes on **iOS** and **Android**.
- [x] I hid my changes behind a **[feature flag]**, or they don't need one.
- [x] I have included **screenshots** or **videos**, or I have not changed the UI.
- [x] I have added **tests**, or my changes don't require any.
- [x] I added an **[app state migration]**, or my changes do not require one.
- [x] I have documented any **follow-up work** that this PR will require, or it does not require any.
- [x] I have added a **changelog entry** below, or my changes do not require one.

### To the reviewers 👀

- [ ] I would like **at least one** of the reviewers to **run** this PR on the simulator or device.

<details><summary>Changelog updates</summary>

#### Cross-platform user-facing changes

-

#### iOS user-facing changes

-

#### Android user-facing changes

-

#### Dev changes

- render new Homeview for artsy people - mounir

</details>

Need help with something? Have a look at our [docs], or get in touch with us.

[app state migration]: ../blob/main/docs/adding_state_migrations.md
[feature flag]: ../blob/main/docs/developing_a_feature.md
[docs]: ../blob/main/docs/README.md

[ONYX-1330]: https://artsyproduct.atlassian.net/browse/ONYX-1330?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #10880, opened 2024-10-02
fix: fetch all home view sections at once to avoid placeholders while scrolling

nickskalkin:
### Description

We received feedback that some sections do not load in time. This issue is most noticeable while scrolling (even when scrolling slowly). The reason for this is that we fetch new sections as we scroll, and it takes time to retrieve their data. One idea that was discussed during the post-QA session with the team was to fetch more sections at once (currently, we fetch five).

The Home view screen is rendered in two steps. In the first step, we fetch a small amount of data – only section titles and __typenames. This information is needed to begin rendering the sections.

```graphql
query {
  homeView {
    sectionsConnection(first: 20) {
      edges {
        node {
          __typename
          internalID
          component {
            title
            type
          }
        }
      }
    }
  }
}
```

After the sections are fetched from the backend, each section renders a query renderer component, which requests its own data.

The first step is quite fast. I haven’t noticed a significant difference between fetching 5 sections and 20. However, it is important to note that this could change – currently, sections are hardcoded in Metaphysics, but in the future, we may want to resolve them dynamically, which could make the fetch operation slower.

Small before/after demo – notice that there are no visible placeholders after the initial render in the ‘after’ video.

| Before | After |
|---|---|
| <video src="https://github.com/user-attachments/assets/953bcda0-a283-4e39-8db3-944146da9e20" /> 
 | <video src="https://github.com/user-attachments/assets/52547e88-7476-4dcc-ab87-60935d08a246" /> |

### PR Checklist

### To the reviewers 👀

### Changelog updates

#nochangelog

</details>

### DIFF
--- a/src/app/Scenes/HomeView/HomeView.tsx
+++ b/src/app/Scenes/HomeView/HomeView.tsx
@@ -24,7 +24,7 @@ import { Suspense, useCallback, useEffect, useState } from "react"
 import { FlatList, RefreshControl } from "react-native"
 import { fetchQuery, graphql, useLazyLoadQuery, usePaginationFragment } from "react-relay"
 
-export const NUMBER_OF_SECTIONS_TO_LOAD = 5
+export const NUMBER_OF_SECTIONS_TO_LOAD = 30
 // Hard coding the value here because 30px is not a valid value for the spacing unit
 // and we need it to be consistent with 60px spacing between sections
 export const HOME_VIEW_SECTIONS_SEPARATOR_HEIGHT = "30px"


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  ✅ **No changelog changes**
  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against da2eae66d3cdcf731b5ab85922eba6e612093229
</p>

damassi:
This somewhat intersects the work that @MounirDhahri was experimenting with which was prefetching the list during the boot phase of the app. I know he's OOO, but was that played with any more? 

Ideally, the UX is such to where by the time the home screen disappears there's no placeholders above the fold; that work was fetched during the boot screen, and the app lands stably.

dblandin:
> I know he's OOO, but was that played with any more?

I believe that was implemented in https://github.com/artsy/eigen/pull/10760 and should be active via beta builds.

As @anandaroop [recently pointed out](https://github.com/artsy/eigen/pull/10851#discussion_r1783566676), the condition backing that prefetch unfortunately drifted from the main condition that enables the new Home View.

That means that in the most recently submitted app store builds, internal Artsy colleagues will see the new Home View by default but _will not_ benefit from this prefetching optimization.

src/app/Scenes/HomeView/HomeView.tsx
```diff
@@ -24,7 +24,7 @@ import { Suspense, useCallback, useEffect, useState } from "react"
 import { FlatList, RefreshControl } from "react-native"
 import { fetchQuery, graphql, useLazyLoadQuery, usePaginationFragment } from "react-relay"
 
-export const NUMBER_OF_SECTIONS_TO_LOAD = 5
+export const NUMBER_OF_SECTIONS_TO_LOAD = 30
```
dblandin, line 27: **question**: Instead of bumping this number, is there any way to more eagerly fetch additional sections as you scroll down the screen? For example, instead of waiting until you reach the end of the currently fetched list, can we trigger the next pagination request when we scroll down to the half way point?

dblandin: Although we should bump this number either way as 5 seems like too small of a chunk. Maybe we can see how 10 or 15 feels in addition to more eagerly requesting that additional page of results?

anandaroop: I thought we had that kind of lazy loading, remembering [this conversation](https://github.com/artsy/eigen/pull/10763#issuecomment-2349364612) from a couple of weeks ago 🤔

damassi: Its a little different, in the sense that when you use `usePaginationFragment` you can keep the 5 there, but then fetch the next 5 (ie, next page). AFAIK, Mounirs implementation will only execute the lazy loaded content query once the content intersects on the visible page, but that doesn't solve the situation around prefetching the sections expected to render themselves, which could get slow later as things become more dynamic. 

It might be over engineering for this pass (due to static list in MP) but def something to keep in mind for later.

nickskalkin: thanks all for feedback! In this commit https://github.com/artsy/eigen/pull/10880/commits/044cc5bf838706778cf74007a8b59baa6e84d637 I:

- request 10 sections at a time
- request second (and following) page earlier by tweaking [onEndReachedThreshold](https://reactnative.dev/docs/virtualizedlist#onendreachedthreshold) option - now second page is requested when user is ~ on Auction Lots for You. Without this change second page is requested ~ a viewport later. When scrolling at moderate speed - users shouldn't see second page's placeholders. If they scroll fast - they will see placeholders.

wdyt?

@dblandin @damassi @anandaroop

anandaroop (approved):
Latest version feels great to me in the simulator (and I assume will feel even better on device).

I say we merge it and try it out 👍🏽

src/app/Scenes/HomeView/HomeView.tsx
damassi: niceee

MounirDhahri: Agree `onEndReachedThreshold ` is a nice solution here 👏

PR #10895, opened 2024-10-06
refactor(auth2): More smoothing out of auth screens 

damassi:
### Description

This cleans things up a bit more:
- Avoid the use of `useFormik` hooks [per the docs](https://formik.org/docs/api/useFormik#use-cases-for-useformik), and instead rely on standard `<Formik>`, which has performance benefits 
- On submit, collapse the keyboard, wait, then show keyboard again when the next screen appears. This is pretty standard native UX for mult-step login flows, and the only way we can get around the UI jank involving the keyboard QuickType bar (suggestions), which **jumps** the modal, vs animating, as the quick type bar appears / disappears instantly. 

Also stubbed out test files for a future PR. 

@iskounen - this one is pretty uncontroversial so gonna merge to get a beta out before monday but happy to follow up on any comments 👍 

cc @artsy/diamond-devs

### DIFF
--- a/src/app/Scenes/Onboarding/Auth2/AuthScenes.tsx
+++ b/src/app/Scenes/Onboarding/Auth2/AuthScenes.tsx
@@ -5,11 +5,12 @@ import { LoginPasswordStep } from "app/Scenes/Onboarding/Auth2/scenes/LoginPassw
 import { SignUpNameStep } from "app/Scenes/Onboarding/Auth2/scenes/SignUpNameStep"
 import { SignUpPasswordStep } from "app/Scenes/Onboarding/Auth2/scenes/SignUpPasswordStep"
 import { WelcomeStep } from "app/Scenes/Onboarding/Auth2/scenes/WelcomeStep"
+import { ScrollView } from "react-native-gesture-handler"
 
 export const AuthScenes: React.FC = () => {
   return (
     <>
-      <>
+      <ScrollView keyboardShouldPersistTaps="always" scrollEnabled={false}>
         <AuthScreen name="WelcomeStep">
           <WelcomeStep />
         </AuthScreen>
@@ -33,7 +34,7 @@ export const AuthScenes: React.FC = () => {
         <AuthScreen name="SignUpPasswordStep">
           <SignUpPasswordStep />
         </AuthScreen>
-      </>
+      </ScrollView>
     </>
   )
 }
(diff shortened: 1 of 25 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 3ed3bc67357e071b2afba1625112520987a06268
</p>

src/app/Scenes/Onboarding/Auth2/AuthScenes.tsx
```diff
@@ -10,7 +10,7 @@ import { ScrollView } from "react-native-gesture-handler"
 export const AuthScenes: React.FC = () => {
   return (
     <>
-      <ScrollView keyboardShouldPersistTaps="always">
+      <ScrollView keyboardShouldPersistTaps="always" scrollEnabled={false}>
```
damassi, line 13: We want keyboard taps to go through and not get weird, but we also don't want users to be able to scroll around -- ui should be stable

src/app/Scenes/Onboarding/Auth2/scenes/SignUpNameStep.tsx
```diff
@@ -98,11 +101,14 @@ const SignUpNameStepForm: React.FC = () => {
         autoCorrect={false}
         blurOnSubmit={false}
         error={errors.name}
+        keyboardType="name-phone-pad"
```
damassi, line 104: Hides quick type bar

src/app/Scenes/Onboarding/Auth2/scenes/SignUpNameStep.tsx
```diff
@@ -103,6 +106,8 @@ const SignUpNameStepForm: React.FC = () => {
         placeholderTextColor={color("black30")}
         ref={nameRef}
         returnKeyType="done"
+        spellCheck={false}
+        textContentType="none"
```
damassi, line 110: Hide quick type bar -- don't need suggestions

src/app/Scenes/Onboarding/Auth2/scenes/WelcomeStep.tsx
```diff
@@ -118,34 +121,34 @@ const WelcomeStepForm: React.FC = () => {
...
         error={errors.email}
         keyboardType="email-address"
-        onSubmitEditing={handleSubmit}
         placeholderTextColor={color("black30")}
         ref={emailRef}
-        returnKeyType="next"
         spellCheck={false}
-        // We need to to set textContentType to username (instead of emailAddress) here
-        // enable autofill of login details from the device keychain.
-        textContentType="username"
```
damassi, line 131: We can discuss this one, but for multi-step forms like this, the QuickType bar + auto-suggest isn't valuable, because we can't auto-sign-in. It _is_ useful for the password step though. So remove it from here and fix some very bad UI jank (quicktype bar appearing / disappearing / being generally wonky on ios)

src/app/Scenes/Onboarding/Auth2/utils/waitForSubmit.tsx
```diff
@@ -0,0 +1,6 @@
+/**
+ * Slight delay to smooth out keyboard transitions between submission steps
+ */
+export const waitForSubmit = async (delay = 1500) => {
+  return new Promise((resolve: any) => setTimeout(resolve, delay))
+}
```
damassi, line 6: This is unintuitive, but fake wait times are the key to making native apps feel smooth. They're all over folio, and that's why it feels so polished.

PR #10896, opened 2024-10-07
fix(auth2): Remove KeyboardAvoidingView to get around iOS 17 bug 

damassi:
### Description

This PR makes me 😢. 

Basically, iOS 17 introduced a bug that is only seen on device around KeyboardAvoidingView and the QuickType bar (password autosuggest, etc) where it will flash and lead to UI positional instability. A way to reproduce it is by simply hitting backspace on the keyboard in an empty password field. With the keyboard avoiding view this would lead to a jump in the modal. It also manifested when transitioning between screens and the quicktype bar would appear and disappear depending on the input type. Its simply not passable. Typing fast would also trigger it! 

See this github issue: https://github.com/facebook/react-native/issues/39411

It impacts all platforms, including fully native apps. 

Since we can't remove the password auto-fill functionality by hiding the quicktype bar, we need to compromise and wait for an iOS version update 😭. 

Fix: 

Remove the keyboard avoiding view and all form of dynamic positioning of the modal. Now, when the sign in/up flow begins we need to animate it to a fixed position at the top of the screen. When transitioning between inputs, nothing stirs. And yet, also, nothing is exactly positioned and centered anymore either, which was nice. All in all a minor compromise given the bug. 

We can revisit this with later iOS updates. 

Before (kinda captures it -- only reproducible on device, but you can see how the quicktype bar appears and hides changing screens -- how apply this to simply entering or backspacing text) 

https://github.com/user-attachments/assets/ba3e1c7e-9b8f-4c34-afd4-9cccb31c1f4a

After: 

https://github.com/user-attachments/assets/33062b77-070b-4e8a-93df-707b9d14f85a

### DIFF
--- a/src/app/Scenes/Onboarding/Auth2/components/AuthModal.tsx
+++ b/src/app/Scenes/Onboarding/Auth2/components/AuthModal.tsx
@@ -1,8 +1,9 @@
 import { Box, Flex, useTheme } from "@artsy/palette-mobile"
 import { AuthContext } from "app/Scenes/Onboarding/Auth2/AuthContext"
 import { MotiView } from "moti"
-import { Dimensions, KeyboardAvoidingView, Platform } from "react-native"
+import { Dimensions } from "react-native"
 import { Easing } from "react-native-reanimated"
+import { useSafeAreaInsets } from "react-native-safe-area-context"
 
 const HEIGHT = {
   LoginWelcomeStep: 320,
@@ -20,6 +21,7 @@ export const AuthModal: React.FC = ({ children }) => {
   const { isModalExpanded, isMounted, currentScreen } = AuthContext.useStoreState((state) => state)
 
   const { space } = useTheme()
+  const insets = useSafeAreaInsets()
 
   const screenHeight = Dimensions.get("window").height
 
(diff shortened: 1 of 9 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against edd9de8380cab4532107c50c7e4dc356a7f4d55c
</p>

src/app/Scenes/Onboarding/Auth2/scenes/LoginWelcomeStep.tsx
```diff
@@ -135,15 +133,25 @@ const LoginWelcomeStepForm: React.FC = () => {
         onSubmitEditing={handleSubmit}
       />
 
-      <Flex display={isModalExpanded ? "flex" : "none"}>
+      <MotiView
+        from={{ opacity: isModalExpanded ? 0 : 1 }}
+        animate={{ opacity: isModalExpanded ? 1 : 0 }}
+        transition={{ type: "timing", duration: 400, easing: Easing.linear }}
+        style={{ display: isModalExpanded ? "flex" : "none" }}
+      >
```
damassi, line 141: Just adds a little smooth transition between the social login options and the email only input step

src/app/Scenes/Onboarding/Auth2/components/AuthModal.tsx
```diff
@@ -34,44 +36,40 @@ export const AuthModal: React.FC = ({ children }) => {
...
+        }}
+        transition={{
+          type: "timing",
+          duration: isModalExpanded ? 800 : 600,
+          delay: isMounted ? 0 : 500,
+          easing: Easing.out(Easing.exp),
+        }}
+        style={{
+          width: "100%",
+          backgroundColor: "white",
```
araujobarret, line 62: nit: backgroundColor: color("white100"),

src/app/Scenes/Onboarding/Auth2/components/AuthModal.tsx
```diff
@@ -34,44 +36,40 @@ export const AuthModal: React.FC = ({ children }) => {
   const translateY = (() => {
```
araujobarret, line 36: shouldn't we memoize it?

src/app/Scenes/Onboarding/Auth2/scenes/LoginWelcomeStep.tsx
```diff
@@ -119,6 +116,7 @@ const LoginWelcomeStepForm: React.FC = () => {
         autoCapitalize="none"
         autoComplete="email"
         autoCorrect={false}
+        blurOnSubmit={false}
```
araujobarret, line 119: I see most of the places you removed the Keyboard.dismiss and added this prop as false, can you explain what it does now? (just curious)

araujobarret (approved):

src/app/Scenes/Onboarding/Auth2/scenes/LoginWelcomeStep.tsx
damassi: Keeps the keyboard up, so that as we transition from screen to screen things aren't jumping around. Before it worked fine because we could use the keyboard avoiding view, and then use flexbox to vertically center in the screen. User would submit, keyboard would slide down, things would fetch, screen would transition. But with the quicktype bar, this broke the UX in a number of different ways. So now just keep things up and persistent from screen to screen.

damassi: It doesn't feel as nice any more having everything at that fixed position, but at least it doesn't feel broken!

src/app/Scenes/Onboarding/Auth2/components/AuthModal.tsx
damassi: ditto!

src/app/Scenes/Onboarding/Auth2/components/AuthModal.tsx
damassi: Will update in next pr, gonna send out a beta for this one.

src/app/Scenes/Onboarding/Auth2/components/AuthModal.tsx
damassi: much better, using `useMemo` i was like... on geez. Updated in here: https://github.com/artsy/eigen/pull/10898

src/app/Scenes/Onboarding/Auth2/scenes/LoginWelcomeStep.tsx
araujobarret: it's a good compromise!

PR #10879, opened 2024-10-01
feat(EMI-2095): Add Swipeable component

MrSltun:
This PR resolves [EMI-2095] 

> [!NOTE]
> This PR is related to artsy/palette-mobile#263, we decided for the time being to move the work to Eigen.
> More info in the [Slack thread](https://artsy.slack.com/archives/C02JHHHKP5K/p1727790098522529?thread_ts=1727779387.115219&cid=C02JHHHKP5K) 🔒

### Description
This PR adds a swipeable component to Eigen

### Demos (from Palette-Mobile)
| Android | iOS |
|---|---|
| <video src="https://github.com/user-attachments/assets/73a223ff-5586-4969-b837-b8a7211e48a9" /> | <video src="https://github.com/user-attachments/assets/a941add6-d28a-4434-ab8b-cf34994a7560" /> |

### PR Checklist

- [X] I have tested my changes on **iOS** and **Android**.
- [x] I hid my changes behind a **[feature flag]**, or they don't need one.
- [x] I have included **screenshots** or **videos**, or I have not changed the UI.
- [x] I have added **tests**, or my changes don't require any.
- [x] I added an **[app state migration]**, or my changes do not require one.
- [x] I have documented any **follow-up work** that this PR will require, or it does not require any.
- [x] I have added a **changelog entry** below, or my changes do not require one.

### To the reviewers 👀

- [ ] I would like **at least one** of the reviewers to **run** this PR on the simulator or device.

<details><summary>Changelog updates</summary>

#### Cross-platform user-facing changes

-

#### iOS user-facing changes

-

#### Android user-facing changes

-

#### Dev changes

- Add swipeable component to Eigen - mrsltun

</details>

Need help with something? Have a look at our [docs], or get in touch with us.

[app state migration]: ../blob/main/docs/adding_state_migrations.md
[feature flag]: ../blob/main/docs/developing_a_feature.md
[docs]: ../blob/main/docs/README.md

[EMI-2095]: https://artsyproduct.atlassian.net/browse/EMI-2095?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #10912, opened 2024-10-09
feat: Add swipe-to-interact functionality to `Swipeable` component

olerichter00:
This PR resolves [EMI-2127] 

* Follow-Up PR for https://github.com/artsy/eigen/pull/10879

### Description

This PR adds swipe-to-interact functionality (swiping until the end without tapping on the action button) to the `Swipeable` component. We might fine-tune the interaction later after testing it on a real device.

https://github.com/user-attachments/assets/b51da737-0184-4f25-a1f4-865886c020b5

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- Add swipe-to-interact functionality to `Swipeable` component (not used yet) - ole

</details>

[EMI-2127]: https://artsyproduct.atlassian.net/browse/EMI-2127?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Components/Swipeable/Swipeable.tsx
+++ b/src/app/Components/Swipeable/Swipeable.tsx
@@ -1,5 +1,5 @@
 import { Color, Flex, Touchable } from "@artsy/palette-mobile"
-import { forwardRef, useEffect, useState } from "react"
+import { forwardRef, useRef } from "react"
 import ReanimatedSwipeable, {
   SwipeableMethods,
   SwipeableRef,
@@ -32,13 +32,12 @@ export const Swipeable = forwardRef((props: SwipeableProps, swipeableRef: Swipea
 
   const width = useSharedValue(0)
 
-  const [hasSwiped, setHasSwiped] = useState(false)
+  const hasSwiped = useRef(false)
 
-  useEffect(() => {
-    if (!hasSwiped) return
-
+  const handleSwipeToInteract = () => {
+    hasSwiped.current = true
     actionOnSwipe?.()
-  }, [hasSwiped])
+  }
 
   const RightActions = (
     _progress: SharedValue<number>,
@@ -48,12 +47,12 @@ export const Swipeable = forwardRef((props: SwipeableProps, swipeableRef: Swipea
     useAnimatedStyle(() => {
       "worklet"
       // Don't do anything if the action is disabled, if the user has already swiped, or if the width is not yet set (on first render)
-      if (!actionOnSwipe || hasSwiped || !width.value) return {}
+      if (!actionOnSwipe || hasSwiped.current || !width.value) return {}
 
       if (width.value + dragX.value * FRICTION <= SWIPE_TO_INTERACT_THRESHOLD) {
         // TODO: Add haptic feedback
 
-        runOnJS(setHasSwiped)(true)
+        runOnJS(handleSwipeToInteract)()
 
         return {}
       }
--- a/src/app/Components/Swipeable/Swipeable.tsx
+++ b/src/app/Components/Swipeable/Swipeable.tsx
@@ -1,16 +1,18 @@
 import { Color, Flex, Touchable } from "@artsy/palette-mobile"
-import { forwardRef, useCallback } from "react"
+import { forwardRef, useEffect, useState } from "react"
 import ReanimatedSwipeable, {
   SwipeableMethods,
   SwipeableRef,
 } from "react-native-gesture-handler/ReanimatedSwipeable"
-import { SharedValue } from "react-native-reanimated"
+import { runOnJS, SharedValue, useAnimatedStyle, useSharedValue } from "react-native-reanimated"
 
 const FRICTION = 1.2
+const SWIPE_TO_INTERACT_THRESHOLD = 80
 
 export interface SwipeableProps {
   children: React.ReactNode
-  actionOnPress: () => void
+  actionOnPress?: () => void
+  actionOnSwipe?: () => void
   actionComponent: React.ReactNode
   actionBackground?: Color
   swipeableProps?: SwipeableProps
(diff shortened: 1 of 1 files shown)


src/app/Components/Swipeable/Swipeable.tsx
```diff
@@ -1,16 +1,18 @@
...
+import { forwardRef, useEffect, useState } from "react"
 import ReanimatedSwipeable, {
   SwipeableMethods,
   SwipeableRef,
 } from "react-native-gesture-handler/ReanimatedSwipeable"
-import { SharedValue } from "react-native-reanimated"
+import { runOnJS, SharedValue, useAnimatedStyle, useSharedValue } from "react-native-reanimated"
 
 const FRICTION = 1.2
+const SWIPE_TO_INTERACT_THRESHOLD = 80
```
olerichter00, line 10: This is the threshold up to which the distance from the screen edge the user has to swipe to trigger the action.

src/app/Components/Swipeable/Swipeable.tsx
```diff
@@ -21,17 +23,44 @@ export const Swipeable = forwardRef((props: SwipeableProps, swipeableRef: Swipea
...
-  const RightActions: (
-    progress: SharedValue<number>,
+  const width = useSharedValue(0)
+
+  const [hasSwiped, setHasSwiped] = useState(false)
+
+  useEffect(() => {
+    if (!hasSwiped) return
+
+    actionOnSwipe?.()
```
olerichter00, line 40: Calling the action in an effect to prevent the action from being called more than once.

ArtsyOpenSource:

- Dev changes (Add swipe-to-interact functionality to `Swipeable` component (not used yet) - ole)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against c9a7e40d8f272b828bcfaa97caa68619b47f3bfe
</p>

src/app/Components/Swipeable/Swipeable.tsx
```diff
@@ -48,18 +77,24 @@ export const Swipeable = forwardRef((props: SwipeableProps, swipeableRef: Swipea
...
-      testID="swipeable-component"
-      ref={swipeableRef}
-      enabled={enabled}
-      renderRightActions={RightActions}
-      friction={FRICTION}
-      {...swipeableProps}
+    <Flex
+      onLayout={(event) => {
+        width.value = event.nativeEvent.layout.width
+      }}
```
olerichter00, line 86: I'm using the component's width to determine when the `dragX` of the drag interaction has reached the end of the component.

araujobarret (commented):
Not sure if I follow the need for this extra function, I think the Reanimated element already provides [onSwipeableOpen](https://docs.swmansion.com/react-native-gesture-handler/docs/components/reanimated_swipeable/#onswipeableopen), where we can control kinda the same of `actionOnSwipe` without any extra code. Then it's up to the caller to control it and the swipeable element stays simpler and more performant.

olerichter00:
> Not sure if I follow the need for this extra function, I think the Reanimated element already provides [onSwipeableOpen](https://docs.swmansion.com/react-native-gesture-handler/docs/components/reanimated_swipeable/#onswipeableopen), where we can control kinda the same of actionOnSwipe without any extra code. Then it's up to the caller to control it and the swipeable element stays simpler and more performant.

Maybe my PR description wasn't very clear. This new functionality is about swiping until the end without tapping on the action button (The [Jira ticket](https://artsyproduct.atlassian.net/browse/EMI-2127) has a video and better description of the interaction). `onswipeableopen` is only called once when the user swipes and the buttons are displayed.

src/app/Components/Swipeable/Swipeable.tsx
```diff
@@ -21,17 +23,44 @@ export const Swipeable = forwardRef((props: SwipeableProps, swipeableRef: Swipea
...
     actionBackground = "red100",
     enabled = true,
     swipeableProps,
   } = props
 
-  const RightActions: (
-    progress: SharedValue<number>,
+  const width = useSharedValue(0)
+
+  const [hasSwiped, setHasSwiped] = useState(false)
```
araujobarret, line 35: this can be replaced by useRef from what I see in the code

araujobarret (commented):
Ah it makes sense, not sure about the UX since there's nothing on the DX, looks legit just one thing we might consider to change.

olerichter00:
> Ah it makes sense, not sure about the UX since there's nothing on the DS, looks legit just one thing we might consider to change.

You are right, the UX might change. The designs for the new home view's Tasks/Notifications section are still wip...

src/app/Components/Swipeable/Swipeable.tsx
olerichter00: I've tried to make it work but it doesn't.... Is this what you would suggest?

```
  const hasSwiped = useRef(false)

  const handleSwipeToInteract = () => {
    hasSwiped.current = true
  }

  useEffect(() => {
    if (!hasSwiped.current) return

    actionOnSwipe?.()
  }, [hasSwiped.current])
  ```

Comment (araujobarret) [reply src/app/Components/Swipeable/Swipeable.tsx:35]: I think that won't work because your useEffect comes from the state change reaction when we call `runOnJs. What could work is:
```jsx
// no useEffect anymore
const handleSwipeToInteract = () => {
  hasSwiped.current = true
  actionOnSwipe?.()
}

// line 51
if (!actionOnSwipe || hasSwiped.current || !width.value) return {}

// line 56
runOnJs(handleSwipeToInteract)()
```

olerichter00: Thanks a lot, that's very clean! Of course, with the ref, there is no need for the effect anymore. I've updated the code 👍

src/app/Components/Swipeable/Swipeable.tsx
```diff
@@ -21,17 +23,43 @@ export const Swipeable = forwardRef((props: SwipeableProps, swipeableRef: Swipea
...
   } = props
 
-  const RightActions: (
-    progress: SharedValue<number>,
+  const width = useSharedValue(0)
+
+  const hasSwiped = useRef(false)
+
+  const handleSwipeToInteract = () => {
+    hasSwiped.current = true
```
olerichter00, line 38: To prevent calling the action more than once.

araujobarret (approved):

damassi:
@olerichter00 - Is there any way to make it so that element doesn't just vanish? Looks a bit clunky just popping out like that. 

In Moti, there's an `<AnimatePresence>` component that can be used to check if something is appearing and disappearing. Ie, if 

```
<AnimatePresence>
  {foo && 
    <MotiView exit={{ ...someExitAnimation }} exitTransition={{ ... }}>
```

We should lean on that, even if its a simple opacity. But better: animate the height of the container to `0`.

olerichter00:
> @olerichter00 - Is there any way to make it so that element doesn't just vanish? Looks a bit clunky just popping out like that.

Yes, there are a couple of options like [LayoutAnimation](https://reactnative.dev/docs/layoutanimation), [Disappearable](https://github.com/artsy/eigen/blob/edbcd3e2085f32acbf6e69b144c3a827f79fd7a8/src/app/Components/Disappearable.tsx#L18), [HeighTransition](https://github.com/artsy/eigen/blob/edbcd3e2085f32acbf6e69b144c3a827f79fd7a8/src/app/utils/animations/HeightTransition.tsx#L20), or [Animated.Flastlist](https://reactnative.dev/docs/animated#animatable-components),that can be used to animate the item disappearing. 

Because it might depend on the context, I thought it would be good to not built the animation into this component.

PR #10914, opened 2024-10-09
feat(DIA-882): add explore by categories to the home view

araujobarret:
This PR resolves [DIA-882] 

### Description

Add the Explore By Categories to the home view

iPhone

https://github.com/user-attachments/assets/428264a2-4b79-4edd-82e3-106bd8b98430

iPad

https://github.com/user-attachments/assets/568609bd-24f4-4289-af2e-d8eab964c37b

### PR Checklist

### To the reviewers 👀

- feat: add explore by section in the homeview - araujobarret

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[DIA-882]: https://artsyproduct.atlassian.net/browse/DIA-882?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/Collections/Collections.tsx
+++ b/src/app/Scenes/Collections/Collections.tsx
@@ -1,0 +1,27 @@
+import { Flex, Screen, Text } from "@artsy/palette-mobile"
+import { RouteProp, useRoute } from "@react-navigation/native"
+import { goBack } from "app/system/navigation/navigate"
+import { FC } from "react"
+
+type CollectionsNavigationRoutes = {
+  collections: { props: { category: string } }
+}
+
+export const Collections: FC = () => {
+  const { params } = useRoute<RouteProp<CollectionsNavigationRoutes, "collections">>()
+
+  const category = params?.props.category ?? ""
+
+  return (
+    <Screen>
+      <Screen.Header onBack={goBack} title="Collections" animated />
+      <Screen.Body fullwidth px={2}>
+        <Screen.ScrollView>
+          <Flex>
+            <Text>{category}</Text>
+          </Flex>
+        </Screen.ScrollView>
+      </Screen.Body>
+    </Screen>
+  )
+}
--- a/src/app/Scenes/CollectionsByCategory/CollectionsByCategory.tsx
+++ b/src/app/Scenes/CollectionsByCategory/CollectionsByCategory.tsx
@@ -10,8 +10,6 @@ type CollectionsByCategoriesNavigationRoutes = {
 export const CollectionsByCategory: FC = () => {
   const { params } = useRoute<RouteProp<CollectionsByCategoriesNavigationRoutes, "collections">>()
 
-  console.log(params)
-
   const category = params?.props.category ?? ""
 
   return (
--- a/src/app/Scenes/CollectionsByCategory/CollectionsByCategory.tsx
+++ b/src/app/Scenes/CollectionsByCategory/CollectionsByCategory.tsx
@@ -3,12 +3,14 @@ import { RouteProp, useRoute } from "@react-navigation/native"
 import { goBack } from "app/system/navigation/navigate"
 import { FC } from "react"
 
-type CollectionsNavigationRoutes = {
+type CollectionsByCategoriesNavigationRoutes = {
   collections: { props: { category: string } }
 }
 
-export const Collections: FC = () => {
-  const { params } = useRoute<RouteProp<CollectionsNavigationRoutes, "collections">>()
+export const CollectionsByCategory: FC = () => {
+  const { params } = useRoute<RouteProp<CollectionsByCategoriesNavigationRoutes, "collections">>()
+
+  console.log(params)
 
   const category = params?.props.category ?? ""
 
--- a/src/app/Scenes/HomeView/Sections/HomeViewSectionExploreBy.tests.tsx
+++ b/src/app/Scenes/HomeView/Sections/HomeViewSectionExploreBy.tests.tsx
@@ -1,0 +1,32 @@
+import { screen } from "@testing-library/react-native"
+import { HomeViewSectionExploreByTestsQuery } from "__generated__/HomeViewSectionExploreByTestsQuery.graphql"
+import { HomeViewSectionExploreBy } from "app/Scenes/HomeView/Sections/HomeViewSectionExploreBy"
+import { setupTestWrapper } from "app/utils/tests/setupTestWrapper"
+import { graphql } from "react-relay"
+
+describe("HomeViewSectionExploreBy", () => {
+  const { renderWithRelay } = setupTestWrapper<HomeViewSectionExploreByTestsQuery>({
+    Component: (props) => <HomeViewSectionExploreBy section={props.homeView.section} />,
+    query: graphql`
+      query HomeViewSectionExploreByTestsQuery @relay_test_operation {
+        homeView @required(action: NONE) {
+          section(id: "home-view-section-explore-by-category") @required(action: NONE) {
+            ...HomeViewSectionExploreBy_section
+          }
+        }
+      }
+    `,
+  })
+
+  it("renders", () => {
+    renderWithRelay({
+      HomeViewSectionCards: () => ({
+        component: {
+          title: "Explore by category",
+        },
+      }),
+    })
+
+    expect(screen.getByText("Explore by category")).toBeOnTheScreen()
+  })
+})
--- a/src/app/Scenes/HomeView/Sections/HomeViewSectionExploreBy.tsx
+++ b/src/app/Scenes/HomeView/Sections/HomeViewSectionExploreBy.tsx
@@ -1,0 +1,128 @@
+import { Flex, Image, Text, Touchable, useScreenDimensions, useSpace } from "@artsy/palette-mobile"
+import { HomeViewSectionExploreByQuery } from "__generated__/HomeViewSectionExploreByQuery.graphql"
+import { HomeViewSectionExploreBy_section$key } from "__generated__/HomeViewSectionExploreBy_section.graphql"
+import { SectionSharedProps } from "app/Scenes/HomeView/Sections/Section"
+import { navigate } from "app/system/navigation/navigate"
+import { extractNodes } from "app/utils/extractNodes"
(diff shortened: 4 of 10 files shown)


src/app/routes.ts
```diff
@@ -234,6 +234,7 @@ export function getDomainMap(): Record<string, RouteMatcher[] | null> {
     addRoute("/city-fair/:citySlug", "CityFairList"),
     addRoute("/city-save/:citySlug", "CitySavedList"),
     addRoute("/city/:citySlug/:section", "CitySectionList"),
+    addRoute("/collections", "Collections"),
```
araujobarret, line 237: @anandaroop any thoughts on this route? I made it in a more REST way so we can navigate with `/collections?category=Medium"`

src/app/Scenes/Collections/Collections.tsx
```diff
@@ -0,0 +1,27 @@
+import { Flex, Screen, Text } from "@artsy/palette-mobile"
+import { RouteProp, useRoute } from "@react-navigation/native"
+import { goBack } from "app/system/navigation/navigate"
+import { FC } from "react"
+
+type CollectionsNavigationRoutes = {
+  collections: { props: { category: string } }
+}
+
+export const Collections: FC = () => {
```
araujobarret, line 10: This is just a temporary placeholder, will be worked properly in upcoming tickets

ArtsyOpenSource:

- Cross-platform user-facing changes (feat: add explore by section in the homeview - araujobarret)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 933f6501f1e8f00ea724c0b649afd5828bf419ff
</p>

src/app/Scenes/HomeView/Sections/HomeViewSectionExploreBy.tsx
```diff
@@ -0,0 +1,128 @@
...
+  }
+`
+
+const HomeViewExploreByPlaceholder: React.FC = () => {
+  return null
+}
+
+export const HomeViewSectionExploreByQueryRenderer: React.FC<SectionSharedProps> = withSuspense(
+  ({ sectionID, index, ...flexProps }) => {
+    const isEnabled = useFeatureFlag("AREnableMarketingCollectionsCategories")
```
dblandin, line 107: **question**: This is something we should align on best practices for as the __section instance__ definition [already specifies a feature flag](https://github.com/artsy/metaphysics/blob/fc6a57aa2d25cb157b68e5620665714a80324b18/src/schema/v2/homeView/sections/ExploreByCategory.ts#L42) and inclusion can be controlled via the Unleash dashboard across different environments.

Seems potentially confusing to coordinate between and understand the differences across these two different feature flags although it does seem useful to continue to have a way within Eigen to signal that section type support is ready to release to end users.

src/app/routes.ts
dblandin: We already have a [/collections](https://artsy.net/collections) route on web and as this route is doing something a bit different and doesn't yet have any web equivalent or deep-linking expectations, maybe something a bit more explicit would make sense?

```ts
addRoute("/collections-by-category/:category", "CollectionsByCategory")
// or
addRoute("/explore-by-category/:category", "ExploreByCategory")
```

src/app/Scenes/HomeView/Sections/HomeViewSectionExploreBy.tests.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { setupTestWrapper } from "app/utils/tests/setupTestWrapper"
+import { graphql } from "react-relay"
+
+describe("HomeViewSectionExploreBy", () => {
+  const { renderWithRelay } = setupTestWrapper<HomeViewSectionExploreByTestsQuery>({
+    Component: (props) => <HomeViewSectionExploreBy section={props.homeView.section} />,
+    query: graphql`
+      query HomeViewSectionExploreByTestsQuery @relay_test_operation {
+        homeView @required(action: NONE) {
+          section(id: "home-view-section-explore-by-category") @required(action: NONE) {
```
dblandin, line 13: **question**: I don't have much familiarity with this `@required` directive but based on the [documentation](https://relay.dev/docs/guides/required-directive/#none-expected), it states that by using `action: NONE` we expect this field to be null sometimes.

The root level `homeView` field is currently [non-nullable](https://github.com/artsy/metaphysics/blob/afb5f5512a520b1021165e482028b1c7155d1600/src/schema/v2/homeView/index.ts#L82).

This isn't the case now but it looks like we should be able tighten up the types for [homeView/section and homeView/sectionsConnection](https://github.com/artsy/metaphysics/blob/afb5f5512a520b1021165e482028b1c7155d1600/src/schema/v2/homeView/index.ts#L76-L77) to also be non-nullable.

When we can't find a section by ID, we currently throw an error rather than responding with a null section value.

dblandin (commented):
Looking good! Left a few comments.

src/app/Scenes/HomeView/Sections/HomeViewSectionExploreBy.tests.tsx
araujobarret: That's just the mocked query on the test, in this scenario will never be null 👍 
(I did not use this directive in the normal file, just here in the test)

dblandin: Ahh, gotcha 👍🏼

src/app/Scenes/HomeView/Sections/HomeViewSectionExploreBy.tsx
araujobarret: That's indeed confusing, but there's a conflict because unleash doesn't control the app versions.

The architecture of the home view with ongoing work we might start enabling new elements when you don't even have the entire flow ready, example: in version 8.53.0 it had no collection by category screen whereas in version 8.54.0 that was delivered, so we enable the ff and will have the flow broken on the 8.53.0 and ok in the 8.54.0. 
There's no way to avoid the feature flag combo here to only deliver when all the feature work is done.
More discussions [here](https://artsy.slack.com/archives/C05EEBNEF71/p1727445772237339?thread_ts=1727336233.733129&cid=C05EEBNEF71)

src/app/routes.ts
araujobarret: Then yeah, it makes sense to have something more specific in this case. I think I'll go with `/collections-by-category/:category` because it was something that we previously talked with @anandaroop too, thanks!

src/app/Scenes/HomeView/Sections/HomeViewSectionExploreBy.tsx
dblandin: Yeah, I totally understand the need. Made we can add some guidance re: feature flags in our [home view docs page in MP](https://github.com/artsy/metaphysics/blob/main/docs/adding_a_new_home_view_section.md).

araujobarret: I think it'll be better, if it's a feature that renders something in a new shape but redirects to older screens/has behaviors supported by past versions we can rely solely on MP ff, however when it adds more stuff on Eigen we cannot avoid, maybe this distinction will be also necessary

src/app/Scenes/CollectionsByCategory/CollectionsByCategory.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { FC } from "react"
+
+type CollectionsByCategoriesNavigationRoutes = {
+  collections: { props: { category: string } }
+}
+
+export const CollectionsByCategory: FC = () => {
+  const { params } = useRoute<RouteProp<CollectionsByCategoriesNavigationRoutes, "collections">>()
+
+  console.log(params)
```
damassi, line 13: 🦑

src/app/Scenes/HomeView/Sections/Section.tsx
```diff
@@ -62,6 +63,8 @@ export const Section: React.FC<SectionProps> = ({ section, ...rest }) => {
       return <HomeViewSectionAuctionResultsQueryRenderer sectionID={section.internalID} {...rest} />
     case "HomeViewSectionHeroUnits":
       return <HomeViewSectionHeroUnitsQueryRenderer sectionID={section.internalID} {...rest} />
+    case "HomeViewSectionCards":
+      return <HomeViewSectionExploreByQueryRenderer sectionID={section.internalID} {...rest} />
```
anandaroop, line 67: **suggestion:** We should make the component name here match the section type:

```suggestion
    case "HomeViewSectionCards":
      return < HomeViewSectionCards QueryRenderer sectionID={section.internalID} {...rest} />
```

That reinforces the expectation that `HomeViewSectionCards` has a default presentation style, the one you are adding here. 

(And then some future section may declare it is also a `HomeViewSectionCards` but has a different presentation, as discussed in [the thread](https://artsy.slack.com/archives/C05EEBNEF71/p1728404106250269?thread_ts=1728401640.597839&cid=C05EEBNEF71) — but the _default_ style should be clearer here from the component naming)

anandaroop (changes requested):
Looking great, just one suggestion to keep naming conventions consistent with the patterns we've established.

araujobarret:
I've addressed all the requested changes, need just a 👍 in order to merge it.

src/app/Scenes/HomeView/Sections/Section.tsx
anandaroop: Thank you! I noticed a very similar issue [over here](https://github.com/artsy/eigen/pull/10927/files#diff-980cb0bae6931064b4372248200cf0c852e8b8d0565ac6393ce60a6e8bfd55ccR46-R52), but I will comment on that separately

anandaroop (dismissed):

src/app/Scenes/HomeView/Sections/Section.tsx
araujobarret: I have to touch again on the discovery by implementing missing features and gonna pack this change there, didn't want to ship here bc it was out of the scope of this PR 👍

PR #10927, opened 2024-10-10
fix(DiscoverSomethingNew): fix graphql type

araujobarret:
### Description

This PR fixes the CI MP integration broken given a type that doesn't exist anymore for the element Discover Something New. Also protects the element from being displayed with ff.

### PR Checklist

### To the reviewers 👀

- fix: Discover something new graphql type

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Scenes/HomeView/Sections/Section.tsx
+++ b/src/app/Scenes/HomeView/Sections/Section.tsx
@@ -43,6 +43,13 @@ export const Section: React.FC<SectionProps> = ({ section, ...rest }) => {
       )
     case "ArticlesCard":
       return <HomeViewSectionArticlesCardsQueryRenderer sectionID={section.internalID} {...rest} />
+    case "Chips":
+      return (
+        <HomeViewSectionDiscoverMarketingCollectionsQueryRenderer
+          sectionID={section.internalID}
+          {...rest}
+        />
+      )
   }
 
   switch (section.__typename) {
@@ -77,13 +84,6 @@ export const Section: React.FC<SectionProps> = ({ section, ...rest }) => {
       return <HomeViewSectionViewingRoomsQueryRenderer sectionID={section.internalID} {...rest} />
     case "HomeViewSectionSales":
       return <HomeViewSectionSalesQueryRenderer sectionID={section.internalID} {...rest} />
-    case "HomeViewSectionDiscoverMarketingCollections":
-      return (
-        <HomeViewSectionDiscoverMarketingCollectionsQueryRenderer
-          sectionID={section.internalID}
-          {...rest}
-        />
-      )
     default:
       if (__DEV__) {
         return (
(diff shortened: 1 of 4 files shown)


egdbear (approved):
🚀 🚀 🚀 🚀 🚀

ArtsyOpenSource:

- Cross-platform user-facing changes (fix: Discover something new graphql type)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 1c7c3eac7f6b7f121a9b4e406c506f908fe736f9
</p>

src/app/Scenes/HomeView/Sections/Section.tsx
```diff
@@ -43,6 +43,13 @@ export const Section: React.FC<SectionProps> = ({ section, ...rest }) => {
       )
     case "ArticlesCard":
       return <HomeViewSectionArticlesCardsQueryRenderer sectionID={section.internalID} {...rest} />
+    case "Chips":
+      return (
+        <HomeViewSectionDiscoverMarketingCollectionsQueryRenderer
+          sectionID={section.internalID}
+          {...rest}
+        />
```
anandaroop, line 51: Note how the component names on L42, L45 above match the the `section.component.type` attribute coming from MP. 

Let's adhere to that pattern to keep things 1:1

I would even go one step further and recommend something like

```suggestion
    case "CardsChips":
      return (
        <HomeViewSectionCardsChipsQueryRenderer
          sectionID={section.internalID}
          {...rest}
        />
```

This is a direct parallel to `ArticlesCard` above — it incorporates the section type (Articles) and the presentational style (Card) in the name.

I realize `CardsChips` doesn't exactly roll off the tongue nicely, so I'm open to suggestion 😅. But at the very least I would prefer to avoid having `section.component.type` drift completely from the component naming — those were always intended to correspond to one another in my view.

(cc @dblandin if you have thoughts)

anandaroop: And just to reiterate the more general point about naming — components that are general and can be re-used in a general way should have naming that reflects that. This component name was overly coupled to the collections-discovery use case imo.

dblandin: Yeah, I agree that ideally the component / query renderer name should match the requested component type value. So `HomeViewSection[Chips/CardChips]QueryRenderer`.

Will let my subconscious mull over "CardsChips" naming over the weekend 😅

araujobarret: This was a fix and not a refactor haha
I'm already doing the renaming in the feat of the discovery something new 😄

araujobarret: I was writing CardsChips, in my head made sense the plural + plural but I ain't no native speaker 🫠

PR #10937, opened 2024-10-10
refactor: unable to load screen to add tracking for screens

gkartalis:
This PR resolves [PHIRE-1241] 

### Description

Adds screen tracking in sentry stack trace message for unable to load screens like so:

```
Unable to load in tab: home params:  {"moduleName": "Artist", "props": {"artistID": "kaws"}, "stackID": "home"}
```

Waiting to trigger a codepush beta first to test out that this works 100% and then we could merge.

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- unable to load screen to add tracking for screens - gkartalis

</details>

[PHIRE-1241]: https://artsyproduct.atlassian.net/browse/PHIRE-1241?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Components/LoadFailureView.tsx
+++ b/src/app/Components/LoadFailureView.tsx
@@ -10,6 +10,7 @@ import {
   DEFAULT_HIT_SLOP,
 } from "@artsy/palette-mobile"
 import * as Sentry from "@sentry/react-native"
+import { __unsafe_mainModalStackRef } from "app/NativeModules/ARScreenPresenterModule"
 import { GlobalStore } from "app/store/GlobalStore"
 import { goBack } from "app/system/navigation/navigate"
 import { useDevToggle } from "app/utils/hooks/useDevToggle"
@@ -55,7 +56,7 @@ export const LoadFailureView: React.FC<LoadFailureViewProps & BoxProps> = ({
 
   const showErrorMessage = __DEV__ || isStaging || showErrorInLoadFailureViewToggle
 
-  const trackLoadFailureView = (error: Error | undefined) => {
+  const trackLoadFailureView = (error: Error | undefined, routeParams: any) => {
     const shouldTrackError = !__DEV__ && !isStaging && trackErrorBoundary
     if (shouldTrackError) {
       Sentry.withScope((scope) => {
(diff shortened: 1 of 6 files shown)


src/app/Components/LoadFailureView.tsx
```diff
@@ -55,22 +56,27 @@ export const LoadFailureView: React.FC<LoadFailureViewProps & BoxProps> = ({
...
     const shouldTrackError = !__DEV__ && !isStaging && trackErrorBoundary
     if (shouldTrackError) {
       Sentry.withScope((scope) => {
         scope.setExtra("user-id", userId)
         if (error) {
           scope.setExtra("error", error)
         }
-        Sentry.captureMessage("Unable to load in tab: " + activeTab, "error")
+        Sentry.captureMessage(
+          "Unable to load in tab: " + activeTab + " params:" + JSON.stringify(routeParams),
```
gkartalis, line 68: the message is open for feedback. It will theoretically looks like this:

```
Unable to load in tab: home params: {"moduleName": "Artist", "props": {"artistID": "kaws"}, "stackID": "home"}
```

I wouldn't change the activeTab part for the first iteration and if we see that the stackID and moduleName are consistent we can then refactor.

gkartalis: canveat of this approach is that we will have only new unable to load tracking due to the change of message but it will be theoretically way easier to debug and locate the screens

ArtsyOpenSource:

- Dev changes (unable to load screen to add tracking for screens - gkartalis)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 6c4c8eb84b9920b676023cbed619e32157f65a1f
</p>

src/app/Scenes/Artist/Artist.tsx
```diff
@@ -125,6 +125,8 @@ export const Artist: React.FC<ArtistProps> = (props) => {
     [artistAboveTheFold, headerHeight]
   )
 
+  throw new Error("This is George testing stuff in staging only")
+
```
gkartalis, line 129: this will get reverted before merging just wanna trigger a codepush beta to see how this error above looks in sentry

src/app/Components/LoadFailureView.tsx
gkartalis: @MounirDhahri Could you elaborate? not sure I understand.

The current implementation will track something like this:

```
Unable to load in tab: home params: {"moduleName": "Artist", "props": {"artistID": "kaws"}, "stackID": "home"}
```

and afterwards I was thinking that if we see that the params part working reliably to keep afterwards the stackID and moduleName that are basically the tab + screen. But before we let it out I would personally refrain from removing the previous tab report and I would just enhance them for now.

cc @brainbicycle @MounirDhahri

MounirDhahri: good point - I got what you mean now. Let's do that

brainbicycle: alternatively keep the name in the message the same `Unable to load in tab: home` and just attach the params for diagnostics to the scope

PR #10938, opened 2024-10-10
fix(DIA-901): tidy-up forgot password step [WIP]

iskounen:
This PR tidies up the Forgot Password? step of the new authentication flow.

- Pre-populate the email address
- Return users to the password step for back navigation and "Return to login" buttons

https://github.com/user-attachments/assets/6d1f0cc5-a248-4669-a01a-adddea92ee5c

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- Added touch-ups to auth2 forgot password step - iskounen

</details>

### DIFF
--- a/src/app/Scenes/Onboarding/Auth2/scenes/ForgotPasswordStep.tsx
+++ b/src/app/Scenes/Onboarding/Auth2/scenes/ForgotPasswordStep.tsx
@@ -20,14 +20,10 @@ export const ForgotPasswordStep: React.FC = () => {
   return (
     <Formik<ForgotPasswordStepFormValues>
       initialValues={{ email: screen.params?.email ?? "" }}
-      onSubmit={async ({ email }, { setErrors, resetForm }) => {
-        console.log("🧸")
-
+      onSubmit={async ({ email }, { setErrors }) => {
         const res = await GlobalStore.actions.auth.forgotPassword({
           email,
         })
-
-        console.log({ res })
 
         if (!res) {
           setErrors({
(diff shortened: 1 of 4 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 3f63a73df65f59e528ea15eaa87bece1be31a97c
</p>

src/app/Scenes/Onboarding/Auth2/scenes/ForgotPasswordStep.tsx
```diff
@@ -167,7 +173,7 @@ const ForgotPasswordStepForm: React.FC = () => {
           block
           variant="fillDark"
           haptic="impactMedium"
-          disabled={!isValid}
+          disabled={!isValid || !values.email}
```
araujobarret, line 176: do we need the extra `!values.email` here? I assume if we don't have the email set it's not valid because it's a required form field, isn't it?

iskounen: You're right about the email **field** not being valid when it is empty, however Formik thinks that the **form** is valid when it is initially displayed because I do not have `validateOnMount` set to true.

I've avoided using `validateOnMount` because it causes validation errors to be displayed when the form is initially loaded. I can mitigate those using `errors={submittedCount > 0 ? errors.email : unknown}` but I don't think that solution is better than this one.

I've also avoided using `validateOnMount` because in a stack navigator, it is only called when the screen is initialized, but I expected it to be called every time the user navigates back to this screen, so that made things more confusing for me.

Having said all of that, I think that I can use `validateOnMount` on this step since I'm not showing an error message when validation fails, so I'll make this change.

PR #10951, opened 2024-10-11
fix: keep splash screen visible for 500ms

MounirDhahri:
### Description

This PR is an attempt to improve our homeview elements rendering logic. 
The idea from this PR is the following, we don't want to show an empty home view then suddenly have multiple sections show up at the same time. We can't avoid that because the home view query is still active. So what can we do about it?
We can wait for the prefetching query to finish. That's good, but! what if it takes forever or never finishes god forbid! 
How about we wait for a bit, enough for the query to finish (_usually_), then hide the splash screen. For how long should we wait, the average [query duration in datadog ](https://app.datadoghq.com/dashboard/z95-ftb-4qe/home-view-health?fromUser=true&refresh_mode=sliding&tpl_var_env%5B0%5D=%2A&from_ts=1728057263324&to_ts=1728662063324&live=true)

Screenshot 2024-10-11 at 18 10 49

**Note: I know this will add 500ms to the splash screen but the user was going to wait anyway and it's pretty subtle. Let me know what you think**

https://github.com/user-attachments/assets/aedf925e-125d-41aa-bba6-04dc295b8174

### PR Checklist

### To the reviewers 👀

- gracefully show home view content after splash screen - mounir

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/utils/useHideSplashScreen.ts
+++ b/src/app/utils/useHideSplashScreen.ts
@@ -6,7 +6,7 @@ import { usePrefetch } from "app/utils/queryPrefetching"
 import { useEffect } from "react"
 import RNBootSplash from "react-native-bootsplash"
 
-const HOME_VIEW_QUERY_AVERAGE_LATENCY = 500
+const HOME_VIEW_SPLASH_SCREEN_DELAY = 500
 
 export const useHideSplashScreen = () => {
   const isNavigationReady = GlobalStore.useAppState((state) => state.sessionState.isNavigationReady)
@@ -32,7 +32,7 @@ export const useHideSplashScreen = () => {
           })
           setTimeout(() => {
             hideSplashScreen()
-          }, HOME_VIEW_QUERY_AVERAGE_LATENCY)
+          }, HOME_VIEW_SPLASH_SCREEN_DELAY)
         }
         return
       }
--- a/src/app/utils/useHideSplashScreen.ts
+++ b/src/app/utils/useHideSplashScreen.ts
@@ -6,6 +6,8 @@ import { usePrefetch } from "app/utils/queryPrefetching"
 import { useEffect } from "react"
 import RNBootSplash from "react-native-bootsplash"
 
+const HOME_VIEW_QUERY_AVERAGE_LATENCY = 500
+
 export const useHideSplashScreen = () => {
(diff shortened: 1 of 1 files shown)


src/app/utils/useHideSplashScreen.ts
```diff
@@ -28,20 +30,15 @@ export const useHideSplashScreen = () => {
...
+            hideSplashScreen()
+          }, HOME_VIEW_QUERY_AVERAGE_LATENCY)
         }
         return
       }
       if (!isLoggedIn) {
         hideSplashScreen()
       }
     }
-    hideSplashScreen()
```
MounirDhahri, line 45: this isn't needed

ArtsyOpenSource:

- Cross-platform user-facing changes (gracefully show home view content after splash screen - mounir)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 7aab59b28ad5a81201129f3c0337afcc9c962bbe
</p>

src/app/utils/useHideSplashScreen.ts
```diff
@@ -6,6 +6,8 @@ import { usePrefetch } from "app/utils/queryPrefetching"
 import { useEffect } from "react"
 import RNBootSplash from "react-native-bootsplash"
 
+const HOME_VIEW_QUERY_AVERAGE_LATENCY = 500
```
anandaroop, line 9: **suggestion (non-blocking):** Just a minor nit about the name. Since this is influenced by the avg latency, but doesn't actually represent it — maybe `HOME_VIEW_SPLASH_SCREEN_DELAY` or something like that would be a more intention-revealing name.

anandaroop (dismissed):
LGTM

dblandin (dismissed):
Nice, sounds like a nice way to reduce perceived delay + visual shifting 👍🏼

src/app/utils/useHideSplashScreen.ts
gkartalis: I think I agree with this one

gkartalis (dismissed):

damassi:
Its unintuitive but fake wait times used well are one of the best tools. Nice!

PR #11005, opened 2024-10-23
feat(ONYX-1294): redesign badges and urgency signals on artwork card

dariakoko:
This PR resolves [ONYX-1294] 

### Description

| Increased Interest | Increased Interest Artwork screen | Curators' pick | Curators' pick Artwork screen | Limited-Time Offer | Limited-Time Offer and Increased interest on the Grid | Hight demand Signal in MyCollection |
| --- | --- | --- | --- | --- | --- | --- |
| image | image | image | image | image | image | image |

Redesign badges and urgency signals on artwork card
- remove Limited-Time Offer badge from the artwork
- add original artwork price to the Limited-Time Offer artwork
- change the copy Expired for Limited-Time Offer artwork
- redesign the Curators pick badge
- redesign Increased interest badge
- move the new badge under the artwork's metadata
- redesign the Hight Demand indicator in MyCollection

### PR Checklist

### To the reviewers 👀

- redesign badges and urgency signals on artwork card - daria

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[ONYX-1294]: https://artsyproduct.atlassian.net/browse/ONYX-1294?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/Components/ArtworkGrids/ArtworkGridItem.tsx
```diff
@@ -288,6 +288,74 @@ export const Artwork: React.FC<ArtworkProps> = ({
     await item._disappearable?.disappear()
   }
 
+  const getSaleMessage = () => {
```
olerichter00, line 291: Because this function returns a component, it would make sense to move it outside of `ArtworkGridItem` and call it `ArtworkSaleMessage` (or something similar). If it uses artwork data, it could have its own fragment as well (`useFragment`).

src/app/Components/ArtworkGrids/ArtworkGridItem.tsx
```diff
@@ -288,6 +288,74 @@ export const Artwork: React.FC<ArtworkProps> = ({
...
+    // TODO: move into a hook, rename getOfferMessage???
+    const parts = saleInfo && saleInfo.split(/(~.*?~)/)
```
olerichter00, line 293: I think it's okay to keep this inline unless it's used in more than one place, in which case it makes sense to create a helper function to share the code between components.

src/app/Components/ArtworkGrids/ArtworkSocialSignal.tsx
```diff
@@ -21,31 +21,19 @@ export const ArtworkSocialSignal: React.FC<ArtworkSocialSignalProps> = ({
...
-          pb="2px"
-          alignSelf="flex-start"
-          borderRadius={3}
-          borderWidth={1}
-          borderColor={primaryColor}
-        >
-          <Text color={primaryColor} variant="xxs">
+        <Box flexDirection="row" alignItems="center">
+          <ArrowUpRightIcon />
+          <Text color={primaryColor} variant="xs" pl="3px">
```
olerichter00, line 36: Would this work as well? IMO it's better to use an empty space character instead of padding. But I don't feel very strong about it.

```suggestion
          <Text color={primaryColor} variant="xs">{" "}Increased Interest</Text>
```

src/app/Components/ArtworkRail/ArtworkRailCardMeta.tsx
```diff
@@ -132,6 +132,69 @@ export const ArtworkRailCardMeta: React.FC<ArtworkRailCardMetaProps> = ({
     onCompleted: onArtworkSavedOrUnSaved,
   })
 
+  const getSaleMessage = () => {
```
olerichter00, line 135: If this is a duplication of the first `getSaleMessage` it would even make more sense to extract the code into it's own component.

src/app/Components/ArtworkRail/ArtworkRailCardMeta.tsx
```diff
@@ -242,6 +260,18 @@ export const ArtworkRailCardMeta: React.FC<ArtworkRailCardMetaProps> = ({
         {!!displayAuctionSignal && !!collectorSignals && (
           <ArtworkAuctionTimer collectorSignals={collectorSignals} inRailCard />
         )}
+
+        {!sale?.isAuction &&
+          !displayLimitedTimeOfferSignal &&
+          !!collectorSignals &&
+          !!enableCuratorsPicksAndInterestSignals && (
```
olerichter00, line 267: What do you think about extracting this into a const like `displayArtworkSocialSignal`?

src/app/Components/ContextMenu/ContextMenuArtworkPreviewCard.tsx
```diff
@@ -122,7 +122,7 @@ export const ContextMenuArtworkPreviewCard: React.FC<ContextMenuArtworkPreviewCa
                   numberOfLines={1}
                   fontWeight={500}
                 >
-                  {saleMessage}
+                  11 {saleMessage}
```
olerichter00, line 125: 👀 

```suggestion
                  {saleMessage}
```

src/app/utils/getSaleMessgeOrBidInfo.ts
```diff
@@ -84,7 +84,9 @@ export const saleMessageOrBidInfo = ({
   }
 
   if (collectorSignals?.partnerOffer?.isAvailable) {
-    return collectorSignals.partnerOffer.priceWithDiscount?.display
+    const salePrice = artwork.saleMessage && `~${artwork.saleMessage}~`
```
olerichter00, line 87: Are these `~` intentional?

src/app/Components/ArtworkRail/ArtworkRailCardMeta.tsx
dariakoko: It is used in only one place

src/app/utils/getSaleMessgeOrBidInfo.ts
dariakoko: Yes. `saleMessageOrBidInfo` returns a string and we display it in a Text component. For the LTO we will need to return discounted price as well as original price and the original price has to have `line-through` text decoration. In order to not change the returned type I came up with this solution - mark the text that has to have line-through decoration with `~` symbol and then parse the returned string. I'm not sure if this is the best approach, but it seems to work and doesn't require a lot of testing

ArtsyOpenSource:

- Cross-platform user-facing changes (redesign badges and urgency signals on artwork card - daria - dariakoko)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against b104397488ddbc624cd94b1387c7bebd21a9e7b8
</p>

src/app/Components/ArtworkGrids/ArtworkGridItem.tsx
dariakoko: It is used in only one place

src/app/Components/ArtworkGrids/ArtworkSocialSignal.tsx
```diff
@@ -21,31 +21,20 @@ export const ArtworkSocialSignal: React.FC<ArtworkSocialSignalProps> = ({
...
-          alignSelf="flex-start"
-          borderRadius={3}
-          borderWidth={1}
-          borderColor={primaryColor}
-        >
-          <Text color={primaryColor} variant="xxs">
+        <Box flexDirection="row" alignItems="center">
+          <ArrowUpRightIcon />
+          <Text color={primaryColor} variant="xs">
+            {" "}
```
anandaroop, line 37: **question:** Is this space character intentional? 

I don't see it on the other one^, but there is an explicit `pl` there. Wondering if these were meant to be consistent.

src/app/utils/getSaleMessgeOrBidInfo.ts
anandaroop: **question:** Should that be handled with a ternary instead of a `&&` though? 

I'm wondering if it's possible that `artwork.saleMessage` could be null or undefined, in which case this could render in an undesirable way with the `&&` trick. 

(It is possible for an artwork sale message to be null in Gravity, [under certain conditions](https://github.com/artsy/gravity/blob/a411591753d330bc50a69167c666d38450402d2e/app/models/concerns/for_sale.rb#L368))

src/app/Components/ArtworkGrids/ArtworkSocialSignal.tsx
dariakoko: The space is intentional and should be replacing pl="3px". Thanks!

src/app/utils/getSaleMessgeOrBidInfo.ts
dariakoko: We don't need `artwork.saleMessage && `here at all actually, I will remove it

dariakoko: You're right, I missed that 👍

olerichter00: That's a great solution. To make it more obvious what's happening it could be good to extract the logic into a serialize and a parse function (maybe just declared within the same file). Alternatively two comments could also help to understand when reading the code. 

Extracting the logic would be my favorite :)

src/app/Components/ArtworkGrids/ArtworkSocialSignal.tsx
```diff
@@ -21,31 +21,21 @@ export const ArtworkSocialSignal: React.FC<ArtworkSocialSignalProps> = ({
...
-          alignSelf="flex-start"
-          borderRadius={3}
-          borderWidth={1}
-          borderColor={primaryColor}
-        >
-          <Text color={primaryColor} variant="xxs">
+        <Box alignItems="center" flexDirection="row">
+          <FireIcon />
+          <Text color={primaryColor} variant="xs" textAlign="center">
+            {" "}
```
MounirDhahri, line 27: is this intentional? you can do `ml={0.5}`

MounirDhahri: okey I see the same question [here](https://github.com/artsy/eigen/pull/11005/files#r1815463364) - I think ml="3px"would make sense here

src/app/Components/ArtworkRail/ArtworkSaleMessageComponent.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+interface ArtworkSaleMessageComponentProps {
+  artwork: ArtworkSaleMessageComponent_artwork$key
+  displayLimitedTimeOfferSignal: boolean | null | undefined
+  saleMessage: string | null | undefined
+  saleInfoTextStyle?: TextProps
+  dark?: boolean
+}
+
+export const ArtworkSaleMessageComponent: React.FC<ArtworkSaleMessageComponentProps> = ({
```
MounirDhahri, line 17:
```suggestion
export const ArtworkSaleMessage: React.FC<ArtworkSaleMessageProps> = ({
```

MounirDhahri (dismissed):

PR #11016, opened 2024-10-24
fix: design QA fixes omnibus

iskounen:
This PR resolves [DIA-938] and [DIA-939] and [DIA-940].

### Description

- Decrease the corner rounding from 20px to 10px
- Increase the line height of the disclaimer text from 14px to 16px
- Return to the email step from the forgot password step
- Memoize the recaptcha component definition

| Before | After |
|--------|------|
| 20px - 14x | 10px - 16px |

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[DIA-938]: https://artsyproduct.atlassian.net/browse/DIA-938?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DIA-939]: https://artsyproduct.atlassian.net/browse/DIA-939?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DIA-940]: https://artsyproduct.atlassian.net/browse/DIA-940?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Components/Recaptcha/Recaptcha.tsx
+++ b/src/app/Components/Recaptcha/Recaptcha.tsx
@@ -1,6 +1,6 @@
 import { Box } from "@artsy/palette-mobile"
 import { RecaptchaWebView } from "app/Components/Recaptcha/RecaptchaWebView"
-import { useState } from "react"
+import { useCallback, useState } from "react"
 
 type State = "idle" | "error" | undefined
 type UseRecaptchaProps = { source: string; action: string }
@@ -18,10 +18,13 @@ export const useRecaptcha = ({ source, action }: UseRecaptchaProps) => {
     setState("error")
   }
 
-  const RecaptchaComponent = () => (
-    <Box height={0}>
-      <RecaptchaWebView action={action} onToken={handleOnToken} onError={handleOnError} />
-    </Box>
+  const RecaptchaComponent = useCallback(
+    () => (
+      <Box height={0}>
+        <RecaptchaWebView action={action} onToken={handleOnToken} onError={handleOnError} />
+      </Box>
+    ),
+    []
   )
 
   return { Recaptcha: RecaptchaComponent, token, state }
--- a/src/app/Components/Recaptcha/Recaptcha.tsx
+++ b/src/app/Components/Recaptcha/Recaptcha.tsx
@@ -18,14 +18,21 @@ export const useRecaptcha = ({ source, action }: UseRecaptchaProps) => {
     setState("error")
   }
 
-  const RecaptchaComponent = useCallback(
-    () => (
+  interface RecaptchaComponentProps {
+    active?: boolean
+  }
+
+  const RecaptchaComponent: React.FC<RecaptchaComponentProps> = useCallback(({ active }) => {
+    if (!active) {
+      return null
+    }
+
+    return (
       <Box height={0}>
         <RecaptchaWebView action={action} onToken={handleOnToken} onError={handleOnError} />
       </Box>
-    ),
-    []
-  )
+    )
+  }, [])
 
   return { Recaptcha: RecaptchaComponent, token, state }
 }
--- a/src/app/Scenes/Onboarding/Auth2/hooks/useAuthNavigation.tsx
+++ b/src/app/Scenes/Onboarding/Auth2/hooks/useAuthNavigation.tsx
@@ -4,10 +4,8 @@ export const useAuthNavigation = () => {
   const actions = AuthContext.useStoreActions((actions) => actions)
 
   return {
-    goBack: (count = 1) => {
-      for (let i = 0; i < count; i++) {
-        actions.goBack()
-      }
+    goBack: () => {
+      actions.goBack()
     },
     navigate: (screen: AuthScreen) => {
       actions.setCurrentScreen(screen)
--- a/src/app/Scenes/Onboarding/Auth2/hooks/useAuthNavigation.tsx
+++ b/src/app/Scenes/Onboarding/Auth2/hooks/useAuthNavigation.tsx
@@ -4,8 +4,10 @@ export const useAuthNavigation = () => {
   const actions = AuthContext.useStoreActions((actions) => actions)
 
   return {
-    goBack: () => {
-      actions.goBack()
+    goBack: (count = 1) => {
+      for (let i = 0; i < count; i++) {
+        actions.goBack()
+      }
     },
     navigate: (screen: AuthScreen) => {
       actions.setCurrentScreen(screen)
--- a/src/app/Scenes/Onboarding/Auth2/scenes/ForgotPasswordStep.tsx
+++ b/src/app/Scenes/Onboarding/Auth2/scenes/ForgotPasswordStep.tsx
@@ -69,6 +69,11 @@ const ForgotPasswordStepForm: React.FC = () => {
     resetForm({ values: { email: screen.params?.email ?? "" } })
   }
 
+  const handleReturnToLoginPress = () => {
+    navigation.goBack(2)
+    resetForm({ values: { email: screen.params?.email ?? "" } })
+  }
+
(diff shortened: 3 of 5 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 74efbc691709232fa85bfaeaa678511ec1e7a103
</p>

src/app/Components/Recaptcha/Recaptcha.tsx
```diff
@@ -18,10 +18,13 @@ export const useRecaptcha = ({ source, action }: UseRecaptchaProps) => {
...
-    <Box height={0}>
-      <RecaptchaWebView action={action} onToken={handleOnToken} onError={handleOnError} />
-    </Box>
+  const RecaptchaComponent = useCallback(
+    () => (
+      <Box height={0}>
+        <RecaptchaWebView action={action} onToken={handleOnToken} onError={handleOnError} />
+      </Box>
+    ),
+    []
```
iskounen, line 27: @araujobarret we memoized this because we found that it was causing the email form to constantly re-render. While this prevents that from happening, it prevents me from being able to perform recaptcha validation more than once.

I need to find some way to refresh this when the user navigates back to the email step...

damassi: Is it a problem that the email form constantly rerenders? 

The answer here is to pass a prop into the recapcha component, and then thread that into the dependency array. So on back do something like:

```
const isFocused = useScreenFocused() // or whatever it is 

<Recaptcha enabled={isFocused} />
```
and then pass that prop into the array.

src/app/Scenes/Onboarding/Auth2/hooks/useAuthNavigation.tsx
```diff
@@ -4,8 +4,10 @@ export const useAuthNavigation = () => {
   const actions = AuthContext.useStoreActions((actions) => actions)
 
   return {
-    goBack: () => {
-      actions.goBack()
+    goBack: (count = 1) => {
+      for (let i = 0; i < count; i++) {
+        actions.goBack()
+      }
```
damassi, line 10: ehhh -- why is this here? And can we use functional programming practices -- if this is necessary! 

In general (as a guiding principle) usually code like this should be considered a code smell -- meaning, the ideal form -- how should it work in the simplest possible case? -- is the best form, and anything to get to that ideal form will ultimately make the system work better. So, solve that and 👨‍🍳

src/app/Scenes/Onboarding/Auth2/scenes/ForgotPasswordStep.tsx
```diff
@@ -69,6 +69,11 @@ const ForgotPasswordStepForm: React.FC = () => {
     resetForm({ values: { email: screen.params?.email ?? "" } })
   }
 
+  const handleReturnToLoginPress = () => {
+    navigation.goBack(2)
```
damassi, line 73: Re the previous comment, if we're going back to a certain step, lets pass the screen id -- and jump straight to that, vs anonymous integers. The ideal form says: if we want to go to the email screen, we'd ```
navigation.navigate("EmailScreen") 
```
or something of that nature.

File: src/app/Scenes/Onboarding/Auth2/scenes/LoginWelcomeStep.tsx [hunk 4]
```diff
@@ -170,21 +170,24 @@ const LoginWelcomeStepForm: React.FC = () => {
 
         <Spacer y={1} />
 
-        <Text variant="xxs" color="black60" textAlign="center">
+        <Text variant="xxs" color="black60" textAlign="center" lineHeight="16px">
           By tapping Continue with Apple, Facebook, or Google, you agree to Artsy’s{" "}
           <LinkText
             variant="xxs"
             onPress={() => navigation.navigate("OnboardingWebView", { url: "/terms" })}
+            lineHeight="16px"
```
damassi, line 178: Why are we stepping out of the design system? If that's feedback from Ryan, its up to us to keep the design team in check with a firm "Unfortunately, no, we can't do that; it's not in the DS". Trivial things like this add up, and devs are the last line of defense.

src/app/Scenes/Onboarding/Auth2/scenes/LoginWelcomeStep.tsx
```diff
@@ -170,21 +170,24 @@ const LoginWelcomeStepForm: React.FC = () => {
...
           >
             Terms and Conditions
           </LinkText>{" "}
           and{" "}
           <LinkText
             variant="xxs"
             onPress={() => navigation.navigate("OnboardingWebView", { url: "/privacy" })}
+            lineHeight="16px"
```
damassi, line 186: ditto.

src/app/Scenes/Onboarding/Auth2/hooks/useAuthNavigation.tsx
damassi: Or... 
```tsx
_.times(2, actions.goBack) 
```

(But even still, we should avoid something like this as this won't scale)

src/app/Components/Recaptcha/Recaptcha.tsx
iskounen: OK, I added an optional `active` prop to the Recaptcha component that tells it when it needs to recreate the RecaptchaWebview.

I tried using useCallback's`[]` parameter to control when that happens, but that means the useRecatpcha hook would need to be aware of the auth modal's steps, which would prevent use form using this hook in other places, so I dropped that idea.

src/app/Components/Recaptcha/Recaptcha.tsx
```diff
@@ -18,11 +18,21 @@ export const useRecaptcha = ({ source, action }: UseRecaptchaProps) => {
...
-      <RecaptchaWebView action={action} onToken={handleOnToken} onError={handleOnError} />
-    </Box>
-  )
+  interface RecaptchaComponentProps {
+    active?: boolean
+  }
+
+  const RecaptchaComponent: React.FC<RecaptchaComponentProps> = useCallback(({ active }) => {
+    if (!active) {
+      return null
```
damassi, line 27: Perfect approach 👍

src/app/Scenes/Onboarding/Auth2/scenes/ForgotPasswordStep.tsx
```diff
@@ -69,6 +69,11 @@ const ForgotPasswordStepForm: React.FC = () => {
     resetForm({ values: { email: screen.params?.email ?? "" } })
   }
 
+  const handleReturnToLoginPress = () => {
+    navigation.navigate({ name: "LoginWelcomeStep" })
```
damassi, line 73: 👌

src/app/Components/Recaptcha/Recaptcha.tsx
damassi: this approach is even better

damassi: Ie, no need for the mental overhead/complexity of callback / memoization hooks

PR #11021, opened 2024-10-25
fix(artwork rail): change image resize mode to "cover"

anandaroop:
This PR resolves this QA launch-blocking bug:
https://www.notion.so/artsy/Border-around-artwork-rail-images-12acab0764a080e298c1d1ca455895c9

### Description

| Before | After |
|--------|--------|
|  |  | 

QA uncovered an issue where some images appear to have hairline rules visible on their borders. 

(See ⬆️ in the **Before** column the sides of the first image and the top/bottom of the second image.)

The hairline itself, if you really zoom in there, turns out to be the blurhash showing through because it is not completely covered by the final image! Notice the gradient from the pale green down to the dark blue here:

<br />
<br />

The root cause appears to be tiny sub-pixel differences between the [calculated size](https://github.com/artsy/eigen/blob/d11b1b2641028e1c41fdbf8260799cf24a4ad5cd/src/app/Components/ArtworkRail/ArtworkRailCardImage.tsx#L42-L53) of the image component and the actual image source file itself.

For example in the "before" image above, the calculated size is `215 x 215` but the image source file is `1100 x 1104`

When resized in `contain` mode, the image is shrunk to <code><b>214.2</b> x 215</code>, which leaves a little strip visible on either side.

Furthermore this difference is really apparent when the artwork image (contrary to our guidelines!) has a white buffer around it, as these images do ([see original](https://d32dm0rphc51dk.cloudfront.net/6J96qGWaeOrgf47eGx4IYg/large.jpg)).

This QA caught a perfect storm of multiple such images which made the problem very visible.

### Solution

If we instead resize this in `cover` mode, the image is scaled down just enough to completely cover the `215 x 215` container, and therefore leaves none of the background blurhash showing through. This fix comes at the expense of "throwing away" a minuscule 1-pixel or so line of image data, which is more acceptable than the alternative imo.

### PR Checklist

### To the reviewers 👀

- Fix hairline image issue on artwork rail images

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Components/ArtworkRail/ArtworkRailCardImage.tsx
+++ b/src/app/Components/ArtworkRail/ArtworkRailCardImage.tsx
@@ -73,7 +73,7 @@ export const ArtworkRailCardImage: React.FC<ArtworkRailCardImageProps> = ({ ...r
           aspectRatio={image.aspectRatio}
           blurhash={showBlurhash ? image.blurhash : undefined}
           performResize={false}
-          resizeMode="contain"
+          resizeMode="cover"
           testID="artwork-rail-card-image"
         />
       )}


ArtsyOpenSource:

- Cross-platform user-facing changes (Fix hairline image issue on artwork rail images - roop - anandaroop)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 67c99a4f29e65fb0130d738bcb3521abfaca07af
</p>

dblandin (approved):
Awesome 💯

Thanks for tracking this down! Going to go ahead and merge so it's included in the Monday morning beta build.

dariakoko:
Thanks @anandaroop !

PR #11027, opened 2024-10-28
chore: remove usage of AREnableAuctionImprovementsSignals feature flag

oxaudo:
This PR is the last code change to resolve [EMI-2139]. 

### Description

This is the third pr to clean up tasks to remove feature flags for signals works. Echo PR is coming next.

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- remove AREnableAuctionImprovementsSignals feature flag - oxaudo

</details>


src/app/Components/ArtworkGrids/ArtworkGridItem.tsx
```diff
@@ -352,7 +342,7 @@ export const Artwork: React.FC<ArtworkProps> = ({
                     <Text variant="xs" numberOfLines={1} caps {...lotLabelTextStyle}>
                       Lot {artwork.saleArtwork.lotLabel}
                     </Text>
-                    {!!artwork.sale?.cascadingEndTimeIntervalMinutes && !displayAuctionSignal && (
+                    {!!artwork.sale?.cascadingEndTimeIntervalMinutes && !isAuction && (
```
oxaudo, line 345: This condition is a bit unclear to me. I don't think there is a possibility for work to have `cascadingEndTimeIntervalMinutes` and not be in an auction? But this logic was here before so I assume I'm just missing something?

ArtsyOpenSource:

- Dev changes (remove AREnableAuctionImprovementsSignals feature flag - oxaudo - oxaudo)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against ac5d42cf280ed23153458d3abaddb7667f8a2fb0
</p>

oxaudo:
> Looks good!
> 
> Not sure about the `isAuction` change in the grid element though, it makes sense but I don't have context to assure it's 100% fine that change, the rest looks 👍

I believe that is not a change as it was checking for 'isAuction + feature flag' before and feature flag is enabled across the board. So I kept it to not change anything but I actually do not understand the intent of this check. Maybe @erikdstock or @starsirius (I believe current auctions pros on the team) can chime in here?

src/app/Components/ArtworkGrids/ArtworkGridItem.tsx
```diff
@@ -303,11 +297,7 @@ export const Artwork: React.FC<ArtworkProps> = ({
                 />
 
                 {Boolean(
```
MrSltun, line 299: We would need to remove this whole component, since it was the one used before we introduced the auction improvements to the artwork grid items/rail cards

src/app/Components/ArtworkRail/ArtworkSaleMessage.tsx
```diff
@@ -22,13 +21,11 @@ export const ArtworkSaleMessage: React.FC<ArtworkSaleMessageProps> = ({
...
   dark = false,
 }) => {
-  const enableAuctionImprovementsSignals = useFeatureFlag("AREnableAuctionImprovementsSignals")
-
   const { collectorSignals, sale } = useFragment(fragment, artwork)
 
   const { primaryTextColor } = useMetaDataTextColor({ dark })
 
-  const displayAuctionSignal = enableAuctionImprovementsSignals && sale?.isAuction
+  const displayAuctionSignal = sale?.isAuction
```
MrSltun, line 28: I think this is more complicated thank just removing the mention of the feature flag 👀 will look into that more tomorrow

src/app/Components/ArtworkGrids/ArtworkGridItem.tsx
oxaudo: Do you mean that the whole `urgencyTag` is not used any more? I do not see how/what it's replaced with?

src/app/Components/ArtworkRail/ArtworkSaleMessage.tsx
oxaudo: I thought about dropping `displayAuctionSignal` and use `sale?.isAuction` where applicable directly but as it's used more than once and this is a local variable figured it's still helpful to store and reuse it.

oxaudo: Actually looking at this again - you have a point - it's used  only in a context of determining if `liveBiddingStarted` - I'll update to a  better name.

src/app/Components/ArtworkGrids/ArtworkGridItem.tsx
MrSltun: Yeah I believe urgency tags were the text that would show up on top of the artwork images when the auction has started/close to start/ending, so the new collector signals for auctions are replacing that, so after we released the auction improvements, we replaced the urgency tags with the new auction signals

src/app/Components/ArtworkRail/ArtworkSaleMessage.tsx
MrSltun: Yes this looks better now :) but it was right to just remove the feature flag 😅 thank you for renaming the variable

src/app/Components/ArtworkGrids/ArtworkGridItem.tsx
MrSltun: And with that, we do have a prop for the components called `urgencyTag` that we would also remove since it won't be used anymore, but we would need to double check

oxaudo:
I'm tempted to merge this as is and add a card to see if we can remove `urgencyTag` and take it next. What you think @MrSltun ?

MrSltun:
@oxaudo I believe that the removal of the feature flag should include the removal of the `urgencyTag`, since I don't know how broken the experience will be if the urgency tag still shows up with the new auction-related collector signals

oxaudo:
@MrSltun - this `urgencyTag` is also used in `ContextMenuArtworkPreviewCard` and removing it causing some more cascading code deletion. I realize I'm not sure where this `ContextMenuArtworkPreviewCard` is even used. Looks like it's rendered from `contextMenuArtwork` which seems to somehow do with caching.

I can totally remove this `urgencyTag` from there as well but first feel like it would be better in a separate PR for potential roll back and also that removal of the `enableAuctionImprovementsSignals` should not matter here. As this flag is set to true across the board and checking it will always evaluate to true which is the same as not checking it at all.

MrSltun (approved):
I think this looks good now! But we need to test to make sure that nothing broke due to these changes

Thank you for powering through Oksana! 🌟

oxaudo:
Ok. Added one more removal. Wonder if we should merge this after we release tomorrow and then maybe even do mocktion just to be on the safe side that we are not messing up things here.

I feel like it's been quite a big chunk of removals. Makes me uneasy :)

MrSltun (approved):

PR #10947, opened 2024-10-11
feat: bump ShowCard image size from large to larger

dblandin:
### Description

Our Image components often trigger on-the-fly image resizing via Gemini and this larger image should be a better original source to crop/resize from.

Follow up to #10944.

### PR Checklist

- [ ] I have tested my changes on **iOS** and **Android**.
- [ ] I hid my changes behind a **[feature flag]**, or they don't need one.
- [ ] I have included **screenshots** or **videos**, or I have not changed the UI.
- [ ] I have added **tests**, or my changes don't require any.
- [ ] I added an **[app state migration]**, or my changes do not require one.
- [ ] I have documented any **follow-up work** that this PR will require, or it does not require any.
- [ ] I have added a **changelog entry** below, or my changes do not require one.

### To the reviewers 👀

- [ ] I would like **at least one** of the reviewers to **run** this PR on the simulator or device.

<details><summary>Changelog updates</summary>

#### Cross-platform user-facing changes

- feat: bump ShowCard image size from large to larger

#### iOS user-facing changes

-

#### Android user-facing changes

-

#### Dev changes

-

</details>

Need help with something? Have a look at our [docs], or get in touch with us.

[app state migration]: ../blob/main/docs/adding_state_migrations.md
[feature flag]: ../blob/main/docs/developing_a_feature.md
[docs]: ../blob/main/docs/README.md

PR #11028, opened 2024-10-28
fix(MyC): use better artwork image version

anandaroop:
This PR resolves the QA launch-blocker https://www.notion.so/artsy/Artworks-in-my-collection-are-displayed-incorrectly-zoomed-in-12acab0764a080d3be3ac407fa0a99d7

### Description

This was marked as a launch-blocker (though it was also noted that this was [not a new issue](https://www.notion.so/Artworks-in-my-collection-are-displayed-incorrectly-zoomed-in-12acab0764a080d3be3ac407fa0a99d7?d=12acab0764a080a380d3001c3d54bd69&pvs=4)):

> [!IMPORTANT]
> I am not 100% sure this is a valid fix within the MyC context, see comments below…

| Before | After |
|--------|--------|
|  |  | 

### PR Checklist

### To the reviewers 👀

- Use better image crop for My Collection artworks

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Scenes/MyCollection/Screens/ArtworkList/MyCollectionArtworkGridItem.tsx
+++ b/src/app/Scenes/MyCollection/Screens/ArtworkList/MyCollectionArtworkGridItem.tsx
@@ -169,7 +169,7 @@ export const MyCollectionArtworkGridItemFragmentContainer = createFragmentContai
           name
         }
         images(includeAll: $includeAllImages) {
-          url
+          url(version: "large")
           isDefault
           internalID
           versions


src/app/Scenes/MyCollection/Screens/ArtworkList/MyCollectionArtworkGridItem.tsx
```diff
@@ -169,7 +169,7 @@ export const MyCollectionArtworkGridItemFragmentContainer = createFragmentContai
           name
         }
         images(includeAll: $includeAllImages) {
-          url
+          url(version: "large")
```
anandaroop, line 172: This brings the MyC-specific artwork grid component into agreement with the regular artwork grid component:

https://github.com/artsy/eigen/blob/56bc08526076e10f0a73f47f320f1add5e7b8683/src/app/Components/ArtworkGrids/ArtworkGridItem.tsx#L554

Without specifying a version, I think we get an indeterminate artwork image version, which when I examined these in the simulator turned out to be `large_rectangle`. This would certainly explain the weird crops!

| large_rectangle | large |
|--------|--------|
|   |  |

anandaroop: I do however recall some complications around image processing for MyCollection so I'm not certain if we can rely on `large` always being available. Lmk if you think there's any reason not to roll with this change.

MounirDhahri: Thanks for this fix! i would just suggest to try out `larger` here instead to avoid on-the-fly image resizing as we did here https://github.com/artsy/eigen/pull/10947

dariakoko (commented):
On ArtworkGridItem.tsx we specify the image version `url(version: "large") `so, i guess, your solution would be the one to go with

However I tried using imageURL instead of url in MyCollectionArtworkGridItemFragmentContainer and it solved the issue
Also, we use imageURL on the Artwork screen. Do you know if there was a reason to use `url` on the grids? I guess, url returns a different size of the image and it might be too hight quality to render on the grid?

src/app/Scenes/MyCollection/Screens/ArtworkList/MyCollectionArtworkGridItem.tsx
dblandin: `larger` should be a good size to start with and important to be explicit with that desired size here. I expect that we're still executing on-the-fly resizing under the hood within the OpaqueImageView component [here](https://github.com/artsy/eigen/blob/main/src/app/Components/OpaqueImageView2/OpaqueImageView2.tsx#L183-L190).

anandaroop: Good call. I had to remind myself (again!) of the various pre-cut image versions over in Gemini:

```ruby
Account.find_by(name: "Gravity").templates.find_by(key: "additional-image").versions.pluck(:key,
 :resize_to, :width, :height, :convert_to, :quality).sort

[["large", "limit", 640, 640, "jpg", 90],          # 👈🏽 current
 ["large_rectangle", "fill", 640, 480, "jpg", 90],
 ["larger", "limit", 1024, 1024, "jpg", 81],       # 👈🏽 proposed
 ["main", "limit", 5120, 5120, "jpg", 90],
 ["medium", "height", nil, 260, "jpg", 90],
 ["medium_rectangle", "fill", 260, 195, "jpg", 90],
 ["normalized", nil, nil, nil, "jpg", 81],
 ["small", "limit", 200, 200, "jpg", 70],
 ["square", "fill", 230, 230, "jpg", 90],
 ["tall", "limit", 260, 800, "jpg", 90]]
```

Most of these (i.e. where `resize_to ≠ fill`) should preserve aspect ratio. And `square` happens on the [highest-priority](https://github.com/artsy/gemini/blob/db4ece0a05bc8b6472baa10e884af2668eb7d15a/app/models/entry_version.rb#L47) queue so is most likely to be available as a fallback.

So I think the preferred order should be: [`larger`, `large`, `medium`, `small`, `square`]

And I think we should do it for both grid components in this case (regular and MyC).

anandaroop: Modified in https://github.com/artsy/eigen/pull/11028/commits/0a91436bd702bf2b3751f224f52d94349fe9c517 (and also https://github.com/artsy/eigen/pull/11028/commits/c197ca39b437e1a0eb803f62561ff781e4742c13 to keep tsc happy)

anandaroop:
> I tried using imageURL instead of url [...] Do you know if there was a reason to use url on the grids

Looks like imageURL would return `square` by default which would be fairly low-res especially on high DPI screens, so I think the current solution is better, especially if these are being resized further by their image components.

And FYI there is a [plan on the horizon](https://www.notion.so/artsy/All-the-migrations-1dbb761c58944fa1ba3d78bf52c2a4d7?p=85052f249d4d48dcb97cc08faa3f54bb&pm=s) to deprecate Gemini image versions in favor of always doing on-the-fly resizing from a high-res original, but that's still very preliminary, so I'm basically ignoring that here.

dblandin:
> And FYI there is a [plan on the horizon](https://www.notion.so/artsy/All-the-migrations-1dbb761c58944fa1ba3d78bf52c2a4d7?p=85052f249d4d48dcb97cc08faa3f54bb&pm=s) to deprecate Gemini image versions in favor of always doing on-the-fly resizing from a high-res original, but that's still very preliminary, so I'm basically ignoring that here.

FWIW, I think that's essentially what this implementation would enable by placing `larger` as the first preferred image version and then within the underlying Image component, executing a separate request for an on-the-fly resized image. From my understanding `normalized` and `larger` image versions will likely stick around with `larger` being the preferred image version to execute on-the-fly resizing requests against. All others would be on the chopping block to deprecate+remove.

dblandin (approved):

anandaroop:
I think the goal _eventually_ would be to always resize artwork images from `main`, which is by modern standards sufficiently large (up to 5120x5120 or 26 megapixels) for all purposes except deep zoom. Resized versions would be cached at Cloudfront, so only expensive on the first request per image, and always tailored for the use case in question. We'll get there!

PR #11031, opened 2024-10-29
feat(EMI-2146): Add Task onboarding animation

MrSltun:
This PR resolves [EMI-2146] 

### Description
This PR adds an onboarding animation for the Act Now section (or tasks) in the home view
It also requires adding a patch for `react-native-gesture-handler/ReanimatedSwipeable`

#### Why patching?
The patch is for a calculation for the `openRight` function since it doesn't work on the first try
There's a PR on the package that fixes the issue (https://github.com/software-mansion/react-native-gesture-handler/pull/3149) but it doesn't seem to be added on the latest release, so we will be removing the patch when the fix gets released

### Demo

| Android | iOS |
|---|---|
| <video src="https://github.com/user-attachments/assets/db4ec509-0482-4800-b6c1-ad232c1de5d8" /> | <video src="https://github.com/user-attachments/assets/72129b91-d447-49c6-a7d8-b444d87f2ad2" /> | 

### PR Checklist

### To the reviewers 👀

- Add onboarding animation for Act Now (tasks) section in the home view

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[EMI-2146]: https://artsyproduct.atlassian.net/browse/EMI-2146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ArtsyOpenSource:

- Cross-platform user-facing changes (Add onboarding animation for Act Now (tasks) section in the home view - MrSltun)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 7f88e116ef1b92c0f08151a11502b117134746cd
</p>

patches/react-native-gesture-handler+2.19.0.patch
```diff
@@ -0,0 +1,13 @@
...
++++ b/node_modules/react-native-gesture-handler/src/components/ReanimatedSwipeable.tsx
+@@ -444,7 +444,7 @@ const Swipeable = forwardRef<SwipeableMethods, SwipeableProps>(
+       openRight() {
+         'worklet';
+         rightWidth.value = rowWidth.value - rightOffset.value;
+-        animateRow(calculateCurrentOffset(), -rightWidth.value);
++        animateRow(calculateCurrentOffset(), -(rowWidth.value - rightOffset.value));
+       },
+       reset() {
+         'worklet';
```
gkartalis, line 13: is this released or not yet??

src/app/Scenes/HomeView/Sections/HomeViewSectionTasks.tsx
```diff
@@ -26,29 +28,55 @@ export const HomeViewSectionTasks: React.FC<HomeViewSectionTasksProps> = ({
...
+  const startOnboardingAnimation = () => {
+    InteractionManager.runAfterInteractions(() => {
+      swipeableRef.current?.openRight()
+    }).then(() => {
+      setTimeout(() => {
+        swipeableRef.current?.close()
+        dismiss("act-now-tasks")
+      }, 1500)
+    })
+  }
```
gkartalis, line 64: **question:**  What happens when the user logs in and has some saved artworks?

Asking for how it is performing when it does multiple things on the home view like:
- you end up in home page
- we ask for receiving notifications
- we display the popover for the saved artworks 
and now we have this new animation

gkartalis (approved):
left some questions but other than that nice one ;)

patches/react-native-gesture-handler+2.19.0.patch
MrSltun: No, I don't believe it has been released, it has been merged but not released yet.
I already checked the commits on the last version of `react-native-gesture-handler`

src/app/Scenes/HomeView/Sections/HomeViewSectionTasks.tsx
MrSltun: Good question, I'll test that scenario, but I assume that we it would animate everything at once 😬

brainbicycle (dismissed):
looks slick!

patches/react-native-gesture-handler+2.19.0.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node_modules/react-native-gesture-handler/src/components/ReanimatedSwipeable.tsx b/node_modules/react-native-gesture-handler/src/components/ReanimatedSwipeable.tsx
```
damassi, line 1: Wanted to start a discussion about this in Slack: https://artsy.slack.com/archives/C02JHHHKP5K/p1730213514746509

Patches should be regarded as last resort, ideally 👀

damassi: (talked in slack, all good)

src/app/Components/Tasks/Task.tsx
```diff
@@ -19,84 +19,79 @@ interface TaskProps {
...
-  ...restProps
-}) => {
-  const { tappedNotification, tappedClearNotification } = useHomeViewTracking()
-  const { submitMutation: dismissTask } = useDismissTask()
-  const { submitMutation: acknowledgeTask } = useAcknowledgeTask()
-  const fontScale = PixelRatio.getFontScale()
+export const Task = forwardRef<SwipeableMethods, TaskProps>(
+  ({ disableSwipeable, onClearTask, onPress, ...restProps }, ref) => {
+    const { tappedNotification, tappedClearNotification } = useHomeViewTracking()
+    const { submitMutation: dismissTask } = useDismissTask()
```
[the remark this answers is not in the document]
MrSltun: I noticed that we have different ways of committing mutations, maybe we should agree on one convention and apply it to all the mutation hooks in Eigen?

src/app/Scenes/HomeView/Sections/HomeViewSectionTasks.tsx
MrSltun: It does break 👀 
I just fixed it, and now it shows after each other

MrSltun: please excuse the slowness, it's a debug on a physical android 😅
<video src="https://github.com/user-attachments/assets/9a6a5d78-f72a-4b1d-80f4-de9a891e70df" />

src/app/Scenes/HomeView/Sections/HomeViewSectionTasks.tsx
```diff
@@ -26,29 +28,55 @@ export const HomeViewSectionTasks: React.FC<HomeViewSectionTasksProps> = ({
 
   const tasks = extractNodes(section.tasksConnection)
 
+  const swipeableRef = useRef<SwipeableMethods>(null)
+  const {
+    isDismissed,
+    sessionState: { isReady },
+  } = GlobalStore.useAppState((state) => state.progressiveOnboarding)
+  const { dismiss } = GlobalStore.actions.progressiveOnboarding
```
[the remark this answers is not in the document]
gkartalis: I would love to see this removed actually 😅 I think the only reason that @MrSltun referenced it in this pr is to work around the actual popover

src/app/Scenes/HomeView/Sections/HomeViewSectionTasks.tsx
gkartalis: noice! Thanks for addressing + testing + recording 👏

src/app/Components/Tasks/Task.tsx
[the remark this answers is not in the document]
MrSltun: I'll add it as follow-up PR if that's okay :)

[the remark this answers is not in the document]
damassi: Don't mean to jump into the mix, but this is one of the plusses of using the [relay version of the useMutation hook](https://relay.dev/docs/api-reference/use-mutation/) -- it returns a tuple, and you can name the variable however:

```tsx
const [commit] = useMutation(...) 
```

Adapting it to our patterns, we'd do something like:

```tsx
const useCreateAlert = () => {
  return useMutation(...)
}
```

Which allows us to use it like:

```tsx
const [createAlert] = useCreateAlert() 
```

@MrSltun and @araujobarret - But aside from ☝️ being more ergonomic, the way things are done in this PR  are exactly idiomatic with all of our other codebases and mutation patterns Artsy-wide and I would argue we should keep it (vs defining one-off hooks for mutations with their own custom return values, breaking convention, vs very simple aliasing. 

Like, this is the hook as its currently being used:

```tsx
export const useDismissTask = () => {
  return useMutation<useDismissTaskMutation>({
    mutation: DismissTaskMutation,
  })
}
```

What it returns is an interface defined by our custom `useMutation` hook. If we need to change something there, it applies _everywhere_ -- this is good. If we start breaking that convention and writing one-off hooks like:

```tsx
const useDismissTask = () => {
  const { submitMutation } = useMutation(...)
  return {
    dismissTask: submitMutation
  }
}
```

We're actually just increasing complexity because now we're just pushing the aliasing up into the hook and away from the call-site, and adding extra loc. 

But the main thing is:

```tsx
const { submitMutation: dismissTask } = useDismissTask() 
```

Provides a lot of signal to future devs who didn't write code, reading the code -- they know that a) we're dealing with a mutation, not a function call -- the return value -- and b) that we're dismissing a task. All of that signal appears at a glance right near the invocation site. The API here was very deliberate in this way.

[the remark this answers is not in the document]
damassi: Totally -- just wanted to note that 
- returning `submitMutation` was deliberate (giving the option to alias -- if needed) and 
- that we should probably try to use the relay `useMutation` hook more and deprecate the old one, if we do anything. 

I know you (@araujobarret) were the first one to spearhead that in volt2! It makes everything cleaner, and while simultaneously leaning on relay and its subtle optimizations and best practices 👍

PR #11052, opened 2024-11-01
fix: logout redirecting to profile in dev

brainbicycle:
This PR resolves [] 

### Description

Our dev tool to reload the nav state was not clearing the state on logout, this does that.
Also adds missed Podfile.lock update from earlier pr.

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- fix logout redirecting to profile in dev

</details>

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -1415,7 +1415,7 @@ PODS:
     - React-Core
   - RNCClipboard (1.14.1):
     - React-Core
-  - RNDeviceInfo (10.3.0):
+  - RNDeviceInfo (14.0.0):
     - React-Core
   - RNFastImage (8.6.3):
     - React-Core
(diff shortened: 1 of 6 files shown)


ArtsyOpenSource:

- Dev changes (fix logout redirecting to profile in dev - brainbicycle)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 09ce59db1bb43e23bcafd5f3f0ae4c4dff4876d6
</p>

iskounen (commented):
1. I've only observed this on Android, so I was expecting an Android-specific solution. Was it happening on iOS as well?
2. How were you able to determine that `clearNavState` only needed to be cleared for dev environments?

brainbicycle:
> 1. I've only observed this on Android, so I was expecting an Android-specific solution. Was it happening on iOS as well?
> 2. How were you able to determine that `clearNavState` only needed to be cleared for dev environments?

On 1. I would expect them both to be affected but I can double check, in either case we don't want to restore nav after logout on either platform so i don't think we need a platform specific solution. 

For 2. I pulled down the beta from firebase and was unable to reproduce, running from main I was able to, the bug is tied to the useReloadedDevNavigationState hook which is only used in dev.

brainbicycle:
> 1. I've only observed this on Android, so I was expecting an Android-specific solution. Was it happening on iOS as well?
> 2. How were you able to determine that `clearNavState` only needed to be cleared for dev environments?

It doesn't affect iOS because we are clearing AsyncStorage via a native module: https://github.com/artsy/eigen/blob/862b469544a3982167f3e219eb9844ca1b726888/ios/Artsy/Networking/API_Modules/ARUserManager.m#L187

brainbicycle:
going to merge this one, let me know if any concerns!

iskounen:
@brainbicycle sorry for the late response. Thanks for explaining how you debugged this. All good from my side 👍

PR #11070, opened 2024-11-07
fix: remove @cacheable from non cacheable requests

MounirDhahri:
### Description

This PR comes to address the comments during the QA session here https://www.notion.so/artsy/2024-10-10-Mobile-Feature-QA-Metaphysics-CDN-and-cacheable-directive-11bcab0764a08097bd33cc1179d4d6a9

Here: I am removing the `@cacheable` flag from any query that is requesting content that is not cacheable. This includes for example:
- Shows for You rail
- Queries containing artwork grids which display (`isSaved`)

### PR Checklist

### To the reviewers 👀

- 

#### iOS user-facing changes

-

-

#### Dev changes

- remove @cacheable from non cacheable requests - mounir

</details>

### DIFF
--- a/src/app/Scenes/Artist/Artist.tsx
+++ b/src/app/Scenes/Artist/Artist.tsx
@@ -199,7 +199,7 @@ interface ArtistQueryRendererProps {
 }
 
 export const ArtistScreenQuery = graphql`
-  query ArtistAboveTheFoldQuery($artistID: String!, $input: FilterArtworksInput) @cacheable {
+  query ArtistAboveTheFoldQuery($artistID: String!, $input: FilterArtworksInput) {
     artist(id: $artistID) @principalField {
       ...ArtistHeader_artist
       ...ArtistArtworks_artist @arguments(input: $input)
--- a/src/app/Scenes/HomeView/HomeView.tsx
+++ b/src/app/Scenes/HomeView/HomeView.tsx
@@ -211,7 +211,7 @@ const sectionsFragment = graphql`
 `
 
 export const homeViewScreenQuery = graphql`
-  query HomeViewQuery($count: Int!, $cursor: String) @cacheable {
+  query HomeViewQuery($count: Int!, $cursor: String) {
     homeView {
       experiments {
         name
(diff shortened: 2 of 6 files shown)


ArtsyOpenSource:

- Dev changes (remove @cacheable from non cacheable requests - mounir - MounirDhahri)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against dd89333d576b708823d5fb6c9272029a1bb5f305
</p>

src/app/Scenes/Artist/Artist.tsx
```diff
@@ -199,7 +199,7 @@ interface ArtistQueryRendererProps {
 }
 
 export const ArtistScreenQuery = graphql`
-  query ArtistAboveTheFoldQuery($artistID: String!, $input: FilterArtworksInput) @cacheable {
+  query ArtistAboveTheFoldQuery($artistID: String!, $input: FilterArtworksInput) {
```
brainbicycle, line 202: i am starting to feel we need some backend solution for identifying which queries are personalized and which are not, it seems pretty easy right now to get this wrong

brainbicycle (approved):
nice!

MounirDhahri:
@MrSltun currently the `@cacheable` is not being used anywhere so this changes shouldn't have any effect afaik

src/app/Scenes/Artist/Artist.tsx
gkartalis: also agree with this, pretty tricky to figure this out quickly and people might spent too much time trying to figure it out

mzikherman: there's middleware that will strip out the access token from the headers used in making backend requests, so there shouldn't ever be a bug where we cache your data for someone else's. There is no such thing as 'your' data b/c the access token is removed.

I'm wondering what people think might be wrong if it's errantly used and how we can improve the experience? If used incorrectly, it would mean that personalized content simply _never_ shows up which I think should be obvious to the dev, but I also agree can be sneaky - for instance `isSaved: false` comes back for everything and you don't realize the save functionality/display is borked?

(as an aside, would be nice to move saves into their own query so grids/rails can be cached, we did that on Force in https://github.com/artsy/force/pull/14533)

PR #11097, opened 2024-11-14
fix: avoid prefetching when following deeplinks

anandaroop:
This PR resolves [ONYX-1392] 

Co-authored-by: @brainbicycle 
Co-authored-by: @MounirDhahri 
Co-authored-by: @olerichter00 

### Description

The goal here is to avoid prefetching when following a deep link, in order to avoid an unnecessary [stampede](https://artsy.slack.com/archives/CA8SANW3W/p1731532192977049?thread_ts=1731531969.773739&cid=CA8SANW3W) that results in API latency.

We observe that prefetching happens in a few ways:

1. A lighter home view prefetch is initiated from the splash screen
    - can be avoided by checking if we are following a deep link
2. A heavier prefetch for a subset of home view sections’ content
    - can also be avoided by declining to render the home view _at all_ when launching via deep link
2. Content for other tabs is initiated from the home view
    - can be avoided thanks to the previous point

### After this change

| Case | Behavior | Screencap |
|--------|--------|--------|
| Launching the app normally | **Performs** prefetching, as before | <video alt="regular launch" src="https://github.com/user-attachments/assets/f31c8ab6-2cb5-40de-940f-f753098fd709" /> |
| Launching via deep link | **Avoids** prefetching | <video alt="deep link" src="https://github.com/user-attachments/assets/c0644b54-ed09-4702-9b63-c4a691adde79" /> |
| Launching via deep link, then returning to home | **Fetches the content just-in-time** rather than prefetch. This is the necessary de-optimization that allows us to avoid the stampede of API requests | <video alt="deep link then home" src="https://github.com/user-attachments/assets/c3d3e2b6-13dc-47ea-a7ff-a96b459eba2c" /> | 
| Launching via link to `/` | **Performs** prefetching, as before | <video alt="home link" src="https://github.com/user-attachments/assets/3e30408d-58c7-456b-902f-d0814249a2e7" /> |

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- Avoids prefetching home view content when opening a deep link

</details>

[ONYX-1392]: https://artsyproduct.atlassian.net/browse/ONYX-1392?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/utils/hooks/useIsDeepLink.tests.ts
+++ b/src/app/utils/hooks/useIsDeepLink.tests.ts
@@ -1,8 +1,8 @@
 import { renderHook } from "@testing-library/react-hooks"
+import { matchRoute } from "app/routes"
 import { useIsDeepLink } from "app/utils/hooks/useIsDeepLink"
 import { Linking } from "react-native"
 
-// Existing mocks
 const mockUseIsFocusedMock = jest.fn()
 jest.mock("@react-navigation/native", () => ({
   useIsFocused: () => mockUseIsFocusedMock(),
@@ -14,13 +14,18 @@ jest.mock("react-native", () => ({
   },
 }))
 
-// Start of the new test case
+jest.mock("app/routes", () => ({
+  matchRoute: jest.fn(),
+}))
+
 describe("useIsDeepLink", () => {
   const mockLinkingGetInitialURL = Linking.getInitialURL as jest.Mock
+  const mockMatchRoute = matchRoute as jest.Mock
 
   it("should return true if opened from a deep link", async () => {
     // Setup the mock to return the specific URL
-    mockLinkingGetInitialURL.mockResolvedValue("artsy:///foo/bar")
+    mockLinkingGetInitialURL.mockResolvedValue("artsy:///artwork/foo")
+    mockMatchRoute.mockReturnValue({ type: "match", module: "Artwork" })
     mockUseIsFocusedMock.mockReturnValue(true)
 
     // Render the hook under test
@@ -34,6 +39,7 @@ describe("useIsDeepLink", () => {
     })
     expect(mockUseIsFocusedMock).toHaveBeenCalled()
     expect(mockLinkingGetInitialURL).toHaveBeenCalled()
+    expect(mockMatchRoute).toHaveBeenCalled()
   })
 
   it("should return false if not opened from a deep link", async () => {
@@ -50,11 +56,13 @@ describe("useIsDeepLink", () => {
     expect(result.current.isDeepLink).toEqual(false)
     expect(mockUseIsFocusedMock).toHaveBeenCalled()
     expect(mockLinkingGetInitialURL).toHaveBeenCalled()
+    expect(mockMatchRoute).toHaveBeenCalled()
   })
 
   it("should return false if opened from a link to /", async () => {
     // Setup the mock to return null
     mockLinkingGetInitialURL.mockResolvedValue("artsy:///")
+    mockMatchRoute.mockReturnValue({ type: "match", module: "Home" })
     mockUseIsFocusedMock.mockReturnValue(true)
 
     // Render the hook under test
@@ -66,5 +74,6 @@ describe("useIsDeepLink", () => {
     expect(result.current.isDeepLink).toEqual(false)
     expect(mockUseIsFocusedMock).toHaveBeenCalled()
     expect(mockLinkingGetInitialURL).toHaveBeenCalled()
+    expect(mockMatchRoute).toHaveBeenCalled()
   })
 })
--- a/src/app/utils/hooks/useIsDeepLink.tests.ts
+++ b/src/app/utils/hooks/useIsDeepLink.tests.ts
@@ -1,0 +1,54 @@
+import { renderHook } from "@testing-library/react-hooks"
+import { useIsDeepLink } from "app/utils/hooks/useIsDeepLink"
+import { Linking } from "react-native"
+
+// Existing mocks
+const mockUseIsFocusedMock = jest.fn()
+jest.mock("@react-navigation/native", () => ({
+  useIsFocused: () => mockUseIsFocusedMock(),
+}))
+
+jest.mock("react-native", () => ({
+  Linking: {
+    getInitialURL: jest.fn(),
+  },
+}))
+
+// Start of the new test case
+describe("useIsDeepLink", () => {
+  const mockLinkingGetInitialURL = Linking.getInitialURL as jest.Mock
+
+  it("should return true if opened from a deep link", async () => {
+    // Setup the mock to return the specific URL
+    mockLinkingGetInitialURL.mockResolvedValue("artst.net/foo/bar")
+    mockUseIsFocusedMock.mockReturnValue(true)
+
+    // Render the hook under test
+    const { result, waitForNextUpdate } = renderHook(() => useIsDeepLink())
+
+    // Wait for async effects to resolve
+    await waitForNextUpdate()
+
+    expect(result.current).toEqual({
+      isDeepLink: true,
+    })
+    expect(mockUseIsFocusedMock).toHaveBeenCalled()
+    expect(mockLinkingGetInitialURL).toHaveBeenCalled()
+  })
+
+  it("should return false if not opened from a deep link", async () => {
+    // Setup the mock to return null
+    mockLinkingGetInitialURL.mockResolvedValue(null)
+    mockUseIsFocusedMock.mockReturnValue(true)
+
+    // Render the hook under test
+    const { result, waitForNextUpdate } = renderHook(() => useIsDeepLink())
+
+    // Wait for async effects to resolve
+    await waitForNextUpdate()
+
+    expect(result.current.isDeepLink).toEqual(false)
+    expect(mockUseIsFocusedMock).toHaveBeenCalled()
+    expect(mockLinkingGetInitialURL).toHaveBeenCalled()
+  })
+})
--- a/src/app/utils/hooks/useIsDeepLink.tests.ts
+++ b/src/app/utils/hooks/useIsDeepLink.tests.ts
@@ -20,7 +20,7 @@ describe("useIsDeepLink", () => {
 
   it("should return true if opened from a deep link", async () => {
     // Setup the mock to return the specific URL
-    mockLinkingGetInitialURL.mockResolvedValue("artst.net/foo/bar")
+    mockLinkingGetInitialURL.mockResolvedValue("artsy:///foo/bar")
     mockUseIsFocusedMock.mockReturnValue(true)
 
     // Render the hook under test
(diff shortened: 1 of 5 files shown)


MounirDhahri (approved):
Nice! thanks for the fix

MounirDhahri:
| Scenario | Vide |
|--------|--------|
| Opened App w/o deep link | <Video src="https://github.com/user-attachments/assets/1875e3f7-afe4-46df-9205-7b749449f7ac" /> |
| Opened App from deep link | <Video src="https://github.com/user-attachments/assets/2ccbe4be-a25c-4951-aa94-dbb538453909" /> |

src/app/Scenes/Home/HomeContainer.tsx
```diff
@@ -32,6 +35,12 @@ export const InnerHomeContainer = () => {
     }
   }, [artQuizState, navigateToArtQuiz, isNavigationReady])
 
+  // We want to avoid rendering the home view when the user comes back from a deep link
+  // Because it triggers a lot of queries that affect the user's experience and can be avoided
+  if (isDeepLink !== false) {
```
brainbicycle, line 40: I wonder if we need a check if the deeplinks destination is the homeview? Otherwise this might lead to home view not rendering in this case?

src/app/utils/hooks/useIsDeepLink.tests.ts
```diff
@@ -0,0 +1,54 @@
...
+  useIsFocused: () => mockUseIsFocusedMock(),
+}))
+
+jest.mock("react-native", () => ({
+  Linking: {
+    getInitialURL: jest.fn(),
+  },
+}))
+
+// Start of the new test case
```
olerichter00, line 17: 👀

src/app/utils/hooks/useIsDeepLink.tests.ts
```diff
@@ -0,0 +1,54 @@
+import { renderHook } from "@testing-library/react-hooks"
+import { useIsDeepLink } from "app/utils/hooks/useIsDeepLink"
+import { Linking } from "react-native"
+
+// Existing mocks
```
olerichter00, line 5: 👀

src/app/Scenes/Home/HomeContainer.tsx
olerichter00: `useIsDeepLink` could be renamed to `useDeepLinkURL` and return the URL of the deep link or `null` to make this check possible.

MounirDhahri: good thought! that makes sense

olerichter00: Instead of `null`, the hook could return `undefined` when waiting for the async call to get the deep link URL (`undefined`: declared but has not yet been assigned a value). This way, we can use `null` when there is no deep link URL (`null`: as a representation of no value) and a string to represent the URL when it exists.

anandaroop: Mobbed on this in Onyx KS today and settled on https://github.com/artsy/eigen/pull/11097/commits/d151919e6ce130302e90f466c329c134d7219421

ArtsyOpenSource:

- Dev changes (Avoids prefetching home view content when opening a deep link - anandaroop)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against d151919e6ce130302e90f466c329c134d7219421
</p>

src/app/utils/hooks/useIsDeepLink.ts
```diff
@@ -0,0 +1,52 @@
...
+  useEffect(() => {
+    Linking.getInitialURL()
+      .then((url) => {
+        if (!url) {
+          setIsDeepLink(false)
+          return
+        }
+
+        const targetURL = url.replace("artsy://", "")
+        const isDeepLink = targetURL !== "/"
```
brainbicycle, line 35: I think this is on the right track but the check is a little too naive still, because we handle universal links the link will not always start with `artsy://`. For example if you tap a link on your mobile device from any of our associated domains: 
```
 https://www.artsy.net
- https://artsy.net
- https://click.artsy.net
- https://m.artsy.net
- https://link.artsy.net
- https://email-link.artsy.net
```
This will open the app, the initial url will be `https://artsy.net` and will still resolve to the home view and then not render. 

One approach might be to reuse the matchRoute function in routes.ts we use for handling deeplinks and see what the route resolves to rather than trying to figure it out here.

brainbicycle (changes requested):
i think we need a little more nuance in our home deeplink detection, happy to pair tomorrow if it would be helpful!

src/app/utils/hooks/useIsDeepLink.ts
anandaroop: Great feedback. When we discussed today, there was some concern about coupling this to a specific module e.g. `HomeView`, but sounds like maybe that _is_ actually what's necessary here. 

Yes, happy to pair!

anandaroop:
Brian and I paired on switching the detection back to `matchRoute` in https://github.com/artsy/eigen/pull/11097/commits/4bf5896a78a27a51f85964ae7503cecbb9fcd720 since it will be more reliable in the edge cases noted above. (h/t @MounirDhahri for suggesting that originally)

I confirmed that the screengrabs above still work the same after these changes, and we also ran through the various cases in this gist, which may also come in handy at Mobile QA time:
- https://gist.github.com/anandaroop/8903e16ced8f962ff1fea60e13266584

brainbicycle (dismissed):
thank you!

src/app/utils/hooks/useIsDeepLink.ts
```diff
@@ -0,0 +1,55 @@
...
+import { useEffect, useState } from "react"
+import { Linking } from "react-native"
+
+/**
+ * This is a hook that returns whether or not the user came from a deep link
+ * (defined as a direct navigation to a route other than "/").
+ *
+ * This can be used to avoid rendering content in previous screens in react-navigation history
+ *
+ * @returns {isDeepLink: boolean | null} isDeepLink is true if the user came from a deep link
```
olerichter00, line 12: I'd add something like this to the description to make sure the hook is used correctly. (`null` vs `false` value).

```suggestion
 * @returns {isDeepLink: boolean | null}` isDeepLink` is true if the user came from a deep link. Initially, it is set to `null` while retrieving the deep link URL asynchronously and will be set to `false` if retrieving the `URL` fails.
```

src/app/utils/hooks/useIsDeepLink.ts
```diff
@@ -0,0 +1,55 @@
...
+ */
+export const useIsDeepLink = () => {
+  const [isDeepLink, setIsDeepLink] = useState<boolean | null>(null)
```
olerichter00, line 15: **very minor:** Shouldn't the initial value be `undefined` instead of `null` because it's not yet defined ([null vs undefined](https://dev.to/sduduzog/null-vs-undefined-what-to-choose-what-to-use-11g))?

src/app/utils/hooks/useIsDeepLink.ts
```diff
@@ -0,0 +1,55 @@
...
+
+        if (shouldTreatAsDeepLink) {
+          setIsDeepLink(true)
+        } else {
+          setIsDeepLink(false)
+        }
+      })
+      .catch((error) => {
+        console.error("Error getting initial URL", error)
+        setIsDeepLink(false)
```
olerichter00, line 48: **also very minor:** Setting the value to `false` when retrieving the URL fails does not seem 100% correct. However, setting it to another value (`undefined` or `null`) or adding an error state will probably make this hook very complicated. 🤷

src/app/utils/useHideSplashScreen.ts
```diff
@@ -20,14 +21,18 @@ export const useHideSplashScreen = () => {
 
     if (isHydrated) {
       if (isLoggedIn && isNavigationReady) {
-        prefetchUrl("/", homeViewScreenQueryVariables(), {
-          force: false,
+        Linking.getInitialURL().then((url) => {
```
olerichter00, line 24: Couldn't we use the new hook `useIsDeepLink` here as well?

anandaroop: I had the same thought and tried this out briefly, but got some errors that suggested this wouldn't work. Do you want to give it a try @olerichter00?

src/app/utils/hooks/useIsDeepLink.ts
```diff
@@ -0,0 +1,55 @@
...
+        const result = matchRoute(url)
+        const isExternalUrl = result.type === "external_url"
+        const isHomeLink = result.type === "match" && result.module === "Home"
+        const shouldTreatAsDeepLink = !isHomeLink && !isExternalUrl
+
+        if (shouldTreatAsDeepLink) {
+          setIsDeepLink(true)
+        } else {
+          setIsDeepLink(false)
+        }
```
MounirDhahri, line 44:
```suggestion
        setIsDeepLink(shouldTreatAsDeepLink)
```

src/app/utils/useHideSplashScreen.ts
olerichter00: It looks like `useIsFocused` (which is used by `useIsDeepLink`) is causing the error because it cannot be used outside of a `NavigationContainer`.

But I wonder whether we need the deep link check here at all. Even when following a deep link, prefetching the home view query would make sense because it's likely that the user will navigate to the home screen within the same session. Removing the check could have performance implications, but I don't think they are significant (which we should check :)).

simulator_screenshot_525B6452-2BB9-4C58-8F34-47B09C50ACD0

brainbicycle: the motivation here is that eigen is very chatty when first launched and this has led to incidents in the past: https://artsy.slack.com/archives/CA8SANW3W/p1731531969773739

It does occur to me though that the concern is more around push notifications and in app messages specifically as these will result in large groups of users launching the app in a small time frame and causing spikes on our servers

brainbicycle: we could possibly distinguish that case specifically 🤔

anandaroop: +1 to that case being our exact motivation here.

How might we distinguish the push/in-app case vs others such as email?

brainbicycle: we have 2 different ways of handling pushes right now for iOS and Android 🙃 but I think the basic idea would be similar to what you have done here, try to find push handling early enough in the lifecycle, set a flag `isNonHomePush` or hopefully a better name, if that flag is true don't do the prefetching, if it is false do it.

brainbicycle: first place I would look is `handleReceivedNotification` in PushNotification.ts, check if that is early enough in lifecycle and works across platforms, if so good to go if not might need to dig in more.

brainbicycle: in app messages I would need to look at, there should be some braze callbacks we can get hopefully in js land.

anandaroop: I'll timebox an attempt at this^ (happy to pair also) — but if that doesn't pan out quickly how would we feel about merging this as-is, and then following up with this further refinement? 

If I'm understanding correctly, the only risk right now is a non-optimized home screen load for people who following an email deeplink, then navigate back to home. 

If that's right I think we stand to gain more than we lose with this current version, since we are still open to this stampede problem, even as we speak.

brainbicycle: I am fine with that personally, I think the more risky thing is the check in the HomeContainer, if we get that wrong we get dead home screens. I will create a ticket to follow up so if you want to timebox fine but fine also to leave it for the time being.

olerichter00: +1 merging this PR as is 👍

anandaroop: Ok, let's go with that. (My timebox didn't really yield anything).

One more rebase and I'll merge this.

PR #11137, opened 2024-11-19
fix: city picker padding on city guide

MounirDhahri:
This PR resolves [https://www.notion.so/artsy/City-guide-drop-down-misaligned-not-sure-if-new-13bcab0764a080838854d7f873a0b78e?pvs=4] 

### Description

This PR fixes the broken padding on the city guide after the latest navigation changes.

| Before | After |
|--------|--------|
| IMG_3D3FA04DBC6D-1 | Screenshot 2024-11-19 at 19 08 41 | 

### PR Checklist

### To the reviewers 👀

- city picker padding on city guide - mounir

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Scenes/Map/GlobalMap.tsx
+++ b/src/app/Scenes/Map/GlobalMap.tsx
@@ -550,10 +550,7 @@ export class GlobalMap extends React.Component<Props, State> {
               style={{ ...this.backgroundImageSize }}
             />
 
-            {/* We do this gymnastics to get accurately the top inset
-              We are injecting two extra margins that on the native side that we need to deduct
-              https://github.com/artsy/eigen/blob/054ef4fea23bf401a53fc3f44bebb30933ca80f2/ios/Artsy/Emission/ViewControllers/ARMapContainerViewController.m#L119
-              https://github.com/artsy/eigen/blob/054ef4fea23bf401a53fc3f44bebb30933ca80f2/ios/Artsy/View_Controllers/Util/ARNavigationController.m#L212
+            {/* TODO: safeArea insets are being reported back incorrectly here, so we have to subtract, we should look further into why, possibly with how we register these components, possibly because they are partially native. 
              */}
             <TopButtonsContainer style={{ top: this.props.safeAreaInsets.top - 12 - 10 }}>
               <Animated.View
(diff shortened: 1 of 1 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (city picker padding on city guide - mounir - MounirDhahri)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 68799d897987491643bd7e97f75c0e27008fbb8d
</p>

src/app/Scenes/Map/GlobalMap.tsx
```diff
@@ -549,7 +549,13 @@ export class GlobalMap extends React.Component<Props, State> {
               resizeMode="cover"
               style={{ ...this.backgroundImageSize }}
             />
-            <TopButtonsContainer style={{ top: this.props.safeAreaInsets.top + 12 }}>
+
+            {/* We do this gymnastics to get accurately the top inset
+              We are injecting two extra margins that on the native side that we need to deduct
+              https://github.com/artsy/eigen/blob/054ef4fea23bf401a53fc3f44bebb30933ca80f2/ios/Artsy/Emission/ViewControllers/ARMapContainerViewController.m#L119
+              https://github.com/artsy/eigen/blob/054ef4fea23bf401a53fc3f44bebb30933ca80f2/ios/Artsy/View_Controllers/Util/ARNavigationController.m#L212
+             */}
```
brainbicycle, line 557: thanks for the comment, I wonder if instead we can remove these from the native side? maybe that will break other things but happy to give a shot tomorrow

MounirDhahri: I did consider doing that but since the padding was part of the `ARNavigationController` and that wraps all other native modules, I was afraid of how many things I would need to fix as well - and if I would know how to do it 😄

brainbicycle: i will give it a quick look and see how it goes, not critical though if it is too hairy

brainbicycle: i am confused still by this but it seems like a rabbit hole, it seems like the values in the native controllers aren't really affecting the position of the button, rather the safe area is being reported back incorrectly, in my testing: `mapProps safeAreaInsets 100.33333333333333` when it should be around 62 on a device with a notch. I can't really understand why it seems like possibly due to how we register these components, is there a conflicting SafeAreaProvider, or because they are partially native that it is breaking things. Dropping for now but does feel a little precarious, also debugging has gotten worse with these modules not sure when but constantly getting unregistered errors on refresh. Will make a ticket to follow up, leaving for now.

brainbicycle (dismissed):

src/app/Scenes/Map/GlobalMap.tsx
brainbicycle: will update the comment with my findings and merge 🙏

PR #11213, opened 2024-11-28
fix: sort & filter modal performance issues

MounirDhahri:
### Description

This PR fixes an outstanding issue we have with the filter modal which still rely on the old `FancyModal` 😮.

`FancyModal` is actually no longer needed and IMO, **should** be deprecated because there is nothing we can't do anymore with the native Modal component and we don't need the cards stack and the **extremely** complicated custom cards layering logic.

### PR Checklist

### To the reviewers 👀

#nochangelog


MounirDhahri:
This worked https://artsy.slack.com/archives/C05EQL4R5N0/p1732796324420449?thread_ts=1732551427.261299&cid=C05EQL4R5N0

PR #11235, opened 2024-12-03
chore: improve tab press event performance

MounirDhahri:
This PR resolves [] 

### Description

In this PR, I am improving the performance of the tab event.

While working on dev, I started investigating why is the transition between tabs slow. We are no longer using the old infra so it should be that bad. This led me to notice that the tabPress event function was being computed everytime and it affected the performance.

**Note:** on prod, the performance is fine. It's only on dev that we notice this.

https://github.com/user-attachments/assets/f5bde761-f18b-406c-8884-5f2ae7820d02

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- improve tab press event performance - mounir

</details>

### DIFF
--- a/src/app/Navigation/AuthenticatedRoutes/Tabs.tsx
+++ b/src/app/Navigation/AuthenticatedRoutes/Tabs.tsx
@@ -19,6 +19,7 @@ import { OnboardingQuiz } from "app/Scenes/Onboarding/OnboardingQuiz/OnboardingQ
 import { GlobalStore } from "app/store/GlobalStore"
 import { useIsStaging } from "app/utils/hooks/useIsStaging"
 import { postEventToProviders } from "app/utils/track/providers"
+import { useCallback } from "react"
 import { Platform } from "react-native"
 import { useSafeAreaInsets } from "react-native-safe-area-context"
 
@@ -51,6 +52,28 @@ const AppTabs: React.FC = () => {
   const color = useColor()
   const isStaging = useIsStaging()
   const insets = useSafeAreaInsets()
+
+  const selectedTab = GlobalStore.useAppState((state) => state.bottomTabs.sessionState.selectedTab)
+
+  const handleTabPress = useCallback(
+    (e) => {
+      // the tab name is saved in e.target postfixed with random string like sell-Nw_wCNTWwOg95v
+      const tabName = Object.keys(bottomTabsConfig).find(
+        (tab) => e.target?.startsWith(tab)
+      ) as BottomTabType
+
+      if (Object.keys(BottomTabOption).includes(tabName) && selectedTab !== tabName) {
+        GlobalStore.actions.bottomTabs.setSelectedTab(tabName)
+        postEventToProviders(
+          tappedTabBar({
+            tab: bottomTabsConfig[tabName as BottomTabType].analyticsDescription,
+            contextScreenOwnerType: BottomTabOption[tabName as BottomTabType],
+          })
+        )
+      }
+    },
+    [selectedTab]
+  )
 
   const stagingTabBarStyle = {
     borderTopColor: color("devpurple"),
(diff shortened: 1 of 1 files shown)


src/app/Navigation/AuthenticatedRoutes/Tabs.tsx
```diff
@@ -109,20 +132,10 @@ const AppTabs: React.FC = () => {
...
-            GlobalStore.actions.bottomTabs.setSelectedTab(tabName)
-            postEventToProviders(
-              tappedTabBar({
-                tab: bottomTabsConfig[tabName as BottomTabType].analyticsDescription,
-                contextScreenOwnerType: BottomTabOption[tabName as BottomTabType],
-              })
-            )
-          }
+          // The goal of this is to queue up the tab press event to be handled after the tab has changed
+          setTimeout(() => {
```
MounirDhahri, line 136: Wrapping this with a `setTimeout` to make sure it is at the end of the JS stack. [Example](https://rehmat-sayany.medium.com/understanding-the-event-loop-task-queue-and-microtask-queue-in-javascript-3a00b61783ec)

ArtsyOpenSource:

- Dev changes (improve tab press event performance - mounir - MounirDhahri)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 8c02b0c8cc27ec6802923ebb4d09480d2a88d5b4
</p>

src/app/Navigation/AuthenticatedRoutes/Tabs.tsx
olerichter00: This part I don't really get; why does it improve the performance if it runs later? Is there something happening before that got blocked by this?

olerichter00 (dismissed):
Thanks a lot, good idea to improve the performance here 🌟 I've only added one comment with a question.

src/app/Navigation/AuthenticatedRoutes/Tabs.tsx
gkartalis: I think that we might be able to get away without using setTimeout and instead use something like [this](https://reactnative.dev/docs/timers#interactionmanager):

```js
InteractionManager.runAfterInteractions(() => {
  // ...long-running synchronous task...
});
```

dariakoko (approved):

src/app/Navigation/AuthenticatedRoutes/Tabs.tsx
MounirDhahri: I actually don't really know the exact details but from my testing, `tabPress` was getting called before changing the tab. So we were waiting for pending operations to finish first. It would have been great if they had an event called `onTabChanged` like `blur` or 'beforeRemove'.

MounirDhahri: A lot of the time `setTimeout` and `runAfterInteractions` are interchangeable but in this case I don't know if they are interchangeable. `InteractionManager` helps running tasks after interactions/animations have completed. In this case, I don't know for sure if what's blocking the progress of the tap is for sure an interaction or a transition. My bet, is that it's the JS thread being busy but I could be wrong here. 

Wrote this ⬆️ then tried it locally and I don't see any performance differences between `setTimeout` and `runAfterInteractions` so the bottleneck was probably not on the JS thread.

MounirDhahri: I will make the update to `InteractionManager`

olerichter00: Good idea to use `InteractionManager.runAfterInteractions` here (and luckily it works as well :)). In my opinion, `runAfterInteractions` is way more descriptive than `setTimeout` or `setImmediate`, which, I guess, only as a side effect do the same, but are meant to be used for setting a, well, timeout.

PR #11240, opened 2024-12-03
chore: deprecate fullbleed prop

MounirDhahri:
This PR resolves [] 

### Description

This PR is a follow-up to previous navigation infra improvements.

In this PR, I work on removing some custom prop that we use for defining screens safe edges: the famous `fullbleed`. This prop although is very intuitive to use. You first need to learn, then you would have to guess that you need to define it with your route **instead** of within the screen itself. Of course that led to it being very coupled to screens layout.

Soo, I removed its usage from the new navigation infra then I went through every screen that doesn't define `fullbleed` around 100 😄 and updated all the ones that broke.

The ideal `viewOptions` props that a screen would need, should be no props other than the default `screenOptions` from `react-navigation`

**Bonus:**
I aligned the navigation headers of so many old screens to match the new patterns 

**Screenshots**
I would like to post screenshots but I don't know if it makes sense to make screen shots to 150 screens - I did though check everything but if anyone raises a broken padding, we will have 2 weeks to fix it

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- deprecate fullbleed - mounir

</details>

### DIFF
--- a/src/app/routes.ts
+++ b/src/app/routes.ts
@@ -265,7 +265,6 @@ export function getDomainMap(): Record<string, RouteMatcher[] | null> {
     addRoute("/feature/:slug", "Feature"),
     addRoute("/galleries-for-you", "GalleriesForYou"),
     addRoute("/gene/:geneID", "Gene"),
-    addRoute("/home-view", "HomeView"),
     addRoute("/home-view/sections/:sectionID", "HomeViewSectionScreen"),
     addRoute("/inbox", "Inbox"),
     addRoute("/inquiry/:artworkID", "Inquiry"),
--- a/src/app/AppRegistry.tsx
+++ b/src/app/AppRegistry.tsx
@@ -71,7 +71,6 @@ import {
   AuctionResultsForArtistsYouFollowQueryRenderer,
 } from "./Scenes/AuctionResults/AuctionResultsForArtistsYouFollow"
 import { BottomTabOption, BottomTabType } from "./Scenes/BottomTabs/BottomTabType"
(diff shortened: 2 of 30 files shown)


ArtsyOpenSource:

- Dev changes (deprecate fullbleed - mounir - MounirDhahri)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 2e6840dfcb8de1ea128ded77ea18d84065ad6e4c
</p>

damassi (dismissed):
Yess, this one has been a long time coming 👍

olerichter00 (approved):
Looks great 🌟

src/app/routes.ts
```diff
@@ -265,7 +265,6 @@ export function getDomainMap(): Record<string, RouteMatcher[] | null> {
     addRoute("/feature/:slug", "Feature"),
     addRoute("/galleries-for-you", "GalleriesForYou"),
     addRoute("/gene/:geneID", "Gene"),
-    addRoute("/home-view", "HomeView"),
```
gkartalis, line 268: 👀👀 intentional?

gkartalis (approved):
thanks! Looks great

src/app/routes.ts
MounirDhahri: Yes, we created this route initially to be able to test the old home view and the new one separately. Then we moved to use the same route with a feature flag and we forgot to remove this one.

PR #11246, opened 2024-12-04
chore: better screens definition

MounirDhahri:
This PR resolves [ONYX-1428] 

### Description

This PR brings some more minor improvements to the navigation infra. 
**Summary of changes:**
- Separate `sharedRoutes` into `modalRoutes` and `commonRoutes`
- Rename `AuthenticatedNavStack` to `ModalNavStack`. Because it's a modal navigation stack 😅 
- Remove a bunch of empty spaces that I introduce the previous PR
- Extract ignore logs - these should have never belonged to `AppRegistry`

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- navigation infra -> better screen definitions - mounir

</details>

[ONYX-1428]: https://artsyproduct.atlassian.net/browse/ONYX-1428?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/Navigation/AuthenticatedRoutes/HomeTab.tsx
```diff
@@ -1,6 +1,6 @@
 import { modules } from "app/AppRegistry"
-import { registerSharedRoutes } from "app/Navigation/AuthenticatedRoutes/SharedRoutes"
 import { registerScreen, StackNavigator } from "app/Navigation/AuthenticatedRoutes/StackNavigator"
+import { commonRoutes } from "app/Navigation/AuthenticatedRoutes/commonRoutes"
```
damassi, line 3: I've vote for keeping this as `sharedRoutes`. 

It's a minor language thing, but "common" implies a bit of randomness. If something is common, it may or may not appear; there's a sort of indeterminism involved that leans towards the likely. Where shared is: "these things are shared". It simply IS. 

As an example: shared, used in variables, is a common idiom 😄 Folks use it a lot, but they also might use something entirely else.

src/app/Navigation/AuthenticatedRoutes/Tabs.tsx
```diff
@@ -148,7 +148,7 @@ const AppTabs: React.FC = () => {
   )
 }
 
-export const AuthenticatedRoutesStack = createNativeStackNavigator()
```
damassi, line 151: For outsiders reading code, there's a lot more informational capacity by keeping this as it was -- AuthenticatedRoutes. 

It 
- demarcates a boundary (auth, not-auth) thats clear at a glance
- is a common idiom
- doesn't require inside knowledge of implementation details. Ie, I have no clue what a "modalRoute" is other than knowing what a Modal is (an overlay), and can't quite understand how the bulk of our routes (authenticated) could suddenly transform into a modal. But that's only because I don't understand! 
- So: we should use language that neutral and free of technical details that aren't quite relevant for most product devs 👍 (that these are modals can be easily communicated in a code comment somewhere)

src/app/Navigation/AuthenticatedRoutes/Tabs.tsx
```diff
@@ -158,15 +158,11 @@ export const AuthenticatedRoutes: React.FC = () => {
...
-          component={AppTabs}
-          options={{ headerShown: false }}
-        />
-        {registerSharedModalRoutes()}
-      </AuthenticatedRoutesStack.Group>
-    </AuthenticatedRoutesStack.Navigator>
+    <ModalStack.Navigator>
+      <ModalStack.Group>
+        <ModalStack.Screen name="AppTabs" component={AppTabs} options={{ headerShown: false }} />
+        {modalRoutes()}
```
damassi, line 164: `authenticatedRoutes`: very strong signal at a glance

damassi (commented):
Left a few comments that aren't necessarily blocking, but something to think about (for future devs who are further away from platform code)

src/app/Navigation/AuthenticatedRoutes/commonRoutes.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { registerScreen, StackNavigator } from "app/Navigation/AuthenticatedRoutes/StackNavigator"
+import { isModalScreen } from "app/Navigation/Utils/isModalScreen"
+
+const nonModalModules = Object.entries(nonTabModules).filter(
+  ([_, module]) => !isModalScreen(module)
+)
+
+/**
+ * This represents the screens that can be accessed from any tab
+ */
```
olerichter00, line 11: Question: What are screens that cannot be accessed from any tab, and why is it that they cannot?

src/app/Navigation/AuthenticatedRoutes/HomeTab.tsx
olerichter00: question: What exactly are shared routes, and what's the difference from routes that are not shared?

olerichter00 (commented):
Thank you! I've added a few questions as comments ❓ ❔ ❓

src/app/Navigation/AuthenticatedRoutes/commonRoutes.tsx
MounirDhahri: Good question. The main reason is to make sure that some screens do not randomly pop up at the wrong tab. For example, we wouldn't want to have the home view showing up in the profile tab. It doesn't make sense as well to show the Inbox in the search tab. 

Screenshot 2024-12-05 at 10 25 51

Basically, the main tab screens and whatever that we would want to make sure is only available at a specific tab.

I spoke about it in further details here https://drive.google.com/drive/u/0/folders/1wUAUHt86WLQMy7Fx3MU7k4bj4y1ngaCk

src/app/Navigation/AuthenticatedRoutes/HomeTab.tsx
MounirDhahri: I answered [here](https://github.com/artsy/eigen/pull/11246#discussion_r1870993749) about the routes that are not shared.

For the routes that are shared, it's the screens that will be available in all tabs - for example, the artwork screen can be opened from every tab

Screenshot 2024-12-05 at 10 34 01

MounirDhahri: @damassi cool - that makes sense, I should have trusted my initial naming more 😄

MounirDhahri: @olerichter00 I will add a comment about it to the code

src/app/Navigation/AuthenticatedRoutes/Tabs.tsx
MounirDhahri: Good point - here I wanted to be more descriptive about what is it and focused less about what it does - I will rename it back. Thanks

src/app/Navigation/AuthenticatedRoutes/HomeTab.tsx
olerichter00: Thanks, Mounir! 🙏

PR #11248, opened 2024-12-04
feat(DIA-1014): create an entry point for infinite discovery

iskounen:
This PR resolves [DIA-1014]

### Description

This PR adds an entry-point for Infinite Discovery behind a feature flag. The feature flag will allow us to enable a new Home View section for Infinite Discovery without exposing it to users who have versions of the app with the feature in-progress.

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- Added infinite discovery scaffolding behind feature flag

</details>

[DIA-1014]: https://artsyproduct.atlassian.net/browse/DIA-1014?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/Scenes/InfiniteDiscovery/InfiniteDiscovery.tsx
```diff
@@ -0,0 +1,25 @@
+import { NavigationContainer } from "@react-navigation/native"
+import { createStackNavigator } from "@react-navigation/stack"
+import { InfiniteDiscoveryContext } from "app/Scenes/InfiniteDiscovery/InfiniteDiscoveryContext"
+import InfiniteDiscoveryHeader from "app/Scenes/InfiniteDiscovery/InfiniteDiscoveryHeader"
+import InfiniteDiscoveryScreen from "app/Scenes/InfiniteDiscovery/InfiniteDiscoveryScreen"
+
+const Stack = createStackNavigator()
+
+export default () => {
```
damassi, line 9: We should def keep to idioms! 

```tsx
export const InfiniteDiscoveryApp: React.FC = () => {
  ... 
}
```

damassi: (And importantly, always avoid [default exports](https://github.com/artsy/force/blob/main/docs/best_practices.md#naming-imports-and-exports))

src/app/Scenes/InfiniteDiscovery/InfiniteDiscoveryScreen.tsx
```diff
@@ -0,0 +1,67 @@
+import { Button, Flex, Screen, Text, useScreenDimensions, useTheme } from "@artsy/palette-mobile"
+import { FancySwiper } from "app/Components/FancySwiper/FancySwiper"
+import { Card } from "app/Components/FancySwiper/FancySwiperCard"
+import { InfiniteDiscoveryContext } from "app/Scenes/InfiniteDiscovery/InfiniteDiscoveryContext"
+
+export default () => {
```
damassi, line 6: Ditto

src/app/Scenes/InfiniteDiscovery/InfiniteDiscoveryScreen.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const artworkIds = InfiniteDiscoveryContext.useStoreState((state) => state.artworkIds)
+  const goNext = InfiniteDiscoveryContext.useStoreActions((actions) => actions.goForward)
```
damassi, line 8: Its best to avoid `1:2` representations of things and stick to 1:1 (good habit to internalize). Meaning, if the action is called `goForward` in the store, the variable should be called `const goForward` locally. This makes refactoring easier later because a find and replace will match all instances in one go, vs needing to hunt for `1:n` representations of things

src/app/Scenes/InfiniteDiscovery/hooks/useInfiniteDiscovery.ts
```diff
@@ -0,0 +1,13 @@
+import { InfiniteDiscoveryContext } from "app/Scenes/InfiniteDiscovery/InfiniteDiscoveryContext"
+
+export default () => {
```
damassi, line 3: Ditto

damassi: The rule of thumb is to name the primary named export after the file name -- that's 1:1. So 
```tsx
export const useInfiniteDiscovery = () => { ... }
```

src/app/Scenes/InfiniteDiscovery/hooks/useInfiniteDiscovery.ts
```diff
@@ -0,0 +1,13 @@
...
+  // state
+  const artworkIds = InfiniteDiscoveryContext.useStoreState((state) => state.artworkIds)
+  const currentArtworkId = InfiniteDiscoveryContext.useStoreState((state) => state.currentArtworkId)
+
+  // actions
+  const goBack = InfiniteDiscoveryContext.useStoreActions((actions) => actions.goBack)
+  const goForward = InfiniteDiscoveryContext.useStoreActions((actions) => actions.goForward)
+
+  return { artworkIds, currentArtworkId, goBack, goForward }
```
damassi, line 12: #non-blocking, but would suggest going with this form:

```tsx
const state = InfiniteDiscoveryContext.useStoreState((state) => state)
const actions = InfiniteDiscoveryContext.useStoreActions((actions) => actions)

return { state, actions } 
```

damassi (approved):
A couple minor naming comments around best practices, but otherwise looks ❇️

src/app/Scenes/InfiniteDiscovery/hooks/__tests__/useInfiniteDiscovery.tests.ts
```diff
@@ -0,0 +1,57 @@
...
+      console.log(selector.toString())
+      if (selector.toString().includes("artworkIds")) return ["1", "2", "3"]
+      if (selector.toString().includes("currentArtworkId")) return "2"
+    })
+
+    mockUseStoreActions.mockImplementation((selector) => {
+      if (selector.toString().includes("goBack")) return mockGoBack
+      if (selector.toString().includes("goForward")) return mockGoForward
+    })
+  })
```
iskounen, line 40: I'm struggling to write tests for components (or hooks) that use `InfiniteDiscoveryContext` actions because I can't mock the state or spy on the actions without hacking its implementation like this.

I have followed the Easy Peasy instructions for testing components, but haven't figured out how to assert that one of its actions was called without this implementation.

src/app/Scenes/InfiniteDiscovery/InfiniteDiscovery.tsx
damassi: Ah: `InfiniteDiscoveryScene`

src/app/Scenes/InfiniteDiscovery/InfiniteDiscoveryScreen.tsx
iskounen: Thanks for pointing this out. I missed it while changing the name back and forth between `goNext` and `goForward`.

src/app/Scenes/InfiniteDiscovery/hooks/useInfiniteDiscovery.ts
iskounen: I like default exports as a way to keep files small and focused on doing one thing, but I'm OK to change this.

damassi: Yah, we've very intentionally refactored out all default exports over the years from all of our codebases. Worth while to browse around to see, as its make the code handshake between compiler, to type-checker entirely locked down (a named export that doesn't exist will error at two separate levels). Happy to explain the reasoning in more detail!

damassi: FWIW, from a TS39 member who helped design the modern JS spec -- tweet can't be found tho TLDR is that the committee regretted supporting default exports, but it was required for backwards compat with Node.js's commonJS: 

Screenshot 2024-12-05 at 10 10 10 AM

src/app/Scenes/HomeView/Sections/Section.tsx
```diff
@@ -33,13 +34,22 @@ export interface SectionSharedProps extends FlexProps {
...
     }
     return null
   }
 
+  if (
+    section.internalID === "home-view-section-infinite-discovery" &&
+    !isInfiniteDiscoveryEnabled
+  ) {
+    return null
+  }
```
iskounen, line 51: @anandaroop this is the best way that I could think of feature-flagging a card until it is ready for release. The reason why I need this feature flag in addition to the Unleash flag is so that users with app versions that contain the WIP Infinite Discovery feature don't see the card in their app when the Unleash flag is enabled on production. I'm curious if you agree with this approach.

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against f108d0b2ca7cad1073f9f1d7f12942c52ec404cb
</p>

src/app/Scenes/InfiniteDiscovery/InfiniteDiscoveryContext.ts
```diff
@@ -0,0 +1,42 @@
...
+    "white100",
+    "green100",
+    "yellow100",
+    "orange100",
+    "red100",
+    "purple100",
+    "blue100",
+  ],
+  currentArtworkId: "black100",
+  goBack: action((state) => {
```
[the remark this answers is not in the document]
iskounen: I switched to `goToPreviousArtwork` and `goToNextArtwork`.

src/app/Scenes/InfiniteDiscovery/hooks/__tests__/useInfiniteDiscovery.tests.ts
iskounen: I was struggling to test components that use Easy Peasy state/actions because I wasn't wrapping them in a context provider. I got rid of the `useInfiniteDiscovery` hook after figuring that out.

src/app/Scenes/HomeView/Sections/Section.tsx
MounirDhahri: I would actually suggest to move this logic to within the section itself to avoid coupling its logic to `Section.tsx`

src/app/Scenes/HomeView/Sections/HomeViewSectionCard.tsx
```diff
@@ -187,6 +187,8 @@ function getRoute(card: any) {
   } else if (card.entityType === "Page" && card.entityID === OwnerType.galleriesForYou) {
     // not a canonical web url, thus the indirection above
     route = "/galleries-for-you"
+  } else if (card.entityType === "Page" && card.entityID === OwnerType.infiniteDiscovery) {
```
MounirDhahri, line 190: this is a good opportunity to switch this logic to be a switch/case statement

src/app/Scenes/InfiniteDiscovery/InfiniteDiscovery.tsx
```diff
@@ -0,0 +1,115 @@
...
+  )
+
+  const canGoBack = artworks.length && currentArtwork !== artworks[0]
+
+  const handleBackPressed = () => {
+    goToPreviousArtwork()
+  }
+
+  const handleExitPressed = () => {
+    navigate("/home-view")
```
MounirDhahri, line 40: RIP `home-view` 😄  - maybe use `/` or `dismissModal` if you present this as a modal

MounirDhahri (commented):
High quality stuff 👏

src/app/Scenes/HomeView/Sections/Section.tsx
iskounen: I would move it above [this line](https://github.com/artsy/eigen/blob/277f98e1028db9f36710c6b73bc4c0e0ce4cebb2/src/app/Scenes/HomeView/Sections/HomeViewSectionCard.tsx#L160) if I were to move this into the section. Does that sound right?

src/app/Scenes/HomeView/Sections/HomeViewSectionCard.tsx
iskounen: I was thinking of refactoring it to a series of if...return statements to avoid nesting.

src/app/Scenes/HomeView/Sections/Section.tsx
MounirDhahri: Yes, that sounds good to me - it's fine as well to keep it inside the component for this section since the section is already public and is stable and fast to render.

src/app/Scenes/InfiniteDiscovery/InfiniteDiscovery.tsx
iskounen: Ah! That's what caused it the exit button to fail. I was going to leave that for a follow-up PR because I want to spend some more time with the new navigation system to see what is possible.

src/app/Scenes/HomeView/Sections/HomeViewSectionCard.tsx
```diff
@@ -187,6 +187,8 @@ function getRoute(card: any) {
...
+    route = "/infinite-discovery"
```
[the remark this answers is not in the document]
anandaroop: That's an interesting idea. My instinct is to push back, on the assumption that canonical urls are preferable. But that bumps up against the reality that we may increasingly be introducing AA features as mobile-first or [even](https://artsy.slack.com/archives/C07ANEV7RNV/p1733416246739459) mobile-only. I'll be curious what others think too.

src/app/Scenes/HomeView/Sections/HomeViewSectionCard.tsx
MounirDhahri: that's also fine for me.

src/app/Scenes/InfiniteDiscovery/InfiniteDiscovery.tsx
MounirDhahri: awesome

src/app/Scenes/HomeView/Sections/HomeViewSectionCard.tsx
```diff
@@ -169,6 +172,10 @@ export const HomeViewSectionCardQueryRenderer: React.FC<SectionSharedProps> = wi
       }
     )
 
+    if (sectionID === "home-view-section-infinite-discovery" && !isInfiniteDiscoveryEnabled) {
+      return null
+    }
```
iskounen, line 177: @MounirDhahri I moved this here, but I have to put it after the `useLazyLoadQuery` call because hooks can't be conditionally rendered. Do you have any concerns about that?

MounirDhahri: 1. that's a good point - how about we actually exit before we even load the data. that way, we can avoid that additional `useLazyLoadQuery`.
I mean something like this. 


Screenshot 2024-12-06 at 17 18 56



2. You might not need to do much mocking if you do the above, however if some day you would want to do that, `setupTestWrapper` accepts also query renderers and suspended components. Along with `__globalStoreTestUtils__?.injectFeatureFlags` you should be able to mock the ff and response from relay

Screenshot 2024-12-06 at 17 21 18

MounirDhahri:
Hey @iskounen - Something just came to my mind
I _think_, we don't need a feature flag here. I actually don't know how we can have efficient feature toggling here without proper versioning.

**What do I mean with that?**
First, assuming you would set the feature flag from MP (let's call that the unleash flag) and Eigen (let's call that the echo flag) as ready for release and release the feature to all users. 

Usually, you would expect only the versions where the echo flag is set as `readyForRelease` to receive the new banner, right? 

**Let's see what will happen for each kind of app user?**

- For users with the latest version, the answer is **Yes**. They will get the new banner and by pressing on it, they will be redirected to the infinite discovery screen(s).
- For users who have old version where the echo flag is not ready for release -the answer is also **Yes**, no banner will be visible since we do explicitly check for that. Nice!
Screenshot 2024-12-09 at 16 33 31

- For users who don't have this logic yet - meaning they don't even have the flag - which is all users that we currently have or that might not upgrade the app until later. The answer is **No**, they will get a `HomeViewSectionCard` with `id` `home-view-section-infinite-discovery` -  we will end up showing them a banner, with no active route/broken route. It's because they don't have this piece of code yet. Something like this 👇 

https://github.com/user-attachments/assets/d4028baf-63ec-40a1-9b10-8825b894bfaf


**How do I suggest to proceed with this**
- Spike on other feature flagging approaches on MP - maybe by version? I don't know, I don't want to start jumping to solutions here.
- For the time being, to avoid disrupting your workflow, we would be better with a new section hidden behind a feature flag (just on EIgen side until we figure out a more appropriate approach)

iskounen:
@MounirDhahri thanks for raising that issue.

We discussed it offline with @anandaroop and decided to:

- Find a way to feature flag Home View sections by app version in Metaphysics
- Keep the feature flag in Eigen for the time being

anandaroop:
@iskounen the [MP solution for this](https://github.com/artsy/metaphysics/pull/6297) has been merged.

Note that PR is being merged [with a version constraint](https://github.com/artsy/metaphysics/pull/6297/files#r1879053756) which will allow this section to appear in current builds of Eigen; you'll probably want to raise that version constraint at some point (or feel free to remove it for now, if that makes things easier)

PR #11256, opened 2024-12-06
feat: consolidate paths and screens

MounirDhahri:
This PR resolves [ONYX-1427]

### Description

This PR suggests a few improvements to our navigation infra and focuses mainly on consolidating `routes` and `AppRegistry`.

**What am I solving here?**
- Currently, in order to define a screen, you would need to add its path to `routes.ts` then to specify the screen in `AppRegistry`. This fragmented experience can be avoided if we define both together just like we do in Force.
In order to define as screen now, you would need to open `routes.tsx` and add a new route like so:
```typescript
// routes.tsx
  {
    path: "/new-path",
    name: "MyNewScreen",
    Component: MyNewScreen,
  },
```

Instead of 

```typescript
// routes.ts
addRoute("/new-path", "MyNewScreen")


// AppRegistry.tsx
const modules = defineModules([
....
   MyNewScreen: {
     Component: MyNewScreen
   }
....

```

**What was the main challenge here?**
- The same module, can be available from two different routes. For example `artist/:artistID` and `/artist/:artistID/auction-results` both point to the artist screen. The opposite is not true - meaning we have a 1..n relationship between module and routes.
- There is some sorcery made to get typescript to work 😄 

**What comes next?**
- `injectParams` can be improved - it's probably not needed anymore since we can define the params inline now withing the `Component` props.
- `getDomainMap()` isn't trivial. What are we doing there? there is too much code and too little done - it should be something like, if not `live.artsy.net`, it's `artsy.net` 🤷 
Screenshot 2024-12-06 at 12 14 08
- Docs update

**What to focus on as reviewer here?**
- How do you find the new API?
- How do you find the folders structure?
- Do you prefer the old API? I hope not hahaha
- Please keep in mind that this is a suggestion and I want to hear your thoughts about it. Thanks @olerichter00 for helping with this.

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- consolidate paths and screens - mounir

</details>

[ONYX-1427]: https://artsyproduct.atlassian.net/browse/ONYX-1427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/Navigation/routes.tsx
MounirDhahri, line 1: `Routes.tsx` used to do so many things - now it only defines routes. Everything else lives separately now

src/app/Navigation/AuthenticatedRoutes/StackNavigator.tsx
```diff
@@ -8,18 +8,18 @@ import {
...
 import { Platform } from "react-native"
 import { isTablet } from "react-native-device-info"
 
 export const StackNavigator = createNativeStackNavigator<AuthenticatedRoutesParams>()
 
 type StackNavigatorScreenProps = {
-  name: keyof AuthenticatedRoutesParams
-  module: ModuleDescriptor
+  name: AppModule
+  module: ModuleDescriptor<any>
```
brainbicycle, line 22: any way to get rid of the <any> and make a more specific type?

MounirDhahri: omg thanks for reminding me about this 😄  - I had sooo many changes going on and I wanted to get back to it later. I will check it out now

MounirDhahri:
Git got confused - it says I have two utils folders but I have only one - 
Screenshot 2024-12-06 at 14 26 23
Screenshot 2024-12-06 at 14 27 00

I might need to rename the folder to helpers then rename it again to avoid breaking devs local environments

src/app/Navigation/AuthenticatedRoutes/StackNavigator.tsx
MounirDhahri: it was so easy 🤩 
Screenshot 2024-12-06 at 14 32 39

damassi (dismissed):
Yes @MounirDhahri! This PR is such a nice simplification all around ✨  1 static route definition that does all the things. Great work 👍

src/app/Navigation/routes.tsx
```diff
@@ -0,0 +1,1504 @@
...
+  {
+    path: "/artist/:artistID",
+    name: "Artist",
+    Component: ArtistQueryRenderer,
+    options: {
+      screenOptions: {
+        headerShown: false,
+      },
+    },
+    Queries: [ArtistScreenQuery],
```
damassi, line 323: @MounirDhahri - with this now in place, we're very close to being able to implement route prefetching like its done in force: see [here](https://github.com/artsy/force/blob/main/src/System/Hooks/usePrefetchRoute.tsx#L24). 

The main step from here would be to generalize, app-wide, a `RouterLink` style component (similar to force) which takes a path. Any tappable surfaces that redirect to a route implements that, and then internal to that, we detect whether that's viewable and then prefetch. 

Check out:
- https://github.com/artsy/force/blob/main/src/System/Components/RouterLink.tsx#L55
- https://github.com/artsy/force/blob/main/src/System/Components/RouterLink.tsx#L74-L78

Once that's in place, a user scrolls down the app, a prefetch is triggered, and then by the time they tap the link (even if its a QueryRenderer) the cache will be warmed and the page should load instantly. 

We're very very close here! Just need to get some kind of 
```tsx
<AppLink to="artist/andy-warhol">Tap Me</AppLink>
```
in the app!

MounirDhahri: Thanks for the hint - that wasn't actually on my mind but heck yeah! it will be fun to implement!

MounirDhahri:
No one minds the new approach.  - I will fix merge conflicts and update our docs tomorrow - you can already review the PR - not much will chang

src/app/Navigation/routes.tsx
olerichter00: That would be amazing! We just need to find a way to determine when an element enters the viewport. When I implemented prefetching in Eigen 3 years ago, I couldn't find a working equivalent to the browser's Intersection Observer API that is used in Force. Instead, I used `FlatList`'s [onViewableItemsChanged](https://reactnative.dev/docs/flatlist#onviewableitemschanged) callback ([implementation](https://github.com/artsy/eigen/blob/b8d894c4729f2bc7527fe5c4ccb8c7582f350f42/src/app/Components/PrefetchFlatList.tsx#L16)).

But I would like to give it another try. [react-native-viewport-detector](https://github.com/thinhkhang97/react-native-viewport-detector) and [react-native-intersection-observer](https://www.npmjs.com/package/react-native-intersection-observer) could be some candidates 💡

src/app/Navigation/routes.tsx
```diff
@@ -0,0 +1,1490 @@
...
+        options: {
+          alwaysPresentModally: true,
+          screenOptions: {
+            gestureEnabled: false,
+            headerShown: false,
+          },
+        },
+        injectParams: (params) => ({ slug: params["*"] }),
+      }
+    : {
```
olerichter00, line 1474: Isn't this a duplicate of [defineWebViewRoute("/*", {....](https://github.com/artsy/eigen/pull/11256/files#diff-7a830bc549b1b6490596e87df987d53f84bc8a335bcc8afb03249dd77d3e3418R1454)?

src/app/Navigation/routes.tsx
```diff
@@ -0,0 +1,1490 @@
...
+    path: "/:slug",
+    name: "VanityURLEntity",
+    Component: VanityURLEntityRenderer,
+  },
+  defineWebViewRoute("/*", {
+    alwaysPresentModally: true,
+  }),
+])
+
+export const liveDotArtsyRoutes = defineRoutes([
```
olerichter00, line 1459: How does it work and how is it different from [defineWebViewRoute("/*", {....](https://github.com/artsy/eigen/pull/11256/files#diff-7a830bc549b1b6490596e87df987d53f84bc8a335bcc8afb03249dd77d3e3418R1454)?

olerichter00 (commented):
This is such a great improvement! It will simplify defining new routes and searching for a route a lot 💯

src/app/Navigation/routes.tsx
MounirDhahri: good thought - I can probably also simply this and use `defineWebViewRoute`

src/app/Navigation/routes.tsx
MounirDhahri: good news @olerichter00 - we actually have something that does that and we are using it [here](https://github.com/artsy/eigen/blob/976159b57ab73559d405562dac2e7eb814948318/src/app/Scenes/HomeView/Components/HomeViewSectionSentinel.tsx#L38).

olerichter00: Thanks, @MounirDhahri, I'll have a look 👀

src/app/Navigation/routes.tsx
olerichter00: And could `liveDotArtsyRoutes` be included in the other routes array? If there is a particular reason to separate them, a small comment above could help to understand what is included there.

olerichter00: And I have an idea for the naming. Although the names `liveDotArtsyRoutes` and `artsyDotNetRoutes` are very descriptive, they are kind of difficult to read. And they are tight to the actual URL, which could change, I guess.

What do you think about changing it to something like:

```
export const liveRoutes = [
  ......
]

export const routes = [
 ....
 ...liveRoutes
 ]

ArtsyOpenSource:

- Dev changes (consolidate paths and screens - mounir - MounirDhahri)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 3f0ae952886d6c530a0713396e6bd61c1d6f6496
</p>

src/app/Navigation/routes.tsx
MounirDhahri: > good thought - I can probably also simply this and use `defineWebViewRoute`

I renamed this to be less confusing because it was simply a `webViewRoute`. `define` made it more confusing

src/app/Navigation/routes.tsx
damassi: Yah @olerichter00 - i looked around for something too (a while back) but couldn't find it, but assumed i was just thinking about it in the wrong way and that we'd be able to find a work around once all of the other details were taken care of 👍

src/app/Navigation/routes.tsx
olerichter00: @damassi, I'd be very interested to hear your thoughts on the variable names :)

olerichter00: @MounirDhahri, this shouldn't block you from merging the PR, but I know you enjoy discussions about naming as well :)

MounirDhahri:
Merging this to give us 3 weeks of testing things out. Thanks everyone for reviewing

PR #11285, opened 2024-12-13
fix: cacheable home view section card query

iskounen:
This PR removes the `@cacheable` directive from the Home View card section query. That directive prevents the `X-ACCESS-TOKEN` from propagating to Metaphysics, so Metaphysics assumes that the request is unauthenticated. This, in turn, prevents home view sections with `requiresAuthentication: true` to not display.

-

#### iOS user-facing changes

-

-

#### Dev changes

- Removed a @cacheable directive from a home view query

</details>

### DIFF
--- a/src/app/Scenes/HomeView/Sections/HomeViewSectionCard.tsx
+++ b/src/app/Scenes/HomeView/Sections/HomeViewSectionCard.tsx
@@ -148,7 +148,7 @@ const HomeViewSectionCardPlaceholder: React.FC<FlexProps> = (flexProps) => {
 }
 
 const homeViewSectionCardQuery = graphql`
-  query HomeViewSectionCardQuery($id: String!) @cacheable {
+  query HomeViewSectionCardQuery($id: String!) {
     homeView {
       section(id: $id) {
         ...HomeViewSectionCard_section


src/app/Scenes/HomeView/Sections/HomeViewSectionCard.tsx
```diff
@@ -148,7 +148,7 @@ const HomeViewSectionCardPlaceholder: React.FC<FlexProps> = (flexProps) => {
 }
 
 const homeViewSectionCardQuery = graphql`
-  query HomeViewSectionCardQuery($id: String!) @cacheable {
```
iskounen, line 151: @mzikherman is it possible to make this @cacheable conditional? I want some cards to be cacheable and some to not.

ArtsyOpenSource:

- Dev changes (Removed a @cacheable directive from a home view query - iskounen)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 192836354f25f5fee2199815233da94d55153b8f
</p>

iskounen:
After debating whether we should (a) disable caching for all card sections, or (b) remove the authentication requirement for this specific card, we decided to do (b). The Infinite Discovery card launches a feature that requires authentication, but the card itself does not require any authentication to display. Furthermore, if we display this card to unauthenticated users in the future, we should prompt them to authenticate instead of hiding the feature from them.

PR #11293, opened 2024-12-17
feat: Introduce RouterLink component to prefetch links as they enter the viewport

olerichter00:
This PR resolves [ONYX-1461] 

### Description

This PR introduces a `RouterLink` component that wraps Palette's `Touchable` and takes an extra `to` prop with a URL or path to link to.

The reason for introducing the new `RouterLink` component is to be able to prefetch (all) app links as they enter the viewport, similar to how it's implemented in Force (https://github.com/artsy/force/pull/14520 & https://github.com/artsy/force/pull/14454).

If prefetching is enabled, the component automatically prefetches the URL with Relay as soon as it enters the viewport.

The new `RouterLink` component will replace `PrefetchFlatList`, which is also capable of prefetching links as they enter the viewport, but is limited and cannot be used as universally as the new link component.

### Next Steps:

* Use the new `RouterLink` component in more places
* Make routes ready to be prefetched by adding the main query to `routes.ts`

### Screenshots / Videos

https://github.com/user-attachments/assets/11ab4c7c-d0b4-4f9a-b3c1-ee281e5cf311

### PR Checklist

### To the reviewers 👀

- Prefetch links as they enter the viewport - ole

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[ONYX-1461]: https://artsyproduct.atlassian.net/browse/ONYX-1461?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/Components/ArtworkGrids/ArtworkGridItem.tsx
```diff
@@ -26,12 +26,12 @@ import { ContextMenuArtwork } from "app/Components/ContextMenu/ContextMenuArtwor
 import { DurationProvider } from "app/Components/Countdown"
 import { Disappearable, DissapearableArtwork } from "app/Components/Disappearable"
 import { ProgressiveOnboardingSaveArtwork } from "app/Components/ProgressiveOnboarding/ProgressiveOnboardingSaveArtwork"
+import { RouterLink } from "app/Components/RouterLink"
```
damassi, line 29: Can we place all system-level things in the System folder? https://github.com/artsy/eigen/tree/main/src/app/system/navigation

(in general, we should leverage this folder more for framework-y things -- cc @MounirDhahri)

src/app/Components/RouterLink.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  const handleVisible = () => {
+    if (!disablePrefetch && enableViewPortPrefetching && to) {
+      prefetchUrl(to)
+    }
+  }
+
+  return (
+    <ElementInView onVisible={handleVisible}>
+      <Touchable {...restProps} onPress={handlePress} />
```
MounirDhahri, line 40: I wonder here if `Touchable` would be our preferred way, sometimes we use `TouchableOpacity` and sometimes we use `TouchableWithoutHighlight`. 
Anyway, that's a problem for later, we can start with this now and extend it later.

MounirDhahri: Maybe this is a call to use `RouterTouchableOpacity` 🤷

src/app/Components/ArtworkGrids/ArtworkGridItem.tsx
olerichter00: Good idea! I'll move the component to the systems folder 👍

src/app/Components/RouterLink.tsx
olerichter00: I used the `Touchable` component from "@artsy/palette-mobile" and figured this is probably the preferred one. It has a prop to adjust the opacity on press.

Yeah, `RouterTouchableOpacity` or `RouterTouchable` could also be good names for this component. I took the name `RouterLink` from [Force](https://github.com/artsy/force/blob/ad95afa02e1657c0b0e7273e5190241719c3c49b/src/System/Components/RouterLink.tsx#L37) but haven't thought much about naming yet. I'm very open for ideas for the naming.

ArtsyOpenSource:

- Cross-platform user-facing changes (Prefetch links as they enter the viewport - ole - olerichter00)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 346d8552ed3d79e2ea0b8380d6aee5bdf8383504
</p>

src/app/system/navigation/RouterLink.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { usePrefetch } from "app/utils/queryPrefetching"
+import { GestureResponderEvent } from "react-native"
+
+interface RouterLinkProps {
+  disablePrefetch?: boolean
+  passProps?: Object
+  to: string | null | undefined
+}
+
+export const RouterLink: React.FC<RouterLinkProps & TouchableProps> = ({
```
olerichter00, line 14: I've named this component `RouterLink`, as it is named in Force (routerlink.ts). However, I am open to other suggestions since this component will be used in many places.

MounirDhahri (dismissed):
Thanks for this addition. I left few suggestions - I would leave to use to decide if they're worth the effort

src/app/system/navigation/RouterLink.tsx
```diff
@@ -0,0 +1,47 @@
...
+    onPress?.(event)
+
+    if (!to) return
+
+    if (passProps) {
+      navigate(to, { passProps })
+    } else {
+      navigate(to)
+    }
+  }
```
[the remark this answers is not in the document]
olerichter00: That's a very interesting thought! I approached it this way to simplify how we implement links (similar to the web) while still having the full flexibility of `onPress`. 

Do you have an example from the app where we would not be able to use the new `to` prop? 

If it's just an edge case, then the workaround could be to do the navigation in `onPress` and not use `to` and not prefetch

src/app/utils/queryPrefetching.ts
```diff
@@ -46,6 +46,7 @@ const prefetchRoute = async <TQuery extends OperationType>(
   const allVariables = { ...result.params, ...variables }
 
   return queries.map((query) => {
+    console.log("[queryPrefetching] Prefetching:", route, JSON.stringify(allVariables))
```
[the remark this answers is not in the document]
olerichter00: Thanks! This doesn't need to be here. Just used it for debugging.

src/app/system/navigation/RouterLink.tsx
```diff
@@ -0,0 +1,47 @@
+import { Touchable, TouchableProps } from "@artsy/palette-mobile"
+import { navigate } from "app/system/navigation/navigate"
+import { ElementInView } from "app/utils/ElementInView"
+import { useFeatureFlag } from "app/utils/hooks/useFeatureFlag"
+import { usePrefetch } from "app/utils/queryPrefetching"
+import { GestureResponderEvent } from "react-native"
+
+interface RouterLinkProps {
+  disablePrefetch?: boolean
+  passProps?: Object
```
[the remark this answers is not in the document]
olerichter00: I like the idea of renaming `passProps` to something like `parameters`. But if we decide to do that, I think it would be better to rename it everywhere and not just in the `RouterLink` component. `passProps` is also used when passing parameters to `navigate`, and simply renaming it within `RouterLink` could be confusing if it's called `passProps` everywhere else.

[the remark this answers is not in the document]
olerichter00: `navigationProps` sounds good to me and it's not too different from `routerProps`. I've opened a PR for the renaming: https://github.com/artsy/eigen/pull/11380

src/app/Components/ArtworkGrids/ArtworkGridItem.tsx
```diff
@@ -307,11 +309,13 @@ export const Artwork: React.FC<ArtworkProps> = ({
...
           hideCreateAlertOnArtworkPreview={hideCreateAlertOnArtworkPreview}
         >
-          <Touchable
+          <RouterLink
             haptic
             underlayColor={color("white100")}
             activeOpacity={0.8}
             onPress={handleTap}
+            passProps={passProps}
+            to={artwork.href}
```
dariakoko, line 318: **question:** I see that you added `to` and we still have `navigate?.(artwork.href)` in the `handleTap` function. How would the navigation happen here? I see that navigation is used in the `RouterLink` component as well: `navigate(to, { passProps })`

olerichter00: Good catch! I forgot to remove the navigation in the `onPress` callback. I've opened a PR with a fix: https://github.com/artsy/eigen/pull/11382

src/app/system/navigation/RouterLink.tsx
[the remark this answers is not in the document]
MounirDhahri: I don't really have an example in mind, it's just something I had in mind to have more flexibility for `onPress`. It's not a blocker so it's totally fine as you do it.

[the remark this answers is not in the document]
olerichter00: Very good idea to keep that in mind. I'm currently converting a lot of links and will see if I find any where we need more flexibility and will try to find a solution if needed. If it's just a view, I suppose deactivating the navigation without passing `to` should work 🤞

PR #11346, opened 2025-01-07
feat(ONYX-1457): add Create Alert Prompt experiment

dariakoko:
This PR resolves [ONYX-1457] 

### Description

Help increase awareness of the Create Alert function from the artist page (high traffic and high impact)
[Figma](https://www.figma.com/design/zRKGxOWHhq3kcNiNAZCig3/OtH-App%3A-Alerts-and-Activity?node-id=340-44256&m=dev)

1. Implement two variants and control variant
- control: nothing changes
- variant-a: popover component
- variant-b: message components with CTA
3. For variants `a` and `b` the component should be shown:
- after 40 artworks is loaded
- max 2 times per user 
- the second time the user will see the popover or message component will be minimum in 7 days
- when after seeing the component the user clicks "Create Alert" CTA, the popover or message should not be shown again
4. Make sure the popover component is not shown together with any of the progressive onboarding components at the same time on the same screen 

Also adding the `payloadSuggestions: ['{"forcePrompt": "true"}', '{"forcePrompt": "false"}']` for testing purposes. Inspired by @anandaroop 🙏 

When `forcePrompt === true` the max times to show the component is 123456 and the waiting time is 30 seconds instead of 3 days. Also, in case of a forced prompt I display a text that says how many times the component has been shown. If the Create Alert CTA was not pressed after seeing the component, on the next render of the component the number will increase by 1. If Create Alert CTA was pressed, the number will be increased by 2. This functionality was added for testing purposes. 

#### iOS
| Popover | Message |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/61346c96-109b-4185-becd-5a0ca1de3a3f" width="250"/> | <video src="https://github.com/user-attachments/assets/d5387c46-643d-49c5-99ef-cb10f84bfdb3" width="250"/> |

#### Android
| Popover | Message |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/bd85636b-b1f4-4abb-80f7-917e412ae28a" width="250"/> | <video src="https://github.com/user-attachments/assets/9d8b43e0-511d-4626-a243-b1a00ffe068d" width="250"/> | 

### PR Checklist

### To the reviewers 👀

- add Create Alert Prompt experiment (onyx_create-alert-prompt-experiment)

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[ONYX-1457]: https://artsyproduct.atlassian.net/browse/ONYX-1457?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
+++ b/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
@@ -28,6 +28,7 @@ import { useArtworkFilters } from "app/Components/ArtworkFilter/useArtworkFilter
 import ArtworkGridItem from "app/Components/ArtworkGrids/ArtworkGridItem"
 import { FilteredArtworkGridZeroState } from "app/Components/ArtworkGrids/FilteredArtworkGridZeroState"
 import { Props as InfiniteScrollGridProps } from "app/Components/ArtworkGrids/InfiniteScrollArtworksGrid"
+import { useExperimentVariant } from "app/utils/experiments/hooks"
 import { extractNodes } from "app/utils/extractNodes"
 import { useFeatureFlag } from "app/utils/hooks/useFeatureFlag"
 import {
@@ -70,6 +71,11 @@ const ArtworksGrid: React.FC<ArtworksGridProps> = ({
   const artworks = useMemo(() => extractNodes(artist.artworks), [artist.artworks])
   const artworksCount = artist.artworks?.counts?.total ?? 0
   const gridRef = useRef<MasonryFlashListRef<(typeof artworks)[0]>>(null)
+  const {
+    enabled,
+    variant,
+    trackExperiment: trackCreateAlertPromptExperiment,
+  } = useExperimentVariant("onyx_create-alert-prompt-experiment")
 
   const appliedFilters = ArtworksFiltersStore.useStoreState((state) => state.appliedFilters)
 
@@ -83,6 +89,10 @@ const ArtworksGrid: React.FC<ArtworksGridProps> = ({
   const setInitialFilterStateAction = ArtworksFiltersStore.useStoreActions(
     (state) => state.setInitialFilterStateAction
   )
+
+  useEffect(() => {
+    trackCreateAlertPromptExperiment()
+  }, [])
 
   useEffect(() => {
     let filters: FilterArray = []
@@ -258,8 +268,8 @@ const ArtworksGrid: React.FC<ArtworksGridProps> = ({
     )
   }
 
-  const variant_a = true
-  const variant_b = true
+  const variant_a = enabled && variant === "variant-a"
+  const variant_b = enabled && variant === "variant-b"
 
   const userDidScroll = artworks.length >= 10 // 40
 
--- a/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
+++ b/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
@@ -337,10 +337,9 @@ const ArtworksGrid: React.FC<ArtworksGridProps> = ({
         visible={isCreateAlertModalVisible}
       />
 
-      <CreateAlertPromptMessage
-        shouldShowCreateAlertPrompt={!!userDidScroll && variant_b}
-        onPress={() => setIsCreateAlertModalVisible(true)}
-      />
+      {!!userDidScroll && !!variant_b && (
+        <CreateAlertPromptMessage onPress={() => setIsCreateAlertModalVisible(true)} />
+      )}
     </>
   )
 }
--- a/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
+++ b/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
@@ -261,7 +261,7 @@ const ArtworksGrid: React.FC<ArtworksGridProps> = ({
   const variant_a = true
   const variant_b = true
 
-  const shouldShowCreateAlertPrompt = artworks.length >= 10 // 40
+  const userDidScroll = artworks.length >= 10 // 40
 
   return (
     <>
@@ -294,7 +294,7 @@ const ArtworksGrid: React.FC<ArtworksGridProps> = ({
               <ArtistArtworksFilterHeader
                 artist={artist}
                 showCreateAlertModal={() => setIsCreateAlertModalVisible(true)}
-                shouldShowCreateAlertPrompt={!!shouldShowCreateAlertPrompt && variant_a}
+                shouldShowCreateAlertPrompt={!!userDidScroll && variant_a}
               />
             </Tabs.SubTabBar>
             <Flex pt={1}>
@@ -328,7 +328,7 @@ const ArtworksGrid: React.FC<ArtworksGridProps> = ({
       />
 
       <CreateAlertPromptMessage
-        shouldShowCreateAlertPrompt={!!shouldShowCreateAlertPrompt && variant_b}
+        shouldShowCreateAlertPrompt={!!userDidScroll && variant_b}
         onPress={() => setIsCreateAlertModalVisible(true)}
       />
     </>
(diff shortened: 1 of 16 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (add Create Alert Prompt experiment (onyx_create-alert-prompt-experiment) - dariakoko)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 246d6873de2ed8e96ede713d9dbf7bdc86c2909b
</p>

src/app/Components/Artist/ArtistArtworks/SavedSearchButtonV2.tsx
```diff
@@ -1,26 +1,41 @@
...
   const handlePress = () => {
     onPress()
+
+    /**
+     * if Create Alert CTA was pressed withing 2 minutes after the prompt was shown
+     *  do not show the prompt again
+     * */
+    if (Date.now() - promptState.dismissDate < 120000 && isFocused) {
+      dontShowCreateAlertPromptAgain()
+    }
```
dariakoko, line 31: This part is still TDB whether to keep it or not

src/app/Components/Artist/ArtistArtworks/SavedSearchButtonV2Popover.tsx
```diff
@@ -0,0 +1,25 @@
...
+}) => {
+  const {
+    sessionState: { activePopover },
+  } = GlobalStore.useAppState((state) => state.progressiveOnboarding)
+  const isFocused = useIsFocused()
+
+  const isOnboardingFinished =
+    isFocused &&
+    // make sure threre is no active onboarding popover on the screen
+    !activePopover
```
dariakoko, line 18: We should not display two popovers on the same screen at the same time, tis is why I am adding this check here. 
From my understanding, this is the easiest ways to check if there is an active popover from the progressive onboarding. Not sure if this is the most reliable implementation.. Please, correct me if I am missing something

dariakoko: Just noticed that `isOnboardingFinished` is not a very descriptive name
Should be something like "canDisplayPopover"

src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
```diff
@@ -150,7 +161,7 @@ const ArtworksGrid: React.FC<ArtworksGridProps> = ({
     }
   }, [relay.hasMore(), relay.isLoading()])
 
-  const CreateAlertButton = () => {
+  const CreateAlertButton: React.FC = () => {
```
olerichter00, line 164: 🌟

src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
```diff
@@ -257,6 +268,11 @@ const ArtworksGrid: React.FC<ArtworksGridProps> = ({
     )
   }
 
+  const variant_a = enabled && variant === "variant-a"
+  const variant_b = enabled && variant === "variant-b"
```
olerichter00, line 272: 🐪 🐍 

```suggestion
  const variantA = enabled && variant === "variant-a"
  const variantB = enabled && variant === "variant-b"
```

src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
```diff
@@ -257,6 +268,11 @@ const ArtworksGrid: React.FC<ArtworksGridProps> = ({
...
+
+  const userDidScroll = artworks.length >= 40
```
olerichter00, line 274: This might work at the moment, but it does not precisely determine whether the user has scrolled or not. I think there are three things here we could improve:  
1. Update the logic to either involve the scroll offset or check whether the number of artworks is greater than the initial number of artworks fetched (I'm not 100% sure if this is necessary, though).  
2. Avoid [magic numbers](https://www.google.com/search?q=magics+numbers+programming).  
3. Change the variable name because the user could have already scrolled when the number of artworks is less than or equal to 40.
```suggestion
  const displayCreateAlertPrompt = artworks.length >= CREATE_ALERT_ONBOARDING_OFFSET && artworks.length > INITIAL_NUMBER_OF_ARTWORKS
```

src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
```diff
@@ -288,6 +304,7 @@ const ArtworksGrid: React.FC<ArtworksGridProps> = ({
               <ArtistArtworksFilterHeader
                 artist={artist}
                 showCreateAlertModal={() => setIsCreateAlertModalVisible(true)}
+                shouldShowCreateAlertPrompt={!!userDidScroll && variant_a}
```
olerichter00, line 307: Not sure if prompt ("a request the software makes in order to obtain information from the end user...") is the correct term here. Maybe onboarding tooltip makes more sense 🤔

src/app/Components/Artist/ArtistArtworks/ArtistArtworksFilterHeader.tsx
```diff
@@ -9,11 +9,13 @@ import { graphql, useFragment } from "react-relay"
 interface ArtistArtworksFilterProps {
   artist: ArtistArtworksFilterHeader_artist$key
   showCreateAlertModal: () => void
+  shouldShowCreateAlertPrompt?: boolean
```
olerichter00, line 12: I wonder if we can avoid prop drilling here by using the already existing `ProgressiveOnboardingModel`'s [setIsReady](https://github.com/artsy/eigen/blob/9e354b20e830bed30dd2662b896c7295886bf031/src/app/store/ProgressiveOnboardingModel.ts#L26) functionality.

src/app/store/CreateAlertPromptModel.ts
```diff
@@ -0,0 +1,33 @@
+import { action, Action } from "easy-peasy"
+
+export interface CreateAlertPromptModel {
```
olerichter00, line 3: Can we use the already existing `ProgressiveOnboardingModel` for this feature?

src/app/Components/Artist/ArtistArtworks/SavedSearchButtonV2.tsx
dariakoko: Barney confirmed this is a good enough approach - if the user interacted with the Create Alert button within 2 minutes after seeing the message or a popover we should not show the message or popover again

src/app/store/GlobalStoreModel.ts
```diff
@@ -57,6 +58,7 @@ interface GlobalStoreStateModel {
   requestedPriceEstimates: RequestedPriceEstimatesModel
   recentPriceRanges: RecentPriceRangesModel
   progressiveOnboarding: ProgressiveOnboardingModel
+  createAlertPrompt: CreateAlertPromptModel
```
MounirDhahri, line 61: models should be ideally names instead of verbs. Maybe something like `alertPrompt`

MounirDhahri: However, I don't think we should add a new model for this. It might be possible for us to use the existing visual clues model to model this. 

We already have `addClue` and `setVisualClueAsSeen`. Maybe we can find a way to save the number of times an item has been seen and combine both models into one

dariakoko:
Closing in favor of https://github.com/artsy/eigen/pull/11377

src/app/store/GlobalStoreModel.ts
dariakoko: Thanks for the review Mounir! We had a call with Ole today to implement the feature using ProgressiveOnboardingModel. You can find the new implementation here
https://github.com/artsy/eigen/pull/11377

PR #11369, opened 2025-01-09
chore: use bidder stored on sale for determining if credit card info or creating bidder info is required.

oxaudo:
This PR comes from investigating https://artsyproduct.atlassian.net/browse/EMI-2224. The issue described there is not exactly the same but in a process discovered the bug where user is required to reenter the credit card after their bid has been outbid right away

### Description

Following the pattern currently present in Force and reading the bidding info from sale on sale_artwork. Seems like there authenticatedUser is automatically applied to the data returns properly.

#### iOS
| Popover | Message |
| --- | --- |
| bidding_old  |  bidding |

### PR Checklist

- [ x ] I have tested my changes on **iOS** and **Android**.
- [ x ] I hid my changes behind a **[feature flag]**, or they don't need one.
- [ x ] I have included **screenshots** or **videos**, or I have not changed the UI.
- [ x ] I have added **tests**, or my changes don't require any.
- [ x ] I added an **[app state migration]**, or my changes do not require one.
- [ x ] I have documented any **follow-up work** that this PR will require, or it does not require any.
- [ x ] I have added a **changelog entry** below, or my changes do not require one.

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- use bidder status from sale instead of me when determining bidder registration when bidding

</details>

### DIFF
--- a/src/app/Components/Bidding/Screens/ConfirmBid.tests.tsx
+++ b/src/app/Components/Bidding/Screens/ConfirmBid.tests.tsx
@@ -981,8 +981,12 @@ describe("ConfirmBid", () => {
 
   const saleArtworkRegisteredForBidding: ConfirmBid_sale_artwork$data = {
     ...saleArtwork,
+    endAt: "2018-05-13T20:22:42+00:00",
+    extendedBiddingEndAt: null,
     sale: {
       ...baseSaleArtwork.sale,
+      live_start_at: null,
+      cascadingEndTimeIntervalMinutes: null,
       bidder: {
         id: "1234567",
       },
--- a/src/app/Components/Bidding/Screens/ConfirmBid.tests.tsx
+++ b/src/app/Components/Bidding/Screens/ConfirmBid.tests.tsx
@@ -595,6 +595,7 @@ describe("ConfirmBid", () => {
                 partner: {
                   name: "Christie's",
                 },
+                bidder: null,
                 slug: "best-art-sale-in-town",
               },
             },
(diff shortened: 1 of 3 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 57fb90add040550bf17dedd1b2003f9e278314c3
</p>

MrSltun (commented):
Looks great!! 🌟  We just might need to do a little cleanup and fix the tests! 
Thank you Oksana!

oxaudo:
@sultan - I did update the specs so that they are passing. So I think this change is safe to merge.

This change does not fix the original but reported where one enters the card and then attempts to change the price without submitting the bid first. Testing this flow seems like the card is not saved on the user before bidding is actually attempted even though it says 'save' on the card input screen. Not sure how to fix that without doing some more refactors here. However think this scenario is much less likely than the one that I'm fixing here. So probably a good update. What you think?

MrSltun (approved):

MrSltun:
I think it's a fix for an important bug in this flow, but the other issue needs a refactor, also this part of the app have been in need for some love for quite some time :)

oxaudo:
Ok. Merging here.

PR #11376, opened 2025-01-10
chore: deprecate FancyModal

MounirDhahri:
#future-friday

### Description

A few years ago, we needed to have our own fancy modal to have native like "presentation sheet" modals overlapping. We called that `FancyModal`. The latter is no longer needed since we get the same behaviour from `react-navigation` `presentation: "formsheet"` and RN `Modal` `prenstationStyle: "formSheet"|"pageSheet"`.

### Benefits of this work
- We don't need to maintain our custom Fancy Modal.
- We can move faster and safer without being afraid of breaking keyboard avoidance in the future.
- We can remove the old `FancyModalContext` and `ArtsyKeyboardAvoidingViewContext` components.

| Screen | Screenshot |
|--------|--------|
| Rarity Modal | Screenshot 2025-01-10 at 17 03 11 |
| Create Alert Modal | Screenshot 2025-01-10 at 16 57 08 |
| Add Location Modal | Screenshot 2025-01-10 at 16 53 36 |
| Share Modal | Screenshot 2025-01-10 at 17 10 56 | 
| Select Artist on Price Insights | Screenshot 2025-01-10 at 17 14 16 | 
| Abandon Modal | Screenshot 2025-01-10 at 17 13 23 | 

In this PR, I am migrating some of the screens we have to the latter.

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- Deprecate FancyModal - mounir

</details>

### DIFF
--- a/src/app/Scenes/MyProfile/MyProfileEditModal.tsx
+++ b/src/app/Scenes/MyProfile/MyProfileEditModal.tsx
@@ -4,7 +4,6 @@ import { BottomSheetScrollView } from "@gorhom/bottom-sheet"
 import { MyProfileEditModal_me$key } from "__generated__/MyProfileEditModal_me.graphql"
 import { AutomountedBottomSheetModal } from "app/Components/BottomSheet/AutomountedBottomSheetModal"
 import { buildLocationDisplay } from "app/Components/LocationAutocomplete"
-import { SNAP_POINTS } from "app/Scenes/MyCollection/Components/MyCollectionBottomSheetModals/MyCollectionBottomSheetModalArtistsPrompt"
 import {
   UserProfileFields,
   UserProfileFormikSchema,
@@ -109,13 +108,7 @@ export const MyProfileEditModal: React.FC<MyProfileEditModalProps> = ({
   }
 
   return (
-    <AutomountedBottomSheetModal
-      visible={visible}
-      onDismiss={handleDismiss}
-      enableDynamicSizing
-      snapPoints={SNAP_POINTS}
-      enableOverDrag={false}
-    >
+    <AutomountedBottomSheetModal visible={visible} onDismiss={handleDismiss} enableDynamicSizing>
       <BottomSheetScrollView keyboardShouldPersistTaps="always">
         <Box p={2}>
           <Text>{message}</Text>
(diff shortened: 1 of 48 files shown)


MrSltun (approved):
Thank you for this effort!!! 🌟🌟 
We just need to test this change properly before we release, since it touches multiple places, especially the webview

MounirDhahri:
Good point @MrSltun for now I would like to get this work out without making changes to the webview or the inquiry modals because I want to have just one for those since they're very sensitive.

The screens i touched have no inputs apart from the shipping modal which I tested it and it seems to be working fine

| Android | iOS |
|--------|--------|
| Screenshot_1736528434 | Simulator Screenshot - iPhone 16 Pro - 2025-01-10 at 18 00 29 |

ArtsyOpenSource:

- Dev changes (Deprecate FancyModal - mounir - MounirDhahri)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 6d68755eb12068c307a9a3f38a6aff796ebaa7ec
</p>

src/app/Scenes/MyProfile/MyProfileEditModal.tsx
```diff
@@ -109,7 +109,13 @@ export const MyProfileEditModal: React.FC<MyProfileEditModalProps> = ({
   }
 
   return (
-    <AutomountedBottomSheetModal visible={visible} onDismiss={handleDismiss} enableDynamicSizing>
+    <AutomountedBottomSheetModal
+      visible={visible}
+      onDismiss={handleDismiss}
+      enableDynamicSizing
+      snapPoints={SNAP_POINTS}
```
iskounen, line 116: **thought:** My understanding of bottom sheets is that they can have either (A) dynamic sizing or (B) snap points, but not both, so I'm surprised to see both here.

iskounen (dismissed):

src/app/Scenes/MyProfile/MyProfileEditModal.tsx
MounirDhahri: good point - I will keep it as it is then.

MounirDhahri:
I did some QA and noticed that the keyboard was broken on the CreateAlert modal and thought it's related to my changes. I believe it was related to the RN upgrade but good news, I fixed those.

src/app/Scenes/MyProfile/MyProfileEditModal.tsx
iskounen: It's worth testing to be sure because I haven't tried it with this PR yet. Just FYI that there is another instance of `SNAP_POINTS` being added like this in ShareSheet.

PR #11383, opened 2025-01-13
feat: Replace Touchable with Routerlink for prefetching on artwork and artist rails

olerichter00:
This PR resolves [ONYX-1461] 

### Description

This PR replaces `Touchable` with `Routerlink` for prefetching on artwork and artist rails, as well as some additional rails on the home screen. It also makes some changes to the RouterLink component and to `routes.ts`. All updates are explained in the comments.

### PR Checklist

### To the reviewers 👀

- Replace Touchable with Routerlink for prefetching on artwork and artist rails, as well as some additional rails on the home screen - ole

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[ONYX-1461]: https://artsyproduct.atlassian.net/browse/ONYX-1461?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Navigation/routes.tsx
+++ b/src/app/Navigation/routes.tsx
@@ -25,7 +25,11 @@ import { ArticleScreen } from "app/Scenes/Article/ArticleScreen"
 import { ArticlesSlideShowScreen } from "app/Scenes/ArticleSlideShow/ArticleSlideShow"
 import { ArticlesScreen, ArticlesScreenQuery } from "app/Scenes/Articles/Articles"
 import { NewsScreen, NewsScreenQuery } from "app/Scenes/Articles/News/News"
-import { ArtistQueryRenderer, ArtistScreenQuery } from "app/Scenes/Artist/Artist"
+import {
+  ArtistQueryRenderer,
+  ArtistScreenQuery,
+  defaultArtistVariables,
+} from "app/Scenes/Artist/Artist"
 import { ArtistArticlesQueryRenderer } from "app/Scenes/ArtistArticles/ArtistArticles"
 import { ArtistSeriesQueryRenderer } from "app/Scenes/ArtistSeries/ArtistSeries"
 import { ArtistSeriesFullArtistSeriesListQueryRenderer } from "app/Scenes/ArtistSeries/ArtistSeriesFullArtistSeriesList"
@@ -192,7 +196,8 @@ export type ModuleDescriptor<T = string> = {
   name: T
   Component: React.ComponentType<any>
   options?: ViewOptions
-  Queries?: GraphQLTaggedNode[]
+  queries?: GraphQLTaggedNode[]
+  queryVariables?: object[]
   injectParams?: (params: any) => any
 }
 
@@ -215,7 +220,7 @@ export const artsyDotNetRoutes = defineRoutes([
         headerShown: false,
       },
     },
-    Queries: [homeViewScreenQuery],
+    queries: [homeViewScreenQuery],
   },
   {
     path: "/about",
@@ -237,7 +242,7 @@ export const artsyDotNetRoutes = defineRoutes([
         headerShown: false,
       },
     },
-    Queries: [activityHeaderQuery, activityContentQuery],
+    queries: [activityHeaderQuery, activityContentQuery],
   },
   {
     path: "/notification/:notificationID",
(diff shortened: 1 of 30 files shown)


src/app/Navigation/routes.tsx
```diff
@@ -192,7 +196,8 @@ export type ModuleDescriptor<T = string> = {
   name: T
   Component: React.ComponentType<any>
   options?: ViewOptions
-  Queries?: GraphQLTaggedNode[]
+  queries?: GraphQLTaggedNode[]
+  queryVariables?: object[]
```
olerichter00, line 200: Renaming `Queries` to `queries` and adding `queryVariables` to enable defining default variables for each query (e.g., for sorting or default filters). This is important to prefetch queries with the exact same variables as the screen.

src/app/Scenes/ViewingRoom/Components/ViewingRoomArtworkRail.tsx
```diff
@@ -36,6 +36,7 @@ export const ViewingRoomArtworkRail: React.FC<ViewingRoomArtworkRailProps> = ({
       </Flex>
       <ArtworkRail
         artworks={artworks}
+        itemHref={(artwork) => `/viewing-room/${viewingRoom.slug}/${artwork.slug}`}
```
olerichter00, line 39: The viewing room artwork rail links to the viewing room artwork screen instead of the artwork screen and uses the `itemHref` prop to customize the linking.

src/app/Navigation/routes.tsx
olerichter00: I will have a look to see if we can specify the default variables within the query instead. This should work in most cases and would be cleaner.

src/app/utils/loggers.ts
```diff
@@ -5,7 +5,7 @@ let metaflags = {
   logNavigation: false,
   logNotification: false,
   logOperation: false,
-  logPrefetching: false,
+  logPrefetching: true,
```
MounirDhahri, line 8: I think we should keep this `false` by default

src/app/Navigation/routes.tsx
MounirDhahri: good idea

MounirDhahri (approved):
Great work! I left a small comment about `routes.ts` for a code preference. It's not blocking though

src/app/utils/loggers.ts
olerichter00: Thanks 🙏 I set it in `metaflags.json` instead, which overrides the defaults and is ignored by Git.

olerichter00:
> **suggestion:**
How about we combine the `queries` and `queryVariables` at a single variable to avoid having to match two arrays.
```typescript
queries: [{name: Screen1Query}, {name: Screen2Query, variables: Screen2Variables }]
```
instead of 
```typescript
queries: [Screen1Query, Screen2Query]
querVariables: [null, Screen2Variables]
```

@MounirDhahri, this is a very good idea. The reason I went with the "simpler" solution is that adding `querVariables` is quite an edge case (only one case so far), and I wanted to keep the route definition as concise as possible. I know it's not very clean, but in almost all cases, it will look like

```
queries: [Screen1Query, Screen2Query]
```

instead of (`query` instead of `name` because we're defining the query itself).

```
queries: [{ query: Screen1Query }, { query: Screen2Query }]
```

Please let me know what you think. I would also be okay with your solution and my urge to keep it concise might not be the best idea.

src/app/Navigation/routes.tsx
olerichter00: I've added comments here: https://github.com/artsy/eigen/pull/11383/commits/d53dd73762f483a91308ba5f44507193798c8dc7

MounirDhahri: Sounds good to me - let's keep an eye on the API here in case more stuff gets added

olerichter00: Yeah, if we need to add more `defaultVariables` in the future or more values, the object should be updated. 👍

src/app/Navigation/routes.tsx
```diff
@@ -336,7 +347,8 @@ export const artsyDotNetRoutes = defineRoutes([
         headerShown: false,
       },
     },
-    Queries: [ArtistScreenQuery],
+    queries: [ArtistScreenQuery],
+    queryVariables: [defaultArtistVariables],
```
damassi, line 351: 👍

PR #11404, opened 2025-01-16
fix: fix progressive onboarding is not showing up

dariakoko:
This PR resolves [] 

### PR Checklist

### To the reviewers 👀

- fix progressive onboarding is not showing up

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>


dariakoko:
I am removing the isActive variable because for me it was always false because isActivePopover is always undefined. If you prefer to keep it, would love to hear your thoughts on how to make it work better 🙏

src/app/Components/ProgressiveOnboarding/ProgressiveOnboardingSaveAlert.tests.tsx
```diff
@@ -54,7 +54,7 @@ describe("ProgressiveOnboardingSaveAlert", () => {
     __globalStoreTestUtils__?.injectState({
       progressiveOnboarding: {
         sessionState: { isReady: true },
-        dismissed: [{ key: "save-highlight", timestamp: Date.now() }],
+        dismissed: [{ key: "offer-settings", timestamp: Date.now() }],
```
olerichter00, line 57: 👀

src/app/Components/ProgressiveOnboarding/ProgressiveOnboardingSaveAlert.tsx
```diff
@@ -16,16 +20,17 @@ export const ProgressiveOnboardingSaveAlert: React.FC = ({ children }) => {
...
     !!progressiveOnboardingAlerts &&
     isFocused
-  const { isActive, clearActivePopover } = useSetActivePopover(isDisplayable)
+
+  const { clearActivePopover } = useSetActivePopover(isDisplayable)
 
   const handleDismiss = () => {
     setIsReady(false)
-    dismiss("alert-create")
+    dismiss(PROGRESSIVE_ONBOARDING_ALERT_CREATE)
```
olerichter00, line 33: 🙏

olerichter00 (commented):
Thanks for all the code cleaning! 🧹

Do you know the purpose of `isActive`? Understanding would be great to avoid issues when we remove it.

src/app/Components/ProgressiveOnboarding/ProgressiveOnboardingSaveAlert.tests.tsx
dariakoko: save-highlight is no longer used, the previously dismissed popover has to be offer-settings

dariakoko:
Sure @olerichter00, isActive is coming from [useSetActivePopover.ts](https://github.com/artsy/eigen/blob/main/src/app/Components/ProgressiveOnboarding/useSetActivePopover.ts) and is used to tell whether there is an active popover on the app. It was done to make sure we don't show 1+ popovers on the screen at the same time. Somehow for me `isActive` is always `false` because `activePopover` variable from `useSetActivePopover` is always `undefined`. I couldn't figure out why and I suspect that it could be a bug that might have been introduced due to the laters additions to the progressive onboarding. I am not completely sure whether it's fine to remove it of if there is another way, would be open to hear feedbacks

src/app/Components/ProgressiveOnboarding/ProgressiveOnboardingSaveAlert.tests.tsx
dariakoko: or are you referring to the string used instead of the const?

ArtsyOpenSource:

- Cross-platform user-facing changes (fix progressive onboarding is not showing up - dariakoko)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 6280a9e31c4e4fbbc457a783fe61b18147602353
</p>

olerichter00:
>  I am not completely sure whether it's fine to remove it of if there is another way, would be open to hear feedbacks

Thanks for the explanation 🙏 

If `isActive` prevents the display of two tooltips at the same time, we may need to fix it, find an alternative solution, or ensure that this cannot happen without an additional mechanism to prevent it.

I would be okay with removing the check for the upcoming release to display tooltips again. But then we should definitely look into it more.

dariakoko:
@olerichter00 agree. I think for now the alternative solution would be making sure that popovers that can be displayed on the same screen are dismissed. We can do it manually. We have only one popover that can conflict with other - `save-artwork`. I added a check to two screens that can display this popover to make sure we do not proceed with the onboarding before it is dismissed.

olerichter00:
@dariakoko Ok, for now let's test manually to make sure that the case of more than one tooltip at the same time cannot happen. But I think we need to have a general mechanism in the onboarding to prevent it from happening in the future when we add more onboarding tooltips.

PR #11420, opened 2025-01-21
fix: Modernize iOS widget config

jonallured:
This PR fixes a number of rough edges with our iOS widget.

<details><summary>screenshots</summary>

sabretooth-000035@2x
sabretooth-000034@2x

</details> 

## contentMarginsDisabled

This configuration turns off an annoying thing where Apple is attempting to limit the developer to a "safe" content area. I guess I see what they are going for here but for us it's all wrong. We want our widget area to extend all the way to the edge and this configuration restores that layout.

## containerBackgroundRemovable

This configuration is used by iOS to decide how to render the widget in various scenarios where a background does not make sense. Think the Lock Screen - those widgets should not have a background. In our case we don't support the Lock Screen and further we do not want our background to ever be removed. Setting this to false ensures we always render correctly.

## containerBackground

This view modifier is only available in iOS 17 or later but is required to be called when it _is_ available. For us the previous setting ensures the background will not be removed but the view doesn't know this. In order to preview the widget in Xcode we have to call this. Because the minimum iOS version is 16 I had to wrap this call with availability checks and so my solution was to extract a function for the actual body of the widget. Once we are able to drop support for iOS 16 then I can reverse this and simplify the views.

/cc @artsy/amber-devs

### DIFF
--- a/ios/Artsy.xcodeproj/project.pbxproj
+++ b/ios/Artsy.xcodeproj/project.pbxproj
@@ -1676,14 +1676,14 @@
 			children = (
 				CB4326B927E8FAD900B29DD6 /* ArtsyWidgets.swift */,
 				CB4326AD27E8FAD900B29DD6 /* Assets.xcassets */,
+				CB43268B27E8FA6B00B29DD6 /* Assets.xcassets */,
 				CB43269D27E8FAD900B29DD6 /* FeaturedArtworks */,
 				CB4326AE27E8FAD900B29DD6 /* Fixtures.swift */,
 				CB43269627E8FAD800B29DD6 /* FullBleed */,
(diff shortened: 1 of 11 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against e9ca7c45dfa102d29fd245ba3e06f75167756835
</p>

jonallured:
Gonna merge so I can see this in the beta. LMK if there are any follow ups!

PR #11428, opened 2025-01-22
chore(tests): end to end testing again

brainbicycle:
This PR resolves [] 

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>


src/app/store/AuthModel.ts
```diff
@@ -992,9 +992,10 @@ export const getAuthModel = (): AuthModel => ({
     SegmentTrackingProvider.identify?.(undefined, { is_temporary_user: 1 })
 
     await Promise.all([
-      Platform.OS === "ios"
-        ? await LegacyNativeModules.ArtsyNativeModule.clearUserData()
-        : Promise.resolve(),
+      // TODO: Figure out why this is breaking for Maestro
+      // Platform.OS === "ios"
+      //   ? await LegacyNativeModules.ArtsyNativeModule.clearUserData()
+      //   : Promise.resolve(),
```
brainbicycle, line 998: fix before merge

ArtsyOpenSource:

- Dev changes (add first end to end tests - brian, george - brainbicycle)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against e8051afafd5adcab558807651a5f2df2b75461e7
</p>

e2e/config.yml
```diff
@@ -1,9 +1,4 @@
-appId: ${MAESTRO_APP_ID}
+appId: net.artsy.artsy
```
araujobarret: [super-minor]: This works for iOS only; on Android, the appId is different and I see in other sections we still use this env, maybe this change was unwanted?

e2e/set_env_vars.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash 
+
+# TODO: make work for android as well 
+export MAESTRO_APP_ID=net.artsy.artsy
+export MAESTRO_TEST_EMAIL=detox+test@example.com
+export MAESTRO_TEST_PASSWORD='$###$someT0ughP4$$'
```
araujobarret: I think we should store any test credentials elsewhere tbh, and fetch during runtime

fastlane/utility_fastlane.rb
```diff
@@ -70,6 +72,49 @@
   )
 end
 
+# Build and upload iOS and Android builds to S3
+
+lane :s3_upload_ios_build do |options|
```
araujobarret, line 77: nice, it was missing in order to work on the CI, thanks for addressing that too!

install_app_in_booted_sims.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
araujobarret: Maybe host it under `scripts/utils`?

ios/Artsy/Networking/API_Modules/ArtsyAPI+DeviceTokens.m
```diff
@@ -13,6 +13,13 @@ + (AFHTTPRequestOperation *)setAPNTokenForCurrentDevice:(NSString *)token succes
     
     if (token && name) {
         NSURLRequest *request = [ARRouter newSetDeviceAPNTokenRequest:token forDevice:name];
+
+        // TODO: Is it a problem that this can be nil?
+        // Should we fail loudly? Wait for the httpClient to be ready?
+        if (!request) {
+            return nil;
+        }
```
MounirDhahri, line 21: Why is this suddenly a problem?

MounirDhahri: is it a separate bug or is it something that started to break just with e2e tests

src/app/system/devTools/useMaestroInitialization.ts
```diff
@@ -0,0 +1,37 @@
...
+      return
+    }
+
+    const args = LaunchArguments.value<MaestroLaunchArguments>()
+    const email = args.email
+    const password = args.password
+    const shouldSignOut = args.shouldSignOut
+
+    if (email && password) {
+      GlobalStore.actions.auth.signIn({
```
MounirDhahri, line 27: nice one 😄 
I think I will add a helper for this on my CLI for quick login

src/app/system/devTools/useMaestroInitialization.ts
```diff
@@ -0,0 +1,37 @@
...
+  password?: string
+  shouldSignOut?: boolean
+}
+
+export const useMaestroInitialization = () => {
+  const isLoggedIn = GlobalStore.useAppState((state) => !!state.auth.userAccessToken)
+  const isHydrated = GlobalStore.useAppState((state) => state.sessionState.isHydrated)
+
+  useEffect(() => {
+    if (!ArtsyNativeModule.isBetaOrDev || !isHydrated) {
```
MounirDhahri, line 17: **nitpick:** probably not too important here. maybe we could also add `isNavigationReadyHere` as well as `isLoggedIn`

MounirDhahri (approved):
LGTM! I am happy we are finally doing this. 

The more use this gets in the future, the more we would have to think to run these tests only on merge to main, but that's a problem for another day.

src/app/system/devTools/useMaestroInitialization.ts
```diff
@@ -0,0 +1,37 @@
...
+        oauthProvider: "email",
+        oauthMode: "email",
+        email,
+        password,
+      })
+    } else if (shouldSignOut && isLoggedIn) {
+      GlobalStore.actions.auth.signOut()
+    }
+  }, [isHydrated])
+}
```
araujobarret, line 37: I understand the reasoning of this to speed up the tests 100%, timewise it's super expensive, but I'm more defensive of reusing accounts for some tasks because of shared state between test scenarios.
For example, using account X, we enter test scenario 1, which fails and leaves the account in a bad state. Consequently, test scenario 2 also fails due to the current state left by test scenario 1.
Reusing state in e2e tests usually introduces flakiness, so I wonder if we really should do it 🤔
It's a tradeoff of speed vs flakiness, I don't think it's a blocker but my fear is the pipeline gets flaky at some point and we just start ignoring it because the noise 🦺

e2e/flows/signup.yml
```diff
@@ -0,0 +1,21 @@
...
+- inputText: ${output.signup.email}
+- tapOn: "Continue.*"
+- assertVisible: "Welcome to Artsy"
+- inputText: ${output.signup.password}
+- tapOn: "Continue.*"
+- inputText: "Test McTest"
+- runFlow:
+    commands:
+      - tapOn:
+          point: 15%,36%
```
araujobarret, line 20: been there, done that 😢 
I can help poking around our checkboxes to make them fully accessible to use with maestro if you want, I totally forgot about it 🫠

ios/Artsy/Networking/API_Modules/ArtsyAPI+DeviceTokens.m
brainbicycle: it only got revealed because with the fast logout we are trying to use the NativeModule to clearUserData really early in the lifecycle so the APIClient isn't ready, I think in real world this should not happen

src/app/system/devTools/useMaestroInitialization.ts
brainbicycle: we can add isNavigationReady but actually took out isLoggedIn deliberately because if `isLoggedIn` = true and at app start and `shouldSignOut` we log out, then the test logs back in and `isLoggedIn` changes to true again causing the hook to rerun causing us to log out again.

src/app/system/devTools/useMaestroInitialization.ts
brainbicycle: i hear the concern and definitely don't want flaky tests either but I am not sure reusing accounts is reasonably avoidable, we are gonna add a huge time cost to every scenario and it will make writing tests more costly because you will need to wait for manual sign up every time or at least account for both signed out and signed in state in your flow writing. 

We could possibly:
- have multiple accounts for different flows
- have some guidelines about relying on account state for tests
- have nightly jobs to reset account states or even create new accounts in staging for tests

I am in agreement we want to avoid flakiness as much as possible though.

e2e/flows/signup.yml
brainbicycle: I actually haven't tried updating this since way back, I think George made some changes here recently, will check if there is already a better way

src/app/system/devTools/useMaestroInitialization.ts
araujobarret: Yeah, it takes a lot of time just from one account to get to a point where you can go somewhere and test the flow.

I like the idea of `having multiple accounts for different flows`, about the guidelines is very broad imo, all dev, QA engineers creating scenarios should be aware to avoid any reusage of other scenarios' state, any overflow should be mitigated, but I don't know how to recommend that.
For example, let's say a test where we create a convo, and we have another scenario that opens a convo and reply to something there, the other scenario shouldn't rely on the `top conversation` to reply, so these kind of things are flags to us reviewers to observe when reviewing any e2e PR, not only Maestro's, Ingetrity too 👍 

We can keep the idea of helpers to reset the account state(follows, etc), maybe more stuff to the dev menu so we can have some cleanup when a scenario finishes or breaks, we can evolve this idea as we go, I believe.

Thanks for the response, as I said I liked the ideas you suggested, I think it's def gonna reduce any flakiness level here 💯

e2e/flows/signup.yml
gkartalis: I did make the checkboxes accessible already

brainbicycle: i checked in maestro studio and still have the problem that all accessible touchables lead to link opening rather than checkbox getting checked, will take a look after this pr

PR #11430, opened 2025-01-22
fix(ONYX-1459): fix progressive onboarding and add new steps

dariakoko:
This PR resolves [ONYX-1459]  and fixed progressive onboarding

### Description

This PR fixes progressive onboarding and adds two additional steps.

What fixed progressive onboarding:
- using isActive consistently for every onboarding popover 
- fixing useDismissSavedArtwork hook - it was missing `setActivePopover(undefined)` action
- removing steps that are no longer in use
 
Additional steps are Create Alert Reminder 1 and Create Alert Reminder 2
- Create Alert Reminder 1 should be shown after the sort and filet onboarding is dismissed
- Create Alert Reminder 2 should be shown 7 days after the Create Alert Reminder 1 was dismissed
- If after seeing the Create Alert Reminder 1 the user interacts with the Create Alert CTA dismiss Create Alert Reminder 2

https://github.com/user-attachments/assets/4a0d7853-c25e-48a7-8570-02a8e3d75a6b

### PR Checklist

### To the reviewers 👀

- fix progressive onboarding, add Create Alert Reminder onboarding step

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[ONYX-1459]: https://artsyproduct.atlassian.net/browse/ONYX-1459?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/Components/ProgressiveOnboarding/useSetActivePopover.ts
```diff
@@ -1,46 +1,24 @@
...
-    return blurListener
-  }, [isDisplayable, isFocused, navigation])
-
-  useEffect(() => {
-    const focusListener = navigation.addListener("focus", () => {
-      setIsPopoverVisible(true)
-    })
-
-    return focusListener
-  }, [isDisplayable, isFocused, navigation])
```
dariakoko, line 34: @gkartalis I know that you added this two useEffects on purpose, unfortunately, I could not make them work which cause to isPopoverVisible always being false. Let me know if you want to bring them back, we could do together it over a call

ArtsyOpenSource:

- Cross-platform user-facing changes (fix progressive onboarding, add Create Alert Reminder onboarding step - dariakoko)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 3188af7fbcbb00b8afeb9c1983f66e9038cb9649
</p>

src/app/store/ProgressiveOnboardingModel.ts
```diff
@@ -65,13 +65,6 @@ export const getProgressiveOnboardingModel = (): ProgressiveOnboardingModel => (
 
 // Saves
 export const PROGRESSIVE_ONBOARDING_SAVE_ARTWORK = "save-artwork"
-export const PROGRESSIVE_ONBOARDING_FIND_SAVED_ARTWORK = "find-saved-artwork"
-export const PROGRESSIVE_ONBOARDING_SAVE_HIGHLIGHT = "save-highlight"
-export const PROGRESSIVE_ONBOARDING_SAVE_CHAIN = [
-  PROGRESSIVE_ONBOARDING_SAVE_ARTWORK,
-  PROGRESSIVE_ONBOARDING_FIND_SAVED_ARTWORK,
-  PROGRESSIVE_ONBOARDING_SAVE_HIGHLIGHT,
-]
```
dariakoko, line 74: **note:** `"find-saved-artwork"` and `"save-highlight"` are no longer in use. Remowing it and replacing with `"save-artwork"` to maintain the order of the onboarding popovers

src/app/store/ProgressiveOnboardingModel.ts
```diff
@@ -97,17 +90,25 @@ export const PROGRESSIVE_ONBOARDING_PARTNER_OFFER_CHAIN = [
 // Act Now Tasks
 export const PROGRESSIVE_ONBOARDING_ACT_NOW_TASKS = "act-now-tasks"
 
+// Alerts Reminder
+export const PROGRESSIVE_ONBOARDING_ALERT_REMINDER_1 = "alert-create-reminder-1"
+export const PROGRESSIVE_ONBOARDING_ALERT_REMINDER_2 = "alert-create-reminder-2"
+export const PROGRESSIVE_ONBOARDING_ALERT_REMINDER_CHAIN = [
+  PROGRESSIVE_ONBOARDING_ALERT_REMINDER_1,
+  PROGRESSIVE_ONBOARDING_ALERT_REMINDER_2,
+] as const
```
olerichter00, line 99: Is this used anywhere in the code?

src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
```diff
@@ -285,10 +297,21 @@ const ArtworksGrid: React.FC<ArtworksGridProps> = ({
...
-              />
+              <Flex flexDirection="row">
+                <ProgressiveOnboardingAlertReminder visible={!!shouldShowCreateAlertReminder} />
+                <Flex flex={1}>
+                  <ArtistArtworksFilterHeader
+                    artist={artist}
+                    showCreateAlertModal={() => {
+                      if (shouldShowCreateAlertReminder) {
+                        dismissChainOfRemindersAfterInteraction()
+                      }
```
olerichter00, line 308: Can't we remove this logic completely? The second reminder should not show up within 7 days after the user has dismissed the first reminder, right?

src/app/Components/ArtworkGrids/InfiniteScrollArtworksGrid.tsx
```diff
@@ -339,7 +338,7 @@ const InfiniteScrollArtworksGrid: React.FC<Props & PrivateProps> = ({
         const displayToolTip =
           enableMyCollectionSellOnboarding &&
           isMyCollection &&
-          !isDismissed(PROGRESSIVE_ONBOARDING_MY_COLLECTION_SELL_THIS_WORK).status &&
+          !isDismissed("my-collection-sell-this-work").status &&
```
olerichter00, line 341: Shouldn't we use the constants instead of the strings everywhere? 

```suggestion
          !isDismissed(PROGRESSIVE_ONBOARDING_MY_COLLECTION_SELL_THIS_WORK).status &&
```

src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
olerichter00: Or maybe I don't fully understand what this is doing....

src/app/store/ProgressiveOnboardingModel.ts
dariakoko: You're right, it can be removed

src/app/Components/ArtworkGrids/InfiniteScrollArtworksGrid.tsx
dariakoko: As Carlos mentioned [here](https://github.com/artsy/eigen/pull/11404#discussion_r1918585174) it's okay either way because it can be auto-completed by IDEs. I prefer a string as it is more readable imo

src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
dariakoko: The second reminder should be shown in 7 days after the first reminder was dismissed. If the user interacts with the Create Alert button shortly after dismissing the first reminder we should dismiss the second one too

I added a condition [here](https://github.com/artsy/eigen/pull/11430/commits/1488abe856b3eac9268aa805160db5728094a25d) - if the second reminder is already dismissed exit the function

olerichter00: Thanks for the explanation 🙏 

I wonder if we could simplify this and dismiss all "Create Alert" reminders whenever the user enters clicks on "Create Alert" on the Artwork screen? 

I suppose this 2-minute rule within `dismissChainOfRemindersAfterInteraction` tries to detect whether the user has clicked on "Create Alert" because of the reminder or not. Although this does not sound wrong, in my opinion, this distinction is not really necessary, and removing it would reduce a lot of complexity and make the code way easier to understand.

dariakoko: I love it and totally agree. It also mimics the approach we took with the save-artwork popover. I will double-check with Barney and adjust. Want to double check because [on Figma](https://www.figma.com/design/zRKGxOWHhq3kcNiNAZCig3/OtH-App%3A-Alerts-and-Activity?node-id=391-10702&m=dev) is says to hide _after_ interaction

olerichter00: > Want to double check because [on Figma](https://www.figma.com/design/zRKGxOWHhq3kcNiNAZCig3/OtH-App%3A-Alerts-and-Activity?node-id=391-10702&m=dev) is says to hide after interaction

👍 

I would say that "after interaction" does not have to be shortly after the interaction.

dariakoko: I confirmed this point with Barney earlier actually, he agreed that 2 minutes is a good time interval

olerichter00: If the product decision is to keep the 2-minute rule, I'm okay with it and will disagree and commit :)

I still believe we should remove the 2-minute rule if possible, as it would simplify implementation and make it easier to understand and test. In my opinion, we should only implement the logic if there are very good reasons not to dismiss the banner outside of this time interval.

src/app/Components/ProgressiveOnboarding/useDismissAlertReminder.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  const dismissChainOfReminders = () => {
+    dismiss("alert-create-reminder-1")
+    dismiss("alert-create-reminder-2")
+  }
+
+  return {
+    isDisplayable,
+    dismissSingleReminder,
+    dismissChainOfReminders,
```
olerichter00, line 43: Perhaps we can rename these helper functions to make their purposes clearer outside the context of this hook. For example, we could name them `dismissNextCreateAlertReminder` and `dismissAllCreateAlertReminder`.

dariakoko:
I will go ahead and merge this one
_cannot believe I'm saying this finally 😄_

src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
dariakoko: [Slack thread](https://artsy.slack.com/archives/C05EQL4R5N0/p1737639653334309) - decided to proceed with: keep showing reminders with at least 7 days in between until the user has seen the reminder two times or has interacted with the 'Create Alert' button

PR #11439, opened 2025-01-23
refactor(DIA-1019): pass a component to FancySwiperCard instead of JSX

iskounen:
This PR refactors FancySwiper to receive an array of React components instead of JSX. I wanted to make this change because I will want to reuse the `useSaveArtwork` hook to allow users to save artworks, and I can't use that hook for each artwork card unless it is inside of a component.

https://github.com/user-attachments/assets/fa9e22e2-9e82-4093-ab13-df9c45168246

https://github.com/user-attachments/assets/5d4b37f1-0dd5-43fb-80f8-b0fa5408ae58

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- Refactored FancySwiper to receive an array of components instead of JSX

</details>

### DIFF
--- a/src/app/Components/FancySwiper/FancySwiper.tsx
+++ b/src/app/Components/FancySwiper/FancySwiper.tsx
@@ -1,13 +1,13 @@
 import { Flex } from "@artsy/palette-mobile"
 import { FancySwiperIcons } from "app/Components/FancySwiper/FancySwiperIcons"
-import { useRef } from "react"
+import React, { useRef } from "react"
 import { PanResponder, Animated } from "react-native"
-import { Card, FancySwiperCard } from "./FancySwiperCard"
+import { FancySwiperCard } from "./FancySwiperCard"
 
 export const OFFSET_X = 100
 
 interface FancySwiperProps {
-  cards: Card[]
+  cards: React.ReactNode[]
   hideActionButtons?: boolean
   onSwipeLeft?: () => void
   onSwipeRight?: () => void
--- a/src/app/Components/FancySwiper/FancySwiper.tsx
+++ b/src/app/Components/FancySwiper/FancySwiper.tsx
@@ -92,7 +92,7 @@ export const FancySwiper = ({
           return (
             <FancySwiperCard
               card={card}
-              key={card.id}
+              key={index}
               swiper={swiper}
               isTopCard={isTopCard}
               {...gestureDraggers}
--- a/src/app/Scenes/ArtQuiz/ArtQuizArtworks.tsx
+++ b/src/app/Scenes/ArtQuiz/ArtQuizArtworks.tsx
@@ -19,6 +19,7 @@ import { graphql, useLazyLoadQuery, useMutation } from "react-relay"
 const ArtQuizArtworksScreen = () => {
   const queryResult = useLazyLoadQuery<ArtQuizArtworksQuery>(artQuizArtworksQuery, {})
   const { userID } = GlobalStore.useAppState((store) => store.auth)
+  const nonNullUserID = userID as string
   const { width } = useScreenDimensions()
   const space = useSpace()
   const artworks = extractNodes(queryResult.me?.quiz.quizArtworkConnection)
@@ -88,7 +89,7 @@ const ArtQuizArtworksScreen = () => {
       variables: {
         input: {
           artworkId: currentArtwork.internalID,
-          userId: userID!,
+          userId: nonNullUserID,
         },
       },
     })
@@ -140,7 +141,7 @@ const ArtQuizArtworksScreen = () => {
         variables: {
           input: {
             artworkId: previousArtwork.internalID,
-            userId: userID!,
+            userId: nonNullUserID,
             clearInteraction: true,
           },
         },
--- a/src/app/Scenes/ArtQuiz/ArtQuizArtworks.tsx
+++ b/src/app/Scenes/ArtQuiz/ArtQuizArtworks.tsx
@@ -1,17 +1,19 @@
 import { Touchable, Flex, Text, useScreenDimensions, useSpace, Screen } from "@artsy/palette-mobile"
 import { ArtQuizArtworksDislikeMutation } from "__generated__/ArtQuizArtworksDislikeMutation.graphql"
-import { ArtQuizArtworksQuery } from "__generated__/ArtQuizArtworksQuery.graphql"
+import {
+  ArtQuizArtworksQuery,
+  ArtQuizArtworksQuery$data,
+} from "__generated__/ArtQuizArtworksQuery.graphql"
 import { ArtQuizArtworksSaveMutation } from "__generated__/ArtQuizArtworksSaveMutation.graphql"
 import { ArtQuizArtworksUpdateQuizMutation } from "__generated__/ArtQuizArtworksUpdateQuizMutation.graphql"
 import { FancySwiper } from "app/Components/FancySwiper/FancySwiper"
-import { Card } from "app/Components/FancySwiper/FancySwiperCard"
 import { usePopoverMessage } from "app/Components/PopoverMessage/popoverMessageHooks"
 import { ArtQuizLoader } from "app/Scenes/ArtQuiz/ArtQuizLoader"
 import { GlobalStore } from "app/store/GlobalStore"
 import { goBack, navigate } from "app/system/navigation/navigate"
 import { extractNodes } from "app/utils/extractNodes"
 import { isEmpty } from "lodash"
-import { Suspense, useEffect, useState } from "react"
+import { Suspense, useEffect, useMemo, useState } from "react"
 import { Image } from "react-native"
 
 import { graphql, useLazyLoadQuery, useMutation } from "react-relay"
@@ -20,8 +22,6 @@ const ArtQuizArtworksScreen = () => {
   const queryResult = useLazyLoadQuery<ArtQuizArtworksQuery>(artQuizArtworksQuery, {})
   const { userID } = GlobalStore.useAppState((store) => store.auth)
   const nonNullUserID = userID as string
-  const { width } = useScreenDimensions()
-  const space = useSpace()
   const artworks = extractNodes(queryResult.me?.quiz.quizArtworkConnection)
 
   const lastInteractedArtworkIndex = queryResult.me?.quiz.quizArtworkConnection?.edges?.findIndex(
@@ -156,19 +156,12 @@ const ArtQuizArtworksScreen = () => {
     })
   }
 
-  const artworkCards: Card[] = artworks.slice(activeCardIndex).map((artwork) => {
-    return {
-      jsx: (
-        <Flex width={width - space(4)} height={500} backgroundColor="white">
-          <Image
-            source={{ uri: artwork.image?.resized?.src }}
-            style={{ flex: 1 }}
-            resizeMode="contain"
-          />
-        </Flex>
-      ),
-    }
-  })
+  const artworkCards: React.ReactNode[] = useMemo(
+    () => artworks.map((artwork, index) => <ArtQuizArtworkCard artwork={artwork} key={index} />),
+    [artworks]
+  )
+
+  const unswipedCards: React.ReactNode[] = artworkCards.slice(activeCardIndex)
 
   return (
     <Screen>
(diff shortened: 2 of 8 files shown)


src/app/Scenes/ArtQuiz/ArtQuizArtworks.tsx
```diff
@@ -155,20 +156,12 @@ const ArtQuizArtworksScreen = () => {
...
-          />
-        </Flex>
-      ),
-      id: artwork.internalID,
-    }
-  })
+  const artworkCards: React.ReactNode[] = useMemo(
+    () => artworks.map((artwork, index) => <ArtQuizArtworkCard artwork={artwork} key={index} />),
+    [artworks]
+  )
```
iskounen, line 162: I memoized this because the artwork cards were being recalculated on each swipe and causing the screen to flicker. This stopped the flickering.

ArtsyOpenSource:

- Dev changes (Refactored FancySwiper to receive an array of components instead of JSX - iskounen)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against f4839031b2be8fca32e3a273298724037537d05e
</p>

src/app/Scenes/ArtQuiz/ArtQuizArtworks.tsx
damassi: Makes sense. I was gonna ask about performance but this makes sense!

src/app/Scenes/InfiniteDiscovery/InfiniteDiscovery.tsx
```diff
@@ -221,6 +165,71 @@ const InfiniteDiscoverySpinner: React.FC = () => (
...
+>
+interface InfiniteDiscoveryArtworkCardProps {
+  artwork: InfiniteDiscoveryArtwork
+}
+
+const InfiniteDiscoveryArtworkCard: React.FC<InfiniteDiscoveryArtworkCardProps> = ({ artwork }) => {
+  const { color } = useTheme()
+  const { width: screenWidth } = useScreenDimensions()
+
+  const src = !!artwork?.images?.[0]?.url ? artwork.images[0].url : undefined
```
damassi, line 181: #superminor - There's a couple simpler ways to do this kind of thing:

If its undefined, its undefined; no need to use a ternary and then assignment: 
```tsx
const src = artwork?.images?.[0]?.url
```

src/app/Scenes/InfiniteDiscovery/InfiniteDiscovery.tsx
```diff
@@ -221,6 +165,71 @@ const InfiniteDiscoverySpinner: React.FC = () => (
...
+  const width = !!artwork?.images?.[0]?.width ? artwork.images[0].width : 0
+  const height = !!artwork?.images?.[0]?.height ? artwork.images[0].height : 0
```
damassi, line 183: as well as here

src/app/Scenes/InfiniteDiscovery/InfiniteDiscovery.tsx
```diff
@@ -221,6 +165,71 @@ const InfiniteDiscoverySpinner: React.FC = () => (
...
+
+  const size = sizeToFit({ width: width, height: height }, { width: screenWidth, height: 500 })
+
+  return (
+    <Flex backgroundColor={color("white100")} width="100%" height={800}>
+      <EntityHeader
+        name={artwork?.artistNames ?? ""}
+        meta={artwork?.artists?.[0]?.formattedNationalityAndBirthday ?? undefined}
```
damassi, line 191: does this stuff not just fall back to undefined? Like do we need to explicitly define `?? undefined` ?

damassi (dismissed):
Just a couple #superminor comments, but nothing blocking 👍

src/app/Scenes/ArtQuiz/ArtQuizArtworks.tsx
MounirDhahri: Same thought here, smart use of memoization

src/app/Scenes/ArtQuiz/ArtQuizArtworks.tsx
```diff
@@ -205,6 +198,33 @@ export const ArtQuizArtworks = () => {
...
+interface ArtQuizArtworkCardProps {
+  artwork: ArtQuizArtwork
+}
+
+const ArtQuizArtworkCard: React.FC<ArtQuizArtworkCardProps> = ({ artwork }) => {
+  const { width } = useScreenDimensions()
+  const space = useSpace()
+
+  return (
+    <Flex width={width - space(4)} height={500} backgroundColor="white">
```
MounirDhahri, line 218: **nitpick:** because this is an important number, it's maybe worth it to assign it to a constant. something like `MAX_HEIGHT` or something like that

src/app/Scenes/InfiniteDiscovery/InfiniteDiscovery.tsx
```diff
@@ -221,6 +165,71 @@ const InfiniteDiscoverySpinner: React.FC = () => (
 
 export const InfiniteDiscoveryQueryRenderer = () => <InfiniteDiscovery />
 
+type InfiniteDiscoveryArtwork = NonNullable<
+  NonNullable<
+    NonNullable<NonNullable<InfiniteDiscoveryQuery$data["discoverArtworks"]>["edges"]>[0]
+  >["node"]
+>
```
MounirDhahri, line 172: **suggestion:** you can use `ExtractNodeType` here. maybe something like this

```typescript
ExtractNodeType<InfiniteDiscoveryQuery$data["discoverArtworks"]>
```

src/app/Scenes/InfiniteDiscovery/InfiniteDiscovery.tsx
```diff
@@ -221,6 +165,71 @@ const InfiniteDiscoverySpinner: React.FC = () => (
...
+
+const InfiniteDiscoveryArtworkCard: React.FC<InfiniteDiscoveryArtworkCardProps> = ({ artwork }) => {
+  const { color } = useTheme()
+  const { width: screenWidth } = useScreenDimensions()
+
+  const src = !!artwork?.images?.[0]?.url ? artwork.images[0].url : undefined
+  const width = !!artwork?.images?.[0]?.width ? artwork.images[0].width : 0
+  const height = !!artwork?.images?.[0]?.height ? artwork.images[0].height : 0
+
+  const size = sizeToFit({ width: width, height: height }, { width: screenWidth, height: 500 })
```
MounirDhahri, line 185: you can the constant that I mentioned before here - this will make it easier to update things

src/app/Scenes/InfiniteDiscovery/InfiniteDiscovery.tsx
```diff
@@ -221,6 +165,71 @@ const InfiniteDiscoverySpinner: React.FC = () => (
...
+
+  return (
+    <Flex backgroundColor={color("white100")} width="100%" height={800}>
```
MounirDhahri, line 188: same thought like above here about `800`

src/app/Scenes/InfiniteDiscovery/InfiniteDiscovery.tsx
```diff
@@ -221,6 +165,71 @@ const InfiniteDiscoverySpinner: React.FC = () => (
...
+      <EntityHeader
```
MounirDhahri, line 189: I suggest you use `ArtistListItem` here. It's a fragment container and It has all the logic you need and it's also using `EntityHeader` under the hood.

src/app/Scenes/InfiniteDiscovery/InfiniteDiscovery.tsx
```diff
@@ -221,6 +165,71 @@ const InfiniteDiscoverySpinner: React.FC = () => (
...
+        {!!src && <Image src={src} height={size.height} width={size.width} />}
+      </Flex>
+      <Flex flexDirection="row" justifyContent="space-between" p={1}>
+        <Flex>
+          <Flex flexDirection="row" maxWidth={screenWidth - 200}>
+            {/* TODO: maxWidth above and ellipsizeMode + numberOfLines below are used to */}
+            {/* prevent long artwork titles from pushing the save button off of the card, */}
+            {/* it doesn't work as expected on Android. */}
+            <Text
+              color={color("black60")}
```
MounirDhahri, line 213: btw, this should work without using `color`
```suggestion
              color={"black60"}
```

src/app/Scenes/InfiniteDiscovery/InfiniteDiscovery.tsx
```diff
@@ -221,6 +165,71 @@ const InfiniteDiscoverySpinner: React.FC = () => (
...
+        RightButton={
+          <Button variant="outlineGray" size="small">
+            Follow
+          </Button>
+        }
+        p={1}
+      />
+      <Spacer y={2} />
+
+      <Flex alignItems="center" backgroundColor={color("purple60")}>
```
MounirDhahri, line 203: no need for this here
```suggestion
      <Flex alignItems="center" backgroundColor={"purple60"}>
```

MounirDhahri (approved):
LGTM, just minor comments.

src/app/Scenes/InfiniteDiscovery/InfiniteDiscovery.tsx
MounirDhahri: It is so useful 
https://github.com/artsy/eigen/blob/e491f493b3a481c928df6baa3ff608c461cbbd7c/src/app/utils/relayHelpers.ts#L16

src/app/Scenes/InfiniteDiscovery/InfiniteDiscovery.tsx
iskounen: Yeah, I needed to explicitly define fallback to undefined for `meta` and `imageUrl` in this component because those were typed as `string | undefined` without `null`.

src/app/Scenes/ArtQuiz/ArtQuizArtworks.tsx
iskounen: I will follow up on the sizing of the images in a follow-up PR. Ryan would like the header and footer for each card to always be at the same position (so that users can easily access the save ♡ button with their thumbs) so I need to figure out how to make the artwork always take up the same vertical space.

src/app/Scenes/InfiniteDiscovery/InfiniteDiscovery.tsx
iskounen: This was a surprisingly easy change to make, and I got the Follow behavior for free! The only issues I see are that that the artist icons are larger and different...

| Before | After |
|--------|------|
| Simulator Screenshot - iPhone 16 Plus - 2025-01-29 at 06 59 48 | Simulator Screenshot - iPhone 16 Plus - 2025-01-29 at 06 59 17 |
| Simulator Screenshot - iPhone 16 Plus - 2025-01-29 at 06 59 51 | Simulator Screenshot - iPhone 16 Plus - 2025-01-29 at 07 00 02 |

src/app/Components/FancySwiper/FancySwiper.tsx
```diff
@@ -92,7 +92,7 @@ export const FancySwiper = ({
           return (
             <FancySwiperCard
               card={card}
-              key={card.id}
+              key={index}
```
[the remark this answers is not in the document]
MounirDhahri: maybe we are not mutating the artworks array here, but if it's possible to avoid using indexes, it's a good idea to avoid them

src/app/Scenes/InfiniteDiscovery/InfiniteDiscovery.tsx
MounirDhahri: nice 🔥 I believe we had at some point a prop around the avatar size version but it might not be part of the `ArtistListItem` yet

MounirDhahri (approved):

src/app/Components/FancySwiper/FancySwiper.tsx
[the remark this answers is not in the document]
iskounen: I tried to fix this by re-introducing a key field together with the React element that is passed to the FancySwiper, but I'm not seeing the artwork cards displaying anymore.

I'm going to merge this now to unblock some other work. I'm not too worried about the indexes that we're passing because in the case of ArtTasteQuiz we are passing the position of each of the 15 artworks, which don't change. For InfiniteDiscovery, we add new artworks to the end of the list and don't change their indexes either. I agree that this should be future-proof, but I will punt on it for today.

PR #11448, opened 2025-01-28
fix(PBRW-375): navigation on the cascade bidding auction screen

dariakoko:
This PR resolves [PBRW-375] 

### Description

Fix navigation on the cascade bidding auction screen

https://github.com/user-attachments/assets/fadadc1c-c229-4830-878e-0234b9e65bad

### PR Checklist

### To the reviewers 👀

- fix navigation on the cascade bidding auction screen

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[PBRW-375]: https://artsyproduct.atlassian.net/browse/PBRW-375?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/Sale/Sale.tsx
+++ b/src/app/Scenes/Sale/Sale.tsx
@@ -137,10 +137,7 @@ export const Sale: React.FC<Props> = ({ sale, me, below, relay }) => {
   const switchToLive = () => {
     const liveBaseURL = unsafe__getEnvironment().predictionURL
     const liveAuctionURL = `${liveBaseURL}/${sale.slug}`
-    navigate(liveAuctionURL, {
-      replaceActiveScreen: true,
-      replaceActiveModal: true,
-    })
+    navigate(liveAuctionURL, {})
   }
 
   const viewConfigRef = useRef({ viewAreaCoveragePercentThreshold: 30 })
--- a/src/app/Scenes/Sale/Sale.tests.tsx
+++ b/src/app/Scenes/Sale/Sale.tests.tsx
@@ -51,10 +51,7 @@ describe("Sale", () => {
     expect(navigate).toHaveBeenCalledTimes(0)
     await waitFor(() => expect(navigate).toHaveBeenCalledTimes(1))
     await waitFor(() =>
-      expect(navigate).toHaveBeenCalledWith("https://live-staging.artsy.net/live-sale-slug", {
-        replaceActiveModal: true,
-        replaceActiveScreen: true,
-      })
+      expect(navigate).toHaveBeenCalledWith("https://live-staging.artsy.net/live-sale-slug", {})
     )
   })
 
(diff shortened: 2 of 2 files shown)


src/app/Scenes/Sale/Sale.tsx
```diff
@@ -137,10 +137,7 @@ export const Sale: React.FC<Props> = ({ sale, me, below, relay }) => {
   const switchToLive = () => {
     const liveBaseURL = unsafe__getEnvironment().predictionURL
     const liveAuctionURL = `${liveBaseURL}/${sale.slug}`
-    navigate(liveAuctionURL, {
-      replaceActiveScreen: true,
-      replaceActiveModal: true,
-    })
+    navigate(liveAuctionURL, {})
```
gkartalis, line 140: 🌟 noice! 

**question for @MounirDhahri**: do you think that we need to remove `replaceActiveScreen` and `replaceActiveModal` from everywhere in the app? Asking because I noticed this being the second PR that fixes a similar issue due to the `replaceActiveScreen` and `replaceActiveModal` in the navigate function

gkartalis (approved):

ArtsyOpenSource:

- Cross-platform user-facing changes (fix navigation on the cascade bidding auction screen - dariakoko)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 8fb58c29533fadb3e2017912170d83e7ff44116c
</p>

src/app/Scenes/Sale/Sale.tsx
MounirDhahri: Good thought but I feel like this might be a symptom of a bigger issue and we should look into fixing `replaceActiveScreen`. 

In this context, for example, navigating and replacing the active screen was a product decision and I think we should keep it as it is.

MounirDhahri (changes requested):
Thanks for working on fixing this but I feel like we should look into fixing the navigate method to avoid the same regression in other screens.

src/app/Scenes/Sale/Sale.tsx
MrSltun: @MounirDhahri I agree with the context of replacing an active screen, but currently I believe this should be removed since it blocks the users, especially on iOS, once you enter the live auction screen, there's no way to get out of it 😅

MrSltun, line 140: Maybe we could remove the empty object? 
```suggestion
    navigate(liveAuctionURL)
```

MrSltun (approved):
Thank you Daria for finding and fixing the issue 🌟🌟

src/app/Scenes/Sale/Sale.tsx
MounirDhahri: I will work on a bug and timebox it for an hour.

MounirDhahri (approved):
Okey, I had a call with @dariakoko and we checked if this was the symptom of a bigger issue with our navigation, but luckily it was not and there was just a bug that came after the navigation changes (we were not replacing routes)

olerichter00 (approved):
🌟

MounirDhahri:
Fwiw, the reason why `replaceActiveScreen` was removing the app tabs stack, is because modally presented screens, are in a separate navigator. So when we replace the stack, the previous screen, in our case the app tabs, get removed and replaced by the modally presented screen.

PR #11459, opened 2025-01-29
feat(DIA-1001): Copy changes for Explore By section

JanaeHijaz:
This PR resolves [DIA-1001] 

### Description

This PR makes a tiny copy change to the Explore By section categories

Before:

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Scenes/CollectionsByCategory/Body.tsx
+++ b/src/app/Scenes/CollectionsByCategory/Body.tsx
@@ -34,7 +34,9 @@ export const Body: React.FC<BodyProps> = ({ marketingCollections }) => {
         <Text variant="xl" px={2}>
           {category}
         </Text>
-        <Text px={2}>Explore collections with {category}</Text>
+        <Text px={2}>
+          Explore collections by {category.charAt(0).toLowerCase() + category.slice(1)}
+        </Text>
         {/* TODO: fix typings broken by some unknown reason here, prob related to @plural */}
         <CollectionsChips marketingCollections={marketingCollections as any} />
       </Flex>
--- a/src/app/Scenes/CollectionsByCategory/__tests__/Body.tests.tsx
+++ b/src/app/Scenes/CollectionsByCategory/__tests__/Body.tests.tsx
@@ -34,7 +34,7 @@ describe("Body", () => {
   it("renders", () => {
     renderWithRelay()
 
-    expect(screen.getByText(/Explore collections with mock-category/)).toBeOnTheScreen()
+    expect(screen.getByText(/Explore collections by mock-category/)).toBeOnTheScreen()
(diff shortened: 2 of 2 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (updated copy in explore by sections (one line only) - JanaeHijaz)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 05e3b23a1084f5e35d136e32a643800e24ed1e21
</p>

src/app/Scenes/CollectionsByCategory/Body.tsx
```diff
@@ -34,7 +34,9 @@ export const Body: React.FC<BodyProps> = ({ marketingCollections }) => {
         <Text variant="xl" px={2}>
           {category}
         </Text>
-        <Text px={2}>Explore collections with {category}</Text>
+        <Text px={2}>
+          Explore collections by {category.charAt(0).toLowerCase() + category.slice(1)}
```
damassi, line 38: Why not just `{category.toLowerCase()}`?

MounirDhahri: agree with that 🆙

MounirDhahri: `{category.charAt(0).toLowerCase() + category.slice(1)}` might work for cases where the "category" is just one word, but `category.toLowerCase()` is more future proof to avoid cases like
```
Explore collections by contemporary Art
```

damassi: Another note here is just to kinda feel into things intuitively (this applies to all code, in general). Something like 
```
category.charAt(0).toLowerCase() + category.slice(1)
```
That's a lot of code to "make lowercase", which "feels" like it should be simple, or that there might be a function ([lodash](https://lodash.com/docs/4.17.15#lowerCase)?) that would do it for you. 

Try to lean into the intuition a bit more, because if it seems like it should be simpler / less code, that feeling is typically 99% of the time right and its just a matter of thinking about it a bit more.

iskounen: Ooohhh, we had the intuition that it should be simpler, but found some existing code that did it this way, and I assumed that it was the right way. I'm glad that there's a `toLowerCase`!

iskounen (approved):
I'm happy that I learned JavaScript has a toLowerCase method!

src/app/Scenes/CollectionsByCategory/Body.tsx
JanaeHijaz: thanks @damassi!

PR #11473, opened 2025-02-03
feat: use npx pod-install #RFC

MounirDhahri:
### Description

This PR comes as an mini RFC to improve our `pod install` DX.

**What is this PR solving?**
More often than not, we have pod install issues, the latter come from having an out of sync pod or being in the wrong folder. This usually slows us down and quite often we have notifications from Sapphire about running `bundle exec pod install --repo-update` to fix something or major upgrades.

We could probably avoid most of the situations below (screenshot from slack) because `npx pod-install` would already do that.
Screenshot 2025-02-03 at 16 22 05

**Suggested solution**
While working on Energy migration to Expo, I was surprised that pods didn't get out of sync and somehow were always working. I noticed then that Expo is using a built-in house solution for that. Good thing, it's open source and works with all kind of projects! (even ionic). It's called `pod-install`.
**Link**: https://www.npmjs.com/package/pod-install

**What is `pod-install`?**
A tool to fix common issues developers have when running `pod install`.

**How does it do it?**
This package will do the following:

- Check if the machine is darwin.
  - If not then it'll quit with a helpful error message.
- Ensure CocoaPods CLI is installed on the machine.
  - If not then it'll try to install CocoaPods CLI, first with gem, then with homebrew.
- Check if there is an Xcode project in the current directory
  - If not then it'll try again in ios/ than macos/ directories (if any exists).
- Run pod install
  - If pod install fails because the repo is out of date, then it'll run pod repo update and try again.

**Risks**
None that comes to mind 🤷  happy to hear your thoughts, we can always roll this back.

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- use npx pod-install instead of raw pod-install - mounir

</details>


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 749e053ab59a12f961de4a20b273cb979fdefc23
</p>

.circleci/config.yml
```diff
@@ -91,7 +91,7 @@ commands:
             - v29-pods-{{ checksum ".manifests/cocoapods" }}
       - run:
           name: Install Pods
-          command: cd ios; bundle exec pod check --ignore-dev-pods || bundle exec pod install; cd ..;
+          command: cd ios; bundle exec pod check --ignore-dev-pods || bundle exec npx pod-install; cd ..;
```
damassi, line 94: `npx` totally works, but it might also prompt the user to install the package on their computer. Could we add it to `devDependencies`?

MounirDhahri: good point, I thought I've done that. thanks

gkartalis:
I would be really curious to see if it actually fixes this issue, because I noticed that the most common issue is basically podfile diffs that are usually solved like so:

```bash
➜ cd ios && bundle exec pod update DoubleConversion --no-repo-update
```

but it is not solving the issue if you run the command without explicitly specifying the pod that has the wrong hash.

```bash
➜ cd ios && bundle exec pod update --no-repo-update
```

gkartalis:
other than my comment above I don't have any other concerns, thanks for working on this

anandaroop (approved):
I think I had the same q as George — will this help with the frequent cases like [this](https://artsy.slack.com/archives/C02BAQ5K7/p1738241847264399)?

If so then I'm all for it 👍🏽

MounirDhahri:
Good note @gkartalis and @anandaroop - I also don't know if it would fix that issue, but it would be great if it does. 

It's using `update-check` [under the hood](https://github.com/expo/expo/tree/main/packages/pod-install/src) so it might work actually, but I think the issue itself should be fixed from `cocoapods`.

MounirDhahri:
Fwiw, the goal from this is not to fix issue with `DoubleConversion`, `Glog`, `Folly`  If we need to fix that, we would need to patch rn with one of the issues suggested [here](https://github.com/facebook/react-native/issues/31121)

brainbicycle (approved):
sounds reasonable to me! thank you 🙏 !

MounirDhahri:
Let's merge this and check it out. I Hope it makes our life easier

PR #11481, opened 2025-02-04
fix(PBRW-397): Dismiss modal on foreground Push Notification tap

MrSltun:
This PR resolves [PBRW-397] 

### Description
This PR fixes the issue where push notifications don't navigate the users when a webview modal is present

### Demo

| | Before | After |
|---|---|---|
| Android | <video src="https://github.com/user-attachments/assets/a875bad3-7708-4767-b6ae-1dbbb1b18796" /> | <video src="https://github.com/user-attachments/assets/df5bc43f-ef2f-4229-9cd5-aae53b230125" /> |
| iOS | <video src="https://github.com/user-attachments/assets/f3eee367-b4f4-4a4d-adcb-51243c3c64a0" /> | <video src="https://github.com/user-attachments/assets/d14367d8-de83-4495-8462-adb235238bf3" /> |

	

	
	

### PR Checklist

### To the reviewers 👀

- navigate the user on foreground push when a webview modal is visible

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[PBRW-397]: https://artsyproduct.atlassian.net/browse/PBRW-397?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/utils/PushNotification.ts
+++ b/src/app/utils/PushNotification.ts
@@ -9,7 +9,7 @@ import {
   unsafe_getUserAccessToken,
 } from "app/store/GlobalStore"
 import { PendingPushNotification } from "app/store/PendingPushNotificationModel"
-import { navigate } from "app/system/navigation/navigate"
+import { navigate, navigationEvents } from "app/system/navigation/navigate"
 import { Platform } from "react-native"
 import DeviceInfo from "react-native-device-info"
 import PushNotification, { ReceivedNotification } from "react-native-push-notification"
@@ -208,6 +208,7 @@ export const handleReceivedNotification = (
     }
     const hasUrl = !!notification.data.url
     if (isLoggedIn && hasUrl) {
+      navigationEvents.emit("requestModalDismiss")
       navigate(notification.data.url as string, {
         passProps: notification.data,
         ignoreDebounce: true,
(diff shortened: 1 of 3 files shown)


gkartalis (approved):

ArtsyOpenSource:

- Cross-platform user-facing changes (navigate the user on foreground push when a webview modal is visible - MrSltun)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 73c8686df8f33904a1c42f6c0dc434a72888252b
</p>

src/app/utils/PushNotification.ts
```diff
@@ -208,6 +208,7 @@ export const handleReceivedNotification = (
     }
     const hasUrl = !!notification.data.url
     if (isLoggedIn && hasUrl) {
+      navigationEvents.emit("requestModalDismiss")
```
MounirDhahri, line 211: I think this is happening because the order screen is presented modally so we can't navigate to whatever is not presented modally as well so the same issue should be also happening if you are in the dev menu or SWA flow. 
Maybe we can check ahead of the navigate if the active screen is presented modally, if so dismiss it

```
const isCurrentRouteModal = () => {
  const currentURL = getCurrentURL()
  const result = matchRoute(targetURL)
  const module = modules[result.module]

  return module.options?.alwaysPresentModally
}
```

We can arguably have this call within the `navigate` method but since this wasn't reported yet, let's avoid it.

MounirDhahri (approved):
we probably ne

src/app/utils/PushNotification.ts
MrSltun: I agree with your idea, but I wasn't able to make it work 🤷 would love to pair on it with you! 

but to unblock the release I'm gonna be merging this PR since it fixes the issue that was reported, please let me know if you think I shouldn't :)

PR #11498, opened 2025-02-06
chore: Replace `ElementInView` with `Sentinel`

olerichter00:
This PR resolves [ONYX-1514](https://artsyproduct.atlassian.net/browse/ONYX-1514)

### Description

This replaces t `ElementInView` with `Sentinel` (in RouterLink, Onboarding, AboutTheWorkTab) ~~[react-native-viewport](https://www.npmjs.com/package/@herberthtk/react-native-viewport)~~. Both do pretty much the same, but `ElementInView` does not always work (mostly when scrolling horizontally).

The `ElementInView` does not work reliably, especially in nested scroll views. It works for the onboarding where it was initially used, but not for prefetching.

~~I propose replacing it with [react-native-viewport](https://www.npmjs.com/package/@herberthtk/react-native-viewport), which has a similar API, similar implementation, and some extra features like a delay.~~

In this PR I remove `ElementInView` and replace it with our `Sentinel` component (thanks @MounirDhahri 🙏 ).

### PR Checklist

### To the reviewers 👀

- Replace ElementInView with Sentinel in onboarding - ole
- Replace ElementInView with Sentinel for prefetching queries on the home screen - ole

#### iOS user-facing changes

- Fix home view query prefetching (it's now more reliable) - ole

-

#### Dev changes

- Replace `ElementInView` with `Sentinel` to check whether an element is on screen in onboarding and query prefetching. - ole

</details>

[ONYX-1514]: https://artsyproduct.atlassian.net/browse/ONYX-1514?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MounirDhahri:
This came in a conversation with @damassi recently, the component we are using currently seems to not be working properly the homeview. 

@olerichter00 do you mind you try out the component which I added to the home view to see if it performs better than `IsElementInView`. You can find it here `https://github.com/artsy/eigen/blob/main/src/app/utils/Sentinel.tsx`

If it doesn't perform better, I would still avoid adding this as a package and instead add it at as a component since it's just one [file](https://github.com/herberthk/react-native-viewport/blob/main/src/index.tsx) and we would be better if we maintain it ourselves.

olerichter00:
@MounirDhahri I've tried out `Sentinel`, and it works as well as the package. Thanks for the suggestion! 🙏 

I've updated the PR and replaced the package with our `Sentinel` component. 👍

MounirDhahri (approved):

MounirDhahri:
Awesome @olerichter00 👏

olerichter00:
> I think you'd need to rebase and replace another element that uses ElementInView that got merged kinda an hour ago 😅

Thanks, I've rebased and updated 👍 

> I don't get why ElementInView isn't working, but that was initially designed to work with the ProgressiveOnboarding, so it might need some tweaks to work properly with other sections but it should be minor. My only concern is Sentinal seems to be a little bit complicated just to check if an element is in View 🤔

The main difference might be the measuring and `ElementInView` only measuring horizontally ([ElementInView](https://github.com/artsy/eigen/blob/a9cf9b285513cbfe91455ce0d986eae82b245732/src/app/utils/ElementInView.tsx#L29) vs [Sentinel](https://github.com/artsy/eigen/blob/a9cf9b285513cbfe91455ce0d986eae82b245732/src/app/utils/Sentinel.tsx#L61).

damassi (commented):
🙏

PR #11520, opened 2025-02-08
fix(DIA-1068): prevent artwork cards from freezing

iskounen:
This PR fixes a bug in Infinite Discovery where, after the first 3 swipes, the 4th card would be on top for an instant before being replaced by another card. When trying to swipe this new card that appeared, users would not be able to swipe it, so the Infinite Discovery would come to a halt. It looked like this:

https://github.com/user-attachments/assets/01942f3d-653c-43ad-97e8-e5614d8be60e

I was struggling to debug this, so I decided to pass each card's artwork ID to FancySwiper so that I could print out the ID of each artwork card in the stack as I was swiping. Lo and behold, the bug stopped happening.

The only real change that I made was to use the artwork IDs for each artwork card's `key` field, so that seems to have fixed the buggy behavior. (@araujobarret and @MounirDhahri, I am haunted by the echoes of [this thread](https://github.com/artsy/eigen/pull/11439#discussion_r1933394560)). Here is how it behaves now:

https://github.com/user-attachments/assets/aeb225d9-7390-4357-a3ea-194b1eca8fa1

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- Fixed a bug where Infinite Discovery cards were freezing

</details>

### DIFF
--- a/src/app/Components/FancySwiper/FancySwiper.tsx
+++ b/src/app/Components/FancySwiper/FancySwiper.tsx
@@ -7,8 +7,13 @@ import { FancySwiperCard } from "./FancySwiperCard"
 // the amount of swiping that is considered a full swipe
 export const SWIPE_MAGNITUDE = 100
 
+export type FancySwiperCard = {
+  content: React.ReactNode
+  artworkId: string
+}
+
 interface FancySwiperProps {
-  cards: React.ReactNode[]
+  cards: FancySwiperCard[]
   hideActionButtons?: boolean
   onSwipeAnywhere?: () => void
   onSwipeLeft?: () => void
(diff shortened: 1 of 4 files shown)


ArtsyOpenSource:

- Dev changes (Fixed a bug where Infinite Discovery cards were freezing - iskounen)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 2bd2689d8684a364fdd2e188de195082d2752468
</p>

src/app/Components/FancySwiper/FancySwiper.tsx
```diff
@@ -117,8 +122,8 @@ export const FancySwiper = ({
 
           return (
             <FancySwiperCard
-              card={card}
-              key={index}
+              card={card.content}
+              key={card.artworkId}
```
iskounen, line 126: This is the only modification that I can attribute the change in behavior to.

PR #11522, opened 2025-02-10
fix(EMI-2224): Refactor Register to Bid flow

MrSltun:
This PR resolves [EMI-2224] 

### Description

This PR fixes an issue with the credit card token not persisting in the app when bidding on an artwork before registering for the auction.
This also refactors the Auction registration flow to be able to add the fix, which should make the flow more performant

> [!NOTE]
> I kept the old code for comparison and to not break the Github diff (hence the big **addition** and **subtraction**), but will remove it before merging this change
Would love your inputs on it :)

>[!TIP]
> To review the code changes, I recommend checking the commits of this PR

### Videos
<details><summary>Videos Here</summary>

#### Before
| | Android | iOS |
|---|---|---|
| Register without CC | <video src="https://github.com/user-attachments/assets/8e8ed357-8cba-4da6-8af5-1657f56244d6" /> | <video src="https://github.com/user-attachments/assets/bcb6b162-a196-4490-8a6d-2583dd0936ba" /> |
| Register with CC | <video src="https://github.com/user-attachments/assets/56ebe674-6c86-4620-819c-b74e2fb1c7ac" /> | <video src="https://github.com/user-attachments/assets/1bfe4df9-bf9b-4ecd-b309-9d508646acae" /> |
| Bid without CC | <video src="https://github.com/user-attachments/assets/306a5d2a-3976-4e91-abeb-ca6a2b6cba2b" /> | <video src="https://github.com/user-attachments/assets/2902bb95-9f35-4d44-b06a-b98224af8af8" /> |
| Bid with CC | <video src="https://github.com/user-attachments/assets/241eb8c4-4337-441f-ba51-e5b3310f0cd6" /> | <video src="https://github.com/user-attachments/assets/316aedeb-48de-40e9-8244-995558771b0a" /> |

#### After
| | Android | iOS |
|---|---|---|
| Register without CC | <video src="https://github.com/user-attachments/assets/cd218b3b-3493-40f3-b182-46ad8163f36f" /> | <video src="https://github.com/user-attachments/assets/23fee497-ab97-48ef-8ed2-ba3866a7fe81" /> |
| Register with CC | <video src="https://github.com/user-attachments/assets/070429dc-10b0-402d-8c21-de5f23893428" /> | <video src="https://github.com/user-attachments/assets/ca092f06-fec5-4bf5-9886-ce9b80e154e7" /> |
| Bid without CC | <video src="https://github.com/user-attachments/assets/612aa860-588d-4576-9e9f-c9e5c29edd57" /> | <video src="https://github.com/user-attachments/assets/4520eec3-eb7f-4897-86ba-6aeb0193a062" /> |
| Bid with CC | <video src="https://github.com/user-attachments/assets/d3def0b1-4aff-46c2-b73f-7c4a6640eba9" /> | <video src="https://github.com/user-attachments/assets/a5d89ff9-a182-49ea-b5e8-cf898c1fef1f" /> |

</details> 

### Performance comparison
This comparison was done using [Flashlight](https://github.com/bamlab/flashlight) on my Android device (OnePlus 8 Pro)
The comparison was made between the Firebase beta builds `1674646046`(before) and `1674646044`(after)
I tested each flow 3 times on each build (18 test in total)

> [!NOTE]
> Flashlight calculates the CPU usage of all cores, my device has 8 so 100% would be 800%, so I normalized it by dividing by 8, aka 159% would be 19.88%

#### Updating Max Bid
| Metric | Before | After |
|---|---|---|
| Avarage FPS | `57.27 FPS` | `56.97 FPS` 🔴 |
| Avarage CPU Usage | `14.10%` | `14.99%` 🔴 |
| Avarage RAM Usage | `576.17 MB` | `255.50 MB` 🟢 |

#### Adding Credit Card
| Metric | Before | After |
|---|---|---|
| Avarage FPS | `39.40 FPS` | `56.80 FPS` 🟢 |
| Avarage CPU Usage | `19.75%` | `16.00%` 🟢 |
| Avarage RAM Usage | `448.17 MB` | `274.10 MB` 🟢 |

#### Whole flow (Bid on artwork without CC before registering)
| Metric | Before | After |
|---|---|---|
| Avarage FPS | `41.60 FPS` | `57.37 FPS` 🟢 |
| Avarage CPU Usage | `19.62%` | `16.36%` 🟢 |
| Avarage RAM Usage | `303.63 MB`  | `594.77 MB` 🔴 |

### PR Checklist

### To the reviewers 👀

- Refactor Auction Registration flow
- Fix removing credit card token when updating max bid 

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### TODO
- [x] Compare performance changes
- [ ] Add Sentry tracking for errors

[EMI-2224]: https://artsyproduct.atlassian.net/browse/EMI-2224?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/Components/Bidding/Context/BidFlowContextProvider.tsx
```diff
@@ -0,0 +1,52 @@
...
+  setBillingAddress: action((state, payload) => {
+    state.billingAddress = payload
+  }),
+  setCreditCardToken: action((state, payload) => {
+    state.creditCardToken = payload
+  }),
+  setBiddingEndAt: action((state, payload) => {
+    state.biddingEndAt = payload
+  }),
+})
```
damassi, line 45: 🙏

src/app/Components/Bidding/Context/BidFlowContextProvider.tsx
```diff
@@ -0,0 +1,52 @@
...
+  biddingEndAt?: string | null
+  selectedBid: Computed<this, Increments[number]>
+  setSaleArtworkIncrements: Action<this, Increments>
+  setSelectedBidIndex: Action<this, number>
+  setBillingAddress: Action<this, Address>
+  setCreditCardToken: Action<this, Token.Result>
+  setBiddingEndAt: Action<this, string | undefined | null>
+}
+
+export const getBidFlowContextStore = (): BidFlowContextModel => ({
```
MounirDhahri, line 23: This is a great intermediate step especially for refactoring, later we can move all this logic to be pulled from MP to avoid having data not in sync and decouple further these screens

src/app/Components/Bidding/Screens/CreditCardForm.tsx
```diff
@@ -33,6 +32,8 @@ interface CreditCardFormProps {
   onSubmit: (t: Token.Result, a: Address) => void
 }
 
+const MemoizedInput = memo(Input)
```
MounirDhahri, line 35: nice!. let's move this to palette-mobile though. This should be the default behaviour imo

damassi: Or use formik?

MrSltun: I don't think it works on React Native since inputs are native elements 😢

MounirDhahri: didn't know about `FastField` before. Thanks for sharing about it

ArtsyOpenSource:

- Cross-platform user-facing changes (Refactor Auction Registration flow,Fix removing credit card token when updating max bid - MrSltun)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 6064a9e2054db4c7060153680d521334733f4dcf
</p>

damassi:
Can't comment too much on the particulars, but amazing PR @MrSltun 🔥

src/app/Components/Bidding/Context/BidFlowContextProvider.tsx
```diff
@@ -0,0 +1,52 @@
+import { Token } from "@stripe/stripe-react-native"
+import { SelectMaxBid_saleArtwork$data } from "__generated__/SelectMaxBid_saleArtwork.graphql"
+import { Address, PaymentCardTextFieldParams } from "app/Components/Bidding/types"
+import { action, Action, Computed, computed, createContextStore } from "easy-peasy"
+
+type Increments = NonNullable<NonNullable<SelectMaxBid_saleArtwork$data["increments"]>[number]>[]
```
damassi, line 6: 👍 👍 Thanks for pulling this from the generated type

src/app/Components/Containers/BiddingNavigator.tsx
```diff
@@ -0,0 +1,89 @@
...
+        />
+        <BiddingNavigationStack.Screen name="PhoneNumberForm" component={PhoneNumberForm} />
+        <BiddingNavigationStack.Screen name="CreditCardForm" component={CreditCardForm} />
+        <BiddingNavigationStack.Screen name="ConfirmBid" component={ConfirmBid} />
+        <BiddingNavigationStack.Screen name="BidResult" component={BidResult} />
+        <BiddingNavigationStack.Screen name="RegistrationResult" component={RegistrationResult} />
+      </BiddingNavigationStack.Navigator>
+    </NavigationContainer>
+  )
+}
```
damassi, line 89: @MrSltun, @MounirDhahri - curious to get y'alls take here. (I'm mostly referencing Force and the ease in which we can discover most things, at a glance, just from viewing the file tree).

This bidding navigator is effectively an "app" with "routes", but its buried in `components/containers`. While this follows existing patterns, its still quite confusing at a glance, when compared to how we can navigate [all around in folio](https://github.com/artsy/energy/tree/main/src/apps) (as an imperfect example). 

It'd be great if we could lay down some patterns to help here, as i think it could naturally simplify the app and make it more tree like -- apps contain routes and screens; makes sense to colocate and make self-describing. 

No need to address here! just figured this is a good example of what I mean and curious about y'alls take.

gkartalis (dismissed):
looks awesome, left some minor comments about some todos but other than that THANK you

src/app/Components/Containers/BiddingNavigator.tsx
MounirDhahri: Very good comment. I am going to add a comment about that to the practice mobile to discuss this together further as a team.
We could definitely do a better job in our [documentation](https://github.com/artsy/eigen/blob/main/docs/adding_a_new_route.md) to limit the use of `NavigationContainer` in Eigen. In almost all cases, it's not really required and if it's there, it's a symptom of high coupling.

MounirDhahri: Decoupling the screens at the level of this PR might be so hard. I do though believe that these screens would benefit a lot from being decoupled from the context store that is wrapping them and be independent routes at some point in the future. 

The main issue that made this screens hard to maintain, along with the navigation, is the state management and how it's being synchronized across the different screens. I believe the synching issue would be entirely solved if we make each of the screens an independent `QueryRenderer` that queries for data itself instead of rely on other screens passing it. It will also give us more confidence in the future to make changes and make QA easier

MounirDhahri (approved):
This is amazing! Great job.

src/app/Components/Containers/BiddingNavigator.tsx
MrSltun: Great point! I think it makes sense to move it to `navigation` instead of `components/containers` since it's just a navigation container and not a component container, I will update that

gkartalis (approved):
so happy to see flashlight being used! thanks for the measures!

MrSltun:
Since there are no issues yet, I'll be merging this in an hour :)

PR #11651, opened 2025-03-11
chore: Introduce `RouterButton` component

olerichter00:
This PR resolves [ONYX-1493] 

### Description

this PR introduces a `RouterButton` component (similar to [RouterLink](https://github.com/artsy/eigen/blob/14113b8c405646512ae7d89537434a4ba3c4cd8c/src/app/system/navigation/RouterLink.tsx#L24)) to support prefetching button links.

To achieve this without duplicating code, I created a generic `RouterLinkComponent` that can be adjusted to use a different Touchable wrapper.

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- Add `RouterButton` component to support prefetching button links - ole

</details>

[ONYX-1493]: https://artsyproduct.atlassian.net/browse/ONYX-1493?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/routing.md
+++ b/docs/routing.md
@@ -12,8 +12,23 @@ The [RouterLink] component wraps `Touchable` and enables navigation to a specifi
 </RouterLink>
 ```
 
-Alternatively, [navigate] can be used to programmatically navigate to a given route.
+For buttons, [RouterButton] can be used, which wraps `Button`:
+
+```typescript
+<RouterButton to="/my-route">
+  {children}
+</RouterButton>
+```
+
+Alternatively, [navigate] can be used to programmatically navigate to a given route:
+
+```typescript
+<Touchable onPress={() => navigate("/my-route")}>
+  {children}
+</Touchable>
+```
 
 [routes.tsx]: /src/app/Navigation/routes.tsx
 [RouterLink]: /src/app/system/navigation/RouterLink.tsx
+[RouterButton]: /src/app/system/navigation/RouterButton.tsx
 [navigate]: /src/app/system/navigation/navigate.ts
--- a/src/app/system/navigation/RouterButton.tsx
+++ b/src/app/system/navigation/RouterButton.tsx
@@ -1,0 +1,40 @@
+import { Button, ButtonProps } from "@artsy/palette-mobile"
+import { RouterLinkProps, usePrefetchOnVisible } from "app/system/navigation/RouterLink"
+import { Sentinel } from "app/utils/Sentinel"
+import React from "react"
+
+export type RouterButtonProps = RouterLinkProps & ButtonProps
+/**
+ * `Button` wrapper component that enables navigation when pressed, using the `to` prop.
+ * It supports optional prefetching.
+ */
+export const RouterButton: React.FC<RouterButtonProps> = ({
+  disablePrefetch,
+  to,
+  prefetchVariables,
+  onPress,
+  navigationProps,
+  children,
+
+  ...restProps
+}) => {
+  const { isPrefetchingEnabled, handlePress, handleVisible } = usePrefetchOnVisible({
+    to,
+    disablePrefetch,
+    onPress,
+    navigationProps,
+    prefetchVariables,
+  })
+
+  if (!isPrefetchingEnabled) {
+    return <Button onPress={handlePress} {...restProps} children={children} />
+  }
+
+  return (
+    <Sentinel onChange={handleVisible}>
+      <Button {...restProps} onPress={handlePress}>
+        {children}
+      </Button>
+    </Sentinel>
+  )
+}
--- a/src/app/system/navigation/RouterButton.tsx
+++ b/src/app/system/navigation/RouterButton.tsx
@@ -1,40 +1,13 @@
 import { Button, ButtonProps } from "@artsy/palette-mobile"
-import { RouterLinkProps, usePrefetchOnVisible } from "app/system/navigation/RouterLink"
-import { Sentinel } from "app/utils/Sentinel"
+import { RouterLinkComponent, RouterLinkComponentProps } from "app/system/navigation/RouterLink"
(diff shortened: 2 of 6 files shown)


olerichter00:
> The abstraction per se is fine, but I think we might be increasing too much the tooling around the prefetching, that was @damassi 's request when I came up with another abstraction of having an extra hook to navigate and prefetch https://github.com/artsy/eigen/pull/11610 and I had to agree, it made sense.

In general, I agree as well, but I see the new `RouterButton` component not as another tool for prefetching, but as part of a shift towards a new pattern. We now use a `to` prop for navigation instead of calling `navigate` within an `onPress` callback. Being able to prefetch links as they enter the viewport is more like a side effect (and the main purpose for which the new pattern is introduced 😅).

> Can't you achieve this wrapping a button with <RouterLink hasChildTouchable><Button>... ?

That's a great idea! I'll have a look 👀

olerichter00:
Using an `as={}` prop instead of creating a new component could be an alternative 👍 

I'm still in favor of a separate component because I think it's easier to use and read (shorter, fewer imports, no new pattern one has to know): `<RouterButton to="/foo">` instead of `<RouterLink as={Button} to={url}>`. But I would be fine with both...

docs/routing.md
```diff
@@ -12,8 +12,23 @@ The [RouterLink] component wraps `Touchable` and enables navigation to a specifi
...
+</RouterButton>
+```
+
+Alternatively, [navigate] can be used to programmatically navigate to a given route:
+
+```typescript
+<Touchable onPress={() => navigate("/my-route")}>
+  {children}
+</Touchable>
+```
```
damassi, line 29: Perhaps we should add a recommendation here, to use RouterLink or RouterButton, as programmatic `navigate` doesn't support prefetching? (though we should support it in this function :)

src/app/system/navigation/RouterButton.tsx
```diff
@@ -0,0 +1,13 @@
+import { Button, ButtonProps } from "@artsy/palette-mobile"
+import { RouterLinkComponent, RouterLinkComponentProps } from "app/system/navigation/RouterLink"
```
damassi, line 2: We should avoid adding `Component` in file names ("FooComponent Component"). If we're using something as a "Base", perhaps something like `RouterLinkBase`?

src/app/system/navigation/RouterLink.tsx
```diff
@@ -7,30 +7,40 @@ import React, { useState } from "react"
...
+  TouchableWrapper?: React.ElementType
+  onPress?: (event: GestureResponderEvent) => void
+} & WrapperProps
 
-/**
- * Wrapper component that enables navigation when pressed, using the `to` prop.
- * It supports optional prefetching and ensures proper touch handling for nested touchable elements.
- */
-export const RouterLink: React.FC<RouterLinkProps & TouchableProps> = ({
+export function RouterLinkComponent<WrapperProps>({
```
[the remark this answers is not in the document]
damassi: Can we keep the api a bit more familiar and use `as={Button}`? Wouldn't want that to interfere with palette props, but `as=` is typically the way this is done

src/app/system/navigation/RouterLink.tsx
```diff
@@ -7,30 +7,40 @@ import React, { useState } from "react"
...
   disablePrefetch,
   to,
   prefetchVariables,
   onPress,
   navigationProps,
   children,
   hasChildTouchable,
```
damassi, line 40: Q: would we still want to support this API, or is this no deprecated? And if so, can we update @araujobarret 's component the new form?

damassi (approved):
Couple of minor comments but looks 👍

damassi:
Ah! so sorry @araujobarret and @olerichter00 - i totally didn't see the comments here before reviewing! 

Yup, i was thinking the same thing, re `as` -- that keeps the API similar to elsewhere, using polymorphism. Could even just fake it 🤷 But its something memorable and common, vs something new. 

I too share @araujobarret's concert about an expanding API. Eigen has too much code in general, so more can often be less. 

Part of what made Force's implementation successful is that we had this one thing that people used everywhere. When we (finally) got around to properly implementing prefetching, it was easy -- just enter that one component, wire it up. An hour later force was fast. So keeping it one-component like will only serve to further. 

That said, web is different than mobile, and having something like `RouterLinkBase` that we can "extend" from is nice. If we think that RouterButton is gonna be used all over and the risk of divergence from the root component isn't too high, i'm not opposed. But in general, agree with @araujobarret 's points that 1 component API is better than 2 (even if the second is basically one).

docs/routing.md
olerichter00: Good idea 👍

olerichter00:
Thanks both for your feedback! 🙏 

I'm hearing that creating new components (`RouterButton`, `RouterTextLink`, etc.) is something we should avoid in favor of another solution. 👂 

I wonder if we can implement the `as={Button}` pattern in a way that the types are correct and allow something like this:

```
<RouterLink as={Button} buttonSpecificProp={...} routerLinkSpecificProp={...} />
```

If implementing the typing correctly doesn't work, we could simply use the new [hasChildTouchable](https://github.com/artsy/eigen/blob/4ecc4beab2070ab598e3beda34afb8194103a649/src/app/system/navigation/RouterLink.tsx#L23) prop and wrap buttons with `RouterLink`:

```
<RouterLink to="/foo" hasChildTouchable>
  <Button>bar</Button>
</RouterLink>
```

olerichter00:
Closing this PR for now and revisiting possibilities to simplify the code later after the RouterLink migration.

PR #11709, opened 2025-03-18
chore: Add all screen queries to routes.tsx for prefetching

olerichter00:
### Description

This PR adds missing screen queries to routes.tsx. The queries will be used for prefetching.

🤖 Generated with [Claude Code](https://claude.ai/code)

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- Add all screen queries to `routes.tsx` for prefetching - ole

</details>

### DIFF
--- a/src/app/Navigation/routes.tsx
+++ b/src/app/Navigation/routes.tsx
@@ -79,7 +79,10 @@ import { FairMoreInfoQueryRenderer } from "app/Scenes/Fair/FairMoreInfo"
 import { FeaturedFairsScreen, featuredFairsScreenQuery } from "app/Scenes/Fair/FeaturedFairsScreen"
 import { Favorites, Favorites as LegacyFavorites } from "app/Scenes/Favorites/Favorites"
 import { FeatureQueryRenderer } from "app/Scenes/Feature/Feature"
-import { GalleriesForYouScreen } from "app/Scenes/GalleriesForYou/GalleriesForYouScreen"
+import {
+  GalleriesForYouQuery,
+  GalleriesForYouScreen,
+} from "app/Scenes/GalleriesForYou/GalleriesForYouScreen"
 import { GeneQueryRenderer, GeneScreenQuery } from "app/Scenes/Gene/Gene"
 import { HomeViewScreen, homeViewScreenQuery } from "app/Scenes/HomeView/HomeView"
 import {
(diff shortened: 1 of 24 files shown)


ArtsyOpenSource:

- Dev changes (Add all screen queries to `routes.tsx` for prefetching - ole - olerichter00)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against f99199d01f58b8b5cc79596c1d0659420fa70a54
</p>

src/app/Scenes/ArtQuiz/ArtQuizResults/ArtQuizResults.tsx
```diff
@@ -33,7 +33,7 @@ export const ArtQuizResults = ({ isCalculatingResult }: { isCalculatingResult?:
   )
 }
 
-const artQuizResultsQuery = graphql`
+export const ArtQuizResultsScreenQuery = graphql`
```
olerichter00, line 36: I chose this naming pattern because it's the one that is currently used for most queries. However, I'm open to changing it.

src/app/Navigation/routes.tsx
```diff
@@ -442,6 +492,7 @@ export const artsyDotNetRoutes = defineRoutes([
     path: "/artist/:artistID/shows",
     name: "ArtistShows",
     Component: ArtistShowsQueryRenderer,
+    queries: [ArtistShowsScreenQuery],
```
MounirDhahri, line 495: In some cases like this query, we need `queryVariables` as well so this would work properly. So we might need to export it as well.

damassi (approved):

src/app/Navigation/routes.tsx
olerichter00: You are right. In this case, the variable `:artistID` will be taken from the path, but for other routes, we will need to either add the variables here or set the defaults within the GraphQL query.

I'm not sure if I can automate this work. So far, I have manually tested whether the prefetching worked and added the variables in case it hasn't... 🤔

damassi: We should go the the route approach as that's our centralized hub which contains all screen dependencies. Over in force this is how we do it:
- `[prepareVariables](https://github.com/artsy/force/blob/main/src/Apps/Artist/artistRoutes.tsx#L146-L162)`: 
- And then in the RouterLink's `usePrefetchRoute` hook, we [look it up](https://github.com/artsy/force/blob/main/src/System/Hooks/usePrefetchRoute.tsx#L42): 

So as things grow, and if we ever need to add anything more, the routes file / route item is the only place you need to go, and then just call `findRoutesByPath` and pass it the route path in question. Has worked well over there.

MounirDhahri: I received some more warnings related to this that I didn't expect.
Screenshot 2025-03-28 at 18 10 56

PR #11730, opened 2025-03-19
feat: add custom context fields to Unleash

dblandin:
### Description

This is meant as a conversation starter – not something on the critical path for feature development.

Reflecting on recent Unleash-backed A/B experiments, it would be nice to update the targeting for an experiment so that a certain group is always given the `experiment` variant. For example the members of a specific product team or all Artsy staff members (as identified via the `team` role).

We've recently built some handling into Metaphysics (https://github.com/artsy/metaphysics/pull/6297 and https://github.com/artsy/metaphysics/pull/6465) to support returning Home View content only when the client is requesting from a supported app version. It seems like we can do similar targeting for Unleash feature flags (including experiments), if we pass along an appVersion context field.

- appPlatformOS: "ios" | "android" | "web" | # to allow targeting based on platform
- appVersion: "8.67.0" # to allow targeting based on semVer constraints
- userRoles: "user,team" # to allow targeting based on user role (for example, rolling out to all Artsy staff)

The updated API request params to the Unleash backend look a bit like this:

```
sessionId: 128917516
appName: eigen
environment: default
properties["appPlatformOS"]: ios
properties["userRoles"]: user,team
userId: <user ID>
```

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- feat: add custom context fields to Unleash

</details>

### DIFF
--- a/src/app/system/flags/Components/WrappedFlagProvider.tsx
+++ b/src/app/system/flags/Components/WrappedFlagProvider.tsx
@@ -3,6 +3,8 @@ import FlagProvider, { IConfig } from "@unleash/proxy-client-react"
 import { useUnleashEnvironment } from "app/system/flags/hooks/useUnleashEnvironment"
 import { useUnleashInitializer } from "app/system/flags/hooks/useUnleashInitializer"
 import { useUnleashListener } from "app/system/flags/hooks/useUnleashListener"
+import { getAppVersion } from "app/utils/appVersion"
+import { Platform } from "react-native"
 import Keys from "react-native-keys"
 
 export const WrappedFlagProvider: React.FC = ({ children }) => {
@@ -11,6 +13,12 @@ export const WrappedFlagProvider: React.FC = ({ children }) => {
   const storageName = (name: string) => `unleash-values:${name}`
 
   const config: IConfig = {
+    context: {
+      properties: {
+        appVersion: getAppVersion(),
+        appPlatformOS: Platform.OS,
+      },
+    },
     url:
       env === "production"
         ? Keys.secureFor("UNLEASH_PROXY_URL_PRODUCTION")
--- a/src/app/system/flags/hooks/__tests__/useUnleashInitializer.tests.ts
+++ b/src/app/system/flags/hooks/__tests__/useUnleashInitializer.tests.ts
@@ -1,0 +1,150 @@
+import { renderHook } from "@testing-library/react-hooks"
+import { useUnleashClient } from "@unleash/proxy-client-react"
+import { GlobalStore } from "app/store/GlobalStore"
+import { useUnleashInitializer } from "app/system/flags/hooks/useUnleashInitializer"
+import { jwtDecode } from "jwt-decode"
+
+jest.mock("@unleash/proxy-client-react")
+jest.mock("app/store/GlobalStore")
+jest.mock("jwt-decode")
+
+jest.mock("react-native", () => ({
+  Platform: { OS: "ios" },
+  NativeModules: {
+    ArtsyNativeModule: {
+      gitCommitShortHash: "abc123",
+      isBetaOrDev: false,
+      clearCache: jest.fn(),
+    },
+  },
+}))
+
+const mockUseUnleashClient = useUnleashClient as jest.MockedFunction<typeof useUnleashClient>
+const mockJwtDecode = jwtDecode as jest.MockedFunction<typeof jwtDecode>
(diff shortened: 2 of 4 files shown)


src/app/system/flags/Components/WrappedFlagProvider.tsx
```diff
@@ -11,6 +13,12 @@ export const WrappedFlagProvider: React.FC = ({ children }) => {
   const storageName = (name: string) => `unleash-values:${name}`
 
   const config: IConfig = {
+    context: {
+      properties: {
+        appVersion: DeviceInfo.getVersion(),
```
dblandin, line 18: Should we be using the version attribute within `app.json` instead? Looks we use that when generating our user agent string. When running in a simulator, `DeviceInfo.getVersion()` doesn't properly return a known version number but we still use this utility function in a few places, including the About screen.

MounirDhahri (approved):
I am definitely in for more granularity with Unleash. In many situations it would have been great to target a specific target group instead of injecting user ids. Do you know how we can set such groups/strategies? does unleash allow for that

dariakoko (approved):

src/app/system/flags/Components/WrappedFlagProvider.tsx
dariakoko: `DeviceInfo.getVersion()` retrieves the version of the app installed from the store. That's probably why we don't see a correct version when running in a simulator.

Do we use version number from `app.json` anywhere?

anandaroop (commented):
I like it! 

Is the idea that the custom `userRoles` context variable being set here would show up in the Unleash targeting UI?

Screenshot 2025-03-19 at 8 53 51 PM

src/app/system/flags/Components/WrappedFlagProvider.tsx
dblandin: We're currently using the app.json value to construct the user agent string. Maybe not very significant before but we're now using that value in Metaphysics to alter the homeView content returned.

dblandin:
> I like it! 
> 
> Is the idea that the custom `userRoles` context variable being set here would show up in the Unleash targeting UI?
> 
> Screenshot 2025-03-19 at 8 53 51 PM
> 

Yes, exactly. You have to define these properties in the Unleash admin UI for them to show up as targeting options. And while Unleash doesn't directly support arrays of strings, we can use a comma-separated value in an "includes" clause for role targeting.

Also interesting to find that they have specific semVer constraints that you can apply to a semVer version value.

dblandin:
> I am definitely in for more granularity with Unleash. In many situations it would have been great to target a specific target group instead of injecting user ids. Do you know how we can set such groups/strategies? does unleash allow for that

I know you can define segments in the Unleash UI which can persist a group of constraints and give it a name. So for Onyx, we can define a named Onyx segment with those userIDs preconfigured.

But with user roles we can also define an "Artsy Staff" segment.

gkartalis:
I am in favour of this approach - will be really useful

ArtsyOpenSource:

- Dev changes (feat: add custom context fields to Unleash - dblandin)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 9ffa57a5b3322deeb0b2ecbffc5202cb8e9f07c6
</p>

src/app/system/flags/Components/WrappedFlagProvider.tsx
dblandin: FYI – I opened [another PR](https://github.com/artsy/eigen/pull/11750) to bring more consistency to how we determine the current app version, preferring the app.json value based on [this Slack conversation](https://artsy.slack.com/archives/C02BAQ5K7/p1742476420628389). 

Holding on this change until that PR is resolved – then I'll rebase here.

MounirDhahri:
Any blockers on getting this merged - it adds a lot of value

dblandin:
> Any blockers on getting this merged - it adds a lot of value

Thanks for the nudge! I'll revisit this change and the other [version number consolidation PR](https://github.com/artsy/eigen/pull/11750).

src/app/system/flags/hooks/__tests__/useUnleashInitializer.tests.ts
MounirDhahri, line 1: thanks for the thorough testing here

PR #11744, opened 2025-03-21
chore: Migrate Relay `{ cacheConfig: force true }` to `store-and-network` fetchPolicy

olerichter00:
This PR resolves [ONYX-1630] 

### Description

This PR replaces Relay's `cacheConfig: force: true` with `fetchPolicy: "store-and-network"` for screens using `useLazyloadQuery`. It also adds a few more queries to routes.tsx for prefetching.

Relay's "store-and-network" fetch policy does not seem to be supported by `QueryRenderer`. During testing, the component did not run a new query when the screen opened, and the query was already cached. I will further investigate and probably migrate the remaining screens that use `QueryRenderer` to `useLazyLoadQuery`.

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- Migrate Relay `{ cacheConfig: force true }` to "store-and-network" fetchPolicy (for prefetching) - ole

</details>

[ONYX-1630]: https://artsyproduct.atlassian.net/browse/ONYX-1630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Components/Containers/Inbox.tsx
+++ b/src/app/Components/Containers/Inbox.tsx
@@ -1,12 +1,12 @@
 import { ActionType } from "@artsy/cohesion"
 import {
-  Spacer,
   Flex,
+  Screen,
   Separator,
-  Tabs,
   Skeleton,
   SkeletonText,
-  Screen,
+  Spacer,
+  Tabs,
 } from "@artsy/palette-mobile"
 import { TabsContainer } from "@artsy/palette-mobile/dist/elements/Tabs/TabsContainer"
 import { StackScreenProps } from "@react-navigation/stack"
@@ -16,14 +16,12 @@ import { Inbox_me$data } from "__generated__/Inbox_me.graphql"
 import { ConversationsContainer } from "app/Scenes/Inbox/Components/Conversations/Conversations"
 import { MyBidsContainer } from "app/Scenes/MyBids/MyBids"
 import { listenToNativeEvents } from "app/store/NativeModel"
-import { getRelayEnvironment } from "app/system/relay/defaultEnvironment"
 import { PlaceholderBox, PlaceholderText } from "app/utils/placeholders"
-import { renderWithPlaceholder } from "app/utils/renderWithPlaceholder"
 import { track } from "app/utils/track"
 import { ActionNames, ActionTypes } from "app/utils/track/schema"
-import React, { memo } from "react"
+import React, { memo, Suspense } from "react"
 import { EmitterSubscription } from "react-native"
-import { createRefetchContainer, graphql, QueryRenderer, RelayRefetchProp } from "react-relay"
+import { createRefetchContainer, graphql, RelayRefetchProp, useLazyLoadQuery } from "react-relay"
 
 enum Tab {
   bids = "bids",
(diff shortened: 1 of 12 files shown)


src/app/Components/Containers/Inbox.tsx
```diff
@@ -147,34 +145,37 @@ export const InboxScreenQuery = graphql`
...
+    </Suspense>
+  )
+}
+
 interface InboxQueryRendererProps extends StackScreenProps<any> {
   isVisible?: boolean
 }
 
 export const InboxQueryRenderer: React.FC<InboxQueryRendererProps> = memo((props) => {
+  const data = useLazyLoadQuery<InboxQuery>(
```
olerichter00, line 161: Migrating the query renderer to `useLazyLoadQuery` to be able to prefetch and reload the data when the user opens the screen.

src/app/Scenes/ArtworkList/ArtworkList.tsx
```diff
@@ -39,19 +39,20 @@ const SORT_OPTIONS: SortOption[] = [
 
 const DEFAULT_SORT_OPTION = SORT_OPTIONS[0].value as CollectionArtworkSorts
 
+export const artworkListVariables = { count: PAGE_SIZE, sort: DEFAULT_SORT_OPTION }
```
olerichter00, line 42: Extracting default variables for prefetching.

MounirDhahri (approved):

src/app/Components/Containers/Inbox.tsx
```diff
@@ -147,34 +145,37 @@ export const InboxScreenQuery = graphql`
   }
 `
 
+export const InboxScreen: React.FC<InboxQueryRendererProps> = (props) => {
+  return (
+    <Suspense fallback={<InboxPlaceholder />}>
+      <InboxQueryRenderer {...props} />
+    </Suspense>
```
brainbicycle, line 152: @olerichter00 sorry for the late drive by review but I _think_ there is a subtle but important difference here vs renderWithPlaceholder. I believe renderWithPlaceholder will render the failure view in event of errors on the same screen we are on whereas without it the error will bubble up and potentially cause app wide unable to load making tabs inaccessible. tl;dr we need an ErrorBoundary in addition to the Suspense: https://github.com/artsy/eigen/blob/e65fbf76e4e339b80c54af89c3bdc32f25362ec9/src/app/utils/hooks/withSuspense.tsx#L55

olerichter00: Thanks for the late review :) 

I tested the error handling of the screen (by throwing an error inside of the component), and it looks fine to me. The "Unable to Load" screen is displayed; it's possible to reload and access other tabs. Or am I missing something?

I guess, the global error [AppWideErrorBoundary](https://github.com/artsy/eigen/blob/59bb4ffb8ba1e7244737a597c96b0c8eb6d13448/src/app/Providers.tsx#L41) handles the error with displaying the tabs correctly 🤔 


https://github.com/user-attachments/assets/de501ad1-632f-4104-9ceb-9e119eccaeb6

brainbicycle: it is good it is showing the tabs, but really we shouldn't hit the app wide error boundary whenever possible which is what the error is saying and instead add an error boundary that is as local as possible to where the error occurred

brainbicycle: https://github.com/artsy/eigen/pull/11035 e.g. trying to maintain as much nav as possible possibly a simpler error message

olerichter00: Yeah, that makes sense 👍 I've created a PR that adds local error boundaries with `withSuspense` to the Inbox and the Collection Artist screens: https://github.com/artsy/eigen/pull/11775

PR #11745, opened 2025-03-21
fix: Increase Artwork grid page size and threshold

olerichter00:
This PR resolves [ONYX-1629] 

* [Slack Thread](https://artsy.slack.com/archives/C02BAQ5K7/p1742550961154439)

### Description

This increases the artwork grid's page size and "end reached threshold" to improve scroll performance.

**Before and after:**

https://github.com/user-attachments/assets/b19628d3-b160-4a4a-8f03-f854ee4a2bbd

### PR Checklist

### To the reviewers 👀

- Increase Artwork grid page size and "end reached" threshold (performance improvement) - ole

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[ONYX-1629]: https://artsyproduct.atlassian.net/browse/ONYX-1629?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/utils/masonryHelpers/index.tsx
+++ b/src/app/utils/masonryHelpers/index.tsx
@@ -6,9 +6,9 @@ export const ESTIMATED_MASONRY_ITEM_SIZE = 272
 
 export const NUM_COLUMNS_MASONRY = isTablet() ? 3 : 2
 
-export const ON_END_REACHED_THRESHOLD_MASONRY = 0.3
+export const ON_END_REACHED_THRESHOLD_MASONRY = 0.9
 
-export const MASONRY_LIST_PAGE_SIZE = 10
+export const MASONRY_LIST_PAGE_SIZE = 15
 
 export interface masonryRenderItemProps {
   item: MasonryArtworkItem
--- a/src/app/utils/masonryHelpers/index.tsx
+++ b/src/app/utils/masonryHelpers/index.tsx
@@ -6,7 +6,7 @@ export const ESTIMATED_MASONRY_ITEM_SIZE = 272
 
 export const NUM_COLUMNS_MASONRY = isTablet() ? 3 : 2
 
-export const ON_END_REACHED_THRESHOLD_MASONRY = 0.9
+export const ON_END_REACHED_THRESHOLD_MASONRY = 2
 
 export const MASONRY_LIST_PAGE_SIZE = 15
 


ArtsyOpenSource:

- Cross-platform user-facing changes (Increase Artwork grid page size and "end reached" threshold (performance improvement) - ole - olerichter00)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against ec762682f658484bfde6456fa1a1a78f32503d78
</p>

gkartalis (dismissed):
good idea ;)

src/app/utils/masonryHelpers/index.tsx
```diff
@@ -6,9 +6,9 @@ export const ESTIMATED_MASONRY_ITEM_SIZE = 272
 
 export const NUM_COLUMNS_MASONRY = isTablet() ? 3 : 2
 
-export const ON_END_REACHED_THRESHOLD_MASONRY = 0.3
+export const ON_END_REACHED_THRESHOLD_MASONRY = 2
```
anandaroop, line 9: I've been trying to debug a QA issue where [pagination is flaky on artwork lists' grids](https://www.notion.so/artsy/My-saved-artworks-list-says-26-artworks-but-in-my-list-are-only-10-1cfcab0764a08032917df9f505a9ce0d).

Sometimes pagination works and sometimes it doesn't. Even when you revisit the _same_ artwork list, the results are unpredictable.

One of the things I tried was setting this threshold to 0.5 (the example value from the docs):
val

That _seems_ to fix the issue. I'll keep poking to see if I can prove that's really it. 

(Not totally obvious to me _why_ this would fix it, or be so inconsistent in the first place. Also I'd expect to see this on other artwork grids too? Need to rule that out too.)

anandaroop: The change to `2` seemed fine on other grids that I checked (artist, fair, marketing collection) 

But it reliably correlates with problems on artwork lists — both the new (`/favorites`) and old (`/favorites/saves`) versions. Which are then fixed by setting this to 0.5

I'm inclined to change this value — thoughts? Or other theories?

olerichter00: The issue seems to be related to the "store-and-network" fetch policy. I can consistently reproduce the error (even with setting the threshold to 0.5) when  
a) I wait long enough before tapping on "Saved Artworks" until the artwork list query is prefetched, or  
b) I come back to the artwork list screen (`/favorites`) after I have opened it before.

I assume the issue arises because the screen immediately renders the artwork list and calls `loadNext` to load the next page, and at the same time, it refetches the data in the background due to the fetch policy. When the refetch query wins, it results in only the first 10 artworks being displayed and `loadNext` never being called again.

olerichter00: Maybe decreasing the threshold reduces the likelihood that `loadNext` is called too soon in some cases which is why it looks like it could fix the bug. 🤔

PR #11750, opened 2025-03-21
chore: consolidate querying of current app version

dblandin:
### Description

Prefer value stored in `app.json` file. Fallback to getVersion from react-native-device-info package.

Allow optional inclusion of buildMetadata as supported [in the SemVer spec](https://semver.org/#spec-item-10).

This will include the build number or the commit sha.

### PR Checklist

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- chore: consolidate querying of current app version

</details>


ArtsyOpenSource:

- Dev changes (chore: consolidate querying of current app version - dblandin)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against d3edd2abe01db5cdaabd17858404f6285c5e34ca
</p>

src/app/store/GlobalStore.tsx
```diff
@@ -114,7 +114,7 @@ export function getCurrentEmissionState() {
   // See: https://github.com/artsy/metaphysics/pull/6297
   const userAgent = `${
     __DEV__ ? "Artsy-Mobile " + Platform.OS : DeviceInfo.getUserAgentSync()
-  } ${DeviceInfo.getSystemName()}/${DeviceInfo.getSystemVersion()} Artsy-Mobile/${version} Eigen/${DeviceInfo.getBuildNumber()}/${version}`
+  } ${DeviceInfo.getSystemName()}/${DeviceInfo.getSystemVersion()} Artsy-Mobile/${AppInfo.version()} Eigen/${DeviceInfo.getBuildNumber()}/${AppInfo.version()}`
```
dblandin, line 117: **walkthrough**: I didn't change this userAgent string but I am wondering why we have two similar looking chunks:

```
Artsy-Mobile/${AppInfo.version()
Eigen/some-build-number/${AppInfo.version()
```

I'm not sure what's relying on this specific format but I wonder if we could eventually consolidate to the same SemVer-compatible format:

**Before**
```
Artsy-Mobile/8.68.0
Eigen/1674646088/8.68.0
```

**After**
```
Artsy-Mobile/8.68.0+1674646088
Eigen/8.68.0+1674646088
```

src/app/system/appInfo.ts
```diff
@@ -0,0 +1,23 @@
...
+import { appJson } from "app/utils/jsonFiles"
+import DeviceInfo, { getVersion } from "react-native-device-info"
+
+const version = (options: { includeBuildMetadata: boolean } = { includeBuildMetadata: false }) => {
+  const version = appJson().version || getVersion()
+
+  let buildMetadata = DeviceInfo.getBuildNumber()
+
+  if (!buildMetadata) {
+    buildMetadata = `sha.${ArtsyNativeModule.gitCommitShortHash}`
```
brainbicycle, line 11: this api is a bit funky to me because at a glance I expect AppInfo.version to just return the app version and I think the build number and hash are pretty distinct so defaulting to hash also seems a but unexpected from that reading
What about an api with multiple fields and one for the fullVersionString like: 

AppInfo: {
  version()
  longVersionString() // this returns the extra metadata
}

brainbicycle (commented):
this looks good to me! one small nitpick about the api

src/app/store/GlobalStore.tsx
brainbicycle: you are def right this is weird, I don't have a hard answer for what relies on it but i know some analytics parse that string, we can try to track those relying on it down but for now I would leave as is

src/app/system/appInfo.ts
dblandin: I went ahead and added a spec file to better illustrate expected behavior. Most calls would simply use `AppInfo.getVersion()`. The dev menu is currently the only place which uses the longer `AppInfo.getVersion({ includeBuildMetadata: true })` invocation.

Based on the [SemVer spec](https://semver.org/#spec-item-10), I'm considering the build metadata part of the version but an optional addition, depending on the needs of the caller. It might be nice to get this longer version string in the userAgent, analytics, and/or Sentry but I've left that out of scope for now.

brainbicycle: as in as a mobile dev I expect a method called getVersion to return 1 thing that is the version: 8.7.0 without anything appended. In my mind if we want to make a longer string for display somewhere that includes extra metadata it should be just a whole distinct method or if we just need it in one place just combine them in place like "AppInfo.getVersion() + AppInfo.getBuildMetadata" so I can see immediately what is happening at the call site. This is nitpicking and non-blocking but just my thoughts.

brainbicycle: and maybe part of confusion is that on iOS the version is referring to CFBundleVersion: https://developer.apple.com/documentation/BundleResources/Information-Property-List/CFBundleVersion
and can only have numbers separated by periods, doesn't really adhere to semver because it is just identifying which version of the binary it is

brainbicycle: coming back to this, this feedback is non-blocking for what it is worth so if you want to merge feel free!

dblandin:
Closing in favor of https://github.com/artsy/eigen/pull/12645 which proposes a simpler implementation.

PR #11780, opened 2025-03-26
[Do Not Merge] feat(prefetching): Add debugger

damassi:
### Description

Not necessarily meant to be merged, but to help with QA / debugging some of the edge-cases where prefetching isn't as sensitive as it should be (like on the horizontal rails). 

https://github.com/user-attachments/assets/78d57a55-1ab7-4707-83b9-b117556e0c7c


MounirDhahri (dismissed):
This is pretty useful for testing. Maybe we can have this behind a dev flag and ship it.

olerichter00:
This is amazing! 🤩 I took the liberty of pushing a commit that updates the border color depending on whether the prefetching query is completed or still in progress.

damassi:
Oh yah, no problem -- commit away 👍 this is just for debugging weird edge cases, a take it or leave it kinda thing.

damassi:
> Maybe we can have this behind a dev flag and ship it.

@MounirDhahri - worth noting that if this is merged, we'd want to guard against the state update triggering the rerender when the flag is off. I've noticed that it can trigger unexpected UI flashes on the images, etc. And the state update has perf implications too. (but maybe the latter is not so bad, compared to the flashing UI -- something that can be fixed)

src/app/system/navigation/RouterLink.tsx
```diff
@@ -49,8 +52,11 @@ export const RouterLink: React.FC<RouterLinkProps & TouchableProps> = ({
   }
 
   const handleVisible = (isVisible: boolean) => {
+    setPrefetchState("started")
```
MounirDhahri, line 55: this is going to be always called which might trigger a re-render as @damassi raised in his last comment. I would avoid this if the flag isn't enabled

MounirDhahri (commented):
Almost there, only a small comment based on Chris's feedback

src/app/system/navigation/RouterLink.tsx
olerichter00: You are right, we have to move `setPrefetchState("started")` inside the condition `if (isPrefetchingEnabled && isVisible)` 👍 🟢

MounirDhahri (dismissed):
LGTM! thanks @olerichter00 

**Edit:**
I still though think that if we want to merge this, we would need a separate flag for developers only - because at its current state this PR would still affect the performance.

I mean something like
```
const showPrefetchingHelpers = useDevFlag("whatever")

if (showPrefetchingHelpers) {
  return <Border>
   {...}
 </Border>
}

```

olerichter00:
> I still though think that if we want to merge this, we would need a separate flag for developers only - because at its current state this PR would still affect the performance.

For me, it would be okay to not merge this PR and just check out the branch when debugging is needed (and close the PR when the project is finished).

damassi:
I _do_ think this would be helpful to have, but not necessarily if it complicates things. We don't have anything like this in Force since we've got RouterLink everywhere, just thinking that finishing up that last 1% this will be helpful, so we don't have to switch between branches. 

Maybe an in between here would be to merge this and then delete the code once QA is complete. Then we (and other users) can toggle it on during independent QA passes and get a feel for what could be improved.

olerichter00:
Alright, I [added a dev toggle](https://github.com/artsy/eigen/pull/11780/commits/27d1361170449d2f58f67b42ce9c7d10e29d4f43) to switch the prefetching indicator on and off, and I added a condition to the state update to avoid rerendering when the indicator is disabled. Like this we should be able to merge it without any performance issues 🤔

damassi:
Sweet, thanks @olerichter00 👍

damassi:
Code updates look great, gonna #MergeOnGreen

olerichter00:
> Code updates look great, gonna #MergeOnGreen


#[Squash On Green](https://github.com/artsy/eigen/labels/Squash%20On%20Green)

PR #11782, opened 2025-03-27
chore(ONYX-1644): Fetch home view experiments async in the background

olerichter00:
Resolves [ONYX-1644](https://artsyproduct.atlassian.net/browse/ONYX-1644)
### Description

> [!NOTE]
> This might not bring much performance improvement because Metaphysics does not seem to make a request to Unleash to fetch the experiments ([code](https://github.com/artsy/metaphysics/blob/20b513b73e7ad0f58b59053c936c840042267a5f/src/schema/v2/homeView/experiments/HomeViewExperiments.ts#L19)). But it's probably good to move everything out of the query to make it smaller.

This PR defers fetching home view experiments (asynchronously and in the background) to potentially increase the time the home screen renders.

I suppose it's fine to fetch the experiments within an effect after the home screen has rendered because they are only used for tracking and don't affect any rendered components.

### PR Checklist

### To the reviewers 👀

- 

#### iOS user-facing changes

-

-

#### Dev changes

- Fching home view experiments async - ole

</details>

[ONYX-1644]: https://artsyproduct.atlassian.net/browse/ONYX-1644?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/HomeView/hooks/useHomeViewExperimentTracking.ts
+++ b/src/app/Scenes/HomeView/hooks/useHomeViewExperimentTracking.ts
@@ -1,20 +1,23 @@
-import { ExperimentViewed, ActionType, OwnerType } from "@artsy/cohesion"
-import { HomeViewQuery$data } from "__generated__/HomeViewQuery.graphql"
+import { ActionType, ExperimentViewed, OwnerType } from "@artsy/cohesion"
+import {
+  useHomeViewExperimentTrackingQuery,
+  useHomeViewExperimentTrackingQuery$data,
+} from "__generated__/useHomeViewExperimentTrackingQuery.graphql"
 import { HomeViewStore } from "app/Scenes/HomeView/HomeViewContext"
+import { getRelayEnvironment } from "app/system/relay/defaultEnvironment"
 import { compact } from "lodash"
 import { useEffect } from "react"
+import { fetchQuery, graphql } from "react-relay"
 import { useTracking } from "react-tracking"
 
-export function useHomeViewExperimentTracking(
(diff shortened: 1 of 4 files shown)


src/app/Scenes/HomeView/hooks/useHomeViewExperimentTracking.ts
```diff
@@ -44,5 +48,34 @@ export function useHomeViewExperimentTracking(
...
+      },
+      next: (data) => {
+        trackExperiments(data.homeView?.experiments)
+      },
+    })
+  }
+
+  useEffect(() => {
+    fetchAndTrackExperiments()
+  }, [])
```
damassi, line 80: Niceeee - these little things certainly add up 👍

anandaroop (approved):
I wonder if we'll be able to track the effect on `HomeViewQuery`. I'd expect the impact of this to be minimal, since MP's call to Unleash is [querying an in-memory store](https://github.com/Unleash/unleash-client-node#:~:text=The%20SDK%20will%20set%20up%20an%20in%2Dmemory%20repository%20and%20poll%20for%20updates%20from%20the%20Unleash%20server%20at%20regular%20intervals), and is likely extremely fast.

So I _think_ the tradeoff here is probably a very marginal user-facing speedup in exchange for a little more overhead and MP traffic. 

But in the interest of faster perceived loading times, no objection here, let's give this a shot 👍🏽

PR #11810, opened 2025-03-31
feat: Decrease visibility theshold to prefetch links earlier

olerichter00:
## Description

This PR adds a `threshold` prop to `Sentinel` to be able to prefetch links as soon as they enter the viewport (vs. when they have fully entered the viewport).

### Behaviour

| Before | After |
| --- | --- |
|https://github.com/user-attachments/assets/af91a3e0-5eb0-43fc-a137-528299732137 | https://github.com/user-attachments/assets/7f0540df-d414-40d0-9f7f-c19718e5b5bd|

### Performance

| Before | After |
| --- | --- |
|https://github.com/user-attachments/assets/82728aee-1fc2-40ed-b293-4b3989b17170 | https://github.com/user-attachments/assets/4564e8a1-f2be-4455-9b8c-b9531f989cab|

### PR Checklist

### To the reviewers 👀

- Prefetch links earlier (when they enter the viewport instead of when they are fully visible)- ole

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/system/navigation/RouterLink.tsx
+++ b/src/app/system/navigation/RouterLink.tsx
@@ -77,7 +77,7 @@ export const RouterLink: React.FC<RouterLinkProps & TouchableProps> = ({
   // If the child component is a touchable element, we don't add another touchable wrapper
   if (hasChildTouchable && isPrefetchingEnabled) {
     return (
-      <Sentinel onChange={handleVisible}>
+      <Sentinel onChange={handleVisible} threshold={0}>
         <Border prefetchState={prefetchState}>
           {React.Children.map(children, (child) => {
             return React.isValidElement(child) ? React.cloneElement(child, cloneProps) : child
@@ -102,7 +102,7 @@ export const RouterLink: React.FC<RouterLinkProps & TouchableProps> = ({
   }
 
   return (
-    <Sentinel onChange={handleVisible}>
+    <Sentinel onChange={handleVisible} threshold={0}>
       <Border prefetchState={prefetchState}>
         <Touchable {...touchableProps}>{children}</Touchable>
       </Border>
--- a/src/app/utils/Sentinel.tsx
+++ b/src/app/utils/Sentinel.tsx
@@ -13,10 +13,15 @@ import { useFocusEffect } from "@react-navigation/native"
 import { FC, ReactNode, useCallback, useRef, useState } from "react"
 import { Dimensions, View } from "react-native"
 
+const DEFAULT_THRESHOLD = 1
+
 export interface IDimensionData {
   rectTop: number
   rectBottom: number
   rectWidth: number
+  rectHeight: number
+  width: number
+  height: number
 }
 
 export interface Props {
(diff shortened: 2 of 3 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 054462151ad4a586091c88af9da16fa225513f98
</p>

src/app/system/navigation/RouterLink.tsx
```diff
@@ -77,7 +77,7 @@ export const RouterLink: React.FC<RouterLinkProps & TouchableProps> = ({
   // If the child component is a touchable element, we don't add another touchable wrapper
   if (hasChildTouchable && isPrefetchingEnabled) {
     return (
-      <Sentinel onChange={handleVisible}>
+      <Sentinel onChange={handleVisible} threshold={0}>
```
olerichter00, line 80: `threshold={0}` means that links will be prefetched as soon as they enter the viewport.

src/app/utils/Sentinel.tsx
```diff
@@ -13,28 +13,38 @@ import { useFocusEffect } from "@react-navigation/native"
 import { FC, ReactNode, useCallback, useRef, useState } from "react"
 import { Dimensions, View } from "react-native"
 
+const DEFAULT_THRESHOLD = 1
```
olerichter00, line 16: Setting the default to how the behaviour was before this update.

src/app/system/navigation/RouterLink.tsx
damassi: And if its a little fussy, maybe even .5, just so its triggered by the edges that are in the horizontal scroll but slightly peaking through. But lets see how this it feels in a beta 👍

MounirDhahri (commented):
I have the same thought as @damassi here - do you mind if you create a beta and add its number to the PR to test things out. I do however think this would probably not affect performance much since it's already happening.
Another thing we can do is to trigger the prefetching on momentum end but that's probably going to require more computing and affect the performance further.

src/app/system/navigation/RouterLink.tsx
olerichter00: I don't expect any performance degradation with this update. We're pretty much fetching the same amount of queries, just a bit earlier and a few more on the initial screen render (links that are completely on the screen + links with edges on the screen).

I've added before and after videos with a visible performance monitor (top left corner) and created new betas to test the changes on a real device 🚀

brainbicycle (approved):

olerichter00:
> Another thing we can do is to trigger the prefetching on momentum end but that's probably going to require more computing and affect the performance further.

@MounirDhahri, Triggering prefetching on [onmomentumscrollend](https://reactnative.dev/docs/scrollview#onmomentumscrollend) sounds interesting! It would probably require some work because, at the moment, prefetching isn't triggered based on scrolling but rather by using an interval that checks the link's position every x milliseconds.

PR #11822, opened 2025-04-02
fix(PBRW-558): Fix error on tapping Make Offer

MrSltun:
This PR resolves [PBRW-558] 

### Description
This PR fixes an issue with artworks with edition sets, where the user can create an order for the artwork before an edition set is selected, which breaks in Exchange

The fix that I implemented here was to hide the commercial buttons if the artwork has _more than one_ edition sets and an edition set is not selected

| Platform | Before | After |
|---|---|---|
| Android | <video src="https://github.com/user-attachments/assets/cc9cbeba-eb40-4e7c-ad56-9563619bc877" width="400" /> | <video src="https://github.com/user-attachments/assets/52ee5075-282a-4e8c-9a89-a70b70bc5613" width="400" /> |
| iOS | <video src="https://github.com/user-attachments/assets/7ab059ef-36b1-4cf1-b319-2c5e1c71a921" width="400" /> | <video src="https://github.com/user-attachments/assets/2a0fd2ae-5e09-4302-b8d2-55b429eee19d" width="400" /> |

| | Android | iOS |
|---|---|---|
| With 1 edition set | image | image |

### PR Checklist

  - [x] **Android**.
  - [x] **iOS**.

### To the reviewers 👀

- Prevent creating an order for artworks with edition sets when the edition set is not selected

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[PBRW-558]: https://artsyproduct.atlassian.net/browse/PBRW-558?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/Artwork/Components/ArtworkCommercialButtons.tsx
+++ b/src/app/Scenes/Artwork/Components/ArtworkCommercialButtons.tsx
@@ -88,7 +88,7 @@ export const ArtworkCommercialButtons: React.FC<ArtworkCommercialButtonsProps> =
     )
   }
 
-  if (!noEditions && !selectedEditionId) {
+  if ((!artworkData.editionSets || artworkData.editionSets.length > 1) && !selectedEditionId) {
     return null
   }
 
--- a/src/app/Scenes/Artwork/Components/ArtworkCommercialButtons.tsx
+++ b/src/app/Scenes/Artwork/Components/ArtworkCommercialButtons.tsx
@@ -88,6 +88,10 @@ export const ArtworkCommercialButtons: React.FC<ArtworkCommercialButtonsProps> =
     )
   }
 
+  if (!noEditions && !selectedEditionId) {
+    return null
+  }
+
   if (artworkData.isOfferable && artworkData.isAcquireable) {
     return (
       <RowContainer>
--- a/src/app/Scenes/Artwork/Components/ArtworkCommercialButtons.tests.tsx
+++ b/src/app/Scenes/Artwork/Components/ArtworkCommercialButtons.tests.tsx
@@ -668,45 +668,55 @@ describe("ArtworkCommercialButtons", () => {
     })
   })
 
-  describe("with edition sets", () => {
-    const editionSet = {
-      internalID: "editionSetID",
-      editionOf: "Edition Set One",
-      saleMessage: "$1000",
-      isAcquireable: true,
-      isOfferable: true,
-      isInquireable: false,
-    }
-    const artworkWithEditionSets = {
-      ...ArtworkFixture,
-      editionSets: [editionSet],
-      isAcquireable: true,
-      isOfferable: true,
-      isInquireable: false,
-    }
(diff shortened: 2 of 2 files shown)


gkartalis (dismissed):
beautiful!

ArtsyOpenSource:

- Cross-platform user-facing changes (Prevent creating an order for artworks with edition sets when the edition set is not selected - MrSltun)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against be4607fae827f49612d0bc3dddf2271e7654ccfa
</p>

leamotta (approved):

src/app/Scenes/Artwork/Components/ArtworkCommercialButtons.tsx
```diff
@@ -88,6 +88,10 @@ export const ArtworkCommercialButtons: React.FC<ArtworkCommercialButtonsProps> =
     )
   }
 
+  if (!noEditions && !selectedEditionId) {
```
leamotta, line 91: actually we would need to check for >1 editionsets length since artworks with 1 type of edition don't have a selected edition set

MounirDhahri:
do you know why there is not edition set and how does it get selected? I don't see you tapped anywhere in the video attached

MrSltun:
> do you know why there is not edition set and how does it get selected? I don't see you tapped anywhere in the video attached

@MounirDhahri it's a network delay, if I tap quick enough (or I have a slow internet connection) before the edition sets data loads, I get the error

Here's a video showing the issue:
<video src="https://github.com/user-attachments/assets/0bbc4c75-2e0f-41a2-8ce3-e227de6ba782" width="400px" />

MrSltun:
> thank you! also a bit confused about how the selection happens? does this just happen automatically when you go to the page?
@brainbicycle yes, once the edition sets data loads, we update the artwork store with the first edition set

brainbicycle:
> > thank you! also a bit confused about how the selection happens? does this just happen automatically when you go to the page?
> 
> @brainbicycle yes, once the edition sets data loads, we update the artwork store with the first edition set

Ahh I see: https://github.com/artsy/eigen/blob/d4c8741edbfbc875b75587acbd670acf3c2b1901/src/app/Scenes/Artwork/Components/ArtworkEditionSets.tsx#L16

Outside scope of this PR but I wonder if the default edition set could just be returned by MP in the artwork payload if we are using this logic elsewhere

MounirDhahri (dismissed):
thanks for explaining @MrSltun 
I already agree with @brainbicycle - it would be great to return the first edition set or maybe even the first edition set price so we don't need to have logic to fire another query on both app and web

leamotta (approved):

MrSltun:
I agree that we should update it to come from MP, but for now to not block the release, I'll merge this PR!

PR #11826, opened 2025-04-02
refactor(ArtworkLists): converts the ArtworkLists into easy-peasy store

araujobarret:
This PR resolves [DIA-1227] 

### Description

This PR improves the performance of Artwork Lists Provider by converting it into an easy-peasy context.

No visual changes are expected, but I'd gladly ask someone with more expertise in the Artwork Lists to test it thoroughly 🧪 

Now Infinite Discovery doesn't suffer from any delay when saving an artwork into a list.

Profiling the usage of saving an Artwork into a list before the changes(we see 2 commit bars and the time was quite high for the first big commit 190ms, 2nd big commit was also expensive):
Screenshot 2025-03-21 at 11 39 52

Profiling the usage of saving an Artwork into a list before the changes(we only see 1 commit bar and the time improved to around 105ms):
Screenshot 2025-03-21 at 11 38 34

Now we see in real-time the interactions with the list from more expensive screens:

https://github.com/user-attachments/assets/84ce0813-f4cb-4e1f-9d15-50e13ffa3d66

Major changes of the PR:
- ArtworkLists converted into easy-peasy store
- Removed unnecessary context API within useToast
- onSave removed from the list that was creating a dependency to another provider and was doing too many tasks, abstracted this logic to use the hook `useArtworkListsModifiedToastHandler` and  added a computed field `modifiedActionType` that identifies what action was performed

### PR Checklist

  - [x] **Android**.
  - [x] **iOS**.

### To the reviewers 👀

- refactor: converts ArtworkLists into easy-peasy store and remove the dependency of toastsProvider from it

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[DIA-1227]: https://artsyproduct.atlassian.net/browse/DIA-1227?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Components/ArtworkLists/ArtworkListsStore.tsx
+++ b/src/app/Components/ArtworkLists/ArtworkListsStore.tsx
@@ -2,15 +2,13 @@ import { BottomSheetModalProvider } from "@gorhom/bottom-sheet"
 import {
   getArtworkListsStoreInitialState,
   ArtworkListEntity,
-  ArtworkEntity,
-  RecentlyAddedArtworkList,
-  ArtworkListState,
 } from "app/Components/ArtworkLists/types"
 import { ArtworkListOfferSettingsView } from "app/Components/ArtworkLists/views/ArtworkListOfferSettingsView/ArtworkListOfferSettingsView"
 import { CreateNewArtworkListView } from "app/Components/ArtworkLists/views/CreateNewArtworkListView/CreateNewArtworkListView"
 import { SelectArtworkListsForArtworkView } from "app/Components/ArtworkLists/views/SelectArtworkListsForArtworkView/SelectArtworkListsForArtworkView"
 import { Action, action, Computed, computed, createContextStore } from "easy-peasy"
 import { FC } from "react"
+import { ArtworkListState } from "./types"
 
 export type ModifiedListType =
   | "GENERIC_CHANGES"
@@ -22,19 +20,7 @@ export type ModifiedListType =
 
 export interface ArtworkListsModel {
   // State
-  selectArtworkListsViewVisible: boolean
-  createNewArtworkListViewVisible: boolean
-  artworkListOfferSettingsViewVisible: boolean
-  artwork: ArtworkEntity | null
-  artworkListID: string | null
-  recentlyAddedArtworkList: RecentlyAddedArtworkList | null
-  selectedTotalCount: number
-  addingArtworkLists: ArtworkListEntity[]
-  removingArtworkLists: ArtworkListEntity[]
-  sharingArtworkLists: ArtworkListEntity[]
-  keepingArtworkListsPrivate: ArtworkListEntity[]
-  hasUnsavedChanges: boolean
-  toastBottomPadding: number | null
+  state: ArtworkListState
   artworkListId?: string
 
   // Computed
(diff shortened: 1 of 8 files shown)


src/app/Components/ArtworkLists/useArtworkListsModifiedToastHandler.ts
```diff
@@ -0,0 +1,58 @@
+import { dispatchArtworkSavedStateChanged } from "app/Components/ArtworkLists/ArtworkListEvents"
+import { ArtworkListsStore } from "app/Components/ArtworkLists/ArtworkListsContext"
+import { useArtworkListToast } from "app/Components/ArtworkLists/useArtworkListsToast"
+import { useCallback } from "react"
+
+export const useArtworkListsModifiedToastHandler = () => {
```
damassi, line 6: Maybe just `useArtworkListToast`?

damassi (dismissed):
Nice work on this! 

I'm 99.9% sure this would remove the benefit, but kinda curious what the response time would look like if all hooks and state were combined into a single hook, similar to the existing `useArtworkListContext`, something like:

```tsx
const useArtworkListContext = () => {
  const actions = ArtworksListContext.useActions(actions => actions)
  const state = ArtworksListContext.useActions(actions => state)
  
  return { ...actions, ...state }
}
```

Like, just how intelligent is the story memoization? 

It makes sense that the perf benefits only come from plucking things out individually, though I've been meaning to investigate this.

src/app/Components/ArtworkLists/useArtworkListsModifiedToastHandler.ts
```diff
@@ -0,0 +1,58 @@
...
+        break
+      case "ADDED_AND_REMOVED_LIST":
+        dispatchArtworkSavedStateChanged(artwork.internalID)
+        toast.changesSaved()
+        break
+      case "GENERIC_CHANGES":
+        toast.changesSaved()
+        break
+    }
+  }, [artwork, addingArtworkLists, removingArtworkLists, modifiedActionType, toast])
```
MounirDhahri, line 55: I would avoid this since we are using all objects here - it would make it easier to add new action types without breaking its logic

MounirDhahri (dismissed):
Great addition! I would suggest test this extensively in other app surfaces and android as well

src/app/Components/ArtworkLists/useArtworkListsModifiedToastHandler.ts
MounirDhahri: I meant the useCallback

src/app/Components/ArtworkLists/components/ArtworkLists.tsx
```diff
@@ -1,6 +1,6 @@
 import { BottomSheetFlatList } from "@gorhom/bottom-sheet"
 import { ArtworkLists_me$data, ArtworkLists_me$key } from "__generated__/ArtworkLists_me.graphql"
-import { useArtworkListsContext } from "app/Components/ArtworkLists/ArtworkListsContext"
+import { ArtworkListsStore } from "app/Components/ArtworkLists/ArtworkListsContext"
```
brainbicycle, line 3: maybe this file should be called ArtworkListsStore or similar now that we are not using context anymore?

src/app/Components/ArtworkLists/views/SelectArtworkListsForArtworkView/components/SelectArtworkListsForArtwork.tsx
```diff
@@ -69,7 +65,7 @@ const ArtworkListsPlaceholder = () => {
   )
 }
 
-const Query = graphql`
+export const selectArtworkListsForArtworkQuery = graphql`
```
brainbicycle, line 68: 🧹

src/app/Components/ArtworkLists/views/SelectArtworkListsForArtworkView/useSaveArtworkListsChanges.ts
```diff
@@ -13,26 +13,36 @@ interface Options {
...
-      artwork_ids: [artwork.internalID],
+      artwork_ids: [artwork?.internalID ?? ""],
       owner_ids: addingArtworkListIDs,
     }
 
     trackEvent(event)
   }
 
   const save = useCallback(() => {
+    if (!artwork) {
```
brainbicycle, line 42: 🦺

brainbicycle (dismissed):
great refactor 🎉 ! minor comment on old file name. Agree with Mounir this probably merits some targeted testing

src/app/Components/ArtworkLists/ArtworkListsStore.tsx
```diff
@@ -0,0 +1,228 @@
...
+import { FC } from "react"
+import { ArtworkListState } from "./types"
+
+export type ModifiedListType =
+  | "GENERIC_CHANGES"
+  | "ADDED_AND_REMOVED_LIST"
+  | "ADDED_SINGLE_LIST"
+  | "ADDED_MULTIPLE_LIST"
+  | "REMOVED_SINGLE_LIST"
+  | "REMOVED_MULTIPLE_LIST"
```
JanaeHijaz, line 19: @araujobarret I'm not sure I understand this, are these referring to specific user behaviors?

PR #11832, opened 2025-04-02
fix(DIA-1173): some users cannot send password reset emails

iskounen:
We have received reports of users that are unable to send themselves password reset emails from the new auth flow. I suspect that the request to send a password reset email is failing for these users because the `AuthModal` is holding on to (and reusing) expired XAPP tokens.

### PR Checklist

  - [x] **Android**.
  - [x] **iOS**.

### To the reviewers 👀

- Fixed a bug where some users couldn't send password reset emails

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/store/AuthModel.ts
+++ b/src/app/store/AuthModel.ts
@@ -1040,6 +1040,12 @@ export const getAuthModel = (): AuthModel => ({
 
 const isTokenExpired = (expiresIn: string) => {
   const expirationDate = new Date(expiresIn)
+
+  if (isNaN(expirationDate.getTime())) {
+    // if expiresIn is not a valid date string, treat it as expired
+    return true
+  }
+
   const now = new Date()
   return expirationDate <= now
 }
--- a/src/app/store/AuthModel.ts
+++ b/src/app/store/AuthModel.ts
@@ -219,8 +219,8 @@ export const getAuthModel = (): AuthModel => ({
   }),
   getXAppToken: thunk(async (actions, _payload, context) => {
     const xAppToken = context.getState().xAppToken
-    if (xAppToken) {
-      // TODO: handle expiry
+    const xAppTokenExpiresIn = context.getState().xApptokenExpiresIn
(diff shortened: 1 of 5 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 9ace9911612acd4bb239ae24a58965ee38e3c2f1
</p>

src/app/store/AuthModel.ts
```diff
@@ -1038,6 +1038,12 @@ export const getAuthModel = (): AuthModel => ({
   }),
 })
 
+const isTokenExpired = (expiresIn: string) => {
+  const expirationDate = new Date(expiresIn)
+  const now = new Date()
+  return expirationDate <= now
```
brainbicycle, line 1044: I think it is probably a good idea to include a buffer like Joey was suggesting here to avoid any race conditions or problems with clock drift: 
```
const isTokenExpired = (expiresIn: string, bufferMs = 60_000) => {
  const expirationDate = new Date(expiresIn).getTime()
  return Date.now() >= (expirationDate - bufferMs)
}
``` 
something like that maybe.
https://artsy.slack.com/archives/C02BAQ5K7/p1743636834716749?thread_ts=1743634138.506409&cid=C02BAQ5K7

brainbicycle (changes requested):
minor comment around adding a buffer, but otherwise looks good! thanks

src/app/store/AuthModel.ts
iskounen: Thanks for the code sample. I originally didn't understand the request, but this helped me learn about clock drift and other factors that can cause unexpected behavior.

The only change I made was to bump the buffer up to 5 mins.

brainbicycle (approved):
nice one! thanks!

PR #11878, opened 2025-04-09
chore: remove brew installed detect secrets

ovasdi:
This PR resolves N/A

### Description

Don't install detect-secrets using brew. Devs should install the tool using pip via [potential](https://github.com/artsy/potential/pull/487).

### PR Checklist

  - [ ] **Android**.
  - [ ] **iOS**.

### To the reviewers 👀

#### Dev changes

- remove brew install of detect-secrets

</details>

### DIFF
--- a/scripts/utils/install_detect_secrets.sh
+++ b/scripts/utils/install_detect_secrets.sh
@@ -13,4 +13,11 @@ done
 echo "legacy_version_file = yes" >> "$HOME/.asdfrc"
 asdf plugin add python
 asdf install python latest
-asdf set python latest --home
+
+if asdf help set &>/dev/null; then
+  # Older asdf version
+  asdf set python latest --home
+else
+  # Newer asdf version
+  asdf global python latest
+fi
--- a/scripts/utils/install_detect_secrets.sh
+++ b/scripts/utils/install_detect_secrets.sh
@@ -11,7 +11,11 @@ for package in "${pip_packages[@]}"; do
 done
 
 echo "legacy_version_file = yes" >> "$HOME/.asdfrc"
-asdf plugin add python
+if asdf plugin add --help &>/dev/null; then
+  asdf plugin add python
+else
+  asdf plugin-add python
+fi
 asdf install python latest
 
 if asdf set --help &>/dev/null; then
--- a/scripts/utils/install_detect_secrets.sh
+++ b/scripts/utils/install_detect_secrets.sh
@@ -14,10 +14,10 @@ echo "legacy_version_file = yes" >> "$HOME/.asdfrc"
 asdf plugin add python
 asdf install python latest
 
-if asdf help set &>/dev/null; then
-  # Older asdf version
+if asdf set --help &>/dev/null; then
+  # Newer asdf version
   asdf set python latest --home
 else
-  # Newer asdf version
+  # Older asdf version
   asdf global python latest
 fi
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,3 +1,4 @@
 nodejs 20.9.0
 ruby 3.1.6
 java zulu-17.50.19
+python 3.13.3t
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,4 +1,3 @@
 nodejs 20.9.0
 ruby 3.1.6
 java zulu-17.50.19
-python 3.13.3t
--- a/Brewfile
+++ b/Brewfile
@@ -1,2 +1,1 @@
 brew "jq"
-brew "detect-secrets"
--- a/install_detect_secrets.sh
+++ b/install_detect_secrets.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+[ ! -f "$HOME/.default-python-packages" ] && touch "$HOME/.default-python-packages"
+
+# Array of pip packages to be added
+pip_packages=("virtualenv" "pre-commit" "detect-secrets==1.5.0")
+for package in "${pip_packages[@]}"; do
+  if ! grep -q "$package" "$HOME/.default-python-packages"; then
+    echo "$package" >> "$HOME/.default-python-packages"
+  fi
+done
+
+echo "legacy_version_file = yes" >> "$HOME/.asdfrc"
+asdf plugin add python
+asdf install python latest
+asdf set python latest --home


ArtsyOpenSource:

- Dev changes (remove brew install of detect-secrets - ovasdi)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 01fe7ebc6af34d816f02802457c4dd8bd7eb2e0f
</p>

brainbicycle:
i think we need python in our tool-versions for the python detect-secrets to work, one minute 🙏

scripts/utils/install_detect_secrets.sh
```diff
@@ -0,0 +1,23 @@
...
+echo "legacy_version_file = yes" >> "$HOME/.asdfrc"
+asdf plugin add python
+asdf install python latest
+
+if asdf help set &>/dev/null; then
+  # Older asdf version
+  asdf set python latest --home
+else
+  # Newer asdf version
+  asdf global python latest
```
brainbicycle, line 22: added this convenience script based on potential for devs to install, I think potential probably needs an if check like this, later asdf versions don't have the `set` command

brainbicycle (dismissed):
this looks good to me, but once this is merged lets let devs know in practice mobile that they should run:
```
brew uninstall detect-sectrets
./scripts/utils/install_detect_secrets.sh
```

ovasdi:
> i think we need python in our tool-versions for the python detect-secrets to work, one minute 🙏

It was working for me without any other changes.

```
➜  eigen git:(main) ✗ git commit -m "foo"
yarn run v1.22.22
$ /Users/ovasdi/code/eigen/node_modules/.bin/lint-staged
✔ Preparing lint-staged...
❯ Running tasks for staged files...
  ❯ package.json — 1 file
    ❯ * — 1 file
      ✖ yarn detect-secrets:hook [FAILED]
    ↓ *.@(ts|tsx) — no files [SKIPPED]
    ↓ *.@(json|md) — no files [SKIPPED]
↓ Skipped because of errors from tasks. [SKIPPED]
✔ Reverting to original state because of errors...
✔ Cleaning up temporary files...

✖ yarn detect-secrets:hook:
error Command failed with exit code 1.
$ scripts/detect-secrets/detect-secrets.sh hook /Users/ovasdi/code/eigen/foo.txt
Executing detect-secrets-hook...
ERROR: Potential secrets about to be committed to git repo!

Secret Type: Secret Keyword
Location:    foo.txt:1
```

brainbicycle:
> > i think we need python in our tool-versions for the python detect-secrets to work, one minute 🙏
> 
> It was working for me without any other changes.
> 
> ```
> ➜  eigen git:(main) ✗ git commit -m "foo"
> yarn run v1.22.22
> $ /Users/ovasdi/code/eigen/node_modules/.bin/lint-staged
> ✔ Preparing lint-staged...
> ❯ Running tasks for staged files...
>   ❯ package.json — 1 file
>     ❯ * — 1 file
>       ✖ yarn detect-secrets:hook [FAILED]
>     ↓ *.@(ts|tsx) — no files [SKIPPED]
>     ↓ *.@(json|md) — no files [SKIPPED]
> ↓ Skipped because of errors from tasks. [SKIPPED]
> ✔ Reverting to original state because of errors...
> ✔ Cleaning up temporary files...
> 
> ✖ yarn detect-secrets:hook:
> error Command failed with exit code 1.
> $ scripts/detect-secrets/detect-secrets.sh hook /Users/ovasdi/code/eigen/foo.txt
> Executing detect-secrets-hook...
> ERROR: Potential secrets about to be committed to git repo!
> 
> Secret Type: Secret Keyword
> Location:    foo.txt:1
> ```

we don't need it in tool-versions but potential script is broken for me because of my detect-secrets version so global python was not set, see the convenience script I added

scripts/utils/install_detect_secrets.sh
ovasdi, line 22: I believe its the other way around, `set` is used in the newer version, starting `0.16.0`, while `global` was used in version prior to `0.16.0` release. https://asdf-vm.com/guide/upgrading-to-v0-16.html#asdf-global-and-asdf-local-commands-have-been-replaced-with-asdf-set
```suggestion
if asdf set --help &>/dev/null; then
  # Newer asdf version
  asdf set python latest --home
else
  # Older asdf version
  asdf global python latest
```

brainbicycle: ahh you are right, thanks

brainbicycle:
> > added this convenience script based on potential for devs to install
> 
> Curious to know why add the install script here instead of expect devs to install python and configure default packages via potential? Is it because this is an open source project?

No, just trying to keep the disruption as low as possible. Without this script what I would do is go to potential, find where detect secrets is installed, isolate that and run it myself rather than rerunning the whole script. I suspect a lot of other devs would do the same so this is just doing it for them to make sure not to block devs in the short term.

ovasdi:
> > > added this convenience script based on potential for devs to install
> > 
> > 
> > Curious to know why add the install script here instead of expect devs to install python and configure default packages via potential? Is it because this is an open source project?
> 
> No, just trying to keep the disruption as low as possible. Without this script what I would do is go to potential, find where detect secrets is installed, isolate that and run it myself rather than rerunning the whole script. I suspect a lot of other devs would do the same so this is just doing it for them to make sure not to block devs in the short term.

So if potential had an easier/better way to specify which specific step to install, would that make it a better experience and keep disruption low? Or would you still prefer to keep this script here?

brainbicycle:
> > > > added this convenience script based on potential for devs to install
> > > 
> > > 
> > > Curious to know why add the install script here instead of expect devs to install python and configure default packages via potential? Is it because this is an open source project?
> > 
> > 
> > No, just trying to keep the disruption as low as possible. Without this script what I would do is go to potential, find where detect secrets is installed, isolate that and run it myself rather than rerunning the whole script. I suspect a lot of other devs would do the same so this is just doing it for them to make sure not to block devs in the short term.
> 
> So if potential had an easier/better way to specify which specific step to install, would that make it a better experience and keep disruption low? Or would you still prefer to keep this script here?

if I could specify easily which step to install then yeah, although the script here is just a short term thing to help with migration, I am not sure it is worth trying to isolate every step in potential but maybe. I just don't run it that often unless I am setting up a new machine, maybe that is wrong 🤷

scripts/utils/install_detect_secrets.sh
```diff
@@ -0,0 +1,23 @@
...
+# Array of pip packages to be added
+pip_packages=("virtualenv" "pre-commit" "detect-secrets==1.5.0")
+for package in "${pip_packages[@]}"; do
+  if ! grep -q "$package" "$HOME/.default-python-packages"; then
+    echo "$package" >> "$HOME/.default-python-packages"
+  fi
+done
+
+echo "legacy_version_file = yes" >> "$HOME/.asdfrc"
+asdf plugin add python
```
ovasdi, line 14: Related to https://github.com/artsy/eigen/pull/11878#discussion_r2037391113

I just remembered that this command is not backwards compatible https://asdf-vm.com/guide/upgrading-to-v0-16.html#hyphenated-commands-have-been-removed

That means if someone with `asdf` version < `0.16.0` is running this script, it will fail to add the plugin. If this script is aiming to be backwards compatible with `asdf` versions prior to `0.16.0` then:
```suggestion
if asdf plugin add --help &>/dev/null; then
  asdf plugin add python
else
  asdf plugin-add python
fi
```

brainbicycle:
```
~/C/eigen (brian+george/expo-again|✔) [1] $ asdf plugin add python                                                                          (base) 
updating plugin repository...HEAD is now at c0369a1 feat: add yasm plugin (#1087)
Plugin named python already added
~/C/eigen (brian+george/expo-again|✔) $ asdf --version                                                                                      (base) 
v0.15.0
``` 
hmm this actually works on my version, maybe earlier versions it didn't, but i don't see any harm in checking

PR #11932, opened 2025-04-17
feat(DIA-1241): Improve performance of Infinite Discovery 

araujobarret:
This PR resolves [DIA-1241] 

### Description

Performance improvements:
- Do not keep more than 2 swiped cards rendered
- Remove workaround to keep artwork's relay data stored in a ref
- Avoid Relay's GCing swiped cards data with retain
- Invalidate cache when leaving the screen

| Platform | Before | After |
|---|---|---|
| Android | <video src="https://github.com/user-attachments/assets/d8a0570a-46e2-4694-9e0e-0508e8838c60" width="400" /> | <video src="https://github.com/user-attachments/assets/14f7591a-ab00-4898-bb3b-398232d006bc" width="400" /> |

Before:
After the 10th artwork, it gets super laggy, we swipe until the 20th card 

After:
No lag during the test, we swipe until the 27th card without lag and performance is boosted comparing both

### PR Checklist

- [X] I have tested my changes on the following platforms:
  - [X] **Android**.
  - [ ] **iOS**.
- [X] I have included **screenshots** or **videos** at least on **Android**, or I have not changed the UI.

### To the reviewers 👀

- feat: virtualise unswiped cards and more fixes for infinite discovery

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[DIA-1241]: https://artsyproduct.atlassian.net/browse/DIA-1241?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/InfiniteDiscovery/InfiniteDiscovery.tsx
+++ b/src/app/Scenes/InfiniteDiscovery/InfiniteDiscovery.tsx
@@ -339,8 +339,9 @@ export const InfiniteDiscoveryQueryRenderer = withSuspense({
       infiniteDiscoveryQuery,
       infiniteDiscoveryVariables
     )
-    const disposableQueries = useRef<Disposable[]>([])
-    const queryIds = useRef<string[][]>([])
+    // Disposable queries to allow them to be GCed by Relay when calling dispose()
+    const queriesForDisposal = useRef<Disposable[]>([])
+    const usedExcludeArtworkIds = useRef<string[][]>([])
     const env = getRelayEnvironment()
     const prefetch = usePrefetch()
 
@@ -348,12 +349,13 @@ export const InfiniteDiscoveryQueryRenderer = withSuspense({
     const initialArtworks = extractNodes(data.discoverArtworks)
     const [artworks, setArtworks] = useState<InfiniteDiscoveryArtwork[]>(initialArtworks)
 
+    // Retain the queries to not have them GCed when swiping many times
     const retainQuery = useCallback((excludeArtworkIds: string[]) => {
       const queryRequest = getRequest(infiniteDiscoveryQuery)
       const descriptor = createOperationDescriptor(queryRequest, { excludeArtworkIds })
       const disposable = env.retain(descriptor)
-      disposableQueries.current.push(disposable)
-      queryIds.current.push(excludeArtworkIds)
+      queriesForDisposal.current.push(disposable)
+      usedExcludeArtworkIds.current.push(excludeArtworkIds)
     }, [])
 
     const fetchMoreArtworks = async (excludeArtworkIds: string[], isRetry = false) => {
@@ -390,12 +392,12 @@ export const InfiniteDiscoveryQueryRenderer = withSuspense({
 
       // Mark the queries to be disposed by GC, invalidate cache and prefetch a infinite discovery again
       return () => {
-        disposableQueries.current.forEach((disposable) => {
-          if (!!disposable.dispose) {
-            disposable.dispose()
+        queriesForDisposal.current.forEach((query) => {
+          if (!!query.dispose) {
+            query.dispose()
           }
         })
-        queryIds.current.forEach((id) => {
+        usedExcludeArtworkIds.current.forEach((id) => {
           commitLocalUpdate(env, (store) => {
             store
               ?.getRoot()


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 9492df934076f86585311d540f0b67ce0c71eb31
</p>

src/app/Scenes/InfiniteDiscovery/InfiniteDiscovery.tsx
```diff
@@ -371,8 +385,27 @@ export const InfiniteDiscoveryQueryRenderer = withSuspense({
...
+        })
+        queryIds.current.forEach((id) => {
+          commitLocalUpdate(env, (store) => {
+            store
+              ?.getRoot()
+              ?.getLinkedRecord("discoverArtworks", { excludeArtworkIds: id })
+              ?.invalidateRecord()
+          })
+        })
+        prefetch("/infinite-discovery", infiniteDiscoveryVariables)
```
iskounen, line 406: **question:** Why are you prefetching again when this component dismounts?

src/app/Scenes/InfiniteDiscovery/InfiniteDiscovery.tsx
```diff
@@ -338,15 +339,27 @@ export const InfiniteDiscoveryQueryRenderer = withSuspense({
       infiniteDiscoveryQuery,
       infiniteDiscoveryVariables
     )
+    const disposableQueries = useRef<Disposable[]>([])
+    const queryIds = useRef<string[][]>([])
```
iskounen, line 343: **suggestion:** I understand what you're doing with `disposableQueries` because I've read the Relay docs, but the `queryIds` is a mystery to me. These two could use some docs to explain what's going on.

iskounen (approved):

MounirDhahri (commented):
Thanks for improving the performance here, it's a bit scary though to do so much store manipulation manually. If we ever need to update this code, it might be tricky given its complexity (the quality is there for sure 👏 ⭐ )

I still feel like there is something wrong with our overall architecture here leading to the performance issues and we are fixing the symptoms. After all, we are showing just a few artworks per session and I find it hard to believe that relay, powering Facebook, can't handle 20 or 30 artwork records. I saw @damassi raising something similar about manually updating the store and its cache size and I must admit I also think we should be careful with it.

src/app/Scenes/InfiniteDiscovery/InfiniteDiscovery.tsx
```diff
@@ -338,15 +339,29 @@ export const InfiniteDiscoveryQueryRenderer = withSuspense({
...
     const [artworks, setArtworks] = useState<InfiniteDiscoveryArtwork[]>(initialArtworks)
 
+    // Retain the queries to not have them GCed when swiping many times
+    const retainQuery = useCallback((excludeArtworkIds: string[]) => {
+      const queryRequest = getRequest(infiniteDiscoveryQuery)
+      const descriptor = createOperationDescriptor(queryRequest, { excludeArtworkIds })
+      const disposable = env.retain(descriptor)
+      queriesForDisposal.current.push(disposable)
+      usedExcludeArtworkIds.current.push(excludeArtworkIds)
+    }, [])
```
olerichter00, line 359: I don't fully understand why we have to retain the queries. Is it because, even if we store all artworks in a state, the fragments could get garbage collected since we only render the most recent 2 fragments?

olerichter00: Thanks for the explanation! 🙏  I assume the garbage collection happens because the component (artwork card) that is referencing the fragment isn't rendered anymore, right?

Although this workaround works fine, I wonder if the performance bug could also be fixed by rendering all cards (to avoid the fragments being garbage collected) but letting the cards that are not in the foreground return null instead of rendering the content 🤔 Just in case we want to simplify and avoid manipulating the garbage collection.

MounirDhahri:
> @MounirDhahri is a paginated concept in this case, but we don't have pagination backed by MP, so we lose all the extras we have in all other paginated surfaces like the grids where we never run into this issue. We're kinda doing manually what the paginated queries from Relay do but manually, that's not something we should do, but it's more right to go down one level and avoid more hacks in our code.

Thanks for explaining this, I think this is worth a comment in the code to make sure it's clear in the future. I wasn't able to understand this until you explained it now.

> Feel free to tag the backend folks to support pagination there(I don't think it's possible given the nature of open search)

My goal wasn't to engineer a solution here because tbh I don't know why it's broken, I wanted to tag Chris because we had a recent conversation with Ole [here](https://artsy.slack.com/archives/C05EQL4R5N0/p1743089915537149?thread_ts=1743011726.845739&cid=C05EQL4R5N0) about manipulating cache that might add value here.

src/app/Scenes/InfiniteDiscovery/InfiniteDiscovery.tsx
damassi: > but letting the cards that are not in the foreground return null instead of rendering the content

Great potential fix, and would agree. 

For a UI / data layer as simple as this, it (ideally) wouldn't require so much complexity. Y'all know this space the best, but would be worth giving Ole's idea a shot because the why of all of this garbage collection and so on is likely going to be lost to time as the next dev comes along and looks at the UI, wondering why things needed to get so granular.  

I guess tho this does depend on the user / limits of the system. Can a user swipe 200 cards? Or is the average 10? But in either case, one could also imagine a collection of 100 artwork rows being fetched in a table with rows; we wouldn't think to manage the GC there because 100 rows really isn't that much. (For something like Folio's offline mode, though, this was required). So it seems to point at an underlying problem elsewhere.

PR #11933, opened 2025-04-17
fix: Set UIUserInterfaceStyle to Automatic to adopt system color theme

olerichter00:
This PR resolves [ONYX-1682] 

### Description

The Dark Mode "Sync with system" option isn't working correctly on iOS. This updates `UIUserInterfaceStyle` to `Automatic` to adopt the system color theme instead of always using the light theme.

* https://developer.apple.com/documentation/bundleresources/information-property-list/uiuserinterfacestyle
* https://stackoverflow.com/questions/61124229/react-native-0-62-2-appearance-return-wrong-color-scheme
Simulator Screenshot - iPhone 16 - 2025-04-17 at 16 49 50

### PR Checklist

  - [ ] **Android**.
  - [ ] **iOS**.

### To the reviewers 👀

-

#### iOS user-facing changes

-  Set UIUserInterfaceStyle to Automatic to adopt system color theme (automatic dark mode) - ole

-

#### Dev changes

-

</details>

[ONYX-1682]: https://artsyproduct.atlassian.net/browse/ONYX-1682?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ios/Artsy/App_Resources/Info.plist
+++ b/ios/Artsy/App_Resources/Info.plist
@@ -3,7 +3,7 @@
 <plist version="1.0">
 <dict>
 	<key>UIUserInterfaceStyle</key>
-	<string>Light</string>
+	<string>Automatic</string>
 	<key>AppIdentifierPrefix</key>
 	<string>$(AppIdentifierPrefix)</string>
 	<key>CFBundleDevelopmentRegion</key>


ArtsyOpenSource:

- iOS user-facing changes (Set UIUserInterfaceStyle to Automatic to adopt system color theme (automatic dark mode) - ole - olerichter00)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against a3f319ba8b80cf80abca89bf9c8ed284a7aca17a
</p>

brainbicycle (commented):
the change looks reasonable but I am wondering if we have looked at the native modules like live auction, city guide, view in room in dark mode to see how they behave? We should probably check before enabling because we won't be able to use a feature flag here so they may be broken

MounirDhahri:
@gkartalis Yes the feature is still behind a feature flag and we are still working on QA

@brainbicycle 

**Live Auctions**
tbh, we didn't look into the implementation within live auctions. I don't think it's worth the effort though given the very little usage of the feature and how complicated it might be. 

**City Guide**
The views were migrated to dark mode, the map itself looked fine without being migrated but it would depend on Design feedback after the QA session if we should make it dark as well. It's common in many maps apps to show the map bright

**View in room**
No work has been discussed there, it will depend on design feedback. The good thing, is that there is nothing bright there except on button so it might not require any updates

olerichter00:
Thanks @brainbicycle & @MounirDhahri! 🙏 I've tested it on the simulator, and the native modules seem to look fine in dark mode. I will merge the changes and do some further testing on real devices (and roll back the change before the next release in case it broke something).

MounirDhahri:
Okey, great looks good to me. I just wanted to check to make sure this change won't affect the app if the feature flag is disabled

MounirDhahri (approved):

PR #11951, opened 2025-04-22
chore: Restrict using `navigate` in favor of `RouterLink` with eslint

olerichter00:
This change will restrict the import of `navigate` using the ESLint "no-restricted-imports" rule ([docs](https://archive.eslint.org/docs/rules/no-restricted-imports)) to ensure that `RouterLink` is used instead. 

If it is necessary to use `navigate` directly, you can do so with an `eslint-ignore-next-line` comment above the import.

```typescript
// eslint-disable-next-line no-restricted-imports
import { navigate } from "app/system/navigation/navigate"

.....

navigate("honolulu")
``` 

### PR Checklist

  - [ ] **Android**.
  - [ ] **iOS**.

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- Restrict using `navigate` in favor of `RouterLink` with eslint - ole

</details>

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -141,6 +141,12 @@ module.exports = {
             importNames: ["graphql"],
             message: "Please import `graphql` from `react-relay`.",
           },
+          {
+            name: "app/system/navigation/navigate",
+            importNames: ["navigate"],
+            message:
+              "Please use `RouterLink` instead for prefetching support. If this is not possible, you can import `navigate` with '// eslint-disable-next-line no-restricted-imports'.",
+          },
         ],
       },
     ],
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -145,7 +145,7 @@ module.exports = {
             name: "app/system/navigation/navigate",
             importNames: ["navigate"],
             message:
-              "Please use `RouterLink` instead for prefetching support. If this is not possible, you can import `navigate` with '// eslint-disable-next-line no-restricted-imports'.",
+              "Please use `RouterLink` instead of `navigate` (to support prefetching). If this is not possible, you can import `navigate` by adding a '// eslint-disable-next-line no-restricted-imports' comment.",
           },
         ],
       },


.eslintrc.js
```diff
@@ -140,6 +140,12 @@ module.exports = {
             importNames: ["graphql"],
             message: "Please import `graphql` from `react-relay`.",
           },
+          {
+            name: "app/system/navigation/navigate",
+            importNames: ["navigate"],
+            message:
+              "Please use `RouterLink` instead for prefetching support. If this is not possible, you can import `navigate` with '// eslint-disable-next-line no-restricted-imports'.",
```
olerichter00, line 147:
```suggestion
              "Please use `RouterLink` instead of `navigate` (to support prefetching). If this is not possible, you can import `navigate` by adding a '// eslint-disable-next-line no-restricted-imports' comment.",
```

ArtsyOpenSource:

- Dev changes (Restrict using `navigate` in favor of `RouterLink` with eslint - ole - olerichter00)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 3494e193fcf676aca4f9fbdb9c0c982a3e6d110b
</p>

MounirDhahri (commented):
This is a tricky one for me. On one side, I agree we should default to `RouterLink` for a better UX. But on the other side, I still don't feel strong enough that `navigate` is an antipattern. 

We currently have 292 navigate calls and a lot of them make sense and could not be replaced with a `RouterLink`. Being unable to restrict the usage of `navigate` only within JSX feels like a deal breaker for me.

Example 1:
```typescript
const handlePress = () => {
 if (x === "a") {
  // this should not throw
  navigate("href-1") 
  return
 }
// this should not throw
navigate("href-2")
}
```

Example 2:
```typescript
<Button onPress={() => { 
// this should throw
 navigate("href-3")
}}
```
I would feel weird in example 1 to add `// eslint-disable-next-line app/system/navigation/navigate`

```typescript
const handlePress = () => {
 if (x === "a") {
  // eslint-disable-next-line app/system/navigation/navigate
  navigate("href-1") 
  return
 }
// eslint-disable-next-line app/system/navigation/navigate
navigate("href-2")
}
```

olerichter00:
I agree that there are cases where using `navigate` is necessary, but from what I see, there are very few where `RouterLink` cannot be used ([useEffect example](https://github.com/artsy/eigen/blob/7edbfe024f4200bffdb2700809d655cc7ee906f5/src/app/Scenes/Activity/ActivityItemScreen.tsx#L47)). If I understand your examples correctly, I think they can be converted like this:

Example 1:
```
const href = x === "a" ? "href-1" : "href-2"

<RouterLink to={href} />
```

Example 2:
```
<Button onPress={() => { 
// does not throw an error and navigates to href-3
 navigate("href-3")
}}

// does not throw an error and navigates to href-3
<RouterLink hasChildTouchale to="href-3>
  <Button />
</RouterLink>
```


About how to use `navigate` with the restriction that this PR introduces, it's a bit easier, and the `eslint-ignore` comment needs to be placed when importing `navigate`.

```typescript
// eslint-disable-next-line no-restricted-imports
import { navigate } from "app/system/navigation/navigate"

.....

navigate("honolulu")
```

I have already noticed that a few PRs have introduced new links using `navigate` instead of `RouterLink` since the migration. This approach does not support prefetching and leads to a mix of both patterns.

Restricting the import is my attempt to avoid this issue by adding a warning to ensure that `navigate` is used only when `RouterLink` cannot be applied, which should be in very few instances.



https://github.com/artsy/eigen/blob/7edbfe024f4200bffdb2700809d655cc7ee906f5/src/app/Scenes/Activity/ActivityItemScreen.tsx#L47

brainbicycle (approved):
I am good with this as long as we have a reasonable easy opt out for legitimate use cases and it seems like we do 👍

MounirDhahri (approved):
@olerichter00 thanks for the clear answer. I am sold, let's try it out and see how it goes. Anyway it's something we can change easily if we want to

olerichter00:
> I'm just thinking if this rule is going to create more noise in the code than helping it out, on web we don't restrict the access to any router navigation functions so I wonder if this is really necessary rather than stricter code reviews when checking elements that navigate without the RouterLink 🤔

That's a valid concern, and I'm completely open to removing this rule if it causes too much noise. However, for now, I believe it's a good idea to restrict the use of `navigate`. The situations where `RouterLink` cannot be used are quite rare, and I assume most developers still tend to use the `onPress={() => navigate("honolulu")}` pattern. The rule is there as a reminder to use `RouterLink` and to move away from the current state where both patterns are still mixed.

PR #11952, opened 2025-04-22
chore: use pagination fragment for infinite discovery [POC]

MounirDhahri:
### Description

This PR is a POC on how we can update the infinite discovery to use a pagination fragment. The goal from this is to fix the memory leaks coming from us saving all artworks fetched in memory as you swipe and appending to them the new fetched artworks, which is something that we can do with relay pagination hook, out of the box, with caching optimised out of the box.

This update should in theory fix the memory leak we have as well as solve the issue of the disappearing ids and fragments.

Thanks @araujobarret for the hint here https://github.com/artsy/eigen/pull/11932#issuecomment-2820898724

https://github.com/user-attachments/assets/b0365b9a-2115-45b3-abf8-26653fafce16

This PR isn't mergeable at its current state, but I tested it on android and iOS and everything seems to be working well. Whoever is interested, you can take it over and get it to a mergeable state.

### PR Checklist

  - [x] **Android**.
  - [x] **iOS**.

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Scenes/InfiniteDiscovery/InfiniteDiscovery.tsx
+++ b/src/app/Scenes/InfiniteDiscovery/InfiniteDiscovery.tsx
@@ -10,11 +10,12 @@ import {
   Text,
   Touchable,
 } from "@artsy/palette-mobile"
-import { addBreadcrumb, captureException, captureMessage } from "@sentry/react-native"
+import { captureMessage } from "@sentry/react-native"
+import { InfiniteDiscoveryQuery } from "__generated__/InfiniteDiscoveryQuery.graphql"
 import {
-  InfiniteDiscoveryQuery,
-  InfiniteDiscoveryQuery$data,
-} from "__generated__/InfiniteDiscoveryQuery.graphql"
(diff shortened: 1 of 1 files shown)


MounirDhahri:
Thanks for looking into this. My theory is that it's fine to send those cursors, even though they are not going to be used by the API. My goal was to not make the API paginated, but interpret its results in a paginated way that tricks relay to keep the data it finds.

True, it might not be straighforward to support `excludeArtworkIds`but that's probably not required anymore in such a paginated API.

Thanks again for looking into this, I know it's tricky code in a tricky surface. I am sure that step by step we would make it easier and more performant.

damassi (commented):
I think this approach is worth a good test, even if its kinda faking it! Very curious if this fixes the issue. At the very least, the code now looks vanilla relay, and that's worth something for maintainability and scale through time.

MounirDhahri:
Closing this for lack of activity. If you want to make this migration some day, please let me know. I would be happy to support with it

PR #11967, opened 2025-04-23
chore: add artsy icons check eslint rule

MounirDhahri:
This PR resolves https://github.com/artsy/palette-mobile/pull/334#issuecomment-2824030547 

### Description

This PR comes as a follow-up to this comment https://github.com/artsy/palette-mobile/pull/334#issuecomment-2824201563

The goal from this PR is to make sure that we no longer import icons from `@artsy/palette-mobile`, and instead always import them from `@artsy/icons`.

This required the addition of a new eslint rule that will be only available for devs, and it required the addition of a new root configuration file. I tried to move that around to a better folder but I failed.

Screenshot 2025-04-23 at 16 51 28

### PR Checklist

  - [x] **Android**.
  - [x] **iOS**.

### To the reviewers 👀

- 

#### iOS user-facing changes

-

-

#### Dev changes

- add artsy icons check eslint rule - mounir

</details>

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -11,8 +11,8 @@ module.exports = {
     "react-hooks",
     "testing-library",
     "unused-imports",
-    "no-autofix",
     "artsy",
+    "local-rules",
   ],
   extends: [
     "eslint:recommended",
@@ -38,11 +38,13 @@ module.exports = {
     react: {
       version: "detect",
     },
+    "local-rules": "./eslint-rules",
   },
   rules: {
     /**
      * Errors
      */
+    "local-rules/no-palette-icon-imports": ERR,
     "artsy/no-uselazyloadquery-inside-suspense": ERR,
     "import/order": [
       ERR,
@@ -57,7 +59,6 @@ module.exports = {
     "react/jsx-no-leaked-render": [ERR, { validStrategies: ["coerce", "ternary"] }],
     "react-hooks/rules-of-hooks": ERR,
     "unused-imports/no-unused-imports": OFF, // look below
-    "no-autofix/unused-imports/no-unused-imports": ERR,
 
     /**
      * Rules for tests see https://github.com/testing-library/eslint-plugin-testing-library#supported-rules for details
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,6 @@
 {
   "eslint.validate": ["javascript", "javascriptreact", "typescript", "typescriptreact"],
-  "eslint.run": "onType", // or "onSave"
+  "eslint.run": "onSave",
   "biome.enabled": false,
   "cSpell.diagnosticLevel": "Hint",
   "editor.codeActionsOnSave": {
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,4 @@
 {
-  "eslint.validate": ["javascript", "javascriptreact", "typescript", "typescriptreact"],
-  "eslint.run": "onSave",
   "biome.enabled": false,
   "cSpell.diagnosticLevel": "Hint",
   "editor.codeActionsOnSave": {
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,6 @@
 {
+  "eslint.validate": ["javascript", "javascriptreact", "typescript", "typescriptreact"],
+  "eslint.run": "onType", // or "onSave"
   "biome.enabled": false,
   "cSpell.diagnosticLevel": "Hint",
   "editor.codeActionsOnSave": {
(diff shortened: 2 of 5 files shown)


.vscode/settings.json
```diff
@@ -1,4 +1,6 @@
 {
+  "eslint.validate": ["javascript", "javascriptreact", "typescript", "typescriptreact"],
+  "eslint.run": "onType", // or "onSave"
```
MounirDhahri, line 3: required for better dx

ArtsyOpenSource:

- Dev changes (add artsy icons check eslint rule - mounir - MounirDhahri)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 39d0aacb3550bc2813b23688b182740353552044
</p>

.eslintrc.js
```diff
@@ -38,11 +38,13 @@ module.exports = {
     react: {
       version: "detect",
     },
+    "local-rules": "./eslint-rules",
   },
   rules: {
     /**
      * Errors
      */
+    "local-rules/no-palette-icon-imports": ERR,
```
damassi, line 47: I know time is valuable and there is so much going on so obviously this is non-blocking! 

Thinking back to our process in force tho, we bulk replaced everything in one go and just got it over with. Can we do this here in Eigen? vs adding additional linting logic and pushing that task off into the future.

MounirDhahri: what you are suggesting is however the best course of action to solve all icon issues in Eigen in one go.

.vscode/settings.json
MounirDhahri: I believe it's actually the `defaultValue` here https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint but I wanted it to be explicit about it.

MounirDhahri: Let's change this to `onSave` and see how it goes, maybe it will make Eigen DX faster in VSCode

.eslintrc.js
damassi: Its not totally ideal, but sometimes is ok to just rip the bandaid off and fix issues as they come up and are raised (by devs or other folks), vs continuing to fragment things with V1s existing alongside V2s (and V3s, etc), which Eigen already does too much of. 

The thinking ☝️ here is that from a pure quality perspective, simple and consistent gets fixes, updates and more, even if the simple solution in the short-term leads to hopefully minor issues. Varied and multiple gets "not my problem!" -- and so complexity grows, and quality decreases, and in the end its a free for all. 

Its counter intuitive, but have seen this scenario play out again and again.

MrSltun: Thank you for the effort Mounir! But I think we should not add an eslint rule for that
The plan would be to migrate the used icons from Palette Mobile to Artsy Icons, after that replace all Eigen icons in one go ([jira ticket](https://artsyproduct.atlassian.net/browse/MOPRAT-837) 🔒) to eliminate having multiple sources of truths for the icons

I would be okay with adding the eslint for the time being, but I'm planning to replace all the icons in one go after removing them from Palette Mobile

.vscode/settings.json
damassi: Ah yah! trying it out over here and its the same thing with type-checking, looking closely. If its the default maybe just omit the setting completely?

.eslintrc.js
MounirDhahri: Your point is valid here. I am not up to date with the last state of the `@artsy/icons` project updates on Eigen but I think in this case, if we don't do the entire migration, we will probably never do it.

> complexity grows, and quality decreases

This is true, I spent the last 20 minutes jumping between Claude and replacing icon instances myself in Eigen but I already see some tech debt, some icons do not have the same name and some other do not exist. I received 96 type check errors. But I would consider that as a win because the other 330 imports were successful.

damassi: Its a judgement call for sure. If there are icons that are missing or named differently / now 1:1 replacements (excluding things like padding) that might create a lot of friction so in that case, agreed, we should merge this just to prevent any further spread 👍

MounirDhahri: @damassi Claude is really great though, I am impressed! I am going to try out https://github.com/openai/codex later

damassi: Codex is so unbearably slow in my experience! But i tried it out just after it was released. Do report yr experience back in #dev-ai 🙏

MrSltun: Okay let's merge this one! Thank you Mounir!! 🌟

dariakoko (dismissed):

.eslintrc.js
MounirDhahri: Sure @damassi

MounirDhahri: done here https://artsy.slack.com/archives/CAPQS2MGX/p1745480185394619

PR #12011, opened 2025-04-29
fix: menu item layout on large text

MounirDhahri:
### Description
This PR fixes the MenuItem layout when the font is too big because of accessibility.

Note: I didn't run this through design but I think this is better than broken text

**Before**

___

**After**
Screenshot 2025-04-29 at 15 06 54

### PR Checklist

  - [x] **Android**.
  - [x] **iOS**.

### To the reviewers 👀

- fix menu item layout on large text - mounir

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Components/MenuItem.tsx
+++ b/src/app/Components/MenuItem.tsx
@@ -9,6 +9,7 @@ import {
   TextProps,
 } from "@artsy/palette-mobile"
 import { RouterLink } from "app/system/navigation/RouterLink"
+import { PixelRatio } from "react-native"
 import { ResponsiveValue } from "styled-system"
 
 export const MenuItem: React.FC<{
@@ -42,12 +43,14 @@ export const MenuItem: React.FC<{
   value,
   alignItems = "center",
 }) => {
+  const showInline = PixelRatio.getFontScale() < 1.5 || !value
+
   return (
     <RouterLink onPress={onPress} to={href} underlayColor="mono5" disabled={disabled}>
       <Flex px={px ?? 2}>
         <Flex
-          flexDirection="row"
-          alignItems={alignItems}
+          flexDirection={!showInline ? "column" : "row"}
+          alignItems={showInline ? alignItems : "flex-start"}
           opacity={disabled && allowDisabledVisualClue ? 0.5 : 1}
           py={2}
         >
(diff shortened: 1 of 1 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (fix menu item layout on large text - mounir - MounirDhahri)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 7f68b2fbe6f5491a2b394559863f7b588fc95ed8
</p>

src/app/Components/MenuItem.tsx
```diff
@@ -74,9 +77,16 @@ export const MenuItem: React.FC<{
...
+            flexDirection="row"
+            justifyContent={showInline ? "flex-end" : "space-between"}
+            flex={1}
+            width="100%"
+            flexGrow={3}
+            height="100%"
+          >
             {!!value && (
-              <Flex width={200}>
+              <Flex width={showInline ? 200 : undefined}>
```
olerichter00, line 89: Is setting a fixed `width` really necessary, or could we remove it (or find a workaround)?

olerichter00 (commented):
Great work! Just let's make sure to avoid introducing too much complexity. In almost all cases, it should be possible to make the components responsive and flexible for scalable text using only CSS, with the occasional addition of `PixelRatio.getFontScale()`.

In this example, the usage of `flex-wrap: wrap` could be a pure CSS solution (https://stackoverflow.com/questions/37930217/flex-when-text-too-long-break-into-new-line).

MounirDhahri:
I tried to actually wrap the `MenuItem` content but that didn't play nice because of how the file is made. I wanted to avoid making a big refactoring here because it's a tricky component used in multiple places in the app in different variant and could easily break.

src/app/Components/MenuItem.tsx
MounirDhahri: That's a good point. tbh, I don't know why this was set here, and I felt if we need to refactor this component, we should do that separately. In multiple cases in the past, removing something that was there for a while didn't play nice

olerichter00:
> I tried to actually wrap the MenuItem content but that didn't play nice because of how the file is made. I wanted to avoid making a big refactoring here because it's a tricky component used in multiple places in the app in different variant and could easily break.

I've looked at the component and fully understand your point. 🤯  Let's proceed with your solution to address the issue. We can refactor and simplify the entire component separately at another time.

olerichter00 (approved):

PR #12019, opened 2025-04-29
releases(build): ios dev build

brainbicycle:
This PR resolves [] 

### PR Checklist

  - [ ] **Android**.
  - [ ] **iOS**.

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>


android/app/build.gradle
```diff
@@ -126,13 +127,23 @@ android {
...
             minifyEnabled enableProguardInReleaseBuilds
             proguardFiles getDefaultProguardFile("proguard-android.txt"), "proguard-rules.pro"
+            resValue "string", "EXUpdatesDisableAntiBrickingMeasures", "false"
+        }
+        beta {
+            initWith release
+            signingConfig signingConfigs.release
+            matchingFallbacks = ['release'] // Important: fallback to release if needed
+            // You can set beta-specific settings here
+            resValue "string", "EXUpdatesDisableAntiBrickingMeasures", "true"
```
brainbicycle, line 145: for now this flag is the primary difference between the builds delivered to firebase and those delivered to testflight. 
https://docs.expo.dev/eas-update/override/

ios/Artsy.xcodeproj/project.pbxproj
```diff
@@ -3802,6 +3777,24 @@
...
+			);
+			name = "[Expo Updates] Update Config Plist";
+			outputFileListPaths = (
+			);
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "DEFAULT_EXPO_PLIST=\"${SRCROOT}/Artsy/Supporting/Expo.plist\"\n\nCONFIG_EXPO_PLIST=\"${EXPO_PLIST_PATH:-$DEFAULT_EXPO_PLIST}\"\n\nTARGET_PLIST=\"${BUILT_PRODUCTS_DIR}/${PRODUCT_NAME}.app/Expo.plist\"\n\necho \"Copying $CONFIG_EXPO_PLIST to $TARGET_PLIST\"\ncp \"$CONFIG_EXPO_PLIST\" \"$TARGET_PLIST\"\n";
+		};
```
brainbicycle, line 3797: More readable version:
```
DEFAULT_EXPO_PLIST="${SRCROOT}/Artsy/Supporting/Expo.plist"

CONFIG_EXPO_PLIST="${EXPO_PLIST_PATH:-$DEFAULT_EXPO_PLIST}"

TARGET_PLIST="${BUILT_PRODUCTS_DIR}/${PRODUCT_NAME}.app/Expo.plist"
```
this swaps out the plist on iOS with one with the flag set to enable dynamically changing channels

android/app/build.gradle
```diff
@@ -126,13 +127,23 @@ android {
...
         }
         release {
             // Caution! In production, you need to generate your own keystore file.
             // see https://reactnative.dev/docs/signed-apk-android.
             signingConfig signingConfigs.release
             minifyEnabled enableProguardInReleaseBuilds
             proguardFiles getDefaultProguardFile("proguard-android.txt"), "proguard-rules.pro"
+            resValue "string", "EXUpdatesDisableAntiBrickingMeasures", "false"
+        }
+        beta {
```
brainbicycle, line 140: for now using a builtType rather than a flavor, the benefit is you don't have to pass a buildType and it will default to debug so we don't have to change all our commands. Downside is we can't change the app id this way so may want to change to use flavors later.

ios/Artsy/AppDelegate.mm
```diff
@@ -131,13 +131,7 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
 
 - (RCTBridge *)createBridgeWithDelegate:(id<RCTBridgeDelegate>)delegate launchOptions:(NSDictionary *)launchOptions {
-    // TODO: This is off the beaten track for expo setup, we should probably find another way
-    // TODO: This is currently broken for expo updates
-    // can switch to this [super createBridgeWithDelegate:delegate launchOptions:launchOptions]; to fix expo updates
-    // however this breaks our stateful notificationManager native module because we are changing the bridge delegate
-    // and as a result extraModulesForBridge is not called causing the module to be instantiated 2x
-    // It will definitely not work on new architecture
-    RCTBridge *bridge = [super createBridgeWithDelegate:self launchOptions:launchOptions];
+    RCTBridge *bridge = [super createBridgeWithDelegate:delegate launchOptions:launchOptions];
```
brainbicycle, line 134: deleted comment explains a bit but expo updates needs to be the bridge delegate, however we relied on being the bridge delegate to register our extra modules and if we don't do that they get autoregistered and for modules that got initialized with state this is a problem because the state gets blown away. New approach: let expo updates be bridge delegate and allow auto init of our modules, for any modules that need to retain state make them singletons and make sure that we return the inited instance with state on second init.

ios/Artsy/Emission/Core/AREmission.h
```diff
@@ -16,7 +15,6 @@ extern NSString *const AREnvTest;
 @interface AREmission : NSObject
 
 @property (nonatomic, strong, readonly) RCTBridge *bridge;
-@property (nonatomic, strong, readonly) AREventsModule *eventsModule;
```
brainbicycle, line 19: this module turned out to be dead code, deleted it and related files

ios/Artsy/Emission/EigenCommunications/ARNotificationsManager.m
```diff
@@ -52,17 +40,60 @@ @interface ARNotificationsManager ()
...
+    if (state) {
+        instance->_state = [state copy];
+    }
+
+    return instance;
+}
+
+- (instancetype)init {
+    if (_sharedInstance) {
+        return _sharedInstance;
```
brainbicycle, line 88: return sharedInstance to retain state, this is the main module that was problematic, didBootstrap would be false on the auto created module so no events would be sent

ios/Artsy/Emission/TemporaryAPI/ARTemporaryAPIModule.m
```diff
@@ -27,8 +29,11 @@ @implementation ARTemporaryAPIModule
 
 RCT_EXPORT_METHOD(markNotificationsRead:(RCTResponseSenderBlock)block)
 {
-    /* In eigen, this should mark the notifications as read using ArtsyAPI */
-    self.notificationReadStatusAssigner(block);
+    [ArtsyAPI markUserNotificationsReadWithSuccess:^(id response) {
+        block(@[[NSNull null]]);
+    } failure:^(NSError *error) {
+        block(@[ RCTJSErrorFromNSError(error)]);
+    }];
```
brainbicycle, line 36: the block assignment was the only bit of state in this module that could be problematic, I believe this was just a historical artifact back when Emission was a library so clients could customize, we don't need it anymore

ios/Artsy/Supporting/QA/ExpoQA.plist
```diff
@@ -0,0 +1,51 @@
...
+	<string>1.0.0</string>
+	<key>EXUpdatesRequestHeaders</key>
+	<dict>
+		<key>expo-channel-name</key>
+		<string>production</string>
+		<key>expo-expect-signature</key>
+		<string>true</string>
+	</dict>
+	<key>EXUpdatesDisableAntiBrickingMeasures</key>
+	<true/>
```
brainbicycle, line 49: QA scheme plist that enables channel switching in the build, substituted at build time via build script

fastlane/Fastfile
```diff
@@ -51,76 +59,103 @@ end
 
 # iOS Lanes
 
-desc 'Ships iOS beta to TestFlight'
-lane :ship_beta_ios do
-  api_key = generate_app_store_connect_api_key
+desc 'Ships iOS beta to TestFlight or Firebase'
+lane :ship_beta_ios do |options|
+  deployment_target = options[:deployment_target] || 'testflight' # default to 'testflight'
```
brainbicycle, line 64: updated the beta lane to support shipping to both testflight and firebase, a fair amount of boilerplate but the main differences are: sign with ad hoc certificates, use the QA scheme when building

ios/Artsy/Emission/Core/AREmission.h
gkartalis: 💀 👋

gkartalis (approved):
🧹 thanks for the additional cleanup ! LEts MEeergeee

PR #12062, opened 2025-05-06
chore: increase parallelism limit and use larger resource class for js tests

MounirDhahri:
[WAITING FOR EARLY FEEDBACK]

### Description

This PR is an attempt to make our JS tests run faster. This is currently just a draft until the job runs and I collect more insights if it makes sense or not.

**Before**
Screenshot 2025-05-06 at 11 18 00

**After**
Screenshot 2025-05-06 at 11 18 08

**Why wasn't the impact of this bigger?**
It looks like our individual tests are slow, some take more than 150 seconds to fully run! this is where I believe we should focus next. Any small improvement there could help us cut a minute or two of the total wait time

**Still open questions**
- Implications on this on 💵 ? any idea @brainbicycle 

### PR Checklist

  - [x] **Android**.
  - [x] **iOS**.

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- increase parallelism limit and use larger resource class for js tests -  mounir

</details>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -247,11 +247,11 @@ jobs:
       #       ./scripts/deploys/expo-updates/deploy-to-expo-updates staging
 
   test-js:
-    parallelism: 8
+    parallelism: 32
     executor:
       name: node/default
       tag: "20.9.0"
-    resource_class: large
+    resource_class: xlarge
     steps:
       - checkout
       - install-node-modules
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -247,11 +247,11 @@ jobs:
       #       ./scripts/deploys/expo-updates/deploy-to-expo-updates staging
 
   test-js:
-    parallelism: 32
+    parallelism: 16
     executor:
       name: node/default
       tag: "20.9.0"
-    resource_class: xlarge
+    resource_class: large
     steps:
       - checkout
       - install-node-modules


ArtsyOpenSource:

- Dev changes (increase parallelism limit and use larger resource class for js tests -  mounir - MounirDhahri)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 2d5dd11a4588023ec44fa282e9b8a2ff5866e436
</p>

.circleci/config.yml
```diff
@@ -247,11 +247,11 @@ jobs:
       #       ./scripts/deploys/expo-updates/deploy-to-expo-updates staging
 
   test-js:
-    parallelism: 8
+    parallelism: 32
     executor:
       name: node/default
       tag: "20.9.0"
-    resource_class: large
+    resource_class: xlarge
```
MounirDhahri, line 254: I will try this out with `large` next.

gkartalis, line 254: does this affect the cost though? and if yes how 👁️

MounirDhahri: yes, that makes sense. I asked @brainbicycle and let's see, I think we have already a performance plan, not sure how stretched we are with it and if we have some extra room

MounirDhahri: Screenshot 2025-05-06 at 15 11 52
We would still be within our plan

brainbicycle: I think we should not do this as it will be a significant price increase. 
xlarge (x86) images use 40 credits per minute and the large (x86) images use 20 credits per minute: https://circleci.com/pricing/price-list/, so it essentially doubles the cost, that plus the increased parallelism and we are adding up quickly. 

**Before:**
8 parallel jobs * 20 credits per minute * ~10minutes per job * 193 jobs per month = 308,800 credits per month

**After:**
32 parallel jobs * 40 credits per minute * ~6minutes per job * 193 jobs per month = 1,482,240 credits per month

**Additional Credits:** 
1,482,240 - 308,800 = 1,173,440 credits

**Additional Cost:**
1,173,440 credits × $0.0006/credit = $704.06 per month

brainbicycle: I do feel the pain of the long test times though so wondering about alternatives:
- have we tried a modest increase in parallelism with the existing resource class?
 - "It looks like our individual tests are slow, some take more than 150 seconds to fully run! this is where I believe we should focus next. Any small improvement there could help us cut a minute or two of the total wait time" I think it would be good to look into why this is, I suspect we might be able to clean up some setupJest / shared setup stuff of find some dangling resources and speed this up
 - (bigger lift) what about faster runners or tools like bun?

MounirDhahri: I have a very intresting finding! with `large` and `16` instead of `32` parallelism, the result is exactly the same!

gkartalis (approved):

.circleci/config.yml
MounirDhahri: tbh, I think we can go with even `medium+` since we don't get over 6gb but I don't think it's worth it given how little that would save us

Screenshot 2025-05-06 at 18 22 09

brainbicycle: much more modest increase! seems reasonable to me!
16 parallel jobs * 20 credits per minute * ~7minutes per job * 193 jobs per month = 432,320

brainbicycle (approved):
thanks!!

PR #12098, opened 2025-05-08
chore(ci): attempt to make flaky test less flaky 

brainbicycle:
This PR resolves [] 

### Description

I can't reproduce this locally but my best guess is the second button press can happen while the first press or UI transition is still in flight leading the 2nd not to be handled. This just explicitly waits for the button to be visible and enabled. 

#nochangelog #trivial


iskounen (approved):
Thank you for fixing this.

PR #12114, opened 2025-05-12
fix: memoize the theme to prevent navigating to home view when changing dark mode settings

dariakoko:
This PR resolves [PBRW-762] 

### Description
Memoize the theme to prevent navigating to home view when changing dark mode settings

| iOS | Android |
| --- | --- |
| https://github.com/user-attachments/assets/933a7689-c9f9-4151-a7f1-93e566c68860 | https://github.com/user-attachments/assets/2ad74613-9460-402e-9a57-1a4a65857b1a | 

### PR Checklist

  - [x] **Android**.
  - [x] **iOS**.

### To the reviewers 👀

- Memoize the theme to prevent navigating to home view when changing dark mode settings

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[PBRW-762]: https://artsyproduct.atlassian.net/browse/PBRW-762?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Providers.tsx
+++ b/src/app/Providers.tsx
@@ -10,7 +10,7 @@ import { useFeatureFlag } from "app/utils/hooks/useFeatureFlag"
 import { ProvideScreenDimensions } from "app/utils/hooks/useScreenDimensions"
 import { NavigationTestsProvider } from "app/utils/tests/NavigationTestsProvider"
 import { postEventToProviders } from "app/utils/track/providers"
-import { Suspense } from "react"
+import { Suspense, useMemo } from "react"
 import { GestureHandlerRootView } from "react-native-gesture-handler"
 import { SafeAreaProvider } from "react-native-safe-area-context"
 import { RelayEnvironmentProvider } from "react-relay"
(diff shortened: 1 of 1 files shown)


src/app/Providers.tsx
```diff
@@ -110,7 +110,10 @@ const TrackingProvider: React.FC = ({ children }) => {
 function ThemeWithDarkModeSupport({ children }: { children?: React.ReactNode }) {
   const supportDarkMode = useFeatureFlag("ARDarkModeSupport")
   const colorScheme = GlobalStore.useAppState((state) => state.devicePrefs.colorScheme)
-  const theme = supportDarkMode ? (colorScheme === "dark" ? "v3dark" : "v3light") : undefined
+
+  const theme = useMemo(() => {
+    return supportDarkMode ? (colorScheme === "dark" ? "v3dark" : "v3light") : undefined
+  }, [colorScheme])
```
olerichter00, line 116: I don't think this is fixing the issue because `theme` would still be updated when `colorScheme` changes and causes a rerender.

I cannot reproduce the bug with the simulator and only see it on a real device. It could be related to the navigation state. I'm seeing [this](https://github.com/artsy/eigen/blob/1ce7ff7e937f2095840c567061a19891a4352b99/src/app/Navigation/Navigation.tsx#L127) message when I update the color scheme in the settings.

olerichter00, line 116:
```suggestion
  }, [colorScheme, supportDarkMode])
```

dariakoko: Yess, it is the same for me. I tested this branch on my phone as well and i couldn't see the issue with these changes. But on the build from testflight I can reproduce it every time

Chat gpt suggested to me this fix saying "Without useMemo, the theme object is recreated on each render, causing the entire NavigationContainer to unmount and reset state." I looked into the navigation states as well, and it feels like it makes sense  

What do you think about merging it to see if it will resolve the issue? I'd test it on the nightly beta and if it doesn't fix the issue I'll revert the PR. If it does fix the issue we cherry pick 🍒

PR #12153, opened 2025-05-18
fix: broken myCollection tests after flag enabled

brainbicycle:
This PR resolves [] 

### Description

- one failing test was a real bug, the analytics call was in Button onPress in ZeroState which was swallowed by RouterLink causing it not to be fired.
- minor capitalization difference in new design
- minor copy difference in new design

### PR Checklist

  - [ ] **Android**.
  - [ ] **iOS**.

### To the reviewers 👀

- fix broken analytics in MyCollection zero state - brian 

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Scenes/MyCollection/Components/MyCollectionZeroArtists.tsx
+++ b/src/app/Scenes/MyCollection/Components/MyCollectionZeroArtists.tsx
@@ -41,14 +41,14 @@ export const MyCollectionZeroArtists: React.FC<{}> = () => {
             flexDirection={isFontScaleLarge() ? "column" : "row"}
             justifyContent="space-evenly"
           >
-            <RouterLink to="/my-collection/collected-artists/new" hasChildTouchable>
-              <Button
-                testID="add-artwork-button-zero-state"
-                block={isFontScaleLarge()}
-                onPress={() => {
-                  // TODO: Implement analytics
-                }}
-              >
+            <RouterLink
+              to="/my-collection/collected-artists/new"
+              onPress={() => {
+                // TODO: Implement analytics
+              }}
+              hasChildTouchable
+            >
+              <Button testID="add-artwork-button-zero-state" block={isFontScaleLarge()}>
                 Add Artists
               </Button>
             </RouterLink>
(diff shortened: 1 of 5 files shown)


src/app/Scenes/MyCollection/Components/MyCollectionZeroArtists.tsx
```diff
@@ -41,14 +41,14 @@ export const MyCollectionZeroArtists: React.FC<{}> = () => {
             flexDirection={isFontScaleLarge() ? "column" : "row"}
             justifyContent="space-evenly"
           >
-            <RouterLink to="/my-collection/collected-artists/new" hasChildTouchable>
-              <Button
-                testID="add-artwork-button-zero-state"
-                block={isFontScaleLarge()}
-                onPress={() => {
```
brainbicycle, line 48: cc @olerichter00 I wonder if this is something we can catch with a lint rule? that is using onPress within a child component of RouterLink? Just a thought

ArtsyOpenSource:

- Cross-platform user-facing changes (fix broken analytics in MyCollection zero state - brian - brainbicycle)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 399e9870a7392658ff7fe89a326c2c4427ab62d1
</p>

src/app/Scenes/MyCollection/Screens/Insights/MyCollectionInsights.tests.tsx
```diff
@@ -45,7 +45,7 @@ describe("MyCollectionInsights", () => {
 
     await flushPromiseQueue()
 
-    expect(getByText("Gain Deeper Knowledge of your Collection")).toBeTruthy()
+    expect(getByText("Gain deeper knowledge of your collection")).toBeTruthy()
```
brainbicycle, line 48: not sure if the previous or the new is the correct case we want here but the test was currently broken so I defaulted to the newer

src/app/Scenes/MyCollection/Components/MyCollectionZeroArtists.tsx
olerichter00: Yes, this is a very good idea 👍 Potentially, we should also consider using another pattern for buttons that use navigate. In `Palette` for web, it is possible to add an `as={RouterLink}` prop to `Button` (instead of `hasChildTouchable`):

```
<Button as={RouterLink} to="/my-url"></Button>
```

olerichter00: https://github.com/artsy/eigen/pull/12157

PR #12221, opened 2025-06-04
chore: fix artwork grid filtering on collect

MounirDhahri:
### Description

https://www.notion.so/artsy/App-Art-under-X-quick-link-and-the-Collect-screen-1fbcab0764a08069833ad682209e0cd5?p=1fbcab0764a080ee8ad9e74b0cb96079&pm=s

This PR fixes some issues happening at the Collect screen and describes how I could debug things.

**TLDR:**
- The query was timing out because we were querying for too many aggregations.

**Debugging process:**
1. I took the query to try it out in Insomnia, and it was still timing out. I assumed there is either something wrong with my input, or the payload we are sending
2. After removing some of the aggregations, the query started to work and the more I removed, the faster the query became
3. I went to Force to inspect the query because it's working fine there
4. I noticed that we are using less aggregations there! Yet still showing the filter options (for example the size and color filters)
5. I got the idea of removing the slow aggregations in Eigen that are not available in Force
6. Some filters disappeared! It turned out in Eigen, we show a lot of filters based on the presence of aggregations.

**Fix:**
- I removed the aggregations that are not used in Force
- I updated the logic for some filters in Eigen to not show them based on the aggregations, just like at Force.
- Bonus: Prefetching was broken, I fixed it for a better UX

<Video src="https://github.com/user-attachments/assets/4f80c17d-245a-4eef-b750-aaccb644dad9" />

### PR Checklist

  - [ ] **Android**.
  - [x] **iOS**.
- [x] I have added a **changelog entry** below, or my changes do not require onex

### To the reviewers 👀

- improve artwork grid performance - mounir

#### iOS user-facing changes

-

-

#### Dev changes

- fix prefetching logic on collect pills - mounir

</details>

### DIFF
--- a/src/app/Scenes/Collect/Collect.tsx
+++ b/src/app/Scenes/Collect/Collect.tsx
@@ -148,7 +148,16 @@ export const viewerFragment = graphql`
       first: $count
       after: $cursor
       input: $input
-      aggregations: [ARTIST, ARTIST_NATIONALITY, LOCATION_CITY, MATERIALS_TERMS, PARTNER, TOTAL]
+      aggregations: [
+        ARTIST
+        ARTIST_NATIONALITY
+        LOCATION_CITY
+        MAJOR_PERIOD
+        MATERIALS_TERMS
+        MEDIUM
+        PARTNER
+        TOTAL
+      ]
     ) @connection(key: "CollectArtworks_artworksConnection") {
       aggregations {
         slice
--- a/src/app/Scenes/Collect/Collect.tsx
+++ b/src/app/Scenes/Collect/Collect.tsx
@@ -148,19 +148,7 @@ export const viewerFragment = graphql`
       first: $count
       after: $cursor
       input: $input
-      aggregations: [
-        ARTIST
-        ARTIST_SERIES
-        COLOR
-        DIMENSION_RANGE
-        LOCATION_CITY
-        MAJOR_PERIOD
-        MATERIALS_TERMS
-        MEDIUM
-        PARTNER
-        PRICE_RANGE
-        ARTIST_NATIONALITY
-      ]
+      aggregations: [ARTIST, ARTIST_NATIONALITY, LOCATION_CITY, MATERIALS_TERMS, PARTNER, TOTAL]
     ) @connection(key: "CollectArtworks_artworksConnection") {
       aggregations {
         slice
--- a/src/app/Scenes/Collect/Collect.tsx
+++ b/src/app/Scenes/Collect/Collect.tsx
@@ -93,7 +93,7 @@ export const CollectContent: React.FC<CollectContentProps> = ({ viewer }) => {
         animated
         artworks={artworks}
         isLoading={isLoadingNext}
-        loadMore={loadNext}
+        loadMore={() => loadNext(PAGE_SIZE)}
         hasMore={hasNext}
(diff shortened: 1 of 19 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (improve artwork grid performance - mounir - MounirDhahri)
- Dev changes (fix prefetching logic on collect pills - mounir - MounirDhahri)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against e9b3ed7bb434e2f62cf365295d2938e79a0d1f23
</p>

src/app/Scenes/Collect/Collect.tsx
```diff
@@ -193,11 +190,19 @@ export const collectQuery = graphql`
 
 type CollectQueryRendererProps = StackScreenProps<any, any>
 
+export const prepareCollectVariables = (params: {
```
MounirDhahri, line 193: I extracted this into a function to make sure that we can use it for the prefetching

olerichter00 (commented):
Thanks for looking into the issue 🙏 

If we really want to make the filter options independent from the aggregations, this is the right approach. If we could benefit from using the aggregations, we should consider adding the missing indexes in Gravity instead ([here](https://github.com/artsy/gravity/blob/main/app/models/search/indexes/artwork_filter_index.rb#L20), [example PR](https://github.com/artsy/gravity/pull/13808)).

MounirDhahri:
That's a very good point about indexing these aggregations. I actually looked into that but colors and prices were already indexed. The only field that wasn't indexed was the sizes (maybe a related field is), but we didn't have any frontend logic that was using that aggregation or displaying the available slices, it's why I deleted it and used the same logic as we have in Force

olerichter00:
> I actually looked into that but colors and prices were already indexed

I wonder if they are indexed correctly for the aggregations. If they are indexed, the query should be fast.

> but we didn't have any frontend logic that was using that aggregation or displaying the available slices

If the aggregations are only used on the collect screen, I suppose it's a good idea to just remove them. However, if Metaphysics returns the aggregations, it may be a good idea to see if we can improve their performance.

anandaroop (approved):
Thanks! I had to refresh my own understanding of how this works. 

I think this is sort of implied in the work done in this PR, but I just want to state this for the record —

On web there are a few filters that are rendered _without_ any consideration of aggregation data because we simply show **_all_** the filter options, instead of limiting it to what's present in the artwork results.

| Rarity | Ways to buy | Size | Color |
|--------|--------|--------|--------|
| rarity |ways |size |color | 

☝🏽 **Rarity** and **Ways to Buy** work this way — as well as **Size** and **Color** (which makes the `DIMENSION_RANGE` and `COLOR` aggregations unnecessary, as you found). 

And I *believe* `PRICE_RANGE` represents an old UI which was replaced by the price histogram+sliders.

Some of the aggregations can be quite expensive, so thanks for weeding out the ones we don't actually want 🙏🏽

anandaroop:
My only concern here would be whether relying on aggregations was a deliberate decision on mobile, to limit the amount of back and forth you do in the UI if you choose filters that result in 0 hits. But addressing the performance issues might be the bigger concern here.

MounirDhahri:
I can follow up on your concerns as well @anandaroop. It was from a long time ago so I hope I find someone with full context

PR #12222, opened 2025-06-04
feat(a11y): Add fontScale tracking

gkartalis:
This PR resolves [] 

### Description

Adds fontscale tracking to the app in order to get an idea of what are the fontscales that our users usually use.
Enhances and renames the `useScreenReaderTracking` to include the above under `useScreenReaderAndFontScaleTracking` name.

Below you can see how we track it in segment:

Screenshot 2025-06-13 at 14 27 14

### PR Checklist

  - [ ] **Android**.
  - [ ] **iOS**.

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- enhance and rename useScreenReader hook with fontscale size to useScreenReaderAndFontScaleTracking

</details>


src/app/utils/useFontScaleTracking.ts
```diff
@@ -0,0 +1,26 @@
...
+// the purpose of this hook is to track once the fontScale of the user's device
+export const useFontScaleTracking = () => {
+  const { trackEvent } = useTracking()
+
+  useEffect(() => {
+    const getFontScale = async () => {
+      try {
+        const fontScale = await DeviceInfo.getFontScale()
+
+        postEventToProviders({ name: "FontScale", value: fontScale })
```
gkartalis: is there a better approach for this?

ArtsyOpenSource:

- Dev changes (adds hook to track users fontscale - gkartalis)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 2b2fc6f66c2f660342f26aebc78fa1fcf9a284f2
</p>

src/app/utils/useFontScaleTracking.ts
MounirDhahri: I am fine with this tbh. Maybe we can even track other things like `isScreenReaderEnabled`.

src/app/utils/useFontScaleTracking.ts
```diff
@@ -0,0 +1,26 @@
...
+import { useTracking } from "react-tracking"
+
+// the purpose of this hook is to track once the fontScale of the user's device
+export const useFontScaleTracking = () => {
+  const { trackEvent } = useTracking()
+
+  useEffect(() => {
+    const getFontScale = async () => {
+      try {
+        const fontScale = await DeviceInfo.getFontScale()
```
MounirDhahri: did you intentionally use `DeviceInfo`, I have been using `PixelRatio` from RN in other parts

src/app/utils/useFontScaleTracking.ts
```diff
@@ -0,0 +1,26 @@
...
+
+        return fontScale
+      } catch (error) {
+        console.error("Error getting font scale: ", error)
+      }
+      return
+    }
+
+    getFontScale()
+  }, [trackEvent])
```
MounirDhahri: is this to make sure tracking is ready?

gkartalis: exactly

src/app/utils/useFontScaleTracking.ts
gkartalis: 🤔 not sure which of the two is the best for this specific job - I can briefly look into the pixelRatio as well

gkartalis:
> I am fine with this tbh. Maybe we can even track other things like isScreenReaderEnabled.

https://github.com/artsy/eigen/blob/main/src/app/utils/useScreenReaderTracking.ts

We do have this in place for two years now :D

MounirDhahri:
helloo little guy 😄 i remember you added it but I didn't know it's still there. Thanks for the hint

src/app/utils/useFontScaleTracking.ts
MounirDhahri: no preferences, I bet it's using the same API anyway

src/app/utils/useFontScaleTracking.ts
brainbicycle: It might make sense to use a similar approach to the screen reader tracking with the identify call, this will add it to the user traits like so: 
```
" "traits": {
    "user interface style": "light",
    "is_temporary_user": 0,
    "screen reader status": "disabled",
    "has enabled notifications": 1
  }," 

```
rather than posting an event. It _seems_ like more of a user trait then an event to me but I am not sure fi that has consequences for analysis, I think we should have access to both

MounirDhahri (approved):

PR #12241, opened 2025-06-10
fix: install node modules before determine-deployment-type

brainbicycle:
This PR resolves [] 

### Description

Fixes failure in main due to fingerprint dep not being available.

Successful job:
https://app.circleci.com/pipelines/github/artsy/eigen/70377/workflows/09500dc8-7ea3-48bd-b96a-0179b8c37735/jobs/245395

### Follow-ups
- [ ] Migrate ci tools to bin pattern to avoid needing to install all node modules

#nochangelog

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -605,7 +605,6 @@ workflows:
             branches:
               only:
                 - main
-                - brian/main-break
 
       - test-js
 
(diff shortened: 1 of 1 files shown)


gkartalis (approved):
thanks

PR #12246, opened 2025-06-10
fix: broken pagination on collect screen

MounirDhahri:
This PR resolves [https://www.notion.so/artsy/App-Art-under-X-quick-link-and-the-Collect-screen-1fbcab0764a08069833ad682209e0cd5] 

### Description

This PR fixes pagination issues on the collect screen. 

**Debugging instructions:**
- Pagination was broken, so i thought we are either sending different params or the query itself is broken. I didn't assume that MasonryGrid is broken since we use it everywhere
- I opened Reactron to debug pagination requests, but everything seems fine. The payload made sense and the cursor and the pages also made sense
- I noticed that when the pagination fails, the query as well fail. Tada! it's a query issue then
- I took the payload from Reactron and pasted it inside Insomnia
- The query timed out!
- I started tweaking the query, then I noticed that we are querying for the aggregations always! which isn't needed
- I looked into the implementation and it was safe to extract the aggregations query to a separate fragment and optimise it by querying for `first:0`and not getting artworks.
- 💯 that did the trick

https://github.com/user-attachments/assets/2957b52b-4de4-4bd0-8426-de9f186d75e3

https://github.com/user-attachments/assets/d882e305-74fa-4464-a80d-0dba77013e78

### PR Checklist

  - [x] **Android**.
  - [x] **iOS**.

### To the reviewers 👀

- broken pagination on collect screen - mounir

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Scenes/Collect/Collect.tsx
+++ b/src/app/Scenes/Collect/Collect.tsx
@@ -2,7 +2,12 @@ import { ContextModule, OwnerType } from "@artsy/cohesion"
 import { Flex, Screen, SimpleMessage, Spacer, Text } from "@artsy/palette-mobile"
 import { StackScreenProps } from "@react-navigation/stack"
 import { CollectArtworks_viewer$key } from "__generated__/CollectArtworks_viewer.graphql"
-import { CollectQuery, FilterArtworksInput } from "__generated__/CollectQuery.graphql"
+import { CollectArtworks_viewerAggregations$key } from "__generated__/CollectArtworks_viewerAggregations.graphql"
+import {
+  CollectQuery,
+  CollectQuery$data,
+  FilterArtworksInput,
+} from "__generated__/CollectQuery.graphql"
 import { ArtworkFilterNavigator, FilterModalMode } from "app/Components/ArtworkFilter"
 import {
   FilterArray,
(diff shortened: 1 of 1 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (broken pagination on collect screen - mounir - MounirDhahri)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 15f10d3a562e28aa414ff01a59b66d645e05cc3c
</p>

nickskalkin (approved):
Thanks! Love that you attached your journey to the PR description!

I guess the only downside of this approach is that whenever a user opens the Collect screen - it will fire 2 separate somewhat heavy requests - one to fetch artworks, another one to fetch aggregations? IMO that's all right, at least for now

MounirDhahri:
>  it will fire 2 separate somewhat heavy requests - one to fetch artworks, another one to fetch aggregations

@nickskalkin that's almost it but for Relay, it is just one query that requests both at the same time. Because I have both the aggregations and the artworks query inside `CollectQuery` and within the same `useLazyLoadQuery`, we will make only one request that gets both. However for pagination, only the artworks query query will be triggered on `loadNext`

PR #12273, opened 2025-06-13
fix: broken edit saved search

MounirDhahri:
This PR resolves [ONYX-1738] 

### Description

This PR fixes the broken edit on saved search.

**Debugging process:**
Untitled-2024-10-09-2321

We have a few possible solutions if we want to keep things as they are:

- When the user updates an alert, mutate the store value 
- Update the store value when the network request is done. This isn't straightforward - so we better just make the query `network-only` 
>> I initially did this but removed it because the solution above was enough for the alert. It remains however fragile in case we query for another interface at some point. I would say we cross that bridge when we reach it but I am open to your thoughts here.

https://github.com/user-attachments/assets/db351dff-3ead-495f-9762-7778abf57c0b

### PR Checklist

  - [x] **Android**.
  - [x] **iOS**.

### To the reviewers 👀

- fix: broken edit saved search - mounir

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[ONYX-1738]: https://artsyproduct.atlassian.net/browse/ONYX-1738?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/SavedSearchAlert/SavedSearchAlert.tsx
+++ b/src/app/Scenes/SavedSearchAlert/SavedSearchAlert.tsx
@@ -6,33 +6,39 @@ export const SavedSearchAlertScreenQuery = graphql`
   query SavedSearchAlertQuery($savedSearchAlertId: String!) {
     me {
       alert(id: $savedSearchAlertId) {
-        acquireable
-        additionalGeneIDs
-        artistIDs
-        artistSeriesIDs
-        atAuction
-        attributionClass
-        colors
-        dimensionRange
-        displayName
-        sizes
-        height
-        inquireableOnly
-        locationCities
-        majorPeriods
-        materialsTerms
-        offerable
-        partnerIDs
-        priceRange
-        settings {
-          email
-          name
-          push
-          details
-        }
-        width
+        ...SavedSearchAlert_alert @relay(mask: false)
       }
     }
+  }
+`
+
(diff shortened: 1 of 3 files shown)


src/app/Scenes/SavedSearchAlert/SavedSearchAlert.tsx
```diff
@@ -6,36 +6,42 @@ export const SavedSearchAlertScreenQuery = graphql`
...
-        partnerIDs
-        priceRange
-        settings {
-          email
-          name
-          push
-          details
-        }
-        width
+        ...SavedSearchAlert_alert @relay(mask: false)
```
MounirDhahri, line 9: I am unmasking this fragment because the values of the response from this mutation are used directly and it led to multiple ts issues.

ArtsyOpenSource:

- Cross-platform user-facing changes (fix: broken edit saved search - mounir - MounirDhahri)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 3a96198bc5de6a76fc992df2370ddd7a6020b7c7
</p>

dblandin (approved):
Looks good 👍🏼

PR #12256, opened 2025-06-11
chore: enforce eas version

brainbicycle:
This PR resolves [] 

### Description

This takes another approach to enforcing eas cli version, adds a bin directory with its own package.json exclusively for shared cli tooling not used in normal dev work.

### PR Checklist

- [ ] I have tested my changes on the following platforms:
  - [ ] **Android**.
  - [ ] **iOS**.
- [x] I hid my changes behind a **[feature flag]**, or they don't need one.
- [x] I have included **screenshots** or **videos** at least on **Android**, or I have not changed the UI.
- [x] I have added **tests**, or my changes don't require any.
- [x] I added an **[app state migration]**, or my changes do not require one.
- [x] I have documented any **follow-up work** that this PR will require, or it does not require any.
- [x] I have added a **changelog entry** below, or my changes do not require one.

### To the reviewers 👀

- [ ] I would like **at least one** of the reviewers to **run** this PR on the simulator or device.

<details><summary>Changelog updates</summary>

#### Cross-platform user-facing changes

-

#### iOS user-facing changes

-

#### Android user-facing changes

-

#### Dev changes

- add bin pattern to enforce cli version - brian

</details>

Need help with something? Have a look at our [docs], or get in touch with us.

[app state migration]: ../blob/main/docs/adding_state_migrations.md
[feature flag]: ../blob/main/docs/developing_a_feature.md
[docs]: ../blob/main/docs/README.md

PR #12275, opened 2025-06-13
fix: sentry cli not found in time during fastlane deploy

gkartalis:
This PR resolves [] 

### Description

Brings back `sh("yarn add @sentry/cli -D")`. It was removed [here](https://github.com/artsy/eigen/pull/12256/files#diff-f93431d15b5981ca9a3ef1538b1bad0a059b541c31ba42500ef6da9e40638982L8)

Reasoning: 

Creating a release for sentry will **always** fail unless you are on the latest `@sentry/cli` version. The reason we are getting these messages [here](https://artsy.slack.com/archives/C02BAQ5K7/p1749797086500379) is because we moved sentry-cli to the `./bin` directory. 

Sentry cli is used only on this fastlane file and IMO it's gonna be way better to install it there only on demand vs having it on the bin file because it is always gonna have the latest version.

If we dont do that the feedback loop is gonna be like so:
- we get similar sentry cli is outdated and creating the sentry release failed errors in slack
- we need to go and open a pr with a dependency bump 
- trigger a new beta with the upgraded sentry-cli version

More context on when we moved sentry-cli installation in the fastlane file [here](https://github.com/artsy/eigen/pull/10767)

Followup of https://github.com/artsy/eigen/pull/12256

#nochangelog

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -491,7 +491,7 @@ lane :deploy_to_expo_updates do |options|
       popd > /dev/null")
   build_folder = File.expand_path("../dist/")
 
-  sh("./scripts/setup/instal-bin")
+  sh("yarn add @sentry/cli -D")
   sentry_cli_path = 'bin/node_modules/@sentry/cli/bin/sentry-cli'
 
   upload_expo_sourcemaps(
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -491,6 +491,7 @@ lane :deploy_to_expo_updates do |options|
       popd > /dev/null")
   build_folder = File.expand_path("../dist/")
 
+  sh("./scripts/setup/instal-bin")
   sentry_cli_path = 'bin/node_modules/@sentry/cli/bin/sentry-cli'
 
   upload_expo_sourcemaps(
--- a/fastlane/sentry_fastlane.rb
+++ b/fastlane/sentry_fastlane.rb
@@ -5,7 +5,8 @@ lane :upload_sentry_artifacts do |options|
   sentry_release_name = options[:sentry_release_name]
   platform = options[:platform]
   dist_version = options[:dist_version]
-  sentry_cli_path="./bin/node_modules/@sentry/cli/bin/sentry-cli"
+  sh("yarn add @sentry/cli -D")
+  sentry_cli_path="node_modules/@sentry/cli/bin/sentry-cli"
 
   project_slug = 'eigen'
   org_slug = 'artsynet'


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  ✅ **No changelog changes**
  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 0bfbecb761e7d3981bd17c29781c19123fcc3893
</p>

fastlane/Fastfile
```diff
@@ -491,6 +491,7 @@ lane :deploy_to_expo_updates do |options|
       popd > /dev/null")
   build_folder = File.expand_path("../dist/")
 
+  sh("./scripts/setup/instal-bin")
```
MounirDhahri:
```suggestion
  sh("./scripts/setup/install-bin")
```

MounirDhahri (commented):
I wonder if it makes sense to add this to `./scripts/deploys/expo-updates/deploy-to-expo-updates` instead. I know we have other install scripts within `FastFile` already but I feel in this case, it's kind of like `setup:releases`, a one time thing that we can extract

fastlane/Fastfile
gkartalis: woops nice catch.

gkartalis:
> I wonder if it makes sense to add this to ./scripts/deploys/expo-updates/deploy-to-expo-updates instead. I know we have other install scripts within FastFile already but I feel in this case, it's kind of like setup:releases, a one time thing that we can extract

Actually this is a different issue kinda - I will update the pr and elaborate in a bit

gkartalis:
@MounirDhahri I updated the pr and the approach and added more context on the issue

MounirDhahri:
Great, thanks @gkartalis

MounirDhahri (approved):
LGTM

gkartalis:
will continue addressing the other two fastlane places that this was called and remove it from `.bin`

MounirDhahri (commented):
Tbh, I would be also fine with us adding `@sentry/cli` as a dev dependency.

gkartalis:
I get the frustration. 

Thinking how we can do to make the experience better:

- keep the `sh("yarn add @sentry/cli -D")` in fastlane so we wont be relying to the pinned version of the `.bin/package.json` dependency thus we wont need to open update prs and avoid the lengthy feedback loop
- keep the `@sentry/cli` in the `./bin/package.json` so we as devs when deploying a beta will not have to revert the `package.json` and `yarn.lock` changes every time but when it fails locally with error like:

```
Your sentry-cli is outdated, please upgrade to at least version 2.46.0 and start your lane again!
```

Devs should be responsible of upgrading + commiting this diff.

How does that sound?

MounirDhahri:
Yes, that sounds reasonable. Do we have to upgrade sentry cli so often though?

gkartalis:
They bump it every ~ month or so generally - it was a huge relief for me after I added it on fastfile back then. https://github.com/artsy/eigen/pulls?q=is%3Apr++is%3Aclosed+%22%40sentry-cli%22 

not all of those are the bump but you get the gist of it - the frequency was from every 1 month mostly to every three months

MrSltun (dismissed):

MounirDhahri:
Clear - thanks @gkartalis

gkartalis:
I opened the ☝️ and will keep this open for discussion for a bit

brainbicycle:
Whoops sorry for the trouble, I forgot about sentry requiring the latest always 😬 

I think that ideally we would not manually add deps this way because:
1. It breaks reproducibility, a build with the exact same settings can behave differently tmrw then today, we are violating this elsewhere I know but it is something I think we should aim for
2. We can't cache it
3. (minor annoyance) running it locally modifies package.json and shouldn't be committed

Thinking of alternatives:
1. we could add "@sentry-cli": "latest" to the bin directory, I think this works but also breaks reproducibility
2. we could configure dependabot to aggressively update sentry cli every time an update is available and automerge

I think 2 is probably best, so would like to look into that one but not opposed to merging this in the meantime if we want to.

gkartalis:
I bumped sentry-cli for now so I don't think this should be merged - its more of a conversation starter, The dependabot configuration does make sense - wandering if we should do anything extra to make it work for the bin directory though

PR #12175, opened 2025-05-23
feat: cross platform city guide

MounirDhahri:
### Description

This PR migrates the city guide from native/react-native to react-native.

**Please note: I kept the logic as it is, I didn't change much to avoid breaking stuff I don't really understand**

### How to QA this?
You can install these specific betas which have the city guide in react-native.
**Android**: 167464192
**iOS**: 2025.5.27.12

### What comes next?
Removing the native code

### Opportunities?
- With this work, we would be able to support city guide, for every city. I migrated the small helper that gives you the nearest city to your location to RN. It would make supporting other cities easier.
- Maybe we don't need a city, just a longitude and lattitude and we get a list of events
- A step closer to a regular react-native app.

### When can this be merged?
Because this touches many things, and can not be hidden behind a feature flag, we can only merge this after wrapping up it QA.

| Android | iOS |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/0f1685c4-b42d-4a66-8a68-e1322aa69c3c" /> | <video src="https://github.com/user-attachments/assets/bf29b1ae-d901-4b20-a2fb-4a4b30693d6b" /> | 

### PR Checklist

- [x] I have tested my changes on the following platforms:
  - [x] **Android**.
  - [x] **iOS**.
- [x] I hid my changes behind a **[feature flag]**, or they don't need one.
- [x] I have included **screenshots** or **videos** at least on **Android**, or I have not changed the UI.
- [x] I have added **tests**, or my changes don't require any.
- [x] I added an **[app state migration]**, or my changes do not require one.
- [x] I have documented any **follow-up work** that this PR will require, or it does not require any.
- [x] I have added a **changelog entry** below, or my changes do not require one.

### To the reviewers 👀

- [ ] I would like **at least one** of the reviewers to **run** this PR on the simulator or device.

<details><summary>Changelog updates</summary>

#### Cross-platform user-facing changes

- migrate cross platform to react-native - mounir

#### iOS user-facing changes

-

#### Android user-facing changes

-

#### Dev changes

-

</details>

Need help with something? Have a look at our [docs], or get in touch with us.

[app state migration]: ../blob/main/docs/adding_state_migrations.md
[feature flag]: ../blob/main/docs/developing_a_feature.md
[docs]: ../blob/main/docs/README.md

PR #12286, opened 2025-06-13
feat(city guide): enable expanded list of cities

anandaroop:
### Description

Co-authored-by: @MounirDhahri 

To accompany the enablement of cross-platform City Guide (#12175) here we also expand the list of cities from the original six that were supported at launch.

Which cities to include? Luckily we already have a concept of "[featured cities](https://github.com/artsy/gravity/blob/93d85b912b466bcd832b95538168be1d4b4326a5/app/services/city_geocoding_service.rb#L3-L27)" which we make use of in various places e.g. on https://www.artsy.net/shows

So here we create an expanded list of cities based on that existing concept.

For QA purposes, this is behind a feature flag: **Enable expanded city list in City Guide**

| iOS | Android |
|---|---|
| <video src="https://github.com/user-attachments/assets/cd0720f2-4c76-48c2-bef9-54b847572408" width="400" /> | <video src="https://github.com/user-attachments/assets/1f649359-efd2-465f-a30d-2fe34366c6da" width="400" /> |

### PR Checklist

  - [x] **Android**.
  - [x] **iOS**.

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- Expanded list of City Guide cities behind feature flag

</details>

### DIFF
--- a/data/cityDataSortedByDisplayPreference-expanded.json
+++ b/data/cityDataSortedByDisplayPreference-expanded.json
@@ -1,0 +1,176 @@
+[
+  {
+    "slug": "new-york-ny-usa",
+    "name": "New York",
+    "coordinates": { "lat": 40.71, "lng": -74.01 },
+    "maxBounds": {
+      "sw": { "lat": 40.144, "lng": -74.754 },
+      "ne": { "lat": 41.272, "lng": -73.266 }
+    }
+  },
+  {
+    "slug": "london-united-kingdom",
+    "name": "London",
+    "coordinates": { "lat": 51.51, "lng": -0.13 },
+    "maxBounds": { "sw": { "lat": 50.895, "lng": -1.118 }, "ne": { "lat": 52.127, "lng": 0.862 } }
+  },
+  {
+    "slug": "los-angeles-ca-usa",
+    "name": "Los Angeles",
+    "coordinates": { "lat": 34.05, "lng": -118.24 },
+    "maxBounds": {
+      "sw": { "lat": 33.453, "lng": -118.959 },
+      "ne": { "lat": 34.644, "lng": -117.521 }
+    }
+  },
+  {
+    "slug": "paris-france",
+    "name": "Paris",
+    "coordinates": { "lat": 48.86, "lng": 2.35 },
+    "maxBounds": { "sw": { "lat": 48.339, "lng": 1.561 }, "ne": { "lat": 49.377, "lng": 3.139 } }
+  },
+  {
+    "slug": "berlin-germany",
+    "name": "Berlin",
+    "coordinates": { "lat": 52.52, "lng": 13.4 },
+    "maxBounds": { "sw": { "lat": 51.882, "lng": 12.352 }, "ne": { "lat": 53.159, "lng": 14.452 } }
+  },
+  {
+    "slug": "miami-fl-usa",
+    "name": "Miami",
+    "coordinates": { "lat": 25.76, "lng": -80.19 },
+    "maxBounds": {
+      "sw": { "lat": 25.131, "lng": -80.887 },
+      "ne": { "lat": 26.386, "lng": -79.493 }
+    }
+  },
+  {
+    "slug": "san-francisco-ca-usa",
+    "name": "San Francisco",
+    "coordinates": { "lat": 37.77, "lng": -122.42 },
+    "maxBounds": {
+      "sw": { "lat": 37.19, "lng": -123.152 },
+      "ne": { "lat": 38.347, "lng": -121.688 }
+    }
+  },
+  {
+    "slug": "hong-kong-hong-kong",
+    "name": "Hong Kong",
+    "coordinates": { "lat": 22.28, "lng": 114.16 },
+    "maxBounds": { "sw": { "lat": 21.64, "lng": 113.47 }, "ne": { "lat": 22.918, "lng": 114.85 } }
+  },
+  {
+    "slug": "milan-italy",
+    "name": "Milan",
+    "coordinates": { "lat": 45.47, "lng": 9.18 },
+    "maxBounds": { "sw": { "lat": 44.929, "lng": 8.412 }, "ne": { "lat": 46.006, "lng": 9.948 } }
+  },
+  {
+    "slug": "sao-paulo-brazil",
+    "name": "São Paulo",
+    "coordinates": { "lat": -23.56, "lng": -46.64 },
+    "maxBounds": {
+      "sw": { "lat": -24.194, "lng": -47.333 },
+      "ne": { "lat": -22.924, "lng": -45.947 }
+    }
+  },
+  {
+    "slug": "tokyo-japan",
+    "name": "Tokyo",
+    "coordinates": { "lat": 35.69, "lng": 139.69 },
+    "maxBounds": { "sw": { "lat": 35.1, "lng": 138.966 }, "ne": { "lat": 36.276, "lng": 140.414 } }
+  },
+  {
+    "slug": "toronto-canada",
+    "name": "Toronto",
+    "coordinates": { "lat": 43.65, "lng": -79.38 },
+    "maxBounds": {
+      "sw": { "lat": 43.099, "lng": -80.138 },
+      "ne": { "lat": 44.196, "lng": -78.622 }
+    }
+  },
+  {
+    "slug": "mexico-city-mexico",
+    "name": "Mexico City",
+    "coordinates": { "lat": 19.43, "lng": -99.13 },
+    "maxBounds": {
+      "sw": { "lat": 18.782, "lng": -99.816 },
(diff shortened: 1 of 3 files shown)


data/cityDataSortedByDisplayPreference-expanded.json
anandaroop, line 1: This file is added directly to Eigen in this PR, but note that we currently have a build process that [grabs this from Metaphysics](https://github.com/artsy/eigen/blob/d18e36a9bb804e54be7927dcbfbe4f8f863900cd/scripts/city-guide/generate-cities-objc.js#L18).

I'm not exactly sure why — perhaps we anticipated this list would evolve more frequently, but it has been unchanged since City Guide's launch in early 2019 😅 

We should decide whether to continue with that pattern, or do away with that kinda superfluous build step and leave this static data file here in Eigen.

src/app/Scenes/City/CityPicker.tsx
```diff
@@ -21,8 +40,11 @@ export const CityPicker: React.FC<Props> = (props) => {
   const { size } = useScreenDimensions()
   const space = useSpace()
   const insets = useSafeAreaInsets()
+  const enabledExpandedList = useFeatureFlag("AREnableExpandedCityGuide")
+
+  const cities = enabledExpandedList ? expandedCities : originalCities
```
anandaroop, line 45: This is the only place we switch between the regular and expanded list, on order to create the expanded city picker menu.

But note that there are [several other mentions of this static data file](https://github.com/search?q=repo%3Aartsy%2Feigen%20cityDataSortedByDisplayPreference&type=code), and we haven't combed through these to see if they also need to change according to this flag.

data/cityDataSortedByDisplayPreference-expanded.json
```diff
@@ -0,0 +1,176 @@
...
+    }
+  },
+  {
+    "slug": "montreal-canada",
+    "name": "Montreal",
+    "coordinates": { "lat": 45.5, "lng": -73.57 },
+    "maxBounds": {
+      "sw": { "lat": 44.959, "lng": -74.338 },
+      "ne": { "lat": 46.036, "lng": -72.802 }
+    }
```
anandaroop, line 174: I could have sworn that these `maxBounds` parameters were in use before, and [governed the behavior of the map in Eigen](https://artsy.slack.com/archives/CP9P4KR35/p1611344096029700?thread_ts=1611344044.029300&cid=CP9P4KR35). 

But that does **not** appear to be the case now. (I'm really not sure what happened here, it's possible that it changed or that I mis-remembered the original feature.)

In any case I generated them for completeness' sake, based on [this notebook](https://observablehq.com/@anandaroop/max-bounds-for-all-featured-cities) which dates back to the original City Guide days.

src/app/Scenes/Map/GlobalMap.tsx
```diff
@@ -518,7 +532,7 @@ export const GlobalMap: React.FC<Props> = (props) => {
         >
           <MapboxGL.Camera
             ref={cameraRef}
-            animationMode="flyTo"
+            animationMode="moveTo"
```
anandaroop, line 544: This change is meant to replace the very melodramatic current animation with a simple instant transition.

The old animation however, due to being so lengthy, was actually masking the query latency.  Now we see a spinner instead, but we'll try to improve upon that too.

| Before | After |
|--------|--------|
| `animationMode="flyTo"` |`animationMode="moveTo"` | 
| <video src="https://github.com/user-attachments/assets/7632788c-a6f8-4ce6-bf27-c7ea075cf3aa" /> | <video src="https://github.com/user-attachments/assets/c73d950b-5ea0-4f62-9e7c-fd0b864cc971" /> |

data/cityDataSortedByDisplayPreference-expanded.json
```diff
@@ -77,7 +77,7 @@
   {
     "slug": "tokyo-japan",
     "name": "Tokyo",
-    "coordinates": { "lat": 35.69, "lng": 139.69 },
+    "coordinates": { "lat": 35.68, "lng": 139.76 },
```
anandaroop: I chose the updated values for Tokyo and Dubai manually by adapting this notebook from the old City Guide days https://observablehq.com/@anandaroop/city-guide-shows

MounirDhahri: Impressive notebook! thanks

MounirDhahri:
FYI, I looked further into the scripts we currently have but are not using to understand if they are worth keeping. I believe we don't need them anymore. The goal, I assume, was to preheat the relay cache so more cities are available for quicker switching as well as sending less payload to MP (using query id instead of the entire query payload). Although it's a nice quality of life improvement, I think it could mean that one might get cached values and we would need to maintain a script that is very hard to know about its existence (which is the case now). Which is why I think it's fair to remove it

data/cityDataSortedByDisplayPreference-expanded.json
anandaroop: Here's **Tokyo** before and after, much improved.

| Before | After |
|--------|--------|
| tokyo | tokyo2 | 

Dubai didn't change, I'd have to look into why, but it wasn't as bad as Tokyo. Imo it's not a blocker, but I will investigate.

anandaroop:
One final thought I have before opening this up for QA — now that we've got a bigger list of cities in the picker UI, it's no longer easy to digest in one glance.

Therefore I think the ordering of the list is important. Currently it's essentially in random order:

before


(Not _really_ random, it follows [this list](https://github.com/artsy/gravity/blob/e0c09bda1452d488f1f857d5e7a2f0d37c34f584/app/services/city_geocoding_service.rb#L3-L27) in Gravity, but I do not know why that list is ordered the way it is, and neither will any user.)

Since we have full control in Eigen thanks to that static file that lives in this repo, we can modify the ordering to make it more sensible.

There's more than one way to do this. I propose keeping the "original 6" cities at the top of the list, followed by the remainder in alphabetical order:

after

Not perfect, but an improvement over the current list I think, a compromise between keeping the original 6 major cities where they always were vs making the other cities easy to find by skimming.

@MounirDhahri if you agree, I will push that change up to this branch.

MounirDhahri:
@anandaroop I think that's a valid concern and I like the suggested solution. 

My only worry is that if you didn't see your city at first glance from the top 6 in the List (`London` for example), you might not find it after you scroll down to the letter `L`. I suggest maybe to also include top cities along with the full list of cities below and make the separation clear. Something like this
Screenshot 2025-09-24 at 07 48 05

MounirDhahri:
Another alternative would be a search bar but I like your suggestion more
Screenshot 2025-09-24 at 07 50 11

anandaroop:
I was going for the lazy approach that merely tweaked the data file, but yeah I'm open to any of that! Either now, or as a a follow-up. We could get design input as well. Maybe a hands-on QA session could surface ideas about the best way forward.

artsyit:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  Changes were made to package.json, but not to yarn.lock.<br/><i>Perhaps you need to run `yarn install`?</i>
  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 26005084905bfaa5bd54cff80306145a944f3091
</p>

anandaroop:
Updated to prefer strict alphabetical ordering of city list, per today's discussion:

Screenshot 2025-11-21 at 11 38 02 AM

src/app/Scenes/Map/GlobalMap.tsx
```diff
@@ -479,6 +491,8 @@ export const GlobalMap: React.FC<Props> = (props) => {
 
   const onSelectCity = (newCity: CityData) => {
     setShowCityPicker(false)
+    console.warn("setPreviouslySelectedCitySlug", newCity.slug)
```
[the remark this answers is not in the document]
MounirDhahri: we can maybe remove this

src/app/store/config/features.ts
```diff
@@ -224,6 +224,12 @@ export const features = {
     description: "Enable feature videos phase 2, falls back to webview if disabled",
     echoFlagKey: "AREnableFeatureVideoPhase2Type",
   },
+  AREnableExpandedCityGuide: {
+    readyForRelease: false,
```
MounirDhahri: we can already set this as ready for release since we merged this

MounirDhahri (dismissed):
Nice! LGTM

src/app/Scenes/Map/GlobalMap.tsx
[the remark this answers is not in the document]
anandaroop: This one is still there, right?

PR #12300, opened 2025-06-18
fix(ONYX-1566): use autoheigh bottom sheet component on the my collection artists screen

dariakoko:
This PR resolves [ONYX-1566] 

### Description

Make the My Collection artist details bottom sheet scale height automatically depending on the content - use AutoHeightBottomSheet component for this purpose

| Platform | Before | After |
|---|---|---|
| Android | <video src="https://github.com/user-attachments/assets/509c1d2a-5329-4082-9c43-6ec0817303dd" width="400" /> | <video src="https://github.com/user-attachments/assets/3890edc8-8e08-41a7-8c32-493a9dc44a69" width="400" /> |
| iOS | <video src="https://github.com/user-attachments/assets/aa826651-1138-4e07-b9d3-ab148c7e7e4f" width="400" /> | <video src="https://github.com/user-attachments/assets/abdca865-0637-4e28-b36c-4a2e890fe889" width="400" /> |

### PR Checklist

  - [x] **Android**.
  - [x] **iOS**.

### To the reviewers 👀

- use autoheigh bottom sheet component on the my collection artists screen

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[ONYX-1566]: https://artsyproduct.atlassian.net/browse/ONYX-1566?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/MyCollection/Components/MyCollectionBottomSheetModals/MyCollectionBottomSheetModalArtistPreview.tsx
+++ b/src/app/Scenes/MyCollection/Components/MyCollectionBottomSheetModals/MyCollectionBottomSheetModalArtistPreview.tsx
@@ -19,8 +19,8 @@ import { QueryRenderer, createFragmentContainer, graphql } from "react-relay"
 import useDebounce from "react-use/lib/useDebounce"
 
 interface MyCollectionBottomSheetModalArtistPreviewProps {
-  artist: MyCollectionBottomSheetModalArtistPreview_artist$data
-  me: MyCollectionBottomSheetModalArtistPreview_me$data
+  artist: MyCollectionBottomSheetModalArtistPreview_artist$data | null
+  me: MyCollectionBottomSheetModalArtistPreview_me$data | null
   interestId: string
 }
 
@@ -224,8 +224,8 @@ export const MyCollectionBottomSheetModalArtistPreviewQueryRenderer: React.FC<{
       render={({ props }) => {
         return (
           <MyCollectionBottomSheetModalArtistPreviewFragmentContainer
-            artist={props?.artist || null || undefined}
-            me={props?.me || null || undefined}
+            artist={props?.artist || null}
+            me={props?.me || null}
           />
         )
       }}
--- a/src/app/Scenes/MyCollection/Components/MyCollectionBottomSheetModals/MyCollectionBottomSheetModalArtistPreview.tsx
+++ b/src/app/Scenes/MyCollection/Components/MyCollectionBottomSheetModals/MyCollectionBottomSheetModalArtistPreview.tsx
@@ -5,7 +5,6 @@ import { MyCollectionBottomSheetModalArtistPreviewQuery } from "__generated__/My
 import { MyCollectionBottomSheetModalArtistPreview_artist$data } from "__generated__/MyCollectionBottomSheetModalArtistPreview_artist.graphql"
 import { MyCollectionBottomSheetModalArtistPreview_me$data } from "__generated__/MyCollectionBottomSheetModalArtistPreview_me.graphql"
 import { ArtistListItemContainer, ArtistListItemPlaceholder } from "app/Components/ArtistListItem"
-import { useArtworkListsBottomOffset } from "app/Components/ArtworkLists/useArtworkListsBottomOffset"
 import { AutoHeightBottomSheet } from "app/Components/BottomSheet/AutoHeightBottomSheet"
 import { useToast } from "app/Components/Toast/toastHook"
 import { ArtistKindPills } from "app/Scenes/MyCollection/Components/MyCollectionBottomSheetModals/MyCollectionBottomSheetModalArtistPreview/ArtistKindPills"
@@ -15,6 +14,7 @@ import { updateUserInterest } from "app/Scenes/MyCollection/mutations/updateUser
 import { getRelayEnvironment } from "app/system/relay/defaultEnvironment"
 import { refreshMyCollection } from "app/utils/refreshHelpers"
 import { useState } from "react"
+import { useSafeAreaInsets } from "react-native-safe-area-context"
 import { QueryRenderer, createFragmentContainer, graphql } from "react-relay"
 import useDebounce from "react-use/lib/useDebounce"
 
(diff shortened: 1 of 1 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (use autoheigh bottom sheet component on the my collection artists screen - dariakoko)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 6ecf270c401c55b3276698bffc929598d6261743
</p>

src/app/Scenes/MyCollection/Components/MyCollectionBottomSheetModals/MyCollectionBottomSheetModalArtistPreview.tsx
```diff
@@ -67,6 +57,26 @@ export const MyCollectionBottomSheetModalArtistPreview: React.FC<
...
   )
 
+  const dismissBottomView = () => {
+    setViewKind({ viewKind: null })
+  }
+
+  if (!artist || !me || !me.userInterest) {
+    return (
+      <AutoHeightBottomSheet
+        visible={!artist || !me || !me.userInterest}
```
olerichter00: Is this safeguard needed? Because we pass the same condition to `visible` 🤔
```suggestion
    return (
      <AutoHeightBottomSheet
        visible={!artist || !me?.userInterest}
```

src/app/Scenes/MyCollection/Components/MyCollectionBottomSheetModals/MyCollectionBottomSheetModalArtistPreview.tsx
```diff
@@ -210,28 +221,14 @@ export const MyCollectionBottomSheetModalArtistPreviewQueryRenderer: React.FC<{
...
-        },
-      })}
+      render={({ props }) => {
+        return (
+          <MyCollectionBottomSheetModalArtistPreviewFragmentContainer
+            artist={props?.artist || null}
+            me={props?.me || null}
+          />
+        )
+      }}
```
MounirDhahri, line 226: I would avoid return the fragment container directly within a `QueryRenderer` because then we would have to handle the placeholder and and error fallback within the component and it makes components bigger. 

If you want to refactor this, I would suggest to update this container to use `useLazyLoadQuery` along with `withSuspense`. The latter has helpers that help you avoid passing null data (`...$data | null`)

src/app/Scenes/MyCollection/Components/MyCollectionBottomSheetModals/MyCollectionBottomSheetModalArtistPreview.tsx
```diff
@@ -39,20 +29,20 @@ export const MyCollectionBottomSheetModalArtistPreview: React.FC<
...
-  const [isPrivate, setIsPrivate] = useState(me.userInterest?.private ?? false)
+  const [isPrivate, setIsPrivate] = useState(me?.userInterest?.private ?? false)
 
   const setViewKind = MyCollectionTabsStore.useStoreActions((actions) => actions.setViewKind)
 
   const toast = useToast()
-  const { bottom } = useSafeAreaInsets()
 
-  const space = useSpace()
+  const bottomOffset = useArtworkListsBottomOffset(2)
```
olerichter00: minor: I would recommend against using the artwork list-specific helper here to decouple the different contexts. `safeAreaInsets` or `space(2)` should do the same.

olerichter00 (approved):
Looks much better with the auto-height modal! 🌟

src/app/Scenes/MyCollection/Components/MyCollectionBottomSheetModals/MyCollectionBottomSheetModalArtistPreview.tsx
MounirDhahri: I was a bit confused initially because I didn't understand if this is 2 as a palette spacings["2"] or 2 pixels. Anyway, maybe here we can simplify this to make `2` as the default value since it's the only used value in Eigen currently
Screenshot 2025-06-18 at 10 14 10

MounirDhahri: We can probably also move this out from `Components/ArtworkLists` since it doesn't seem to have much to do with `ArtworkLists`. We can rename it to `useBottomSheetViewOffset` or something like that

olerichter00: Wouldn't it make sense to add the bottom offset to the modal component? It looks like it's always needed 🤔

MounirDhahri: @olerichter00 yes - I agree. think that's a better suggestion

dariakoko: Agree, will address it in a separate PR

src/app/Scenes/MyCollection/Components/MyCollectionBottomSheetModals/MyCollectionBottomSheetModalArtistPreview.tsx
dariakoko: I tried this approach in [this commit](https://github.com/artsy/eigen/pull/12300/commits/9a298cf73ab568d7a14cd1c728289579d775a1a2) 
It looks cleaner, but it introduces a bug unfortunately. With the way my collection bottom sheet compoenets are set up, the placeholder has to be a bottom sheet components as well that has to be dismissed separately and the hooks that dismiss the MyCollectionBottomSheetModalArtist bottom sheet are ignored win this setup

PR #10319, opened 2024-05-31
fix: crash on masonry dynamic columns empty state

gkartalis:
This PR resolves [APPL-933] 

### Description

Fixes a crash that was happening on masonry surfaces that we were changing the `numColumns` dynamically. This occurred only when the artworks array was empty.

#### Android

|Before|After|
|---|---|
|<video src="https://github.com/artsy/eigen/assets/21178754/f9d921dd-542a-432b-8715-0a3c193771fe" width="400" />|<video src="https://github.com/artsy/eigen/assets/21178754/bd4928bc-8e75-4695-b031-2635c3261c2f" width="400" />|

### iOS

|Before|After|
|---|---|
|<video src="https://github.com/artsy/eigen/assets/21178754/710fde47-a9ed-4763-880e-2210ef9a05c3" width="400" />|<video src="https://github.com/artsy/eigen/assets/21178754/a3b39d89-c756-4213-9c0e-d8b9d6e6694e" width="400" />|

### PR Checklist

- [x] I have tested my changes on **iOS** and **Android**.
- [x] I hid my changes behind a **[feature flag]**, or they don't need one.
- [x] I have included **screenshots** or **videos**, or I have not changed the UI.
- [x] I have added **tests**, or my changes don't require any.
- [x] I added an **[app state migration]**, or my changes do not require one.
- [x] I have documented any **follow-up work** that this PR will require, or it does not require any.
- [x] I have added a **changelog entry** below, or my changes do not require one.

### To the reviewers 👀

- [ ] I would like **at least one** of the reviewers to **run** this PR on the simulator or device.

<details><summary>Changelog updates</summary>

#### Cross-platform user-facing changes

- crash on masonry dynamic columns empty state - gkartalis

#### iOS user-facing changes

-

#### Android user-facing changes

-

#### Dev changes

-

</details>

Need help with something? Have a look at our [docs], or get in touch with us.

[app state migration]: ../blob/main/docs/adding_state_migrations.md
[feature flag]: ../blob/main/docs/developing_a_feature.md
[docs]: ../blob/main/docs/README.md

[APPL-933]: https://artsyproduct.atlassian.net/browse/APPL-933?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #12304, opened 2025-06-18
fix: collect screen breaks on empty state

MounirDhahri:
### Description

This PR fixes an issue affecting masonry artwork grids in App. When a user selects a set of filters with no results, they get an error instead of the empty state. This usually works fine in most cases, except in the case where one changes the number of columns. 

We had a first go at fixing this [here](https://github.com/artsy/eigen/pull/10319) which did the trick for a long time. But we didn't notice back then is that it wouldn't be enough in case where the artworks are dynamic, which is the case for the first time within the Collect screen.

**Suggested Solution**
Instead of returning a single `Flashlist` handling the case of artworks and no artworks, we would return two separate Flashlists sharing all props, except the dynamic `numColumns`.   

**How's that different from what we have now?**
By making `numColumns` static in the case where there are no artworks, the Flashlist won't complain if the `numColumn` prop changed, because it thinks it didn't change. This is fine in our case because the empty state is the same regardless of the view mode (masonry grid/list)

**When can we remove this logic?**
When the issue is fixed inside Flashlist
https://github.com/Shopify/flash-list/issues/1219

Tbh, I tried to solve this in a reasonable amount of time but wasn't able, I wanted to keep `numColumns` as a ref but it felt worse as a tech debt than this approach.

| Platform | Before | After |
|---|---|---|
| Android | <video src="https://github.com/user-attachments/assets/e0270dbb-845e-40e8-8bee-20339c012efd" > | <video src="https://github.com/user-attachments/assets/2f4f0568-3c32-4ada-8304-476e9158fba3" > |
| iOS | <video src="https://github.com/user-attachments/assets/e2aa61eb-6e66-4481-ae28-2ea4e9cc76c1" > | <video src="https://github.com/user-attachments/assets/891b011f-8ca5-49ed-b189-1ff1bff0ff7c" > |

### PR Checklist

  - [x] **Android**.
  - [x] **iOS**.

### To the reviewers 👀

- collect screen breaks on empty state - mounir

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Components/ArtworkGrids/MasonryInfiniteScrollArtworkGrid.tsx
+++ b/src/app/Components/ArtworkGrids/MasonryInfiniteScrollArtworkGrid.tsx
@@ -1,6 +1,6 @@
 import { ContextModule, ScreenOwnerType } from "@artsy/cohesion"
 import { useSpace } from "@artsy/palette-mobile"
-import { MasonryFlashList, MasonryFlashListProps } from "@shopify/flash-list"
+import { FlashListProps, MasonryFlashList, MasonryFlashListProps } from "@shopify/flash-list"
 import { PriceOfferMessage } from "app/Components/ArtworkGrids/ArtworkGridItem"
 import { MasonryArtworkGridItem } from "app/Components/ArtworkGrids/MasonryArtworkGridItem"
 import { PartnerOffer } from "app/Scenes/Activity/components/PartnerOfferCreatedNotification"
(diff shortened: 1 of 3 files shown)


MounirDhahri:
@araujobarret that’s also a nice solution, i wanted to still use the Flashlist because I wanted to keep the refresh control and some style from the main container. 

I am still not happy with my solution and I will try again to see if I can fix this in the library and open a PR there

ArtsyOpenSource:

- Cross-platform user-facing changes (collect screen breaks on empty state - mounir - MounirDhahri)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 687afca9ee1a58f9c6858662ca0a90f0fb92ff38
</p>

PR #12352, opened 2025-06-25
fix: broken navigation bar color

MounirDhahri:
WIP

### Description
This PR fixes the bottom navigation tabs issues on android

https://github.com/user-attachments/assets/55886252-95e9-4476-9d0f-49e2f8972432

### PR Checklist

  - [ ] **Android**.
  - [ ] **iOS**.

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/package.json
+++ b/package.json
@@ -139,6 +139,8 @@
     "easy-peasy": "6.0.5",
     "events": "3.3.0",
     "expo": "^52.0.0",
+    "expo-navigation-bar": "~4.0.9",
+    "expo-status-bar": "~2.0.1",
     "expo-updates": "~0.27.4",
     "formik": "2.2.9",
     "html-entities": "2.3.3",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -139,6 +139,8 @@
     "easy-peasy": "6.0.5",
     "events": "3.3.0",
     "expo": "^52.0.0",
+    "expo-navigation-bar": "~4.0.9",
+    "expo-status-bar": "~2.0.1",
```
gkartalis, line 143: **blocking:** I would say let's avoid adding expo dependencies especially for things that can be done with the react native api

MounirDhahri: Absolutely, this was just a spike to see if this would be working better than React Native `StatusBar` and own our `ArtsyNativeModule` code. Unfortunately it failed the test, and still, both approaches don't work.

gkartalis: I looked into this issue a while back, its complicated but not impossible to address ~ I think that our issue is that we are overriding the status bar from the native side of things but also on the JS side of things multiple times and with many ways that conflict with each other and makes it difficult to address and test, happy to pair on it one day

MounirDhahri: Yes, along with some stuff over at `styles.xml`, it became hard to debug

PR #12356, opened 2025-06-25
RFC: align on folder structure best practices

MounirDhahri:
Co-authored-by: @MrSltun 

### Description

This RFC comes as an action item on https://artsy.slack.com/archives/C02BAQ5K7/p1749213588102729

Our current docs are kinda outdated and we started to have multiple patterns of organising folders in the app. So the goal from this RFC is to align again on best practices. IMO, there is not best way to organise folders, but there are bad ways, and inconsistency is one of them. 

### PR Checklist

  - [x] **Android**.
  - [x] **iOS**.

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- chore update folder structure best practices - mounir

</details>

### DIFF
--- a/docs/best_practices.md
+++ b/docs/best_practices.md
@@ -91,23 +91,17 @@ For example `mutations`, `hooks` and `utils` would be **camelCase** folders, whi
 │ │ ├── MyScreenComponentA.tsx
 │ │ ├── MyScreenComponentB.tsx
 
-│ ├── mutations
-│ │ ├── __tests__
-│ │ │ ├── mutationFunction.tests.ts
-│ │ │ ├── useMyMutation.tests.ts
-│ │ ├── mutationFunction.ts
-│ │ ├── useMyMutation.ts  👈 hook mutations still go here
-
 │ ├── hooks
 │ │ ├── __tests__
 │ │ │ ├── useMyHook.tests.ts
+│ │ │ ├── useMyMutation.tests.ts
 │ │ ├── useMyHook.ts
-
-│ ├──├── utilFunction.tests.ts
-│ │ ├── utilFunction.ts
+│ │ ├── useMyMutation.ts  👈 hook mutations still go here
 
 ├── … utils
 │ │ ├── __tests__
+│ ├─│ ├── utilFunction.tests.ts
+│ │ ├── utilFunction.ts
 │ │ │
 
 ```
--- a/docs/best_practices.md
+++ b/docs/best_practices.md
@@ -11,11 +11,12 @@ _Please note: Links should point to specific commits, and not a branch (in case 
 
 - [:world_map: Our Best Practices](#world_map-our-best-practices)
 - [Contents](#contents)
-
   - [Examples \& Hacks](#examples--hacks)
   - [History](#history)
     - [File Structure Organization](#file-structure-organization)
-      - [index.ts files](#indexts-files)
+    - [Example when creating a new screen](#example-when-creating-a-new-screen)
+    - [Example when adding a component(s) to src/app/components](#example-when-adding-a-components-to-srcappcomponents)
+      - [AVOID index.ts(x) files](#avoid-indextsx-files)
     - [When committing code](#when-committing-code)
   - [Frontend](#frontend)
     - [Styling](#styling)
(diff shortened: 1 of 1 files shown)


ArtsyOpenSource:

- Dev changes (chore update folder structure best practices - mounir - MounirDhahri)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 1802e7a6d89f952c08d705f92489b33c811880d5
</p>

MounirDhahri:
@MrSltun I updated the PR's title to be RFC instead of chore to get some feedback from the rest of the team. Also to make sure that this is the way to go from now on

docs/best_practices.md
```diff
@@ -61,64 +58,86 @@ The Artsy app was initially written in Objective-C and Swift and React Native wa
...
+
+│ ├── hooks
+│ │ ├── __tests__
+│ │ │ ├── useMyHook.tests.ts
+│ │ ├── useMyHook.ts
+
+│ ├──├── utilFunction.tests.ts
+│ │ ├── utilFunction.ts
+
+├── … utils
```
iskounen, line 102: I'm curious about the difference between what should go in a `utilFunction.ts` v.s. a `utils` folder. Do you have any examples you could share?

MounirDhahri: this must have been a typo, `utilFuntions` go inside `utils`

docs/best_practices.md
```diff
@@ -61,64 +58,86 @@ The Artsy app was initially written in Objective-C and Swift and React Native wa
...
-│   ├── buttons.stories.tsx
-├── …
+
+├── MyScreen
+
+│ ├── __tests__
+│ │ ├── MyScreen.tests.tsx
+
+│ ├── MyScreen.tsx
+│ ├── MyScreenStoreModel.tsx
```
araujobarret, line 86: I think these stores should be in a folder `/stores`, same in case you have context providers, some repos also store contexts/redux stores all under `/stores`.

docs/best_practices.md
araujobarret: imo utils could be anything, constants, helper functions that transform data, types, it's a flexible folder

docs/best_practices.md
```diff
@@ -61,64 +58,86 @@ The Artsy app was initially written in Objective-C and Swift and React Native wa
...
+│ │ | ├── MyScreenComponentB.tests.tsx
+│ │ ├── MyScreenComponentA.tsx
+│ │ ├── MyScreenComponentB.tsx
+
+│ ├── mutations
+│ │ ├── __tests__
+│ │ │ ├── mutationFunction.tests.ts
+│ │ │ ├── useMyMutation.tests.ts
+│ │ ├── mutationFunction.ts
+│ │ ├── useMyMutation.ts  👈 hook mutations still go here
```
araujobarret: I'm not sure if I agree much here, if it's a hook should be in the hooks folder, look at other repos as https://github.com/artsy/volt/tree/main/app/javascript/apps/bulk-update/hooks or https://github.com/artsy/force/tree/main/src/Apps/ArtQuiz/Hooks, I think we should get rid of `/mutations`

dariakoko (commented):
Thanks for taking your time to look into this! 

I just have one comment - maybe we should add the scenario when we have a component/hook/mutation within a Screen and there is a need to use it elsewhere? I think this is one of the most common scenarios that leads to the inconsistent files structure on Eigen

docs/best_practices.md
MounirDhahri: yes, that's the goal actually. Do you think we should be more specific?

MounirDhahri: IMO, this is kind of similar to semantic commits, if we would be more specific with `style:`, `refactor:`, `perf:`, `test:`... etc, it starts to take effort to guess the folder name.

docs/best_practices.md
MounirDhahri: So far, in almost all cases, our store models, even in the case of the ones used in Globalstore, has been a single file. I don't feel strong about any approach and I am fine with moving this to a folder in case we all agree about it. That's what matters the most to me.

docs/best_practices.md
MounirDhahri: that's a good idea, let's follow the patterns in `Force` and `Volt` then

MounirDhahri:
> I just have one comment - maybe we should add the scenario when we have a component/hook/mutation within a Screen and there is a need to use it elsewhere? I think this is one of the most common scenarios that leads to the inconsistent files structure on Eigen

That's a good point, In that case, the shared component/hook needs to be extracted from the original component and should be moved to `src/app/Components` or `src/app/hooks` etc..

I will add a note about it

docs/best_practices.md
araujobarret: nah, generic as utils should do the job, I agree with it as a generic folder, Rails apps also use the same idea for more general stuff 👍

docs/best_practices.md
araujobarret: Sorry, I meant the local easy-peasy context or react context providers, not the global ones. We could use `contexts` as we have in Volt, Force.

araujobarret: But I consider it a minor one, we don't have many context/stores defined in local level of elements, so def not a blocker for this RFC

gkartalis (commented):
thanks for pushing this forwards! no strong opinions from my side

docs/best_practices.md
MounirDhahri: awesome

brainbicycle (approved):
This seems reasonable to me! Thank you!

dariakoko (approved):

MrSltun:
Merging this since we all agree on the new structure, we will mention it in the upcoming Engineering Stand up

PR #12358, opened 2025-06-25
fix: pass the correct query to useSearchQuery refetch

iskounen:
This PR attempts to fix an issue that Claude Code noticed while I was debugging the Explore by Category component.

The refetch returned by that hook uses the `SearchScreenQuery` query, even when the `GlobalSearchOverlay` is using it, instead of using the same query that is passed to the hook (which is named `globalSearchInputOverlayQuery` in the case of `GlobalSearchOverlay`). 

### PR Checklist

  - [ ] **Android**.
  - [x] **iOS**.

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- Pass the correct query to useSearchQuery refetch

</details>

### DIFF
--- a/src/app/Scenes/Search/useSearchQuery.tsx
+++ b/src/app/Scenes/Search/useSearchQuery.tsx
@@ -1,4 +1,3 @@
-import { SearchScreenQuery } from "app/Scenes/Search/Search"
 import { useState } from "react"
 import { GraphQLTaggedNode, useLazyLoadQuery, useRelayEnvironment } from "react-relay"
 import { FetchPolicy, fetchQuery, OperationType, VariablesOf } from "relay-runtime"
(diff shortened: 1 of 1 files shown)


ArtsyOpenSource:

- Dev changes (Pass the correct query to useSearchQuery refetch - iskounen)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 7bf596e70fdb1321c5b6ac934b3c057a832b3c07
</p>

MounirDhahri (approved):
Good find. How did you use Claude to identify this?

iskounen:
@MounirDhahri I was using Claude to figure out why this error was being thrown on the Search page after I added the _Explore by Category_ widget to it:

```
[1]  ERROR  Warning: Relay: Unexpected fetch while using a null fragment ref for fragment `HomeViewSectionsConnection_viewer` in `usePaginationFragment()`. When fetching more items, we expect initial fragment data to be non-null. Please make sure you're passing a valid fragment ref to `usePaginationFragment()` before paginating.
```

One idea that Claude was pursuing was that this was caused by the Home tab (via the GlobalSearchOverlay) and the Search tab sharing the same `viewer` query, and therefore corrupting the Relay cache.

That didn't turn out to be the cause of that error, and luckily the two `viewer` queries are identical so it's not a problem for the search functionality, but I thought this would be some good cleanup.

MounirDhahri:
Awesome! thanks @iskounen

PR #12364, opened 2025-06-26
feat(DIA-1363): add explore by category

iskounen:
This PR adds _Explore by Category_ to the Search tab.

It differs a bit from the original implementation on the Home tab in that it hard-codes the categories instead of fetching them from Metaphysics.

https://github.com/user-attachments/assets/960191e1-6377-4ccd-b15c-6eace0844237

### PR Checklist

  - [x] **Android**.
  - [x] **iOS**.

### To the reviewers 👀

- Permanently added Expore by Category to the Search tab

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Scenes/CollectionsByCategory/Body.tsx
+++ b/src/app/Scenes/CollectionsByCategory/Body.tsx
@@ -12,6 +12,7 @@ import {
   CollectionsChips,
   CollectionsChipsPlaceholder,
 } from "app/Scenes/CollectionsByCategory/CollectionsChips"
+import { getTitleForCategory } from "app/Scenes/Search/components/ExploreByCategory/constants"
 import { NoFallback, withSuspense } from "app/utils/hooks/withSuspense"
 import { graphql, useFragment, useLazyLoadQuery } from "react-relay"
 
@@ -22,7 +23,8 @@ interface BodyProps {
 export const Body: React.FC<BodyProps> = ({ viewer }) => {
   const data = useFragment(fragment, viewer)
   const { params } = useRoute<CollectionsByCategoriesRouteProp>()
-  const category = params.category
+  const category = decodeURI(params.category)
+  const title = getTitleForCategory(category)
 
   if (!data?.marketingCollections) {
     return null
@@ -32,9 +34,9 @@ export const Body: React.FC<BodyProps> = ({ viewer }) => {
     <Flex gap={4}>
       <Flex gap={2}>
         <Text variant="xl" px={2}>
-          {category}
+          {title}
         </Text>
-        <Text px={2}>Explore collections by {category.toLowerCase()}</Text>
+        <Text px={2}>Explore collections by {title.toLowerCase()}</Text>
         {/* TODO: fix typings broken by some unknown reason here, prob related to @plural */}
         <CollectionsChips marketingCollections={data.marketingCollections as any} />
       </Flex>
@@ -101,8 +103,9 @@ export const collectionsByCategoryQuery = graphql`
 export const BodyWithSuspense = withSuspense({
   Component: () => {
     const { params } = useRoute<CollectionsByCategoriesRouteProp>()
+    const category = decodeURI(params.category)
     const data = useLazyLoadQuery<BodyCollectionsByCategoryQuery>(collectionsByCategoryQuery, {
-      category: params.entityID,
+      category,
     })
 
     if (!data.viewer) {
--- a/src/app/Scenes/CollectionsByCategory/CollectionsByCategory.tsx
+++ b/src/app/Scenes/CollectionsByCategory/CollectionsByCategory.tsx
@@ -2,29 +2,33 @@ import { OwnerType } from "@artsy/cohesion"
 import { Flex, Screen } from "@artsy/palette-mobile"
 import { RouteProp, useRoute } from "@react-navigation/native"
 import { BodyWithSuspense } from "app/Scenes/CollectionsByCategory/Body"
-import { FooterWithSuspense } from "app/Scenes/CollectionsByCategory/Footer"
+import { Footer } from "app/Scenes/CollectionsByCategory/Footer"
+import { getTitleForCategory } from "app/Scenes/Search/components/ExploreByCategory/constants"
 import { goBack } from "app/system/navigation/navigate"
 import { ProvideScreenTrackingWithCohesionSchema } from "app/utils/track"
 import { screen } from "app/utils/track/helpers"
 import { FC } from "react"
-
-type CollectionsByCategoriesNavigationRoutes = {
-  collections: {
-    category: string
-    entityID: string
-    homeViewSectionId: string
-  }
-}
+import { Platform } from "react-native"
+import { useSafeAreaInsets } from "react-native-safe-area-context"
 
 export type CollectionsByCategoriesRouteProp = RouteProp<
-  CollectionsByCategoriesNavigationRoutes,
+  {
+    collections: {
+      category: string
+    }
+  },
   "collections"
 >
 
 export const CollectionsByCategory: FC = () => {
   const { params } = useRoute<CollectionsByCategoriesRouteProp>()
+  const category = decodeURI(params.category)
+  const insets = useSafeAreaInsets()
 
-  const category = params.category
+  const title = getTitleForCategory(category)
+
+  const bottomPadding =
+    Platform.OS === "ios" ? Math.max(insets.bottom + 60, 80) : Math.max(insets.bottom + 40, 60)
 
   return (
     <ProvideScreenTrackingWithCohesionSchema
@@ -34,13 +38,13 @@ export const CollectionsByCategory: FC = () => {
       })}
     >
       <Screen>
-        <Screen.Header onBack={goBack} title={category} animated />
+        <Screen.Header onBack={goBack} title={title} animated />
         <Screen.Body fullwidth flex={1}>
-          <Screen.ScrollView>
+          <Screen.ScrollView contentContainerStyle={{ paddingBottom: bottomPadding }}>
             <Flex gap={4}>
               <BodyWithSuspense />
 
-              <FooterWithSuspense homeViewSectionId={params.homeViewSectionId} />
+              <Footer />
             </Flex>
           </Screen.ScrollView>
         </Screen.Body>
--- a/src/app/Scenes/Search/Search.tsx
+++ b/src/app/Scenes/Search/Search.tsx
@@ -8,13 +8,15 @@ import { SearchQuery, SearchQuery$variables } from "__generated__/SearchQuery.gr
 import { GlobalSearchInput } from "app/Components/GlobalSearchInput/GlobalSearchInput"
 import { SearchPills } from "app/Scenes/Search/SearchPills"
 import { DiscoverSomethingNew } from "app/Scenes/Search/components/DiscoverSomethingNew/DiscoverSomethingNew"
+import { ExploreByCategory } from "app/Scenes/Search/components/ExploreByCategory/ExploreByCategory"
 import { useRefetchWhenQueryChanged } from "app/Scenes/Search/useRefetchWhenQueryChanged"
 import { useSearchQuery } from "app/Scenes/Search/useSearchQuery"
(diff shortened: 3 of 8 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (Permanently added Expore by Category to the Search tab - iskounen)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 26b6a4d83643a9b1b1f8a64d6eda336bb4c647db
</p>

src/app/Scenes/CollectionsByCategory/CollectionsByCategory.tsx
```diff
@@ -34,13 +38,13 @@ export const CollectionsByCategory: FC = () => {
       })}
     >
       <Screen>
-        <Screen.Header onBack={goBack} title={category} animated />
+        <Screen.Header onBack={goBack} title={title} animated />
         <Screen.Body fullwidth flex={1}>
-          <Screen.ScrollView>
+          <Screen.ScrollView contentContainerStyle={{ paddingBottom: bottomPadding }}>
```
MounirDhahri: btw, we have a contastant that we usually use for such case called `SCROLLVIEW_PADDING_BOTTOM_OFFSET`. It's less smart that what you implemented but it's more straighforward and aligned with other surfaces

src/app/Scenes/CollectionsByCategory/Body.tsx
MounirDhahri, line 1: I know this is existing code, but do you mind at some point to rename this to be something more easy in IDEs
as per https://github.com/artsy/eigen/pull/12356

src/app/Scenes/Search/components/ExploreByCategory/ExploreByCategory.tsx
MounirDhahri, line 1: same comment here about naming, although it's a long name, I would still suggest `SearchExploreByCategory`

MounirDhahri (dismissed):
just some minor comments, but looks good to me overall

src/app/Scenes/CollectionsByCategory/Body.tsx
araujobarret: I think it might be a good time for a small folder reorg refactor too 👍

src/app/Scenes/Search/components/ExploreByCategory/constants.ts
```diff
@@ -0,0 +1,42 @@
+export interface MarketingCollectionCategory {
+  category: string
+  title: string
+  imageUrl: string
+}
+
+export const MARKETING_COLLECTION_CATEGORIES: MarketingCollectionCategory[] = [
```
araujobarret: Since we're moving away from getting this info from the homeView it should come from somewhere else in MP, I don't think we should hard code it here to be honest 🤔

araujobarret: Also another opportunity to convert these images from jpeg to webp, just noticed we're still getting them as jpeg which are way bigger than webp

iskounen: My thought process was that we won't change these 6 categories nearly as often as we may change the "Discover Something New" categories. Most of our UI is hard-coded in Eigen (as opposed to being configurable like the home feed) so I don't think it's unusual.

However! I am having a hard time getting the CollectionsByCategory component to correctly resize now that its `Footer` isn't dynamically loaded, so I am open to fetching this from Metaphysics again.

src/app/Scenes/CollectionsByCategory/CollectionsByCategory.tsx
iskounen: I tried to use that, but it wasn't nearly enough space, so the `Footer` was still being cut short. I think the cause of this is that the `Body` is loaded with suspense but the `Footer` is not anymore ... which makes the `ScrollView` struggle to resize itself.

src/app/Scenes/CollectionsByCategory/CollectionsByCategoryBody.tsx
```diff
@@ -90,27 +92,28 @@ const BodyPlaceholder: React.FC = () => {
...
-export const collectionsByCategoryQuery = graphql`
-  query BodyCollectionsByCategoryQuery($category: String!) {
-    viewer {
-      ...BodyCollectionsByCategory_viewer @arguments(category: $category)
-    }
-  }
-`
-
-export const BodyWithSuspense = withSuspense({
+export const CollectionsByCategoryBodyWithSuspense = withSuspense({
```
araujobarret, line 103: super-minor: by default we've been putting the queries here before these "query renderers" components.

src/app/Scenes/CollectionsByCategory/CollectionsByCategoryFooter.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+const query = graphql`
+  query CollectionsByCategoryFooterQuery {
+    categories: discoveryCategoriesConnection {
+      ...CollectionsByCategoryFooter_category
+    }
+  }
+`
+
+const fragment = graphql`
```
araujobarret: ditto here about the order

src/app/Scenes/Search/Search.tsx
```diff
@@ -50,6 +53,9 @@ export const Search: React.FC = () => {
 
   const shouldShowCityGuide = !isTablet()
 
+  const bottomPadding =
```
araujobarret: shouldn't we use the `BOTTOM_TABS_HEIGHT` + space(1 or 2) here?

src/app/Scenes/Search/components/ExploreByCategory/ExploreByCategoryCard.tsx
```diff
@@ -0,0 +1,81 @@
...
+    }
+  }
+
+  return (
+    /**
+     * A prefetchVariable is passed to RouterLink so that Relay doesn't confuse the viewer used by
+     * CollectionsByCategory with other viewers it has in the cache. By providing the category ID as
+     * a variable, its viewer will be stored in its own cache key.
+     */
+    <RouterLink to={href} onPress={handleCardPress} prefetchVariables={{ category: card.category }}>
```
araujobarret: Have you checked if using prefetchVariables instead of appending the category as a URL parameter on line 29 works using this prop? Maybe the URL param will encode and will break the MP query because we don't expect encoded params with `%20` for example 🤔

MounirDhahri:
Please let me know when I can review this

src/app/Scenes/Search/components/ExploreByCategory/ExploreByCategory.tsx
```diff
@@ -0,0 +1,30 @@
...
+      const data = useLazyLoadQuery<ExploreByCategoryQuery>(query, {})
+
+      if (!data?.categories) {
+        return null
+      }
+
+      return <ExploreByCategoryCards categories={data.categories} />
+    },
+    LoadingFallback: ExploreByCategoryCardsPlaceholder,
+    ErrorFallback: NoFallback,
```
MounirDhahri, line 20: I would suggest to have minimal fallback here in case the query failed. Something like

```typescript
  ErrorFallback: (fallbackProps) => {
    return (
      <LoadFailureView
        onRetry={fallbackProps.resetErrorBoundary}
        useSafeArea={false}
        showCloseButton
        error={fallbackProps.error}
        showBackButton={true}
        trackErrorBoundary={false}
      />
    )
  },
```

src/app/Scenes/Search/components/ExploreByCategory/ExploreByCategory.tsx
```diff
@@ -0,0 +1,30 @@
...
+  })
+)
+
+const query = graphql`
+  query ExploreByCategoryQuery {
+    categories: discoveryCategoriesConnection {
```
MounirDhahri, line 26: **no-blocker:** I am not used to seeing connections without arguments

src/app/Scenes/Search/components/ExploreByCategory/ExploreByCategoryCard.tsx
MounirDhahri: It might not work because in `routes.tsx` we have the route defined as `/collections-by-category/:category`. 

On the other side, you have a good point about encoding, I wish we had better ids for categories to avoid such cases. Things are in MP now so I don't see why we can't support that already,

src/app/Scenes/Search/components/ExploreByCategory/ExploreByCategory.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+      return <ExploreByCategoryCards categories={data.categories} />
+    },
+    LoadingFallback: ExploreByCategoryCardsPlaceholder,
+    ErrorFallback: NoFallback,
+  })
+)
+
+const query = graphql`
+  query ExploreByCategoryQuery {
```
MounirDhahri: This query contains no user related fields or timely fields and can be cached
```suggestion
  query ExploreByCategoryQuery @cacheable {
```

src/app/Scenes/Search/components/ExploreByCategory/__tests__/ExploreByCategoryCard.tests.tsx
```diff
@@ -0,0 +1,172 @@
+import { fireEvent, screen } from "@testing-library/react-native"
+import { ExploreByCategoryCardTestQuery } from "__generated__/ExploreByCategoryCardTestQuery.graphql"
+import { ExploreByCategoryCard } from "app/Scenes/Search/components/ExploreByCategory/ExploreByCategoryCard"
+import { mockTrackEvent } from "app/utils/tests/globallyMockedStuff"
+import { setupTestWrapper } from "app/utils/tests/setupTestWrapper"
+import { graphql } from "react-relay"
+
+const mockRouterLinkProps = { to: "", prefetchVariables: {} }
+
+jest.mock("app/system/navigation/RouterLink", () => ({
```
MounirDhahri, line 10: **Q:** is there a specific reason why you mocked `RouterLink`

src/app/Scenes/Search/components/ExploreByCategory/ExploreByCategoryCard.tsx
```diff
@@ -0,0 +1,90 @@
...
+          <Text variant="md">{card.title}</Text>
+        </Flex>
+      </Flex>
+    </RouterLink>
+  )
+}
+
+const fragment = graphql`
+  fragment ExploreByCategoryCard_category on DiscoveryCategory {
+    category
```
MounirDhahri, line 70: imagine if we had an `internalID` here 
It would mean we will no longer need to do stuff like:
```
 const category = decodeURI(params.category)
```

MounirDhahri: If we would add it here 
https://github.com/artsy/metaphysics/blob/56fbd37ac6056a32a9b67971271b7c3285b51f3a/src/schema/v2/discoveryCategoriesConnection.ts#L35

and use it in the resolver, it would solve all our fetching/prefetching issues and if we would port the same logic to force, we wouldn't need to copy the `decodeURI` logic we do in multiple cases here

src/app/Scenes/Search/components/ExploreByCategory/ExploreByCategory.tsx
iskounen: I'm interested in using a LoadFailureView (especially for sending errors to Sentry), but I'm not sure what the props all mean. I'm happy to follow-up with this change after discussing it with you.

src/app/Scenes/Search/components/ExploreByCategory/__tests__/ExploreByCategoryCard.tests.tsx
iskounen: This allows me to assert that proper values are being passed to the RouterLink's props because I was very confused by the different values of `category` that were being passed from ExploreByCategory to CollectionsByCategory components.

PR #12373, opened 2025-06-27
chore: improve performance of artist screen

MounirDhahri:
### Description

This PR brings multiple improvements to artist screen in order to improve its performance

- Separate Artworks Query From the above the fold query
- Do not query for aggregations at each page
- Do not query for artworks before the artist cover image and important metadata
- Prefetch artworks query
- Refactor the component into a functional component
- Better skeleton - especially for when an artist is featured
- Fade into withSuspense content: this looks sooo good - it fixed the flash between content and placeholder

According to Reactotron, I see a drop of 30% in the query time, I do however think the drop will be bigger. I will report back after we get this in prod and see things in Datadog

**More Context:**
https://artsy.slack.com/archives/C05EQL4R5N0/p1751035602351999

| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/c43126b0-f6a9-4e0c-bd9f-528769fd63b7" />  | <video src="https://github.com/user-attachments/assets/f47a5c89-e0ec-4586-90c4-f1aa73008fe0" /> | 

### PR Checklist

  - [x] **Android**.
  - [x] **iOS**.

### To the reviewers 👀

- improve the performance of the artist screen - mounir

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
+++ b/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
@@ -465,10 +465,10 @@ const ArtistArtworksPlaceholder = () => {
       scrollEnabled={false}
       contentContainerStyle={{ flex: 1, paddingHorizontal: 0, paddingTop: space(2) }}
     >
-      <Flex px={2}>
+      <Flex px={2} testID="ArtistArtworksPlaceholder">
         <Flex flexDirection="row" justifyContent="space-between">
           <SkeletonText>Create Alert</SkeletonText>
-          <SkeletonText>Sory & Filter</SkeletonText>
+          <SkeletonText>Sort & Filter</SkeletonText>
         </Flex>
 
         <Spacer y={2} />
--- a/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
+++ b/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
@@ -5,6 +5,7 @@ import {
   Button,
   Flex,
   Message,
+  SkeletonText,
   Spacer,
   Tabs,
   Text,
@@ -457,12 +458,31 @@ export const ArtistArtworksQueryRenderer = withSuspense<ArtistArtworksQueryRende
 })
 
 const ArtistArtworksPlaceholder = () => {
+  const space = useSpace()
+
   return (
     <Tabs.ScrollView
       scrollEnabled={false}
-      contentContainerStyle={{ flex: 1, paddingHorizontal: 0, paddingTop: 20 }}
+      contentContainerStyle={{ flex: 1, paddingHorizontal: 0, paddingTop: space(2) }}
     >
-      <PlaceholderGrid />
+      <Flex px={2}>
+        <Flex flexDirection="row" justifyContent="space-between">
+          <SkeletonText>Create Alert</SkeletonText>
+          <SkeletonText>Sory & Filter</SkeletonText>
+        </Flex>
+
+        <Spacer y={2} />
+
+        <Flex borderBottomWidth={1} borderBottomColor="mono100" mx={-2} />
+
+        <Spacer y={2} />
+
+        <SkeletonText variant="xs">XX Artworks</SkeletonText>
+
+        <Spacer y={2} />
+
+        <PlaceholderGrid mx={0} />
+      </Flex>
     </Tabs.ScrollView>
   )
 }
--- a/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
+++ b/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
@@ -13,7 +13,13 @@ import {
   useSpace,
 } from "@artsy/palette-mobile"
 import { MasonryFlashListRef } from "@shopify/flash-list"
-import { ArtistArtworks_artist$data } from "__generated__/ArtistArtworks_artist.graphql"
+import {
+  ArtistArtworksQuery,
+  ArtistArtworksQuery$data,
+  FilterArtworksInput,
+} from "__generated__/ArtistArtworksQuery.graphql"
+import { ArtistArtworks_artist$key } from "__generated__/ArtistArtworks_artist.graphql"
+import { ArtistArtworks_artistAggregation$key } from "__generated__/ArtistArtworks_artistAggregation.graphql"
 import { ArtistArtworksFilterHeader } from "app/Components/Artist/ArtistArtworks/ArtistArtworksFilterHeader"
 import { useCreateSavedSearchModalFilters } from "app/Components/Artist/ArtistArtworks/hooks/useCreateSavedSearchModalFilters"
 import { useShowArtworksFilterModal } from "app/Components/Artist/ArtistArtworks/hooks/useShowArtworksFilterModal"
@@ -27,6 +33,7 @@ import { useArtworkFilters } from "app/Components/ArtworkFilter/useArtworkFilter
 import ArtworkGridItem from "app/Components/ArtworkGrids/ArtworkGridItem"
 import { FilteredArtworkGridZeroState } from "app/Components/ArtworkGrids/FilteredArtworkGridZeroState"
 import { Props as InfiniteScrollGridProps } from "app/Components/ArtworkGrids/InfiniteScrollArtworksGrid"
+import { LoadFailureView, LoadFailureViewProps } from "app/Components/LoadFailureView"
 import { ProgressiveOnboardingAlertReminder } from "app/Components/ProgressiveOnboarding/ProgressiveOnboardingAlertReminder"
 import {
   CREATE_ALERT_REMINDER_ARTWORK_THRESHOLD,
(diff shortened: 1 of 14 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (improve the performance of the artist screen - mounir - MounirDhahri)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against a2de40c507d6cfad52ee3350bb74525c11ea267b
</p>

src/app/Components/FadeIn.tsx
```diff
@@ -1,30 +1,36 @@
...
         {
           transform: [
             {
-              translateY: slide
-                ? showing.interpolate({
-                    inputRange: [0, 1],
-                    outputRange: [10, 0],
-                  })
-                : 0,
+              translateY: slide ? interpolate(showing.value, [0, 1], [10, 0]) : 0,
```
araujobarret, line 33: I wonder if this will trigger that annoying message from reanimated:
```
[Reanimated] Reading from `value` during component render. Please ensure that you do not access the `value` property or use `get` method of a shared value while React is rendering a component.
```
Do you know if it's popping up?

src/app/Components/SkeletonPill/SkeletonPill.tsx
```diff
@@ -0,0 +1,14 @@
+import { SkeletonBox } from "@artsy/palette-mobile"
+import { PixelRatio } from "react-native"
+
+export const SkeletonPill = () => {
```
araujobarret, line 4: I wouldn't put it here, I'd rather use it as you defined, without creating an extra Skeleton element. If we want to reuse this across different areas, we should place it in the palette-mobile, as with the rest of the skeleton elements.

src/app/Components/ArtistListItemShort.tsx
```diff
@@ -17,6 +17,11 @@ interface ArtistListItemShortProps {
 export const ArtistListItemShort: FC<ArtistListItemShortProps> = ({ artist, onPress }) => {
   const data = useFragment(fragment, artist)
 
+  const followButton = useMemo(
+    () => <ArtistFollowButtonQueryRenderer artistID={data?.internalID ?? ""} />,
+    [data?.internalID]
+  )
```
olerichter00, line 23: Does this have a measurable effect? If it doesn't, I'd suggest avoiding useMemo if possible to prevent bugs, e.g., by not updating the dependencies array correctly.

src/app/Navigation/routes.tsx
```diff
@@ -450,7 +451,7 @@ export const artsyDotNetRoutes = defineRoutes([
         headerShown: false,
       },
     },
-    queries: [ArtistScreenQuery],
+    queries: [ArtistScreenQuery, artistArtworksQuery],
```
olerichter00, line 454: 💯

src/app/Scenes/Artist/Artist.tsx
```diff
@@ -348,15 +357,34 @@ const ArtistSkeleton: React.FC = () => {
       <Screen.Body fullwidth>
         <Skeleton>
           <SkeletonBox width={width} height={height} />
+
           <Spacer y={2} />
+
           <Flex px={2}>
-            <SkeletonText variant="lg">Artist Name Artist Name</SkeletonText>
+            <SkeletonText variant="lg-display">Artist Name</SkeletonText>
             <Spacer y={0.5} />
```
olerichter00, line 365: Do we still need the spacer after the change to `lg-display`?

MounirDhahri: I actually tried removing it but it was neeeded. The two `SkeletonText` blocks were showing sticking to each other and it didn't look nice

src/app/Components/ArtistListItemShort.tsx
MounirDhahri: yes, it does have a measureable effect actually. Because this wasn't memoized, each state change here was leading to the FollowButton to be computed again. Since `ArtistFollowButtonQueryRenderer` is a query renderer, it meant a new query at each state update

MounirDhahri: I was able to notice that especially during tests here https://app.circleci.com/pipelines/github/artsy/eigen/71280/workflows/748a7c56-a6ae-4961-a663-727f1f9468f9/jobs/248758

MounirDhahri:
I am excited about tomorrow's release. We are now almost capable to migrate the above the fold query to be `@cacheable` meaning we will very soon have close to no skeletons here.

src/app/Components/ArtistListItemShort.tsx
araujobarret: maybe the better approach would be wrap the original component with React.memo and in the diff condition check only for the internalID changes

src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
```diff
@@ -350,92 +361,155 @@ const ArtworksGrid: React.FC<ArtworksGridProps> = ({
...
+
+const ArtistArtworksPlaceholder = () => {
+  const space = useSpace()
+
+  return (
+    <Tabs.ScrollView
+      scrollEnabled={false}
+      contentContainerStyle={{ flex: 1, paddingHorizontal: 0, paddingTop: space(2) }}
+    >
+      <Flex px={2} testID="ArtistArtworksPlaceholder">
```
araujobarret, line 468: you can use gap prop instead of many Spacers inside

src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
```diff
@@ -266,7 +277,7 @@ const ArtworksGrid: React.FC<ArtworksGridProps> = ({
             iconPosition="right"
           />
         )}
-        <AnimatedMasonryListFooter shouldDisplaySpinner={shouldDisplaySpinner} />
+        <AnimatedMasonryListFooter shouldDisplaySpinner={!!hasNext && isLoadingNext} />
```
araujobarret, line 280: nit: you don't need the !! for hasNext, from the types it can be only a boolean

src/app/Scenes/Artwork/Components/ArtworkMakerTitle.tsx
```diff
@@ -32,6 +37,7 @@ const ArtworkMaker: React.FC<ArtworkMakerProps> = ({ artistName, href }) => {
       disabled={!href}
       to={href}
       onPress={handleArtistTap}
+      navigationProps={{ verifiedRepresentativesCount }}
```
araujobarret, line 40: I see we're sending it via navigation(I don't think it's recommended to send relay data instead of fetching it in the component, but I can't see a way to know ahead of time if they have verified representatives or not), but do you know where we consume it?

MounirDhahri: thanks for raising this - this is a good point, we should not send relay stuff as navigationProps because we don't want to increase coupling between our components. 
**So why did I do it here?** 
In this case, `verifiedRepresentativesCount` is a number that is not required in the Artist screen, if it's there, we can generate a more accurate skeleton, instead of the existent broken skeleton (because of featured artist), if it's not there, we assume the user has no featured representation.

You can see how it was like vs how it became. 

| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/1b550c29-7cb2-4e7a-a19d-0485fbf5421a" />  | <video src="https://github.com/user-attachments/assets/a2ffdde7-39ce-4cdb-8231-cd0f7894c95b" />  | 

It's a subtle change but in this case, I feel like the devil is in the details, I think this helps us further nail the experience

src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
MounirDhahri: true, I will update

MounirDhahri: oh it's autosave that added this not me 😄  it will be added by the next person who touches this as well so I think it's fine to keep it
https://github.com/user-attachments/assets/c12893c9-9e33-4b5c-ac58-57901f408f9b

src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
MounirDhahri: good idea

src/app/Components/FadeIn.tsx
MounirDhahri: you are right, I think it would pop up. I didn't see it because i am setting slide to `false`. I will update this. we don't need more of this annoying warning 😄

src/app/Components/ArtistListItemShort.tsx
MounirDhahri: Let me try it out

MounirDhahri: okey it I feel like it sounds dangerous to me to check for the internalID change within the `memo`, it felt like I am adding tech debt. However, I will wrap it with `memo` because I believe that's something we should have done

damassi:
Eigen is looking 🔥 🔥 🔥 !

PR #12488, opened 2025-07-17
chore(DIA-1378): improve fair screen performance

MounirDhahri:
**Context:** https://artsy.slack.com/archives/C02BAQ5K7/p1752761208038099

### Foundations is business critical and it's why we should get these fixes in this release. 
I promise to hold a KS to go through these. 

### Description

This PR brings a bunch of improvements to the Fair screen such as:
- Make TTI faster
  - Went down from 10/11 seconds for non prefetched to under 200ms
- Improve the scroll performance
  - FPS from occasional 0 drops and under 10fps to above 50-60fps
- Fix some UI bugs

### Changes Made
<details><summary>Broken padding on artwork rails</summary>
<p>

Simulator Screenshot - iPhone 16 Pro - 2025-07-18 at 14 29 23

</p>
</details> 

<details><summary>Dropped frames when scrolling</summary>
<p>

<video src="https://github.com/user-attachments/assets/30cee2df-a273-407b-8798-6a063f523f64" />

| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/6796b127-e038-4659-a8c6-5047e0788b97" /> | <video src="https://github.com/user-attachments/assets/8b7988d7-dd63-42d7-a10d-b01648d84aec" /> | 

</p>
</details> 

<details><summary>Multiple non-needed re-renders</summary>
<p>

<video src="https://github.com/user-attachments/assets/3cf44a86-a95d-4281-bd76-f5e41056cf90" />

</p>
</details> 

<details><summary>Screen jump or blank space on switching tab</summary>
<p>

<video src="https://github.com/user-attachments/assets/f0066c10-f6dd-4579-a357-6354d8360354" />

</p>
</details> 

<details><summary>Slow Query ‼️ More than 10 seconds</summary>
<p>

Screenshot 2025-07-18 at 14 34 59

**The TTI is going to be less 300 ms now**

</p>
</details> 

| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/403985f3-d278-488e-ac7e-4f4eef938b51" /> | <video src="https://github.com/user-attachments/assets/6a21851e-992c-4f1c-b945-b15bcb3e0990" /> | 

# Android video coming in a bit 

### PR Checklist

  - [x] **Android**.
  - [x] **iOS**.

### To the reviewers 👀

- improve fair screen performance- mounir

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Components/ArtworkRail/ArtworkRail.tsx
+++ b/src/app/Components/ArtworkRail/ArtworkRail.tsx
@@ -113,7 +113,7 @@ export const ArtworkRail: React.FC<ArtworkRailProps> = memo(
         showsHorizontalScrollIndicator={false}
         onViewableItemsChanged={onViewableItemsChanged}
         viewabilityConfig={viewabilityConfig}
-        windowSize={isTablet() ? 10 : 5}
+        windowSize={isTablet() ? 10 : 6}
       />
     )
   }
--- a/src/app/Components/ArtworkRail/ArtworkRail.tsx
+++ b/src/app/Components/ArtworkRail/ArtworkRail.tsx
@@ -113,7 +113,9 @@ export const ArtworkRail: React.FC<ArtworkRailProps> = memo(
         showsHorizontalScrollIndicator={false}
         onViewableItemsChanged={onViewableItemsChanged}
         viewabilityConfig={viewabilityConfig}
-        windowSize={isTablet() ? 10 : 6}
+        // Because this is a horizontal rail, we don't want to load more than 3 screens before and after
+        // The default 21 (10 before and 10 after) is too much
+        windowSize={isTablet() ? 10 : 5}
       />
     )
   }
--- a/src/app/Components/ArtworkRail/ArtworkRail.tsx
+++ b/src/app/Components/ArtworkRail/ArtworkRail.tsx
@@ -113,6 +113,7 @@ export const ArtworkRail: React.FC<ArtworkRailProps> = memo(
         showsHorizontalScrollIndicator={false}
         onViewableItemsChanged={onViewableItemsChanged}
         viewabilityConfig={viewabilityConfig}
+        windowSize={isTablet() ? 10 : 5}
       />
     )
   }
--- a/src/app/Components/ArtworkRail/ArtworkRail.tsx
+++ b/src/app/Components/ArtworkRail/ArtworkRail.tsx
@@ -91,6 +91,9 @@ export const ArtworkRail: React.FC<ArtworkRailProps> = memo(
       <FlatList
         data={artworks}
         horizontal
+        // This is required to avoid broken virtualization on nested flatlists
+        // See https://artsy.slack.com/archives/C02BAQ5K7/p1752833523972209?thread_ts=1752761208.038099&cid=C02BAQ5K7
+        disableVirtualization
         keyExtractor={(item: Artwork) => item.internalID}
         ListFooterComponent={
           <>
--- a/src/app/Components/ArtworkRail/ArtworkRailCard.tsx
+++ b/src/app/Components/ArtworkRail/ArtworkRailCard.tsx
@@ -17,7 +17,7 @@ import { AnalyticsContextProvider } from "app/system/analytics/AnalyticsContext"
 import { RouterLink } from "app/system/navigation/RouterLink"
 import { useFeatureFlag } from "app/utils/hooks/useFeatureFlag"
 import { ArtworkActionTrackingProps } from "app/utils/track/ArtworkActions"
-import { useRef, useState } from "react"
+import { memo, useRef, useState } from "react"
 import { GestureResponderEvent, PixelRatio, Platform } from "react-native"
 import { graphql, useFragment } from "react-relay"
 import { useTracking } from "react-tracking"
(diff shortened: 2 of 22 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (improve fair screen performance- mounir - MounirDhahri)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against de0dd4b69ceffb017cc539a4c7fadb9c38bece9e
</p>

src/app/Scenes/Fair/__tests__/FairFollowedArtistsRail.tests.tsx
```diff
@@ -201,7 +212,7 @@ describe("FairFollowedArtistsRail", () => {
 
     const viewAllButton = await wrapper.root.findAllByType(RouterLink)
 
-    expect(viewAllButton.length).toBe(2)
+    expect(viewAllButton.length).toBe(3)
```
MounirDhahri, line 215: because now we are using `SectionTitle` and the latter contains a `RouterLink`

src/app/Scenes/Fair/FairOverview.tsx
```diff
@@ -99,10 +117,49 @@ const fragment = graphql`
     marketingCollections(size: 5) {
       __typename
     }
-    filterArtworksConnection(first: 20, input: { includeArtworksByFollowedArtists: true }) {
-      edges {
-        __typename
+    filterArtworksConnectionCounts: filterArtworksConnection(
+      first: 1
```
MounirDhahri, line 121: we only need to know that there is more than 1 artwork. We have no need to query for 20

src/app/Scenes/Fair/Fair.tsx
```diff
@@ -158,8 +174,8 @@ const fragment = graphql`
 `
 
 export const FairScreenQuery = graphql`
-  query FairQuery($fairID: String!) {
-    fair(id: $fairID) @principalField {
+  query FairQuery($fairID: String!) @cacheable {
```
MounirDhahri, line 181: the query can now be cacheable because it doesn't have any use specific logic

src/app/Scenes/Fair/Fair.tsx
```diff
@@ -106,14 +126,14 @@ export const Fair: React.FC<FairProps> = ({ fair }) => {
       >
         <Tabs.Tab name="Overview" label="Overview">
           <Tabs.Lazy>
-            <FairOverview fair={data} />
+            <FairOverviewQueryRenderer fairID={data.internalID} />
```
MounirDhahri, line 133: 3 separate query renderers - each tab doesn't need to wait for the others to show its content

src/app/Scenes/Fair/Fair.tsx
```diff
@@ -34,8 +41,21 @@ interface FairProps {
...
+  useEffect(() => {
+    prefetchQuery({
+      query: fairExhibitorsQuery,
+      variables: { id: data.internalID },
+    })
+
+    prefetchQuery({
+      query: fairArtworksQuery,
+      variables: { id: data.internalID },
+    })
```
MounirDhahri: prefetech the other tabs query so it's ready by the time you open them

src/app/Scenes/Fair/Components/FairExhibitors.tsx
```diff
@@ -58,6 +59,11 @@ const FairExhibitors: React.FC<FairExhibitorsProps> = ({ fair, relay }) => {
           </Box>
         ) : null
       }
+      // We want to limit the number of loaded windows to 2 because the items are pretty heavy
```
MounirDhahri: the active screen plus and minus 1

src/app/Scenes/Fair/Components/FairExhibitorRail.tsx
```diff
@@ -77,7 +84,7 @@ export const FairExhibitorRailFragmentContainer = createFragmentContainer(FairEx
         internalID
         slug
       }
-      artworksConnection(first: 20) {
+      artworksConnection(first: 10) {
```
MounirDhahri: align this with other artwork rails in the home view - anyway here we show browse all / view all button in case the user wants to see more

src/app/Scenes/Fair/Components/FairExhibitorRail.tsx
```diff
@@ -51,10 +52,16 @@ const FairExhibitorRail: React.FC<FairExhibitorRailProps> = ({ show }) => {
             )
           )
         }}
+        onMorePress={() => {
+          if (!viewAllUrl) return
+
+          trackEvent(tracks.tappedShow(show))
+        }}
+        showSaveIcon
```
MounirDhahri, line 70: this was crazy with me! we didn't have a save icon and it's a no brainer

src/app/Scenes/Fair/Components/FairCollections.tsx
```diff
@@ -15,69 +18,89 @@ interface FairCollectionsProps extends BoxProps {
...
-                  width={CARD_WIDTH}
-                  images={images}
-                  title={collection.title}
-                  subtitle={collection.category}
-                />
-              </TouchableWithScale>
-            </RouterLink>
-          )
-        }}
+        renderItem={renderItem}
```
MounirDhahri, line 65: extracted this so it's not computed again at each render

src/app/Components/ArtworkRail/ArtworkRail.tsx
```diff
@@ -91,6 +91,9 @@ export const ArtworkRail: React.FC<ArtworkRailProps> = memo(
       <FlatList
         data={artworks}
         horizontal
+        // This is required to avoid broken virtualization on nested flatlists
+        // See https://artsy.slack.com/archives/C02BAQ5K7/p1752833523972209?thread_ts=1752761208.038099&cid=C02BAQ5K7
+        disableVirtualization
```
MounirDhahri, line 96: this was magical, See video in slack

MounirDhahri: Inspiration from https://youtu.be/5Q3VCgKV0GE?si=emUk9aDfhHUwtY1r&t=1230

araujobarret: I wonder about the issue, react native docs doesn't point out best practices when having 2 nested lists with different directions. I wonder if it's a Flatlist problem per se, if for example, the Flashlist wouldn't have this issue 🤔

src/app/Components/ArtworkRail/ArtworkRailCard.tsx
```diff
@@ -34,127 +34,129 @@ export interface ArtworkRailCardProps
...
-  showPartnerName = false,
-  onPress,
-  SalePriceComponent,
-  showSaveIcon = false,
-  testID,
-  ...restProps
-}) => {
-  const enableContextMenuIOS = useFeatureFlag("AREnableArtworkCardContextMenuIOS")
-  const isIOS = Platform.OS === "ios"
+export const ArtworkRailCard: React.FC<ArtworkRailCardProps> = memo(
```
araujobarret, line 37: I think we need to start observing that more on PRs, https://reactnative.dev/docs/optimizing-flatlist-configuration#use-memo advocates in favor of memoizing the lists items as you did here

src/app/Scenes/Fair/Components/FairCollections.tsx
```diff
@@ -15,69 +18,92 @@ interface FairCollectionsProps extends BoxProps {
   fair: FairCollections_fair$data
 }
 
-export const FairCollections: React.FC<FairCollectionsProps> = ({ fair, ...rest }) => {
+export const FairCollections: React.FC<FairCollectionsProps> = memo(({ fair, ...rest }) => {
```
araujobarret, line 21: Do we really need to memoize the whole component and not only the smaller elements as you did below with `useCallback`?

src/app/Components/ArtworkRail/ArtworkRail.tsx
MounirDhahri: I think there is no issue with it. It's just hard to virtualize things when both are together. I think the same issue is also present with the `Flashlist`. It's tricky to do the virtualization properly because you would need to basically memoize everything. I added a link to a conference talk with the exact minute it's being explained.

araujobarret: After watching the yt video, it makes sense to disable given how many cards we render on a Rail, it's just not needed at all 🍖

MounirDhahri:
Thanks for the review @araujobarret 

> Have a small concern about the number of memo usage, since it's everywhere it's kinda hard to say if some are not necessary. Also about disabling virtualization, did you try after the memos to re-enable to see if the re-renders are still happening?

This is a valid concern, I did actually go step by step adding memos and leveraging "Highlight Component Updates" in RN debugger until nothing started to re-render again. I linked a video about it in the PR description under "Dropped frames when scrolling"
The option is available as so:
Screenshot 2025-07-18 at 15 28 49

> About the memos, a good exercise could be having a FPS or profiler running right now and use it as a baseline, then start removing the memos(not from the Flatlist items, but the bigger memos over the entire Rails elements) to see if they're really needed.

Maybe you started reviewing the PR before I added it but there is as well a video showing the dropped frames using the FPS counter. Overall, the experience became much faster and we are above 50fps now 

**Edit:** I forgot to explain the rationale behind the memos, sorry:
So RN recommends to [memoise item components](https://reactnative.dev/docs/optimizing-flatlist-configuration#use-memo) in order to leverage the virtualization the most. In our case, since the exhibitors tab has a list of show rails, we needed to make each rail a `memo`.

Then why adding the memo for individual artwork cards? 
The same answere kind of here, each rail is a `Flatlist`, to get its virtualization to work properly, we needed to make the items, the artwork cards a `memo`. 

The question here becomes, isn't it too much for the memory to save all of this?
Not really, Virtualization kicks in and saves components, it just doesn't save them properly in a way that makes retrieving them fast and therefore they re-render and lead to performance issues

src/app/Components/ArtworkRail/ArtworkRail.tsx
MounirDhahri: exactly, also we don't have an infinite scroll, so we don't really need it

src/app/Scenes/Fair/Components/FairCollections.tsx
MounirDhahri: We spoke about this further and it's related to the comment below. I will resolve this but please feel free to open again if there is something else

src/app/Components/ArtworkRail/ArtworkRailCard.tsx
MounirDhahri: Hopefully with the new compiler we won't need to, hopefully _very soon_

araujobarret:
About the memos, yeah, since we use these rails as an item of a vertical virtualized list it makes sense to have them memorized all the time since the docs also recommend having the items always memorized 👍

MounirDhahri:
**Update:** We QA'd the new screen over at the sprintly QA and everything looks good overall apart from the browse all button (which I fixed). I will go ahead and merge this and create betas with the fixes so it gets QA'd again by Applause

MounirDhahri:
For future references, in case someone wants to optimise another querying logic screen, these were mostly the steps I took

### Step 1: Diagnosis
All our tabs as well as the header were part of the same query. This query can not be cached because it has personalized contente (an artwork is liked or not)

Screenshot 2025-07-18 at 17 09 34

### Step 2: Separating the queries into multiple ones
By coupling all our queries, the first contentful paint comes only after the longest request is done. We can avoid that by decoupling them. This means that the FCP is as low as when we get the fastest query, in our case the header query
Screenshot 2025-07-18 at 17 11 41

### Step 3: Prefetch the different tabs separately

Prefetch on the overview tab the different tab queries 

### Step 4: Make the exhibitors query faster

The exhibitors tab query is super low - around 10 seconds! We query for all shows for a fair! then for each show we query for X artworks. 

Screenshot 2025-07-18 at 17 14 19

From our learnings in the Home view, we can optimise this by querying for the available shows, then querying for the artworks separately. Meaning we would do two queries, one to get the available shows that has artworks, then separate queries to get the list of artworks for each show

Screenshot 2025-07-18 at 17 16 46

### Step 5: 🍒 `@cacheable` FTW
Remember step 2, we make a first query to get the fair header, we can drastically minimize that by using `@cacheable`. It's in fact perfectly a great candidate for cacheable. Meaning we can almost always guarantee the screen opening in less than `300ms`

In step 4, we separated the `get available exhibitors query`, the latter was the culprit of making the exhibitors tab very slow! we can optimise it as well! Yes, we can cache it now since it has no personalized content
Screenshot 2025-07-18 at 17 21 21
 
Screenshot 2025-07-18 at 17 21 17


### Step 5.5 (doesn't apply to deep links) 
Prefetch all different tabs (instead of just main tab)

## Result


Screenshot 2025-07-18 at 17 33 39

PR #12491, opened 2025-07-18
feat(DIA-1360): add negative signals

araujobarret:
This PR resolves [DIA-1360]

### Description

Add negative signals to the discover daily screen.

Will add tracking in a follow-up PR.

cc @artsy/diamond-devs 

| Platform | Demo |
|---|---|
| Android | <video src="https://github.com/user-attachments/assets/0cf758d4-4b9b-4e71-b42b-6d7a4eafd864" width="400" /> |
| iOS | <video src="https://github.com/user-attachments/assets/233b159e-01a5-4e98-9993-b002b437e11f" width="400" /> |

### PR Checklist

  - [x] **Android**.
  - [x] **iOS**.

### To the reviewers 👀

- feat: negative signals to discover daily(hidden with ff)

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[DIA-1360]: https://artsyproduct.atlassian.net/browse/DIA-1360?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryHeader.tsx
+++ b/src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryHeader.tsx
@@ -1,5 +1,5 @@
 import { ChevronDownIcon, MoreIcon, ShareIcon } from "@artsy/icons/native"
-import { DEFAULT_HIT_SLOP, Flex, Screen, Text, Touchable } from "@artsy/palette-mobile"
+import { DEFAULT_HIT_SLOP, Flex, LinkText, Screen, Touchable } from "@artsy/palette-mobile"
 import { getShareURL } from "app/Components/ShareSheet/helpers"
 import { useToast } from "app/Components/Toast/toastHook"
 import { InfiniteDiscoveryArtwork } from "app/Scenes/InfiniteDiscovery/InfiniteDiscovery"
@@ -38,14 +38,9 @@ export const InfiniteDiscoveryHeader: React.FC<InfiniteDiscoveryHeaderProps> = (
           },
           backgroundColor: "green100",
           description: (
-            <Text
-              variant="xs"
-              color="mono0"
-              style={{ textDecorationLine: "underline" }}
-              onPress={() => navigate("/favorites/saves")}
-            >
+            <LinkText variant="xs" color="mono0" onPress={() => navigate("/favorites/saves")}>
               Tap to see all of your saved artworks.
-            </Text>
+            </LinkText>
           ),
           duration: "long",
         }
--- a/src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryNegativeSignals.tsx
+++ b/src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryNegativeSignals.tsx
@@ -14,6 +14,7 @@ import { captureException } from "@sentry/react-native"
 import { InfiniteDiscoveryNegativeSignals_artwork$key } from "__generated__/InfiniteDiscoveryNegativeSignals_artwork.graphql"
 import { getShareURL } from "app/Components/ShareSheet/helpers"
 import { useToast } from "app/Components/Toast/toastHook"
+import { ACCESSIBLE_DEFAULT_ICON_SIZE } from "app/Components/constants"
 import { useExcludeArtistFromDiscovery } from "app/Scenes/InfiniteDiscovery/hooks/useExcludeArtistFromDiscovery"
 import { useInfiniteDiscoveryTracking } from "app/Scenes/InfiniteDiscovery/hooks/useInfiniteDiscoveryTracking"
 import { GlobalStore } from "app/store/GlobalStore"
@@ -163,7 +164,7 @@ export const InfiniteDiscoveryNegativeSignals: FC<InfiniteDiscoveryNegativeSigna
           hitSlop={DEFAULT_HIT_SLOP}
         >
           <Flex flexDirection="row" alignItems="center" gap={1} flexShrink={0}>
-            <ShareIcon width={24} height={24} />
+            <ShareIcon width={ACCESSIBLE_DEFAULT_ICON_SIZE} height={ACCESSIBLE_DEFAULT_ICON_SIZE} />
             <Text>Share Artist</Text>
           </Flex>
         </Touchable>
@@ -176,7 +177,7 @@ export const InfiniteDiscoveryNegativeSignals: FC<InfiniteDiscoveryNegativeSigna
         hitSlop={DEFAULT_HIT_SLOP}
       >
         <Flex flexDirection="row" alignItems="center" gap={1}>
-          <NoArtIcon width={24} height={24} />
+          <NoArtIcon width={ACCESSIBLE_DEFAULT_ICON_SIZE} height={ACCESSIBLE_DEFAULT_ICON_SIZE} />
           <Text>See fewer artworks by this artist</Text>
         </Flex>
       </Touchable>
--- a/src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryNegativeSignalsBottomSheet.tsx
+++ b/src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryNegativeSignalsBottomSheet.tsx
@@ -1,13 +1,15 @@
 import { useColor } from "@artsy/palette-mobile"
 import BottomSheet from "@gorhom/bottom-sheet"
 import { InfiniteDiscoveryNegativeSignalsBottomSheetQuery } from "__generated__/InfiniteDiscoveryNegativeSignalsBottomSheetQuery.graphql"
+import { LoadFailureView } from "app/Components/LoadFailureView"
 import { InfiniteDiscoveryBottomSheetBackdrop } from "app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryBottomSheetBackdrop"
 import {
   InfiniteDiscoveryNegativeSignals,
   InfiniteDiscoveryNegativeSignalsPlaceholder,
 } from "app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryNegativeSignals"
 import { GlobalStore } from "app/store/GlobalStore"
-import { FC, Suspense, useRef } from "react"
+import { withSuspense } from "app/utils/hooks/withSuspense"
+import { FC, useRef } from "react"
 import { graphql, useLazyLoadQuery } from "react-relay"
 
 interface NegativeSignalsBottomSheetProps {
@@ -30,7 +32,7 @@ export const NegativeSignalsBottomSheet: FC<NegativeSignalsBottomSheetProps> = (
 }
 
 export const negativeSignalsQuery = graphql`
-  query InfiniteDiscoveryNegativeSignalsBottomSheetQuery($id: String!) {
+  query InfiniteDiscoveryNegativeSignalsBottomSheetQuery($id: String!) @cacheable {
     artwork(id: $id) {
       ...InfiniteDiscoveryNegativeSignals_artwork
     }
@@ -39,47 +41,53 @@ export const negativeSignalsQuery = graphql`
 
 interface InfiniteDiscoveryNegativeSignalsBottomSheetProps {
   artworkID: string
-  artworkSlug: string
 }
 
-export const InfiniteDiscoveryNegativeSignalsBottomSheet: FC<
-  InfiniteDiscoveryNegativeSignalsBottomSheetProps
-> = ({ artworkID }) => {
-  const color = useColor()
-  const ref = useRef<BottomSheet>(null)
-  const moreInfoSheetVisible = GlobalStore.useAppState(
-    (state) => state.infiniteDiscovery.sessionState.moreInfoSheetVisible
-  )
-  const { setMoreInfoSheetVisible } = GlobalStore.actions.infiniteDiscovery
+export const InfiniteDiscoveryNegativeSignalsBottomSheet =
+  withSuspense<InfiniteDiscoveryNegativeSignalsBottomSheetProps>({
+    Component: ({ artworkID }) => {
+      const color = useColor()
+      const ref = useRef<BottomSheet>(null)
+      const moreInfoSheetVisible = GlobalStore.useAppState(
+        (state) => state.infiniteDiscovery.sessionState.moreInfoSheetVisible
+      )
+      const { setMoreInfoSheetVisible } = GlobalStore.actions.infiniteDiscovery
 
-  const handleOnSheetChange = (index: number) => {
-    if (index === 0 && moreInfoSheetVisible) {
-      setMoreInfoSheetVisible(false)
-    }
-  }
+      const handleOnSheetChange = (index: number) => {
+        if (index === 0 && moreInfoSheetVisible) {
+          setMoreInfoSheetVisible(false)
+        }
+      }
 
-  if (!moreInfoSheetVisible) {
-    return null
-  }
+      if (!moreInfoSheetVisible) {
+        return null
+      }
 
-  return (
-    <BottomSheet
-      ref={ref}
-      enableDynamicSizing={false}
-      enablePanDownToClose={true}
-      snapPoints={[1, 260]}
-      onChange={handleOnSheetChange}
-      index={1}
-      backgroundStyle={{
-        backgroundColor: color("mono0"),
-      }}
-      backdropComponent={(props) => (
-        <InfiniteDiscoveryBottomSheetBackdrop {...props} disappearsOnIndex={0} appearsOnIndex={1} />
-      )}
-    >
-      <Suspense fallback={<InfiniteDiscoveryNegativeSignalsPlaceholder />}>
-        <NegativeSignalsBottomSheet artworkID={artworkID} />
-      </Suspense>
-    </BottomSheet>
-  )
-}
+      return (
+        <BottomSheet
+          ref={ref}
+          enableDynamicSizing={false}
+          enablePanDownToClose={true}
+          snapPoints={[1, 260]}
+          onChange={handleOnSheetChange}
+          index={1}
+          backgroundStyle={{
+            backgroundColor: color("mono0"),
+          }}
+          backdropComponent={(props) => (
+            <InfiniteDiscoveryBottomSheetBackdrop
+              {...props}
+              disappearsOnIndex={0}
+              appearsOnIndex={1}
+            />
+          )}
+        >
+          <NegativeSignalsBottomSheet artworkID={artworkID} />
+        </BottomSheet>
+      )
+    },
+    LoadingFallback: InfiniteDiscoveryNegativeSignalsPlaceholder,
+    ErrorFallback: () => {
+      return <LoadFailureView />
+    },
+  })
--- a/src/app/Scenes/InfiniteDiscovery/InfiniteDiscovery.tsx
+++ b/src/app/Scenes/InfiniteDiscovery/InfiniteDiscovery.tsx
@@ -39,7 +39,6 @@ export const InfiniteDiscovery: React.FC<InfiniteDiscoveryProps> = ({
   const hasInteractedWithOnboarding = GlobalStore.useAppState(
     (state) => state.infiniteDiscovery.hasInteractedWithOnboarding
   )
-  const { setMoreInfoSheetVisible } = GlobalStore.actions.infiniteDiscovery
 
   const [topArtworkId, setTopArtworkId] = useState<string | null>(null)
   const topArtwork = useMemo(
@@ -72,10 +71,6 @@ export const InfiniteDiscovery: React.FC<InfiniteDiscoveryProps> = ({
       })
     }
   }, [artworks])
-
-  useEffect(() => {
-    return () => setMoreInfoSheetVisible(false)
-  }, [setMoreInfoSheetVisible])
 
   // Prefetches the negative signals bottom sheet when top card changes
   useEffect(() => {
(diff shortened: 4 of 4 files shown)


src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryHeader.tsx
```diff
@@ -1,27 +1,130 @@
...
+        if (result.success) {
+          track.share(topArtwork.internalID, topArtwork.slug, result.message)
+        }
+      })
+      .catch((error) => {
+        console.error("InfiniteDiscovery.tsx", error)
+      })
+  }
+
+  const handleOnRightButtonPressed = () => {
```
araujobarret, line 82: this was moved from InfiniteDiscovery element, the only diff here is the feature flag checks to trigger share/show more actions

ArtsyOpenSource:

- Cross-platform user-facing changes (feat: negative signals to discover daily(hidden with ff) - araujobarret)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 9f66996fbdf1afb277630f50fb7dfeadd06c4d08
</p>

src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryHeader.tsx
```diff
@@ -1,27 +1,130 @@
...
+
+export const InfiniteDiscoveryHeader: React.FC<InfiniteDiscoveryHeaderProps> = ({ topArtwork }) => {
+  const negativeSignalsEnabled = useFeatureFlag("AREnabledDiscoverDailyNegativeSignals")
+  const toast = useToast()
+  const track = useInfiniteDiscoveryTracking()
+  const savedArtworksCount = GlobalStore.useAppState(
+    (state) => state.infiniteDiscovery.savedArtworksCount
+  )
+  const { setMoreInfoSheetVisible } = GlobalStore.actions.infiniteDiscovery
+  const hideRightButton = !topArtwork || !topArtwork.slug || !topArtwork.title
```
iskounen, line 26: **suggestion:** Should `InfiniteDiscovery` just not render `InfiniteDiscoveryHeader` unless `!!topArtwork`? That will be consistent with how it renders the bottom sheet and remove a bit of complexity from the codebase.

src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryHeader.tsx
iskounen: **comment:** I didn't have the foresight to put this behind a feature flag. Good call 👍

src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryHeader.tsx
```diff
@@ -1,27 +1,130 @@
...
+            haptic
+          >
+            <ChevronDownIcon />
+          </Touchable>
+        }
+        hideRightElements={hideRightButton}
+        rightElements={
+          <Touchable
+            accessibilityRole="button"
+            accessibilityLabel="More information"
```
iskounen: **nitpick:** It would be nice to render a different label if we disable the feature flag.

src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryHeader.tsx
```diff
@@ -1,27 +1,130 @@
...
+            onPress={handleOnRightButtonPressed}
+            testID="share-icon"
```
iskounen: **nitpick:** We should update this so that the tests make sense.

src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryNegativeSignals.tsx
```diff
@@ -0,0 +1,237 @@
...
+      message: message + "\n" + url,
+      failOnCancel: false,
+    })
+      .then((result) => {
+        if (result.success) {
+          track.share(data.internalID, data.slug, result.message)
+        }
+      })
+      .catch((error) => {
+        console.error("InfiniteDiscovery.tsx", error)
```
iskounen: **nitpick:** It would be good to update this to "InfiniteDiscoveryNegativeSignals.tsx" so we can differentiate it from those coming from older versions (or InfiniteDiscoveryHeader.tsx) if there are any issues.

src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryNegativeSignals.tsx
```diff
@@ -0,0 +1,237 @@
...
+      { text: "No" },
+      {
+        text: "Yes",
+        onPress: () => {
+          data.artists.forEach((artist) => {
+            if (artist) {
+              commitMutation({
+                variables: { input: { artistId: artist.internalID } },
+                onError: () =>
+                  show("Something went wrong, try again", "bottom", { backgroundColor: "red100" }),
```
iskounen: **suggestion:** Should we bubble this up to Sentry?

src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryNegativeSignals.tsx
```diff
@@ -0,0 +1,237 @@
...
+          >
+            <ShareIcon width={24} height={24} />
+          </Touchable>
+        </Flex>
+      </Flex>
+
+      <Touchable
+        accessibilityRole="button"
+        accessibilityLabel="Share artwork"
+        onPress={handleSharePressed}
```
iskounen: **issue:** This button is meant to share an artist, so it'll need a separate `onPress` event from the artwork share (and a different `accessibilityLabel`).

src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryNegativeSignals.tsx
```diff
@@ -0,0 +1,237 @@
...
+      >
+        <Flex flexDirection="row" alignItems="center" gap={1} flexShrink={0}>
+          <ShareIcon width={24} height={24} />
+          <Text>Share Artist</Text>
+        </Flex>
+      </Touchable>
+
+      <Touchable
+        accessibilityRole="button"
+        accessibilityLabel="Share artwork"
```
iskounen: **nitpick:** This should be updated.

src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryNegativeSignals.tsx
```diff
@@ -0,0 +1,237 @@
...
+
+const fragment = graphql`
+  fragment InfiniteDiscoveryNegativeSignals_artwork on Artwork {
+    artistNames @required(action: NONE)
+    date
+    href @required(action: NONE)
+    internalID @required(action: NONE)
+    slug @required(action: NONE)
+    title @required(action: NONE)
+    saleMessage @required(action: NONE)
```
iskounen, line 196: **thought (non-blocking):** We expect each artwork to have a `saleMessage` because the backend has promised to only recommend for-sale artworks, but there's nothing in the artwork schema or query that guarantees they will be for-sale. We've seen artworks that aren't for-sale appear in Discover Daily because of backend bugs in the past, so I'm not sure if we should annotate this.

src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryNegativeSignalsBottomSheet.tsx
```diff
@@ -0,0 +1,85 @@
...
+    {
+      id: artworkID,
+    }
+  )
+
+  if (!data.artwork) {
+    return null
+  }
+
+  return <InfiniteDiscoveryNegativeSignals artwork={data.artwork} />
```
iskounen, line 31: **question:** When I saw this setup, I expected to see a `useFragment` hook in `InfiniteDiscoveryNegativeSignals`, but it doesn't. Is there a reason why these need to be two separate components?

src/app/Scenes/InfiniteDiscovery/hooks/useInfiniteDiscoveryTracking.ts
```diff
@@ -49,6 +49,16 @@ export const useInfiniteDiscoveryTracking = () => {
         mode: "swipe",
       })
     },
+    tappedMore: () => {
+      // TODO: ActionType
```
iskounen, line 53: **question:**

1. Is this going to be done in another PR?
2. Do we need an event for "tapped exclude artist"?

src/app/Scenes/InfiniteDiscovery/Components/__tests__/InfiniteDiscoveryHeader.tests.tsx
```diff
@@ -0,0 +1,248 @@
...
+
+    expect(screen.queryByTestId("share-icon")).not.toBeOnTheScreen()
+  })
+
+  it("hides more button when topArtwork lacks title", () => {
+    const artworkWithoutTitle = { ...mockTopArtwork, title: undefined }
+    renderWithWrappers(<InfiniteDiscoveryHeader topArtwork={artworkWithoutTitle} />)
+
+    expect(screen.queryByTestId("share-icon")).not.toBeOnTheScreen()
+  })
```
iskounen: **suggestion:** I like that you've added test coverage for Discover Daily, and I think the tests for InfiniteDiscoveryNegativeSignals are great, but some of these test cases related to `topArtwork` seem like filler. Eigen's test suite is already slow, so I would remove them.

src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryHeader.tsx
araujobarret: that's not necessary to be honest, these safe guards are just protectors that were in place before, in theory the swiper will never be render with anything null but relay type assume some elements can be null

src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryNegativeSignals.tsx
araujobarret: I don't think so, that's backend's responsibility so the client side shouldn't worry imo, only to trust.

src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryNegativeSignalsBottomSheet.tsx
araujobarret: there's a _useFragment_ in `InfiniteDiscoveryNegativeSignals` 😬 
https://github.com/artsy/eigen/pull/12491/files#diff-2e629b9d160ea8ac096301723ec49bf00c0f677e691e2f8fc817bfefdd94ab0dR32

This component is just the wrapper, the bottom sheet handler, the part that calls the query; the other is the body of the sheet where it uses the artwork information and displays it.

src/app/Scenes/InfiniteDiscovery/hooks/useInfiniteDiscoveryTracking.ts
araujobarret: 1 - yes
2 - I think so, I need to talk to Louis today

src/app/Scenes/InfiniteDiscovery/Components/__tests__/InfiniteDiscoveryHeader.tests.tsx
araujobarret: Good call, I'm chopping the redundant ones. Keeping 4 of them so far 👍

araujobarret:
Thanks for the nice review @iskounen! I've addressed the requested changes here https://github.com/artsy/eigen/pull/12491/commits/d86959091001e0798e3a0430de7d3454c1dfa7f0
I think the share button was a blocker though but fixed in this last commit 👍

src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryHeader.tsx
egdbear: petition to add :comment: to the artsy fork of conventional commits 😅

src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryHeader.tsx
```diff
@@ -1,27 +1,131 @@
...
+          backgroundColor: "green100",
+          description: (
+            <Text
+              variant="xs"
+              color="mono0"
+              style={{ textDecorationLine: "underline" }}
+              onPress={() => navigate("/favorites/saves")}
+            >
+              Tap to see all of your saved artworks.
+            </Text>
```
MounirDhahri: **nitpick:** you could use `LinkText` here

src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryHeader.tsx
```diff
@@ -1,27 +1,131 @@
...
+
+  const handleSharePressed = () => {
+    if (!topArtwork || !topArtwork.slug || !topArtwork.title) {
+      return
+    }
+
+    track.tappedShare(topArtwork.internalID, topArtwork.slug)
+
+    const url = getShareURL(
+      `/artwork/${topArtwork.slug}?utm_content=discover-daily-share&utm_medium=product-share`
```
MounirDhahri, line 63: I was expecting a `utm_source` since that's usually more popular than content

src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryHeader.tsx
```diff
@@ -1,27 +1,131 @@
...
+          <Touchable
+            accessibilityRole="button"
+            accessibilityLabel={rightButtonLabel}
+            onPress={handleOnRightButtonPressed}
+            testID="top-right-icon"
+            hitSlop={DEFAULT_HIT_SLOP}
+            haptic
+          >
+            {negativeSignalsEnabled ? (
+              <MoreIcon width={24} height={24} />
```
MounirDhahri, line 117:
```suggestion
              <MoreIcon width={ACCESSIBLE_DEFAULT_ICON_SIZE} height={ACCESSIBLE_DEFAULT_ICON_SIZE} />
```

src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryNegativeSignals.tsx
```diff
@@ -0,0 +1,252 @@
...
```
MounirDhahri, line 196: The multiple `@required` here look awkward to me. Are you doing this to trick ts for nullability?

MounirDhahri: I am worried about this because I am afraid we end up with more of this pattern all over the app and it's something we should be handling in MP using `new GraphQLNonNull(...)`

src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryNegativeSignals.tsx
```diff
@@ -0,0 +1,252 @@
...
+          >
+            <ShareIcon width={24} height={24} />
+          </Touchable>
+        </Flex>
+      </Flex>
+
+      {data.artists.length === 1 && !!artist && (
+        <Touchable
+          accessibilityRole="button"
+          accessibilityLabel="Share artist"
```
MounirDhahri, line 162: nice 👏

src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryNegativeSignals.tsx
```diff
@@ -0,0 +1,252 @@
...
+  const track = useInfiniteDiscoveryTracking()
+  const { show } = useToast()
+  const { setMoreInfoSheetVisible } = GlobalStore.actions.infiniteDiscovery
+  const [commitMutation, isInFlight] = useExcludeArtistFromDiscovery()
+
+  if (!data) {
+    return null
+  }
+
+  const handleSharePressed = (
```
MounirDhahri, line 45: **nitpick:** we have multiple methods doing the same share artist in the app. It would be great to look into consolidating these in the future into a helper

src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryNegativeSignalsBottomSheet.tsx
```diff
@@ -0,0 +1,85 @@
...
+      backgroundStyle={{
+        backgroundColor: color("mono0"),
+      }}
+      backdropComponent={(props) => (
+        <InfiniteDiscoveryBottomSheetBackdrop {...props} disappearsOnIndex={0} appearsOnIndex={1} />
+      )}
+    >
+      <Suspense fallback={<InfiniteDiscoveryNegativeSignalsPlaceholder />}>
+        <NegativeSignalsBottomSheet artworkID={artworkID} />
+      </Suspense>
```
MounirDhahri: fyi, we have a `withSuspense` wrapper that can be used inside `NegativeSignalsBottomSheet` and it also includes an error boundary that could be useful when it comes to not letting errors bubble up and break the full screen.

src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryNegativeSignalsBottomSheet.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  if (!data.artwork) {
+    return null
+  }
+
+  return <InfiniteDiscoveryNegativeSignals artwork={data.artwork} />
+}
+
+export const negativeSignalsQuery = graphql`
+  query InfiniteDiscoveryNegativeSignalsBottomSheetQuery($id: String!) {
```
MounirDhahri: this query does not define any personalized fields. it should be fine to use `@cacheable` here even if it means that we might be getting slightly stale data - since it would **only** show up in case of someone expressing they are not interested

src/app/Scenes/InfiniteDiscovery/InfiniteDiscovery.tsx
```diff
@@ -74,6 +73,20 @@ export const InfiniteDiscovery: React.FC<InfiniteDiscoveryProps> = ({
     }
   }, [artworks])
 
+  useEffect(() => {
+    return () => setMoreInfoSheetVisible(false)
+  }, [setMoreInfoSheetVisible])
```
MounirDhahri: Why do you need to set this to false on component mount? its default value is already `false` right?

src/app/utils/queryPrefetching.ts
```diff
@@ -67,14 +67,14 @@ const prefetchRoute = async <TQuery extends OperationType>(
   })
 }
 
-export const prefetchQuery = async ({
+export const prefetchQuery = async <T extends Variables>({
```
MounirDhahri, line 70: 👏

MounirDhahri (approved):
Nice PR, just a few comments from my side

src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryNegativeSignals.tsx
araujobarret: I've seen these abstractions creating a mess for the sake of saving a few lines of code, tbh. I prefer repeating ourselves here rather than to create an abstraction for this small share call.

src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryNegativeSignalsBottomSheet.tsx
araujobarret: I started with that and due to some reason I removed, gonna add it back 👍

src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryNegativeSignals.tsx
MounirDhahri: oki - sounds good to me

src/app/Scenes/InfiniteDiscovery/InfiniteDiscovery.tsx
araujobarret: That's on the unmount, actually, it shouldn't be necessary, going to 🧹

src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryHeader.tsx
araujobarret: We have a tracking follow-up, will flag that to Louis, was already like this and idk if they're aware 👍

src/app/Scenes/InfiniteDiscovery/Components/InfiniteDiscoveryNegativeSignals.tsx
araujobarret: I definitely agree we have wrong types on MP, it should come as non-nullable 👍 
On the other hand, this approach is better than the extra `if`s or `?` in our code just to check if a field is null to make TS happy.
This is traceable, and once we have a plan to get rid of all these fake null fields on the MP types, when that happens, it'll be easy to search for `field @required(action: NONE)` and remove that rather than removing the `?`, `if`s, etc.

PR #12518, opened 2025-07-23
chore: [HomeView Performance] do not re-render artwork grid items each second

MounirDhahri:
This PR resolves [] 

### Description

While profiling scroll behaviour on the `HomeView`, I noticed that we had an instance of `ArtworkAuctionTimer` being called each second. After debugging what's going on there, I found out that we are updating the components state at each second. Meaning that if we would have 20 lots shown in the home view, each of those would be updating each second, leading to potential scroll degradation. This is mostly noticeable when a lots rail shows up and a ton of js calls are being made.

**Suggested solution:**
- If a lot is finishing in 1 week, we **don't update** the timer value each second, actually we don't need to update it at all.
- If a lot is closing in 50 minutes, then we can update the timer value, but **each minute** instead of each second (anyway our UI shows minutes left in this case and does not show seconds left)
- If a lot is finishing in under one minute, **keep the same logic** and update each second

This PR fixes an issue with lots re-rendering each second because of a `setInterval`

### PR Checklist

  - [x] **Android**.
  - [x] **iOS**.

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- [HomeView Performance] do not re-render artwork grid items each second - mounir

</details>

### DIFF
--- a/src/app/Components/ArtworkGrids/ArtworkAuctionTimer.tsx
+++ b/src/app/Components/ArtworkGrids/ArtworkAuctionTimer.tsx
@@ -12,8 +12,6 @@ interface ArtworkAuctionTimerProps {
   inRailCard?: boolean
 }
 
-const INTERVAL = 1000
-
 export const ArtworkAuctionTimer: React.FC<ArtworkAuctionTimerProps> = ({
   collectorSignals,
   hideRegisterBySignal,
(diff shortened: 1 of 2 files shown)


src/app/Components/ArtworkGrids/ArtworkAuctionTimer.tsx
```diff
@@ -24,39 +22,75 @@ export const ArtworkAuctionTimer: React.FC<ArtworkAuctionTimerProps> = ({
   const { lotClosesAt, onlineBiddingExtended, registrationEndsAt } = data.auction ?? {}
 
   const lotEndAt = DateTime.fromISO(lotClosesAt ?? "")
-  const intervalId = useRef<ReturnType<typeof setInterval> | null>(null)
-  const [time, setTime] = useState(getTimer(lotClosesAt ?? "")["time"])
-
-  useEffect(() => {
```
MounirDhahri, line 30: I am avoiding this entire logic by moving the `useEffect` to the component itself, meaning we don't need anymore to set an interval in cases where we are not going to show the end time (for example register to bid)

ArtsyOpenSource:

- Dev changes ([HomeView Performance] do not re-render artwork grid items each second - mounir - MounirDhahri)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against debb624ef1db8c3abdedcfee652fea5536ccb442
</p>

src/app/Components/ArtworkGrids/ArtworkAuctionTimer.tsx
```diff
@@ -24,39 +22,75 @@ export const ArtworkAuctionTimer: React.FC<ArtworkAuctionTimerProps> = ({
...
     DateTime.fromISO(registrationEndsAt).diffNow().as("seconds") > 0 &&
     !hideRegisterBySignal
   ) {
-    const formattedRegistrationEndsAt = DateTime.fromISO(registrationEndsAt).toFormat("MMM d")
-
     return (
-      <Text lineHeight={lineHeight} variant="xs" numberOfLines={1} color="mono100">
-        Register by {formattedRegistrationEndsAt}
-      </Text>
+      <AuctionTimerRegisterBy lineHeight={lineHeight} registrationEndsAt={registrationEndsAt} />
```
araujobarret, line 32: Can we add a TODO to remove this lineHeight? It's out of our DS tokens, and we shouldn't be manipulating these line heights individually; it's always linked to the font size through the variant

src/app/Components/ArtworkGrids/ArtworkAuctionTimer.tsx
```diff
@@ -24,39 +22,75 @@ export const ArtworkAuctionTimer: React.FC<ArtworkAuctionTimerProps> = ({
...
+    const UPDATE_INTERVAL = lotEndAt.diffNow().as("minutes") < 5 ? 1000 : 60000
+
+    intervalId.current = setInterval(() => {
+      const { time: timerTime } = getTimer(lotClosesAt ?? "")
+
+      setTime(timerTime)
+    }, UPDATE_INTERVAL)
+
+    return () => {
+      if (intervalId.current) clearInterval(intervalId.current)
```
araujobarret, line 90: nit: you can simplify with:
```suggestion
      clearInterval(intervalId?.current)
```

src/app/Components/ArtworkGrids/ArtworkAuctionTimer.tsx
```diff
@@ -24,39 +22,75 @@ export const ArtworkAuctionTimer: React.FC<ArtworkAuctionTimerProps> = ({
...
+  const [time, setTime] = useState(getTimer(lotClosesAt ?? "")["time"])
+
+  useEffect(() => {
+    // Do not update the timer if the lot is closing in longer than an hour
+    if (lotEndAt.diffNow().as("hours") > 1 && !onlineBiddingExtended) {
+      return
+    }
+
+    // Update the timer every second if the lot is closing in less than 5 minute, otherwise update every minute
+    const UPDATE_INTERVAL = lotEndAt.diffNow().as("minutes") < 5 ? 1000 : 60000
```
araujobarret: nit: maybe camelCase it, usually, these upper-cased constants are declared at top-level

src/app/Scenes/Activity/components/ExpiresInTimer.tsx
```diff
@@ -27,6 +26,22 @@ export const ExpiresInTimer: FC<ExpiresInTimerProps> = ({ item }) => {
   const [hasEnded, setHasEnded] = useState(getTimer(expiresAt)["hasEnded"])
 
   useEffect(() => {
+    // Do not update the timer if the offer is expiring in longer than an hour
+    if (
+      DateTime.fromISO(expiresAt ?? "")
+        .diffNow()
+        .as("hours") > 1
+    ) {
```
araujobarret: I like comments but we can also:
```suggestion
    const timerLongerThan1Hour = DateTime.fromISO(expiresAt ?? "").diffNow()
      .as("hours") >  1
    if (timerLongerThan1Hour) {
```

src/app/Scenes/Activity/components/ExpiresInTimer.tsx
```diff
@@ -27,6 +26,22 @@ export const ExpiresInTimer: FC<ExpiresInTimerProps> = ({ item }) => {
...
+      return
+    }
+
+    const INTERVAL =
```
araujobarret: Ditto here about the var name upper-cased

src/app/Components/ArtworkGrids/ArtworkAuctionTimer.tsx
MounirDhahri: Good idea. I wanted to limit the scope of my changes to keep it easy for reviewers

src/app/Components/ArtworkGrids/ArtworkAuctionTimer.tsx
MounirDhahri: Will do that

src/app/Components/ArtworkGrids/ArtworkAuctionTimer.tsx
MounirDhahri: I wanted to keep this logic as it is but that's a nice idea - I will update

src/app/Scenes/Activity/components/ExpiresInTimer.tsx
MounirDhahri: Hmm, do you feel strong about this? I understand the rationale behind having this as a variable, but my goal here wasn't to describe the variable itself, but the behaviour inside the hook.

MounirDhahri: Let me see, maybe there is a middleground between having a variable and a comment at the same time

araujobarret: No strong feelings, it's just faster to read rather than the comment, very minor imo

src/app/Components/ArtworkGrids/ArtworkAuctionTimer.tsx
MounirDhahri: this wasn't possible although it runs in browser
Screenshot 2025-07-23 at 12 12 14
Screenshot 2025-07-23 at 12 12 39

src/app/Scenes/Activity/components/ExpiresInTimer.tsx
MounirDhahri: I will update this - happy to do it

PR #12576, opened 2025-08-01
fix: bring back missing artworks from collection screen

MounirDhahri:
This PR resolves [ONYX-1862] 

### Description
This PR fixes an issue affecting artwork grids and leading to missing artworks. After some investigation, I believe this is affecting at least 4 other screens but potentially more

### What went wrong
**1. Initial Query**
When a user initially opens an artwork screen we load 10 artworks
**2. Second Query: Pagination**
When the user reaches the end, we load 15 more artworks

The way relay does both 1 and 2 is by sending a query with `first` and `cursor` arguments . 
Inside MP, we convert `first` and `cursor` to gravity supported arguments, meaning `size`, `page` and `offset`.

Let's break down both steps
**1. Inital Query**
- We send to MP `first: 10` then `cursor:null` - meaning the first 10 artworks
- Based on the arguments we sent, we use `convertConnectionArgsToGravityArgs` giving us `page:1` and `size:10` and `offset:0`

**Second QueryL Pagination**
- We send to MP `first: 15` then `cursor:last-artwork-cursor` - meaning the next 15 artworks
- Based on the arguments we sent, we use `convertConnectionArgsToGravityArgs` giving us `page:2` and `size:15` and `offset:15`. **BUT**, becaues offset is 15! we have skipped 5 artworks since offset is computed based on `first`.

**Suggested solution**
Align count on initial query and next queries like we always did before here https://github.com/artsy/eigen/pull/11745. 

simulator_screenshot_BC216E3E-0003-4BF4-A59B-DDA9565698D9

### PR Checklist

  - [x] **Android**.
  - [x] **iOS**.

### To the reviewers 👀

- fix bring back missing artworks from collection screen - mounir

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[ONYX-1862]: https://artsyproduct.atlassian.net/browse/ONYX-1862?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/utils/masonryHelpers/index.tsx
+++ b/src/app/utils/masonryHelpers/index.tsx
@@ -8,7 +8,7 @@ export const NUM_COLUMNS_MASONRY = isTablet() ? 3 : 2
 
 export const ON_END_REACHED_THRESHOLD_MASONRY = 2
 
-export const MASONRY_LIST_PAGE_SIZE = 15
+export const MASONRY_LIST_PAGE_SIZE = 10
 
 export interface masonryRenderItemProps {
   item: MasonryArtworkItem


ArtsyOpenSource:

- Cross-platform user-facing changes (fix bring back missing artworks from collection screen - mounir - MounirDhahri)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 91165f2afa7077fa05b3a177b312bad5dae32de6
</p>

iskounen (approved):
> and offset:15

😱 

Nice catch! Assuming that the artworks in these grids are sorted by merchandisability, we definitely want to be showing numbers 11 - 15 to users.

MrSltun (approved):
Nice catch! 🌟
@MounirDhahri  Should we include this fix in the new release?

MounirDhahri:
@MrSltun yes, I discussed this with @evaschilken and we think it's a launch blocker and worth including in the release

PR #12593, opened 2025-08-05
fix(PBRW-1082): Fix show's Unable to Load

MrSltun:
This PR resolves [PBRW-1082] 

### Description
This PR resolves an unable to load in the Shows screen
It is happening due to the React Native image is getting an `undefined` in `src`, which is causing the screen to break
This PR replaces it with an empty string

### Screenshots
| | Before | After |
|---|---|---|
| Android | image | image |
| iOS | image | image |

### PR Checklist

  - [x] **Android**.
  - [x] **iOS**.

### To the reviewers 👀

- fix Shows screen unable to load when image is not available

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

[PBRW-1082]: https://artsyproduct.atlassian.net/browse/PBRW-1082?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/Scenes/Show/Components/ShowContextCard.tsx
+++ b/src/app/Scenes/Show/Components/ShowContextCard.tsx
@@ -3,6 +3,7 @@ import { Flex, Box, BoxProps, Text, useScreenDimensions, Image } from "@artsy/pa
 import { ShowContextCard_show$data } from "__generated__/ShowContextCard_show.graphql"
 import { SmallCard } from "app/Components/Cards"
 import { SectionTitle } from "app/Components/SectionTitle"
+// eslint-disable-next-line no-restricted-imports
 import { navigate } from "app/system/navigation/navigate"
 import { TouchableOpacity } from "react-native"
 import { createFragmentContainer, graphql } from "react-relay"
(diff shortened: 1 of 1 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (fix Shows screen unable to load when image is not available - MrSltun)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 09230f3c3e19ccef22ab23d1e5f783b1fc49cd21
</p>

dariakoko:
Thanks!

src/app/Scenes/Show/Components/ShowContextCard.tsx
```diff
@@ -99,7 +100,13 @@ const ContextCard: React.FC<ContextCardProps> = ({
...
   ) : (
     <Flex width="100%" borderRadius={4} overflow="hidden">
-      <Image testID="main-image" width={width} aspectRatio={1.5} src={imageUrls[0]} />
+      <Image
+        testID="main-image"
+        width={width}
+        aspectRatio={1.5}
+        // we need to pass an empty string in case imageUrls is empty
+        src={imageUrls[0] || ""}
+      />
```
MounirDhahri, line 109: **nitpick:** the logic above seems to be handling the case of one image or multiple images. I think we don't even need to set `src` as `''` when no image is available.

Maybe something like
```
const imageElement = imageURLs.length === 0 ? null ? hasMultipleImages ? .... // We can use something better this because nested ternaries are hard to read
```

MounirDhahri (commented):
LGTM - just a minor nitpick that isn't blocking

src/app/Scenes/Show/Components/ShowContextCard.tsx
MrSltun: I like your approach, but I would want us to have a placeholder of the image in case it doesn't exist, maybe a switch case would be good here with a gray `Box` instead of `Image` with empty `src`, I can do that in a separate PR

MounirDhahri: take your time with this change - i am happy to review it.

PR #11895, opened 2025-04-11
fix: Do not memoize MasonryInfiniteScrollArtworkGrids onEndReached

olerichter00:
### Description

Do not memoize MasonryInfiniteScrollArtworkGrids onEndReached.

### PR Checklist

- [ ] I have tested my changes on the following platforms:
  - [ ] **Android**.
  - [ ] **iOS**.
- [ ] I hid my changes behind a **[feature flag]**, or they don't need one.
- [ ] I have included **screenshots** or **videos** at least on **Android**, or I have not changed the UI.
- [ ] I have added **tests**, or my changes don't require any.
- [ ] I added an **[app state migration]**, or my changes do not require one.
- [ ] I have documented any **follow-up work** that this PR will require, or it does not require any.
- [ ] I have added a **changelog entry** below, or my changes do not require one.

### To the reviewers 👀

- [ ] I would like **at least one** of the reviewers to **run** this PR on the simulator or device.

<details><summary>Changelog updates</summary>

#### Cross-platform user-facing changes

-

#### iOS user-facing changes

-

#### Android user-facing changes

-

#### Dev changes

-

</details>

Need help with something? Have a look at our [docs], or get in touch with us.

[app state migration]: ../blob/main/docs/adding_state_migrations.md
[feature flag]: ../blob/main/docs/developing_a_feature.md
[docs]: ../blob/main/docs/README.md

PR #12617, opened 2025-08-11
fix(ONYX-1778): pagination is broken on long grids, decrease onEndReachedThreshold

dariakoko:
This PR resolves [ONYX-1778] 

### Description

Pagination on screens with long artwork grids is not working properly - the screen either freezes on a loading state or does not show all available results. It looks like `onEndReached` fails to be called every time when `onEndReachedThreshold` is too high

I noticed that decreasing the `onEndReachedThreshold` value solves the issue without harming the performance - the lists are still rendering fairly quickly, without freezing and showing all available results

We can experiment with the `onEndReachedThreshold`, I notised that anything below 1.6 is an improvement, but I found 1.2 performing best

I’m also wondering if lowering the onEndReachedThreshold is actually the right way to address this issue. The discussion in this PR got me thinking: https://github.com/artsy/eigen/pull/11745 

| Platform | Before | After |
| --- | --- | --- |
| Android | <video src="https://github.com/user-attachments/assets/844725a6-c4fe-4143-8972-aaa7bd7b136f" width="300" /> | <video src="https://github.com/user-attachments/assets/6a2bad98-59e4-4b78-abf2-a4a37fcd1f1c" width="300" /> |
| iOS | <video src="https://github.com/user-attachments/assets/dbe5f878-b4aa-442c-bcd2-19624e11e8e7" width="300" /> | <video src="https://github.com/user-attachments/assets/b4e7e8af-dc64-4927-93ad-6557d8f83c32" width="300" /> |

### PR Checklist

  - [ ] **Android**.
  - [ ] **iOS**.

### To the reviewers 👀

- pagination is broken on long grids, decrease onEndReachedThreshold

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Components/ArtworkGrids/ArtworksFilterHeader.tsx
+++ b/src/app/Components/ArtworkGrids/ArtworksFilterHeader.tsx
@@ -11,7 +11,7 @@ interface FilterHeaderProps {
   showSeparator?: boolean
 }
 
-const HEADER_HEIGHT = 50
+export const HEADER_HEIGHT = 50
 
 export const ArtworksFilterHeader: React.FC<FilterHeaderProps> = ({
   children,
--- a/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
+++ b/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx
@@ -35,6 +35,7 @@ import { convertSavedSearchCriteriaToFilterParams } from "app/Components/Artwork
 import { SearchCriteriaAttributes } from "app/Components/ArtworkFilter/SavedSearch/types"
 import { useArtworkFilters } from "app/Components/ArtworkFilter/useArtworkFilters"
 import ArtworkGridItem from "app/Components/ArtworkGrids/ArtworkGridItem"
+import { HEADER_HEIGHT } from "app/Components/ArtworkGrids/ArtworksFilterHeader"
 import { FilteredArtworkGridZeroState } from "app/Components/ArtworkGrids/FilteredArtworkGridZeroState"
 import { Props as InfiniteScrollGridProps } from "app/Components/ArtworkGrids/InfiniteScrollArtworksGrid"
 import { LoadFailureView, LoadFailureViewProps } from "app/Components/LoadFailureView"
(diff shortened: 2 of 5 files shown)


ArtsyOpenSource:

- Cross-platform user-facing changes (pagination is broken on long grids, decrease onEndReachedThreshold - dariakoko)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against b0ba6923bc86e0bb75c196456cf8ae257b2a8910
</p>

src/app/Components/ArtworkGrids/ArtworksFilterHeader.tsx
```diff
@@ -11,7 +11,7 @@ interface FilterHeaderProps {
   showSeparator?: boolean
 }
 
-const HEADER_HEIGHT = 50
+export const HEADER_HEIGHT = 50
```
brainbicycle, line 14: possibly better to move this to constants.ts if reused?

brainbicycle (approved):
~~i think the first priority is always functioning correctly so if this value fixes the bug with the artwork grids not loading i say go for it and we can try to tweak for performance later. thanks!~~

edit: hold off merging just a bit if alright, reviewing the discussion and seeing if another path forward.

brainbicycle (changes requested):
After reading through the discussion this was the ultimate fix in the past from @olerichter00 : https://github.com/artsy/eigen/pull/11895 

See comments in the PR  but the issue is useCallback in combination with onEndReached, this can cause the onEndReached to be memoized and use the wrong cursor for the fetch. 

I think we can either A) remove useCallback from onEndReached B) add the cursor to the deps array of the useCallback. B might be slightly better for performance, but if too complicated A probably fine. 

Here is where we would want to do that: https://github.com/artsy/eigen/blob/b0ba6923bc86e0bb75c196456cf8ae257b2a8910/src/app/Scenes/Search/SearchArtworksGrid.tsx#L67

dariakoko:
Thanks for the review, @brainbicycle
I tried the options you suggested, but didn’t notice any improvement - Search results list still freezes at some point.
On the ArtistArtworks screen, the issue is more noticeable, and that screen has a different setup: https://github.com/artsy/eigen/blob/main/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx#L174. I also tried fixing it there without lowering onEndReachedThreshold, but no luck. Lowering that value seems to resolve (or at least hide) the issue.

What do you think about merging this PR to improve the browsing experience for now, and then digging deeper into the root cause during Monday’s Mob meeting?

brainbicycle:
> Thanks for the review, @brainbicycle I tried the options you suggested, but didn’t notice any improvement - Search results list still freezes at some point. On the ArtistArtworks screen, the issue is more noticeable, and that screen has a different setup: https://github.com/artsy/eigen/blob/main/src/app/Components/Artist/ArtistArtworks/ArtistArtworks.tsx#L174. I also tried fixing it there without lowering onEndReachedThreshold, but no luck. Lowering that value seems to resolve (or at least hide) the issue.
> 
> What do you think about merging this PR to improve the browsing experience for now, and then digging deeper into the root cause during Monday’s Mob meeting?

that sounds good to me!

PR #12646, opened 2025-08-19
fix(ONYX-1870): fix the screen freezing when adding custom artist 

dariakoko:
This PR resolves [ONYX-1870] 

### Description

The issue could not be reproduced in development mode (localhost), so debugging was not easy. After some digging, I discovered that the error was caused by the use of `Promise.allSettled`, which in release builds on device was throwing "undefined is not a function".

This could be happening because, in release mode, the app may fall back to JavaScriptCore (JSC), and some JSC versions do not support `Promise.allSettled`.

Replacing `Promise.allSettled` with `Promise.all` solved the issue

| Before | After | 
| --- | --- |
| <video src="https://github.com/user-attachments/assets/48a97949-ab4f-430b-8625-2148abfff7e5" width="400" /> | <video src="https://github.com/user-attachments/assets/72924f37-ec37-4f30-83ad-795874e030f0" width="400" /> |

### PR Checklist

  - [ ] **Android**.
  - [x] **iOS**.
- [x  I have added a **changelog entry** below, or my changes do not require one.

### To the reviewers 👀

- fix the screen freezing when adding custom artist 

#### iOS user-facing changes

-

-

#### Dev changes

-

</details>

### DIFF
--- a/src/app/Scenes/MyCollection/hooks/useCreateArtists.ts
+++ b/src/app/Scenes/MyCollection/hooks/useCreateArtists.ts
@@ -13,7 +13,7 @@ export const useCreateArtists = () => {
       return
     }
 
-    const results = await Promise.allSettled(
+    const results = await Promise.all(
       customArtists.map(async (customArtist) => {
         try {
           return await createArtist({


dariakoko:
Could someone with an Android device test it before merging, please 🙏

ArtsyOpenSource:

- Cross-platform user-facing changes (fix the screen freezing when adding custom artist - dariakoko)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against cb7ff9bd142bfa54f83f8ddeb3886ea0fd056e12
</p>

MrSltun:
> Could someone with an Android device test it before merging, please 🙏

@dariakoko I can do that, but is there a specific way to test? Like maybe building a release version on a device and testing the changes?

dariakoko:
Thanks @MrSltun ! Yes, exactly, a release build is required for testing this change

MrSltun:
@dariakoko I can confirm that it works, but I found another bug (in the video), but it's not related to this PR :) 
<video src="https://github.com/user-attachments/assets/a97394b6-9290-46ef-ae0f-8b272d566283" />

MrSltun (approved):

dariakoko:
Thanks for testing @MrSltun!

By the bug on the video you mean the disabled "Add selected artists" CTA? If so, it is expected - we enable the CTA only when selected artists count is greater than 0. So, if you select an artist from the autosuggest list or add one manually (as you did on the video), the CTA becomes active

MrSltun:
@dariakoko I do understand the idea of disabling the button, but on keyboard dismiss, the button is being cut because it's behind the navigation bar on Android

image

dariakoko:
Aaah this is what you mean. Good catch! I will create a ticket for it to onyx board

brainbicycle:
This is a bit confusing, I don't think it is possible for RN to fall back to JSC if we have hermes enabled, it looks like this has been an on and off issue in RN + hermes for some time but we should be on a react native version that support it: https://github.com/facebook/hermes/issues/766
Just not using that method is a valid fix in any case just wondering if it warrants more investigation what is actually going on here.

dariakoko:
I saw someone recommend checking if `Promise.allSettled` exists. And it does exist in dev build, but it does not in release. But it _should_ exists, because we are using hermes, and rn 77, and both support ES2020 features (Promise.allSettled is part of it)
Wondering if it has something to do with the Expo setup on Eigen?

PR #12668, opened 2025-08-22
feat: migrate to use Notifee 

MounirDhahri:
## Description

This PR migrates Eigen to use `notifee` instead of `react-native-push-notification` to handle our push notifications.

### Current state analysis

Our push notifications handling logic is different between iOS and Android, making it tricky to keep the same logic between both. The iOS code for example, has a lot of complexity from the days where we had plenty of native screens, that is no longer required. 

### Changes made

After this PR we will be consolidating many parts of our notifications infra across both platforms. This is how the logic will be now.

We will have a single hook that is shared between both platforms to manage push notifications (`usePushNotifications`). This hook will contain logic to listen to push notification events and handle it.

**Note:**
- The listening logic is different between platforms. We use Google Cloud Messaging Services on Android but Apple Push Notification Services for iOS.

**QA Doc here:** https://www.notion.so/artsy/2025-09-29-Mobile-Feature-QA-Notifee-272cab0764a080b69da1cfbbcffa9616

### PR Checklist

  - [x] **Android**.
  - [x] **iOS**.

### To the reviewers 👀

-

#### iOS user-facing changes

-

-

#### Dev changes

- use notifee - mounir

</details>


MounirDhahri:
@cursor review

src/app/system/notifications/useHandlePushNotifications.ts
```diff
@@ -0,0 +1,128 @@
...
+      navigate(url, {
+        passProps: notificationPayload.current.data,
+        ignoreDebounce: true,
+      })
+
+      // Reset the notification payload after navigation
+      notificationPayload.current = null
+    }
+  }, [isLoggedIn])
+}
```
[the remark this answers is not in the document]
MounirDhahri: very good note!

araujobarret (commented):
I think we need to tweak Expo related config https://rnfirebase.io/messaging/usage#expo

MounirDhahri:
@araujobarret I believe that's only required when you use Expo prebuild. In our case, we handle our entitlements separately here https://github.com/artsy/eigen/blob/9e7622208c0afcbda2700b7b9d56c4878bc089f3/ios/Artsy/App_Resources/Artsy.entitlements and we will probably not be using FCM on iOS

araujobarret:
About the entitlement it's right, but for the remote notifications we need something https://rnfirebase.io/messaging/usage#ios---remote-notification, or maybe we're not using the remote notifications?

ArtsyOpenSource:

- Dev changes (use notifee - mounir - MounirDhahri)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 3321846c1b182d07fb3d51b904023767dc6ba5eb
</p>

MounirDhahri:
> About the entitlement it's right, but for the remote notifications we need something https://rnfirebase.io/messaging/usage#ios---remote-notification, or maybe we're not using the remote notifications?

Yes and no actually, we are using remote notifications but we are not using firebase cloud messaging remote notifications on iOS and instead we are sticking with Apple Push Notifications without changes there

android/app/build.gradle
```diff
@@ -204,10 +204,8 @@ dependencies {
     api("com.squareup.okhttp3:logging-interceptor")
 
     implementation 'com.segment.analytics.android:analytics:4.11.3'
-    implementation platform('com.google.firebase:firebase-bom:28.1.0')
-
-    implementation "com.google.firebase:firebase-analytics:19.0.0" // so we can see push notification reports on firebase
```
brainbicycle, line 209: should this be removed?

android/app/src/main/java/net/artsy/app/MainActivity.kt
```diff
@@ -54,43 +54,6 @@ class MainActivity : ReactActivity() {
...
-                if (source != null && source.equals("appboy", ignoreCase = true)) {
-                    if (intent.hasExtra("uri")) {
-                        val bundle = Bundle()
-                        val uriBundle = Bundle()
-                        val uri = intent.getStringExtra("uri")
-                        uriBundle.putString("url", uri)
-                        bundle.putBundle("data", uriBundle)
-                        return bundle
-                    }
-                }
```
brainbicycle, line 89: not sure if it is still necessary but we should definitely double check braze android notifications and tracking are working after removing this

src/app/Scenes/MyProfile/__tests__/MyProfilePushNotifications.tests.tsx
```diff
@@ -57,6 +66,8 @@ describe(SwitchMenu, () => {
 })
 
 describe("MyProfilePushNotificationsQueryRenderer", () => {
+  // const mockFetchNotificationPermissions = getNotificationPermissionsStatus as jest.Mock
```
brainbicycle, line 69:
```suggestion
```

src/app/App.tsx
```diff
@@ -117,12 +111,6 @@ const Main = () => {
   useListenToThemeChange()
   useTrackAppState()
 
-  useEffect(() => {
-    if (isLoggedIn) {
-      savePendingToken()
-    }
-  }, [isLoggedIn])
```
brainbicycle, line 124: i think the old logic was if you were not logged in but we got a token for you we need to save it and wait for you to log in to send off to gravity, do we not need logic like that anymore?

ios/Podfile
```diff
@@ -121,9 +121,6 @@ target 'Artsy' do
   pod 'MultiDelegate'
   pod 'ObjectiveSugar'
 
-  # Google & Firebase
-  pod 'Firebase/Auth', '10.23.0'
```
brainbicycle, line 125: we don't need for google sign in? maybe we get from node module, but a bit surprised it is not in the Podfile.lock

src/app/system/notifications/__tests__/useAndroidListenToPushNotifications.tests.ts
```diff
@@ -0,0 +1,320 @@
...
+        body: "No channel specified",
+      },
+      android: {
+        smallIcon: "ic_notification",
+        channelId: "Default",
+        badgeCount: undefined,
+      },
+    })
+  })
+})
```
brainbicycle, line 320: ✨

src/app/utils/PushNotification.ts
```diff
@@ -1,325 +1,84 @@
...
     }
+    return false
   }
-}
-
-export const saveToken = (token: string, ignoreSameTokenCheck = false) => {
-  // eslint-disable-next-line no-async-promise-executor
-  return new Promise<boolean>(async (resolve, reject) => {
-    const previousToken = await AsyncStorage.getItem(PUSH_NOTIFICATION_TOKEN)
-    if (token !== previousToken || ignoreSameTokenCheck) {
```
brainbicycle, line 56: the reason for the same token check is because update events can overwhelm gravity unnecessarily, I think we should keep similar logic

src/app/App.tsx
MounirDhahri: Yes, this should still be the case now as well. it's handled in a different location inside `useRegisterForPushNotifications` 

https://github.com/artsy/eigen/blob/be3c336467528b245028511f008da0c1269f546a/src/app/system/notifications/useRegisterForRemoteMessages.ts#L18

ios/Podfile
MounirDhahri: this was pretty surprising to me - it wasn't needed to get google sign in to work.


https://github.com/user-attachments/assets/aecc2e40-2f97-41fc-89ae-09cc08ced560

src/app/utils/PushNotification.ts
MounirDhahri: Thanks for the context - I will bring this back

ios/Podfile
brainbicycle: alright, nice! thanks for checking!

src/app/App.tsx
brainbicycle: aha thanks!

MounirDhahri:
🚀 🚀 🚀 🚀 🚀

iskounen:
@MounirDhahri some of the console.errors that were added in this PR are being noisy - specifically the one in `registerForRemoteMessages` that fires after authentication. Is it cool if I remove the ones that are noisy?

MounirDhahri:
yes please @iskounen - i always wanted to remove those. The android specifc one doesn't even make sense. thank you

